PR #915, opened 2019-10-21
Rename skip-defaults, and add ability to exclude actual defaults

### LINKED ISSUE #523: dict(skip_defaults=True) doesn't apply to nested models
```python
from typing import List
from pydantic import BaseModel, Schema

DIGITAL_INPUT = 'digital_input'


class Payload(BaseModel):
    """Base class for all Combox payloads"""


class PointData(Payload):
    name: str = Schema(..., title='Name', min_length=1, max_length=32)
    gios_type: str = Schema(DIGITAL_INPUT, title='GIOS Type')


class PanelData(Payload):
    name: str = Schema(..., title='Name', min_length=1, max_length=32)
    points: List[PointData] = Schema(..., title='Points', min_length=1)


pd = PanelData.validate({'name': 'Dummy Panel', 'points': [{'name': 'Dummy Point', 'gios_type': DIGITAL_INPUT}]})
assert pd.dict(skip_defaults=True)['points'][0]['gios_type'] == DIGITAL_INPUT
print(pd.dict(skip_defaults=True))
```

```
{'name': 'Dummy Panel', 'points': [{'name': 'Dummy Point', 'gios_type': 'digital_input'}]}
```

Notice how the `dict` contains 'gios_type' field.

dmontagu:
## Change Summary

Rename `skip_defaults` to `exclude_unset`, and add `exclude_defaults` argument.

Still need to add tests and docs, but I wanted to get some feedback on the approach first.

## Related issue number

fix #523

## Checklist

### DIFF
--- a/docs/usage/exporting_models.md
+++ b/docs/usage/exporting_models.md
@@ -10,7 +10,7 @@ Arguments:
 * `include`: fields to include in the returned dictionary; see [below](#advanced-include-exclude)
 * `exclude`: fields to exclude from the returned dictionary; see [below](#advanced-include-exclude)
 * `by_alias`: whether field aliases should be used as keys in the returned dictionary; default `False`
-* `skip_defaults`: whether fields which were not explicitly set when creating the model should
+* `exclude_unset`: whether fields which were not explicitly set when creating the model should
   be excluded from the returned dictionary; default `False`
 
 Example:
@@ -65,7 +65,7 @@ Arguments:
 * `include`: fields to include in the returned dictionary; see [below](#advanced-include-exclude)
 * `exclude`: fields to exclude from the returned dictionary; see [below](#advanced-include-exclude)
 * `by_alias`: whether field aliases should be used as keys in the returned dictionary; default `False`
-* `skip_defaults`: whether fields which were not set when creating the model and have their default values should
+* `exclude_unset`: whether fields which were not set when creating the model and have their default values should
   be excluded from the returned dictionary; default `False`
 * `encoder`: a custom encoder function passed to the `default` argument of `json.dumps()`; defaults to a custom
   encoder designed to take care of all common types
--- a/docs/usage/exporting_models.md
+++ b/docs/usage/exporting_models.md
@@ -12,6 +12,9 @@ Arguments:
 * `by_alias`: whether field aliases should be used as keys in the returned dictionary; default `False`
 * `exclude_unset`: whether fields which were not explicitly set when creating the model should
   be excluded from the returned dictionary; default `False`
+  * Prior to **v1.0**, `exclude_unset` was known as `skip_defaults`; use of `skip_defaults` is now deprecated
+* `exclude_defaults`: whether fields which are equal to their default values (whether set or otherwise) should
+  be excluded from the returned dictionary; default `False`
 
 Example:
 
@@ -66,6 +69,9 @@ Arguments:
 * `exclude`: fields to exclude from the returned dictionary; see [below](#advanced-include-exclude)
 * `by_alias`: whether field aliases should be used as keys in the returned dictionary; default `False`
 * `exclude_unset`: whether fields which were not set when creating the model and have their default values should
+  be excluded from the returned dictionary; default `False`
+  * Prior to **v1.0**, `exclude_unset` was known as `skip_defaults`; use of `skip_defaults` is now deprecated
+* `exclude_defaults`: whether fields which are equal to their default values (whether set or otherwise) should
   be excluded from the returned dictionary; default `False`
 * `encoder`: a custom encoder function passed to the `default` argument of `json.dumps()`; defaults to a custom
   encoder designed to take care of all common types
--- a/docs/usage/exporting_models.md
+++ b/docs/usage/exporting_models.md
@@ -11,8 +11,8 @@ Arguments:
 * `exclude`: fields to exclude from the returned dictionary; see [below](#advanced-include-exclude)
 * `by_alias`: whether field aliases should be used as keys in the returned dictionary; default `False`
 * `exclude_unset`: whether fields which were not explicitly set when creating the model should
-  be excluded from the returned dictionary; default `False`
-  * Prior to **v1.0**, `exclude_unset` was known as `skip_defaults`; use of `skip_defaults` is now deprecated
+  be excluded from the returned dictionary; default `False`.
+  Prior to **v1.0**, `exclude_unset` was known as `skip_defaults`; use of `skip_defaults` is now deprecated
 * `exclude_defaults`: whether fields which are equal to their default values (whether set or otherwise) should
   be excluded from the returned dictionary; default `False`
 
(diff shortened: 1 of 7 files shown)


pydantic/main.py
```diff
@@ -584,34 +602,45 @@ def _iter(
...
 
         value_exclude = ValueItems(self, exclude) if exclude else None
         value_include = ValueItems(self, include) if include else None
 
+        if exclude_defaults:
+            if allowed_keys is None:
+                allowed_keys = set(self.__fields__)
+            for k in allowed_keys:
+                if self.__dict__[k] == self.__fields__[k].default:
+                    allowed_keys.discard(k)
```
dmontagu: @samuelcolvin does the above logic seem reasonable to you for how to perform this? I initially put it in the `for k, v in self.__dict__.items()` loop, but this approach results in fewer checks when `exclude_defaults` is `False` (which I think will be the most common scenario).

Also, this feels like another small nudge toward storing the field defaults in their own dict for fast lookup. Given https://github.com/samuelcolvin/pydantic/pull/898, do you think it could make sense to add that?

I'm fine either way.

samuelcolvin: this looks like it's duplicated below - presumably you meant that to demonstrate both approaches.

This route looks preferable to me.

---

I agree with you on `self.__defaults__` especially if the values are somehow shared between `self.__fields__[k].default` and `self.__defaults__[k]`. 

It would be as simple as adding:

```py
'__defaults__': {n: f.default for n, f in fields.items()},
```
to `new_namespace` in `ModelMetaclass`

dmontagu: I didn't actually mean it in both cases 🤦‍♂️ hard to parse the diffs with all the renames. Thanks for all this feedback.

dmontagu: @samuelcolvin how should this be handled if the field is required? Should I just put the non-required defaults in `__defaults__`? That's what we'd want for #898 too.

docs/usage/exporting_models.md
```diff
@@ -10,7 +10,7 @@ Arguments:
 * `include`: fields to include in the returned dictionary; see [below](#advanced-include-exclude)
 * `exclude`: fields to exclude from the returned dictionary; see [below](#advanced-include-exclude)
 * `by_alias`: whether field aliases should be used as keys in the returned dictionary; default `False`
-* `skip_defaults`: whether fields which were not explicitly set when creating the model should
+* `exclude_unset`: whether fields which were not explicitly set when creating the model should
```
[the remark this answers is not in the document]
dmontagu: Done -- does the added note look good to you?

pydantic/main.py
```diff
@@ -345,8 +358,20 @@ def json(
 
         `encoder` is an optional function to supply as `default` to json.dumps(), other arguments as per `json.dumps()`.
         """
+        if skip_defaults is not None:
+            warnings.warn(
+                f'{type(self).__name__}.dict: "skip_defaults" is deprecated and replaced by "exclude_unset"',
```
```suggestion
                f'{type(self).__name__}.json: "skip_defaults" is deprecated and replaced by "exclude_unset"',
```
?

pydantic/main.py
```diff
@@ -303,15 +303,16 @@ def dict(
         include: Union['SetIntStr', 'DictIntStrAny'] = None,
         exclude: Union['SetIntStr', 'DictIntStrAny'] = None,
         by_alias: bool = False,
-        skip_defaults: bool = False,
+        exclude_unset: bool = False,
+        exclude_defaults: bool = False,
```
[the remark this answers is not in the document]
dmontagu: I changed the default value to None so I can tell if it got set -- that way, you get a warning if you were specifying it as True *or* False (since, if you were specifying as `False`, it would start erroring if/when we removed it.) I think a more robust approach might be to pick up `**kwargs`, but that feels like overkill (and might be computationally expensive too).

I would imagine there are some neat decorators that could be used to perform this, but I'm also fine with the approach implemented here, or with leaving everything defaulting to False and just not warning people if they are passing `skip_defaults=False` somewhere. I leave it up to you

pydantic/main.py
samuelcolvin: yes, `if not f.required` makes sense.

I just tried this on #898 and pycharm complained that `__defaults__` is a special name, see [here](https://docs.python.org/3/reference/datamodel.html). 

Do you think it's okay to call it that? I'm not at all clear whether it might break something.

pydantic/main.py
dmontagu: Does this warning look good to you?

samuelcolvin:
```suggestion
                f'{self.__class__.__name__}.dict(): "skip_defaults" is deprecated and replaced by "exclude_unset"',
```
I would do this, but otherwise looks fine.

pydantic/main.py
dmontagu: I implemented this where `__defaults__` only holds the values for non-required fields. I think it's basically optimally fast when checking to exclude defaults -- you'd have to check them all, and they are cached. This should also simplify the logic for #898.

docs/usage/exporting_models.md
```diff
@@ -10,7 +10,10 @@ Arguments:
...
+  be excluded from the returned dictionary; default `False`
+  * Prior to **v1.0**, `exclude_unset` was known as `skip_defaults`; use of `skip_defaults` is now deprecated
```
samuelcolvin: This isn't working because you need 4 spaces for mkdocs (:gun:), but also I think better to just have one long sentence than a sub-point.

pydantic/main.py
dmontagu: Good catch on the naming -- let's call it `__field_defaults__`? I'll rename it to that, but let me know if you'd prefer something else.

docs/usage/exporting_models.md
```diff
@@ -10,7 +10,10 @@ Arguments:
 * `include`: fields to include in the returned dictionary; see [below](#advanced-include-exclude)
 * `exclude`: fields to exclude from the returned dictionary; see [below](#advanced-include-exclude)
 * `by_alias`: whether field aliases should be used as keys in the returned dictionary; default `False`
-* `skip_defaults`: whether fields which were not explicitly set when creating the model should
+* `exclude_unset`: whether fields which were not explicitly set when creating the model should
+  be excluded from the returned dictionary; default `False`.
+  Prior to **v1.0**, `exclude_unset` was known as `skip_defaults`; use of `skip_defaults` is now deprecated
```
dmontagu: Does this look better?

pydantic/main.py
```diff
@@ -345,8 +358,20 @@ def json(
 
         `encoder` is an optional function to supply as `default` to json.dumps(), other arguments as per `json.dumps()`.
         """
+        if skip_defaults is not None:
+            warnings.warn(
+                f'{self.__class__.__name__}.dict(): "skip_defaults" is deprecated and replaced by "exclude_unset"',
```
```suggestion
                f'{self.__class__.__name__}.json(): "skip_defaults" is deprecated and replaced by "exclude_unset"',
```

samuelcolvin:
need tests for the warning raise by `skip_defaults` and that it does the same thing as `exclude_defaults`

samuelcolvin:
also need a change file.

samuelcolvin:
I think this is the only thing v1 is waiting for. If possible it would be great if this could get finished.

Otherwise hope you don't if I push to this PR to finish it tomorrow.

samuelcolvin:
awesome, thank you.

PR #926, opened 2019-10-23
Clarify documentation and error message about keep_untouched

### LINKED ISSUE #924: Behaviour of keep_untouched is unclear in docs
# Bug

* Python version `import sys; print(sys.version)`: 3.7.3
* Pydantic version `import pydantic; print(pydantic.VERSION)`: 0.32.2

The `keep_untouched` attribute on a model's Config affects the behaviour for properties' **values**, not their **type annotations**. This is unclear, and is also made particularly misleading by the wording of [this error message](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/validators.py#L572):
```
RuntimeError: no validator found for {type_} see `keep_untouched` or `arbitrary_types_allowed` in Config
```
These should both be improved, and perhaps the config attribute's name should also be changed?

retnikt:
Resolves #924.

### DIFF
--- a/changes/926-retnikt.md
+++ b/changes/926-retnikt.md
@@ -1,0 +1,1 @@
+Clarify usage of `remove_untouched`, improve error message for types with no validators.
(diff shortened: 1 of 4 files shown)


dmontagu:
I don’t think it’s strictly necessary, but I do think this could benefit from some practical examples of usage.

PR #812, opened 2019-09-15
Add ability to parse and dump objects as the specified type

dmontagu:
## Change Summary

Adds the capability to parse and dump objects as a specified type.

This can be useful for the sake of dumping a model as though it were one of its superclasses in order to prevent inclusion of subfields, and works even with generic containers.

## Related issue number

#811 (see that issue for more discussion/motivation)

## Checklist

Although the tests currently pass with full coverage, I haven't thought about or added appropriate error messages in casing parsing/dumping fails, (and obviously haven't added any tests for those error messages).

Also, I'm aware that the current tests don't do a good job of stress-testing the capability; I'll add more once I get feedback about whether this might have a chance at getting merged.

* [ ] Unit tests for the changes exist
* [ ] Tests pass on CI and coverage remains at 100%
* [ ] Documentation reflects the changes where applicable
* [ ] `changes/<pull request or issue id>-<github username>.rst` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)

Issue #811, opened 2019-09-15
Serialize as a specified model / field type


dmontagu:
# Feature Request

I would like to be able to serialize/dictify models in a way that ensures the result complies with a specified model schema, even if some of the attributes, or even the model itself, are subclasses of the desired result.

I wouldn't necessarily want to get rid of the current behavior of including all fields, but it would be nice if there was either a boolean flag or a separate function that could be called to ensure the serialized/unstructured output matched the field's schema.

-------
### More detail:

Right now, both dict- and json-generation happen in a "recursive" way that causes models to potentially include extra fields. Consider the following example:

```python
from pydantic import BaseModel

class ModelA(BaseModel):
    integer: int

class ModelB(ModelA):
    string: str

class ModelC(BaseModel):
    a: ModelA

model_b = ModelB(integer=1, string="b")
model_c = ModelC(a=model_b)
print(model_c.dict())  # desired output: {'a': {'integer': 1}}
# {'a': {'integer': 1, 'string': 'b'}}

print(model_c.json())  # desired output: {"a": {"integer": 1}}
# {"a": {"integer": 1, "string": "b"}}
```

It is definitely useful (for both performance and non-performance related reasons) to allow subclass instances to pass validation, but in a context where the goal is to return json that matches the model schema, this can be problematic.

FastAPI has dealt with this by creating a "clone" of the desired response field for which none of the fields / subfields are part of the same class hierarchy. Then, when validating an endpoint response, it is essentially unstructured and reparsed into the "cloned" field, which ensures it only has the fields as described in the model schema. This works, but:

1. It adds a great deal of overhead to the model serialization process, as it means known-to-be-valid models/field values go through three unnecessary steps during serialization: conversion to unstructured format, instantiation of the cloned type(s), and re-parsing of the unstructured data into the cloned type(s).

2. It makes use of a lot of infrastructure, so that one-off serialization/unstructuring-with-a-desired-schema is not really feasible to perform on demand.

-----

### Proposed implementation:
I propose the addition of a function `dump_as_type` with the following signature:
```python
def dump_as_type(obj: T, type_: Type[T]) -> Any: ...
```

This function will take a model, list of models, etc., and a desired type to use while dumping the model the model. The result will be as though a strict instance of `Type` was dumped, removing any fields present in subclasses, and ignoring any field-type overrides.

For more concrete examples, see (the tests in) #812, which also includes a related function `parse_as_type`.

samuelcolvin:
[...]

Issue #908, opened 2019-10-17
Allow dict like types to __root__


knowBalpreet:
Hi @samuelcolvin , Thank you for this amazing library. I am using it with fastapi for defining input models.
```
"preview": {
		"9:16": {
			"thumbnail": "",
			"video": ""
		}, "16:9": {
			"thumbnail": "",
			"video": ""
		}, "1:1": {
			"thumbnail": "",
			"video": ""
		}
	}
```
I have the above data coming in the request for which I want to create a model. I tried implementing wildcard fields but didn't get any success. In the above case, the aspect ratio is dynamic and can be anything but will always be in *:* format. In case that is coming, thumbnail and video are in nested model.

I have implemented this in JSON schema like the following:
```
template_schema = {
    "type": "object",
    "properties": {
        "preview": {
            "minProperties": 1,
            "type": "object",
            "patternProperties": {
               "^[0-9]{1,2}:\d{1,2}$": {
                    "type": "object",
                    "properties": {
                        "thumbnail": {"type": "string"},
                        "video": {"type": "string"}
                    },
                    "required": ["thumbnail", "video"],
                }
            },
        },
    }
}```

Please guide me on how to get it to work? 

Any leads would be appreciated. Thanks in advance!

**Comment by samuelcolvin** [comment]
If you want change the input data to confirm to a more conventional shape, you'd be best of with something like:

``py
class ChildModel(BaseModel):
    thumbnail: str
    video: str

class MainModel(BaseModel):
    __root__: Dict[str, ChildModel]
```
Then you'll need a validator to check the format of the keys.

You'll then need to manually [customise the schema](https://pydantic-docs.helpmanual.io/usage/schema/#schema-customization)

knowBalpreet:
Hi @samuelcolvin, thank you for reverting back so quick. I tried to follow the link and the example shared you above. but wasn't able to achieve the use case as mentioned above. Can you please help me with this? A code example will be a great help. 

Thanks in advance!

knowBalpreet:
Bump!

samuelcolvin:
Sorry, I forgot dictionaries can't be used as the root type.

You need to use a key for the dictionaries of video formats, in your above example that's fine since you have `{"preview: {...}}`.

With that you might want something like:

```py
from typing import Dict
from pydantic import BaseModel, validator
from devtools import debug

class ChildModel(BaseModel):
    thumbnail: str
    video: str

class MainModel(BaseModel):
[...]

Issue #481, opened 2019-04-17
utility methods


samuelcolvin:
See https://github.com/samuelcolvin/pydantic/pull/475#discussion_r276064521, @NoAnyLove's comment:

I think the solution is to add `pydantic.tools`:
* `schema(dataclass_or_model)`
* `to_json(dataclass_or_model)`
* `parse_obj(dataclass_or_model, obj)`
* `parse_file(dataclass_or_model, file)`
* etc.

That would provide a simple way of generating a schema that's unified for dataclasses and models without corrupting (making less vanilla) pydantic dataclasses.

NoAnyLove:
Cool, this idea sounds great.

dmontagu:
I will note that we still may add a `methods` module or similar, and `AnalyzedType` also provides a wrapper exposing similar functionality tied to a specific type. (And has the upshot of encouraging more performant implementations by providing an obvious way to cache the "type analysis", unlike v1's `parse_obj_as`, etc.)

PR #934, opened 2019-10-25
Add parse_as_type function

dmontagu:
## Change Summary

Adds just the `parse_as_type` function from #812, along with some documentation.

The documentation may need some work; @samuelcolvin if you give me some direction on what you'd like I'd be happy to modify it. I also think it may make sense to flesh it out a little more
and/or place it on a separate "tools" page once the `dump_as_type` functionality is implemented.

## Related issue number

Addresses minor parts of #811, but to close that issue the `dump_as_type` functionality will be more important.

Given the lengthy back and forth on #812 I figured it might make more sense to just start fresh and break the pull request into smaller chunks.

## Checklist

### DIFF
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -695,3 +695,30 @@ to get validators to parse and validate the input data.
 {!.tmp_examples/types_custom_type.py!}
 ```
 _(This script is complete, it should run "as is")_
+
+## Parsing data into a specified type
+
+Pydantic includes a utility function `parse_as_type` that can be used to apply pydantic's
+parsing logic in an ad-hoc way.
+
+This is especially useful when you want to parse results into a type that is not a direct subclass of `BaseModel`.
+For example: 
+
+```python
+from typing import List
+
+import requests
+from pydantic import BaseModel, parse_as_type
+
+class Item(BaseModel):
+    id: int
+    name: str
+
+item_data = requests.get('https://my-api.com/items').json()
+# item_data == [{'id': 1, 'name': 'My Item'}, ...]
+
+items = parse_as_type(item_data, List[Item])
+# items == [Item(id=1, name='My Item'), ...]
+```
+
+This function is capable of parsing data into any of the types pydantic can handle as fields of a `BaseModel`.
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -305,11 +305,11 @@ _(This script is complete, it should run "as is")_
 ## Literal Type
 
 !!! note
-    This is not strictly part of the python standard library; 
-    it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
+    This is a new feature of the python standard library as of python 3.8; 
+    prior to python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
 
-*pydantic* supports the use of `typing_extensions.Literal` as a lightweight way to specify that a field
-may accept only specific literal values:
+*pydantic* supports the use of `typing.Literal` (or `typing_extensions.Literal` prior to python 3.8)
+as a lightweight way to specify that a field may accept only specific literal values:
 
 ```py
 {!.tmp_examples/types_literal1.py!}
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -695,30 +695,3 @@ to get validators to parse and validate the input data.
 {!.tmp_examples/types_custom_type.py!}
 ```
 _(This script is complete, it should run "as is")_
-
-## Parsing data into a specified type
-
-Pydantic includes a utility function `parse_as_type` that can be used to apply pydantic's
-parsing logic in an ad-hoc way.
-
-This is especially useful when you want to parse results into a type that is not a direct subclass of `BaseModel`.
-For example: 
-
-```python
-from typing import List
-
-import requests
-from pydantic import BaseModel, parse_as_type
-
-class Item(BaseModel):
-    id: int
-    name: str
-
-item_data = requests.get('https://my-api.com/items').json()
-# item_data == [{'id': 1, 'name': 'My Item'}, ...]
-
-items = parse_as_type(item_data, List[Item])
-# items == [Item(id=1, name='My Item'), ...]
-```
-
-This function is capable of parsing data into any of the types pydantic can handle as fields of a `BaseModel`.
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -1,14 +1,17 @@
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Callable, Type, TypeVar, Union
+from typing import Any, Callable, Optional, Type, TypeVar, Union
 
 from pydantic.parse import Protocol, load_file
 
 from .typing import display_as_type
 
-__all__ = ('parse_obj',)
+__all__ = (
+    'parse_file',
+    'parse_obj',
+)
 
-NameGenerator = Callable[[Type[Any]], str]
+NameFactory = Union[str, Callable[[Type[Any]], str]]
 
 
 def _generate_parsing_type_name(type_: Any) -> str:
@@ -16,7 +19,7 @@ def _generate_parsing_type_name(type_: Any) -> str:
 
 
 @lru_cache(maxsize=2048)
-def _get_parsing_type(type_: Any, type_name: Union[str, NameGenerator] = None) -> Any:
+def _get_parsing_type(type_: Any, type_name: Optional[NameFactory] = None) -> Any:
     from pydantic.main import create_model
 
     if type_name is None:
@@ -29,7 +32,7 @@ def _get_parsing_type(type_: Any, type_name: Union[str, NameGenerator] = None) -
 T = TypeVar('T')
 
 
-def parse_obj(type_: Type[T], obj: Any, type_name: Union[str, NameGenerator] = None) -> T:
+def parse_obj(type_: Type[T], obj: Any, type_name: Optional[NameFactory] = None) -> T:
     model_type = _get_parsing_type(type_, type_name=type_name)
     return model_type(obj=obj).obj
 
@@ -42,7 +45,7 @@ def parse_file(
     encoding: str = 'utf8',
     proto: Protocol = None,
     allow_pickle: bool = False,
-    type_name: Union[str, NameGenerator] = None,
+    type_name: Optional[NameFactory] = None,
 ) -> T:
     obj = load_file(path, proto=proto, content_type=content_type, encoding=encoding, allow_pickle=allow_pickle)
     return parse_obj(type_, obj, type_name)
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -5,7 +5,7 @@ from .typing import display_as_type
 
 __all__ = ('parse_as_type',)
 
-NameGenerator = Callable[[Any], str]
+NameGenerator = Callable[[Type[Any]], str]
 
 
 def _generate_parsing_type_name(type_: Any) -> str:
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -1,20 +1,31 @@
 from functools import lru_cache
-from typing import Any, Type, TypeVar
+from typing import Any, Callable, Type, TypeVar, Union
+
+from .typing import display_as_type
 
 __all__ = ('parse_as_type',)
 
+NameGenerator = Callable[[Any], str]
 
-@lru_cache(maxsize=None)
-def _get_parsing_type(type_: Any, source: str) -> Any:
+
+def _generate_parsing_type_name(type_: Any) -> str:
+    return f'ParsingModel[{display_as_type(type_)}]'
+
+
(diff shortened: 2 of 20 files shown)


aldanor:
This doesn't seem to work with dataclasses, is it supposed to?

dmontagu:
It should work if and only if you could put dataclasses as fields on a pydantic model.

I don't know whether @samuelcolvin would want to support that or not.

docs/usage/types.md
```diff
@@ -660,3 +660,30 @@ to get validators to parse and validate the input data.
...
+
+## Parsing data into a specified type
+
+Pydantic includes a utility function `parse_as_type` that can be used to apply pydantic's
+parsing logic in an ad-hoc way.
+
+This is especially useful when you want to parse results into a type that is not a direct subclass of `BaseModel`.
+For example: 
+
+```python
```
samuelcolvin: can you make this a file in examples so, then use `print()` below, then print output will automatically be updated.

You'll need to add requests to `docs/requirements.txt` and use a real (very stable) url below. Might be easier to just define a variable `response_data` and add a comment saying it could come from an API call.

pydantic/tools.py
```diff
@@ -0,0 +1,20 @@
+from functools import lru_cache
+from typing import Any, Type, TypeVar
+
+__all__ = ('parse_as_type',)
+
+
+@lru_cache(maxsize=None)
```
samuelcolvin: `maxsize=None` a bit dangerous? Maybe better to just use a large maxsize?

pydantic/tools.py
```diff
@@ -0,0 +1,20 @@
...
+def _get_parsing_type(type_: Any, source: str) -> Any:
+    from pydantic.main import create_model
+
+    type_name = getattr(type_, '__name__', str(type_))
```
samuelcolvin:
```suggestion
    type_name = display_as_type(type_)
```
?

pydantic/tools.py
```diff
@@ -0,0 +1,20 @@
...
+    return create_model(f'ParsingModel[{type_name}] (for {source})', obj=(type_, ...))
```
samuelcolvin: is this name that isn't valid python going to cause confusion or anything to break?

Maybe better to remove ` (for {source})`? We could add `docstring` as an argument to `create_model` and add more information there?

docs/usage/types.md
```diff
@@ -660,3 +660,30 @@ to get validators to parse and validate the input data.
 {!.tmp_examples/custom_data_types.py!}
 ```
+
+## Parsing data into a specified type
```
samuelcolvin: I think this should be in `models.md` or a new usage file.

pydantic/tools.py
dmontagu: I want to use this function when generating API clients, so I would expect it to be used for every possible response model from a given API. Given the usefulness of inheritance when defining slightly different models, hitting 128 different response shapes seems possible even for not-*especially*-large APIs.

That said, I guess in almost all cases it would be fine if only the top 100 or so most commonly used response types were typically cached 😄.

It's easy enough to change later if anyone reports performance issues that could be improved by increasing it. I'll switch it to the default for now, but let me know if you think it should be set higher for any reason (I don't see a reason to set it lower than the default.)

dmontagu: Actually, I think another good application of this function would be request parsing in server-side frameworks. In that case, I think you'd generally want to cache all possible parsing types for speed (given the relatively limited marginal memory requirements). So maybe it does make sense to use a higher number than  128.

Thoughts?

pydantic/tools.py
dmontagu: I don't think the naming will cause anything to break; the generated classes are really an implementation detail and are basically private to this function.

Currently, the only way their existence should even be publicly detectable is if you get a parsing error when calling the function, and even then it will only show up in the traceback where we print the model name.

This was originally added to make it easier to figure out where this `ParsingModel` came from -- if you just saw a validation error for, `ParsingModel[int]`, and hadn't defined `ParsingModel` in your code, I think it could be relatively painful to discover what was going on (and that the error was coming from the call to `parse_as_model`).

But now that I think about it, if your error came from a call to this , you'd ultimately get the `parse_as_model` call in the traceback, and that's probably good enough for discoverability.

------

Thinking about this more though, I think we should probably just make it possible to override the generated name (similar to the recent change to GenericModel), since I could imagine framework writers wanting a little more control over the generated error messages. 

I'll think about if there is an easy way to accomplish this.

docs/usage/types.md
dmontagu: Yeah, I think it definitely makes sense to create a new usage file if/when we add `dump_as_type`, but it was hard for me to justify a new docs file for just this short little section.

I'll move it to `models.md` for now.

pydantic/tools.py
samuelcolvin: I think we should remove ` (for {source})` and instead allow people to set their own names if they wish?

pydantic/tools.py
samuelcolvin: My concern was someone calling this with some dynamic or semi dynamic type, e.g. an end user can directly or indirectly decide what `type_` is, and therefore create a memory leak.

Imagine a spreadsheet like application where the user can decide the types of each column, the application developer then calls `parse_as_type(Tuple[int, str, str, str, datetime], data)` (where the tuple is used to represent the types of the columns)  to validate the data. By changing the types of columns you could in theory fill up memory.

Why don't we set a maximum to 2048? That makes it unlikely the limit will ever be reached but gives a safety net.

pydantic/tools.py
samuelcolvin: Not not sure about `obj` I think it could be confusing, I think we should use custom root types, e.g. `__root__`.

That might involve fixing #908 but I think that's a good idea anyway.

ahirner:
Is this a partial solution for #481 and what is missing?

dmontagu:
> Is this a partial solution for #481 and what is missing?

This is orthogonal to the points raised in that issue, but I think it would be nice if it could support pydantic dataclasses too. I'm not sure how I'd make that happen though. I think if you added the functionality discussed in that issue, it would probably make it easier to have the `parse_as_type` function handle pydantic dataclasses.

ahirner:
i like the tests, especially `test_parse_as_type_preserves_subclasses` and think it would be no problem to add functionality to achieve the same with dataclasses.
so correct me if I'm wrong: `parse_as_type(obj, type(dataclass_or_model)) == parse_obj(dataclass_or_model, obj)` ?

dmontagu:
> so correct me if I'm wrong: `parse_as_type(obj, dataclass_or_model_type) == parse_obj(dataclass_or_model_type, obj)` ?

Yeah, I ordered arguments based on what seemed intuitive to me from the function name `parse_as_type` -> parse `obj` as type `type_`), but I am happy to reverse if preferable. The other order would be more consistent with `typing.cast`, so maybe that's better.

dmontagu:
Just waiting on https://github.com/samuelcolvin/pydantic/pull/958 to replace `obj` with `__root__`.

pydantic/tools.py
```diff
@@ -0,0 +1,31 @@
...
+        type_name = _generate_parsing_type_name
+    if not isinstance(type_name, str):
+        type_name = type_name(type_)
+    return create_model(type_name, obj=(type_, ...))
+
+
+T = TypeVar('T')
+
+
+def parse_as_type(type_: Type[T], obj: Any, type_name: Union[str, NameGenerator] = None) -> T:
```
dmontagu: I reordered the parameters for consistency with `typing.cast` and the ordering discussed in #481

dmontagu:
@ahirner Actually, to my surprise, it looks like this works out of the box with dataclasses. (I added a test.)

This will also handle the `parse_file` function.

I *might* hold off on implementing the `to_json()` function because we'll want to add `dump_as_type` from #811, and that may affect the best way to implement the `to_json` function. On the other hand, there *should* be a `to_json` function that doesn't depend on the type you are dumping, so maybe it's fine to add now.

pydantic/tools.py
dmontagu: Also, @samuelcolvin should I rename this to `parse_obj` in line with #481?

pydantic/tools.py
```diff
@@ -0,0 +1,48 @@
...
+        type_name = _generate_parsing_type_name
+    if not isinstance(type_name, str):
+        type_name = type_name(type_)
+    return create_model(type_name, obj=(type_, ...))
+
+
+T = TypeVar('T')
+
+
+def parse_obj(type_: Type[T], obj: Any, type_name: Union[str, NameGenerator] = None) -> T:
```
dmontagu: I went ahead and renamed this to `parse_obj` for consistency with #481 and the `BaseModel` method; let me know if I should change it back.

pydantic/tools.py
```diff
@@ -0,0 +1,48 @@
...
+from pathlib import Path
+from typing import Any, Callable, Type, TypeVar, Union
+
+from pydantic.parse import Protocol, load_file
+
+from .typing import display_as_type
+
+__all__ = ('parse_obj',)
+
+NameGenerator = Callable[[Type[Any]], str]
```
samuelcolvin:
```suggestion
NameFactory = Callable[[Type[Any]], str]
```
Would make more sense.

samuelcolvin: but it's only ever used with a union on str, maybe better to have

```py
NameType = Union[None, str, Callable[[Type[Any]], str]]
```
?

pydantic/tools.py
```diff
@@ -0,0 +1,48 @@
+from functools import lru_cache
+from pathlib import Path
+from typing import Any, Callable, Type, TypeVar, Union
+
+from pydantic.parse import Protocol, load_file
+
+from .typing import display_as_type
+
+__all__ = ('parse_obj',)
```
samuelcolvin:
```suggestion
__all__ = ('parse_obj', 'parse_file')
```

tests/test_tools.py
```diff
@@ -0,0 +1,78 @@
...
+    inputs = {'x': '1'}
+    assert parse_obj(PydanticDataclass, inputs) == PydanticDataclass(1)
+
+
+def test_parse_as_mapping():
+    inputs = {'1': '2'}
+    assert parse_obj(Dict[int, int], inputs) == {1: 2}
+
+
+def test_parse_file(tmpdir):
```
samuelcolvin: Might work better to use `tmp_path` as it implements the modern `Path` interface.

pydantic/tools.py
```diff
@@ -0,0 +1,51 @@
...
+
+def parse_file(
+    type_: Type[T],
+    path: Union[str, Path],
+    *,
+    content_type: str = None,
+    encoding: str = 'utf8',
+    proto: Protocol = None,
+    allow_pickle: bool = False,
+    type_name: Optional[NameFactory] = None,
```
dmontagu: In this function, `type_name` is keyword-only; but in `parse_obj` it is positional or keyword. I think it would be a little nicer if it were the same for both.

@samuelcolvin do you have an opinion of how to handle this?
1. Keyword only for both
2. Positional or keyword for both
3. Keep it like it is now

samuelcolvin: I'm easy, but probably Keyword only for both.

tests/test_tools.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_parse_as_mapping():
+    inputs = {'1': '2'}
+    assert parse_obj(Dict[int, int], inputs) == {1: 2}
+
+
+def test_parse_file(tmp_path):
+    p = tmp_path / 'test.json'
+    p.write_text('{"1": "2"}')
+    assert parse_file(Dict[int, int], str(p)) == {1: 2}
```
samuelcolvin:
```suggestion
    assert parse_file(Dict[int, int], p) == {1: 2}
```

samuelcolvin:
I'm still not sure about `parse_obj` as a name, I understand why you chose it but:
1. It's not really that similar to `BaseModel.parse_obj()`
2. If you're new to pydantic it's less understandable than `parse_as_type`or whatever

I guess I'd prefer `parse_as_type()` and `parse_file_as_type()`. (Or maybe better `parse_object_as()` and `parse_file_as()`). Thoughts?

dmontagu:
I'm fine with that; it seemed similar to me since it is basically like the unbound version of the `BaseModel.parse_obj` classmethod (that doesn't need to have a BaseModel passed as the first argument). But I have no qualms with using different names.

I'd be in favor of `parse_object_as` and `parse_file_as`. I can change it to that if you confirm that's what you want. (If you want to think about it a little longer that's fine too.)

samuelcolvin:
Sorry to be indecisive. Let's go with `parse_object_as` and `parse_file_as`.

dmontagu:
@samuelcolvin I did the renames and added some docs; assuming you have no issues with the docs, I think this is just waiting now on changing `obj` to `__root__`, which requires #950 

(To be clear, I'm fine with this whether we use `__root__` or `obj` though; I don't know if there is a performance implication to using a custom root model for this, but I'm guessing not?.)

PR #941, opened 2019-10-25
Add better support for validator reuse

### LINKED ISSUE #940: Re-use validators?
Is it possible to create a validator and import it everywhere else it's used?

I have validators for certain fields that are present in multiple classes that I'd like to call simply without defining them.

dmontagu:
## Change Summary

Adds a few things:
* Add `allow_reuse` argument to `validator` and `root_validator` indicating that the validator should not be tracked for accidental name collisions.
* Always allow reuse of validators defined at the module level, since those are not likely to be accidents
* Support directly passing classmethods to `root_validator` and `validator` so that you can declare a reused validator/root validator as a classmethod prior to passing it to the decorators.

(I still need to add documentation.)

## Related issue number

fix #940

## Checklist

### DIFF
--- a/changes/940-dmontagu.md
+++ b/changes/940-dmontagu.md
@@ -1,0 +1,1 @@
+Add better support for reusable validators
--- a/changes/940-dmontagu.md
+++ b/changes/940-dmontagu.md
@@ -1,1 +1,1 @@
-Add better support for reusable validators
+Add `allow_reuse` argument to validators, thus allowing validator reuse
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -123,10 +123,10 @@ def _check_validator_name(f: AnyCallable, allow_reuse: bool) -> None:
     """
     if in_ipython():  # pragma: no cover
         return
-    if not allow_reuse and '.' in f.__qualname__:
+    if not allow_reuse:
         ref = f.__module__ + '.' + f.__qualname__
         if ref in _FUNCS:
-            raise ConfigError(f'duplicate validator function "{ref}"')
+            raise ConfigError(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
         _FUNCS.add(ref)
 
 
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -121,7 +121,7 @@ def _check_validator_name(f: AnyCallable, allow_reuse: bool) -> None:
     does not check for duplicated names in functions defined at the module-global level
     since such validators are likely intended for reuse
     """
-    if in_ipython():  # pragma: no branch
+    if in_ipython():  # pragma: no cover
         return
     if not allow_reuse and '.' in f.__qualname__:
         ref = f.__module__ + '.' + f.__qualname__
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -80,13 +80,14 @@ def validator(
         each_item = not whole
 
     def dec(f: AnyCallable) -> classmethod:
-        f_cls = classmethod(f) if not isinstance(f, classmethod) else f
-        f = f_cls.__func__
-        _check_validator_name(f, allow_reuse)
+        f_cls = _prepare_validator(f, allow_reuse)
         setattr(
             f_cls,
             VALIDATOR_CONFIG_KEY,
-            (fields, Validator(func=f, pre=pre, each_item=each_item, always=always, check_fields=check_fields)),
+            (
+                fields,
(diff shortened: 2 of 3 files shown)


tests/test_validators.py
```diff
@@ -906,3 +907,99 @@ def root_validator_child(cls, values):
...
+    if duplication_count > 1:
+        with pytest.raises(ConfigError) as exc_info:
+            declare_with_reused_validators(include_root, allow_1, allow_2, allow_3)
+        assert str(exc_info.value).startswith('duplicate validator function')
+    else:
+        declare_with_reused_validators(include_root, allow_1, allow_2, allow_3)
+
+
+@pytest.mark.parametrize('validator_classmethod,root_validator_classmethod', product(*[[True, False]] * 2))
+def test_root_validator_classmethod(validator_classmethod, root_validator_classmethod, reset_tracked_validators):
```
dmontagu: This is basically copied from above, but adds the `reset_tracked_validators` and parametrized arguments. I could merge this with the test from above, but I wasn't sure if that was desirable since it makes it a little uglier with the conditional decorator logic.

tests/test_validators.py
```diff
@@ -756,7 +757,7 @@ def repeat_b(cls, v):
             return v * 2
 
         @root_validator
-        def root_validator(cls, values):
+        def example_root_validator(cls, values):
```
dmontagu: I noticed the namespace "conflict" (I ran into the issue below when I tried to move the decorator after the function definition). It doesn't need to be changed, but figured it was safe. I can change it back if desired.

tests/test_validators.py
```diff
@@ -906,3 +907,99 @@ def root_validator_child(cls, values):
...
+
+    original_tracked_validators = set(_FUNCS)
+    yield
+    _FUNCS.clear()
+    _FUNCS.update(original_tracked_validators)
+
+
+@pytest.mark.parametrize('include_root,allow_1,allow_2,allow_3', product(*[[True, False]] * 4))
+def test_allow_reuse(include_root, allow_1, allow_2, allow_3, reset_tracked_validators):
+    duplication_count = int(not allow_1) + int(not allow_2) + int(include_root and not allow_3)
```
dmontagu: If the use of `product` here or the `reset_tracked_validators` is more "magical" than you'd like in the tests, I can change it, just let me know.

changes/940-dmontagu.md
```diff
@@ -0,0 +1 @@
+Add better support for reusable validators
```
samuelcolvin:
```suggestion
Add `allow_reuse` argument to validators, thus allowing validator reuse
```

pydantic/class_validators.py
```diff
@@ -91,29 +94,36 @@ def dec(f: AnyCallable) -> classmethod:
...
     avoid validators with duplicated names since without this, validators can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython - see #312
+
+    does not check for duplicated names in functions defined at the module-global level
+    since such validators are likely intended for reuse
     """
-    if not in_ipython():  # pragma: no branch
+    if in_ipython():  # pragma: no cover
+        return
+    if not allow_reuse and '.' in f.__qualname__:
```
samuelcolvin: This could be confusing. We need to explicitly document this behaviour.

dmontagu: @samuelcolvin Should we just remove this and explicitly require the validator to set `allow_reuse=True`, even at the global level? Seems like it would be easier to document and ultimately more intuitive.

dmontagu:
@samuelcolvin if you agree with this implementation let me know and I'll add an appropriate section to the validator docs

pydantic/class_validators.py
```diff
@@ -78,8 +80,9 @@ def validator(
         each_item = not whole
 
     def dec(f: AnyCallable) -> classmethod:
-        _check_validator_name(f)
-        f_cls = classmethod(f)
+        f_cls = classmethod(f) if not isinstance(f, classmethod) else f
```
samuelcolvin:
```suggestion
        f_cls = f if isinstance(f, classmethod) else classmethod(f)
```
No need to use a `not`.

samuelcolvin: Also below. 

But we do the same thing 3 times here, could we move this logic into `_check_validator_name`? Perhaps renamed to 

```py
def _prepare_validator(f: Any, allow_reuse: bool = False) -> classmethod:
    ...
```

skewty:
Will this feature land in 1.2?

samuelcolvin:
yes.

PR #958, opened 2019-11-02
Add support for mapping types as custom root

### LINKED ISSUE #908: Allow dict like types to __root__

dmontagu:
## Change Summary

Modifies `parse_obj` and `MetaModel` to allow mapping types with a custom root.

I needed this to address some of the feedback on #934 

## Related issue number

Closes #908

## Checklist

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -325,7 +325,7 @@ extending a base model with extra fields.
 ## Custom Root Types
 
 Pydantic models can be defined with a custom root type by declaring the `__root__` field.
-The root type can be of any type: list, float, int, etc.
+The root type can be of any type: dict, list, float, int, etc.
 
 The root type is defined via the type hint on the `__root__` field.
 The root value can be passed to model `__init__` via the `__root__` keyword argument or as
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -334,6 +334,25 @@ the first and only argument to `parse_obj`.
 {!.tmp_examples/models_custom_root_field.py!}
 ```
 
+If you call the `parse_obj` method for a model with a custom root type with a *dict* as the first argument,
+the following logic is used:
+
+* If the custom root type is a mapping type (eg., `Dict` or `Mapping`),
+  the argument itself is always validated against the custom root type.
+* For other custom root types, if the dict has precisely one key with the value `__root__`,
+  the corresponding value will be validated against the custom root type.
+* Otherwise, the dict itself is validated against the custom root type.    
+
+This is demonstrated in the following example:
+
+```py
+{!.tmp_examples/models_custom_root_field_parse_obj.py!}
+```
+
+!!! warning
+    Calling the `parse_obj` method on a dict with the single key `"__root__"` for non-mapping custom root types
+    is currently supported for backwards compatibility, but is not recommended and may be dropped in a future version.
+
 ## Faux Immutability
 
 Models can be configured to be immutable via `allow_mutation = False`. When this is set, attempting to change the
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -324,11 +324,10 @@ extending a base model with extra fields.
 
 ## Custom Root Types
 
-Pydantic models can be defined with a custom root type by declaring the `__root__` field.
-The root type can be of any type: dict, list, float, int, etc.
+Pydantic models can be defined with a custom root type by declaring the `__root__` field. 
 
-The root type is defined via the type hint on the `__root__` field.
-The root value can be passed to model `__init__` via the `__root__` keyword argument or as
+The root type can be any type supported by pydantic, and is specified by the type hint on the `__root__` field.
+The root value can be passed to the model `__init__` via the `__root__` keyword argument, or as
 the first and only argument to `parse_obj`.
 
 ```py
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -324,8 +324,8 @@ extending a base model with extra fields.
 
 ## Custom Root Types
 
-Pydantic models which do not represent a `dict` ("object" in JSON parlance) can have a custom
-root type defined via the `__root__` field. The root type can be of any type: list, float, int, etc.
+Pydantic models can be defined with a custom root type by declaring the `__root__` field.
+The root type can be of any type: list, float, int, etc.
 
 The root type is defined via the type hint on the `__root__` field.
 The root value can be passed to model `__init__` via the `__root__` keyword argument or as
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -384,7 +384,7 @@ class BaseModel(metaclass=ModelMetaclass):
             not (isinstance(obj, dict) and obj.keys() == {ROOT_KEY}) or cls.__fields__[ROOT_KEY].shape == SHAPE_MAPPING
         ):
             obj = {ROOT_KEY: obj}
-        if not isinstance(obj, dict):
+        elif not isinstance(obj, dict):
             try:
                 obj = dict(obj)
             except (TypeError, ValueError) as e:
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -136,8 +136,6 @@ def is_valid_field(name: str) -> bool:
 def validate_custom_root_type(fields: Dict[str, ModelField]) -> None:
     if len(fields) > 1:
         raise ValueError('__root__ cannot be mixed with other fields')
-    if fields[ROOT_KEY].shape == SHAPE_MAPPING:
-        raise TypeError('custom root type cannot allow mapping')
 
 
 UNTOUCHED_TYPES = FunctionType, property, type, classmethod, staticmethod
@@ -240,7 +238,15 @@ class ModelMetaclass(ABCMeta):
             '__custom_root_type__': _custom_root_type,
             **{n: v for n, v in namespace.items() if n not in fields},
         }
-        return super().__new__(mcs, name, bases, new_namespace, **kwargs)
+        cls = super().__new__(mcs, name, bases, new_namespace, **kwargs)
+        if _custom_root_type and cls.__fields__[ROOT_KEY].shape == SHAPE_MAPPING:
+            original_parse_obj = cls.parse_obj
+
+            def parse_obj(obj: Any) -> 'Model':
+                return original_parse_obj({ROOT_KEY: obj})
+
+            setattr(cls, 'parse_obj', staticmethod(parse_obj))
+        return cls
 
 
 class BaseModel(metaclass=ModelMetaclass):
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Ty
 from .class_validators import ROOT_KEY, ValidatorGroup, extract_root_validators, extract_validators, inherit_validators
 from .error_wrappers import ErrorWrapper, ValidationError
 from .errors import ConfigError, DictError, ExtraError, MissingError
-from .fields import SHAPE_MAPPING, ModelField
+from .fields import ModelField
 from .json import custom_pydantic_encoder, pydantic_encoder
 from .parse import Protocol, load_file, load_str_bytes
 from .schema import model_schema
@@ -238,15 +238,7 @@ class ModelMetaclass(ABCMeta):
             '__custom_root_type__': _custom_root_type,
             **{n: v for n, v in namespace.items() if n not in fields},
         }
-        cls = super().__new__(mcs, name, bases, new_namespace, **kwargs)
-        if _custom_root_type and cls.__fields__[ROOT_KEY].shape == SHAPE_MAPPING:
-            original_parse_obj = cls.parse_obj
-
-            def parse_obj(obj: Any) -> 'Model':
-                return original_parse_obj({ROOT_KEY: obj})
-
-            setattr(cls, 'parse_obj', staticmethod(parse_obj))
-        return cls
+        return super().__new__(mcs, name, bases, new_namespace, **kwargs)
 
 
 class BaseModel(metaclass=ModelMetaclass):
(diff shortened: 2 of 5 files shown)


pydantic/main.py
```diff
@@ -379,14 +394,11 @@ def json(
...
-                    obj = dict(obj)
-                except (TypeError, ValueError) as e:
-                    exc = TypeError(f'{cls.__name__} expected dict not {type(obj).__name__}')
-                    raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
+            try:
+                obj = dict(obj)
+            except (TypeError, ValueError) as e:
+                exc = TypeError(f'{cls.__name__} expected dict not {type(obj).__name__}')
+                raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
         return cls(**obj)
```
dmontagu: We could instead use
```python
    def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
        if cls.__custom_root_type__:
            obj = {ROOT_KEY: obj}
        if not isinstance(obj, dict):
            try:
                obj = dict(obj)
            except (TypeError, ValueError) as e:
                exc = TypeError(f'{cls.__name__} expected dict not {type(obj).__name__}')
                raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
```
and drop the `_setup_custom_root_type` function, but this would have a breaking change: you'd have to stop specifying `__root__` as a key in `parse_obj`.

I personally think this would be a preferable api, but it's not a big deal and I'm assuming the breaking change is not an option.

pydantic/main.py
```diff
@@ -246,7 +244,24 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
...
+
+
+def _setup_custom_root_type(cls: Type['Model']) -> None:
+    original_parse_obj = cls.parse_obj
+
+    root_field = cls.__fields__[ROOT_KEY]
+    root_field_is_mapping = root_field.shape == SHAPE_MAPPING
+
+    def parse_obj(obj: Any) -> 'Model':
+        if not isinstance(obj, dict) or root_field_is_mapping:
```
dmontagu: By checking specifically for `root_field_is_mapping`, we ensure there are no changes to code paths that were previously valid.

-----

This check could be done inside the current `parse_obj` function, and that would allow us to drop the `_setup_custom_root_type` function. But the use of this `_setup_custom_root_type` function ensures that models *without* a custom root type are never forced to experience any root-type-checking overhead.

@samuelcolvin I'm expecting you to be opposed to this on the grounds of the complexity it introduces, but figured there was a chance you might like the pattern for the performance benefits. I am happy to unwind it and just put the checks inside the `parse_obj` function where it is defined if that's preferable.

pydantic/main.py
```diff
@@ -246,7 +244,15 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
             '__repr__': Representation.__repr__,
             **{n: v for n, v in namespace.items() if n not in fields},
         }
-        return super().__new__(mcs, name, bases, new_namespace, **kwargs)
+        cls = super().__new__(mcs, name, bases, new_namespace, **kwargs)
+        if _custom_root_type and cls.__fields__[ROOT_KEY].shape == SHAPE_MAPPING:
```
dmontagu: The check for the `SHAPE_MAPPING` could go into the `parse_obj` method definition, but then you'd pay the overhead on every call. 

@samuelcolvin I'm guessing you might find that preferable since it's easier to trace the logic, but I figured I'd propose this approach in case you prefer the performance benefits.

If you like this approach, I think we could actually move all of the `_custom_root_type` checks here instead of in the method definition to reduce the number of checks used when working with non-custom-root-type models, but it quickly makes it harder to follow the logic of the method if there is a custom root type.

Maybe it would be okay if we factored this logic out of the metaclass `__new__` and into a separate `_setup_custom_root` function, and just added a reference to that function in the `parse_obj` docstring?

samuelcolvin: Interesting idea, but I think too confusing to avoid one `if <boolean>`.

I think:
```py
    @classmethod
    def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
        if cls.__custom_root_type__:
            obj = {ROOT_KEY: obj}
        elif not isinstance(obj, dict):
            try:
                obj = dict(obj)
            except (TypeError, ValueError) as e:
                exc = TypeError(f'{cls.__name__} expected dict not {type(obj).__name__}')
                raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
        return cls(**obj)
```

Would be better.

---

But this (and your solution) lead to a backwards incompatibility:

If you call `parse_obj({'__root__': whatever})` validation would now fail since you'd really be calling 

```py
Model(**{'__root__': {'__root__': whatever}})`
```

I think this confusion is why we introduced the check on root type in the first place. We should have changed this pre-v1, but too late.

Solutions:
* wait for version 2 to make this change :-(
* change the above function definition to 

```py
    @classmethod
    def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
        if cls.__custom_root_type__ and not (isinstance(obj, dict) and obj.keys() == {'__root__'}):
        ...
```
That wouldn't be any slower for the standard case, but would involve some more logic if `__custom_root_type__ = True`. I think that would be best for now.

We might want to leave a note saying we'll remove the `and not (isinstance(obj, dict) and obj.keys() == {'__root__'})` in future.

Thoughts?

dmontagu: I actually think this is a really clean way to handle this; I didn't think about the fact that the check is always being performed now anyway.

I think I'd be in favor of modifying this behavior in v2, but we can worry about that when the time comes.

docs/usage/models.md
```diff
@@ -324,8 +324,8 @@ extending a base model with extra fields.
 
 ## Custom Root Types
 
-Pydantic models which do not represent a `dict` ("object" in JSON parlance) can have a custom
-root type defined via the `__root__` field. The root type can be of any type: list, float, int, etc.
+Pydantic models can be defined with a custom root type by declaring the `__root__` field.
```
dmontagu: @samuelcolvin Do you think we need any more ceremony than this documenting that `__root__` can now be used with mapping types?

samuelcolvin: I think this is fine.

docs/usage/models.md
```diff
@@ -324,8 +324,8 @@ extending a base model with extra fields.
...
+The root type can be of any type: dict, list, float, int, etc.
```
samuelcolvin: I think this list is now confusing since AFAIK custom root types work with everything.

samuelcolvin:
On second thoughts I think we should have both tests and documentation describing the peculiarities of root validation if you provide `{'__root__': 'x'}` vs `{'__root__': 'x', '__boot__': 'y'}` or `{'__foot__': 'x'}` etc.

dmontagu:
> On second thoughts I think we should have both tests and documentation describing the peculiarities of root validation if you provide `{'__root__': 'x'}` vs `{'__root__': 'x', '__boot__': 'y'}` or `{'__foot__': 'x'}` etc.

I thought about this more while attempting to address the feedback, and I think there is a problem here that we need to address: if the `__root__` type is `Dict[str, str]` (or has other value type), and an external input is provided with only a single key `__root__`, you may get unexpected behavior.

Because of this, I think when the root type is a mapping, we should *not* have different behavior depending on the presence of the `__root__` key: we should either always add it prior to parsing, or always require it.

-----

If we had no backwards compatibility requirements, I think it would be best to change the behavior so that we *always* assumed there was no `__root__` key in the input to `parse_obj`, and always inject it if the model has a custom root. So you would always use `__root__` in the `__init__`, and never use it in `parse_obj`. Let me know if I'm missing something, but I can't think of any functionality this would limit. (I think this might be a good change to make in v2?)

However, given the backwards compatibility requirement, we obviously need to continue supporting both approaches for non-mapping root models. But I think since mapping root-models aren't currently supported, we can be more opinionated here (and just document the choice).

-----

* Option 1: Never place the data provided to `parse_obj` inside a new dict with the key `__root__` (for a mapping-type custom root model). (So you'd have to manually create the input with `__root__` as a key.)
  * Pro: We'd have to do it this way if we wanted to be able to call `parse_obj` on the output of `instance.dict()`
  * Con: Seems less ergonomic since you have to special case calls to `parse_obj` when dealing with a `__root__` type

* Option 2: Always place the data provided to `parse_obj` inside a new dict with the key `__root__` (for a mapping-type custom root model).
  * Pro: You can provide inputs directly to `parse_obj` whether there is a custom root or not.
  * Con: Can't call `parse_obj` on `instance.dict()`

-----

I personally would prefer "Option 2" above, since the handling a dump-reparse cycle seems like a substantially rarer requirement than parsing in the first place, so I'd rather prioritize having a clean API for `parse_obj`.

The behavior would also be easy to change if desired by users by just overriding the relevant methods (i.e., `parse_obj`, `dict`, etc.).

We could also (eventually) address the lack of symmetry by introducing a `dump_obj` public method that uses the same logic as `.json()` for handling custom root models, which would produce output that could be passed back to `parse_obj` to produce the same object. (Or adding a kwarg to `dict`, though I think that's not preferable). (This would probably depend on / be related to #1001)

samuelcolvin:
I think right now we should do the thing which maximises backwards comparability, that is:

``py
if cls.__custom_root_type__:
    obj = {ROOT_KEY: obj}
```

This would mean `Model.parse_obj(model.dict())` wouldn't work, but `Model(model.dict())` would work. I think this is fine.

I'm happy to omit the "`and not (isinstance(obj, dict) and obj.keys() == {'__root__'})`" magic for clarity. We should just be clear about this

I think this is the same as your option 2 above, so we're agreeing.

**Comment by dmontagu** [comment]
> I'm happy to omit the "and not (isinstance(obj, dict) and obj.keys() == {'__root__'})" magic for clarity. We should just be clear about this in the documentation.

@samuelcolvin To be clear, is that a v2 change? Otherwise, it could break existing code that relies on this behavior, right?

I'm currently planning to use the following check:
```python
        if cls.__custom_root_type__ and (
            not (isinstance(obj, dict) and obj.keys() == {'__root__'})
            or cls.__fields__["__root__"].shape == SHAPE_MAPPING
        ):
            obj = {ROOT_KEY: obj}
```

samuelcolvin:
Sorry, I was being dumb. You check looks good.

We can change/reconsider before v2.

dmontagu:
> On second thoughts I think we should have both tests and documentation describing the peculiarities of root validation if you provide `{'__root__': 'x'}` vs `{'__root__': 'x', '__boot__': 'y'}` or `{'__foot__': 'x'}` etc.

@samuelcolvin I added some tests and docs discussing this. Let me know if you think anything is wrong / missing.

pydantic/main.py
```diff
@@ -382,15 +380,16 @@ def json(
...
         if not isinstance(obj, dict):
-            if cls.__custom_root_type__:
-                obj = {ROOT_KEY: obj}
-            else:
-                try:
-                    obj = dict(obj)
-                except (TypeError, ValueError) as e:
-                    exc = TypeError(f'{cls.__name__} expected dict not {type(obj).__name__}')
-                    raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
+            try:
```
samuelcolvin: I think this line above and be `elif not isinstance(obj, dict):` for a tiny performance improvement.

tests/test_main.py
```diff
@@ -900,10 +900,24 @@ class MyModel(BaseModel):
...
+        {'loc': ('__root__', '__root__'), 'msg': 'str type expected', 'type': 'type_error.str'}
+    ]
+
+
+def test_parse_obj_non_mapping_root():
+    class MyModel(BaseModel):
+        __root__: List[str]
+
+    assert MyModel.parse_obj(['a']).__root__ == ['a']
+    assert MyModel.parse_obj({'__root__': ['a']}).__root__ == ['a']
```
samuelcolvin:
```suggestion
    assert MyModel.parse_obj({'__root__': ['a']}).__root__ == ['a']
    with pytest.raises(ValidationError, match='...'):
        MyModel.parse_obj({'__not_root__': ['a']})
    with pytest.raises(ValidationError, match='...'):
        MyModel.parse_obj({'__root__': ['a'], 'other': 1})
```

tests/test_main.py
```diff
@@ -918,6 +918,10 @@ class MyModel(BaseModel):
 
     assert MyModel.parse_obj(['a']).__root__ == ['a']
     assert MyModel.parse_obj({'__root__': ['a']}).__root__ == ['a']
+    with pytest.raises(ValidationError, match='...'):
+        MyModel.parse_obj({'__not_root__': ['a']})
+    with pytest.raises(ValidationError, match='...'):
```
samuelcolvin: We need something in the match! :smile: (I was being lazy).

tests/test_main.py
dmontagu: Just made a commit with this change. I'm not sure if the `match='...'` was intended to be used verbatim or you meant the `...` as a placeholder. But it worked verbatim, so I committed it like that; let me know if you want a more precise check.

tests/test_main.py
dmontagu: Haha okay hold on

tests/test_main.py
samuelcolvin: sorry :see_no_evil: I was being lazy. `...` as a regex is pretty relaxed!

tests/test_main.py
dmontagu: Should be better now

tests/test_main.py
dmontagu: I hadn't seen that syntax before, wasn't sure what exactly it meant. But yeah, I get it now.

dmontagu:
I need to go to bed now, feel free to finish this up (and #934 if desired) if you notice any remaining minor issues, otherwise I'll finish it some time tomorrow.

samuelcolvin:
great, thank you so much.

PR #966, opened 2019-11-05
mypy plugin support for dataclasses

koxudaxi:
## Change Summary
The PR supports `dataclasses` on mypy plugin

## Related issue number
https://github.com/samuelcolvin/pydantic/issues/957

## Checklist


tests/mypy/outputs/plugin-success.txt
```diff
@@ -0,0 +1,3 @@
+121: error: Unexpected keyword argument "name" for "AddProject"  [call-arg]
```
dmontagu: Weird -- is the dataclass_class_maker_callback not properly adding fields?

koxudaxi: Sorry, I might make a mistake with the filename.
The result has not used the plugin.
Should I change the name to `fail4.txt`
```
    ('mypy-default.ini', 'fail3.py', 'fail3.txt'),
    ('mypy-default.ini', 'plugin_success.py', 'plugin-success.txt'),
```

pydantic/dataclasses.py
```diff
@@ -105,33 +108,28 @@ def _pydantic_post_init(self: 'DataclassType', *initvars: Any) -> None:
...
+def dataclass(
+    _cls: Optional[AnyType] = None,
+    *,
+    init: bool = True,
+    repr: bool = True,
+    eq: bool = True,
+    order: bool = False,
+    unsafe_hash: bool = False,
+    frozen: bool = False,
+    config: Type['BaseConfig'] = None,
```
koxudaxi: We usually give a `Config` class as a `config` argument.
But, The class is not inherited from `BaseConfig`.
Should we change the type to  `Type`?

tests/mypy/outputs/plugin-success.txt
dmontagu: Ohh I see.

I think your naming is fine -- I originally named it based on whether the file would execute successfully, not whether the plugin would have zero errors.

At any rate, I wonder if there might be a way to get mypy to treat pydantic dataclasses like regular dataclasses if the plugin is NOT installed, and only override the behavior if the plugin is active. That way it won't break mypy for anyone currently using pydantic dataclasses that don't want to deal with plugins.

But I suppose if it isn't possible it would just push more people to use the plugin, which is maybe a good thing anyway.

koxudaxi: I agree.
I have deleted the logic. But, we may keep it to get same result by mypy. I will think about it tomorrow.
https://github.com/samuelcolvin/pydantic/blob/0c18619769e3b4a975dea01f6e7f189a15639863/pydantic/dataclasses.py#L108-L112

dmontagu: Also, the mypy devs are telling me on gitter it's probably not possible to have it resolve to the `dataclasses.dataclass` decorator with the plugin off, but with our custom behavior with the plugin on.

I'll let you know if I hear anything different, but I guess if we want to add introspection of the pydantic-specific `dataclass` decorator, we'll probably just need to force people to use the mypy plugin. It's probably for the best to get more plugin usage anyway.

pydantic/dataclasses.py
samuelcolvin: Yes, I think it should just be `Type`

pydantic/mypy.py
```diff
@@ -85,6 +87,11 @@ def get_method_hook(self, fullname: str) -> Optional[Callable[[MethodContext], T
             return from_orm_callback
         return None
 
+    def get_class_decorator_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
+        if fullname == DATACLASS_FULLNAME:
+            return dataclasses.dataclass_class_maker_callback
```
samuelcolvin: I'm being dumb, how do this add support for the `config` argument?

tests/mypy/outputs/plugin-success.txt
samuelcolvin: sounds good to me.

pydantic/mypy.py
dmontagu: I don't think it does yet; you'd need to use a different callback that called that one. I'm assuming this is just a first step in this direction

samuelcolvin: makes sense. Sorry was being dense.

pydantic/dataclasses.py
koxudaxi: Thank you for your reply.
Mypy rejects `Type`  which is a generic type that needs an argument.
I changed `Type['BaseConfig']` to `Type[Any]`.

pydantic/mypy.py
koxudaxi: @samuelcolvin 
Yes, this is the first step.
This callback checks `dataclasses` behavior like `frozen`.

> I'm being dumb, how do this add support for the config argument?

Mypy checks the config argument by the type-hint `config: Type[Any]`
If we need to inspect additional features then, we should append logic which is @dmontagu  said.

Would you want to check other behavior? 
I see a list that the plugin supports configs for `BaseModel`.
https://github.com/samuelcolvin/pydantic/blob/093474ae27d9e1737d6aa76bd6b75f3918acd2e2/pydantic/mypy.py#L132-L138
However, these configs are not supported by `dataclasses`
Sorry, I'm confusing what should I check.

samuelcolvin: I thought with this that `@dataclass(config=Config)` would fail, but now I see that this inspection is done with standard type hints.

tests/mypy/modules/fail4.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@dataclass(config=Config)
+class AddProject:
+    name: str
+    slug: Optional[str]
+    description: Optional[str]
+
+
+p = AddProject(name='x', slug='y', description='z')
```
samuelcolvin: Is this failing because the plugin is not installed?

Please can we add a test to passes when the plugin is installed.

docs/mypy_plugin.md
```diff
@@ -53,7 +53,12 @@ There are other benefits too! See below for more details.
 #### Respect `Config.orm_mode`
 * If `Config.orm_mode` is `False`, you'll get a mypy error if you try to call `.from_orm()`;
   cf. [ORM mode](usage/models.md#orm-mode-aka-arbitrary-class-instances)
- 
+
+#### Generate a signature for `dataclasses`
+* The [`dataclasses`](usage/dataclasses.md) is decorated class with `@pydantic.dataclasess.dataclass` which is inspected
```
samuelcolvin:
```suggestion
* classes decorated with [`@pydantic.dataclasess.dataclass`](usage/dataclasses.md) are type checked the same as standard python dataclasses
```

docs/mypy_plugin.md
```diff
@@ -53,7 +53,12 @@ There are other benefits too! See below for more details.
...
+ as python [dataclasses](https://docs.python.org/3/library/dataclasses.html)
```
samuelcolvin:
```suggestion
```

docs/mypy_plugin.md
```diff
@@ -53,7 +53,12 @@ There are other benefits too! See below for more details.
 #### Respect `Config.orm_mode`
 * If `Config.orm_mode` is `False`, you'll get a mypy error if you try to call `.from_orm()`;
   cf. [ORM mode](usage/models.md#orm-mode-aka-arbitrary-class-instances)
- 
+
+#### Generate a signature for `dataclasses`
+* The [`dataclasses`](usage/dataclasses.md) is decorated class with `@pydantic.dataclasess.dataclass` which is inspected
+ as python [dataclasses](https://docs.python.org/3/library/dataclasses.html) 
+* The `@pydantic.dataclasess.dataclass` decorator accepts `Config` as a keyword argument which has the same meaning as [Config](usage/model_config.md).
```
samuelcolvin:
```suggestion
* The `@pydantic.dataclasess.dataclass` decorator accepts a `config` keyword argument which has the same meaning as [the `Config` sub-class](usage/model_config.md).
```

tests/mypy/modules/fail4.py
koxudaxi: Sorry, I fixed the test cases.

pydantic/mypy.py
koxudaxi: Thank you. It's clear.

koxudaxi:
@dmontagu 
Would you help me?
I get an unexpected result from the `mypy` test.
This problem is added empty lines on error out. 
Also, The problem is happend only first time after `make clean`.
```
>           assert actual_out == expected_out, actual_out
E           AssertionError: 119: error: Unexpected keyword argument "name" for "AddProject"  [call-arg]
E             
E             119: error: Unexpected keyword argument "slug" for "AddProject"  [call-arg]
E             
E             119: error: Unexpected keyword argument "description" for "AddProject"  [call-arg]
E           assert '119: error: ..."  [call-arg]' == '119: error: ..."  [call-arg]'
E               119: error: Unexpected keyword argument "name" for "AddProject"  [call-arg]
E             - 
E               119: error: Unexpected keyword argument "slug" for "AddProject"  [call-arg]
E             - 
E               119: error: Unexpected keyword argument "description" for "AddProject"  [call-arg]

tests/mypy/test_mypy.py:61: AssertionError

```

dmontagu:
Huh, that's very weird. If it's just whitespace differences, it seems like it's probably an issue with mypy?

If the issue *is* just extra whitespace, I would just apply `actual_output = re.sub(r'\n\s*\n', r'\n', actual_output)` or whatever actually works to remove the blank lines (didn't check if that statement actually works).

koxudaxi:
@dmontagu 

Thank you for your advice.
I tested this command. but, I did not look extra whitespaces.
```shell
mypy --config-file tests/mypy/configs/mypy-default.ini  --show-error-codes tests/mypy/modules/success.py
```
I use this method  to resolve the error. ` actual_output = re.sub(r'\n\s*\n', r'\n', actual_output) `

tests/mypy/modules/success.py
```diff
@@ -113,3 +103,17 @@ class WrapperModel(GenericModel, Generic[T]):
...
     age: int
     first_name: str = Field('John', const=True)
+
+
+class Config:
+    validate_assignment = True
+
+
+@dataclass(config=Config)
+class AddProject:
```
samuelcolvin: this should be removed since it fails and `tests/mypy/outputs/success.txt` should not exist.

tests/mypy/outputs/plugin-success.txt
samuelcolvin: surely this file should not exist. As per the name, mypy should pass on this.

tests/mypy/modules/success.py
koxudaxi: I removed it.

tests/mypy/outputs/plugin-success.txt
koxudaxi: I deleted this file. And, I added these error messages on `fail4.txt` which is used by only mypy test case without the plugin.

tests/mypy/modules/success.py
dmontagu: @samuelcolvin @koxudaxi When I named this file originally, the idea was that `success.py` would run without exceptions, and other files wouldn't necessarily run without exceptions. It didn't mean that mypy wouldn't necessarily complain about things inside it, it was just a signal that we could check that mypy wasn't reporting success when the file wouldn't actually run.

But I'm amenable to other conventions here (especially since this one clearly wasn't obvious).

koxudaxi: @dmontagu 
I'm sorry, I should have discussed it with quoted your comment before I did it.

docs/usage/dataclasses.md
```diff
@@ -22,11 +22,8 @@ created by the standard library `dataclass` decorator.
 `pydantic.dataclasses.dataclass`'s arguments are the same as the standard decorator, except one extra
 keyword argument `config` which has the same meaning as [Config](model_config.md).
 
-!!! note
-    As a side effect of getting pydantic dataclasses to play nicely with mypy, the `config` argument will show
-    as invalid in IDEs and mypy. Use `@dataclass(..., config=Config) # type: ignore` as a workaround. 
-
-    See [python/mypy#6239](https://github.com/python/mypy/issues/6239) for an explanation of this issue.
+!!! warning
+    After v1.2, [The Mypy plugin](mypy.md) must be installed to type check pydantic dataclasses.
```
samuelcolvin:
```suggestion
    After v1.2, [The Mypy plugin](/mypy_plugin.md) must be installed to type check pydantic dataclasses.
```

samuelcolvin:
@dmontagu are you happy with this?

dmontagu:
@koxudaxi It looks to me like the config isn't currently *used* by mypy, it's just allowed to be passed without raising an error -- is that right?

@samuelcolvin This looks good to me as a first release ensuring compatibility with the config argument. Eventually it would be nice to have the config values affect the type checking where relevant (I guess I'm not totally sure it affects anything mypy would detect anyway), but I don't think that needs to block a release.

Looks good to me!

pydantic/dataclasses.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         def __validate__(cls, v: Any) -> 'DataclassType':
             pass
 
+        def __call__(self, *args: Any, **kwargs: Any) -> 'DataclassType':
```
dmontagu: Should this use a TypeVar?

tests/mypy/outputs/plugin-fail-strict.txt
```diff
@@ -1,34 +1,35 @@
-23: error: Unexpected keyword argument "z" for "Model"  [call-arg]
```
dmontagu: We need to come up with a better solution for these diffs :/

pydantic/dataclasses.py
koxudaxi: Did you expect this?
```python
    def __call__(self: T, *args: Any, **kwargs: Any) -> T:
        pass
```
Also, Did you want to replace other `-> 'DataclassType'`  like `__init__` ?

tests/mypy/modules/plugin_success.py
```diff
@@ -104,3 +105,17 @@ class ClassVarModel(BaseModel):
...
 
 ClassVarModel(x=1)
+
+
+class Config:
+    validate_assignment = True
+
+
+@dataclass(config=Config)
+class AddProject:
```
koxudaxi: @dmontagu 
>  It looks to me like the received config isn't currently used by mypy, it's just allowed to be passed without raising an error -- is that right?

I understand it is right because mypy passes the tests without any error.
Mypy can parse a type declaration of `config=`. 
Of course,  mypy don't use the contents of `Config` like `orm_mode` yet.

tests/mypy/outputs/plugin-fail-strict.txt
koxudaxi: Sorry, I could not understand why we should change it.
Extra argument is not allowed here by `config`. Is it a problem? :thinking:

dmontagu: Sorry, this wasn't directed at you, or specific to this PR -- your changes are fine. It would just be good if the diff was easier to make sense of, or if we could more easily test one thing at a time.

Something to fix in a future PR, not this one. 👍

tests/mypy/modules/plugin_success.py
dmontagu: Yeah, I think that's fine for now. We can revisit eventually if we think it matters.

pydantic/dataclasses.py
dmontagu: Yeah, that's what I meant, I think it would be better (assuming that is the right signature, which I think it is?).

I don't think the `__init__` signature needs to change, but maybe the `_validate_dataclass` one should.

(Any other places where `'DataclassType'` is currently used probably should; I believe this is what is done for `BaseModel` -- I don't see why it shouldn't be that way for the dataclasses.)

Also, I would recommend using something like `DataclassT` as the name of the `TypeVar`, to make it slightly more obvious what's happening, and give it `bound=DataclassType`. (The TypeVar for `BaseModel` in `pydantic.main` is called `Model`; I think that's nice.)

samuelcolvin:
As far as I can tell, we're just waiting for a tweak to the signature of `__call__`. Is that correct?

If not, let me know and I'll merge.

koxudaxi:
@samuelcolvin 
Sorry, I have pushed it which is changed by @dmontagu's review

samuelcolvin:
Woop woop. Thank you very much.

Issue #971, opened 2019-11-07
Byte and Time Conversion Types


dgasmith:
# Feature Request

For many of our `BaseSettings` classes we have fields that either describe a time duration or byte size, typically we have to pick a unit ('GB' or 's') that these should be represented in and users must pass in a value in that unit. It would be more convenient if users could specify `3 TiB` or `30 m` and have this be cast to the correct type. I see two possible implementations, the first where the units type are parametrized and cast to this value:

```
class Config(BaseSettings):
    size: ByteSize['GB']

c = Config(size='1TB')
assert isinstance(c.size, float)
assert c.size == 1000
```

The other where the value remains an object:

```
class Config(BaseSettings):
    size: ByteSize

c = Config(size='1TB')
assert isinstance(c.size, ByteSize)
assert c.size.to('GB') == 1000
assert int(c.size) == int(1e9)
```

Similar structures can be built for a `TimeUnit` object. Several questions:
1) Anyone have thoughts on the implementation or names?
2) Would you want this contributed to pydantic?

samuelcolvin:
I'd be happy with `ByteSize`, but I think it should always just be an int. As with `AnyUrl` it could have some method to format it or change the base, but it should inherit from and/or act like an int representing bits.

Regarding `1s` or `30m`, this would be a change to the `timedelta` parsing, fine by me as long as it is clear and obviously deterministic.

dgasmith:
Ok, I can add the `ByteSize['GB']` implementation that would inherit from a `float` for cases like `size: ByteSize['GB'] = '5 MB'`.

Makes sense for `timedelta`.

samuelcolvin:
I real think any disk or memory size should inherit from int. I wouldn't accept something that inherited from float as I think it would lead to problems.

Also:

`MB` vs `MiB`, should we be technically accurate and call `1 MB = 1,000,000` or follow convention and use `1 MB = 1,048,576`?

I guess we should give the technically correct meaning to `MB` and MiB` (1,000,000 and 1,048,576 respectively) and document it clearly.

But this isn't going to be useful to many people, from [here](https://en.wikipedia.org/wiki/Kilobyte):

[...]

PR #977, opened 2019-11-07
Bytesize

dgasmith:
## Change Summary

Adds a `ByteSize` object to automatically cast common byte string to bytes.

## Related issue number

#971

## Checklist


docs/usage/types.md
```diff
@@ -139,6 +139,9 @@ with custom properties and validation.
 : strings and bytes (converted to strings) are passed to `UUID(v)`;
   see [Pydantic Types](#pydantic-types) for other stricter UUID types
 
+`ByteSize`
+: converts a bytes string with units to bytes
```
samuelcolvin: `ByteSize` will need a section describing it's usage.

pydantic/errors.py
```diff
@@ -499,3 +501,11 @@ class LuhnValidationError(PydanticValueError):
 class InvalidLengthForBrand(PydanticValueError):
     code = 'payment_card_number.invalid_length_for_brand'
     msg_template = 'Length for a {brand} card must be {required_length}'
+
+
+class InvalidByteString(PydanticValueError):
```
samuelcolvin:
```suggestion
class InvalidByteSize(PydanticValueError):
```
and below

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+    'EB': 10 ** 18,
+    'KiB': 2 ** 10,
+    'MiB': 2 ** 20,
+    'GiB': 2 ** 30,
+    'TiB': 2 ** 40,
+    'PiB': 2 ** 50,
+    'EiB': 2 ** 60,
+}
+BYTE_SIZES.update({k.lower(): v for k, v in BYTE_SIZES.items()})
+byte_string_re = re.compile(r'^\s*(\d*\.?\d+)\s*(\w+)?', re.IGNORECASE)
```
samuelcolvin:
```suggestion
byte_string_re = re.compile(r'^\s*(\d*\.?\d+)\s*(\w+)\s*', re.IGNORECASE)
```
`' 123   '` will be parsed to int anyway.

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
         else:
             brand = PaymentCardBrand.other
         return brand
+
+
+BYTE_SIZES = {
+    '': 1,
```
samuelcolvin: I think can be removed.

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+            return v
+
+        try:
+            return cls(int(v))
+        except ValueError:
+            pass
+
+        str_match = byte_string_re.match(v)
+        if str_match is None:
+            raise errors.InvalidByteString
```
samuelcolvin:
```suggestion
            raise errors.InvalidByteString()
```

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+
+        scalar, unit = str_match.groups()
+        if unit is None:
+            unit = "B"
```
samuelcolvin: single quotes please.

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+
+        try:
+            unit_mult = BYTE_SIZES[unit.lower()]
```
samuelcolvin: are we sure about the lower here?

`1b` generally means "1 bit" not "1 byte".

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+        except KeyError:
+            raise errors.InvalidByteUnit(unit=unit)
+
+        return cls(int(float(scalar) * unit_mult))
+
+    def human_readable(self):
+
+        num = float(self.value)
+        for unit in ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB']:
```
samuelcolvin: I think this method should allow decimal or binary display. with the default being binary as it is here.

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+    'TiB': 2 ** 40,
+    'PiB': 2 ** 50,
+    'EiB': 2 ** 60,
+}
+BYTE_SIZES.update({k.lower(): v for k, v in BYTE_SIZES.items()})
+byte_string_re = re.compile(r'^\s*(\d*\.?\d+)\s*(\w+)?', re.IGNORECASE)
+
+
+class ByteSize(int):
+    def __init__(self, value):
```
samuelcolvin: do we need this?

pydantic/types.py
dgasmith: Hard to say, I see it quite a few cases where it is stated as `1gb` meaning `1GB`, but your bits comment is very valid. I would say casting lower is fine for my domain, but network engineers would likely be less than happy about this.

samuelcolvin: Let's keep it like this. It's called `ByteSize` after call.

But please add an `info` section to the docs explaining this.

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
         else:
             brand = PaymentCardBrand.other
         return brand
+
+
+BYTE_SIZES = {
+    '': 1,
+    'B': 1,
+    'kB': 10 ** 3,
```
samuelcolvin: about about if someone enters `10k`, that's what `ls` used by default.

Perhaps the `B` is superfluous in the `kB...EB` cases since it's already called `ByteSize`?

pydantic/types.py
```diff
@@ -606,3 +607,68 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+    'MB': 10 ** 6,
+    'GB': 10 ** 9,
+    'TB': 10 ** 12,
+    'PB': 10 ** 15,
+    'EB': 10 ** 18,
```
samuelcolvin: I'm not sure about `EB`, it's pretty rare. If we go this far we could also add Zetta and Yotta.

Maybe best to stop at `PB` which everyone knows?

I'm not sure myself.

pydantic/types.py
dgasmith: `int(1.0)` will not be however.

samuelcolvin: okay.

pydantic/types.py
dgasmith: Added.

pydantic/types.py
dgasmith: We sometimes deal with EB (high energy physics) which is why I had it in, but my perspective is likely a bit skewed.

pydantic/types.py
dgasmith: No, got it.

pydantic/types.py
samuelcolvin: let's keep it then.

pydantic/types.py
dgasmith: Added.

pydantic/types.py
dgasmith: Added.

samuelcolvin:
can you fix tests please. I think it's just double quotes.

Best to run `make test` before pushing.

docs/usage/types.md
```diff
@@ -668,6 +671,19 @@ The following caveats apply:
 ```
 
+## ByteSize
+
+You can use the `ByteSize` data type to convert byte string representation to
+raw bytes and print out human readable versions of the bytes as well. Note that
```
samuelcolvin: best to use an actual info block

```
!!! info
    Note that size suffixes are parsed in a case insensitive way, so `1b` will be parsed as "1 byte" and not "1 bit".
```

docs/examples/types_bytesize.py
```diff
@@ -0,0 +1,12 @@
+from pydantic import BaseModel, ByteSize
+
+class MyModel(BaseModel):
+    size: ByteSize
+
+print(MyModel(size=52000).size)
+print(MyModel(size="3000 KiB").size)
```
samuelcolvin: single quotes please.

pydantic/types.py
```diff
@@ -606,3 +607,73 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+    'KiB': 2 ** 10,
+    'MiB': 2 ** 20,
+    'GiB': 2 ** 30,
+    'TiB': 2 ** 40,
+    'PiB': 2 ** 50,
+    'EiB': 2 ** 60,
+}
+BYTE_SIZES.update({k.lower(): v for k, v in BYTE_SIZES.items()})
+BYTE_SIZES.update({k.lower()[0]: v for k, v in BYTE_SIZES.items() if "i" not in k})
+BYTE_SIZES.update({k.lower()[:2]: v for k, v in BYTE_SIZES.items() if "i" in k})
```
samuelcolvin: I don't think this should be added. `Ei` doesn't mean anything, in the way `G` does.

pydantic/types.py
```diff
@@ -606,3 +607,73 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+    'PB': 10 ** 15,
+    'EB': 10 ** 18,
+    'KiB': 2 ** 10,
+    'MiB': 2 ** 20,
+    'GiB': 2 ** 30,
+    'TiB': 2 ** 40,
+    'PiB': 2 ** 50,
+    'EiB': 2 ** 60,
+}
+BYTE_SIZES.update({k.lower(): v for k, v in BYTE_SIZES.items()})
```
samuelcolvin: surely we don't need this, just define them as lower case in the main dictionary?

pydantic/types.py
```diff
@@ -606,3 +607,73 @@ def _get_brand(card_number: str) -> PaymentCardBrand:
...
+BYTE_SIZES.update({k.lower()[0]: v for k, v in BYTE_SIZES.items() if "i" not in k})
```
samuelcolvin: maybe add a test for this case.

samuelcolvin:
thanks so much.

samuelcolvin:
Id of this matches [my favourite xkcd](https://xkcd.com/977/)

PR #978, opened 2019-11-07
add buy-sell-ads advert to docs

samuelcolvin:
## Change Summary

Advertisements via buy-sell-ads.

Easy enough to add since [helpmanual.io](https://helpmanual.io/) already has the ads. 

I'll leave this for a few days, let me know if there's a problem.

### DIFF
--- a/docs/extra/ad.css
+++ b/docs/extra/ad.css
@@ -1,0 +1,11 @@
+#bsa-cpc {
+  border-radius: .1rem;;
+  padding: 0.525rem 0.6rem;
+  min-height: 80px;
+}
+#bsa-cpc.loaded {
+  background: hsla(0, 0%, 92.5%, 0.5);
+}
+#_default_ a._default_, #_default_ .default-text {
+  width: 100% !important;
+}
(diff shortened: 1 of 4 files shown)


dmontagu:
I have no problem with this, but Is there an easy way to see what it will look like? E.g., are there other mkdocs sites with it there? If I just run `make docs-serve` will I see it?

I'm assuming this is meant to cover hosting costs, which would be fine with me.

samuelcolvin:
Odd that it's not working on netlify, will fix tomorrow or add a screenshot.

Hosting is free, this is more about my belief that open source development should pay (at least a little). More the principal and an experiment than any expectation of significant income.

I'd be happy to share any revenue with other regular developers, if it ever amounts at anything. My guess would be that this will generate <$10/month.

samuelcolvin:
`make docs-serve` should work fine.

samuelcolvin:
The netlify site works fine and shows the ad, click "deploy/netlify" in the list of checks above.

There are two caveats:
1. The ads don't show on mobile, I think this is decision of ad network
2.  There's no ad on the home page, this was a decision of mine so the landing page looks clean

samuelcolvin:
image

PR #980, opened 2019-11-08
Feature/dataclass schema defaults

### LINKED ISSUE #968: dataclass field doesn't change required schema
# Bug

Please complete:
* OS: Mac OS
* Python version `import sys; print(sys.version)`: **3.6.7** with dataclasses backport
* Pydantic version `import pydantic; print(pydantic.VERSION)`: **1.0**

It is known that `@pydantic.dataclasses.dataclass` doesn't fully mimic the `BaseModel`. However, we'd like use `.schema()` representation faithfully. I'm wondering what's the best workaround.

Adding a `Config` for `b` with `required=False` doesn't carry over to `.schema()` either.

```py
from typing import Dict

from dataclasses import field
from pydantic import Field, ValidationError

@dataclass
class D1:
    a: int
    # b: Dict[str, int] = {"x":10} # mutable default, not allowed
@dataclass
class D2:
    a: int
    b: Dict[str, int] = Field({"x":10}) # pydantic.Field, unsafe and invalid default
@dataclass        
class D3:
    a: int
    b: Dict[str, int] = field(default_factory=lambda: {"x":10}) # std lib, safe and invalid default
        
class C:
    fields = {"b": { "required": False}} # doesn't carry over!!

@dataclass(config=C)        
class D4:
    a: int
    b: Dict[str, int] = field(default_factory=lambda: {"x":10})

print(D1(10))
try:
    print(D2(10))
except ValidationError as e:
    print(e) # pydantic/dataclasses.py:77

print(D3(10))
req = D3.__pydantic_model__.schema()["required"]
if req != ["a"]:
    print('only ["a"] should be required, is:', req)
    
print()
print(D4.__pydantic_model__.schema()["required"]) # same as D3
```

Output
```
D1(a=10)
1 validation error for D2
b
  value is not a valid dict (type=type_error.dict)
D3(a=10, b={'x': 10})
only ["a"] should be required, is: ['a', 'b']

['a', 'b']
```

ahirner:
## Change Summary
* Make use of default_factory lambda when creating pydantic fields
* Add separate test for this kind of default
* Add field for this kind of default in test_schema

## Related issue number
Fixes #968 
If fine and complete, lets discuss docs

## Checklist

* [X] Tests pass on CI and coverage remains

### DIFF
--- a/docs/examples/dataclasses_default_schema.py
+++ b/docs/examples/dataclasses_default_schema.py
@@ -9,4 +9,4 @@ class User:
     friends: List[int] = dataclasses.field(default_factory=lambda: [0])
 
 user = User(id='42')
-print(user.__pydantic_model__.schema()['properties']['friends'])
+print(user.__pydantic_model__.schema())
--- a/docs/examples/dataclasses_default_schema.py
+++ b/docs/examples/dataclasses_default_schema.py
@@ -1,0 +1,12 @@
+import dataclasses
+from typing import List
+from pydantic.dataclasses import dataclass
+
+@dataclass
+class User:
+    id: int
+    name: str = 'John Doe'
+    friends: List[int] = dataclasses.field(default_factory=lambda: [0])
+
+user = User(id='42')
+print(user.__pydantic_model__.schema()['properties']['friends'])
--- a/changes/968-ahirner.md
+++ b/changes/968-ahirner.md
@@ -1,0 +1,1 @@
+Add support for dataclasses default factory
--- a/docs/usage/dataclasses.md
+++ b/docs/usage/dataclasses.md
@@ -19,6 +19,14 @@ _(This script is complete, it should run "as is")_
 You can use all the standard pydantic field types, and the resulting dataclass will be identical to the one
 created by the standard library `dataclass` decorator.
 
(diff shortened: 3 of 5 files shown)


docs/examples/dataclasses_default_schema.py
```diff
@@ -0,0 +1,12 @@
...
+from pydantic.dataclasses import dataclass
+
+@dataclass
+class User:
+    id: int
+    name: str = 'John Doe'
+    friends: List[int] = dataclasses.field(default_factory=lambda: [0])
+
+user = User(id='42')
+print(user.__pydantic_model__.schema()['properties']['friends'])
```
samuelcolvin:
```suggestion
print(user.__pydantic_model__.schema())
```
Is probably enough, best to keep documentation as simple as possible while demonstrating the point.

ahirner: I had this at first then thought the prints use up too much space. So you prefer the complete schema for sure?

samuelcolvin: yes please.

samuelcolvin:
great, thank you very much.

PR #985, opened 2019-11-08
fix mypy complaint on @root_validator(pre=True)

### LINKED ISSUE #984: mypy complains about pre=True root validators
# Bug

Please complete:
* OS: **Ubuntu 19.04**
* Python version `import sys; print(sys.version)`: **3.7.3**
* Pydantic version `import pydantic; print(pydantic.VERSION)`: **1.0**

With the code found in the [root validators](https://pydantic-docs.helpmanual.io/usage/validators/#root-validators) documentation:

```py


class UserModel(BaseModel):
    username: str
    password1: str
    password2: str

    @root_validator(pre=True)
    def check_card_number_omitted(cls, values):  # type: ignore
        assert 'card_number' not in values, 'card_number should not be included'
        return values
```

Mypy has the following to say:

```
9: error: "classmethod" not callable
```

The following change pleases mypy, but of course that changes the semantics:

```diff
-    @root_validator(pre=True)
+    @root_validator
```

samuelcolvin:
## Change Summary

Change signature of `root_validator`. Not entirely clear why mypy works this way, I would have thought the previous signature was correct, but this works

## Related issue number

fix #984

## Checklist

### DIFF
--- a/changes/984-samuelcolvin.md
+++ b/changes/984-samuelcolvin.md
@@ -1,0 +1,1 @@
+Fix mypy complaint about `@root_validator(pre=True)`
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -4,7 +4,7 @@ from functools import wraps
 from inspect import Signature, signature
 from itertools import chain
 from types import FunctionType
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Type
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Type, Union, overload
 
 from .errors import ConfigError
 from .typing import AnyCallable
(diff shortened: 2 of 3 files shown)


bochecha:
> Not entirely clear why mypy works this way, I would have thought the previous signature was correct, but this works

Could it be a bug in mypy then?

I must admit I hesitated to report it there, but figured I'd start here so we can figure out where the issue is.

samuelcolvin:
Perhaps, feel free to report it to mypy and link to this PR if you wish.

dmontagu:
@bochecha I don't think this is a bug in mypy -- it's because mypy checks all cases of a `Union`.

@samuelcolvin There is a better fix for this using `@overload` that doesn't require a # type: ignore and ensures the result is treated by mypy as a classmethod in all cases.

Give me a second and I'll add it as a comment here; you can decide which version you prefer.

dmontagu:
If you add `reveal_type` calls to the success.py, you'll see the version in this PR is actually still getting the type wrong:

```python
class Model(BaseModel):
    age: int
    first_name = 'John'
    last_name: NoneStr = None
    signup_ts: Optional[datetime] = None
    list_of_ints: List[int]

    @validator('age')
    def check_age(cls, value: int) -> int:
        assert value < 100, 'too old'
        return value

    @root_validator
    def root_check(cls, values: Dict[str, Any]) -> Dict[str, Any]:
        return values

    @root_validator(pre=True)
    def pre_root_check(cls, values: Dict[str, Any]) -> Dict[str, Any]:
        return values

    reveal_type(root_check)
    reveal_type(pre_root_check)
```

Gives the output:
```
tests/mypy/modules/success.py:41: note: Revealed type is 'def (def (*Any, **Any) -> Any) -> builtins.classmethod'
tests/mypy/modules/success.py:42: note: Revealed type is 'builtins.classmethod'
```
-----

Here's the overloaded approach, which gives correct type hints:
```python
@overload
def root_validator(_func: AnyCallable) -> classmethod:
    ...

@overload
def root_validator(*, pre: bool = False) -> Callable[[AnyCallable], classmethod]:
    ...

def root_validator(_func: Optional[AnyCallable] = None, *, pre: bool = False) -> Union[classmethod, Callable[[AnyCallable], classmethod]]:
    if _func:
        _check_validator_name(_func)
        f_cls = classmethod(_func)
        setattr(f_cls, ROOT_VALIDATOR_CONFIG_KEY, Validator(func=_func, pre=pre))
        return f_cls

    def dec(f: AnyCallable) -> classmethod:
        _check_validator_name(f)
        f_cls = classmethod(f)
        setattr(f_cls, ROOT_VALIDATOR_CONFIG_KEY, Validator(func=f, pre=pre))
        return f_cls

    return dec
```
If you run the above check with this version, you get:

```
tests/mypy/modules/success.py:41: note: Revealed type is 'builtins.classmethod'
tests/mypy/modules/success.py:42: note: Revealed type is 'builtins.classmethod'
```

samuelcolvin:
@dmontagu thanks so much for the explanation, makes sense.

I've made the changes as you suggested.

bochecha:
Indeed, thanks @dmontagu !

PR #987, opened 2019-11-09
Improve benchmarks

samuelcolvin:
* Adding benchmarks for [cerberus](https://docs.python-cerberus.org/en/stable/)
* remove `toasted-marshmallow`, it wasn't any faster than marshmallow and prevented use of the latest version of marshamllow
* upgrade marshmallow tests to support marshmallow 3


samuelcolvin:
Cerberus is **slooooow**. Really slow. ~26x slower than pydantic and twice as slow as the next slowest library drf which in turn is massively slower than all the other libraries tested.

Am I doing something wrong or is this just how it is?

samuelcolvin:
No reply in 4 days, merging this. If anyone comes here from the SO issue (or elsewhere) and has suggestions of improvements, feel free to create a PR.

PR #998, opened 2019-11-15
fix typo: 'frozonset' -> 'frozenset'

nwalsh1995:


samuelcolvin:
Good catch, thanks so much.

PR #1003, opened 2019-11-17
Add exclude_none option

### LINKED ISSUE #587: add skip_none argument
# Question / Feature Request 

Suppose that I have a class like this:

```
class Jake(BaseModel):
    foo: Optional[str]
```

When I serialize it, I only want to serialize optional values when they are not null.  The `skip_defaults` flag does something _similar_ to what I am looking for but has this weird case:

```
>>> j = Jake()
>>> 
>>> # This is the behavior I expect: 
... j.dict(skip_defaults=True)
{}
>>> 
>>> # Set the foo property: 
>>> j.foo = "baz"
>>> 
>>> # This is also the behaviour I expect: 
... j.dict(skip_defaults=True)
{'foo': 'baz'}
>>> 
>>> 
>>> # Unfortunately, this is not the behavior that I expect:
... j.foo = None
>>> j.dict(skip_defaults=True)
{'foo': None}
```

---

Right now after I serialize to dict, I walk the dict and remove all of the `None` values recursively, but I would rather not have to do this.  Is this built into Pydantic?  Is the behavior that I am seeing a bug?

niknetniko:
## Change Summary

- Add `exclude_none` option to `dict()` and friends.

The original issue talked about `skip_none`, but since all options are now in the `exclude_x` format, that is also what I've done.

## Related issue number

Resolves #587

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -659,7 +659,7 @@ class BaseModel(metaclass=ModelMetaclass):
                     exclude_defaults=exclude_defaults,
                     exclude_none=exclude_none,
                 )
-                if (exclude_none and value is not None) or not exclude_none:
+                if not (exclude_none and value is None):
                     yield k, value
 
     def _calculate_keys(
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -564,7 +564,7 @@ class BaseModel(metaclass=ModelMetaclass):
                     exclude_defaults=exclude_defaults,
                     include=include,
                     exclude=exclude,
-                    exclude_none=exclude_none
+                    exclude_none=exclude_none,
                 )
             else:
                 return v.copy(include=include, exclude=exclude)
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -308,6 +308,7 @@ class BaseModel(metaclass=ModelMetaclass):
         skip_defaults: bool = None,
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
+        exclude_none: bool = False,
     ) -> 'DictStrAny':
         """
         Generate a dictionary representation of the model, optionally specifying which fields to include or exclude.
@@ -332,6 +333,7 @@ class BaseModel(metaclass=ModelMetaclass):
                 exclude=exclude,
                 exclude_unset=exclude_unset,
                 exclude_defaults=exclude_defaults,
+                exclude_none=exclude_none,
             )
         }
 
@@ -350,6 +352,7 @@ class BaseModel(metaclass=ModelMetaclass):
         skip_defaults: bool = None,
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
+        exclude_none: bool = False,
         encoder: Optional[Callable[[Any], Any]] = None,
         **dumps_kwargs: Any,
     ) -> str:
@@ -371,6 +374,7 @@ class BaseModel(metaclass=ModelMetaclass):
             by_alias=by_alias,
             exclude_unset=exclude_unset,
             exclude_defaults=exclude_defaults,
+            exclude_none=exclude_none,
         )
         if self.__custom_root_type__:
             data = data[ROOT_KEY]
@@ -549,6 +553,7 @@ class BaseModel(metaclass=ModelMetaclass):
         exclude: Optional[Union['AbstractSetIntStr', 'DictIntStrAny']],
         exclude_unset: bool,
         exclude_defaults: bool,
+        exclude_none: bool,
     ) -> Any:
 
         if isinstance(v, BaseModel):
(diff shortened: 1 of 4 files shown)


pydantic/main.py
```diff
@@ -640,15 +649,18 @@ def _iter(
...
                     v,
                     to_dict=to_dict,
                     by_alias=by_alias,
                     include=value_include and value_include.for_element(k),
                     exclude=value_exclude and value_exclude.for_element(k),
                     exclude_unset=exclude_unset,
                     exclude_defaults=exclude_defaults,
+                    exclude_none=exclude_none,
                 )
+                if (exclude_none and value is not None) or not exclude_none:
```
```suggestion
                if not (exclude_none and value is None):
```
I think slightly clearer and more efficient.

tests/test_edge_cases.py
```diff
@@ -1303,3 +1303,89 @@ class MyModel(BaseModel):
...
+    assert m.dict(exclude_none=True) == {'c': 5, 'd': 2, 'e': {'a': 0, 'b': 1}}
+    assert dict(m) == {'c': 5, 'd': 2, 'e': {'a': 0, 'b': 1}, 'f': None}
+
+    m = ModelB(c=5, e={'b': 20}, f='test')
+    assert m.dict() == {'c': 5, 'd': 2, 'e': {'a': None, 'b': 20}, 'f': 'test'}
+    assert m.dict(exclude_none=True) == {'c': 5, 'd': 2, 'e': {'b': 20}, 'f': 'test'}
+    assert dict(m) == {'c': 5, 'd': 2, 'e': {'a': None, 'b': 20}, 'f': 'test'}
+
+
+def test_dict_exclude_none_populated_by_alias():
```
samuelcolvin: this name is confusing, but also I think the test is unnecessary.

tests/test_edge_cases.py
```diff
@@ -1303,3 +1303,89 @@ class MyModel(BaseModel):
...
+            allow_population_by_field_name = True
+
+    m = MyModel(alias_a='a')
+
+    assert m.dict(exclude_none=True) == {'a': 'a'}
+    assert m.dict(exclude_none=True, by_alias=True) == {'alias_a': 'a'}
+    assert m.dict(by_alias=True) == {'alias_a': 'a', 'alias_b': None}
+
+
+def test_dict_exclude_none_populated_by_alias_with_extra():
```
samuelcolvin: Best to test just one thing at a time, e.g. here just `test_exclude_none_with_extra`

tests/test_edge_cases.py
```diff
@@ -1303,3 +1303,89 @@ class MyModel(BaseModel):
...
+            extra = 'allow'
+
+    m = MyModel(alias_a='a', c='c')
+
+    assert m.dict(exclude_none=True) == {'a': 'a', 'c': 'c'}
+    assert m.dict(exclude_none=True, by_alias=True) == {'alias_a': 'a', 'c': 'c'}
+    assert m.dict(by_alias=True) == {'alias_a': 'a', 'alias_b': None, 'c': 'c'}
+
+
+def test_dict_exclude_none_populated_with_extra():
```
samuelcolvin: I don't think both this test and the above are required.

tests/test_edge_cases.py
```diff
@@ -1303,3 +1303,89 @@ class MyModel(BaseModel):
 
     m = MyModel(foo={'x': 'a', 'y': None})
     assert m.foo == {'x': 'a', 'y': None}
+
+
+def test_exclude_none_dict():
```
samuelcolvin:
```suggestion
def test_exclude_none():
```

tests/test_edge_cases.py
```diff
@@ -1303,3 +1303,89 @@ class MyModel(BaseModel):
...
+    class MyModel(BaseModel):
+        a: Optional[int] = None
+        b: int = 2
+
+    m = MyModel(a=5)
+    assert m.dict(exclude_none=True) == {'a': 5, 'b': 2}
+
+    m = MyModel(b=3)
+    assert m.dict(exclude_none=True) == {'b': 3}
```
samuelcolvin:
```suggestion
    assert m.dict(exclude_none=True) == {'b': 3}
    assert m.json(exclude_none=True) == '{"b": 3}'
```

tests/test_edge_cases.py
```diff
@@ -1303,3 +1303,89 @@ class MyModel(BaseModel):
...
+
+    m = MyModel()
```
samuelcolvin: This case doesn't test anything, can be removed.

niknetniko:
Thanks for the speedy review; I've applied the suggestions.

samuelcolvin:
awesome, thank you very much.

Issue #607, opened 2019-06-20
.env file like crea when using BaseSettings


samuelcolvin:
so secrets can be kept out of git without having to always activate an environment.

phy25:
Looks like this can be closed since #1011 is merged?

samuelcolvin:
Yes, see directly above.

Will be released soon.

PR #1011, opened 2019-11-19
allow use of a `.env`-style files in BaseSettings (#607)

acnebs:
## Change Summary

Update `BaseSettings` class to allow the use of `.env`-style files for storing sensitive environment variables. 

Just pass in the path (either absolute or relative to the working directory) to such a file in the Config subclass, and pydantic will do the rest for you.

#### .env
```
foo=baz
```

#### settings.py
```python3

class Settings(BaseSettings):
    foo: str = 'bar'

    class Config:
        env_file = '.env'
```

#### test.py
```python3
from settings import Settings

s = Settings()
print(s.foo)
# output: baz
```

## Related issue number

#607 

## Checklist


pydantic/env_settings.py
```diff
@@ -38,6 +38,10 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
         else:
             env_vars = {k.lower(): v for k, v in os.environ.items()}
 
+        env_file = self.__config__.env_file
```
samuelcolvin: Please use a `Path` object here.

Eg.

```py
if self.__config__.env_file:
    env_path = Path(self.__config__.env_file)
    env_vars.update(self._read_from_env_file(env_path))
```

pydantic/env_settings.py
```diff
@@ -56,8 +60,21 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
             d[field.alias] = env_val
         return d
 
+    def _read_from_env_file(self, file_path: str) -> Dict[str, str]:
```
samuelcolvin: I think we should have something like

```py
logger = logging.getLogger('pydantic.settings')
...
    if not env_path.exists():
        logger.warning...
    elif not env_path.is_file():
        raise exception...
```
Or perhaps we should use [warnings](https://docs.python.org/3/library/warnings.html)?

pydantic/env_settings.py
```diff
@@ -56,8 +60,21 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
...
+        file_vars = {}  # type: Dict[str, str]
```
samuelcolvin:
```suggestion
        file_vars: Dict[str, str] = {}
```

pydantic/env_settings.py
samuelcolvin: I think `read_env_file` should be a standalone file to allow easier testing.

I guess signature will end up being:

```py
def read_env_file(file_path: Path, case_sensitive: bool = False) -> Dict[str, str]:
```

pydantic/env_settings.py
samuelcolvin: I think it should also be possble to set `env_file` via a keyword argument to `__init__`, e.g. `_env_file`.

The keyword argument should take precedence over the config argument including if `_env_file=None`

pydantic/env_settings.py
```diff
@@ -56,8 +60,21 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
...
+        with open(file_path) as env_file:
+            for line in env_file.readlines():
+                line = line.strip()
+                if '=' in line and not line.startswith('#'):
```
samuelcolvin:
```suggestion
                if line.startswith('#'):
                    continue
                elif '=' in line:
                    raise ...
```
better to raise an error if the line does not start with '#' and does not contain '='

pydantic/env_settings.py
```diff
@@ -56,8 +60,21 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
             d[field.alias] = env_val
         return d
 
+    def _read_from_env_file(self, file_path: str) -> Dict[str, str]:
+        file_vars = {}  # type: Dict[str, str]
+        with open(file_path) as env_file:
+            for line in env_file.readlines():
```
samuelcolvin: probably easier to just do `enumerate(file_path.read_text().split('\n'), start=1)` (tracking line number for errors below).

pydantic/env_settings.py
samuelcolvin: This might all be easier with a regex something like `re.match('\s*(\w+)\s*=\s*(.+?)\s*', line)` and raise an error if the regex can't be found.

acnebs: I did it this way because it covers all use-cases (AFAIK) and just ignores blank lines, which are fine to have in a `.env` I think.

pydantic/env_settings.py
acnebs: I was under the impression that `.readlines()` is better for cross-platform support than using `.split()`.

pydantic/env_settings.py
samuelcolvin: okay, don't mind which.

samuelcolvin: yes, blank lines (or lines with only white space) should be skipped.

pydantic/env_settings.py
samuelcolvin: maybe, alternatively we could use `re.split('[\r\n]+', file_path.read_text())` which avoids keeping the file open while we process lines and I guess should be a little faster.

Don't really mind.

pydantic/env_settings.py
acnebs: How does `_env_file=None` override the config argument without making it a required argument? i.e. How are we differentiating between not passing in `_env_file` and passing in `_env_file=None`? Potentially we need the disable functionality to be `_env_file=False`, no?

samuelcolvin: Generally people do this using a "sentinel" value.

Something like

```py
env_file_sentinel = object()
...
def __init__(..., _env_file: Union[None, str, Path] = env_file_sentinel):
    ...
    if _env_file != env_file_sentinel:
        ...
```

dmontagu: I think you could also just use `Ellipsis` (aka `...`) for this purpose here.

samuelcolvin: ellipsis is fine with me.

pydantic/env_settings.py
acnebs: I'm thinking proper syntax validation of your `.env` file is beyond the scope of pydantic, so would you be fine with just logging a warning if we hit a line that looks bad?

pydantic/env_settings.py
acnebs: For this case I think it makes sense for both issues with the `env_path` to raise an Exception (especially because Errors for each of those events already exist in `errors.py`).

samuelcolvin: Definitely don't use errors from `errors.py`, they're specifically for validation.

Up to you whether a missing `env_file` is an error or not.

pydantic/env_settings.py
samuelcolvin: If we can't check the format of env files, then we can't accomplish this feature and we should close this PR. But personally I don't think it should be very difficult.

Definitely parsing env files should either succeed or fail, warnings would just make things more complicated.

pydantic/env_settings.py
samuelcolvin: A missing env file should not cause an exception, maybe even the warning is too much. Initialisation should only fail if validation fails.

pydantic/env_settings.py
acnebs: Probably we should just include `python-dotenv`'s implementation of a parser then, no?

https://github.com/theskumar/python-dotenv/blob/master/src/dotenv/parser.py

pydantic/env_settings.py
acnebs: Seems strange to care about the format of the file but not whether or not it actually exists.

pydantic/env_settings.py
acnebs: Maybe I'm doing something wrong, but `Ellipsis` doesn't seem to play well with `typing`.

pydantic/env_settings.py
samuelcolvin: No, it's very simple, please just implement directly as discussed above.

pydantic/env_settings.py
samuelcolvin: No it doesn't this is very common.

pydantic/env_settings.py
samuelcolvin: You'll need `#  type: ignore`, otherwise you could use the sentinel approach with either:
* `env_file_sentinel: str = object()  # type: ignore`
* or `env_file_sentinel = str(object())`

samuelcolvin: or you could do

```py
from pathlib import Path
from typing import Union

EllipsisType = type(...)

def foobar(a: Union[None, str, Path, EllipsisType] = ...) -> None:
    return None
```

acnebs:
Still haven't written any tests yet, but I figure it's best to leave that until we've settled on the rest of the implementation.

pydantic/env_settings.py
```diff
@@ -20,8 +24,9 @@ class BaseSettings(BaseModel):
     Heroku and any 12 factor app design.
     """
 
-    def __init__(__pydantic_self__, **values: Any) -> None:
+    def __init__(__pydantic_self__, _env_file: Union[None, str, Path] = env_file_sentinel, **values: Any) -> None:
         # Uses something other than `self` the first arg to allow "self" as a settable attribute
+        object.__setattr__(__pydantic_self__, '_env_file', _env_file)
```
samuelcolvin: instead of doing this, pass `_env_file` to `_build_values`

pydantic/env_settings.py
```diff
@@ -33,11 +38,19 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
         """
         d: Dict[str, Optional[str]] = {}
 
+        env_vars: Dict[str, str]
         if self.__config__.case_sensitive:
-            env_vars: Mapping[str, str] = os.environ
+            env_vars = dict(os.environ)
```
samuelcolvin: Please don't change this, in theory it could lead to an unnecessary memory increase.

Instead do:

```py
env_vars = {**env_vars, **read_env_file(env_path, self.__config__.case_sensitive)}
```
or something, below.

pydantic/env_settings.py
```diff
@@ -33,11 +38,19 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
...
         else:
             env_vars = {k.lower(): v for k, v in os.environ.items()}
 
+        _env_file = getattr(self, '_env_file')
+        env_file = _env_file if _env_file != env_file_sentinel else self.__config__.env_file
+        if env_file is not None:
+            env_path = Path(env_file)
+            if env_path.exists() and env_path.is_file():
```
samuelcolvin:
```suggestion
            if env_path.is_file():
```

pydantic/read_env_file.py
```diff
@@ -0,0 +1,23 @@
...
+class EnvFileError(ValueError):
+    pass
+
+
+def read_env_file(file_path: Path, case_sensitive: bool = False) -> Dict[str, str]:
+    file_vars: Dict[str, str] = {}
+    for n, line in enumerate(re.split('[\r\n]+', file_path.read_text()), start=1):
+        line = line.strip()
+        if line and not line.startswith('#'):
+            if re.match(r'\s*(\w+)\s*=\s*(.+?)\s*', line):
```
samuelcolvin:
```py
m = re.match(r'\s*(\w+)\s*=\s*(.+?)\s*', line)
if m:
    key, value = m.groups()
    ...
else:
    ...
```

pydantic/read_env_file.py
```diff
@@ -0,0 +1,23 @@
...
+                key, value = line.split('=', 1)
+                key = key.strip()
```
samuelcolvin: no need for this since it's done by the regex.

pydantic/read_env_file.py
```diff
@@ -0,0 +1,23 @@
+import re
```
samuelcolvin: I don't think this needs a separate file, please move to the bottom of `env_settings.py`.

pydantic/env_settings.py
```diff
@@ -1,12 +1,16 @@
...
+from pathlib import Path
+from typing import Any, Dict, Iterable, Optional, Union
 
 from .fields import ModelField
 from .main import BaseModel, Extra
+from .read_env_file import read_env_file
 from .typing import display_as_type
 from .utils import deep_update
 
+env_file_sentinel: str = object()  # type: ignore
```
samuelcolvin: on second thoughts I think this is a big weird/dangerous. Best to use the ellipsis and add `EllipsisType` to the union as described above.

acnebs: Tried that, didn't work. mypy complains about it not being a type when I try `EllipsisType = type(...)` or `type(Ellipsis)`.

pydantic/read_env_file.py
acnebs: No need for line 14? I'm doing that because a failure to match throws an error, but I don't want to throw an error for comments / blanklines.

samuelcolvin: no need for line 17 which is where this comment is...

acnebs: Interesting, on my end the highlight is showing it as lines 14-17.

pydantic/env_settings.py
samuelcolvin: odd, okay then probably best to use `env_file_sentinel = str(object())`, looks weird, but I think should work well.

Or you can just type ignore the function definition.

pydantic/env_settings.py
samuelcolvin: you shouldn't. Please try to work this out yourself.

If you have trouble I'll take over the PR.

euri10:
Sorry to highjack the PR, I'm currently having a simple function to push .env fies into objects inheriting BaseSettings using `load_dotenv` from the dotenv package.

Stuff like this for instance:
```
class DatabaseSetting(BaseSettings):
    host: str = "localhost"
    port: int = 5432
    user: str
    password: str
    db: str

    class Config:
        env_prefix = "POSTGRES_"

def get_db_settings(dotenv_path: Union[str, PathLike]) -> DatabaseSetting:
    loaded = load_dotenv(dotenv_path=dotenv_path, verbose=True, override=True)
    return DatabaseSetting()
```

maybe the "loading" part could take inspiration from it, just an idea, 

well either way that's a good idea and will reduce dependencies so thankls for that !

pydantic/env_settings.py
samuelcolvin: Please remove this and simply pass `_env_file` to `_build_values` as a second argument, then to `_build_environ`

acnebs:
@euri10 Thanks for the suggestion! I think in the end the `dotenv` approach is probably a bit much for this builtin functionality at the moment, which is why I initially went for something as simple as possible. I had also found [this library](https://github.com/David-Lor/Python-DotEnv-Settings-Handler) while researching different approaches, which does something similar to what you're doing.

But yeah, since Settings handling is a first-class use-case of Pydantic, I felt that it was much nicer to have it integrated into the lib without dependencies.

acnebs:
Ok, I think I've sorted everything out on the PR checklist. Let me know if anything else needs changing.

pydantic/env_settings.py
```diff
@@ -38,6 +46,12 @@ def _build_environ(self) -> Dict[str, Optional[str]]:
         else:
             env_vars = {k.lower(): v for k, v in os.environ.items()}
 
+        env_file = _env_file if _env_file != env_file_sentinel else self.__config__.env_file
+        if env_file is not None:
+            env_path = Path(env_file)
+            if env_path.is_file():
+                env_vars = {**env_vars, **read_env_file(env_path, self.__config__.case_sensitive)}
```
samuelcolvin:
```suggestion
                env_vars = {**read_env_file(env_path, self.__config__.case_sensitive), **env_vars}
```
my mistake, environment variables should definitely take precedence over env files.

tests/test_settings.py
```diff
@@ -348,6 +348,65 @@ def _build_values(self, init_kwargs):
...
+        b: str
+        c: str
+
+        class Config:
+            env_file = 'tests/env_files/test_settings_1.env'
+
+    env.set('A', 'overridden var')
+
+    s = Settings()
+    assert s.a == 'good string'
```
samuelcolvin: this should change once `_build_environ` is fixed.

acnebs:
@samuelcolvin Do we really want existing environment variables to take precedence over the `.env` file?

One of the major benefits of using `dotenv` is that it makes it far easier to ensure consistency of your settings. In an actual environment, there can be many different places / programs / scripts etc. modifying your vars. If you use the `.env` file, you can be sure that everything in that file will take precedence. The idea is that you're taking the file, and if it's valid, you're setting all the values inside it as env vars. At least that's how other libs like `python-dotenv` behave.

That makes the most sense to me, as otherwise you might have a bunch of vars in this file you're loading that are silently ignored, which seems like it would do more harm than good.

euri10:
> One of the major benefits of using `dotenv` is that it makes it far easier to ensure consistency of your settings. In an actual environment, there can be many different places / programs / scripts etc. modifying your vars. If you use the `.env` file, you can be sure that everything in that file will take precedence. The idea is that you're taking the file, and if it's valid, you're setting all the values inside it as env vars. At least that's how other libs like `python-dotenv` behave.

Don't quote me because  on that, I'm not too sure about it, but I think in python-dotenv you need to be explicit about precedence  using the override kwarg.

samuelcolvin:
No, environment variables should definitely take precedence over variables in `.env`.  As they do in other libraries:
* with create-react-app environment variables take priority, see [here](https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used)
* with starlette environment variables take priority, see [here](https://www.starlette.io/config/)
* I've just tried with `python-dotenv`, as expected environment variables take priority:

```
(env) pydantic 1 333ms ➤  pip install python-dotenv
...
(env) pydantic 0 971ms ➤  echo FOO=whatever >> .env
(env) pydantic 0  13ms ➤  echo BAR=other >> .env
(env) pydantic 05ms ➤  cat .env 
FOO=whatever
BAR=other
(env) pydantic 08ms ➤  export FOO=different
(env) pydantic 06ms ➤  ipython

Python 3.7.4
import asyncio, base64, math, hashlib, json, os, pickle, random, re, secrets, sys, time
from datetime import datetime, date, timedelta, timezone
from pathlib import Path

autoreload enabled, use `%autoreload 0` to disable

In [1]: from dotenv import load_dotenv

In [2]: load_dotenv()
Out[2]: True

In [3]: import os

In [4]: os.getenv('FOO')
Out[4]: 'different'

In [5]: os.getenv('BAR')
Out[5]: 'other'
```

samuelcolvin:
> In an actual environment, there can be many different places / programs / scripts etc. modifying your vars.

That what the env variable prefix is for.

samuelcolvin:
Hi @acnebs, are you still okay to finish this?

acnebs:
@samuelcolvin Yep, just been a bit busy over the past couple weeks. Should be able to get to it soon.

pydantic/env_settings.py
```diff
@@ -88,3 +103,19 @@ def prepare_field(cls, field: ModelField) -> None:
...
 
     __config__: Config  # type: ignore
+
+
+def read_env_file(file_path: Path, case_sensitive: bool = False) -> Dict[str, str]:
+    file_vars: Dict[str, str] = {}
+    for n, line in enumerate(re.split('[\r\n]+', file_path.read_text()), start=1):
+        line = line.strip()
+        if line and not line.startswith('#'):
+            m = re.match(r'^\s*(\w+)\s*=\s*(.+?)\s*$', line)
```
samuelcolvin:
```suggestion
            m = re.match(r'^(?:export )?\s*(\w+)\s*=\s*(.+?)\s*$', line)
```
Just occurred to me, as with python-dotenv we should allow and ignore lines starting with `export ` so `.env` files also work with `source`.

sm-Fifteen:
Any news about that PR? I'm pretty excided about getting rid of my custom `.env` config data extraction and casting logic so I can replace it with Pydantic instead.

samuelcolvin:
me too, if there's no progress I'll try and finish it myself when I get a chance.

@sm-Fifteen if you'd like to take over this, by all means create a new PR from this work. I'll make sure we give credit to both developers.

tests/env_files/test_settings_1.env
```diff
@@ -0,0 +1,6 @@
+# this is a comment
```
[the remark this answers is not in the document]
acnebs: Is it? It creates a lot of repetition in the tests to create similar files for each different one

docs/usage/settings.md
```diff
@@ -66,6 +66,45 @@ be if passed directly to the initialiser (as a string).
...
 by treating the environment variable's value as a JSON-encoded string.
 
+## Dotenv (.env) support
+
+Dotenv files (generally named `.env`) are a common pattern that make it easy to use environment variables in a platform-independent manner. A dotenv file follows the same general principles of all environment variables, and looks something like this:
+
+#### `.env`
+
+```
+# ignore comment
```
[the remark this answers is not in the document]
acnebs: As in just include the actual code we're using to parse env files in the docs, or...?

acnebs:
Sorry about the long delay, getting back to this now!

sm-Fifteen:
Looking at the code more closely, shouldn't we just use `python-dotenv` (as an optional dependency) for the file parsing instead of making our own spin on it, possibly with its own quirks and issues? Because parsing those files *correctly* looks awfully tricky.

I just checked and `python-dotenv`'s quoting behavior doesn't even match `dotenv`'s since mismatched quotes ([see this file](https://github.com/motdotla/dotenv/blob/master/tests/.env)) appear to assume the current line continues as a multiline string in the python lib, but the JS lib treats the lone quote as a literal character. I don't think the current logic implemented in this PR even handles mismatched quotes at all.

`python-dotenv` would at least ensure that the parsing is done consistently with other Python applications, without adding an extra maintenance burden on pydantic.

```py
import dotenv
# A dict, which can then be used to populate BaseSettings
env_dict = dotenv.dotenv_values('.env')
```

samuelcolvin:
No. For reasons I think I explained above.

We can cover 99.9% of usage very easily, if you want more you should probably use yaml or toml.

samuelcolvin:
upon reflection, I've decided to take on @sm-Fifteen's comment and switch to sing python-dotenv. While I don't like the extra (optional) dependency for one use, I can't ignore the problems you discuss regarding quotes etc.

@acnebs I'm sorry we spent so long building a dotenv parser that's now not being used, but I really do think this is a better approach.

sm-Fifteen:
I'm not seeing tests for types other than strings, does the current implementation support other types of field, such as ints, floats, booleans, optionals and possibly JSON-formatted `BaseModel`s?

samuelcolvin:
Yes of course, same as pydantic always does. Those values will be coerced to the field's type.

There is definitely at least one test which demonstrates this.

Lists etc. would require the [`Json`](https://pydantic-docs.helpmanual.io/usage/types/#json-type) type.

sm-Fifteen:
Yeah, ok, thank you for confirming this. That pretty much covers what I needed this to do, then.

acnebs:
@samuelcolvin Sounds good. Thanks for finishing it up!

layday:
`BaseSettings._build_values` used to be part of the public interface.  The docs said:

> By default ``BaseSettings`` considers field values in the following priority ... This behaviour can be changed by overriding the ``_build_values`` method on ``BaseSettings``.

Mention of this was removed in v1 and this PR breaks code which relied on `_build_values` to change the evaluation order.

samuelcolvin:
sorry about that I wasn't aware that `_build_values` had at one point been publicly documented.

PR welcome if you wish to update the release notes to include this.

layday:
Would you also accept a PR to allow customising the evaluation order in a future version?

samuelcolvin:
I guess, though I imagine it could be quite complex.

Quite related: I also want to allow multiple dot-env files with different priorities which are then merged, same as CRA.

Issue #997, opened 2019-11-14
Add public BaseModel.iter() method


Bobronium:
# Feature Request
Add `BaseModel.iter()` method which will behave exactly like a dict, but without converting nested values to dict by default and will be a generator instead

# Why
Imagine you want to iterate through values in model, but also need to exclude some, or use aliases instead of attributes. So what you need to do is write something like this:

```py
for attr, value in model.dict(exclude={...}, by_alias=True).items():
    do_something(attr, value)
```
Obviously, in this case python will firet evaluate `model.dict(exclude={...}, by_alias=True).` part, going through all values and only then you will be able to iterate, going through the same values again. 

Also, by using dict nested values are force-converted to dict, and this might be very slow when we just want to iterate through top-level values.

We also cannot use existing `_iter()`, because some work is actually done in `dict()`

# How we can solve it
By having `iter()` method we can bypass this double-iteration problem and converting-to-dict problems:
```py
for attr, value in model.iter(exclude={...}, by_alias=True):
    do_something(attr, value)
```

Selected Assignee: @hramezani

samuelcolvin:
This already exists!

It's called `._iter()`.

samuelcolvin:
Should be documented I guess.

MrMrRobat:
> We also cannot use existing _iter(), because some work is actually done in dict()

samuelcolvin:
I don't want to add more public method to models, best I think to implement this yourself.

If there's a simple way to achieve this with changes to existing functions of happy review a PR.

samuelcolvin:
Sorry, I missed your comment when replying in a hurry.

MrMrRobat:
> I don't want to add more public method to models, best I think to implement this yourself.

Just to be clear, thats valid even if `._iter()` itself would become public without adding any new methods? 

> If there's a simple way to achieve this with changes to existing functions of happy review a PR.

I guess the simplest way would be to move those excluding and aliases stuff from `.dict()` to `._iter()`, that will also simplify `.dict()` too.


I'll try to play with it soon and will be happy to make a PR once it'll be done.

dmontagu:
For what it's worth, I would be in favor of moving logic from `dict` into `_iter` -- I think it would simplify/result in a smaller diff for the implementation of `dump_as_type` from #812, and the related serialization performance improvements (even when not dumping as a specified type).

Kludex:
On V2, we can leverage `.model_dump()` as follows:

```python
[...]

Issue #655, opened 2019-07-11
Question: add private attribute


rrbarbosa:
I have a use case that I'd to add an attribute when initialising the instance which is not part of the model, thus should not be validated. Is that possible?

Here's a practical example:

```py
from pydantic import BaseModel
from datetime import datetime

class Test(BaseModel):
    a: int

class TestExtra(BaseModel):
    a: int

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._processed_at = datetime.utcnow()

test = {"a": 1}
Test(**test)
TestExtra(**test)  # ValueError: "TestExtra" object has no field "_processed_at"
```

samuelcolvin:
Yes, use underscore or `ClassVar`>

rrbarbosa:
This works, but it's not what I want, as every instance will have the same timestamp:
```
class TestExtra(BaseModel):
    a: int
    _procesed_at: datetime.utcnow()
```

This is what I want, but it fails with the same error:
```
class TestExtra(BaseModel):
    a: int
    _processed_at: datetime

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._processed_at = datetime.utcnow()

TestExtra(a=1) 
```

Or am I missing something?

H4dr1en:
@samuelcolvin can this be reopened and solved please?

samuelcolvin:
This should work:

```py
from datetime import datetime

from pydantic import BaseModel


class Test(BaseModel):
    a: int


class TestExtra(BaseModel):
    a: int

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        object.__setattr__(self, '_processed_at', datetime.utcnow())


test = {"a": 1}
t1 = Test(**test)
debug(t1)
t2 = TestExtra(**test)
debug(t2)
debug(t2._processed_at)
```

samuelcolvin:
`_processed_at` gets added to the model, so it will be included in `.dict()` etc. I'm not sure whether that's what you want?

We could add a `__internal__` property to models to cases like this (I think there are other user-cases but can't quite remember what they right now).

The usage would be more like:

```py
class TestExtra(BaseModel):
    a: int

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.__internal__['processed_at'] = datetime.utcnow()


t2 = TestExtra(a=123)
debug(t2.dict())
debug(t2.__internal__['processed_at'])
```
[...]

PR #1017, opened 2019-11-21
Refactor ._iter() method, 10x speed boost for dict(model)

MrMrRobat:
## Changes
- Fast yield from `.__dict__` on plain `.iter()` (x10 boost)
- As per #997, moved all keys-related stuff (`include`, `exclude`, etc.) to `._iter()`
- Removed iteration through default values in `.iter()`
- Almost all arguments checks moved out of loops, so checks happen once
- Removed redundant `set(dict.keys())` in `._calculate_keys()`
- Removed redundant `._keys_factory()` method

For tracking performance I used dataset from benchmarks. Seems like performance of `BaseModel.dict()` didn't change much.

```py
# benchmarks.run.generate_case

CASES = [generate_case() for _ in range(10)]


def test():
    for case in CASES:
        try:
            # benchmarks.test_pydantic.TestPydantic().model
            m = Model(**case)
        except ValidationError:
            pass
        else:
            print(timeit(partial(dict, m), number=100000))
 ```
`dict(m)` on `master`:

```py
    2.8283871970000005
    3.3839481399999993
    3.3260206760000006
    3.629596341000001
    4.001494594
    2.797958307000002
    2.6821828930000002
    3.092909658
    3.724842543000001
    3.0450761390000025
  ```
`dict(m)` on `_iter-refactor`:
```py
    0.2933802639999996
    0.28522739900000005
    0.28913957199999984
    0.29106095399999976
    0.2844331260000006
    0.289195844
    0.2902462040000007
    0.2863789990000001
    0.29688859900000075
    0.30014120099999886

```

## Checklist

* [x] Unit tests for the changes exist          | _do we need any tests for that_?


MrMrRobat:
Oh, no. `cython` CI just failed with 
```py
pydantic/main.py:669: in _iter
    yield from iteritems
pydantic/main.py:666: in genexpr
    for k, v in iteritems
E   ValueError: generator already executing
```
Seems like it's an __old__ cython bug, that won't be fixed until `cython 3.0`:
**Original issue:** https://github.com/cython/cython/issues/1159
**Possible workaround:** https://github.com/cython/cython/issues/1479

Or we can bet back to old-style iteration.

@samuelcolvin, @dmontagu what do you think?

MrMrRobat:
Ok, looks like we can't fix it, so I see these options:
1. Check all conditions on each iteration like it was before
2. Have to versions for `cython` and `python` (not preferred I guess)
3. Make things faster some other way

I have an idea how we can achieve 3rd option:
Instead of checking each value on iteration for every condition, we can perform it once on model creation, store it in special fields and then simply get allowed_keys in `._calculate_keys`:
`exclude_defaults` — `BaseModel().__default_fields__` 
`exclude_none` — `BaseModel().__null_fields__`
and then:
```py
if exclude_none:
    keys -= self.__null_fields__

if exclude_defaults:
    keys -= self.__default_fields__
```

samuelcolvin:
I'm not sure:
1. this would slow down validation even though very often `__null_fields__` and `__default_fields__` won't be used
2. They would also need to be kept up to date in `__setattr__`

Surely to work around the cython issue, the simplest solution would be to use a list for intermediate values of `iteritems`?

E.g. something like:
```py
    iteritems = [(k, v) for k, v in iteritems if k in allowed_keys]
```
?

Or would this loose all the performance improvement?

pydantic/main.py
```diff
@@ -629,53 +621,68 @@ def _iter(
...
+
         value_exclude = ValueItems(self, exclude) if exclude else None
         value_include = ValueItems(self, include) if include else None
 
+        iteritems: Union[ItemsView[str, Any], 'TupleGenerator'] = self.__dict__.items()
+        # adding only needed checks/conversions to generator
+        if allowed_keys is not None:
+            iteritems = ((k, v) for k, v in iteritems if k in allowed_keys)
+        if exclude_none:
+            iteritems = ((k, v) for k, v in iteritems if v is not None)
```
samuelcolvin: I guess even if we have one loop we could move this check to before `_get_value`.

pydantic/main.py
```diff
@@ -629,53 +621,68 @@ def _iter(
...
-                    include=value_include and value_include.for_element(k),
-                    exclude=value_exclude and value_exclude.for_element(k),
-                    exclude_unset=exclude_unset,
-                    exclude_defaults=exclude_defaults,
-                    exclude_none=exclude_none,
+            iteritems = ((k, v) for k, v in iteritems if self.__field_defaults__.get(k, _not_found) != v)
+        if by_alias:
+            fields = self.__fields__
+            iteritems = ((fields[k].alias if k in fields else k, v) for k, v in iteritems)
+        if to_dict or allowed_keys is not None:
```
samuelcolvin: I guess even if we have one loop we could avoid calling `_get_value` in many cases?

pydantic/main.py
MrMrRobat: Yeah, that surely good idea

pydantic/main.py
MrMrRobat: Yep. that's my point too 👍

MrMrRobat:
> Surely to work around the cython issue, the simplest solution would be to use a list for intermediate values of iteritems?
E.g. something like:
```py
    iteritems = [(k, v) for k, v in iteritems if k in allowed_keys]
```
It definitely will kill the performance, as it will actually construct new list in place for every condition. So not an option

> Then maybe we could change __iter__ to just return self.__dict__.items()?

Seems like the right thing to do 👍 


#### About `__null_fields__` and `__default_fields__`:
> this would slow down validation even though very often `__null_fields__` and `__default_fields__` won't be used

It will add only this to `validate_model`:
```py
if v_ is None:
    null_fields.add(name)
if v_ == field.default:
    default_fields.add(name)
```
I don't think it'll notably hit the performance. On the other hand `__null_fields__` and `__default_fields__` will be used at every `.dict()`, `.json()`, `.copy()` to improve the performance

> They would also need to be kept up to date in __setattr__

Good point. I guess, it's true only if `validate_assignments` enabled and won't be so hard to do.


How about I commit changes to a new branch, so you'll be able to see it and run needed  benchmarks? Or keep it in this one?

dmontagu:
For what it's worth, in general, I *would* be in favor of special casing the "default" cases for some of these recursive methods in order to achieve better performance. I think currently we are paying an unreasonably high price for checks that are nearly always unused. (Though, to be fair, pydantic is still obviously very fast.)

But I think we need to be aggressive about making sure it's done in a way that is maintainable.

------

Given the complexities noted by @samuelcolvin, I worry that adding `__null_fields__` and modifying various (mostly-unrelated) functions to keep things in sync is likely to be rather painful to maintain over time.

To the extent that it addresses the same performance issues, I would be much more in favor of just declaring a separate module-level function (e.g., called something like `_default_iter()`) which is used when the various kwargs don't add runtime checks to each recursive call, and delegating to that function when possible. That way, maintenance just amounts to ensuring this function is equivalent to `BaseModel._iter` when called with no arguments, so correctness can be verified in a single place (and it would be easy to write many test cases).

The approach described above seems to imply that you'd need to be thinking about the effects of many distantly-related functions to ensure correctness, which adds both to the cognitive burden of maintenance *and* to the difficulty of writing a comprehensive test suite since the logic becomes much more path-dependent (i.e., "unit" tests no longer cover most cases).

samuelcolvin:
Your rebase seems to have gone wrong, you have lots of files here which don't relate to this PR.

MrMrRobat:
Hope last commits fixed the issue. Sorry for inconvenience!

pydantic/main.py
```diff
@@ -674,23 +668,23 @@ def _iter(
...
+            yield k, v
 
     def _calculate_keys(
         self,
         include: Optional[Union['AbstractSetIntStr', 'DictIntStrAny']],
         exclude: Optional[Union['AbstractSetIntStr', 'DictIntStrAny']],
         exclude_unset: bool,
         update: Optional['DictStrAny'] = None,
-    ) -> Optional['SetStr']:
+    ) -> Optional[Union['SetStr', KeysView[str]]]:
```
```suggestion
    ) -> Optional[AbstractSet[str]]:
```
?

pydantic/main.py
```diff
@@ -638,33 +626,39 @@ def __iter__(self) -> 'TupleGenerator':
...
-            if allowed_keys is None:
-                allowed_keys = set(self.__fields__)
-            for k, v in self.__field_defaults__.items():
-                if self.__dict__[k] == v:
-                    allowed_keys.discard(k)
-
-        for k, v in self.__dict__.items():
-            if allowed_keys is None or k in allowed_keys:
-                value = self._get_value(
+        for k, v in self.__iter__():
```
```suggestion
        for k, v in self.__dict__.items():
```

pydantic/main.py
```diff
@@ -638,33 +626,39 @@ def __iter__(self) -> 'TupleGenerator':
...
+            if (
+                (allowed_keys is not None and k not in allowed_keys)
+                or (exclude_none and v is None)
+                or (exclude_defaults and self.__field_defaults__.get(k, _missing) == v)
+            ):
+                continue
+            if by_alias:
```
```suggestion
            if by_alias and k in self.__fields__:
                k = self.__fields__[k].alias
```

samuelcolvin: or perhaps
```suggestion
            if by_alias:
                try:
                    k = self.__fields__[k].alias
                except KeyError:
                    pass
```
Would be faster as you'd only need to do one look up of `self.__fields`

MrMrRobat: Catching exceptions is actually very slow, so it will be much slower if k is not in fields and slightly slower otherwise:
```py
In [1]: d = {str(i): i for i in range(1000)}
In [2]: %%timeit
   ...: try:
   ...:     k = d[1]
   ...: except KeyError:
   ...:     pass
   ...:
218 ns ± 1.83 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %%timeit
   ...: if 1 in d:
   ...:     k = d[1]
   ...:
32.8 ns ± 0.201 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %%timeit
   ...: try:
   ...:     k = d["1"]
   ...: except KeyError:
   ...:     pass
   ...:
52.7 ns ± 0.43 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: %%timeit
   ...: if 1 in d:
   ...:     k = d["1"]
   ...:
32.7 ns ± 0.303 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

So I think, better stick to first approach.

File: pydantic/main.py (hunk 3 above)
Comment (MrMrRobat) [reply pydantic/main.py]: Looks great. Should I change it in other places?

File: pydantic/main.py [hunk 6]
```diff
@@ -638,33 +626,39 @@ def __iter__(self) -> 'TupleGenerator':
         """
         so `dict(model)` works
         """
-        yield from self._iter()
+        yield from self.__dict__.items()
```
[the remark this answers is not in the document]
MrMrRobat: Isn't #1139 already conflict with current logic? 
I think it would be easier to solve conflicts after PR is merged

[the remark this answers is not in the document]
samuelcolvin: agree, see my suggestion on #655.

pydantic/main.py
samuelcolvin: Yes please, if in this file.

pydantic/main.py
samuelcolvin: ok

MrMrRobat:
I guess, all ready 🎉

pydantic/main.py
```diff
@@ -638,33 +626,40 @@ def __iter__(self) -> 'TupleGenerator':
...
-                value = self._get_value(
+            if (
+                (allowed_keys is not None and k not in allowed_keys)
+                or (exclude_none and v is None)
+                or (exclude_defaults and self.__field_defaults__.get(k, _missing) == v)
+            ):
+                continue
+            if by_alias and k in self.__fields__:
+                k = self.__fields__[k].alias
+                k = self.__fields__[k].alias if k in self.__fields__ else k
```
```suggestion
```
I don't think this line is required?

MrMrRobat: Yep, good catch!

samuelcolvin:
awesome, thank you.

PR #1027, opened 2019-11-24
Support typing.Literal in python 3.8

### LINKED ISSUE #1026: Cannot use Literal from standard library in 3.8
# Bug

* OS: macOS
* Python version `import sys; print(sys.version)`: 3.8.0
* Pydantic version `import pydantic; print(pydantic.VERSION)`: 1.1.1

Annotating models with the built-in literal in Python 3.8 in the absence of `typing_extensions` does not work because `is_literal_type` only checks if `Literal` is available from `typing_extensions`.  Please see the snippet below.

```term
$ cat test.py
from typing import Literal


class Foo(BaseModel):
    bar: Literal['']


print('ok!')
$ python3 test.py
  File "test.py", line 5, in <module>
    class Foo(BaseModel):
  File "/Users/layday/code/instawow/.venvs/3.8/lib/python3.8/site-packages/pydantic/main.py", line 192, in __new__
  File "/Users/layday/code/instawow/.venvs/3.8/lib/python3.8/site-packages/pydantic/fields.py", line 260, in infer
    return cls(
  File "/Users/layday/code/instawow/.venvs/3.8/lib/python3.8/site-packages/pydantic/fields.py", line 237, in __init__
    self.prepare()
  File "/Users/layday/code/instawow/.venvs/3.8/lib/python3.8/site-packages/pydantic/fields.py", line 307, in prepare
    self._type_analysis()
  File "/Users/layday/code/instawow/.venvs/3.8/lib/python3.8/site-packages/pydantic/fields.py", line 359, in _type_analysis
    if issubclass(origin, Tuple):  # type: ignore
  File "/nix/store/mnl1njrv6g2jcs69kq0cy19766kcgc5n-python3-3.8.0/lib/python3.8/typing.py", line 771, in __subclasscheck__
    return issubclass(cls, self.__origin__)
$ pip3 install typing_extensions
Collecting typing_extensions
  Using cached https://files.pythonhosted.org/packages/03/92/705fe8aca27678e01bbdd7738173b8e7df0088a2202c80352f664630d638/typing_extensions-3.7.4.1-py3-none-any.whl
Installing collected packages: typing-extensions
Successfully installed typing-extensions-3.7.4.1
WARNING: You are using pip version 19.2.3, however version 19.3.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
$ python3 test.py
ok!
```

dmontagu:
## Change Summary

Attempt to use `typing.Literal` if available (e.g., with python 3.8).

## Related issue number

Closes #1026 

## Checklist

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -28,11 +28,6 @@ from .typing import AnyType, Callable, ForwardRef, NoneType, display_as_type, is
 from .utils import PyObjectStr, Representation, lenient_issubclass, sequence_like
 from .validators import constant_validator, dict_validator, find_validators, validate_json
 
-try:
-    from typing_extensions import Literal
-except ImportError:
-    Literal = None  # type: ignore
-
 Required: Any = Ellipsis
 
 if TYPE_CHECKING:
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -48,18 +48,13 @@ else:
 
     AnyCallable = TypingCallable[..., Any]
 
-try:
-    from typing import Literal as TypingLiteral
-except ImportError:
-    TypingLiteral = None  # type: ignore
-
-try:
-    from typing_extensions import Literal as TypingExtensionsLiteral
-except ImportError:
-    TypingExtensionsLiteral = None  # type: ignore
-
-Literal = TypingLiteral or TypingExtensionsLiteral  # prefer TypingLiteral if possible
-
+if sys.version_info < (3, 8):
+    try:
+        from typing_extensions import Literal
+    except ImportError:
+        Literal = None  # type: ignore
+else:
+    from typing import Literal
 
 if TYPE_CHECKING:
     from .fields import ModelField
@@ -159,10 +154,9 @@ def is_callable_type(type_: AnyType) -> bool:
 
 
 if sys.version_info >= (3, 7):
-    literal_types = {type_ for type_ in (TypingLiteral, TypingExtensionsLiteral) if type_ is not None}
 
     def is_literal_type(type_: AnyType) -> bool:
-        return Literal is not None and getattr(type_, '__origin__', None) in literal_types
+        return Literal is not None and getattr(type_, '__origin__', None) is Literal
 
     def literal_values(type_: AnyType) -> Tuple[Any, ...]:
         return type_.__args__
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -49,9 +49,16 @@ else:
     AnyCallable = TypingCallable[..., Any]
 
 try:
-    from typing_extensions import Literal
+    from typing import Literal as TypingLiteral
 except ImportError:
-    Literal = None  # type: ignore
+    TypingLiteral = None  # type: ignore
+
+try:
+    from typing_extensions import Literal as TypingExtensionsLiteral
+except ImportError:
+    TypingExtensionsLiteral = None  # type: ignore
+
+Literal = TypingLiteral or TypingExtensionsLiteral  # prefer TypingLiteral if possible
 
 
 if TYPE_CHECKING:
@@ -152,9 +159,10 @@ def is_callable_type(type_: AnyType) -> bool:
 
 
 if sys.version_info >= (3, 7):
+    literal_types = {type_ for type_ in (TypingLiteral, TypingExtensionsLiteral) if type_ is not None}
 
     def is_literal_type(type_: AnyType) -> bool:
-        return Literal is not None and getattr(type_, '__origin__', None) is Literal
+        return Literal is not None and getattr(type_, '__origin__', None) in literal_types
 
     def literal_values(type_: AnyType) -> Tuple[Any, ...]:
         return type_.__args__
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -49,10 +49,13 @@ else:
     AnyCallable = TypingCallable[..., Any]
 
 if sys.version_info < (3, 8):
-    try:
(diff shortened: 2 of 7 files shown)


pydantic/typing.py
```diff
@@ -152,9 +159,10 @@ def is_callable_type(type_: AnyType) -> bool:
 
 
 if sys.version_info >= (3, 7):
+    literal_types = {type_ for type_ in (TypingLiteral, TypingExtensionsLiteral) if type_ is not None}
 
     def is_literal_type(type_: AnyType) -> bool:
-        return Literal is not None and getattr(type_, '__origin__', None) is Literal
+        return Literal is not None and getattr(type_, '__origin__', None) in literal_types
```
dmontagu: In the case where the set has a single element (i.e., python 3.7), I benchmarked the `in` check vs the old `is` check at ~0.4s vs 0.5s slower over 10,000,000 iterations. This absolute magnitude of the time difference seems small enough to be worth the simplification, rather than adding another branch for `sys.version_info >= (3, 8)`.

But I could refactor to do that check instead if preferable.

pydantic/typing.py
```diff
@@ -49,9 +49,16 @@ def evaluate_forwardref(type_, globalns, localns):  # type: ignore
...
 except ImportError:
-    Literal = None  # type: ignore
+    TypingLiteral = None  # type: ignore
+
+try:
+    from typing_extensions import Literal as TypingExtensionsLiteral
+except ImportError:
+    TypingExtensionsLiteral = None  # type: ignore
+
+Literal = TypingLiteral or TypingExtensionsLiteral  # prefer TypingLiteral if possible
```
dmontagu: Since we use the `Literal` type from this package in some places in the code, we prioritize the built-in, which should always be available if supported.

pydantic/typing.py
dmontagu: I also confirmed having `literal_types` be a set was ~20% faster than having it be a tuple, even with only two elements.

dmontagu: Also, I'd rather not *require* the use of `typing.Literal` instead of `typing_extensions.Literal` when using python 3.8, since that will make it more annoying to migrate to python 3.8.

layday: `typing_extensions` imports `Literal` from `typing` in 3.8 so you can keep the old identity check and the effect will be the same.

samuelcolvin: > `typing_extensions` imports `Literal` from `typing` in 3.8 so you can keep the old identity check and the effect will be the same.

great, that should make all these checks simpler.

pydantic/fields.py
```diff
@@ -28,11 +28,6 @@
 from .utils import PyObjectStr, Representation, lenient_issubclass, sequence_like
 from .validators import constant_validator, dict_validator, find_validators, validate_json
 
-try:
-    from typing_extensions import Literal
-except ImportError:
-    Literal = None  # type: ignore
```
dmontagu: looks like this was not actually used in this module

pydantic/typing.py
```diff
@@ -48,11 +48,13 @@ def evaluate_forwardref(type_, globalns, localns):  # type: ignore
 
     AnyCallable = TypingCallable[..., Any]
 
-try:
-    from typing_extensions import Literal
-except ImportError:
-    Literal = None  # type: ignore
-
+if sys.version_info < (3, 8):
```
dmontagu: Needed to use the `sys.version_info` check for mypy compatibility with both 3.8 and <=3.7

pydantic/utils.py
```diff
@@ -21,11 +21,6 @@
 
 from .typing import AnyType, display_as_type
 
-try:
-    from typing_extensions import Literal
-except ImportError:
-    Literal = None  # type: ignore
```
dmontagu: Looks like it was unused in this module too.

pydantic/typing.py
```diff
@@ -48,11 +48,16 @@ def evaluate_forwardref(type_, globalns, localns):  # type: ignore
...
+    if TYPE_CHECKING:
+        from typing_extensions import Literal
+    else:  # due to different mypy warnings raised during CI for python 3.7 and 3.8
```
dmontagu: Had to add this gross `if TYPE_CHECKING:` block here because the `Literal = None` line was causing an "unnecessary ignore" warning during the 3.8 build. This approach seems fine to me here, but I'm open to alternatives.

dmontagu:
@samuelcolvin Modulo the remaining ugly `if TYPE_CHECKING:` block in the import of `Literal`, I think this is good to go.

samuelcolvin:
awesome, thank you.

pydantic/typing.py
layday: This is a [known issue](https://github.com/python/mypy/issues/5940) in mypy with one developer commenting that `--warn-unused-ignores` is not recommended for use in CI. But barring that you can trick mypy into interpreting `Literal` as a `Union` with:

```py
if sys.version_info < (3, 8):
    try:
        from typing_extensions import Literal as _Literal
    except ImportError:
        Literal = None
    else:
        Literal = _Literal
else:
    from typing import Literal
```

dmontagu: Thanks for sharing the issue.

I definitely would prefer to keep that check on in CI though; otherwise it's too easy to accidentally forget, and then end up with confusion about what is being ignored and/or ignore a future error that should have been caught.

samuelcolvin: agreed, I hoped mypy would have a `type: ignore flexible` option, but couldn't find anything.

PR #1034, opened 2019-11-26
Add model signature generation

### LINKED ISSUE #1032: Generate actual signature for BaseModel.__init__
# Feature Request
That idea came for me when I saw this issue:
https://github.com/tiangolo/fastapi/issues/318
And this particular comment that shows that standart inspect tools unable to determine signature of a pydantic model:
> @tiangolo, I had an opportunity to fiddle a little bit more with this today.
> Here is a full example of the different ways to specify parameters using depends: https://gist.github.com/Atheuz/075f4d8fe3b56d034741301ba2574ef1
> 
> You can run it using this command `uvicorn run:app --reload --host 0.0.0.0 --port 8080`
> 
> I essence: dataclass classes work if I specify `Query(None)` and `Depends`, pure classes work if I specify `Query(None)` and use `Depends`, and input parameters in the signature just works.
> 
> Pydantic classes do not work, at least in terms of the generated docs, it just says `data` instead of the expected `dt` and `to_sum`.
> 
> dataclasses in the generated docs:
> billede
> 
> pydantic in the generated docs:
> billede

This, however is not true for dataclasses, where `__init__` is generated on class creation.
```py
@dataclass
class Dataclass:
    id: int
    name: str = 'John'

class Model(BaseModel):
    id: int
    name: str = 'John'

help(Dataclass.__init__)
'''
Help on function __init__:
__init__(self, id: int, name: str = 'John') -> None
'''
help(Model.__init__)
'''
Help on function __init__ in module pydantic.main:
__init__(__pydantic_self__, **data: Any) -> None
    Initialize self.  See help(type(self)) for accurate signature.
'''
```

Of course, we cannot follow dataclasses in this part and generate `__init__` of `BaseModel`, since it will cause getting `TypeError` instead of `ValidationError` on missing values, etc., but we still can help inspect tools to see the right signature of our models.

```py
from functools import update_wrapper
import pydantic

def __init__(self, *, id: int, name: str = 'John') -> None: ...
update_wrapper(Model.__init__, __init__)
help(Model.__init__)
'''
Help on function __init__ in module __main__:
__init__(self, *, id: int, name: str = 'John') -> None
'''
```
# Possible approaches
I see this ways of creating typed `__init__`:
- Generate it as a string and use `eval()`
- Create function normally and modify its __code__ signature 
and then use `functools.update_wrapper(Model.__init__, typed_func)` 
# Example
```py
# using `Model` declared above

def f(): ...

f.__code__ = f.__code__.replace(   # __code__.replace() available only in python >= 3.8
    co_kwonlyargcount=len(Model.__fields__),
    co_argcount=1,
    co_varnames=('__pydantic_self__',) + tuple(Model.__fields__)
)

defaults, annotations = {}, {}
for name, field in Model.__fields__.items():
    if field.default:
        defaults[name] = field.default
    annotations[name] = field.type_
    
f.__kwdefaults__ = defaults
f.__annotations__ = annotations
help(f)
'''
Help on function f in module __main__:
f(__pydantic_self__, *, id: int, name: str = 'John')
'''
```

MrMrRobat:
## Change Summary
Generating model signature based on its fields

```py
import inspect

class FooModel(BaseModel):
    id: int
    name: str = None
    description: str = 'Foo'

print(inspect.signature(FooModel))
'''
(*, id: int, name: str = None, description: str = 'Foo') -> None
'''
```

### Respects custom-defined init signature:
```py
import inspect

class MyModel(BaseModel):
    id: int
    info: str = 'Foo'

    def __init__(self, id: int = 1, *, bar: str, **data) -> None:
        super().__init__(id=id, bar=bar, **data)

print(inspect.signature(MyModel))
'''
(id: int = 1, *, bar: str, info: str = 'Foo') -> None
'''
```

## Related issue number
Closes #1032, https://github.com/timothycrosley/hypothesis-auto/issues/10
Relates to tiangolo/fastapi#318
# ToDo:
* [x] Handle overlays in custom `__init__`
* [x] Rewrite signature tests
* [x] Fix issues with cython
* [x] Replace/ignore invalid params names in fields (https://github.com/samuelcolvin/pydantic/issues/1032#issuecomment-558538776)
* [x] Add something in docs?

## Checklist


pydantic/main.py
```diff
@@ -238,7 +249,46 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
...
+        # cannot modify cls.__init__ directly,
+        # as it shared between children classes, so we copy it
+        cls.__init__ = new_init = FunctionType(
+            orig_code, orig_init.__globals__, closure=orig_init.__closure__, name=orig_init.__name__
+        )
+
+        orig_posonlycount = orig_code.co_posonlyargcount if PY38 else 0
+        orig_argnames = orig_code.co_varnames[: orig_code.co_argcount + orig_code.co_kwonlyargcount + orig_posonlycount]
+        orig_argnames_set = set(orig_argnames)
+        fields_names = tuple([name for name in fields if name not in orig_argnames_set])
```
samuelcolvin: this might not maintain order of `orig_argnames` I believe.

pydantic/main.py
```diff
@@ -238,7 +249,46 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
...
+        del orig_argnames_set
+
+        def fake_init():
```
samuelcolvin: I'm not clear why we need this?

pydantic/utils.py
```diff
@@ -114,6 +135,12 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
     return abs(value_1 - value_2) <= delta
 
 
+def copy_code(code: CodeType, **update: Any) -> CodeType:
+    if PY38:  # pragma: no cover
```
samuelcolvin:
```suggestion
    if PY38:
```
This should be tested.

tests/test_main.py
```diff
@@ -1022,3 +1023,72 @@ class NewModel(DerivedModel, something=2):
         something = 1
 
     assert NewModel.something == 2
+
+
+def test_init_signature(capsys):
```
samuelcolvin: new test file please.

tests/test_main.py
```diff
@@ -1022,3 +1023,72 @@ class NewModel(DerivedModel, something=2):
...
+        ('a', kw_only, int, empty),
+        ('b', kw_only, float, 10),
+        ('c', kw_only, SubModel, empty),
+    ):
+        assert name in params
+        param: inspect.Parameter = params[name]
+        assert param.kind is kind
+        assert param.default == default
+
+    assert Model.__init__.__name__ == '__init__'
```
samuelcolvin: can we add

```py
assert help(Model.__init__) == """
...
"""
```
Or equivalent to provide an easy to read check?

pydantic/main.py
```diff
@@ -238,7 +249,46 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
...
+                co_kwonlyargcount=orig_code.co_kwonlyargcount + len(fields_names),
+                co_names=orig_code.co_names + fields_names,
+                co_varnames=orig_argnames + fields_names,
+                co_name=orig_code.co_name,
+                **({'co_posonlyargcount': orig_posonlycount} if PY38 else {}),
+            ),
+            orig_init.__globals__,
+            name=orig_init.__name__,
+        )
+        fake_init.__doc__ = new_init.__doc__ = orig_init.__doc__
```
samuelcolvin: Would it be a good idea to add a helpful docstring to the real `__init__` which by default appears in `help(MyModel.__init__)`?

pydantic/main.py
```diff
@@ -238,7 +249,46 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
...
-        return super().__new__(mcs, name, bases, new_namespace, **kwargs)
+
+        cls = super().__new__(mcs, name, bases, new_namespace, **kwargs)
+
+        # Generating right signature for __init__:
+        orig_init = cls.__init__
+        orig_code = orig_init.__code__
+        # cannot modify cls.__init__ directly,
+        # as it shared between children classes, so we copy it
+        cls.__init__ = new_init = FunctionType(
```
samuelcolvin: could we move from here on into a separate function?

pydantic/main.py
MrMrRobat: I weren’t able to construct it from scratch. Do you have any ideas how it should be don?

pydantic/main.py
MrMrRobat: It doesn’t do anything with orig_argnames, but excluding them from fields. This is actually done to maintain their order.

pydantic/main.py
MrMrRobat: Yeah, good idea. Where do you think is the best place for it? main, utils?

tests/test_main.py
MrMrRobat: I’ve done in first commit, but because of different formatting, tests on CI failed. There were inconsistency with spaces between arguments. Probably we can fix it by comparing strings without spaces

pydantic/utils.py
MrMrRobat: I guess I need a little help here. How do we test such things?

tests/test_main.py
samuelcolvin: Or different tests for different versions of python.

pydantic/utils.py
samuelcolvin: Ci already runs tests for all supported versions of python which are then combined.

pydantic/main.py
MrMrRobat: It’s not bad idea for sure. What should be written there?

pydantic/main.py
samuelcolvin: Don't really mind, either typing or utils I guess.

pydantic/main.py
```diff
@@ -238,7 +249,46 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
...
+        orig_argnames_set = set(orig_argnames)
+        fields_names = tuple([name for name in fields if name not in orig_argnames_set])
+        del orig_argnames_set
+
+        def fake_init():
+            ...
+
+        fake_init = FunctionType(
+            copy_code(
+                fake_init.__code__,
```
dmontagu: Can you not just use `cls.__init__.__code__` here, and drop the `def fake_init(): ...` above?

pydantic/main.py
dmontagu: Does the above not result in `cls.__init__` being modified too? Not sure why `new_init` is necessary here.

pydantic/main.py
MrMrRobat: Nope, because they have incompatible signatures it fails

pydantic/main.py
MrMrRobat: It not modifies parent classes `__init__`, it replaces it in a new class.
`new_init` here just as a variable, so we don't have to write `cls.__init__` each time

pydantic/main.py
dmontagu: Makes sense

pydantic/main.py
samuelcolvin: Not sure, something like:

> Create a new model by parsing and validating input data from keyword arguments, a `ValidationError` will be raised if the input data cannot be parsed to for a valid model.

Or something.

pydantic/main.py
MrMrRobat: Managed to get rid of this one :)

pydantic/main.py
MrMrRobat: Also adding info about original signature to docstring on generation

tests/test_main.py
MrMrRobat: Done!

pydantic/main.py
MrMrRobat: Done!

tests/test_main.py
MrMrRobat: Done!

pydantic/main.py
```diff
@@ -229,7 +240,7 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
         new_namespace = {
             '__config__': config,
             '__fields__': fields,
-            '__field_defaults__': {n: f.default for n, f in fields.items() if not f.required},
+            '__field_defaults__': fields_defaults,
```
samuelcolvin: Has `__field_defaults__` changed at all?

pydantic/utils.py
```diff
@@ -114,6 +137,84 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+    valid_fields_args = tuple([name for name in fields if name not in orig_argnames_set and name.isidentifier()])
+    del orig_argnames_set
+
+    fake_init = FunctionType(
+        copy_code(
+            orig_code,
+            co_argcount=orig_code.co_argcount,
+            co_kwonlyargcount=orig_code.co_kwonlyargcount + len(valid_fields_args),
+            co_names=orig_code.co_names,
+            co_varnames=orig_argnames + valid_fields_args + orig_othernames,
```
samuelcolvin: Field order is wrong if we define a model like

```py
class Foo(BaseModel):
    a: int 
    b: str = 'xxx' 
    c: bool 
```
Which gives a signature of 
```
(__pydantic_self__, *, a: int, b: str = 'xxx', c: bool, **data: Any) -> None
```

pydantic/utils.py
```diff
@@ -114,6 +137,84 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+            **({'co_posonlyargcount': orig_posonlycount} if PY38 else {}),
+        ),
+        orig_init.__globals__,
+        name=orig_init.__name__,
+        closure=orig_init.__closure__ or () if PY38 else (),
+    )
+
+    doc = orig_init.__doc__ or ''
+    real_signature = inspect.signature(orig_init)
+    new_doc = f'Signature is generated based on model fields. Real signature:\n    {real_signature}\n'
```
samuelcolvin: I'm not sure adding the real signature here is very helpful.

I think we should just use `orig_init.__doc__`? Perhaps with `...\n(signature auto generated from model fields)` appended or something.

tests/test_init_signature.py
```diff
@@ -0,0 +1,125 @@
...
+def test_init_signature():
+    class Model(BaseModel):
+        a: float
+        b: int = 10
+
+    assert BaseModel.__init__ is not Model.__init__
+    sig = signature(Model.__init__)
+    assert sig != signature(BaseModel.__init__)
+
+    params = sig.parameters
```
samuelcolvin: I think this would be much easier to read and debug with something like 

```py
assert {n: str(p) for n, p in s.parameters.items()} = {
    ...
}
```

tests/test_init_signature.py
```diff
@@ -0,0 +1,125 @@
...
+            self.bar = bar
+            self.baz = baz
+
+    m = MyModel(foo=2, id=1, baz='Ok!')
+    assert m.id == 1
+    assert m.bar == 2
+    assert m.baz == 'Ok!'
+    assert m.name == 'John Doe'
+
+    sig = signature(MyModel.__init__)
```
samuelcolvin: as above.

pydantic/utils.py
MrMrRobat: I don't see anything wrong here. Where's the wrong part?

pydantic/main.py
MrMrRobat: Nope, just saw that there was unnecessary iteration through fields and replaced it.

pydantic/utils.py
MrMrRobat: If you talking about the non-default parameter `c` after default `b`, there's nothing wrong since all of them are keyword-only:
```py
def f(*, a=1, b): ...
help(f)
'''
Help on function f in module __main__:

f(*, a=1, b)
'''
```

tests/test_init_signature.py
MrMrRobat: Yeah, thats' right. But I'd went with
```py
assert [str(p) for p in sig.parameters.values()] == [
        'self',
        'id: int = 1',
        'bar=2',
        'baz: Any',
        "name: str = 'John Doe'",
        '**data',
    ]
```
to maintain the order.

MrMrRobat:
Oops, probably merging that test was a bad idea...

tests/test_init_signature.py
samuelcolvin: Okay

MrMrRobat: Done!

tests/test_init_signature.py
MrMrRobat: Done!

pydantic/utils.py
MrMrRobat: Done!

pydantic/utils.py
```diff
@@ -114,6 +137,79 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+    """
+    if not compiled:
+        # can perform this in python, however in cython this will cause "unknown opcode" on call
+        orig_init = current_init
+        orig_code = orig_init.__code__
+        new_init = FunctionType(
+            orig_code, orig_init.__globals__, closure=orig_init.__closure__, name=orig_init.__name__
+        )
+    else:
+        orig_init = getattr(current_init, '__origin_init__', current_init)
```
dmontagu: Does this fork not work if not compiled? Just wondering why we can't use the same logic in both cases (it's totally fine if we have to, I'm just trying to understand)

MrMrRobat: I guess this would work, but it would also add redundant call to a stack and probably may cause confusion on step into model creation while debugging. Current behavior just will lead debugger right into original __init__ and won’t add any redundant calls, so I decided to stick with it.

dmontagu: I see where you are coming from.

Possible counter-point: this also means that there could be bugs that show up in the compiled version and not the un-compiled version, or vice versa. I think it would be extremely annoying to figure out what was going wrong in a reported issue if it dependended on the compiled status. Do you still think it makes sense to keep the fork given that consideration?

@samuelcolvin I'd also be interested to get your opinion here.

MrMrRobat: Good point. Personally I don't see a problem there, but this is surely need to be considered.

pydantic/utils.py
```diff
@@ -114,6 +137,79 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+        new_init = __init__  # type: ignore
+        new_init.__origin_init__ = orig_init  # type: ignore
+
+    orig_posonlycount = orig_code.co_posonlyargcount if PY38 else 0  # type: ignore
+    orig_allargscount = orig_code.co_argcount + orig_code.co_kwonlyargcount + orig_posonlycount
+    orig_argnames = orig_code.co_varnames[:orig_allargscount]
+    orig_othernames = orig_code.co_varnames[orig_allargscount:]
+
+    orig_argnames_set = set(orig_argnames)  # just for fast lookups
+    valid_fields_args = tuple([name for name in fields if name not in orig_argnames_set and name.isidentifier()])
```
dmontagu: Does this generate the right signature if a field has an alias and doesn't have `allow_population_by_field_name=True`?

pydantic/utils.py
samuelcolvin: good point, just me getting confused.

pydantic/utils.py
MrMrRobat: Aliases are not considered at all.

And thanks for that question, I found a bug in current implementation. Model with fields declared in namespace would have this in signature:
```py
__init__(__pydantic_self__, *, a=FieldInfo(default=1, alias='A', extra={}), **data: Any) -> None
```

Can you please give me example of such model and what signature should look like?

pydantic/utils.py
samuelcolvin: My problem here is not just about different routes, but also about different signatures:

Running the following code:
```py
from datetime import datetime
from typing import List

class User(BaseModel):
    id: int
    name = 'John Doe'
    signup_ts: datetime = None
    friends: List[int] = []

user = User()
```

With pydantic not compiled I get:
```
  File "test.py", line 14, in <module>
    user = User()
  File "/home/samuel/code/pydantic/pydantic/main.py", line 293, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for User
id
```

But when compiled I get:
```
  File "test.py", line 14, in <module>
    user = User()
  File "pydantic/utils.py", line 169, in pydantic.utils.generate_typed_init.__init__
    orig_init(*args, **kwargs)
  File "pydantic/main.py", line 293, in pydantic.main.BaseModel.__init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for User
id
```

I think it's important these two tracebacks look the same and I think the traceback should **not** have the extra `pydantic.utils.generate_typed_init.__init__...` line.

I see three possible solutions to this:
1. Find a way to re arrange this function to avoid this extra step. maybe not possible.
2. Find another work around for this problem, e.g.:
  * just modify the docstring to detail the arguments
  * implement our own custom `signature()` method that provides a sensible signature for a model, without changing the actual model code
3. Manually customise the traceback so it's always the same, and perhaps even more simplified, e.g. just

```
  File "test.py", line 14, in <module>
    user = User()
pydantic.error_wrappers.ValidationError: 1 validation error for User
id
```
or something.

What do you think?

MrMrRobat: You should look here for reasons why in cpython we cannot achieve same traceback as in python:
https://stackoverflow.com/questions/59072875/how-to-clone-function-in-cython-i-getting-systemerror-unknown-opcode/59087135#59087135

samuelcolvin: Yes, i see that, I tried this myself I got something to work see 523038348177258bf83f8f75466bdc0a3a8c34c1 or the [init-signature-sandbox](https://github.com/samuelcolvin/pydantic/compare/init-signature-sandbox?expand=1) branch. But I think that approach has other problems.

I think the best solution is our own utility function to generate the signature for an init function. That `inspect.signature()` would be the same as now, but `pydantic.inspect.signature()` (or whatever) would give the new improved signature. Would that work for the original use case?

MrMrRobat: I don't get how exactly `pydantic.inspect.signature()` would help us to solve copying problem.

Also, your approach would break models like this:

```py

class MyModelsBase(BaseModel):
    def __init__(self, **data):
        super().__init__(self, **data)
        print(f'Initialized {self} with {data}')

class MyModel(MyModelsBase):
    a: int

MyModel(a=a)   # no output, as on MyModel creation '__init__' was not in namespace and was replaced

Comment (MrMrRobat) [reply pydantic/utils.py]: > 2\. Find another work around for this problem, e.g.:
> 
> * just modify the docstring to detail the arguments
> * implement our own custom `signature()` method that provides a sensible signature for a model, without changing the actual model code

Even if we modify docstring to detail the arguments, or implement our own argument, they still will be modified on the same `__init__` of a base model, so every declared model will rewrite it and eventually it'll end up with the docstring and signature of the last model created.

Comment (MrMrRobat) [reply pydantic/utils.py]: I also thought about declaring `__annotations__`, `__kwdefaults__`, `__wrapped__` and `__signature__` as python `property` so they will give signatures of current models, but that doesn't seem possible too, because `WhateverModel.__init__` would still be bound to `BaseModel` and I don't see any way how these properties could know what model's signature to give on call

File: pydantic/utils.py (hunk 15 above)
Comment (dmontagu) [reply pydantic/utils.py]:
```python
class A(BaseModel):
    a: int = Field(alias='b')
```
It looks like your generated signature will be (ignoring any type hints):
```python
__init__(__pydantic_self__, a)
```
but if you call `__init__(a=1)` you'll get an error. I would think the signature *should * be `__init__(__pydantic_self__, b)` in this case. This is what I meant by aliases.

dmontagu: Also, note that if you are using the type annotation from the field, the generated signature will actually also be wrong for any type that is not idempotent. In particular, at least currently, if you have a `SecretStr` field, you *have* to pass a `str` to the `__init__` function and *not* a `SecretStr`.  I opened an issue to fix this for `SecretStr`, but there are other types for which this is also an issue.

(Admittedly, this is also a problem in the PyCharm plugin and, in strict mode, the mypy plugin too.)

MrMrRobat: Ok, now I see. Aliases should be considered.

MrMrRobat: As for `SecretStr` and other custom types: I don't see any problems here since these annotations will be available only at runtime and won't bother type-checkers. Until some code checks incoming types at runtime and don't know how to deal with `pydantic`, I don't think having `pydantic` specific types in signature would be a problem. For example, `FastAPI` understands such annotations just fine:

```py
class GetUserFilters(BaseModel):
    name: Optional[constr(strip_whitespace=True, max_length=70)]
    email: Optional[SecretStr]

 
@app.get('/users')
def get_users(params=Depends(GetUserFilters)):
    ...
```
image

dmontagu: I agree they won't bother type checkers, but it still feels bad for it to be generated "wrong". I think there is a tradeoff here between it faithfully representing the valid signature of the method (i.e., including a type annotation of `Any` or similar to allow for parsing), and providing useful information.

In general, I would prefer to err on the side of "correctness" (e.g., generating an `__init__` signature with `Any` as the annotation), but from a practical perspective I'm okay with it being occasionally wrong in a minor way like this if it is substantially more useful (and preferred in practice).

samuelcolvin:
Sorry having thought about this I think it's too complicated and risky and could lead to intractable problems.

I'm not not happy about the differences between compiled and un-compiled cases and the extra step in the traceback.

---

In stead I think we should take `generate_typed_init` and move/change it into a new standalone function in `pydantic.tools` called `inspect` which returns a `Signature` object when passed a `Model` or `Model.__init__` if possible.

This will be less intuitive since `inspect.signature` won't give the best signature possible, but provided it's properly document it should suffice and avoids lots of thorny problems.

@dmontagu is fastapi currently using `inspect.signature`? If so could it be changed to use `pydantic.tools.signature`?

`pydantic.tools.signature` could be setup to fall back to `inspect.signature` if the argument is not a subclass of `BaseModel` etc.

dmontagu:
@samuelcolvin FastAPI does use `inspect.signature`. For what it's worth, though, I think you can inject a custom signature just by setting `Model.__signature__`. At least, if the `__signature__` property exists, I believe `inspect.signature(Model)` will return it.

It would probably good to check whether setting `__signature__` on the model would address the original motivating problem in #1032.

If so, I think it would be a good start to just implement a function that generates the desired signature for the model; after that, we could figure out whether we want fastapi to make use of that function, if we want to perform more automated signature generation entirely within pydantic, or if we want to leave dealing with signatures entirely to fastapi.

-----

It would also be great if there was a way for us to share, at least to some degree, the signature-generation logic between the mypy plugin and runtime code.

samuelcolvin:
> It would also be great if there was a way for us to share, at least to some degree, the signature-generation logic between the mypy plugin and runtime code.

agreed, any idea how we would do that?

---

> For what it's worth, though, I think you can inject a custom signature just by setting `Model.__signature__`

Yes! It should be trivial to set the `__signature__` attribute on a model. I have no idea though whether this would fix #1032.

MrMrRobat:
Actually, it's a great idea to set signature on model. I guess it would work just as signature in `__init__` when inspecting the class itself. Can I implement it in this PR instead of signature on `__init__` or should we open another one?

samuelcolvin:
Same PR is fine with me, maybe change the title.

But are we sure it will solve the original problem?

MrMrRobat:
> But are we sure it will solve the original problem?

Anything that uses `inspect.signature` would work with `__signature__` attr as well.

Also, maybe we should add new signature representation into `__doc__` like it's done in dataclasses?
```py
@dataclass
class Model:
    a: int
    b: float = 1

help(Model)
'''
class Model(builtins.object)
 |  Model(a:int, b:float=1)
...
'''
```

samuelcolvin:
> Anything that uses inspect.signature would work with __signature__ attr as well.

I realise that, I just wondered if fastapi had some hard coded check on `Model.__init__`. I guess not.

> Also, maybe we should add new signature representation into __doc__ like it's done in dataclasses?

Maybe, it might look a little confusing to `schema` which uses that docstring to generate description.

Maybe the auto-generated docstring starts or ends with `(auto-generated)` and schema generation looks for that and skips the docstring as description in that case?

pydantic/main.py
```diff
@@ -147,20 +155,28 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
         fields: Dict[str, ModelField] = {}
         config = BaseConfig
         validators: 'ValidatorListDict' = {}
+        fields_defaults: Dict[str, Any] = {}
+
         pre_root_validators, post_root_validators = [], []
+        init = namespace.get('__init__')
```
samuelcolvin: I think cleaner to skip this and do something like 

```py
cls = super().__new__(mcs, name, bases, new_namespace, **kwargs)
cls.__signature__ = generate_model_signature(cls.__init__, fields, config)  # or just generate_model_signature(cls)
return cls
```

changes/1034-MrMrRobat.md
```diff
@@ -0,0 +1 @@
+Add `__init__` signature generation based on model fields
```
samuelcolvin: Add signature to models.

pydantic/utils.py
```diff
@@ -118,6 +121,50 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
 
+def generate_model_signature(
+    init: Callable[..., None], fields: Dict[str, 'ModelField'], config: Type['BaseConfig']
+) -> Signature:
+    """
+    Generate signature for model based on its fields
+    """
+
+    present_params = signature(init).parameters.values()
+    merged_params: OrderedDict[str, Parameter] = OrderedDict()
```
samuelcolvin:
```suggestion
    merged_params: OrderedDict[str, Parameter] = {}
```

No need for ordered dict.

pydantic/utils.py
```diff
@@ -118,6 +121,50 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+                param_name = field_name
+            else:
+                use_var_kw = True
+                continue
+
+        # TODO: replace annotation with actual expected types once #1055 solved
+        kwargs = {'default': field.default} if not field.required else {}
+        merged_params[param_name] = Parameter(param_name, Parameter.KEYWORD_ONLY, annotation=field.type_, **kwargs)
+
+    extra = config.extra
```
samuelcolvin: no need for this assignment.

pydantic/utils.py
```diff
@@ -118,6 +121,50 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+
+    for param in islice(present_params, 1, len(present_params)):  # skip self arg
+        if param.kind is param.VAR_KEYWORD:
+            var_kw = param
+            continue
+        merged_params[param.name] = param
+
+    allow_names = config.allow_population_by_field_name
+    for field_name, field in fields.items():
+        param_name = field.alias
```
samuelcolvin: no need for this assignment I don't think.

tests/test_model_signature.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_model_signature():
+    class Model(BaseModel):
+        a: float = Field(..., title='A')
+        b = Field(10)
+
+    sig = signature(Model)
+    assert sig != signature(BaseModel)
+
+    assert [str(p).replace(' ', '') for p in sig.parameters.values()] == [
```
samuelcolvin: given that you do this a lot, maybe best to implement your own comparison function:

```py
def _eq(a: Union[str, List[str]], b: Union[str, List[str]]) -> bool:
    """
    compare strings with spaces removed
    """
    if isinstance(a, str) and isinstance(b, str):
        return a.replace(' ', '') == b.replace(' ', '')
    elif isinstance(a, list) and isinstance(b, list):
        return all(_eq(a_, b_) for a_, b_ in zip(a, b))
    else:
        raise TypeError(f'arguments must be both strings or both lists, not {type(a)}, {type(b)}')
```

pydantic/utils.py
MrMrRobat: Why? It can be changed later to `field_name` if it's not correct identifier and field names are allowed, and used more than once.

samuelcolvin: Okay.

samuelcolvin:
@MrMrRobat any chance of finishing this? I'd like to get it released, same with #1017.

MrMrRobat:
Yes, it will be finished, but don't know exactly when. Possibly this month.

samuelcolvin:
Great, thanks. No great hurry.

MrMrRobat:
Seems like all done!

MrMrRobat:
@samuelcolvin, any feedback here?

samuelcolvin:
Sorry, been busy. Will look.

MrMrRobat:
It's totally fine, thanks! Just was afraid you was not aware that PR is ready.

docs/examples/models_signature.py
```diff
@@ -0,0 +1,9 @@
+import inspect
+from pydantic import BaseModel
+
+class FooModel(BaseModel):
+    id: int
+    name: str = None
+    description: str = 'Foo'
```
samuelcolvin:
```suggestion
    description: str = 'Foo'
    apple: int = Field(..., alias='pear')
```
(or something more sensible) good to include an example of an alias.

docs/usage/models.md
```diff
@@ -495,3 +495,27 @@ _(This script is complete, it should run "as is")_
...
+
+Accurate signature is useful for introspection purposes and libraries like `FastAPI` or `hypothesis-auto`.
+
+It also respects your custom defined `__init__` and its arguments:
+
+```py
+{!.tmp_examples/models_signature_custom_init.py!}
+```
+
+!!! note
```
samuelcolvin: I don't think this or the below need to be notes.

pydantic/main.py
```diff
@@ -1,3 +1,4 @@
+import inspect
```
samuelcolvin: Please more this to `if TYPE_CHECKING` and use `from inspect import Signature`.

Inspect is relatively slow to import, though pydantic makes an effort to only import it when used.

see #1127

pydantic/utils.py
```diff
@@ -1,4 +1,6 @@
 import warnings
+from inspect import Parameter, Signature, signature
```
samuelcolvin: please remove this from top level imports.

* `Signature` is only required as a type hint so can go into `if TYPE_CHECKING:`
* `Parameter` and `signature` can be moved into `generate_model_signature`

MrMrRobat: But It used each time when model is created. Why put relatively slow import in a function that gonna be called so many times instead of importing it just once here? That way it's gonna impact on dynamic model creation.

samuelcolvin: no, if you repeat an import which is already loaded, it's cached.

Also model's aren't immediately created every time pydantic is imported.

We could decide that inspect is used so commonly that we just accept it will always be imported by pydantic, but that's another discussion

docs/usage/models.md
```diff
@@ -495,3 +495,27 @@ _(This script is complete, it should run "as is")_
...
+
+## Model signature
+
+Any *pydantic* model have its own signature generated based on its fields:
+
+```py
+{!.tmp_examples/models_signature.py!}
+```
+
+Accurate signature is useful for introspection purposes and libraries like `FastAPI` or `hypothesis-auto`.
```
Zac-HD: The signature introspection actually lives in upstream in `hypothesis` (in the `builds()` and `from_type()` strategies) - `hypothesis-auto` just wraps them.

pydantic/utils.py
```diff
@@ -136,6 +139,49 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+    """
+    Generate signature for model based on its fields
+    """
+
+    present_params = signature(init).parameters.values()
+    merged_params: Dict[str, Parameter] = {}
+    var_kw = None
+    use_var_kw = False
+
+    for param in islice(present_params, 1, len(present_params)):  # skip self arg
```
Zac-HD: You can `islice(present_params, 1, None)` to take to end-of-iterable, which avoids readers having to think about indexing semantics.

MrMrRobat: Good point, thank you! Would you like to suggest the change?

docs/usage/models.md
MrMrRobat: Thank you, will correct. Or maybe, you could suggest the change here and I'll commit it, if you want

pydantic/utils.py
MrMrRobat: Ok, I get it. Will fix soon

pydantic/utils.py
Zac-HD: Nah, go ahead - I appreciate the offer but don't need credit 🙂

docs/usage/models.md
```diff
@@ -495,3 +495,25 @@ _(This script is complete, it should run "as is")_
...
+It also respects your custom defined `__init__` and its arguments:
+
+```py
+{!.tmp_examples/models_signature_custom_init.py!}
+```
+
+To be included in signature, field alias or name must be a valid python identifier. *pydantic* prefer aliases over names, but may use field name if alias is not suitable. 
+If any field's alias and name are both invalid identifiers, `**data` argument will be added.
+
+Also, `**data` argument is always present in signature if `Config.extra` is `Extra.allow`
```
MrMrRobat:
```suggestion
Also, `**data` argument is always present in signature if `Config.extra` is `Extra.allow`.

!!! note
    Types in signature are same as declared in model annotations. They are not necessarily the ones that can be parsed. That may change once #1055 solved.
```

MrMrRobat: @samuelcolvin, after this one gets resolved, I guess, we're ready to go 🎉

samuelcolvin: I've rewritten the docs to improve the english, and incorporated this suggestion.

samuelcolvin:
For me this is ready to merge if you agree @MrMrRobat?

MrMrRobat:
@samuelcolvin, is there any reason for changes in af3dd4d? Added test pass even without explicit declaration of `**data`

Moreover, it may lead to confusion in such cases as:
```py
    class Model(BaseModel):
        a: float
        b: int = 2

        def __init__(self, a: float, b: int):
            super().__init__(a=a, b=b)

        class Config:
            extra = 'allow'
```
Even though `Config.extra` is `'allow'`, there is no space for `**data` in custom declared init, so I think signature should be real `(a: float, b: int) -> None` rather than `(a: float, b: int = 2, **data: Any) -> None`

pydantic/utils.py
```diff
@@ -149,7 +149,7 @@ def generate_model_signature(
 
     present_params = signature(init).parameters.values()
     merged_params: Dict[str, Parameter] = {}
-    var_kw = None
+    var_kw = Parameter('data', Parameter.VAR_KEYWORD, annotation=Any)
```
MrMrRobat: My thoughts on this: https://github.com/samuelcolvin/pydantic/pull/1034#issuecomment-592161978

tests/test_model_signature.py
```diff
@@ -57,3 +57,14 @@ def test_invalid_identifiers_signature():
     assert _equals(str(signature(model)), '(*, valid_identifier: int = 123, yeah: int = 0) -> None')
     model = create_model('Model', **{'123 invalid identifier!': 123, '!': Field(0, alias='yeah')})
     assert _equals(str(signature(model)), '(*, yeah: int = 0, **data: Any) -> None')
+
+
+def test_kwargs():
```
MrMrRobat: As I said in https://github.com/samuelcolvin/pydantic/pull/1034#issuecomment-592161978, this test would already work without changes in this commit

samuelcolvin:
Yes sorry, maybe you're right. Feel free to revert this change or I will.

---

Sorry to mess with your PR, just trying to get it merged as I know I'm slow to respond here

MrMrRobat:
That's totally fine, I don't mind any edits from your side.
I'll revert just in a few minutes and will add test for case above

Zac-HD:
Why not make it an error to specify both a custom init method and config for extras-allowed? 

As far as I can see it's always redundant (just check for a VAR_KW param), and may be inconsistent (usually indicating a bug).

pydantic/utils.py
```diff
@@ -136,6 +139,51 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+                    param_name = field_name
+                else:
+                    use_var_kw = True
+                    continue
+
+            # TODO: replace annotation with actual expected types once #1055 solved
+            kwargs = {'default': field.default} if not field.required else {}
+            merged_params[param_name] = Parameter(param_name, Parameter.KEYWORD_ONLY, annotation=field.type_, **kwargs)
+
+    if config.extra is config.extra.allow:
```
MrMrRobat: As proposed in https://github.com/samuelcolvin/pydantic/pull/1034#issuecomment-592181568
```suggestion
    if config.extra is config.extra.allow:
        if not var_kw:
            raise TypeError("Using `Config.extra == Extra.allow` without **data parameter in __init__ is ambiguous")

```
@samuelcolvin, what your thoughts on this?

samuelcolvin: I disagree with this, people should be allowed to create an init method that doesn't allow extra:
* we don't do any other checks that `__init__` matches model
* I can imagine a scenario where a super method has `extra = allow` but a submodel wants a custom init method without having to bother with modifying the config

pydantic/utils.py
```diff
@@ -136,6 +139,51 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+        allow_names = config.allow_population_by_field_name
+        for field_name, field in fields.items():
+            param_name = field.alias
+            if field_name in merged_params or param_name in merged_params:
+                continue
+            elif not param_name.isidentifier():
+                if allow_names and field_name.isidentifier():
+                    param_name = field_name
+                else:
+                    use_var_kw = True
```
samuelcolvin: coverage is being weird, but it [looks](https://codecov.io/gh/samuelcolvin/pydantic/compare/d8d8fc1efd818e5f1114caa95bc17168d874eef1...d8d8fc1efd818e5f1114caa95bc17168d874eef1/src/pydantic/utils.py) like this line might not be coverd

MrMrRobat: Seems like it's covered in `test_invalid_identifiers_signature`, maybe coverage struggles to detect it.

pydantic/utils.py
```diff
@@ -136,6 +139,51 @@ def almost_equal_floats(value_1: float, value_2: float, *, delta: float = 1e-8)
...
+    if var_kw:  # if custom init has no var_kw, fields which are not declared in it cannot be passed through
+        allow_names = config.allow_population_by_field_name
+        for field_name, field in fields.items():
+            param_name = field.alias
+            if field_name in merged_params or param_name in merged_params:
+                continue
+            elif not param_name.isidentifier():
+                if allow_names and field_name.isidentifier():
+                    param_name = field_name
+                else:
```
MrMrRobat:
```suggestion
                else:  # pragma: no cover
 
```

pydantic/utils.py
samuelcolvin: if it's covered, don't worry.

PR #1054, opened 2019-11-30
Allow schema_extra to be a callable

### LINKED ISSUE #889: Is it possible to have a schema with no $ref references ?
Versions:
* OS: Ubuntu 18.04
* Python: 3.7.4
* Pydantic: 0.32.2

Hi,

First of all a huge thanks for the great work done on this package, glad that you are reaching version 1.0 soon !

I have the following simple models: a car and an engine, a car holds an engine.

```py

class Engine(BaseModel):
    size: int

class Car(BaseModel):
    engine: Engine
```

If I print the car's schema I get this:

```py
print(Car.schema_json(indent=4, sort_keys=True))

# >>> {
#     "definitions": {
#         "Engine": {
#             "properties": {
#                 "size": {
#                     "title": "Size",
#                     "type": "integer"
#                 }
#             },
#             "required": [
#                 "size"
#             ],
#             "title": "Engine",
#             "type": "object"
#         }
#     },
#     "properties": {
#         "engine": {
#             "$ref": "#/definitions/Engine"
#         }
#     },
#     "required": [
#         "engine"
#     ],
#     "title": "Car",
#     "type": "object"
# }
```

I would prefer very much to have this, a recursive engine schema inside the car's one:

```py
# >>> {
#     "definitions": {},
#     "properties": {
#         "engine": {
#             "properties": {
#                 "size": {
#                     "title": "Size",
#                     "type": "integer"
#                 }
#             },
#             "required": [
#                 "size"
#             ],
#             "title": "Engine",
#             "type": "object"
#         }
#     },
#     "required": [
#         "engine"
#     ],
#     "title": "Car",
#     "type": "object"
# }
```

The pydantic documentation states that:
> “sub-models” with modifications (via the Schema class) like a custom title, description or default value, are recursively included instead of referenced.

So I tried and modified my code by using the Schema class to trigger something but still did not obtain what I wanted:

```py
from pydantic import Schema

class Engine(BaseModel):
    size: int = Schema(..., title="a size")

class Car(BaseModel):
    engine: Engine = Schema(..., title="an engine")

print(Car.schema_json(indent=4, sort_keys=True))

# >>> {
#     "definitions": {
#         "Engine": {
#             "properties": {
#                 "size": {
#                     "title": "a size",
#                     "type": "integer"
#                 }
#             },
#             "required": [
#                 "size"
#             ],
#             "title": "Engine",
#             "type": "object"
#         }
#     },
#     "properties": {
#         "engine": {
#             "allOf": [
#                 {
#                     "$ref": "#/definitions/Engine"
#                 }
#             ],
#             "title": "an engine"
#         }
#     },
#     "required": [
#         "engine"
#     ],
#     "title": "Car",
#     "type": "object"
# }
```

Instead I got a "allOf" in the car's schema which I do not understand and still no recursive engine schema.

My question is: is it possible to achieve what I want ? No definitions sections, just recursive schemas.

I know I could have some problems with cross referencing schemas (A referencing B and B referencing A), but I do not care of this detail since I know my schemas do not have this kind of thing.

Did I miss or misunderstood something in the documentation (which I read several times) ?
I even tried to inspect pydantic's code to find an answer but the section (schema.py) which generates the schemas is a bit too complicated for me.

Thank you for your time and go on with the greak work :)

### LINKED ISSUE #1051: Supress title generation
# Feature Request

* OS: **Windows 10**
* Python version `import sys; print(sys.version)`: **3.7.1 (v3.7.1:260ec2c36a, Oct 20 2018, 14:57:15) [MSC v.1915 64 bit (AMD64)]**

There's no way to suppress/opt-out of the `title` property generation in `schema()`.
```py
# main.py
import pprint

import pydantic

class User(pydantic.BaseModel):
    name: str = pydantic.Field(..., title="", description="User name")
    password: str = pydantic.Field(..., description="User password")
    age: int = pydantic.Field(..., title=None, description="User age")


pprint.pprint(User.schema())
```
spits out:
```py
{'properties': {'age': {'description': 'User age',
                        'title': 'Age',
                        'type': 'integer'},
                'name': {'description': 'User name',
                         'title': 'Name',
                         'type': 'string'},
                'password': {'description': 'User password',
                             'title': 'Password',
                             'type': 'string'}},
 'required': ['name', 'password', 'age'],
 'title': 'User',
 'type': 'object'}
```
notice how all `properties` are given the auto-generated `title`.

I care about this mostly as a [fastapi](https://github.com/tiangolo/fastapi) user. I don't like how Swagger UI renders the title, and find the title to be redundant with `description` anyway (although maybe I'm missing something?)

This can be seen by adding the following to the code above:
```py
import fastapi

app = fastapi.FastAPI()

@app.post("/users")
def create_user(user: User):
    pass
```
and running the server with `uvicorn main:app` (note this will require `pip install fastapi uvicorn`).
`http://127.0.0.1:8000/docs` show:
image

If I patch `pydantic/schema.py` with the following:
```py
    schema_overrides = False
    # s = dict(title=field.field_info.title or field.alias.title().replace('_', ' '))
    s = {}
    if field.field_info.title:
```
the output is much nicer (IMO):

image

I'd be happy to submit a PR. Adding a `Config` option probably makes sense, but maybe there should also be a way to disable the auto-title only for a specific field?

selimb:
## Change Summary

Allow `Config.schema_extra` to be a callable so that the generated schema can be post-processed.
Also includes some fixes to the dev scripts for Windows development (I'm on Windows with cmder). *Those can be moved to another PR if needed.*

## Related issue number

fix #889, fix #1051 

## Checklist

* [x] `changes/1054-selimb.md` file added describing change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ black = black -S -l 120 --target-version py36 pydantic tests
 
 .PHONY: install
 install:
-	pip install -U setuptools pip
+	python -m pip install -U setuptools pip
 	pip install -U -r requirements.txt
 	SKIP_CYTHON=1 pip install -e .
 
@@ -91,7 +91,7 @@ clean:
 
 .PHONY: docs
 docs:
-	./docs/build/main.py
+	python docs/build/main.py
 	mkdocs build
 	@# to work with the old sphinx build and deploy:
 	@rm -rf docs/_build/
@@ -100,7 +100,7 @@ docs:
 
 .PHONY: docs-serve
 docs-serve:
-	./docs/build/main.py
+	python docs/build/main.py
 	mkdocs serve
 
 .PHONY: publish
--- a/docs/examples/schema_extra_callable.py
+++ b/docs/examples/schema_extra_callable.py
@@ -1,0 +1,15 @@
+# output-json
+from typing import Dict, Any
+from pydantic import BaseModel
+
+class Person(BaseModel):
+    name: str
+    age: int
+
+    class Config:
+        @staticmethod
+        def schema_extra(schema: Dict[str, Any]) -> None:
+            for prop in schema.get('properties', {}).values():
+                prop.pop('title', None)
+
+print(Person.schema_json(indent=2))
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -146,3 +146,20 @@ Outputs:
 ```json
 {!.tmp_examples/schema_with_example.json!}
 ```
+
+For more fine-grained control, you can alternatively set `schema_extra` to a callable and post-process the generated schema.
+The callable is passed the schema dictionary as the first and only argument, and is expected to mutate it *in-place*; the return value is not used.
+
+For example, the `title` key can be removed from the model's `properties`:
+
+```py
+{!.tmp_examples/schema_extra_callable.py!}
+```
+
+_(This script is complete, it should run "as is")_
+
+Outputs:
+
+```json
+{!.tmp_examples/schema_extra_callable.json!}
+```
--- a/changes/1054-selimb.md
+++ b/changes/1054-selimb.md
@@ -1,0 +1,1 @@
+Allow `Config.schema_extra` to be a callable so that the generated schema can be post-processed.
--- a/docs/build/exec_examples.py
+++ b/docs/build/exec_examples.py
@@ -91,7 +91,6 @@ def exec_examples():
     errors = []
     all_md = all_md_contents()
     new_files = {}
-    os.environ.clear()
     os.environ.update({'my_auth_key': 'xxx', 'my_api_key': 'xxx'})
 
     sys.path.append(str(EXAMPLES_DIR))
@@ -113,7 +112,7 @@ def exec_examples():
         if f'{{!.tmp_examples/{file.name}!}}' not in all_md:
             error('file not used anywhere')
 
-        file_text = file.read_text()
+        file_text = file.read_text('utf-8')
         if '\n\n\n' in file_text:
             error('too many new lines')
         if not file_text.endswith('\n'):
@@ -184,7 +183,7 @@ def exec_examples():
     print(f'writing {len(new_files)} example files to {TMP_EXAMPLES_DIR}')
     TMP_EXAMPLES_DIR.mkdir()
     for file_name, content in new_files.items():
-        (TMP_EXAMPLES_DIR / file_name).write_text(content)
+        (TMP_EXAMPLES_DIR / file_name).write_text(content, 'utf-8')
     gen_ansi_output()
     return 0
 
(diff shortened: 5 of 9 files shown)


selimb:
Was going to include functions for `schema_extra` to address both #889 and #1051, but it proved less trivial than I thought to handle all cases.

docs/examples/schema_extra_callable.py
```diff
@@ -0,0 +1,16 @@
...
+def remove_titles(schema: Dict[str, Any]) -> None:
+    for prop in schema.get('properties', {}).values():
+        prop.pop('title', None)
+
+class Person(BaseModel):
+    name: str
+    age: int
+
+    class Config:
+        schema_extra = remove_titles
```
samuelcolvin: I would guess clearer to have something like
```suggestion
        @staticmethod
        def schema_extra(schema: Dict[str, Any]) -> None:
            for prop in schema.get('properties', {}).values():
                prop.pop('title', None)
```

docs/usage/schema.md
```diff
@@ -146,3 +146,17 @@ Outputs:
 ```json
 {!.tmp_examples/schema_with_example.json!}
 ```
+
+For more fine-grained control, you can alternatively set `schema_extra` to a callable and post-process the generated schema.
```
samuelcolvin: Maybe worth a note saying explicitly `schema_extra` allows the schema argument to be mutated, rather than returning a new altered dictionary. For notices to miss this potential stumbling block

docs/examples/schema_extra_callable.py
```diff
@@ -0,0 +1,16 @@
+# output-json
```
samuelcolvin: :+1:

Makefile
```diff
@@ -91,7 +91,7 @@ clean:
 
 .PHONY: docs
 docs:
-	./docs/build/main.py
+	python docs/build/main.py
```
samuelcolvin: LGTM, you might want to change `docs-serve` below too.

samuelcolvin:
> Was going to include functions for `schema_extra` to address both #889 and #1051, but it proved less trivial than I thought to handle all cases.
> 
> Could take a stab at it in another PR if you think it'd be useful?

Humm, thanks but I think it get's complicated if we start providing functions for these cases within pydantic, there are hundreds of different changes that people might want. I think best to leave things like this up to package users.

docs/examples/schema_extra_callable.py
samuelcolvin: might be good to change the test too to confirm this works.

selimb:
Thanks! Believe the comments have been addressed.

> Humm, thanks but I think it get's complicated if we start providing functions for these cases within pydantic, there are hundreds of different changes that people might want. I think best to leave things like this up to package users.

Agreed. Sounds good.

samuelcolvin:
great, thank you.

selimb:
Thanks to you! Cheers.

PR #1061, opened 2019-12-02
Trick Cython into allowing str subclassing

### LINKED ISSUE #1060: Subclassing StrictStr / ConstrainedStr causes ValidationError
# Bug

Please complete:
* OS: **Manjaro Linux**
* Python version: **3.7.4**
* Pydantic version: **1.2.0**

------------------------------------------------------------------
### Original bug discussion in #799 
------------------------------------------------------------------

I am sorta lost as to why this isn't working as expected for me in v1.2:

```py
from typing import Type, TypeVar
import pydantic


class ConnectionName(pydantic.StrictStr):
    min_length = 1


class EndpointStr(pydantic.StrictStr):
    min_length = 1


DEFAULT_ENDPOINT = EndpointStr("default")
T = TypeVar("T", bound="UpRequestEvent")


class UpRequestEvent(pydantic.BaseModel):
    connection_name: ConnectionName  # connection name
    connection_endpoint: EndpointStr  # connection name's endpoint (for server type connections)

    @classmethod
    def build_from(
        cls: Type[T],
        connection_name: ConnectionName,
        connection_endpoint: EndpointStr = DEFAULT_ENDPOINT,
        **kwargs
    ) -> T:
        return cls(
            connection_name=connection_name,
            connection_endpoint=connection_endpoint,
        )

    class Config:
        use_enum_values = True


# This works
UpRequestEvent.build_from(
    connection_name="test", connection_endpoint="default",
)

# This fails
UpRequestEvent.build_from(
    connection_name=ConnectionName("test"), connection_endpoint=EndpointStr("default"),
)
```

```
  File "/opt/pycharm-eap/plugins/python/helpers/pydev/pydevd.py", line 1415, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/opt/pycharm-eap/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/spalmer1009/.PyCharm2019.3/config/scratches/scratch.py", line 44, in <module>
    connection_name=ConnectionName("test"), connection_endpoint=EndpointStr("default"),
  File "/home/spalmer1009/.PyCharm2019.3/config/scratches/scratch.py", line 30, in build_from
    connection_endpoint=connection_endpoint,
  File "pydantic/main.py", line 274, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for UpRequestEvent
connection_name
  Expected unicode, got ConnectionName (type=type_error)
connection_endpoint
  Expected unicode, got EndpointStr (type=type_error)
```

My special types ConnectionName and EndpointStr are a subclass of `str` (through `StrictStr` -> `ConstrainedStr`)

---------------------------------------

As discovered by @dmontagu the issue only seems to exist when the code is Cython compiled.

skewty:
## Change Summary

Tricks Cython into allowing str subclassing

## Related issue number

Fixes #1060

## Checklist

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -977,18 +977,6 @@ def test_strict_bool():
         Model(v=b'1')
 
 
-# def test_strict_bool_subclass():
-#     class MyStrictBool(StrictBool):
-#         pass
-#
-#     class Model(BaseModel):
-#         v: MyStrictBool
-#
-#     m = Model(v=MyStrictBool(False))
-#     assert isinstance(m.v, MyStrictBool)
-#     assert m.v is False
-
-
 def test_strict_int():
     class Model(BaseModel):
         v: StrictInt
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -948,6 +948,24 @@ def test_strict_str():
         Model(v=b'foobar')
 
 
+def test_strict_str_subclass():
+    class MyStrictStr(StrictStr):
+        pass
+
+    class Model(BaseModel):
+        v: MyStrictStr
+
+    m = Model(v=MyStrictStr('foobar'))
+    assert isinstance(m.v, MyStrictStr)
+    assert m.v == 'foobar'
+
+    with pytest.raises(ValidationError):
+        Model(v=123)
+
+    with pytest.raises(ValidationError):
+        Model(v=b'foobar')
+
+
 def test_strict_bool():
     class Model(BaseModel):
         v: StrictBool
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -959,12 +959,6 @@ def test_strict_str_subclass():
     assert isinstance(m.v, MyStrictStr)
     assert m.v == 'foobar'
 
-    with pytest.raises(ValidationError):
-        Model(v=123)
-
-    with pytest.raises(ValidationError):
-        Model(v=b'foobar')
-
 
 def test_strict_bool():
     class Model(BaseModel):
@@ -983,6 +977,18 @@ def test_strict_bool():
         Model(v=b'1')
 
 
+# def test_strict_bool_subclass():
+#     class MyStrictBool(StrictBool):
+#         pass
+#
+#     class Model(BaseModel):
+#         v: MyStrictBool
+#
+#     m = Model(v=MyStrictBool(False))
+#     assert isinstance(m.v, MyStrictBool)
+#     assert m.v is False
+
+
 def test_strict_int():
     class Model(BaseModel):
         v: StrictInt
@@ -996,6 +1002,18 @@ def test_strict_int():
         Model(v=3.14159)
 
 
+def test_strict_int_subclass():
+    class MyStrictInt(StrictInt):
+        pass
+
+    class Model(BaseModel):
+        v: MyStrictInt
+
+    m = Model(v=MyStrictInt(123456))
+    assert isinstance(m.v, MyStrictInt)
+    assert m.v == 123456
+
+
 def test_strict_float():
     class Model(BaseModel):
         v: StrictFloat
@@ -1007,6 +1025,18 @@ def test_strict_float():
 
     with pytest.raises(ValidationError, match='value is not a valid float'):
         Model(v=123456)
+
+
+def test_strict_float_subclass():
+    class MyStrictFloat(StrictFloat):
+        pass
+
+    class Model(BaseModel):
+        v: MyStrictFloat
+
+    m = Model(v=MyStrictFloat(3.14159))
+    assert isinstance(m.v, MyStrictFloat)
+    assert m.v == 3.14159
 
 
 def test_bool_unhashable_fails():
--- a/changes/1061-skewty.md
+++ b/changes/1061-skewty.md
@@ -1,0 +1,4 @@
+Trick Cython into allowing str subclassing.
+
+Without this "trick" passing in values that subclassed `ConstrainedStr` always raised a `Field` `ValidationError`.  
+This `ValidationError` issue only seems to manifest in the **Cython** compiled variant (not in **CPython**).
--- a/changes/1061-skewty.md
+++ b/changes/1061-skewty.md
@@ -1,4 +1,1 @@
 Trick Cython into allowing str subclassing.
-
-Without this "trick" passing in values that subclassed `ConstrainedStr` always raised a `Field` `ValidationError`.  
-This `ValidationError` issue only seems to manifest in the **Cython** compiled variant (not in **CPython**).
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -157,7 +157,7 @@ class ConstrainedStr(str):
         yield cls.validate
 
     @classmethod
-    def validate(cls, value: str) -> str:
+    def validate(cls, value: Union[str]) -> Union[str]:
         if cls.curtail_length and len(value) > cls.curtail_length:
             value = value[: cls.curtail_length]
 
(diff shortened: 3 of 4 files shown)


dmontagu:
@skewty would you mind augmenting the test you've added to check for subclasses of each of the strict types?

Based on the response in https://github.com/cython/cython/issues/3256 it sounds like there is a chance a similar issue could exist for the other strict type validators (or really anything in pydantic annotated as returning a builtin when it might actually return a subclass).

I don't think we need to go crazy, and there might not actually be anything to change here, but I think it's worth checking for issues with `StrictInt`, `StrictBool`, etc. while we are fixing this.

tests/test_types.py
```diff
@@ -948,6 +948,24 @@ class Model(BaseModel):
         Model(v=b'foobar')
 
 
+def test_strict_str_subclass():
+    class MyStrictStr(StrictStr):
+        pass
+
+    class Model(BaseModel):
+        v: MyStrictStr
```
dmontagu: In particular, could you add `class MyStrictInt(StrictInt): pass` (and similar for `StrictFloat` and `StrictBool`), add them as fields on `Model`, and pass appropriate values in the `m = Model(...` line?

No need to add any more `with pytest.raises(ValidationError)` checks, since those shouldn't be affected by this issue.

samuelcolvin:
Humm, I've checked and this seems to have a small but measurable detrimental effect on performance, perhaps ~5%.

Is it something that really useful enough to make the performance decrease a price worth paying?

samuelcolvin:
humm, maybe I'm wrong about the performance impact. I can't get it to be repeatable.

skewty:
Subclassing `StrictBool` doesn't quite work so nicely because `class StrictBool(int)`  The value a subclass gets for `False` is therefore just `0` instead of `False`.  You cannot set `bool` as a base type.  You get "TypeError: type 'bool' is not an acceptable base type".  And, obviously, returning True or False would return the "wrong type".  Should a warning be generated if somebody tries to subclass Constrained/StrictBool or would you prefer an attempt at making it subclassable?

dmontagu:
I’m fine with not addressing StrictBool in this PR.

I think it probably makes sense to add a validator that checks for a subclass of StrictBool and returns the actual bool, or something along those lines, but I think that’s less likely to matter in practice and can be addressed later.

If you can get this to work for int str and float following this pattern let’s just go with that (assuming no substantial performance degradations).

samuelcolvin:
@dmontagu would it be possible for you to try this too and see if you can identify a performance impact?

From you comment https://github.com/samuelcolvin/pydantic/pull/799#issuecomment-560168708 I can imagine it makes sense for there to be no performance impact, but I'm still hesitant.

dmontagu:
@samuelcolvin I did a lot of runs using the following script to compare master vs a rebased version of skewty/master:

```python
import numpy as np
import time


from pydantic import StrictStr, BaseModel, ValidationError


class Model(BaseModel):
    a: StrictStr


def benchmark_str():
    n_rounds = 15
    n_iterations_per_round = 100000
    success_times = []
    fail_times = []
    for _ in range(n_rounds):
        t0 = time.time()
        for _ in range(n_iterations_per_round):
            Model(a="hello")

        t1 = time.time()
        for _ in range(n_iterations_per_round):
            try:
                Model(a=123)
            except ValidationError:
                pass
        t2 = time.time()
        success_times.append(t1 - t0)
        fail_times.append(t2 - t1)
        print(f"StrictStr: {t1 - t0:.3f}s success, {t2 - t1:.3f}s fail")
    print(np.mean(success_times[5:]))
    print(np.mean(fail_times[5:]))

benchmark_str()
```

There is a lot of fluctuation from run to run, but after doing many runs back and forth, it seems to me that there is not a meaningful difference between the "success" times, but the "fail" times average about 1.5% higher.

Given the rarity of use of this type, and the relatively small increase in validation times, I'm personally okay with it.

------

After thinking about this more, if we wanted to keep all of the performance benefits but still allow arbitrary strict types, I think there is another fundamentally different route we could go: replacing the `StrictX` types with `NewType`-generated types, and adding a flag to the `NewType` that pydantic could detect during model creation. This could look something like:

```python
StrictStr = NewType("StrictStr", str)
StrictStr.__pydantic_strict__ = True
```
This would mean that at runtime there is no difference between a str and a `StrictStr`, but we could still make use of the `__pydantic_strict__` flag to modify the validator-generation logic during initialization.

You wouldn't be able to subclass, but if you are only doing it for typing-related reasons, you can create nested NewTypes, so you could do:
```python
MyStrictStr = NewType("MyStrictStr", StrictStr)
```
and it would work like a subclass for type-checking purposes.

Given it would break subclasses I guess this would need to be a v2 thing, but I think it's worth thinking about. It would make it much easier to make arbitrary `StrictX` types -- we could just do `isinstance` checks for the `NewType`'s parent type.

skewty:
As being discussed in #1065, it may look more like this:

```py
StrictStr = NewType("StrictStr", str)
StrictStr.__schema_attributes__ = SchemaAttributes(x, y, z)  # could just be a dict
```

I would like the define more complex validation inside the class declaration if possible.
Having the validation always included for free when I use the constrained / strict variant of a variable type is the big win for me.

`typing.NewType` doesn't play nice with the `class` keyword.

```py
from typing import NewType
Str = NewType("Str", str)
class MyStr(Str):
...    __foo__ = "bar"
...
  File "<input>", line 1, in <module>
TypeError: function() argument 1 must be code, not str
```

But this works:

```py
MyStr = NewType("MyStr", NewType("Str", str))
```

so it follows that:

```py
from typing import NewType
from pydantic import BaseModel, SchemaAttributes

StrictStr = NewType("StrictStr", str)
StrictStr.__schema_attributes__ = SchemaAttributes(min_length=15)

class MyStrictStr(StrictStr):

    @classmethod
    def __get_validators__(cls):
        yield cls.validate_prefix

    @classmethod
    def validate_prefix(cls, value):
        if str(value).startswith("foo:"):
            return value
        raise ValueError('Nope!')
```
would also fail with `TypeError: function() argument 1 must be code, not str`.

What are you proposing for custom validation?  Re-usable validator could be manually added to the class where the StrictX is being included.

dmontagu:
I was imagining that `StrictX` didn’t get any custom validation beyond ensuring it was the right type. If you wanted custom validation, you could use your own custom type not derived from `StrictX`, and manually add the `isinstance` check. If that’s less useful or otherwise not what we want, I’m also happy to drop the NewType idea entirely.

skewty:
That would work for me.  I would like to make use of `SchemaAttributes` though.  I like the idea of some free JSON schema and validation.

So maybe something like this?

```py
from pydantic import BaseModel, SchemaAttributes

class FooStr(str):
    __schema_attributes__ = SchemaAttributes(min_length=15)

    @classmethod
    def __get_validators__(cls):
        yield cls.validate_prefix

    @classmethod
    def validate_prefix(cls, value):
        if str(value).startswith("foo:"):
            return value
        raise ValueError('Nope!')
```

samuelcolvin:
This discussion went slightly off-topic. 

Is this ready to merge? Are we sure it doesn't effect performance?

dmontagu:
@samuelcolvin

> There is a lot of fluctuation from run to run, but after doing many runs back and forth, it seems to me that there is not a meaningful difference between the "success" times, but the "fail" times average about 1.5% higher.

By fail I mean when validation fails. This is specific to the StrictStr.

samuelcolvin:
Sorry @dmontagu, I got confused.

@skewty thank you very much.

PR #1065, opened 2019-12-02
Prevent type attributes being added to schema unintentionally

### LINKED ISSUE #667: What is the proper way to create a numpy ndarray field
# Question

For bugs/questions:
* OS: **Mac**
* Python version `import sys; print(sys.version)`: **Python 3.6.8 :: Anaconda, Inc.**
* Pydantic version `import pydantic; print(pydantic.VERSION)`: **0.30**

I would like to make a custom data type like done here: https://pydantic-docs.helpmanual.io/#custom-data-types but use numpy.ndarray. I have made validators so that the object can construct but I can't seem to figure out how to properly define the schema for this. Is this something I should be able to do and how?

```py
import pydantic
import numpy

class NumpyNDArray(numpy.ndarray):
    @classmethod
    def __get_validators__(cls):
        yield cls.validate
    @classmethod
    def validate(cls, v: Any) -> str:
        # validate data...
        return v


class RemoteExample(pydantic.BaseModel):
    numpy_sample: NumpyNDArray = pydantic.Schema(None, alias='numpy_sample')


a = numpy.ndarray((1,2,3))
re = RemoteExample(numpy_sample=a)
print(re.schema())
```

```
  File "test_resources.py", line 8, in <module>
    print(re.schema())
  File ".../python3.6/site-packages/pydantic/main.py", line 463, in schema
    s = model_schema(cls, by_alias=by_alias)
  File ".../python3.6/site-packages/pydantic/schema.py", line 235, in model_schema
    model, by_alias=by_alias, model_name_map=model_name_map, ref_prefix=ref_prefix
  File ".../python3.6/site-packages/pydantic/schema.py", line 553, in model_process_schema
    model, by_alias=by_alias, model_name_map=model_name_map, ref_prefix=ref_prefix, known_models=known_models
  File ".../python3.6/site-packages/pydantic/schema.py", line 581, in model_type_schema
    f, by_alias=by_alias, model_name_map=model_name_map, ref_prefix=ref_prefix, known_models=known_models
  File ".../python3.6/site-packages/pydantic/schema.py", line 294, in field_schema
    known_models=known_models or set(),
  File ".../python3.6/site-packages/pydantic/schema.py", line 524, in field_type_schema
    known_models=known_models,
  File ".../python3.6/site-packages/pydantic/schema.py", line 794, in field_singleton_schema
    raise ValueError(f'Value not declarable with JSON Schema, field: {field}')
ValueError: Value not declarable with JSON Schema, field: numpy_sample type=NumpyNDArray default=None
```

I tried looking through `field_singleton_schema` and couldn't find anything obvious to do.

### LINKED ISSUE #789: More complex, practical examples of custom types in the docs
(This issue is a follow-up to the point I made and ensuing discussion in https://github.com/samuelcolvin/pydantic/issues/788; I wanted to open a new issue rather than derailing that one.)

@samuelcolvin
The reason I would like a more sophisticated example like that from #788 in the docs is that, until I had actually worked through writing pull requests for a handful of incrementally more complex types (e.g., Strict/RelaxedBool, Base64Encoded, etc.), I didn't feel like I had a good sense of what it would look like to build a complex, practical type with pydantic. 

Even now, I am rarely inclined to create a custom type in my own code/projects, and would have to think somewhat deeply / reference the docs to implement a type like the one in #788, despite the clean, simple API for doing so.

I currently attribute this to the lack of powerful, practical, domain-specific examples in the documentation.

Most of the examples in the docs perform more general validation (things like string length, etc.). Recalling my perspective as a new user (which wasn't actually that long ago 😅),  the examples in the docs 1) were a little too abstract for my imagination to grasp all of the possibilities, and 2) their simplicity led me to implicitly believe that "proper" usage of the library involved a small number of more lightweight, simple validations. Clearly its capabilities go well beyond that; it just wasn't my first impression.

I think an example like #788 would really help inspire intermediate pydantic users to build and use custom types where appropriate in their projects.

-----

On the other hand, I do understand the desire to show clearly unrealistic types in the "Custom Data Types" section -- better for people to use a shared, maintained implementation, rather than copy/paste. However, I think there may be something to be said for doing more to guide users toward self-sufficiency (going beyond a description of the available APIs).

Sqlalchemy is an example of a project that I think follows this strategy: for example, their [custom type docs](https://docs.sqlalchemy.org/en/13/core/custom_types.html) include a "Backend-agnostic GUID Type". That type *could* be integrated directly into the library, would probably get more usage if it *were* integrated into the library, and there are several stackoverflow answers that essentially recommend copy/pasting it into your project. But I think by including it *only as an example in the docs*, it actually helps push users toward understanding the internals, and becoming self-sufficient in creating their own custom types.

I think a similar "advanced examples" section like that in the pydantic docs would result in fewer pull/feature-requests for custom types, and more users just using the example as a starting point for implementing their own types.

-----

I would be very interested to hear if anyone else has had an experience similar to mine. If I am an outlier, the docs shouldn't be changed just for me!

### LINKED ISSUE #1064: Enum field raises AttributeError or produces invalid schema with specific names
# Bug
*When calling `BaseModel.schema()` or `BaseModel.schema_json()`*

An `Enum` field with the name `regex` raises an AttributeError.

An `Enum` field will produce an invalid schema if any of the following are used as an Enum name: `gt`; `lt`; `ge`; `le`; `max_length`; or `multiple_of`. 

It looks like the Enum name is being interpreted as a `ModelField` property

Please complete:
* OS: **Ubuntu 18.04.3 LTS**
* Python version `import sys; print(sys.version)`: **3.8.0 (default, Oct 28 2019, 16:14:01)**

**Please read the [docs](https://pydantic-docs.helpmanual.io/) and search through issues to
confirm your bug hasn't already been reported.**

Where possible please include a self contained code snippet describing your bug:

```py
import enum
import pydantic

class ExampleEnum(enum.Enum):
    just_a_value = "Just a normal value"

    # These names are interpreted as properties of the ModelField
    gt = "GT"
    lt = "LT"
    ge = "GE"
    le = "LE"
    max_length = "ML"
    multiple_of = "MO"

    # This name will cause an AttributeError (when uncommented)
    # regex = "RE"

class Example(pydantic.BaseModel):
    example: ExampleEnum

Example.schema()
...
```

The example above will produce this schema dict; where properties have been derived from the Enum names (rather than a `Field` model).

```py
{'title': 'Example',
 'type': 'object',
 'properties': {'example': {'title': 'Example',
   'enum': ['Just a normal value', 'GT', 'LT', 'GE', 'LE', 'ML', 'MO'],
   'maxLength': <ExampleEnum.max_length: 'ML'>,
   'exclusiveMinimum': <ExampleEnum.gt: 'GT'>,
   'exclusiveMaximum': <ExampleEnum.lt: 'LT'>,
   'minimum': <ExampleEnum.ge: 'GE'>,
   'maximum': <ExampleEnum.le: 'LE'>,
   'multipleOf': <ExampleEnum.multiple_of: 'MO'>}},
 'required': ['example']}
```

samuelcolvin:
## Change Summary

Constrained types which add properties to schema now ~~need to have `__schema_attributes__ = True`~~ implement `__modify_schema__` to update the schema, this prevents unexpected conflicts like on enums attributes named liked `.regex`.

Technically this could not be backwards compatible if people have created their own constrained types. Is anyone concerned about this? Otherwise I think the chance of a problem is remote enough to allow this in a minor release.

## Related issue number

* fix #1064
* fix #789
* fix #667

## Checklist

* [x] documentation updated

### DIFF
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -73,7 +73,7 @@ to set all of the arguments above except `default`.
 
 ### Unenforced Field constraints
 
-If *pydantic* finds constraints which are not being enforced, an error will be raised. If you want to force the 
+If *pydantic* finds constraints which are not being enforced, an error will be raised. If you want to force the
 constraint to appear in the schema, even though it's not being checked upon parsing, you can use variadic arguments
 to `Field()` with the raw schema attribute name:
 
@@ -81,6 +81,11 @@ to `Field()` with the raw schema attribute name:
 {!.tmp_examples/schema_unenforced_constraints.py!}
 ```
 _(This script is complete, it should run "as is")_
+
+## Modifying schema in custom fields
+
+Custom field types can customise the schema generate for them using the `__modify_schema__` class method,
+see [Custom Data Types](types.md#custom-data-types) for more details.
 
 ## JSON Schema Types
 
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -84,7 +84,7 @@ _(This script is complete, it should run "as is")_
 
 ## Modifying schema in custom fields
 
-Custom field types can customise the schema generate for them using the `__modify_schema__` class method,
+Custom field types can customise the schema generated for them using the `__modify_schema__` class method;
 see [Custom Data Types](types.md#custom-data-types) for more details.
 
 ## JSON Schema Types
--- a/changes/1064-samuelcolvin.md
+++ b/changes/1064-samuelcolvin.md
@@ -1,0 +1,1 @@
+Prevent type attributes being added to schema unless the attribute `__schema_attributes__` is `True` 
--- a/docs/examples/types_custom_type.py
+++ b/docs/examples/types_custom_type.py
@@ -1,21 +1,64 @@
-from pydantic import BaseModel, ValidationError
+import re
+from pydantic import BaseModel
 
-class StrictStr(str):
+# https://en.wikipedia.org/wiki/Postcodes_in_the_United_Kingdom#Validation
+post_code_regex = re.compile(
+    r'(?:'
+    r'([A-Z]{1,2}[0-9][A-Z0-9]?|ASCN|STHL|TDCU|BBND|[BFS]IQQ|PCRN|TKCA) ?'
+    r'([0-9][A-Z]{2})|'
+    r'(BFPO) ?([0-9]{1,4})|'
+    r'(KY[0-9]|MSR|VG|AI)[ -]?[0-9]{4}|'
+    r'([A-Z]{2}) ?([0-9]{2})|'
+    r'(GE) ?(CX)|'
+    r'(GIR) ?(0A{2})|'
+    r'(SAN) ?(TA1)'
+    r')'
+)
+
+class PostCode(str):
+    """
+    Partial UK postcode validation. Note: this is just an example, and is not
+    intended for use in production; in particular this does NOT guarantee
+    a postcode exists, just that it has a valid format.
+    """
     @classmethod
     def __get_validators__(cls):
+        # one or more validators may be yielded which will be called in the
+        # order to validate the input, each validator will receive as an input
+        # the value returned from the previous validator
         yield cls.validate
+
+    @classmethod
+    def __modify_schema__(cls, field_schema):
+        # __modify_schema__ should mutate the dict it receives in place,
+        # the returned value will be ignored
+        field_schema.update(
+            # simplified regex here for brevity, see the wikipedia link above
+            pattern='^[A-Z]{1,2}[0-9][A-Z0-9]? ?[0-9][A-Z]{2}$',
+            # some example postcodes
+            examples=['SP11 9DG', 'w1j7bu'],
+        )
 
     @classmethod
     def validate(cls, v):
         if not isinstance(v, str):
-            raise ValueError(f'strict string: str expected not {type(v)}')
-        return v
+            raise TypeError('string required')
+        m = post_code_regex.fullmatch(v.upper())
+        if not m:
+            raise ValueError('invalid postcode format')
+        # you could also return a string here which would mean model.post_code
+        # would be a string, pydantic won't care but you could end up with some
+        # confusion since the value's type won't match the type annotation
+        # exactly
+        return cls(f'{m.group(1)} {m.group(2)}')
+
+    def __repr__(self):
+        return f'PostCode({super().__repr__()})'
 
 class Model(BaseModel):
-    s: StrictStr
+    post_code: PostCode
 
-print(Model(s='hello'))
-try:
-    print(Model(s=123))
-except ValidationError as e:
-    print(e.json())
+model = Model(post_code='sw8 5el')
+print(model)
+print(model.post_code)
+print(Model.schema())
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -1,5 +1,5 @@
-Where possible *pydantic* uses [standard library types](#standard-library-types) to define fields, thus smoothing 
-the learning curve. For many useful applications, however, no standard library type exists, 
+Where possible *pydantic* uses [standard library types](#standard-library-types) to define fields, thus smoothing
+the learning curve. For many useful applications, however, no standard library type exists,
 so *pydantic* implements [many commonly used types](#pydantic-types).
 
 If no existing type suits your purpose you can also implement your [own pydantic-compatible types](#custom-data-types)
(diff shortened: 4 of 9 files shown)


dmontagu:
Paging @skewty -- seemed like you might fall under the category of people who "have created their own constrained types"?

skewty:
> seemed like you might fall under the category of people who "have created their own constrained types"?

Yup, we do.  I was busy after work yesterday but will be able to look through this at work today.

skewty:
I am torn between the *magic boolean* presented above and something more standard / pythonic like a mixin. (Maybe named something like: `HasSchema` or `HasSchemaAttributes`.)  

The mixin is another import but it coveys knowledge and understanding of the other values (names to avoid) that come into play.  I understand that the mixin idea is a bit more compute intensive to check than a Boolean but felt the idea was worth presentation.

Using the *magic boolean* along with `schema_extra` being a function a lot of the naming issues can be worked around.

Both options above could be done in a minor release.  Neither would require any change to the code my team has.  Especially given #1061, I think that the impact that this "breaking change" would have is minor as those affected would likely be in testing phase still.

samuelcolvin:
I wouldn't mind `HasSchemaAttributes` too much. I don't care much about performance here since not during validation, only at model creation time.

My only question is: is a special mixin which does nothing except allow an `isinstance` check really more pythonic? I honestly don't know.

Probably the best solution would be to add a prefix to all these attributes or put them in some kind of namespace e.g. 
```py
class ConstrainedFoobar:
    __schema_regex__ = ...
    __schema_gt__ = ...
```
or
```py
class ConstrainedFoobar:
    __schema_attributes__ = SchemaAttributes(x, y, z)  # could just be a dict
```
(I prefer the latter)

This seems like a bigger change, but now I come to think about it, I can't see how it would break any case that isn't broken by either the magic mixin or the magic attribute.

What do you think?

skewty:
Yes, I additionally thought of that but didn't mention it as it was more breaking.
That is, in my opinion, a better "long term" approach.

samuelcolvin:
Agreed.

Why did you think it was more breaking?

skewty:
Renaming all fields as such was more breaking:
```py
class ConstrainedFoobar:
    __schema_regex__ = ...
    __schema_gt__ = ...
```

Not your 2nd option in the post above.  I never thought of that option.

I like the following option the best as it is not breaking (the chance of name overlap is very small):
```py
class ConstrainedFoobar:
    __schema_attributes__ = SchemaAttributes(x, y, z)  # could just be a dict
```

samuelcolvin:
Still needs docs. Will do soon.

 But I think in general this is a much better wait of implementing this.

samuelcolvin:
need to fix coverage.

docs/usage/schema.md
```diff
@@ -82,6 +82,11 @@ to `Field()` with the raw schema attribute name:
 ```
 
+## Modifying schema in custom fields
+
+Custom field types can customise the schema generate for them using the `__modify_schema__` class method,
```
dmontagu:
```suggestion
Custom field types can customise the schema generated for them using the `__modify_schema__` class method,
```

laevilgenius:
@samuelcolvin current implementation causes issues with `Path` subclasses (same goes for other classes in `field_class_to_schema_enum_disabled` I presume):
```python
class ConstrainedPath(Path):
    extensions = None

    @classmethod
    def __modify_schema__(cls, schema: dict):
        update_not_none(schema, extensions=cls.extensions)


class ImagePath(ConstrainedPath):
    extensions = ['.jpg']


class Model(BaseModel):
    path: ConstrainedPath  # 'path': {'title': 'Path', 'type': 'string', 'format': 'path'}
    image: ImagePath  # 'image': {'title': 'Image', 'extensions': ['.jpg']}
```
As you can see, `ImagePath` loses `type` and `format` keys from schema.

samuelcolvin:
thank you for pointing this out @laevilgenius, I've created a new issue #1102 and just working on a fix now.

caniko:
I am not sure I follow how this fixes the issue we have with np.ndarray and pydantic. I am still getting this error:
```
RuntimeError: no validator found for <class 'numpy.ndarray'>, see `arbitrary_types_allowed` in Config
```
Furthermore, I find myself forced to adopt [this code-snippet](https://gist.github.com/danielhfrank/00e6b8556eed73fb4053450e602d2434) for numpy arrays.

Correct me if I am wrong, but shouldn't there be something native to popular packages such as numpy, pandas, tensorflow, pytorch, etc?

PR #1093, opened 2019-12-09
Settings config inheritance

### LINKED ISSUE #1091: Settings inheritance with ENV prefix does not work
# Bug

* OS: ubuntu 18.04
* Python version `import sys; print(sys.version)`: 3.7.5 (default, Nov  7 2019, 10:50:52) 
* Pydantic version `import pydantic; print(pydantic.VERSION)`: `1.0b1` or `1.2`, but works fine with `0.32`

```py
import os


class MyBaseSettings(BaseSettings):
    var: int = 10


class MySubSettings(MyBaseSettings):
    class Config:
        env_prefix = 'PREFIX_'


os.environ['PREFIX_VAR'] = '2'
k = MySubSettings()
print(k.var)  # 10, but should be 2
```

We expect that `k.var` will be equal to 2, but it has default value 10.
It's reproduced since 1.0b1.

samuelcolvin:
## Related issue number

fix #1091

## Checklist

### DIFF
--- a/changes/1091-samuelcolvin.md
+++ b/changes/1091-samuelcolvin.md
@@ -1,0 +1,1 @@
+Fix `Config` inheritance on `BaseSettings` when used with `env_prefix`
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -66,7 +66,7 @@ class BaseSettings(BaseModel):
         @classmethod
         def prepare_field(cls, field: ModelField) -> None:
             env_names: Iterable[str]
-            env = field.field_info.extra.pop('env', None)
+            env = field.field_info.extra.get('env')
             if env is None:
                 if field.has_alias:
                     warnings.warn(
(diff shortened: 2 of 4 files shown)


aeldaly-tc:
This is still not working for me.

```
import os
import pydantic


class Thing(pydantic.BaseModel):
    some_var: int = 10

    class Config:
        case_sensitive = True
        env_prefix = "TEST_"


print(Thing())

os.environ["TEST_SOME_VAR"] = "20"
print(os.environ.get("TEST_SOME_VAR"))

print(Thing())

print(pydantic.VERSION)
```

Outputs:

```
 ✔ python3 test.py
some_var=10
20
some_var=10
1.3a1
```

dmontagu:
@aeldaly-tc there are two issues with your snippet:

1. You need to inherit from `pydantic.BaseSettings`, not `pydantic.BaseModel`.

2. You need to set `case_sensitive=False` (since `some_var` is lower case, but `TEST_SOME_VAR` is all uppercase).

After making those changes, it worked for me on MacOS. Let us know if that doesn't fix it for you.

aeldaly-tc:
@dmontagu Thanks so much. It was the `case_sensitive` is the one that I needed to change

PR #1104, opened 2019-12-16
Add support for nested generics

dmontagu:
## Change Summary

Modifies `GenericModel` to behave more similarly to standard python generics, including the ability to reference typevars in nested GenericModels, and to make use of partial specifications with new typevars.

There are a couple outstanding issues:

* [x] As this PR stands, an error will be raised if you try to instantiate a GenericModel after just passing `TypeVar`s as the parameters. Currently, I believe this is allowed, and will perform the same logic that would happen if you were to use a `TypeVar` as a field annotation elsewhere (e.g., for purposes of a constraint).
    * [x] I believe the right (backwards-compatible) approach here is to treat unspecified parameters the same way we would treat it as if the `TypeVar` was used as the annotation. I think this is similar to what we do for `Dict`, `List`, etc., since those `TypeVar`s have no constraints and so are basically equivalent to `Any`.
    * [x] This should be easy to implement, but I wanted to get confirmation before putting in more effort. In particular, I think the "right" implementation would be to allow instantiating before you've provided *any* concrete types for the parameters, similar to how `Dict`, etc., work. **But right now that raises an error. I personally would consider this a fix rather than a breaking change, but I want to reach consensus before moving forward.**
    * [x] (It should also be easy to keep the current behavior, but allow instantiation after explicitly specifying `TypeVar` parameters, I just think that is a wrong API choice -- the question is whether preserving the current behavior is sufficiently worthwhile.)

* [ ] The generated names come out a little weird when you make use of partially-specified models. It would be nice to fix this, but since 1) this is already an edge case, 2) type naming is usually not critical, and 3) we already allow you to override `__concrete_name__`, I am happy to leave this to users for now.

## Related issue number

Closes https://github.com/samuelcolvin/pydantic/issues/967

## Checklist

### DIFF
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -20,12 +20,6 @@ class GenericModel(BaseModel):
         # `GenericModel` to also inherit from `Generic`, which would require changes to the use of `create_model` below.
         __parameters__: ClassVar[Tuple[TypeVarType, ...]]
 
-    def __new__(cls, *args: Any, **kwargs: Any) -> Any:
-        if cls.__concrete__:
-            return super().__new__(cls)
-        else:
-            raise TypeError(f'Type {cls.__name__} cannot be used without generic parameters, e.g. {cls.__name__}[T]')
-
     # Setting the return type as Type[Any] instead of Type[BaseModel] prevents PyCharm warnings
     def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[Type[Any], ...]]) -> Type[Any]:
         cached = _generic_types_cache.get((cls, params))
@@ -63,7 +57,7 @@ class GenericModel(BaseModel):
             ),
         )
         created_model.Config = cls.Config
-        concrete = all(not isinstance(v, TypeVar) for v in concrete_type_hints.values())  # type: ignore
+        concrete = all(not _is_typevar(v) for v in concrete_type_hints.values())
         created_model.__concrete__ = concrete
         if not concrete:
             parameters = tuple(v for v in concrete_type_hints.values() if _is_typevar(v))
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -1,15 +1,24 @@
-from typing import Any, ClassVar, Dict, Generic, Tuple, Type, TypeVar, Union, get_type_hints
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, Tuple, Type, TypeVar, Union, cast, get_type_hints
 
 from .class_validators import gather_all_validators
+from .fields import FieldInfo, ModelField
 from .main import BaseModel, create_model
+from .utils import lenient_issubclass
 
 _generic_types_cache: Dict[Tuple[Type[Any], Union[Any, Tuple[Any, ...]]], Type[BaseModel]] = {}
 GenericModelT = TypeVar('GenericModelT', bound='GenericModel')
+TypeVarType = Any  # since mypy doesn't allow the use of TypeVar as a type
 
 
 class GenericModel(BaseModel):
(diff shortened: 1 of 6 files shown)


pydantic/generics.py
```diff
@@ -78,7 +93,30 @@ def resolve_type_hint(type_: Any, typevars_map: Dict[Any, Any]) -> Type[Any]:
...
+
+
+def _parameterize_generic_field(field_type: Type[Any], typevars_map: Dict[TypeVarType, Type[Any]]) -> Type[Any]:
+    if lenient_issubclass(field_type, GenericModel) and not field_type.__concrete__:
+        parameters = tuple(typevars_map.get(param, param) for param in field_type.__parameters__)
+        field_type = field_type[parameters]
+    return field_type
+
+
+def _is_typevar(v: Any) -> bool:
```
samuelcolvin: there's at least one more place above this can be used where we currently have `isinstance(param, TypeVar)`

dmontagu: Thanks, fixed

pydantic/generics.py
```diff
@@ -1,21 +1,24 @@
...
 _generic_types_cache: Dict[Tuple[Type[Any], Union[Any, Tuple[Any, ...]]], Type[BaseModel]] = {}
 GenericModelT = TypeVar('GenericModelT', bound='GenericModel')
+TypeVarType = Any  # since mypy doesn't allow the use of TypeVar as a type
 
 
 class GenericModel(BaseModel):
     __slots__ = ()
     __concrete__: ClassVar[bool] = False
 
-    def __new__(cls, *args: Any, **kwargs: Any) -> Any:
```
dmontagu: (This was only necessary to prevent instantiation with unspecified parameters.)

tests/test_generics.py
```diff
@@ -444,3 +425,104 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+    BT = TypeVar('BT')
+
+    class Model(GenericModel, Generic[AT, BT]):
+        a: AT
+        b: BT
+
+    partial_model = Model[int, BT]
+    assert partial_model.__name__ == 'Model[int, BT]'
+    concrete_model = partial_model[str]
+    assert concrete_model.__name__ == 'Model[int, BT][str]'
```
dmontagu: This is what I mean by the names are kind of weird.

It would be a relatively involved refactor to make it generate the names properly. Not impossible though if you'd prefer we make it work.

tests/test_generics.py
```diff
@@ -444,3 +425,137 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+
+    with pytest.raises(ValidationError) as exc_info:
+        partial_model(a='a', b=2)
+    assert exc_info.value.errors() == [
+        {'loc': ('a',), 'msg': 'value is not a valid integer', 'type': 'type_error.integer'}
+    ]
+
+
+@skip_36
+def test_partial_specification_instantiation_bounded():
```
dmontagu: This test demonstrates that the typevars for partially-parametrized models are respected properly.

samuelcolvin:
I think this looks fine, but I'm not going to pretend I've followed every line of the discussion.

Does it need any updates to docs? Otherwise can it be merged?

dmontagu:
I think the docs should be updated, but I think it can be a small update. Give me 15 mins and I'll respond here an estimate of how long it will take to finish.

dmontagu:
@samuelcolvin Okay, I've added docs, let me know if you want any changes.

samuelcolvin:
awesome, thank you.

PR #1110, opened 2019-12-18
docs: Fix explanation of case sensitivity for `Settings`

### LINKED ISSUE #1105: Settings - case sensitivity - misleading example in docs
# Bug

Please complete:
* OS: **Linux**
* Python version `import sys; print(sys.version)`: **3.6.9**

There is an example in docs about case sensitivity of environment variable names: https://pydantic-docs.helpmanual.io/usage/settings/#environment-variable-names.

I cite it here for convenience:

> Case-sensitivity can be turned on through the Config:

```python

class Settings(BaseSettings):
    redis_host = 'localhost'

    class Config:
        case_sensitive = True
```

> When `case_sensitive` is `True`, the environment variable must be in all-caps, so in this example `redis_host` could only be modified via `export REDIS_HOST`.

This example is simply incorrect. It leads that you can name `Settings` attrs using "snake_case", and export env variable named "UPPER_CASE", and `Settings` will catch them and populate proper attr with value taken from this env variable. This is not the case. In fact, if you want case sensitive environment variable names, you *must* name `Settings` attributes all "UPPER_CASE" as well.

## Proof

Let's take this example "as is" and try to override settings attribute using environment variable:
```python
import os

class Settings(BaseSettings):
    redis_host = 'localhost'

    class Config:
        case_sensitive = True

Settings().redis_host
'localhost'  # default

os.environ['REDIS_HOST'] = 'myredis'  # according to example, we must name env var all UPPER_CASE
Settings().redis_host
'localhost'  # still default
os.environ['redis_host'] = 'myredis'
Settings().redis_host
'myredis'  # now it's working
```

tribals:
## Change Summary

Fix incorrect explanation of how `BaseSettings` treats environment variable names when populating attributes.

## Related issue number

Closes #1105 

## Checklist

* [ ] Unit tests for the changes exist (not appropriate?)
* [ ] Tests pass on CI and coverage remains at 100% (not appropriate?)
* [x] `changes/1105-tribals.md` file added describing change

### DIFF
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -50,7 +50,8 @@ Case-sensitivity can be turned on through the `Config`:
 {!.tmp_examples/settings_case_sensitive.py!}
 ```
 
-When `case_sensitive` is `True`, the environment variable must be exactly the same as attribute, so in this example
+When `case_sensitive` is `True`, the environment variable names must match field names (optionally with a prefix),
+so in this example
 `redis_host` could only be modified via `export redis_host`. And if you want to name environment variables
 all upper-case, you should name attribute all upper-case too. You can still name environment variables anything
 you like through `Field(..., env=...)`.
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -52,7 +52,7 @@ Case-sensitivity can be turned on through the `Config`:
 
 When `case_sensitive` is `True`, the environment variable must be exactly the same as attribute, so in this example
 `redis_host` could only be modified via `export redis_host`. And if you want to name environment variables
-all upper-case, you should name attribute all upper-case too. I can still name environment variable anything
+all upper-case, you should name attribute all upper-case too. You can still name environment variables anything
 you like through `Field(..., env=...)`.
 
 !!! note
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -50,8 +50,10 @@ Case-sensitivity can be turned on through the `Config`:
 {!.tmp_examples/settings_case_sensitive.py!}
 ```
 
-When `case_sensitive` is `True`, the environment variable must be in all-caps,
-so in this example `redis_host` could only be modified via `export REDIS_HOST`.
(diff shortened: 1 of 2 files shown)


docs/usage/settings.md
```diff
@@ -50,8 +50,10 @@ Case-sensitivity can be turned on through the `Config`:
 {!.tmp_examples/settings_case_sensitive.py!}
 ```
 
-When `case_sensitive` is `True`, the environment variable must be in all-caps,
-so in this example `redis_host` could only be modified via `export REDIS_HOST`.
+When `case_sensitive` is `True`, the environment variable must be exactly the same as attribute, so in this example
+`redis_host` could only be modified via `export redis_host`. And if you want to name environment variables
+all upper-case, you should name attribute all upper-case too. I can still name environment variable anything
```
samuelcolvin:
```suggestion
all upper-case, you should name attribute all upper-case too. You can still name environment variables anything
```

docs/usage/settings.md
```diff
@@ -50,8 +50,10 @@ Case-sensitivity can be turned on through the `Config`:
 {!.tmp_examples/settings_case_sensitive.py!}
 ```
 
-When `case_sensitive` is `True`, the environment variable must be in all-caps,
-so in this example `redis_host` could only be modified via `export REDIS_HOST`.
+When `case_sensitive` is `True`, the environment variable must be exactly the same as attribute, so in this example
```
samuelcolvin:
```suggestion
When `case_sensitive` is `True`, the environment variable names must match field names (optionally with a prefix),
so in this example
```

docs/usage/settings.md
```diff
@@ -50,8 +50,10 @@ Case-sensitivity can be turned on through the `Config`:
...
+`redis_host` could only be modified via `export redis_host`. And if you want to name environment variables
```
samuelcolvin:
```suggestion
`redis_host` could only be modified via `export redis_host`. If you want to name environment variables
```

samuelcolvin:
great, thank you.

PR #1114, opened 2019-12-18
Trivial - Fix comment spelling

therefromhere:


docs/examples/schema_main.py
```diff
@@ -29,5 +29,5 @@ class MainModel(BaseModel):
     class Config:
         title = 'Main'
 
-# this is equivilant of json.dumps(MainModel.schema(), indent=2):
+# this is equivalent of json.dumps(MainModel.schema(), indent=2):
```
dmontagu:
```suggestion
# this is equivalent to json.dumps(MainModel.schema(), indent=2):
```

Seems better to me? Thanks for submitting the fix!

samuelcolvin:
thank you.

PR #1119, opened 2019-12-20
Documentation: rename benchmark django-restful-framework -> django-rest-framework

frankie567:
## Change Summary

I came across this while reading the [benchmark page](https://pydantic-docs.helpmanual.io/benchmarks/) in the docs: benchmark for [`django-rest-framework`](https://www.django-rest-framework.org/) was named `django-restful-framework`. This confused me a bit.

I changed the package name in the benchmark script and run the script instead of modifying directly the generated MD file, as specified.

Problem is that, on my machine, the benchmark is not in favor of pydantic 😟So I don't know what you prefer for this (I would be okay to just edit the previous results).

## Related issue number

Nope.

## Checklist

### DIFF
--- a/benchmarks/test_drf.py
+++ b/benchmarks/test_drf.py
@@ -10,7 +10,7 @@ from rest_framework import __version__, serializers
 
 
 class TestDRF:
-    package = 'django-restful-framework'
+    package = 'django-rest-framework'
     version = __version__
 
     def __init__(self, allow_extra):
--- a/changes/1119-frankie567.md
+++ b/changes/1119-frankie567.md
@@ -1,0 +1,1 @@
+Rename django-rest-framework benchmark in documentation
--- a/docs/.benchmarks_table.md
+++ b/docs/.benchmarks_table.md
@@ -2,10 +2,10 @@
 
 Package | Version | Relative Performance | Mean validation time
 --- | --- | --- | ---
-valideer | `0.4.2` |  | 83.1μs
-attrs + cattrs | `19.3.0` | 1.0x slower | 84.1μs
-pydantic | `1.2` | 1.1x slower | 92.7μs
-marshmallow | `3.3.0` | 1.7x slower | 141.8μs
-trafaret | `2.0.2` | 2.4x slower | 202.0μs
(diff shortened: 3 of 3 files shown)


dmontagu:
Did you cythonize before running the benchmark locally? That will make pydantic about 2x faster.

samuelcolvin:
Also, I always use the same `cases.json` for consistency, you can download it from [here](https://gist.github.com/samuelcolvin/b99b6c1bea779f31197382ffab367bb8).

---

I know that `.benchmarks_table.md` says "do not edit", but I think here for simplicity you can just do a find a replace on "django-restful-framework" and thereby edit both files without regenerating benchmarks.

frankie567:
@dmontagu I didn't, and TBH I never did that. Shouldn't it be mentioned in the benchmarks page? It just says "run with Python 3.7.4". That doesn't seem very fair regarding the other libs.

@samuelcolvin Ok, I'll do that 👍

samuelcolvin:
> @dmontagu I didn't, and TBH I never did that. Shouldn't it be mentioned in the benchmarks page? It just says "run with Python 3.7.4". That doesn't seem very fair regarding the other libs.

The benchmarks say:

> Benchmarks were run with Python 3.7.4 and the package versions listed above installed via pypi on Ubuntu 18.04.

I think that's the fair way of running benchmarks - with the library versions distributed via pypi. 

Pydantic provides binaries compiled with cython from python which is modified in numerous ways to be compilable and as fast as possible when compiled - that's one of the reasons it's fast. If other libraries provided similar binaries for linux, they would be used too; some may already be, I haven't checked.

frankie567:
@samuelcolvin Thank you, I understand better now. So, since I ran the benchmark with a pydantic version pulled from PyPi (not the one on my local machine), I guess then it actually ran with the Cython optimizations?

samuelcolvin:
Only if you're on linux, you can checked via `print(pydantic.compiled)`.

frankie567:
I didn't then, i was on macOS. Thank you for providing those details 👍

samuelcolvin:
thanks.

PR #1132, opened 2019-12-26
improve pydantic import time

### LINKED ISSUE #1039: Change `make` commands order in Makefile
# Question

Currently calling `make` in project will first run all tests and only after that will run lint checks.
Is there any reason for that?
It doesn't make sense to me as code analysis are complete much faster. Now I have to either run code analyses separately before running make, or wait while all tests will pass.

So can we change this:
```bash
.PHONY: all
all: testcov lint mypy
```
to this:
```
.PHONY: all
all:  lint mypy testcov
```
?

### LINKED ISSUE #1127: Import slowdown after v1
After upgrading to v1, I've noticed that it takes almost double the time to import Pydantic.  Observe:

```
$ python3 -m pip install -U 'pydantic<1'
[...]
Successfully installed pydantic-0.32.2

$ seq 2 | xargs -I -- python3 -X importtime -c '
  import pydantic
  print("pydantic version:", pydantic.VERSION)
  ' 2>&1 | begin
      head -n 1
      tail -n 2
  end
import time: self [us] | cumulative | imported package
import time:      1058 |      70730 | pydantic
pydantic version: 0.32.2

$ python3 -m pip install -U pydantic
[...]
Successfully installed pydantic-1.3

$ seq 2 | xargs -I -- python3 -X importtime -c '
  import pydantic
  print("pydantic version:", pydantic.VERSION)
  ' 2>&1 | begin
      head -n 1
      tail -n 2
  end
import time: self [us] | cumulative | imported package
import time:       802 |     128663 | pydantic
pydantic version: 1.3
```

v0.32.2 takes approximately 70ms to import on my MBP; v1.3 takes 130ms.

attrs for comparison takes 36ms:

```
$ seq 2 | xargs -I -- python3 -X importtime -c '
  import attr
  print("attrs version:", attr.__version__)
  ' 2>&1 | begin
      head -n 1
      tail -n 2
  end
import time: self [us] | cumulative | imported package
import time:       882 |      35659 | attr
attrs version: 19.3.0
```

I can understand that this might not be something you'd want to spend time on, depending on what your direction for Pydantic is; but slow warmups have been plaguing Python CLIs for a very long time.  If you'd like to investigate this further, [tuna](https://github.com/nschloe/tuna) can help you visualise your dependency tree.  Here's what it's showing me for `tuna (python3 -X importtime -c 'import pydantic' 2>| psub)`:

Screenshot_2019-12-24 tuna - var folders 18 r2nky_v137d3fwhtx0644m940000gn T psub WY0ulb7Mci

There is the unfortunate practice in Python of importing the entire package's contents in `__init__` and (unfortunately) once you've done that, there's no going back - not without a breaking release anyway.

As always, thanks for all your work on Pydantic.

samuelcolvin:
## Change Summary

Roughly halve the time to import pydantic, see #1127 for discussion.

Changes:
* `dataclasses` is only imported when used
* three regexes `url_regex`, `ascii_domain_regex` and `int_domain_regex` in `network.py` now only compiled when used. Thus `url_regex` etc. are now accessible as a function `url_regex()`
* `email_validator` is only imported when it's used.
* `pydantic.VERSION` is no longer a `StrictVersion`, just a standard str
* `version_info` will only import `platform` and `sys` when it's called
* `Color` regexes are not compiled until used.

## Related issue number

fix #1127 and also fix #1039

## Checklist

* [ ] ~~Documentation reflects the changes where applicable~~

### DIFF
--- a/pydantic/color.py
+++ b/pydantic/color.py
@@ -42,17 +42,18 @@ class RGBA:
         return self._tuple[item]
 
 
-r_hex_short = re.compile(r'\s*(?:#|0x)?([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])?\s*')
-r_hex_long = re.compile(r'\s*(?:#|0x)?([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})?\s*')
+# these are not compiled here to avoid import slowdown, they'll be compiled the first time they're used, then cached
+r_hex_short = r'\s*(?:#|0x)?([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])?\s*'
+r_hex_long = r'\s*(?:#|0x)?([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})?\s*'
 _r_255 = r'(\d{1,3}(?:\.\d+)?)'
 _r_comma = r'\s*,\s*'
-r_rgb = re.compile(fr'\s*rgb\(\s*{_r_255}{_r_comma}{_r_255}{_r_comma}{_r_255}\)\s*')
+r_rgb = fr'\s*rgb\(\s*{_r_255}{_r_comma}{_r_255}{_r_comma}{_r_255}\)\s*'
 _r_alpha = r'(\d(?:\.\d+)?|\.\d+|\d{1,2}%)'
-r_rgba = re.compile(fr'\s*rgba\(\s*{_r_255}{_r_comma}{_r_255}{_r_comma}{_r_255}{_r_comma}{_r_alpha}\s*\)\s*')
+r_rgba = fr'\s*rgba\(\s*{_r_255}{_r_comma}{_r_255}{_r_comma}{_r_255}{_r_comma}{_r_alpha}\s*\)\s*'
 _r_h = r'(-?\d+(?:\.\d+)?|-?\.\d+)(deg|rad|turn)?'
 _r_sl = r'(\d{1,3}(?:\.\d+)?)%'
-r_hsl = re.compile(fr'\s*hsl\(\s*{_r_h}{_r_comma}{_r_sl}{_r_comma}{_r_sl}\s*\)\s*')
-r_hsla = re.compile(fr'\s*hsl\(\s*{_r_h}{_r_comma}{_r_sl}{_r_comma}{_r_sl}{_r_comma}{_r_alpha}\s*\)\s*')
+r_hsl = fr'\s*hsl\(\s*{_r_h}{_r_comma}{_r_sl}{_r_comma}{_r_sl}\s*\)\s*'
+r_hsla = fr'\s*hsl\(\s*{_r_h}{_r_comma}{_r_sl}{_r_comma}{_r_sl}{_r_comma}{_r_alpha}\s*\)\s*'
 
 # colors where the two hex characters are the same, if all colors match this the short version of hex colors can be used
 repeat_colors = {int(c * 2, 16) for c in '0123456789abcdef'}
@@ -226,7 +227,7 @@ def parse_str(value: str) -> RGBA:
     else:
         return ints_to_rgba(r, g, b, None)
 
-    m = r_hex_short.fullmatch(value_lower)
+    m = re.fullmatch(r_hex_short, value_lower)
     if m:
         *rgb, a = m.groups()
         r, g, b = [int(v * 2, 16) for v in rgb]
@@ -236,7 +237,7 @@ def parse_str(value: str) -> RGBA:
             alpha = None
         return ints_to_rgba(r, g, b, alpha)
 
-    m = r_hex_long.fullmatch(value_lower)
+    m = re.fullmatch(r_hex_long, value_lower)
     if m:
         *rgb, a = m.groups()
         r, g, b = [int(v, 16) for v in rgb]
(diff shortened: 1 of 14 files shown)


samuelcolvin:
For me on my laptop the import time has dropped from ~143md to ~53ms

samuelcolvin:
okay I've just tried this again on my desktop in the office (after rebasing):

I get the following difference:
```
master           89-94ms
import-speedup   28-49ms
```

I tried with both simple python and pydantic compiled and get similar results. I don't know what @layday's numbers [here](https://github.com/samuelcolvin/pydantic/issues/1127#issuecomment-569341353) are different.

I think that's pretty good considering there's no functionality changes.

layday:
It might just be you've got better specs, I don't really know.  As for suggestions for improvement, I was able to shave off about 10 milliseconds by nesting `inspect`.  The call to `inspect.isgenerator` in utils can be inlined:

```py
from types import GeneratorType

def sequence_like(v: AnyType) -> bool:
    return isinstance(v, (list, tuple, set, frozenset, types.GeneratorType))
```

samuelcolvin:
Altering `inspect` imports isn't making any difference for me.

For some reason the log for `pydantic.main` seems to always include importing `inspect`, even though there's no visible "import inspect..":

Logs:
```
...
import time:       297 |        297 |     pydantic.fields
import time:       134 |        134 |           _weakrefset
import time:       249 |        383 |         weakref
import time:        60 |         60 |             org
import time:        11 |         70 |           org.python
import time:        12 |         82 |         org.python.core
import time:       138 |        602 |       copy
import time:        60 |         60 |             _struct
import time:       103 |        162 |           struct
import time:       162 |        162 |           _compat_pickle
import time:        53 |         53 |               org
import time:        88 |        140 |             org.python
import time:        11 |        150 |           org.python.core
import time:        73 |         73 |           _pickle
import time:       548 |       1094 |         pickle
import time:       200 |       1294 |       pydantic.parse
import time:       551 |        551 |         pydantic.networks
import time:       421 |        971 |       pydantic.schema
import time:        58 |         58 |       cython
import time:        85 |         85 |             _opcode
import time:       172 |        256 |           opcode
import time:       258 |        514 |         dis
import time:       112 |        112 |           importlib
import time:       111 |        222 |         importlib.machinery
import time:       158 |        158 |             token
import time:       641 |        798 |           tokenize
import time:        88 |        885 |         linecache
import time:      1379 |       2999 |       inspect   <--- why is this happening!?
import time:       923 |       6844 |     pydantic.main
import time:       291 |      27556 |   pydantic.dataclasses
import time:       194 |        194 |   pydantic.env_settings
import time:       603 |        603 |   pydantic.tools
import time:        71 |         71 |   pydantic.version
import time:       196 |      28618 | pydantic
```

see the [`import-speedup-inspect` branch](https://github.com/samuelcolvin/pydantic/tree/import-speedup-inspect) for what I changed.

layday:
I should've probably been clearer, `inspect` is also used in `class_validators` and `schema`.

```patch
diff --git a/pydantic/class_validators.py b/pydantic/class_validators.py
index 8ac62f0..12f9e36 100644
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -1,7 +1,6 @@
 import warnings
 from collections import ChainMap
 from functools import wraps
-from inspect import Signature, signature
 from itertools import chain
 from types import FunctionType
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Type, Union, overload
@@ -10,6 +9,9 @@ from .errors import ConfigError
 from .typing import AnyCallable
 from .utils import in_ipython
 
+if TYPE_CHECKING:
+    from inspect import Signature
+
 
 class Validator:
     __slots__ = 'func', 'pre', 'each_item', 'always', 'check_fields', 'skip_on_failure'
@@ -196,6 +198,8 @@ def extract_root_validators(namespace: Dict[str, Any]) -> Tuple[List[AnyCallable
     for name, value in namespace.items():
         validator_config: Optional[Validator] = getattr(value, ROOT_VALIDATOR_CONFIG_KEY, None)
         if validator_config:
+            from inspect import signature
+
             sig = signature(validator_config.func)
             args = list(sig.parameters.keys())
             if args[0] == 'self':
@@ -231,6 +235,8 @@ def make_generic_validator(validator: AnyCallable) -> 'ValidatorCallable':
     It's done like this so validators don't all need **kwargs in their signature, eg. any combination of
     the arguments "values", "fields" and/or "config" are permitted.
     """
+    from inspect import signature
+
     sig = signature(validator)
     args = list(sig.parameters.keys())
     first_arg = args.pop(0)
@@ -254,7 +260,7 @@ def prep_validators(v_funcs: Iterable[AnyCallable]) -> 'ValidatorsList':
 all_kwargs = {'values', 'field', 'config'}
 
 
-def _generic_validator_cls(validator: AnyCallable, sig: Signature, args: Set[str]) -> 'ValidatorCallable':
+def _generic_validator_cls(validator: AnyCallable, sig: 'Signature', args: Set[str]) -> 'ValidatorCallable':
     # assume the first argument is value
     has_kwargs = False
     if 'kwargs' in args:
@@ -288,7 +294,7 @@ def _generic_validator_cls(validator: AnyCallable, sig: Signature, args: Set[str
         return lambda cls, v, values, field, config: validator(cls, v, values=values, field=field, config=config)
 
 
-def _generic_validator_basic(validator: AnyCallable, sig: Signature, args: Set[str]) -> 'ValidatorCallable':
+def _generic_validator_basic(validator: AnyCallable, sig: 'Signature', args: Set[str]) -> 'ValidatorCallable':
     has_kwargs = False
     if 'kwargs' in args:
         has_kwargs = True
diff --git a/pydantic/schema.py b/pydantic/schema.py
index 528525a..9f9da83 100644
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -1,4 +1,3 @@
-import inspect
 import re
 import warnings
 from datetime import date, datetime, time, timedelta
@@ -451,6 +450,8 @@ def model_process_schema(
     sub-models of the returned schema will be referenced, but their definitions will not be included in the schema. All
     the definitions are returned as the second value.
     """
+    import inspect
+
     ref_prefix = ref_prefix or default_prefix
     known_models = known_models or set()
     s = {'title': model.__config__.title or model.__name__}

```

samuelcolvin:
of course, that's what I did, see the branch I linked to, I've now added that commit here.

samuelcolvin:
I still don't understand the inspect thing, but I think this is a good start.

layday:
Great, thanks for working on this.

pydantic/color.py
```diff
@@ -246,20 +247,20 @@ def parse_str(value: str) -> RGBA:
...
     if m:
         return ints_to_rgba(*m.groups(), None)  # type: ignore
 
-    m = r_rgba.fullmatch(value_lower)
+    m = re.fullmatch(r_rgba, value_lower)
     if m:
         return ints_to_rgba(*m.groups())  # type: ignore
 
-    m = r_hsl.fullmatch(value_lower)
+    m = re.fullmatch(r_hsl, value_lower)
```
MrMrRobat: Wait, isn't use of `re.fullmatch` instead of using compiled patterns slowed down function calls?

samuelcolvin: only very small, `re.fullmatch` keeps a cache of compiled regexes to use after the first call.

So the old slow down is looking up the cache of regexes.

PR #1135, opened 2019-12-30
docs: Add docs for `allow_reuse`

pmav99:
## Change Summary

adding docs for `enable_reuse`

## Related issue number

Issue #940 
Continuation of #941

### DIFF
--- a/docs/examples/validators_allow_reuse.py
+++ b/docs/examples/validators_allow_reuse.py
@@ -1,0 +1,21 @@
+from pydantic import BaseModel, validator
+
+def normalize(name: str) -> str:
+    return ' '.join((word.capitalize()) for word in name.split(' '))
+
+class Producer(BaseModel):
+    name: str
+
+    # validators
+    _normalize_name = validator('name', allow_reuse=True)(normalize)
+
+class Consumer(BaseModel):
+    name: str
+
+    # validators
+    _normalize_name = validator('name', allow_reuse=True)(normalize)
+
+jane_doe = Producer(name='JaNe DOE')
+john_doe = Consumer(name='joHN dOe')
+assert jane_doe.name == 'Jane Doe'
+assert john_doe.name == 'John Doe'
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -64,6 +64,27 @@ _(This script is complete, it should run "as is")_
 You'll often want to use this together with `pre`, since otherwise with `always=True`
 *pydantic* would try to validate the default `None` which would cause an error.
 
+## Reuse validators
+
+Occasionally, you will want to use the same validator on multiple fields/models (e.g. to
+normalize some input data). The "naive" approach would be to write a separate function,
+then call it from multiple decorators.  Obviously, this entails a lot of repetition and
+boiler plate code. To circumvent this, the `allow_reuse` parameter has been added to
+`pydantic.validator` in **v1.2** (`False` by default):
+
+```py
+{!.tmp_examples/validators_allow_reuse.py!}
+```
+_(This script is complete, it should run "as is")_
+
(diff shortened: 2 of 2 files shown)


dmontagu:
Overall I think this is great!

The one semi-large thing I would change would be to replace the examples with examples where you use the same validator on *different* models, rather than multiple fields of the *same* model.

You can already use the same validator for multiple fields by just listing out the name of each field manually, and so I worry it might confuse people / establish the wrong habit if the only examples we show are something that can (and arguably should) be done using a different approach.

That said, I think it would still make sense to add a note that you can also do this to add the same validator multiple times to multiple fields of the same model.

docs/examples/validators_allow_reuse.py
```diff
@@ -0,0 +1,25 @@
+from pydantic import BaseModel, validator
+
+def normalize(name: str) -> str:
+    return " ".join((word.capitalize()) for word in name.split(" "))
```
samuelcolvin:
```suggestion
    return ' '.join(word.capitalize() for word in name.split(' '))
```

single quotes throughout please.

docs/usage/validators.md
```diff
@@ -64,6 +64,51 @@ _(This script is complete, it should run "as is")_
 You'll often want to use this together with `pre`, since otherwise with `always=True`
 *pydantic* would try to validate the default `None` which would cause an error.
 
+## Reuse validators
+
+Occasionally, you will want to use the same validator on multiple fields/models (e.g. to
+normalize some input data). The "naive" approach would be to write a separate
```
samuelcolvin: I don't think we need the "naive" example, you can either omit this sentence completely or say something like

> the naive approach would be to implement a separate function, then call it from multiple decorators. Obviously, this entails a lot of repetition and ...

docs/usage/validators.md
```diff
@@ -64,6 +64,51 @@ _(This script is complete, it should run "as is")_
...
+classmethod for each field. E.g.:
+
+```py
+{!.tmp_examples/validators_allow_reuse.py!}
+```
+_(This script is complete, it should run "as is")_
+
+Obviously, this entails a lot of repetition and boiler plate code. To circumvent this,
+the `allow_reuse` parameter has been added to `pydantic.validator` in version 1.2
```
samuelcolvin:
```suggestion
the `allow_reuse` parameter has been added to `pydantic.validator` in **v1.2**
```
to be consistent with the rest of the docs.

docs/usage/validators.md
```diff
@@ -64,6 +64,51 @@ _(This script is complete, it should run "as is")_
...
+```py
+{!.tmp_examples/validators_allow_reuse2.py!}
+```
+_(This script is complete, it should run "as is")_
+
+As it is obvious, repetition has been reduced and the models become again almost
+declarative.
+
+!!! note
+    In case you find the syntax strange - it is indeed unusual! - you should [read a bit
```
samuelcolvin: I don't think we need this, also I don't think just linking to a google search is helpful. It feels a bit passive aggressive like lmgtfy

docs/usage/validators.md
```diff
@@ -64,6 +64,49 @@ _(This script is complete, it should run "as is")_
...
+    more](https://www.google.com/search?q=python+decorators+syntax) about python
+    decorators.
+
+!!! tip
+    If you have a lot of fields that you want to validate, it usually makes sense to
```
samuelcolvin: good to have this tip, but I don't think it needs it's own code example. Also all code should be in python files so they can be linted etc.

pmav99: It was a complete example that's why I chose to have it inline. I will remove it.

samuelcolvin:
also probably good to add a change description.

samuelcolvin:
great, thank you.

PR #1139, opened 2020-01-03
add exclude_private parameter

H4dr1en:
## Change Summary

Add `exclude_private` parameter to `BaseModel.dict` and `BaseModel.json`, allow `BaseModel.__setattr__` to register properties if `extra == Extra.ignore`.

This PR would solve the problem described in the issue linked below: User can add private properties at runtime and exclude them from export.

## Related issue number
#655 

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -277,7 +277,7 @@ class BaseModel(metaclass=ModelMetaclass):
 
     @no_type_check
     def __setattr__(self, name, value):
-        if self.__config__.extra is not Extra.allow and name not in self.__fields__:
+        if self.__config__.extra is Extra.forbid and name not in self.__fields__:
             raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')
         elif not self.__config__.allow_mutation:
             raise TypeError(f'"{self.__class__.__name__}" is immutable and does not support item assignment')
@@ -307,6 +307,7 @@ class BaseModel(metaclass=ModelMetaclass):
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
         exclude_none: bool = False,
+        exclude_private: bool = True,
     ) -> 'DictStrAny':
         """
         Generate a dictionary representation of the model, optionally specifying which fields to include or exclude.
@@ -332,6 +333,7 @@ class BaseModel(metaclass=ModelMetaclass):
                 exclude_unset=exclude_unset,
                 exclude_defaults=exclude_defaults,
                 exclude_none=exclude_none,
+                exclude_private=exclude_private
             )
         }
 
@@ -351,6 +353,7 @@ class BaseModel(metaclass=ModelMetaclass):
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
         exclude_none: bool = False,
+        exclude_private: bool = True,
         encoder: Optional[Callable[[Any], Any]] = None,
         **dumps_kwargs: Any,
     ) -> str:
@@ -373,6 +376,7 @@ class BaseModel(metaclass=ModelMetaclass):
             exclude_unset=exclude_unset,
             exclude_defaults=exclude_defaults,
             exclude_none=exclude_none,
+            exclude_private=exclude_private,
         )
         if self.__custom_root_type__:
             data = data[ROOT_KEY]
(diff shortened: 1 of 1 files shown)


samuelcolvin:
This is failing, but even if it was passing I'm not sure I agree with it.

It adds even more logic and arguments to `.dict()`, I think you can either use `exclude` or implement your own logic (perhaps on `_iter`, see #997).

pydantic/main.py
```diff
@@ -277,7 +277,7 @@ def __init__(__pydantic_self__, **data: Any) -> None:
 
     @no_type_check
     def __setattr__(self, name, value):
-        if self.__config__.extra is not Extra.allow and name not in self.__fields__:
+        if self.__config__.extra is Extra.forbid and name not in self.__fields__:
```
```suggestion
        if name.startswith('_'):
            self.__dict__[name] = value
            return
        if self.__config__.extra is not Extra.allow and name not in self.__fields__:
```

pydantic/main.py
```diff
@@ -307,6 +307,7 @@ def dict(
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
         exclude_none: bool = False,
+        exclude_private: bool = True,
```
samuelcolvin: remove this, I think it's safe to always exclude anything in `__dict__` where the key startswith with `_`.

MrMrRobat: But this would break concept when we have only parsed values in `__dict__` and will force additional checks, etc.

MrMrRobat:
```py
from pydantic import BaseModel, Extra

class Record(BaseModel):
    class Config:
        extra = Extra.allow

print(Record(_a='b'))  # _a='b'
```

MrMrRobat:
Personally, I think that the best solution for this problem, when you want to set your own private attrs which must not be included in fields is to use `__slots__`, as proposed in https://github.com/samuelcolvin/pydantic/issues/655#issuecomment-570312649

Maybe it should be documented as a standard way of doing it to avoid confusion in future. I don't see any disadvantages of `__slots__` approach, it solves the problem, it keeps `__dict__` only for fields and makes things simpler with no excuses.

samuelcolvin:
I agree with @MrMrRobat and also see the performance concerns, let's discuss other solutions on #655.

MaxNoe:
@MrMrRobat 

> I don't see any disadvantages of __slots__ approach

Maybe a disadvantage is, that it does not really work?

```

In [2]: class Component(BaseModel): 
   ...:     __slots__ = ('log', ) 
   ...:     foo: int 
   ...:     bar: float 
   ...:                                                                                                                                       

In [3]: c = Component(foo=1, bar=3.5, log='test')                                                                                             

In [4]: c.log                                                                                                                                 
AttributeError                            Traceback (most recent call last)
<ipython-input-4-706395a84d01> in <module>
----> 1 c.log

AttributeError: log

In [5]: c.log = 'test'                                                                                                                        
ValueError                                Traceback (most recent call last)
<ipython-input-5-8fce80748249> in <module>
----> 1 c.log = 'test'

~/.local/lib/python3.8/site-packages/pydantic/main.cpython-38-x86_64-linux-gnu.so in pydantic.main.BaseModel.__setattr__()

ValueError: "Component" object has no field "log"
```

samuelcolvin:
> Maybe a disadvantage is, that it does not really work?

No, not good enough. Please reword your comments to be more polite. 

You might also think about crafting a more articulate description of your point here - I'm not really sure what you point is.

MaxNoe:
Rereading the comment mentioning slots, it also adds `object.__setattr__` to actually set the field, sorry I missed that in the first read.

But that does not really address private variables to be used like normal members, which was the point of this PR right?

At least for us, the slots approach does not really solve our problem.
We'd reallly like to use pydantic for configuration but need to have additional properties or members not handled by pydantic. For this, there is currently not really a way to get this working, right?

This PR would have been a great step into this direction.

trallnag:
@maxnoe, maybe it's an option to convert the Pydantic model to a Box tree (see this project <https://github.com/cdgriffith/Box>) after the initial setup? 

But in general I agree with you, my use case is a slight modification of data send to an API endpoint. What I currently do is create a second version of my whole model that inherits from the base model and only includes the additional fields. Then I use `construct()` to create the updated model. The biggest disadvantage with this is that `construct()` does not handle nesting that well

misuzu:
I just got bitten by this issue, I'm using pydantic model to describe aggregate result from mongodb:

```python
class CollectionAggregateId(BaseModel):
    field1: str
    field2: str
    field3: str


class CollectionAggregateModel(BaseModel):
    _id: CollectionAggregateId

    count: int
```

```python
AttributeError: 'CollectionAggregateModel' object has no attribute '_id'
```

Using alias is not an option for me because I have data with both `_id` and `id` fields in my database.

Issue #800, opened 2019-09-09
[Discussion] Recoverable fields


DerRidda:
# Feature Request

I want to discuss the idea of potentially having a recoverable type or recovery mechanic for invalid fields in pydantic.

The specific use case is, that we have data with a ton of optional fields that may be provided to us with invalid data. We want to detect and report about these invalid fields but without discarding everything else that is valid.

I have tried several approaches of bolting this kind of logic on top of the library as purely a user of it but ended up with hacky and incomplete solutions that at a bare minimum always lose the location information ("loc") field of the invalid data.

Any solution to be truly viable for us needs to still include the loc field and it needs to be collectable in a structured fashion, e.g. a method to retrieve these errors on the parsed object or via warnings (though capturing warnings is a thread-safety issue).

I wasn't sure if that was really in line with the idea of pydantic but since I read that pydantic is actually a parsing library instead of validation I would say, that this is aligns with that idea.

samuelcolvin:
I'm pro this.

I think some kinds of error (eg. `const` maybe?) have `given_value` in the context on each error.

I would suggest we add `given_value` (perhaps better name if possible) to `ctx` of all errors.

Would this suit your purpose?

DerRidda:
I don"t think it would serve my person and I do believe I didn't quite make my intention clear.
Let me just draft a pseudo-code example of what I imagine:

```python
from typing import List
from pydantic import BaseModel, recoverable

class NestedModel(BaseModel):
    int_value: recoverable(int, recover_value=None)
    # recover_value would be the value that is applied when the field is invalid


class Model(BaseModel):
    float_value: float
    nested: NestedModel


data = {"float_value": 13.56 ,"nested": { "int_value": "not an int"}}

m = Model(**data)  # Does not raise a ValidationError

assert m.float_value == 13.56 # passes
assert m.nested.int_value == None # also passes

warnings = m.warnings() # This would contain list or similar of ValidationError that wasn't thrown but
                                       # raised by recoverable values

assert w[0].loc == ("nested", "int_value",) # location information is still resent

`¨

I hope I get the idea across. This of course is by no means the way this functionality would necessary look like.

samuelcolvin:
I see what you mean, makes more sense.

Personally I think the warning/recoverable think is too esoteric.

How about a `Config` setting for keeping a reference to the original data:
* that original data would be available in the sub-model, so you could do something like `m.nested.original['int_value']`
[...]

PR #1144, opened 2020-01-05
[DRAFT] Add loc to validatiors and refactor so that the true nested loc is passed down validation hierarchy.

DerRidda:
## Change Summary

This is a draft for adding the loc to validators as a parameter.
That change alone was fairly straight forward but made me realize that that it wasn't enough
to actually achieve the goal of knowing the nested loc before flattening errors.

The issue is that nested modles get instantiated via the BaseModel __init__ method which accepts no loc, as essentially all parameters there are considered fields for validation.
My solution was to create a new classmethod for construction that simply accepts the data and loc information which is then used by both the __init__ method and the validator to construct the model.

This PR should be seen as a basis for discussion if this is even something @samuelcolvin is willing to pursue further. If the answer is yes the complete impact needs to be assessed and further TODOs defined to make this a clean and robust change without any breaking changes and minimal to no performance impact.

So far it is incomplete, not optimized and tests have not been updated.

## Related issue number

#800 

## Checklist


pydantic/main.py
```diff
@@ -266,14 +267,21 @@ class BaseModel(metaclass=ModelMetaclass):
 
     def __init__(__pydantic_self__, **data: Any) -> None:
         # Uses something other than `self` the first arg to allow "self" as a settable attribute
+        __pydantic_self__.__class__.__from_dict__(data, __pydantic_self__)
```
```suggestion
        __pydantic_self__.__from_dict__(data, __pydantic_self__)
```

if we do it this way.

pydantic/main.py
```diff
@@ -266,14 +267,21 @@ class BaseModel(metaclass=ModelMetaclass):
...
+
+    @classmethod
+    def __from_dict__(cls, data: Dict[str, Any], __pydantic_self__ = None, loc: 'LocStr' = None):
```
samuelcolvin: I'm not sure about this, we'll need to make sure it has no performance impact.

Also I'm sure we already have a way of fixing kinks in `loc` even in sub-models. Are you sure this method is required?

I guess something like it would be required to make `loc` right in validators.

Does implementing `loc` this way allow us to simplify the type of `ErrorList` and maybe `flatten_errors`? If so it might be worth it.

pydantic/main.py
```diff
@@ -799,11 +807,16 @@ def validate_model(  # noqa: C901 (ignore complexity)
...
+        loc = tuple()
+    elif not isinstance(loc, tuple):
+        loc = (loc, )
+
     for validator in model.__pre_root_validators__:
         try:
             input_data = validator(cls_, input_data)
         except (ValueError, TypeError, AssertionError) as exc:
-            return {}, set(), ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls_)
+            return {}, set(), ValidationError([ErrorWrapper(exc, loc=loc + (ROOT_KEY,))], cls_)
```
samuelcolvin: it might be more performant to add a method

`def _build_loc(current_loc, new_key):
    if current_loc:
        return current_loc + (new_key,)
    else:
        return new_key  # or (new_key,)
```

DerRidda: Yeah, there is still some optimization to be done in the hot path here.

What I wanted to know about root validator locs is: Should the root loc of nested models actually contain the ROOT_KEY or should they be claled simply by the name of the field that nested model is in? In the case highlighted here it would mean simply passing in the loc when not none or empty and only passing the ROOT:KEY otherwise.

samuelcolvin: It's important that loc as shown in exceptions should not change from what it is now.

I guess the point in this change is to allow that same loc value that would appear in the exception to be available to the validator.

pydantic/main.py
DerRidda: I had an idea about how to avoid this method when I remembered that you can actually call __init__ statically as well and pass the self reference. That could at least clean the code up, no statement about performance yet though. This code is not very optimized yet.

I don't know how you would fix the loc inside the validation chain as the call to validate submodels is simply a call to ModelClass(**dict) which drops the loc chain then and there.

The reason why the flatten_errors function can fix the loc is because the entire hierarchy is validated and errors exist for each nested model while the root model is passed as well. From there all the information is derived.

A validator has inherently no awareness of where the field it is validating is nested and can't derive that from the exception hierarchy as that doesn't get passed into the validator and even if it was, would be incomplete as validation is still ongoing.

As far as ErrorList is concerned, the flattening would at the very least no longer have to derive the loc at all. So that step can be removed.
It might also be possible to translate ErrorList into a simple 1D list as nesting information is no longer needed with the loc being correct already. My concern here is ordering, so that errors appear in an order that makes sense to the human reading it.

I will see what is possible here. There could be some serious performance gains on the exception path here saving on nested looping in the cases where library users need to call the errors() method (which frankly should be the majority of cases, as you need that to log errors or inform users about what they did wrong).

samuelcolvin: agreed, good luck.

DerRidda:
I slightly cleaned up this approach with the `__init__` method still being the way the BaseModel is instantiated and `__from_dict__` now simply calling it statically.

I think this has a considerably smaller 'wtf' factor

samuelcolvin:
I think you've messed up the rebase, also tests are still failing. Could you fix? 

(Might easiest to create a new PR)

DerRidda:
Sorry for not getting back to you sooner, yeah I will create a new PR with a cleaner version of this.

PR #246, opened 2018-08-19
add ``UrlStr`` and ``urlstr`` types #236

Gr1N:
fix #236

Hi, 

I've ported URL validation from marshmallow library, to my mind it's all we need. @samuelcolvin what do you think?

PR #1148, opened 2020-01-06
correct url_regex

### LINKED ISSUE #1142: url_regex doesn't correctly match all allowed characters in a URI
# Bug

```python
from pydantic.networks import url_regex
m=url_regex.match("postgresql+psycopg2://postgres:postgres@localhost:5432/hatch")
parts = m.groupdict()
parts
```
produces the following incorrect matching:

```python
{'scheme': None,
 'user': 'postgresql+psycopg2',
 'password': '//postgres:postgres',
 'ipv4': None,
 'ipv6': None,
 'domain': 'localhost',
 'port': '5432',
 'path': '/hatch',
 'query': None,
 'fragment': None}
```

This is with Python 3.7.5 on OSX, pydantic version 1.2

samuelcolvin:
## Related issue number

fix #1142

## Checklist

* [ ] ~~Documentation reflects the changes where applicable~~


jonathanunderwood:
Yup, this works. Many thanks for getting to this so quickly.

For my own curiosity - what was behind the decision to manually parse with a regex rather than using urllib?

dmontagu:
@jonathanunderwood I believe the reason was to make it possible to make it easier to call out specific parts in the case of an invalid input. But @samuelcolvin would know better.

samuelcolvin:
> For my own curiosity - what was behind the decision to manually parse with a regex rather than using urllib?

Originally @Gr1N copied the URL parsing logic from marshmallow in #246 which used a regex.

That worked fine for a bit, then people complained (understandably) that there was no hint about what was wrong with an invalid URL. I looked into using urllib but I couldn't get it to parse correctly and/or provide useful errors, hence rewriting with a more sophisticated regex.

But looking into this, I've just noticed that django [uses](https://github.com/django/django/blob/07f6ff09e431cec48be5e46d843a4bbf10bfeac6/django/forms/fields.py#L676-L705) `urlsplit`. If we could get the same interface using `urllib.urlsplit` or similar without the massive regex, I'd be keen to switch over. That said, since we completely rewrote the URL logic for v1, I'm not that keen to change it again significantly even in a v2 release.

The other advantage of the current approach, is flexibility. If we find an issue like this (or something more subtle like we want to follow convention rather than the RFC for international domains or something), a regex gives us that flexibility. Using `urllib` is a bit "all or nothing" I guess.

PR #1154, opened 2020-01-07
Bug Fix: field shape value is wrong for native sequence types #1112

### LINKED ISSUE #1112: field shape wrong for native sequence types
Please complete:
* Pydantic version `import pydantic; print(pydantic.VERSION)`: **master**

While working on #1111 I noticed that `field.shape` is not set correctly for native sequence types:

```py


class Model(BaseModel):
    a: list
    b: set
    c: frozenset
    d: dict
    e: tuple


for name, field in Model.__fields__.items():
    debug(name, field, field.shape)

"""
test.py:13 <module>
    name: 'a' (str) len=1
    field: ModelField(
        name='a',
        type=list,
        required=True,
    ) (ModelField)
    field.shape: 1 (int)
test.py:13 <module>
    name: 'b' (str) len=1
    field: ModelField(
        name='b',
        type=set,
        required=True,
    ) (ModelField)
    field.shape: 1 (int)
test.py:13 <module>
    name: 'c' (str) len=1
    field: ModelField(
        name='c',
        type=frozenset,
        required=True,
    ) (ModelField)
    field.shape: 1 (int)
test.py:13 <module>
    name: 'd' (str) len=1
    field: ModelField(
        name='d',
        type=dict,
        required=True,
    ) (ModelField)
    field.shape: 1 (int)
test.py:13 <module>
    name: 'e' (str) len=1
    field: ModelField(
        name='e',
        type=tuple,
        required=True,
    ) (ModelField)
    field.shape: 1 (int)
"""
```

Currently the only known consequence of this is that the schema will be wrong, I guess `per_item` also won't work.

Should be possible to fix without breaking other functionality.

kousikmitra:
## Change Summary
In `pydantic/fields.py:ModelField._type_analysis()` where the `field.shape`  field get assign depending on the `field` type. `getattr(self.type_, '__origin__', None)` is returning `None` so the function is getting returned from there without assigning the value to `field.shape` variable.

https://github.com/samuelcolvin/pydantic/blob/e65d11240a6febcd569ad44a61e2374eb00b9ba6/pydantic/fields.py#L355-L358

So I added a check which checks if the `origin` is `None` then it checks if the `self.type_` is a native sequence type and based on that it set the value of `self.shape`.

## Related issue number

Fix #1112 

## Checklist

* [x] Added Check for Checking if `type_` is a native sequence and set `field` value based on that.

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -361,7 +361,7 @@ class ModelField(Representation):
         origin = getattr(self.type_, '__origin__', None)
         if origin is None:
             # field is not "typing" object eg. Union, Dict, List etc.
-            self.shape = TYPES_TO_SHAPE.get(self.type_)
+            self.shape = TYPES_TO_SHAPE.get(self.type_, self.shape)
             return
         if origin is Callable:
             return
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -181,6 +181,12 @@ SHAPE_NAME_LOOKUP = {
     SHAPE_SEQUENCE: 'Sequence[{}]',
     SHAPE_FROZENSET: 'FrozenSet[{}]',
 }
+TYPES_TO_SHAPE = {
+    tuple: SHAPE_TUPLE,
+    list: SHAPE_LIST,
+    set: SHAPE_SET,
+    frozenset: SHAPE_FROZENSET,
+}
 
 
 class ModelField(Representation):
@@ -355,18 +361,7 @@ class ModelField(Representation):
         origin = getattr(self.type_, '__origin__', None)
         if origin is None:
             # field is not "typing" object eg. Union, Dict, List etc.
-            if self.type_ is tuple:
-                self.shape = SHAPE_TUPLE
-                return
-            if self.type_ is list:
-                self.shape = SHAPE_LIST
-                return
-            if self.type_ is set:
-                self.shape = SHAPE_SET
-                return
-            if self.type_ is frozenset:
-                self.shape = SHAPE_FROZENSET
-                return
+            self.shape = TYPES_TO_SHAPE.get(self.type_)
             return
         if origin is Callable:
             return
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -355,6 +355,18 @@ class ModelField(Representation):
         origin = getattr(self.type_, '__origin__', None)
         if origin is None:
             # field is not "typing" object eg. Union, Dict, List etc.
+            if self.type_ is tuple:
+                self.shape = SHAPE_TUPLE
+                return
+            if self.type_ is list:
+                self.shape = SHAPE_LIST
+                return
+            if self.type_ is set:
+                self.shape = SHAPE_SET
+                return
+            if self.type_ is frozenset:
+                self.shape = SHAPE_FROZENSET
+                return
             return
         if origin is Callable:
             return


pydantic/fields.py
```diff
@@ -355,6 +355,18 @@ def _type_analysis(self) -> None:  # noqa: C901 (ignore complexity)
...
+                return
+            if self.type_ is list:
+                self.shape = SHAPE_LIST
+                return
+            if self.type_ is set:
+                self.shape = SHAPE_SET
+                return
+            if self.type_ is frozenset:
+                self.shape = SHAPE_FROZENSET
+                return
```
dmontagu: 1. I think this would be cleaner using `elif` and only a single return (the one that was already there), closer to:
```
if self.type_ is x:
    self.shape = SHAPE_X
elif ...
elif ...
...
return
```

2. @samuelcolvin Should we test for subclasses? (@kousikmitra in case this change is desired/necessary, it should be with `if lenient_issubclass(self.type_, x)` (need to use `lenient_issubclass`, which is a pydantic util function, in order to prevent certain errors).

pmav99: Maybe even
``` py
TYPES_TO_SHAPES = {
    tuple: SHAPE_TUPLE,
    list: SHAPE_LIST,
    set: SHAPE_SET,
    frozenset: SHAPE_FROZENSET,
}
# ...
if origin is None:
    self.shape = TYPES_TO_SHAPE[self.type_]
```

dmontagu: Well, should probably use `TYPES_TO_SHAPE.get(self.type_)` and check whether it is `None` in case the value isn't one of those types. But yes, that would be cleaner.

kousikmitra: Yes, It's looking much better. I am making the changes.

kousikmitra: I have made the changes. Do I need to update the test cases as well?

dmontagu: @kousikmitra do you know why this is failing the checks? Are tests and linting etc passing for you locally? (If you run `make` locally, does it succeed? See https://pydantic-docs.helpmanual.io/contributing/ for more detail about how to run the various CI tests locally if you haven't done that yet.)

dmontagu: If there are  any failing tests then yes, you should update them. Also, it would be great if you could add a basic test demonstrating that this actually fixes #1112 (probably need that before it gets merged).

If you don't have time or aren't sure how to do that, I can take care of it at some point soon, just let me know.

kousikmitra: I have ran `make` locally, There is almost 20 test cases which are failing but other checks like code format, linting etc are passing.  I will look into the test cases and revert you back if I need any help.

kousikmitra: Hi @dmontagu! I think there is still a thing needs to be addressed. There is also a member `sub_fields` which is getting initialized in case of `typing` module objects. But in case of standard sequence type it's not getting initialized here. Can you just check that?

samuelcolvin:
@dmontagu you assigned yourself to this. Any chance you could take a look? No problem if not or you're not available right now.

samuelcolvin:
Sorry this has taken so long. I don't think @dmontagu has time to review this at the moment.

I think it's best we close this and I have another go at solving this during the v2 release process when we can afford to break old functionality slightly more.

Thank you for working on this and sorry to close the PR.

PR #1178, opened 2020-01-19
change alias priority logic

### LINKED ISSUE #1177: alias_generator used instead of field's alias on inheritance
# Bug

When `alias` is specified on a field for a class where `alias_generator` is also defined, `alias` is respected for that class but not any derived classes.

```python
             pydantic version: 1.3
            pydantic compiled: False
                 install path: /<snip>/site-packages/pydantic
               python version: 3.7.6 (default, Dec 30 2019, 19:38:28)  [Clang 11.0.0 (clang-1100.0.33.16)]
```

## Example code

```python


class MyBaseModel(BaseModel):
    class Config:
        alias_generator = lambda x: x.upper()

class MyModel(MyBaseModel):
    my_attribute: bool = Field(..., alias='myATTRIBUTE')

class MyOtherModel(MyModel):
    pass
```

## Examples

```py
>>> MyModel()
  File "<snip>/site-packages/pydantic/main.py", line 274, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for MyModel
myATTRIBUTE
>>> MyOtherModel()
  File "<snip>/site-packages/pydantic/main.py", line 274, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for MyOtherModel
MY_ATTRIBUTE
```

Notice that `MyModel` requires `myATTRIBUTE`, as expected, but `MyOtherModel` requires `MY_ATTRIBUTE`, which is from the `alias_generator`

samuelcolvin:
## Change Summary

This changes the priority of aliases to fix #1177, however it introduces a potential breaking change.

For example:

https://github.com/samuelcolvin/pydantic/blob/80152f337191ba7539e077c02a236bb713fdddd2/tests/test_aliases.py#L189-L204

before `alias_generator` on the child would take taken priority. While it's arguable whether this is correct and we could allow `alias_generator` to return a tuple `(<alias>, <alias priority>)` to allow customisation, I'm aware this is a breaking change.

However I'm not sure how else to fix the issue in a cleaner way. @tiangolo, @dmontagu, anyone else involved - do you think we should:
* merge this with a "potential breaking change" warning
* wait to merge this until v2
* provide some other fix for #1177 

?

## Checklist


dmontagu:
I personally view this as a bugfix -- I think manually labeled fields should *always* retain their manually labeled alias unless specifically overridden. It doesn't really surprise me that this hasn't caused issues because I think it is rare that the two are used together. 

I am sensitive to the idea that it is a breaking change, but in my opinion this seems pretty clearly like incorrect behavior instead of a change from valid implementation to another.

----

It would be different if a lot of people were now dependent on the incorrect logic. That would very much surprise me, but I'm not sure how to confirm it's not the case.

Do you have a sense of how hard it would be to raise a model-creation-time warning in the case where this PR would actually result in a different (i.e., incompatible) alias being used? The warning message could say something like `"your model may be dependeing on buggy behavior that will be fixed in an upcoming minor version"`. That way, if anyone is depending on this behavior, they will have the chance to let us know they would prefer we didn't change anything.

pydantic/main.py
```diff
@@ -73,14 +73,22 @@ class BaseConfig:
...
     @classmethod
     def get_field_info(cls, name: str) -> Dict[str, Any]:
-        field_info = cls.fields.get(name) or {}
-        if isinstance(field_info, str):
-            field_info = {'alias': field_info}
-        elif cls.alias_generator and 'alias' not in field_info:
+        fields_value = cls.fields.get(name)
+
+        if isinstance(fields_value, str):
+            field_info: Dict[str, Any] = {'alias': fields_value, 'alias_priority': 0}
```
```suggestion
            field_info: Dict[str, Any] = {'alias': fields_value}
```

pydantic/main.py
```diff
@@ -73,14 +73,22 @@ class BaseConfig:
...
+        elif isinstance(fields_value, dict):
+            field_info = fields_value
+        else:
+            field_info = {}
+
+        if 'alias' in field_info:
+            field_info['alias_priority'] = 2
```
```suggestion
            field_info.set_default('alias_priority', 2)
```

samuelcolvin:
@dmontagu very interesting.

I think in theory the warning approach would be possible, we would change this PR to raise the warning on the new behaviour but not change the alias. I guess we'd then either need to tell people "wait for v1.5 when the new behaviour will come in" or "add the `use_new_alias_behaviour` flag to your alias definitions".

This all sounds like a lot of work and complexity for something which:
* might well not be used by many people
* might be what people are expecting is happening anyway.

I think if we give an appropriate warning in the release notes and are explicit about what's changed, people can't be that upset by the change.

aljets:
That seems ok. I could see a pretty reasonable argument it should work that way. I could also see a pretty reasonable argument that children should always override the parent. But following specificity, as you are, seems reasonable to me. So long as it is clearly documented.

My 2 cents: I'm not sure how I feel about adding an `alias_priority` (in your latest commits), however, as that seems like a decent complication that could quickly get messy, so it may be worth thinking through use cases and alternatives--it may be cleaner for consumers to work around priority in other ways.

Thanks for working on this.

samuelcolvin:
> So long as it is clearly documented.

Done I hope, feedback welcome on the docs.

> My 2 cents: I'm not sure how I feel about adding an alias_priority

Me neither, I haven't documented it and I don't think people should ever use it themselves

> so it may be worth thinking through use cases and alternatives

again I completely agree, but that will have to wait for v2, see #1186

docs/usage/model_config.md
```diff
@@ -103,16 +103,33 @@ _(This script is complete, it should run "as is")_
...
+it should be trivial to modify the `to_camel` function above.
 
 ## Alias Precedence
 
-Aliases defined on the `Config` class of child models will take priority over any aliases defined on `Config` of a
-parent model:
+!!! warning
+    Alias priority logic changed in **v1.4** to resolve buggy and unexpected behaviour in previous versions.
+    In some circumstances this may represent a **breaking change**,
+    see [#1178](https://github.com/samuelcolvin/pydantic/issues/1177) and the precedence order below for details.
```
aljets: Did you intend to indicate issue `1178` but link to `1177`? Pretty minor since they are related though.

docs/usage/model_config.md
```diff
@@ -103,16 +103,33 @@ _(This script is complete, it should run "as is")_
...
+
+A field alias's is chosen based on the following
```
aljets: I think you meant `A field's alias` (or maybe this line is unnecessary/duplicated below)

aljets:
Sounds great. Documentation looks good. Thanks!

mat02:
Hi everyone!
Is there any option to increase priority of alias_generator for particular model?
I have a situation, where I am extending old API of some legacy software with some additional functionality under a new, sort of "unified" API.

I have a base model with aliases that match legacy API. I also have extended models that inherit from that base models and these will be served for client application. Now the problem is that I would like to use alias_generator in that extended models to replace snake case field names with camel case. Obviously, just as explained in the docs, this works only for the fields defined in child model.
Here's the simplified code example:
`
def to_camel(string: str) -> str:
    return ''.join(word.capitalize() for word in string.split('_'))

class OldAPIModel(BaseModel):
    id: int = Field(..., alias='Identyfikator')
    name: str = Field(..., alias='Nazwa')
    description: str = Field(..., alias='DlugiOpis')
    
class NewAPIModel(OldAPIModel):
    product_keywords: Optional[List[str]] = Field([])
    class Config:
        alias_generator = to_camel
`

Now, the use case for that is as following:
I get data from legacy api using OldAPIModel, then gather addidional data from other database and pass all of that to the client using NewAPIModel. As I see it, it would be nice to set some flag (that is disabled by default) to force use of alias_generator in child models. Of course I can just copy all fields from OldAPIModel to NewAPIModel, but I would like to avoid code duplication.
Or maybe there is some other way/usage pattern that I should follow in this case?
BR,
Mateusz

samuelcolvin:
no way I know of to force use of a different alias.

You have the options of:
* copy them like you said
* generate the model yourself using `create_model()`, with `Field()` and alias's set on the field
* look through the pydantic code and find some way to hack it, no idea if this is possible, might break in minor version updates

ainsleymcgrath:
> Hi everyone!
> Is there any option to increase priority of alias_generator for particular model?
> I have a situation, where I am extending old API of some legacy software with some additional functionality under a new, sort of "unified" API.
> 
> I have a base model with aliases that match legacy API. I also have extended models that inherit from that base models and these will be served for client application. Now the problem is that I would like to use alias_generator in that extended models to replace snake case field names with camel case. Obviously, just as explained in the docs, this works only for the fields defined in child model.
> Here's the simplified code example:
> 
> ```
> def to_camel(string: str) -> str:
>     return ''.join(word.capitalize() for word in string.split('_'))
> 
> class OldAPIModel(BaseModel):
>     id: int = Field(..., alias='Identyfikator')
>     name: str = Field(..., alias='Nazwa')
>     description: str = Field(..., alias='DlugiOpis')
>     
> class NewAPIModel(OldAPIModel):
>     product_keywords: Optional[List[str]] = Field([])
>     class Config:
>         alias_generator = to_camel
> ```
> 
> Now, the use case for that is as following:
> I get data from legacy api using OldAPIModel, then gather addidional data from other database and pass all of that to the client using NewAPIModel. As I see it, it would be nice to set some flag (that is disabled by default) to force use of alias_generator in child models. Of course I can just copy all fields from OldAPIModel to NewAPIModel, but I would like to avoid code duplication.
> Or maybe there is some other way/usage pattern that I should follow in this case?
> BR,
> Mateusz

@mat02, if you're still looking for a way to do this, I figured out a reasonable solution using `create_model`.
If you already figured it out, then this can just hang out here for posterity 😸

```python
from itertools import chain

class Bread(BaseModel):
    class Config:
        alias_generator = lambda s: f"bread_{s}"

    name: str
    id: int


class Butter(BaseModel):
    class Config:
        alias_generator = lambda s: f"butter_{s}"

    name: str
    id: int


print(Butter.__fields__)
print(Bread.__fields__)


class BreadAndButterDoesntWork(Bread, Butter):
    """This won't work :("""


print(BreadAndButterDoesntWork.__fields__)


def BreadAndButterWorks():
    return create_model(
        "BreadAndButter",
        **{
            f.alias: (f.type_, ...)
            for f in chain.from_iterable(
                getattr(m, "__fields__").values() for m in [Bread, Butter]
            )
        },
    )


print(BreadAndButterWorks().__fields__)


print(BreadAndButterWorks().__fields__)
```

Hopefully this helps someone!

Fwiw, I was dealing with a use case where we are using Pydantic models to get bits of data from various APIs and databases in order to ultimately stitch them together into one big data warehouse table. 

Imagine in the above, that `Bread` and `Butter` were each responses from separate apis The idea was to compose all the models together in order to form that one big table with minimal repetition and a lot of consistency. Since each of the "parts" had a `source_id` and a `natural_key`, disambiguation with `alias_generator` made a ton of sense. What a surprise when the big rolled up model lost all those aliases!

My experience is limited with this library, and my problem is somewhat unusual, but while I'm here I'll toss 2 cents in the bucket for raising the priority of `alias_generator` in the future. :)

PR #1179, opened 2020-01-19
Validation Decorator

### LINKED ISSUE #347: type checking decorator
## Dream Feature:

Using pydantic to check types without building Models or dataclasses.

Four modes of usage:

### 1. decorators

```py
@typecheck
def foobar(x: int, y: float, x: List[Dict[str, Tuple[int, int]]]):
   ...
```
The `typecheck` decorator would check types when `foobar` is called, optional arguments for the decorator to:
* check x percent of the time
* what to do on an error: exception, warning, logging etc.
* whether to coerce types or just check them

(Usage: at library boundaries to guarentee correct usage)

### 2 context manager

A context manager which uses `sys.settrace` to check types on all function calls inside your code base.

This won't be at all easy and will slow down code execution a lot, but would be a big win. (I think?)

Would need to with with existing trace functions eg. coverage and allow that to still work properly.

Also the arguments passed would need to be checked but not modified so behavior isn't changed.

(usage: see below)

### 3 pytest extension

which uses the context manager above to do runtime checks during testing.

### 4 custom python "executable"

use `runpy.run_path` to run a python program inside the context manager above and therefore with type checks.

(probably most useful as a sanity check or or adhock checks)

-----------------------------------------

Is this something anyone would fine useful?

The `settrace`/pytest extension approach might end up being infeasible, which would just leave the decorator.

samuelcolvin:
## Change Summary

This adds a `validate_arguments` function decorator which checks the arguments to a function matches type annotations.

I'm intending to get an MVP ready and merge that with a note saying this is in open beta and may change.

Still, there's quite a lot here. Feedback very welcome.

## Related issue number

fix #347

## TODO

* [x] proper error on position only arguments
* [x] better error when extra `kwargs`, or `args` are provided.
* [x] Unit tests complete
* [x] evaluate string annotations, or raise an error?
* [x] check async functions work
* [x] agree on the name, is `validate_arguments` the best option? unless someone has a better idea???
* [x] add documentation
* [x] `changes/<pull request or issue id>-<github username>.md` file

To be done in future (I think this should be done after the initial release of this feature): see #1205


dmontagu:
This looks great to me!

samuelcolvin:
For me this is ready to merge.

I'll leave it a few days for any further feedback, then merge.

tiangolo:
I hadn't seen this. It looks awesome!

StephenBrown2:
Just noticed that the decorator is called `validate_arguments`, but in the docs it's referred to as `validate_assignment`.

samuelcolvin:
@tiangolo thanks!

@StephenBrown2 thanks for pointing that out 🤦‍♂ PR welcome to fix that.

samuelcolvin:
don't worry, I'll fix the docs.

mpkocher:
Are there any concerns about memory issues with this? 

Currently, every function that is wrapped via `validate_arguments` has an instance of `ValidatedFunction` as well as instance of a `BaseModel` bolted on to it. 

```
In [31]: @validate_arguments
    ...: def f2(a: int, b: int) -> int:
    ...:     return a + b
    ...: 

In [32]: f2
Out[32]: <function __main__.f2(a: int, b: int) -> int>

In [33]: f2.__dict__
Out[33]: 
{'__wrapped__': <function __main__.f2(a: int, b: int) -> int>,
 'vd': <pydantic.decorator.ValidatedFunction at 0x7fd0586e64f0>,
 'raw_function': <function __main__.f2(a: int, b: int) -> int>,
 'model': F2}

In [34]: f2.model.__fields__
Out[34]: 
{'a': ModelField(name='a', type=int, required=True),
 'b': ModelField(name='b', type=int, required=True),
 'args': ModelField(name='args', type=Optional[List[Any]], required=False, default=None),
 'kwargs': ModelField(name='kwargs', type=Optional[Mapping[Any, Any]], required=False, default=None)}

```

stonecharioteer:
This feature seems to allow values that can be cast from one form to another.
```python

@validate_arguments
def something(x: str, y: Dict[str, str]):
    print(x, y)
```

Testing this with the following: `something(10, {"x": 20})` doesn't give a ValidationError,
but testing it with `something("x", {"x":None})` does. 

Is that by design? Is there a way to enable _stricter_ type checking?

PR #1183, opened 2020-01-21
Support unicode and punycode when validating TLDs

### LINKED ISSUE #1182: Support unicode and punycode when validating TLDs
# Feature Request

```
pydantic version: 1.3
            pydantic compiled: True
               python version: 3.8.1 (default, Jan  3 2020, 22:55:55)  [GCC 8.3.0]
                     platform: Linux-4.9.184-linuxkit-x86_64-with-glibc2.2.5
```

The IANA has approved many TLDs that are not matched by the [TLD domain ending regex](https://github.com/samuelcolvin/pydantic/blob/07271ca5bc89aff88957274d2de1a06b28dd070c/pydantic/networks.py#L49) used for `HttpUrl` validation. There are currently ~152 such TLDs: see the entries in the [authoritative list of TLDs](http://data.iana.org/TLD/tlds-alpha-by-domain.txt) containing the ASCII Compatible Encoding prefix `xn--`.

One approach to adding compatibility for such TLDs would be to modify the domain ending regex pattern to allow for Unicode characters, as well as the corresponding internationalized ASCII strings. For example:

 ```python
_domain_ending = r"(?P<tld>(\.[^\W\d_]{2,63})|(\.(?:xn--)[_0-9a-z-]{2,63}))?\.?"
```

Such a change would allow for the following to run successfully

```python
from pydantic import BaseModel, HttpUrl, ValidationError 


class Domain(BaseModel):
    domain: HttpUrl
        

ascii_domains = ["https://example.com"]

idna_domains = [
    "https://example.xn--p1ai",
    "https://example.xn--vermgensberatung-pwb",
    "https://example.xn--zfr164b",
]

unicode_domains = [str.encode(domain).decode("idna") for domain in idna_domains]
        
valid_domains = ascii_domains + idna_domains + unicode_domains

invalid_domains = ["https://example.123", "https://example.ab34"]

for domain in valid_domains:
    Domain(domain=domain)
    
for invalid_domain in invalid_domains:
    try:
        Domain(domain=invalid_domain)
    except ValidationError:
        pass
```

Would you accept a PR for this?

jamescurtin:
## Change Summary

Add support for unicode and punycode in TLDs

## Related issue number

Closes #1182

## Checklist


pydantic/networks.py
```diff
@@ -46,7 +46,7 @@
 _url_regex_cache = None
 _ascii_domain_regex_cache = None
 _int_domain_regex_cache = None
-_domain_ending = r'(?P<tld>\.[a-z]{2,63})?\.?'
+_domain_ending = r'(?P<tld>(\.[^\W\d_]{2,63})|(\.(?:xn--)[_0-9a-z-]{2,63}))?\.?'
```
samuelcolvin: This needs to go into a function so it's only compiled when it's needed. Same as `url_regex`.

pydantic/networks.py
```diff
@@ -227,11 +227,16 @@ def validate_host(cls, parts: Dict[str, str]) -> Tuple[str, Optional[str], str,
                     raise errors.UrlHostError()
                 host_type = 'int_domain'
                 rebuild = True
-                host = host.encode('idna').decode('ascii')
+            host = host.encode('idna').decode('ascii')
```
samuelcolvin: Why is this change required?

pydantic/networks.py
jamescurtin: When `_domain_ending` is defined, it is a raw string; regex compilation only happens within the functions `ascii_domain_regex` and `int_domain_regex`. I think this means the goal of only calling `re.compile` when necessary is already satisfied, without requiring a change.

Is there an additional performance concern? If so, happy to pull this into a function that `ascii_domain_regex` and `int_domain_regex` call.

pydantic/networks.py
jamescurtin: The [enclosing block](https://github.com/samuelcolvin/pydantic/blob/5cbf7627c3dbe5fcb26b8f1ae734b418ed413c83/pydantic/networks.py#L224) is only entered if the `host` component of the URL contains a non-ASCII character. If non-ASCII characters are only contained in the TLD (_e.g._ `https://example.рф`), this block is not entered. 

Dedenting ensures the full URL string, including the TLD, are escaped in the above case. The tests added on lines 268-270 of `tests/test_networks.py` are meant to address this change.

pydantic/networks.py
samuelcolvin: sorry I was being dumb, reviewing on my phone and not looking properly. 

all good :+1:

pydantic/networks.py
samuelcolvin: Makes sense, but surely better to have:
* `_ascii_domain_ending` - used in `ascii_domain_regex`
* `_int_domain_ending` - used in `int_domain_regex`

Then the below change to the tld is only required in the "international" case?

jamescurtin: Good point; I can split the regex up in two to make the use-cases clearer. With that said, I don't think that would be sufficient to solve the problem.

The `_ascii_domain_ending` (`(?P<tld>\.[a-z]{2,63})?`) can match against some internationalized URLs. For example, it matches `xn` in `https://example.xn--p1ai`, causing the block with international logic to be skipped.

One solution would be to update `_ascii_domain_ending` to explicitly not match non-ASCII TLDs, but I'd need to think about the right way to do this (I'm not sure how accurately we would be able to isolate the TLD from the remainder of the URL).

Another option would be to first execute both `ascii_domain_regex().fullmatch(host)` and `int_domain_regex().fullmatch(host)` and enter into the international logic block if the results of the two regexes differed.

Thoughts? I think that latter option may be easiest but would defer to you on the preferred approach.

samuelcolvin: Surely as long as `ascii_domain_regex` uses the (old) ascii tld regex, it will fail if either the domain or tld are non-ascii e.g. "international"?

jamescurtin: Correct--current state is that it would fail with a `UrlHostTldError`.  The issue here is that the original `ascii_domain_regex` pattern matches the host, but does not match the TLD. That is:

```python
_ascii_domain_ending = r'(?P<tld>\.[a-z]{2,63})?\.?'

...
host = 'https://example.xn--p1ai'

d = ascii_domain_regex().fullmatch(host)
assert d is not None

tld = d.group('tld')
assert tld is None
```

For the current international logic to correctly handle this example, `d` must be None (but it isn't). By making the change you suggest, validation of domains with an international TLD would no longer raise a `UrlHostTldError`, but the last three test cases for `test_coerse_url` would fail.

I'm pushing a change now that I think should address the concern: it keeps all internationalization logic in a single block while also ensuring the edge cases described above are handled correctly. (Also adds a few additional tests). Let me know what you think!

samuelcolvin:
LGTM

I think this is ready to merge?

jamescurtin:
Yes--ready to merge. Thanks for the review--this is a fantastic project!

samuelcolvin:
awesome, thank you.

Issue #1190, opened 2020-01-26
Broken loading list of tuples.


DMantis:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
pydantic version: 1.4
pydantic compiled: True
install path: /home/**removed**/venv/lib/python3.7/site-packages/pydantic
python version: 3.7.3 (default, Apr  3 2019, 19:16:38)  [GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
platform: Linux-4.19.86-041986-generic-x86_64-with-Ubuntu-18.04-bionic
     optional deps. installed: []
```

There is pretty strange behaviour on loading nested list of tuples. I firstly think that this might be intended, but then found out that parse_obj and parse_obj_as give different execution flow which  frustrates me.

```py
from pydantic import BaseModel

class OperationData(BaseModel):
    id: str

class Operation(BaseModel):
    __root__: Tuple[int, OperationData]

data = [0, {'id': '1.11.0'}]
# this one works as expected
print(Operation.parse_obj(data))
# printed: __root__=(0, OperationData(id='1.11.0'))

# However, this one doesn't
print(parse_obj_as(Operation, data))
# Traceback (most recent call last):
#  File "/home/**removed**/protocol/base.py", line 238, in <module>
#    print(parse_obj_as(Operation, data))
#  File "pydantic/tools.py", line 35, in pydantic.tools.parse_obj_as
#  File "pydantic/main.py", line 283, in pydantic.main.BaseModel.__init__
# pydantic.error_wrappers.ValidationError: 1 validation error for ParsingModel[Operation]
#__root__
# value is not a valid dict (type=type_error.dict)

# Which is not a big problem. The problem is that I have nested class

class OperationsBatch(BaseModel):
    batch_desc: str
    operations: List[Operation]

# and it produces same exception on
print(OperationsBatch.parse_obj({'batch_desc': '123', 'operations': [data, data]}))

# Traceback (most recent call last):
# File "/home/**removed**/protocol/base.py", line 243, in <module>
#    OperationsBatch.parse_obj({'batch_desc': '123', 'operations': [data, data]})
#  File "pydantic/main.py", line 402, in pydantic.main.BaseModel.parse_obj
#  File "pydantic/main.py", line 283, in pydantic.main.BaseModel.__init__
# pydantic.error_wrappers.ValidationError: 2 validation errors for OperationsBatch
# operations -> 0
[...]

PR #1200, opened 2020-01-31
Fix parse obj root

dmontagu:
## Change Summary

Makes nested root models work properly. In particular, the code raising errors in #1190 doesn't raise errors with this patch. I suspect this may resolve other issues related to custom root models as well.

Also required a new flag to make nested root models generate proper json.

I believe all of the newly added checks for the `__custom_root_type__` could be moved to model creation time by making use of some method-wrapping logic in the metaclass.

The fact that the `unpack_custom_root` argument is passed in the various recursive calls to `.dict` is the only performance thing that I think is likely hard to eliminate (without breaking the json generation). But it *would* be possible to move the actual *check* (and subsequent unwrapping) into a wrapper function so that it is only performed for types that actually have a custom root. I'm not sure if adding another argument itself adds an amount of performance overhead that is worth fretting over. (But I would also understand if you don't want to add the flag at all.)

## Related issue number

Addresses #1190 

## Checklist

### DIFF
--- a/tests/test_parse.py
+++ b/tests/test_parse.py
@@ -166,10 +166,10 @@ def test_parse_nested_custom_root_mapping():
 
 def test_parse_nested_custom_root_tuple():
     class NestedRootTuple(BaseModel):
-        __root__: Tuple[int, Optional["NestedRootTuple"]]
+        __root__: Tuple[int, Optional['NestedRootTuple']]
 
     NestedRootTuple.update_forward_refs()
 
     obj = NestedRootTuple.parse_obj(('1', ('2', ('3', None))))
     assert obj.dict() == {'__root__': (1, {'__root__': (2, {'__root__': (3, None)})})}
-    assert obj.json() == "[1, [2, [3, null]]]"
+    assert obj.json() == '[1, [2, [3, null]]]'
--- a/tests/test_parse.py
+++ b/tests/test_parse.py
@@ -1,6 +1,6 @@
 import json
 import pickle
-from typing import List, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/test_parse.py
```diff
@@ -151,3 +151,25 @@ class Model(BaseModel):
...
+    obj = NestedRootMapping.parse_obj({'1': {'2': {}}})
+    assert obj.dict() == {'__root__': {1: {'__root__': {2: {'__root__': {}}}}}}
+    assert obj.json() == '{"1": {"2": {}}}'
+
+
+def test_parse_nested_custom_root_tuple():
+    class NestedRootTuple(BaseModel):
+        __root__: Tuple[int, Optional["NestedRootTuple"]]
+
+    NestedRootTuple.update_forward_refs()
```
dmontagu: I discovered a strange bug here -- if you don't call this `update_forward_refs`, it doesn't raise an error, but then below you actually get:

```python
    obj = NestedRootTuple.parse_obj(('1', ('2', ('3', None))))
    assert obj.dict() == {'__root__': (1, {'__root__': ('2', {'__root__': ('3', None)})})}  # note 2 and 3 became strings
```

Might be worth looking into separately.

dmontagu: Yep, I confirmed this is not related to custom root; creating a separate issue.

samuelcolvin:
This seems like a backwards incompatible change to `__init__` logic? So I guess, if we make this change, we should wait for v2.

I'm also concerned about the madness of `dict(unpack_custom_root=True)` not returning a dict, but then again I'm conscious that `dict()` and `json()` should have the same (or as similar as possible) interfaces and `json()` need to return the root object not `{"__root__": thing}` in v2.

In summary, are we sure there isn't a similar way of fixing the original problem?

dmontagu:
I think the main problem is that there are some places in the code where object initialization happens via `Model(**data)` or similar. This generally makes it difficult to handle issues with a custom root model properly because validation *requires* a dict.

I'm not saying that a backwards-compatible fix isn't possible, but I think it would help a lot if there was a single model-validation logic path that involved passing a raw object, rather than making use of the `__init__` function and requiring a dict as input. It seems like some of these issues might be easier to address, for example, if parse_obj was the main way to create a validated instance, and `__init__` delegated to that function, rather than the other way around (as it is now).

But I haven't thought through things like performance implications etc. where the current approach may have marginal benefits for the most common use cases.

At any rate, I agree that this PR is likely unworkable in the near term, so I'm going to close it.

Issue #866, opened 2019-10-05
Simplify setting dynamic default values


ValentinCalomme:
# Feature Request

Please complete:
* OS: Windows
* Python version 3.7.4 (default, Aug  9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)]
* Pydantic version: 0.30

# Feature Request

Currently, if I want to have a dynamic default value, it has to be done the following way:

```py
from datetime import datetime

from pydantic import BaseModel, validator
from uuid import UUID, uuid4

class DemoModel(BaseModel):
    ts: datetime = None
    id: UUID = None

    @validator('id', pre=True, always=True)
    def set_id(cls, v):
        return v or uuid4()

    @validator('ts', pre=True, always=True)
    def set_ts_now(cls, v):
        return v or datetime.now()
```

This is fine for cases where actual validation has to be provided, however, in many cases like the one portrayed below, this leads to unnecessary code, especially if there are multiple dynamically set values. For instance, a timestamp and an id.

It would be great to be able to replace this usage pattern by a pydantic field called `Dynamic` for instance such that one can pass a callable as a default value. This would look like this:

```py
from datetime import datetime
from uuid import UUID, uuid4

from pydantic import BaseModel, validator, Dynamic

class DemoModel(BaseModel):
    ts: Dynamic[datetime] = datetime.now
    id: Dynamic[UUID] = uuid4
```

I don't expect that this would need to modify any existing functionalities. Probably just a need to create some sort of wrapper.

MrMrRobat:
I guess it can be done with some kind of wrapper

samuelcolvin:
I'm hesitant about pseudo types, without reading the docs it's not clear what this does.

What we could do is just is the function as the default/initial value? Then unless the type of the field was `Callable` it would call the function to generate a value.

This would not be completely backwards, but I think it would be more intuitive.

ValentinCalomme:
This would work as well!

dmontagu:
I think this might be better suited as a keyword argument to Schema (or Field now), or similar. That’s how this functionality is handled by both attrs and dataclasses.

samuelcolvin:
> I think this might be better suited as a keyword argument to Schema (or Field now), or similar. That’s how this functionality is handled by both attrs and dataclasses.

agreed.

samuelcolvin:
#1210 will implement `Field(default_factory=...)`.

For use of `ts: datetime = datetime.now` we should wait for v2.

[...]

Issue #155, opened 2018-04-12
Callable defaults


Gr1N:
Hi,

`pydantic` works fine with defaults defined using annotations in case of values like:
```
class Test(BaseModel):
    k: uuid.UUID = uuid.uuid4()
```

But in many validation libraries possible to define callable defaults, with `pydantic` it's also possible but not works as expected from my point of view:
```
class Test(BaseModel):
    k: uuid.UUID = uuid.uuid4

test = Test()
assert isinstance(test.k, uuid.UUID)
```
This code will fail because `test.k` is `<bound method uuid4 of <Test >>`.

Is it possible to fix this issue?

samuelcolvin:
yes, use a validator with `alwasy=True` and return `uuid.uuid4()`

Gr1N:
Read docs first :) Thanks

samuelcolvin:
No problem, is not that obvious in the docs

PR #1210, opened 2020-02-06
Dynamic default value

PrettyWood:
## Change Summary
Following #155 and #866, here is a proposal of implementation to simplify dynamic default values

Right now, to have a dynamic default value, we need to use `validators` like
```python
class DynamicValueModel(BaseModel):
    ts: datetime = None

    @validator('ts', pre=True, always=True)
        def set_ts_now(cls, v):
            return v or datetime.now()
```

Now, we can use `default_factory` (same name as [dataclasses](https://docs.python.org/3/library/dataclasses.html#dataclasses.field).to be consistent)
```python
class DynamicValueModel(BaseModel):
    ts: datetime = Field(default_factory=datetime.now)
```

## Related issue number
#155 and #866

## Checklist

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -1,5 +1,6 @@
 import warnings
 from collections.abc import Iterable as CollectionsIterable
+from copy import deepcopy
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -260,6 +261,16 @@ class ModelField(Representation):
         self.model_config.prepare_field(self)
         self.prepare()
 
+    def get_default(self) -> Any:
+        if self.default_factory is not None:
+            value = self.default_factory()
+        elif self.default is None:
+            # deepcopy is quite slow on None
+            value = None
+        else:
+            value = deepcopy(self.default)
+        return value
+
     @classmethod
     def infer(
         cls,
@@ -320,7 +331,7 @@ class ModelField(Representation):
         Note: this method is **not** idempotent (because _type_analysis is not idempotent),
         e.g. calling it it multiple times may modify the field and configure it incorrectly.
         """
-        default_value = self.default_factory() if self.default_factory is not None else self.default
+        default_value = self.get_default()
         if default_value is not None and self.type_ is None:
             self.type_ = type(default_value)
             self.outer_type_ = self.type_
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -296,7 +296,7 @@ class ModelField(Representation):
 
         if isinstance(value, FieldInfo):
             field_info = value
-            value = field_info.default_factory if field_info.default_factory is not None else field_info.default
+            value = field_info.default_factory() if field_info.default_factory is not None else field_info.default
         else:
             field_info = FieldInfo(value, **field_info_from_config)
         required: 'BoolUndefined' = Undefined
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -25,7 +25,7 @@ from .class_validators import Validator, make_generic_validator, prep_validators
 from .error_wrappers import ErrorWrapper
 from .errors import NoneIsNotAllowedError
 from .types import Json, JsonWrapper
-from .typing import AnyType, Callable, ForwardRef, NoneType, display_as_type, is_literal_type
+from .typing import AnyType, Callable, ForwardRef, NoArgAnyCallable, NoneType, display_as_type, is_literal_type
 from .utils import PyObjectStr, Representation, lenient_issubclass, sequence_like
 from .validators import constant_validator, dict_validator, find_validators, validate_json
 
@@ -58,6 +58,7 @@ class FieldInfo(Representation):
 
     __slots__ = (
         'default',
+        'default_factory',
         'alias',
         'alias_priority',
         'title',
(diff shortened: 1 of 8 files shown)


MrMrRobat:
Discussed in https://github.com/samuelcolvin/pydantic/issues/866

So I would like to see this:
```py
from datetime import datetime

class Record(BaseModel):
    name: str
    description: str
    ts: datetime = datetime.now

record = Record(name='Foo', description='Bar and Baz also.')
print(record.ts)
#> 2020-02-06 22:11:43.454720
```

However if type of field is callable, it must remain untouched:
```py
class IntGenerator(BaseModel):
    fabric: Callable[[], int] = randint

generator = IntGenerator(name='Foo', description='Bar and Baz also.')
print(generator.fabric(0, 1))
#> 1
```

@samuelcolvin, any thoughts?

PrettyWood:
@MrMrRobat Excellent ! I thought about this implementation but I was afraid it was not wanted. Love it !

pydantic/fields.py
```diff
@@ -76,8 +77,12 @@ class FieldInfo(Representation):
         'extra',
     )
 
-    def __init__(self, default: Any, **kwargs: Any) -> None:
+    def __init__(self, default: Any = Undefined, *, default_factory: Any = None, **kwargs: Any) -> None:
+        if default is not Undefined and default_factory is not None:
```
samuelcolvin: perhaps better to move this check into `Field`m then you can use `pop()` as is done elsewhere.

pydantic/fields.py
```diff
@@ -97,8 +102,9 @@ def __init__(self, default: Any, **kwargs: Any) -> None:
 
 
 def Field(
-    default: Any,
+    default: Any = Undefined,
     *,
+    default_factory: Any = None,
```
samuelcolvin:
```suggestion
    default_factory: Optional[Callable[[], Any]] = None,
```
or something

samuelcolvin:
I'm keen on the approach of

```py
from datetime import datetime

class Record(BaseModel):
    ts: datetime = datetime.now
```

However I think it could cause subtle changes in behaviour in edge cases so let's wait for v2.

Getting `Field(default_factory=...)` to work now is a great first step.

tests/test_main.py
```diff
@@ -970,3 +993,35 @@ class NewModel(DerivedModel, something=2):
...
+
+    m1 = DynamicValueModel()
+    m2 = DynamicValueModel()
+    assert isinstance(m1.uid, UUID)
+    assert m1.uid != m2.uid
+
+    # With a callable: we still should be able to set callables as defaults
+    class FunctionModel(BaseModel):
+        a: int = 1
+        uid: Callable[[], UUID] = Field(uuid4)
```
PrettyWood: Had to use `= Field(uuid)` instead of `= uuid4` as functions are ignored when creating fields and the default is hence not properly set
I think this behaviour should be fixed but in another PR (maybe for v2 with the other syntax for dynamic values)

pydantic/fields.py
```diff
@@ -97,8 +99,9 @@ def __init__(self, default: Any, **kwargs: Any) -> None:
 
 
 def Field(
-    default: Any,
+    default: Any = Undefined,
     *,
+    default_factory: Optional['Callable[[], Any]'] = None,
```
PrettyWood: Had to use `'Callable...'` because it would raise
`TypeError: 'ABCMeta' object is not subscriptable`

MrMrRobat: How is that possible? It already used in this module without an issues. That could be the case for collections.Callable, but it's a different thing.

PrettyWood: Ah right ! I didnt' see the `Callable` came from `.typing`. Thanks !

pydantic/fields.py
```diff
@@ -3,6 +3,7 @@
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
```
samuelcolvin: if the below is `TypingCallable`, what is this? Maybe good to give this an alias.

pydantic/fields.py
```diff
@@ -307,8 +321,9 @@ def prepare(self) -> None:
         Note: this method is **not** idempotent (because _type_analysis is not idempotent),
         e.g. calling it it multiple times may modify the field and configure it incorrectly.
         """
-        if self.default is not None and self.type_ is None:
-            self.type_ = type(self.default)
+        default_value = self.default_factory() if self.default_factory is not None else self.default
```
samuelcolvin: perhaps useful to have a `get_default()` function on this class which is used here and in `main.py` and perhaps deals with deepcopy?

pydantic/fields.py
PrettyWood: I decided to add a `NoArgCallable` directly in the local `typing` module. Makes things clearer imo

PrettyWood:
@samuelcolvin Thanks for the review ! Changes done

pydantic/fields.py
```diff
@@ -272,13 +296,13 @@ def infer(
 
         if isinstance(value, FieldInfo):
             field_info = value
-            value = field_info.default
+            value = field_info.default_factory if field_info.default_factory is not None else field_info.default
```
samuelcolvin: This looks a bit weird here, shouldn't we use

```suggestion
            value = field_info.default_factory() if field_info.default_factory is not None else field_info.default
```

or otherwise modify the logic here? I know `value` being a callable won't break the logic below, but it's not what you would expect and could cause errors in future?

PrettyWood: Of course ! Sorry forgot to change it here

jmagnusson:
@samuelcolvin When do you think a release will be ready with this PR? Looking forward to it!

samuelcolvin:
@jmagnusson see #1341 for details on release date.

PR #1253, opened 2020-02-25
Fix parsing nested custom root type models

### LINKED ISSUE #1190: Broken loading list of tuples.

Shados:
## Change Summary

When validating fields that are pydantic models, if all else fails explicitly check if `__custom_root_type__` is true, and try to `parse_obj` if it is.

## Related issue number

Fixes #1190.

## Checklist

### DIFF
--- a/changes/1190-Shados.md
+++ b/changes/1190-Shados.md
@@ -1,0 +1,1 @@
+Fixed parsing of nested 'custom root type' models.
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -546,6 +546,8 @@ class BaseModel(metaclass=ModelMetaclass):
             return value.copy()
         elif cls.__config__.orm_mode:
             return cls.from_orm(value)
+        elif cls.__custom_root_type__:
+            return cls.parse_obj(value)
         else:
             try:
                 value_as_dict = dict(value)
--- a/tests/test_parse.py
+++ b/tests/test_parse.py
@@ -1,10 +1,10 @@
 import json
 import pickle
-from typing import List, Union
+from typing import List, Tuple, Union
 
 import pytest
 
-from pydantic import BaseModel, Field, Protocol, ValidationError
+from pydantic import BaseModel, Field, Protocol, ValidationError, parse_obj_as
 
 
 class Model(BaseModel):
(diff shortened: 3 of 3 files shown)


Shados:
> looks like we still don't have a test for nested root models like i put on the issue.
> 
> Otherwise LGTM.

Not sure which you mean. I do have the example `parse_as_obj` case included. I guess I could split that out into its own test?

samuelcolvin:
There's still no explicit test for a custom root type as as the type of a custom root type:

```py

class Foo(BaseModel):
    __root__: Dict[str, int]

class Bar(BaseModel):
    __root__: Foo
```

Shados:
@samuelcolvin I have no idea why anyone would want do that *explicitly*, but I've added a test covering it.

samuelcolvin:
thanks so much.

PR #1258, opened 2020-02-25
add pydantic.color.Color objects as available input for Color fields

leosussan:
## Change Summary

If you pass a `pydantic.color.Color` object to a `Color` field, it will fail b/c the object 

This small edit allows for `Color` objects as acceptable input for Color fields.

## Checklist

### DIFF
--- a/pydantic/color.py
+++ b/pydantic/color.py
@@ -70,11 +70,13 @@ class Color(Representation):
             self._rgba = parse_tuple(value)
         elif isinstance(value, str):
             self._rgba = parse_str(value)
+        elif isinstance(value, Color):
+            self._rgba = value._rgba
         else:
-            raise ColorError(reason='value must be a tuple, list or string')
+            raise ColorError(reason='value must be a tuple, list, string or pydantic.color.Color')
 
         # if we've got here value must be a valid color
-        self._original = value
+        self._original = value if not isinstance(value, Color) else value._original
 
     @classmethod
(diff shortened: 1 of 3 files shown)


pydantic/color.py
```diff
@@ -70,11 +70,13 @@ def __init__(self, value: ColorType) -> None:
             self._rgba = parse_tuple(value)
         elif isinstance(value, str):
             self._rgba = parse_str(value)
+        elif isinstance(value, Color):
+            self._rgba = value._rgba
```
samuelcolvin: Maybe add `value = value._original` here?

pydantic/color.py
```diff
@@ -70,11 +70,13 @@ def __init__(self, value: ColorType) -> None:
...
         else:
-            raise ColorError(reason='value must be a tuple, list or string')
+            raise ColorError(reason='value must be a tuple, list, string or pydantic.color.Color')
```
samuelcolvin: I think no need to change this, I think it's expected and this could be a confusing message to end users.

leosussan: Gotcha - changing now.

pydantic/color.py
leosussan: Added & removed the conditional below it - thank you for the suggest

samuelcolvin:
awesome, thank you. Sorry about the docs build, I'll fix it when I'm back in the UK.

PR #1263, opened 2020-02-27
validate_arguments decorator example implementation

### LINKED ISSUE #1222: @validate_arguments on instance methods
# Bug

Hello, I tried using the new `@validate_arguments` decorator and it doesn't work when used on instance methods.

I didn't see it on the ToDo in #1205 and it seems like an oversight, maybe due to the special treatment of `self`.

```
$ python3 -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.4a1
            pydantic compiled: False
                 install path: /home/[user]/git/pydantic/pydantic
               python version: 3.7.5 (default, Nov 20 2019, 09:21:52)  [GCC 9.2.1 20191008]
                     platform: Linux-5.3.0-29-generic-x86_64-with-Ubuntu-19.10-eoan
     optional deps. installed: []
```

```py
from pydantic import validate_arguments


class SomeObject:
    @validate_arguments
    def some_function(self, i: int):
        print(type(self), self)
        print(type(i), i)

o = SomeObject()
o.some_function(1)  # doesn't work, instead of `i` `self` becomes 1
#pydantic.error_wrappers.ValidationError: 1 validation error for SomeFunction
#i
#  field required (type=value_error.missing)

o.some_function(o, 1)  # works, but not the way instance methods are meant to be used
#<class '__main__.SomeObject'> <__main__.SomeObject object at 0x7f32911af3d0>
#<class 'int'> 1
```

BLooperZ:
This is an alternative implementation for the validate_arguments decorator #1179
Which also happen to fix #1222
(It also does not relay on arguments naming conventions) 

I would not recommend merging this as it does not wrap raised TypeError as ValidationError (making relevant tests fail).
But might be useful for extracting logic and finding out what causes the above issue.

### DIFF
--- a/pydantic/decorator.py
+++ b/pydantic/decorator.py
@@ -62,17 +62,31 @@ def validate_arguments(func: Callable[..., T]) -> Callable[..., T]:
         **var_keyword,
     )
 
+    sig_pos = tuple(positional_only) + tuple(positional_or_keyword)
+    sig_kw = set(positional_or_keyword) | set(keyword_only)
+
     class SignatureCheck(BaseModel):
         args: Dict[str, Any]
         kwargs: Dict[str, Any]
 
         @validator('args', pre=True, allow_reuse=True)
         def validate_positional(cls, args: Any) -> Dict[str, Any]:
-            return sig.bind_partial(*args).arguments
+            try:
+                return sig.bind_partial(*args).arguments
+            except TypeError:
+                raise TypeError(f'{len(sig_pos)} positional arguments expected but {len(args)} given')
 
         @validator('kwargs', pre=True, allow_reuse=True)
         def validate_keyword(cls, kwargs: Any) -> Any:
-            return sig.bind_partial(**kwargs).arguments
+            try:
+                return sig.bind_partial(**kwargs).arguments
+            except TypeError as e:
+                unexpected = set(kwargs) - sig_kw
+                if unexpected:
+                    plural = '' if len(unexpected) == 1 else 's'
+                    keys = ', '.join(map(repr, unexpected))
+                    raise TypeError(f'unexpected keyword argument{plural}: {keys}')
+                raise e
 
     @wraps(func)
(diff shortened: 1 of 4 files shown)


BLooperZ:
Update:
Now this implementation support wrapping TypeError in ValidationError.
Validation for calling the function happens on separate stage, so those errors won't mix with value errors (which raised after the model has instantiated)

I have edited the tests to highlight the differences, please have a look.

samuelcolvin:
Still seems to be failing, also I'm not sure what the advantages of this approach are?

I've provided a simpler solution to #1222 in #1272, please let me know if that's acceptable.

BLooperZ:
is there something else failing beside string annotations? (I have left the test case active to reflect that)

I liked in this approach to have python itself bind the arguments
(SignatureCheck model is only used to wrap error message around signature.bind - can ignore)

and that we don't have to rely on nor restrict arguments names,
that's achieved by using 2 step approach for input validation:
calling -> throw error if the function can't be called the way it was called (too many positional arguments, multiple values for same argument, unexpected keyword, positional only passed as keyword).
values -> arguments are not valid for annotated types (pydantic's BaseModel)

The logic part is pretty straightforward (getting arguments by kind -> check bindings -> validate arguments on matching model -> call function with validated arguments) - from entry to model creation is just transformation over a mapping, and the apply wrapper itself, which does signature check and validation is also very easy to understand and short.

of course you may decide whatever works, so all is acceptable, but it is also worth to look at this (I suggest looking the file `decorator.py` itself, not the diff).

Thank you for your response and this amazing project.

samuelcolvin:
This looks like an interesting approach, but how does it effect performance?

I guess we need some kind of benchmark to compare it with. If you don't have time, I'll try to get around to building a benchmark when I have more time.

samuelcolvin:
in the interests of getting the number of open PRs down to a manageable level, I'm closing this but could be reopened if performance benchmarks show it's a more performant approach.

PR #1266, opened 2020-02-28
Allow None for (virtual) superclasses of NoneType

### LINKED ISSUE #1262: "none is not an allowed value" for Hashable field
Although `None` is `Hashable`, it fails to validate below. Even stranger, it still fails with the same error if the field is explicitly annotated as `Optional[Hashable]`

```py
from collections.abc import Hashable
from typing import Optional


class Config:

@dataclass(config=Config)
class Model:
    v: Hashable
    # v: Optional[Hashable]

assert isinstance(None, Hashable)

Model(None)

# output:
#
#pydantic.error_wrappers.ValidationError: 1 validation error for Model
#v
#  none is not an allowed value (type=type_error.none.not_allowed)
```

```
pydantic version: 1.4
pydantic compiled: True
python version: 3.6.9 (default, Nov  7 2019, 10:44:02)  [GCC 8.3.0]
platform: Linux-4.15.0-76-generic-x86_64-with-Ubuntu-18.04-bionic
```

gsakkis:
Fixes #1262

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -401,6 +401,9 @@ class ModelField(Representation):
         origin = getattr(self.type_, '__origin__', None)
         if origin is None:
             # field is not "typing" object eg. Union, Dict, List etc.
+            # allow None for virtual superclasses of NoneType, e.g. Hashable
+            if isinstance(self.type_, type) and isinstance(None, self.type_):
+                self.allow_none = True
             return
         if origin is Callable:
             return
--- a/pydantic/errors.py
+++ b/pydantic/errors.py
@@ -61,6 +61,7 @@ __all__ = (
     'DateError',
     'TimeError',
     'DurationError',
+    'HashableError',
     'UUIDError',
     'UUIDVersionError',
     'ArbitraryTypeError',
@@ -390,6 +391,10 @@ class TimeError(PydanticValueError):
 
 class DurationError(PydanticValueError):
     msg_template = 'invalid duration format'
+
+
+class HashableError(PydanticTypeError):
+    msg_template = 'value is not a valid hashable'
 
 
 class UUIDError(PydanticTypeError):
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -1,6 +1,7 @@
 import re
 import sys
 from collections import OrderedDict
+from collections.abc import Hashable
 from datetime import date, datetime, time, timedelta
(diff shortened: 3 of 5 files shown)


pydantic/fields.py
```diff
@@ -364,6 +364,9 @@ def _type_analysis(self) -> None:  # noqa: C901 (ignore complexity)
         origin = getattr(self.type_, '__origin__', None)
         if origin is None:
             # field is not "typing" object eg. Union, Dict, List etc.
+            # allow None for virtual superclasses of NoneType, e.g. Hashable
+            if isinstance(self.type_, type) and isinstance(None, self.type_):
```
dmontagu: Are there any other virtual superclasses of `NoneType`? Is there a reference of this behavior anywhere?

gsakkis: I'm not aware of other virtual subclasses, however the one non-virtual superclass of `NoneType` - `object` - is also handled by this change. 

The behavior is provided by [ABCMeta.register](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register) or the [ABCMeta.__subclasshook__](https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__) classmethod. Toy example:
```py

In [1]: from abc import ABCMeta                                                                                                                                          

In [2]: class Singleton(metaclass=ABCMeta): 
   ...:     pass 

In [3]: for obj in None, Ellipsis, NotImplemented, True, False: 
   ...:     Singleton.register(type(obj)) 

In [4]: isinstance(None, Singleton)                                                                                                                                      
Out[4]: True

In [5]: isinstance(True, Singleton)                                                                                                                                      
Out[5]: True

In [6]: isinstance(1, Singleton)                                                                                                                                         
Out[6]: False
```

samuelcolvin:
This needs a change description.

But also, would it not be a good idea to fully support `Hashable` e.g. check for the `__hash__()` method, at the same time?

samuelcolvin:
@gsakkis any chance you could fix this before v1.5? See #1341.

gsakkis:
@samuelcolvin I added full support for Hashable. 

By the way, there's something wrong in the CI for Windows builds. Initially I thought tests on Linux builds were failing due to Cython but actually the same tests were failing in all platforms; they were just misleadingly reported as passing for Windows.

samuelcolvin:
awesome, thank you. So sorry this took so long to get merged.

Issue #1283, opened 2020-03-02
Model export fails when excluding fields from a List of nested models


masalim2:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.4
            pydantic compiled: False
                 install path: /Users/misha/workflow/balsam/env/lib/python3.7/site-packages/pydantic
               python version: 3.7.4 (default, Sep  7 2019, 18:27:02)  [Clang 10.0.1 (clang-1001.0.46.4)]
                     platform: Darwin-18.7.0-x86_64-i386-64bit
     optional deps. installed: []
```

Hi all, thanks for the terrific library -- it has been a pleasure to use so far.  I am attempting to export a nested model via ``dict()`` with the ``exclude`` kwarg to omit some fields in the inner child model.  This ``exclude`` seems to cause a ``TypeError`` when the exclusion field pertains to a *list* of nested models.  

I wrote a simple reproducer for the issue below.  I am encountering a TypeError inside the the ``_normalize_indexes`` method of ``pydantic.utils.ValueItems``. 

```py
from pydantic import BaseModel
from typing import List

class Foo(BaseModel):
    id: int

class Bar(BaseModel):
    id: int
    foos: List[Foo] = []

my_bar = Bar(id=3)
my_bar.dict()  # This is ok 
my_bar.dict(exclude={"foos": {"id"}}) # This causes TypeError
```

The Traceback is:
```
Traceback (most recent call last):
  File "tester.py", line 13, in <module>
    my_bar.dict(exclude={"foos": {"id"}})
  File "/Users/misha/workflow/balsam/env/lib/python3.7/site-packages/pydantic/main.py", line 343, in dict
    exclude_none=exclude_none,
  File "/Users/misha/workflow/balsam/env/lib/python3.7/site-packages/pydantic/main.py", line 333, in <dictcomp>
    return {
  File "/Users/misha/workflow/balsam/env/lib/python3.7/site-packages/pydantic/main.py", line 675, in _iter
    exclude_none=exclude_none,
  File "/Users/misha/workflow/balsam/env/lib/python3.7/site-packages/pydantic/main.py", line 587, in _get_value
    value_exclude = ValueItems(v, exclude) if exclude else None
  File "/Users/misha/workflow/balsam/env/lib/python3.7/site-packages/pydantic/utils.py", line 295, in __init__
    items = self._normalize_indexes(items, len(value))
[...]

PR #1286, opened 2020-03-03
Use __all__ to exclude fields from all elements of a list/tuple of submodels.  Raise more informative TypeError when passing bad exclude kwargs.

masalim2:
## Change Summary

1. Raise TypeError in ``pydantic.utils.ValueItems`` when ``value`` is a list or tuple and the provided exclusion dict is not keyed by index.
2. Add cautionary sentence to the **Advanced include and exclude** section of docs to point out correct usage of ``exclude`` in the following example. 
3. Use `'__all__'` keyword  to exclude fields from all elements of a sequence of submodels or dicts.

## Related issue number

This PR addresses a couple of my comments in #1283  -- adding a more detailed TypeError and a sentence pointing out the subtle example of correct usage.  Some logic has been added to enable usage of `'__all__'` in the `exclude` kwarg of model export methods.

## Checklist

### DIFF
--- a/changes/1286-masalim2.md
+++ b/changes/1286-masalim2.md
@@ -1,0 +1,5 @@
+Raise more informative ``TypeError`` in ``pydantic.utils.ValueItems.__init__``
+when ``value`` is a list/tuple and the exclusion kwarg is improperly indexed.
+Added special handling for the exclusion key ``'__all__'``, which allows one to
+exclude fields from all elements of a list/tuple of submodels. Added example to
+Exporting Models docs.
--- a/changes/1286-masalim2.md
+++ b/changes/1286-masalim2.md
@@ -1,5 +1,1 @@
-Raise more informative ``TypeError`` in ``pydantic.utils.ValueItems.__init__``
-when ``value`` is a list/tuple and the exclusion kwarg is improperly indexed.
-Added special handling for the exclusion key ``'__all__'``, which allows one to
-exclude fields from all elements of a list/tuple of submodels. Added example to
-Exporting Models docs.
+Exclude exported fields from all elements of a list/tuple of submodels with `'__all__'`.
--- a/changes/1286-masalim2.md
+++ b/changes/1286-masalim2.md
@@ -1,1 +1,1 @@
-Exclude exported fields from all elements of a list/tuple of submodels with `'__all__'`.
+Exclude exported fields from all elements of a list/tuple of submodels/dicts with `'__all__'`.
--- a/docs/usage/exporting_models.md
+++ b/docs/usage/exporting_models.md
@@ -113,13 +113,19 @@ sets or dictionaries. This allows nested selection of which fields to export:
 ```
 
 The ellipsis (``...``) indicates that we want to exclude or include an entire key, just as if we included it in a set.
-Of course, the same can be done at any depth level:
+Of course, the same can be done at any depth level.
+Special care must be taken when including or excluding fields from a list or tuple of submodels.  In this scenario,
+`dict` and related methods expect integer keys for element-wise inclusion or exclusion:
 
 ```py
 {!.tmp_examples/exporting_models_exclude2.py!}
 ```
 
 The same holds for the `json` and `copy` methods.
+
+
+```py
+{!.tmp_examples/exporting_models_exclude3
 
 ## Custom JSON (de)serialisation
 
--- a/docs/usage/exporting_models.md
+++ b/docs/usage/exporting_models.md
@@ -124,9 +124,6 @@ Special care must be taken when including or excluding fields from a list or tup
 The same holds for the `json` and `copy` methods.
 
 
-```py
-{!.tmp_examples/exporting_models_exclude3
-
 ## Custom JSON (de)serialisation
 
 To improve the performance of encoding and decoding JSON, alternative JSON implementations
--- a/docs/usage/exporting_models.md
+++ b/docs/usage/exporting_models.md
@@ -115,7 +115,7 @@ sets or dictionaries. This allows nested selection of which fields to export:
 The ellipsis (``...``) indicates that we want to exclude or include an entire key, just as if we included it in a set.
 Of course, the same can be done at any depth level.
 
-Special care must be taken when including or excluding fields from a list or tuple of submodels.  In this scenario,
+Special care must be taken when including or excluding fields from a list or tuple of submodels or dictionaries.  In this scenario,
 `dict` and related methods expect integer keys for element-wise inclusion or exclusion. To exclude a field from **every**
 member of a list or tuple, the dictionary key `'__all__'` can be used as follows:
 
--- a/docs/usage/exporting_models.md
+++ b/docs/usage/exporting_models.md
@@ -117,14 +117,13 @@ Of course, the same can be done at any depth level.
 
 Special care must be taken when including or excluding fields from a list or tuple of submodels.  In this scenario,
 `dict` and related methods expect integer keys for element-wise inclusion or exclusion. To exclude a field from **every**
-member of a list or tuple, the dictionary key ``"__all__"`` can be used as follows:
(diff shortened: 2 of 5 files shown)


changes/1286-masalim2.md
```diff
@@ -0,0 +1,5 @@
+Raise more informative ``TypeError`` in ``pydantic.utils.ValueItems.__init__``
```
samuelcolvin: single back quotes please, this is markdown. Also I try to keep this to one line (roughly 140 max.) remember there's a link to the actual issue/PR for those who want to understand more.

docs/usage/exporting_models.md
```diff
@@ -113,14 +113,19 @@ sets or dictionaries. This allows nested selection of which fields to export:
 ```
 
 The ellipsis (``...``) indicates that we want to exclude or include an entire key, just as if we included it in a set.
-Of course, the same can be done at any depth level:
+Of course, the same can be done at any depth level.
+
+Special care must be taken when including or excluding fields from a list or tuple of submodels.  In this scenario,
+`dict` and related methods expect integer keys for element-wise inclusion or exclusion. To exclude a field from **every**
+member of a list or tuple, the dictionary key ``"__all__"`` can be used as follows:
```
samuelcolvin:
```suggestion
member of a list or tuple, the dictionary key `'__all__'` can be used as follows:
```

docs/usage/exporting_models.md
```diff
@@ -113,14 +113,19 @@ sets or dictionaries. This allows nested selection of which fields to export:
...
 
 ```py
 {!.tmp_examples/exporting_models_exclude2.py!}
 ```
 
 The same holds for the `json` and `copy` methods.
 
+
```
samuelcolvin: remove double line break, try to keep changes constrained to the matter in question.

pydantic/utils.py
```diff
@@ -345,7 +345,13 @@ def __init__(self, value: Any, items: Union['AbstractSetIntStr', 'DictIntStrAny'
             raise TypeError(f'Unexpected type of exclude value {type(items)}')
 
         if isinstance(value, (list, tuple)):
-            items = self._normalize_indexes(items, len(value))
+            try:
+                items = self._normalize_indexes(items, len(value))
+            except TypeError as e:
+                raise TypeError(
+                    f'Excluding fields from a list or tuple of sub-models must be performed index-wise: '
+                    f'expected integer keys or keyword __all__'
```
samuelcolvin:
```suggestion
                    f'expected integer keys or keyword "__all__"'
```

tests/test_main.py
```diff
@@ -956,6 +956,44 @@ class Bar(BaseModel):
...
+    class Foo(BaseModel):
+        a: int = 1
+        b: int = 2
+
+    class Bar(BaseModel):
+        c: int
+        foos: List[Foo]
+
+    m = Bar(c=3, foos=[Foo(), Foo()])
+    exclusion = {idx: {'a'} for idx in range(len(m.foos))}
```
samuelcolvin: better to be explicit here (we're in a test)
```suggestion
    exclusion = {0: {'a'}, 1: {'a'}}
```

tests/test_main.py
```diff
@@ -956,6 +956,44 @@ class Bar(BaseModel):
...
+def test_model_export_nested_list_with_all():
+    class Foo(BaseModel):
+        a: int = 1
+        b: int = 2
+
+    class Bar(BaseModel):
+        c: int
+        foos: List[Foo]
+
+    m = Bar(c=3, foos=[Foo(), Foo()])
```
samuelcolvin:
```suggestion
    m = Bar(c=3, foos=[Foo(a=1, b=2), Foo(a=3, b=4)])
```
to make sure the below logic is working correctly.

tests/test_main.py
```diff
@@ -956,6 +956,44 @@ class Bar(BaseModel):
...
+    exclusion = {idx: {'a'} for idx in range(len(m.foos))}
+    assert m.dict(exclude={'foos': exclusion}) == {'c': 3, 'foos': [{'b': 2}, {'b': 2}]}
+
+    with pytest.raises(TypeError) as e:
+        m.dict(exclude={'foos': {'a'}})
+    assert 'expected integer keys' in str(e.value)
+
+
+def test_model_export_nested_list_with_all():
```
samuelcolvin: I think this test and the above can be combined. Generally I'm anti combining tests, but here the set is the same and we're just testing different calls.

tests/test_main.py
```diff
@@ -956,6 +956,44 @@ class Bar(BaseModel):
...
+        a: int = 1
+        b: int = 2
+
+    class Bar(BaseModel):
+        c: int
+        foos: List[Foo]
+
+    m = Bar(c=3, foos=[Foo(), Foo()])
+
+    exclusion = {0: {'b'}, '__all__': {'a'}}
```
samuelcolvin: I think no need to create a separate variable, just run `make format` after writing and everything will be formatted nicely.

or you could use parameterisation for all these cases?

pydantic/utils.py
```diff
@@ -393,10 +399,19 @@ def _normalize_indexes(
 
         >>> self._normalize_indexes({0, -2, -1}, 4)
         {0, 2, 3}
+        >>> self._normalize_indexes({'__all__'}, 4)
+        {0, 1, 2, 3}
         """
         if self._type is set:
+            if '__all__' in items:
+                return {i for i in range(v_length)}
```
samuelcolvin: should it be an error if there's anything else in the set besides `'__all__'`?

masalim2: Hmm, don't have a strong opinion either way.  It certainly doesn't make sense to intentionally put `__all__` along with other indices in the set.  But I don't think the behavior would be surprising or unexpected in this case either.

masalim2:
Thanks for reviewing!  Addressed all but the last question in the last commit.

docs/usage/exporting_models.md
```diff
@@ -113,7 +113,11 @@ sets or dictionaries. This allows nested selection of which fields to export:
 ```
 
 The ellipsis (``...``) indicates that we want to exclude or include an entire key, just as if we included it in a set.
-Of course, the same can be done at any depth level:
+Of course, the same can be done at any depth level.
+
+Special care must be taken when including or excluding fields from a list or tuple of submodels.  In this scenario,
```
samuelcolvin: This should work on dictionaries as fields as well as sub-models

pydantic/utils.py
```diff
@@ -345,7 +345,13 @@ def __init__(self, value: Any, items: Union['AbstractSetIntStr', 'DictIntStrAny'
             raise TypeError(f'Unexpected type of exclude value {type(items)}')
 
         if isinstance(value, (list, tuple)):
-            items = self._normalize_indexes(items, len(value))
+            try:
+                items = self._normalize_indexes(items, len(value))
+            except TypeError as e:
+                raise TypeError(
+                    f'Excluding fields from a list or tuple of sub-models must be performed index-wise: '
```
samuelcolvin: again I think we should mention dicts as well as sub-models.

pydantic/utils.py
samuelcolvin: Let's raise an error.

tests/test_main.py
```diff
@@ -956,6 +956,28 @@ class Bar(BaseModel):
...
 
 
+def test_model_export_nested_list():
+    class Foo(BaseModel):
+        a: int = 1
+        b: int = 2
+
+    class Bar(BaseModel):
+        c: int
+        foos: List[Foo]
```
samuelcolvin: again can we add another test with `foos: List[Dict[str, int]]` or similar?

masalim2:
Done.  Should I rebase on upstream master to resolve conflicts, or do you want to handle that yourself?

samuelcolvin:
I've made a few tiny tweaks to avoid wasting your time with another review.  Will merge once tests pass.

Issue #1157, opened 2020-01-08
Serialization of UUID childs


unitto1:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.3
            pydantic compiled: False
                 install path: /usr/local/lib/python3.8/site-packages/pydantic
               python version: 3.8.0 (default, Nov 15 2019, 02:22:06)  [GCC 8.3.0]
                     platform: Linux-5.3.8-133.current-x86_64-with
     optional deps. installed: ['typing-extensions', 'email-validator']

```

```py
from uuid import UUID

import asyncpg
from pydantic import BaseModel

struct_dict = {
   "id": asyncpg.pgproto.pgproto.UUID('c9ecc20f-acd6-429a-b781-c59f803af240')
}

class StructSchema(BaseModel):
    id: UUID

struct = StructSchema(**struct_dict)
assert type(struct.id) == asyncpg.pgproto.pgproto.UUID
assert isinstance(struct.id, UUID)

try:
    struct.json()  # raises error
except Exception as err:
    print(err)

struct.id = UUID(str(struct.id))

struct.json()  # ok, because now it really UUID instead of asyncpg's one
```

Funny one, but little annoying =)

dmontagu:
I believe this is related to how python's built-in `json.dumps` function works -- the `default` keyword argument takes a dictionary of type to encoder. Pydantic adds an encoder for `uuid.UUID` by default, but I believe the lookup is done by actual type rather than subclass (presumably for performance).

You can set a custom encoder on the model `Config` via `json_encoders`, which should fix this issue: https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeljson

It would be nice if subclasses were handled automatically, but I believe there are nontrivial performance implications since you'd have to loop over all possible encoders looking for one that could encode, rather than being able to do a dict-based lookup. There is actually an open PR in FastAPI trying to address this (and looking at switching to subclass-based checks), but I'm not sure the performance implications are worth it.

elprans:
> the default keyword argument takes a dictionary of (exact) type to encoder

Not true.  The `default` arguments takes a callback.

> It would be nice if subclasses were handled automatically, but I believe there are not-insignificant performance implications since you'd have to loop over all possible encoders looking for one that could encode, rather than being able to do a dict-based lookup.

[...]

PR #1291, opened 2020-03-04
Allow subclasses of known types to be encoded with superclass encoder

### LINKED ISSUE #1281: Subclasses of known types are not JSON serializable
# Feature Request/(Bug?)

```
        pydantic version: 1.4
       pydantic compiled: False
            install path: /Users/step7212/.pyenv/versions/3.8.1/envs/fastdate/lib/python3.8/site-packages/pydantic
          python version: 3.8.1 (default, Jan 10 2020, 09:36:37)  [Clang 11.0.0 (clang-1100.0.33.16)]
optional deps. installed: []
```

I would like to use a package that subclasses a standard python library class, adding convenience features but maintaining the interface of it's inheritance. Currently, pydantic will error when attempting to serialize the subclass, as it is not known in `ENCODERS_BY_TYPE`: https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/json.py#L20-L40

Example:
```py
>>> import pendulum
>>> import pydantic

>>> class MyModel(pydantic.BaseModel):
...     date_field: pendulum.DateTime
... 
>>> m = MyModel(date_field=pendulum.now('UTC'))
>>> m
MyModel(date_field=DateTime(2020, 3, 2, 16, 44, 42, 977836, tzinfo=Timezone('UTC')))
>>> m.dict()
{'date_field': DateTime(2020, 3, 2, 16, 44, 42, 977836, tzinfo=Timezone('UTC'))}
>>> m.json()
  File "/Users/step7212/git/hub/pydantic/pydantic/json.py", line 57, in pydantic_encoder
    encoder = ENCODERS_BY_TYPE[type(obj)]
KeyError: <class 'pendulum.datetime.DateTime'>

  File "/Users/step7212/git/hub/pydantic/pydantic/main.py", line 419, in json
    return self.__config__.json_dumps(data, default=encoder, **dumps_kwargs)
  File "/Users/step7212/.pyenv/versions/3.8.1/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/Users/step7212/.pyenv/versions/3.8.1/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/step7212/.pyenv/versions/3.8.1/lib/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/step7212/git/hub/pydantic/pydantic/json.py", line 59, in pydantic_encoder
    raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
TypeError: Object of type 'DateTime' is not JSON serializable
```

Since all pendulum types are children of the std `datetime` types, they should be serializable in the same way:
```py
>>> import pendulum
>>> import pydantic
>>> 
>>> class MyModel(pydantic.BaseModel):
...     date_field: pendulum.DateTime
... 
>>> m = MyModel(date_field=pendulum.now('UTC'))
>>> m
MyModel(date_field=DateTime(2020, 3, 2, 16, 42, 22, 403965, tzinfo=Timezone('UTC')))
>>> m.dict()
{'date_field': DateTime(2020, 3, 2, 16, 42, 22, 403965, tzinfo=Timezone('UTC'))}
>>> m.json()
'{"date_field": "2020-03-02T16:42:22.403965+00:00"}'
```
This is doable by iterating over the [method resolution order](https://docs.python.org/3/library/stdtypes.html#class.__mro__) attribute, and I've created a branch to demonstrate, which was used for the desired output above: https://github.com/samuelcolvin/pydantic/compare/master...StephenBrown2:encode_known_subclasses
<details>
<summary>Benchmarks for funsies, this actually turns out to be a little faster than before:</summary>

```
❯ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
❯ make benchmark-pydantic
python benchmarks/run.py pydantic-only
generating test cases...
    pydantic time=1.043s, success=50.10%
    pydantic time=0.982s, success=50.10%
    pydantic time=0.788s, success=50.10%
    pydantic time=0.732s, success=50.10%
    pydantic time=0.741s, success=50.10%
    pydantic best=0.732s, avg=0.857s, stdev=0.145s

    pydantic best=121.947μs/iter avg=142.857μs/iter stdev=24.164μs/iter version=1.4a1
❯ git checkout encode_known_subclasses
Switched to branch 'encode_known_subclasses'
Your branch is up to date with 'mine/encode_known_subclasses'.
❯ make benchmark-pydantic
python benchmarks/run.py pydantic-only
    pydantic time=0.973s, success=50.10%
    pydantic time=0.742s, success=50.10%
    pydantic time=0.731s, success=50.10%
    pydantic time=0.764s, success=50.10%
    pydantic time=0.734s, success=50.10%
    pydantic best=0.731s, avg=0.789s, stdev=0.104s

    pydantic best=121.800μs/iter avg=131.485μs/iter stdev=17.306μs/iter version=1.4a1
```

Perhaps this could also encourage the merge of using the `__class__` attribute instead of `type()` as well: https://github.com/samuelcolvin/pydantic/compare/samuelcolvin:master...use-class-attribute EDITED Rebased on current master: https://github.com/samuelcolvin/pydantic/compare/master...StephenBrown2:use-class-attribute

</details>

<details>
<summary>Aside on usage of __mro__:</summary>

I noticed that [`__mro__`](https://docs.python.org/3/library/stdtypes.html?highlight=__mro__#class.__mro__) is not used in many places, though it is referenced in a few places, mostly in mypy.py:

https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/class_validators.py#L331-L332
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/main.py#L153
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/mypy.py#L84
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/mypy.py#L174
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/mypy.py#L204
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/mypy.py#L275

Rather, [typing's `__supertype__`](https://github.com/python/cpython/blob/ab6423fe2de0ed5f8a0dc86a9c7070229326b0f0/Lib/typing.py#L1929) is used, which I assumed was custom only because I couldn't find it in the [Python docs](https://docs.python.org/3/search.html?q=__supertype__) or anywhere outside of https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/typing.py#L180-L187

Would it be prudent/performant to replace the usage of `__supertype__` with a fast-breaking iteration of `__mro__` or perhaps more common usage of `lenient_subclass` in `typing.py`? and subsequently:
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/fields.py#L361-L362
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/schema.py#L641-L642
In particular, other checks for inheritance:
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/validators.py#L554-L565
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/pydantic/utils.py#L103-L104
and of course, the tests:
https://github.com/samuelcolvin/pydantic/blob/e3243d267b06bda2d7b2213a8bad70f82171033c/tests/test_utils.py#L141-L153

</details>

StephenBrown2:
## Change Summary

After further research, it seems my aside on `__mro__` usage in #1281 was unfounded and unnecessary, and the only other place I figured this change might be useful ([validators.py](https://github.com/samuelcolvin/pydantic/blob/8727914ced9d4546f0f95a453aff78f03b778275/pydantic/validators.py#L557-L568)) seems to already be handled alright. Thus, I'm submitting this PR as-is to resolve the issue found with JSON serialization of subclasses of known types.

While adding documentation, I found that the custom `json_encoders` were being ignored, so I added the same fix to `custom_pydantic_encoder` as well.

## Related issue number

Resolves #1281 

## Checklist


samuelcolvin:
I'm concerned about the performance impact of this as discussed on #1157, I hope to provide a better solution for this kind of thing in future.

You can achieve what you want by using your own `config.json_encoders`

StephenBrown2:
> I'm concerned about the performance impact of this as discussed on #1157, I hope to provide a better solution for this kind of thing in future.

I'm not sure what performance impacts you are seeing, I am seeing the same or slightly faster performance with my tests (cleaning and compiling before each run):
```
Benchmark on master:
python benchmarks/run.py pydantic-only
testing pydantic, 10 times each
    pydantic ( 1/10) time=0.437s, success=50.10%
    pydantic ( 2/10) time=0.451s, success=50.10%
    pydantic ( 3/10) time=0.437s, success=50.10%
    pydantic ( 4/10) time=0.428s, success=50.10%
    pydantic ( 5/10) time=0.433s, success=50.10%
    pydantic ( 6/10) time=0.432s, success=50.10%
    pydantic ( 7/10) time=0.431s, success=50.10%
    pydantic ( 8/10) time=0.419s, success=50.10%
    pydantic ( 9/10) time=0.531s, success=50.10%
    pydantic (10/10) time=0.449s, success=50.10%
    pydantic best=0.419s, avg=0.445s, stdev=0.032s

    pydantic best=69.874μs/iter avg=74.140μs/iter stdev=5.273μs/iter version=1.4a1

Benchmark on encode_known_subclasses:
python benchmarks/run.py pydantic-only
testing pydantic, 10 times each
    pydantic ( 1/10) time=0.467s, success=50.10%
    pydantic ( 2/10) time=0.471s, success=50.10%
    pydantic ( 3/10) time=0.431s, success=50.10%
    pydantic ( 4/10) time=0.427s, success=50.10%
    pydantic ( 5/10) time=0.434s, success=50.10%
    pydantic ( 6/10) time=0.434s, success=50.10%
    pydantic ( 7/10) time=0.423s, success=50.10%
    pydantic ( 8/10) time=0.417s, success=50.10%
    pydantic ( 9/10) time=0.433s, success=50.10%
    pydantic (10/10) time=0.454s, success=50.10%
    pydantic best=0.417s, avg=0.439s, stdev=0.019s

    pydantic best=69.490μs/iter avg=73.185μs/iter stdev=3.087μs/iter version=1.4a1
```

Using this script to make sure things were fair
<details><summary><code>compare_branch_perf.sh</code>: </summary>

```bash
#!/usr/bin/env bash
branch=${1:-$(git branch --show-current)}
export BENCHMARK_REPEATS=${2:-10}
output=$(mktemp -t "benchmark-${branch}")

if [[ "${branch}" == "master" ]]; then
  echo "Currently on master, and no branch specified."
  echo "Either check out the branch to compare or specify it as the first argument."
  break
fi

echo "Preparing to compare benchmarks on master vs ${branch}, ${BENCHMARK_REPEATS} repetitions"

git checkout master
echo "Cleaning and building with cython"
make clean build-cython 2>&1 | awk '{printf "."; system("")} END{print "\n"}'
echo "Benchmarking master..."
echo "Benchmark on master:" > ${output}
make benchmark-pydantic | tee -a ${output}

git checkout ${branch}
echo "Cleaning and building with cython"
make clean build-cython 2>&1 | awk '{printf "."; system("")} END{print "\n"}'
echo "Benchmarking ${branch}..."
echo -e "\nBenchmark on ${branch}:" >> ${output}
make benchmark-pydantic | tee -a ${output}

clear
cat ${output}
```

</details>

Can you show performance benchmarks that show significant degredation?
[This branch in Travis](https://travis-ci.org/samuelcolvin/pydantic/jobs/658387558#L656): `pydantic best=56.288μs/iter avg=56.288μs/iter stdev=0.000μs/iter version=1.4a1`
[Master in Travis](https://travis-ci.org/samuelcolvin/pydantic/jobs/659158749#L670): `                                pydantic best=56.039μs/iter avg=56.039μs/iter stdev=0.000μs/iter version=1.4a1`

A difference of `0.249μs`.

> You can achieve what you want by using your own `config.json_encoders`

Yes, but that's exactly the point of this. In case I want to use `pendulum` objects instead of naive datetimes everywhere, I would have to add the custom `Config.json_encoders` to every single model that has one. This is a simple, performant solution that allows me to use subclasses without having to add a lot of boilerplate to my code, and also resolves #1157.

Of note, I attempted to try a faster solution that still compared all the elements but shortcircuited using set logic to avoid a loop, and came out just a tiny bit slower (a rounding error), using this logic:
```py
    if set(obj.__class__.__mro__) & set(ENCODERS_BY_TYPE.keys()):
        return next(ENCODERS_BY_TYPE[base](obj) for base in obj.__class__.__mro__ if base in ENCODERS_BY_TYPE)
    else:
        raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
```

```
Benchmark on master:
python benchmarks/run.py pydantic-only
testing pydantic, 10 times each
    pydantic ( 1/10) time=0.444s, success=50.10%
    pydantic ( 2/10) time=0.422s, success=50.10%
    pydantic ( 3/10) time=0.415s, success=50.10%
    pydantic ( 4/10) time=0.415s, success=50.10%
    pydantic ( 5/10) time=0.413s, success=50.10%
    pydantic ( 6/10) time=0.420s, success=50.10%
    pydantic ( 7/10) time=0.455s, success=50.10%
    pydantic ( 8/10) time=0.432s, success=50.10%
    pydantic ( 9/10) time=0.420s, success=50.10%
    pydantic (10/10) time=0.442s, success=50.10%
    pydantic best=0.413s, avg=0.428s, stdev=0.015s

    pydantic best=68.833μs/iter avg=71.283μs/iter stdev=2.440μs/iter version=1.4a1

Benchmark on encode_known_subclasses:
python benchmarks/run.py pydantic-only
testing pydantic, 10 times each
    pydantic ( 1/10) time=0.464s, success=50.10%
    pydantic ( 2/10) time=0.425s, success=50.10%
    pydantic ( 3/10) time=0.404s, success=50.10%
    pydantic ( 4/10) time=0.449s, success=50.10%
    pydantic ( 5/10) time=0.429s, success=50.10%
    pydantic ( 6/10) time=0.438s, success=50.10%
    pydantic ( 7/10) time=0.446s, success=50.10%
    pydantic ( 8/10) time=0.421s, success=50.10%
    pydantic ( 9/10) time=0.416s, success=50.10%
    pydantic (10/10) time=0.404s, success=50.10%
    pydantic best=0.404s, avg=0.430s, stdev=0.020s

    pydantic best=67.315μs/iter avg=71.630μs/iter stdev=3.276μs/iter version=1.4a1
```

StephenBrown2:
As it is, FastAPI implements [nearly the same](https://github.com/samuelcolvin/pydantic/issues/1157#issuecomment-572511480) logic [in its `jsonable_encoder`](https://github.com/tiangolo/fastapi/blob/025b38df40a18f94322fab445aee13024fe783da/fastapi/encoders.py#L128-L140), and while I would rather not add a custom encoder to all my classes, I suppose I could subclass `BaseModel` and add it there.

I would argue it is not "always going to be slower", as it will be exactly the same performance for native python types (as a loop over a single element array `O(n/n)` is the same as accessing a value by key `O(1)`), and only get slower as the inheritance chain grows. Thus on average, the performance hit would be negligible as demonstrated, and only applicable if one has a very deeply inherited object. In which case, one should certainly create a custom encoder, which will speed up the execution back to native levels.

Thank you for your time and consideration, and the efforts you've put into making this one of the best libraries for what it does.

pydantic/json.py
```diff
@@ -53,18 +53,17 @@ def pydantic_encoder(obj: Any) -> Any:
...
-        return encoder(obj)
+        raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
 
 
 def custom_pydantic_encoder(type_encoders: Dict[Any, Callable[[Type[Any]], Any]], obj: Any) -> Any:
-    encoder = type_encoders.get(type(obj))
-    if encoder:
-        return encoder(obj)
+    for base in obj.__class__.__mro__:
+        if base in type_encoders:
```
dmontagu: For what it's worth, I think it would be slightly faster to do:
```
encoder = type_encoders.get(base)
if encoder is not None:
    return encoder(obj)
```
Since that way you only do one lookup instead of two.

dmontagu:
@samuelcolvin For what it's worth, I think @StephenBrown2 is right that this should generally be very cheap since it only matters for types with a long inheritance chain (where you could get better performance if desired through specifying an encoder).

I do think it would be nice to support this style of lookup, and the benchmark execution times seem pretty minimally affected. It seems to me much more likely that this change would reduce confusion around json encoding than that it would result in even a measurable reduction in encoding performance.

(That said, @StephenBrown2 it might be good to see what the performance impact looks like for types that *do* have a longer inheritance chain leading back to something encodable.)

On the other hand, I could see an argument in favor of preventing "surprising" performance issues by just requiring the type to be specified explicitly. I would also understand if you have some fundamentally different approach in mind that you intend to implement down the line... 😬

samuelcolvin:
```py
    for base in obj.__class__.__mro__:
        if base in ENCODERS_BY_TYPE:
            return ENCODERS_BY_TYPE[base](obj)
```

Mea culpa I had missed that this trick avoids iterating through `ENCODERS_BY_TYPE` as was suggested in #1157. I therefore think it's probably a good idea.

I also think we can make that slightly faster by switching to

```py
for base in obj.__class__.__mro__:
    try:
        f = ENCODERS_BY_TYPE[base]
    except KeyError:
        continue
    return f(obj)
```
So we're only doing one lookup on `ENCODERS_BY_TYPE` (unless I'm missing something and the above is somehow faster than this?).

However I'm still a bit concerned about performance of this and think we should put the time into a small benchmark (or extension to the current benchmarks). Unless I'm missing something `benchmarks/run.py` never calls `.json()` so does nothing to measure the performance of this PR.

In summary "yes", but we should invest the time in benchmarking `.json()` before we make this change.

StephenBrown2:
Also alternatively, the loop could be put in the exception block, and the list sliced to reduce unnecessary lookups
<details><summary>something like:</summary>

```py
    try:
        encoder = ENCODERS_BY_TYPE[obj.__class__]
    except KeyError:
        for base in obj.__class__.__mro__[1:-1]:
            encoder = ENCODERS_BY_TYPE.get(base):
            if encoder:
                return encoder(obj)
        else:
            raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
```
</details>

since `obj.__class__.__mro__[0] == type(obj)` and `obj.__class__.__mro__[-1] == object`, however, that may be unnecessary, and we can just trim the last element (`object`) from the loop with your suggestion:

```py
for base in obj.__class__.__mro__[:-1]:
    try:
        encoder = ENCODERS_BY_TYPE[base]
    except KeyError:
        continue
    return encoder(obj)
else:
    raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
```
(Still raising `TypeError` if we exit the for loop without finding anything)

I find the above try/except inside the loop to be fairly readable and only does one lookup as both @samuelcolvin and @dmontagu suggested (Thanks!)

> That said, @StephenBrown2 it might be good to see what the performance impact looks like for types that do have a longer inheritance chain leading back to something encodable.

Yes, though I wouldn't want to add something like pendulum or asyncpg or bson to the benchmarks just to test inheritability. Something like the [tests I added to this PR](https://github.com/samuelcolvin/pydantic/pull/1291/files#diff-0f8529efc62011b2f7a649151c0f6984) would be more like what I'd go for. I can however put together a gist to test and see how the inheritance chain affects things.

> I'm still a bit concerned about performance of this and think we should put the time into a small benchmark (or extension to the current benchmarks). Unless I'm missing something benchmarks/run.py never calls .json() so does nothing to measure the performance of this PR.

@dmontagu had mentioned testing "fastapi's `jsonable_encoder` when called on a relatively large list of dicts of int to int" in the other PR. Would that be a decent test case, or would calling `.json()` on the existing test cases be enough? It's coming from json for the benchmarks, so possibly modifying [generate_case](https://github.com/samuelcolvin/pydantic/blob/938335c46f7e31da71e5c2a45eb3cf1ad51030b5/benchmarks/run.py#L121) to add a couple inherited classes and running it through `.json()` instead of [json.dump](https://github.com/samuelcolvin/pydantic/blob/938335c46f7e31da71e5c2a45eb3cf1ad51030b5/benchmarks/run.py#L162)?

pydantic/json.py
StephenBrown2: Edited based on @samuelcolvin 's suggestion for the same, using KeyError like the prior logic.

samuelcolvin:
Regarding the loop let's go with

```py
for base in obj.__class__.__mro__[:-1]:
    try:
        f = ENCODERS_BY_TYPE[base]
    except KeyError:
        continue
    return f(obj)
```

for now, we can always optimise in future if we find a faster solution.

regarding benchmarking, let's start off with an extension of the current benchmarks that calls `.json()` and go from there.

StephenBrown2:
Went with your version of the loop and added the start of a json benchmark.

benchmarks/run.py
```diff
@@ -198,6 +188,26 @@ def main():
...
+    else:
+        with json_path.open() as f:
+            cases = json.load(f)
+
+    tests = [TestPydantic]
+    if 'pydantic-only' not in sys.argv:
+        tests += other_tests
+
+    repeats = int(os.getenv('BENCHMARK_REPEATS', '5'))
+    results, csv_results = run_tests(tests, cases, repeats, 'json' in sys.argv)
```
samuelcolvin:
```suggestion
    test_json = TEST_JSON in os.environ
    results, csv_results = run_tests(tests, cases, repeats, test_json)
```

That way:
1. it's consistent with `BENCHMARK_REPEATS`
2. you can alter it from outside the makefile.

docs/examples/exporting_models_json.py
```diff
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+import pendulum
```
samuelcolvin: I think too much extra confusion for those who haven't heard of pendulum to introduce another package for one test. Either find an example in another package we use or create a synthetic example.

You can add a note about pendulum, asyncpg etc. in text.

Makefile
```diff
@@ -68,6 +68,10 @@ benchmark-all:
 benchmark-pydantic:
 	python benchmarks/run.py pydantic-only
 
+.PHONY: benchmark-json
```
samuelcolvin: the problem with having these addition as part of this PR, is that we can't easily compare the performance with master before merging this PR.

Sorry to be so annoying, but would it be possible to move the json benchmarks to a new PR?

StephenBrown2: 👍 Sure thing.

StephenBrown2: Created https://github.com/samuelcolvin/pydantic/pull/1344 for just the benchmarks.

StephenBrown2:
Made a new PR and force-pushed this branch back a commit to avoid conflicts with travis.yml (which had been removed)

StephenBrown2:
So, some examples for subclassed types I've seen from other comments include:

- `pendulum.DateTime` -> `datetime.datetime`
- `ColorOriginal` -> `pydantic.Color`
- `asyncpg.pgproto.pgproto.UUID` -> `uuid.UUID`

Two out of three of these involve third-party packages, and one of them (asyncpg) is a cdef, but I suppose synthetic benchmarks could be made to imitate them.

pydantic/json.py
```diff
@@ -53,19 +53,26 @@ def pydantic_encoder(obj: Any) -> Any:
     elif is_dataclass(obj):
         return asdict(obj)
 
-    try:
-        encoder = ENCODERS_BY_TYPE[obj.__class__]
-    except KeyError:
-        raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
-    else:
+    # Check the class type and its superclasses for a matching encoder
```
samuelcolvin, line 56: I think with these changes `Path` and `Enum` above can be moved into `ENCODERS_BY_TYPE`

StephenBrown2: Done in https://github.com/samuelcolvin/pydantic/pull/1291/commits/3370971b79abe15ce09261b49d07875cac15e401

StephenBrown2:
Just a heads up, I should have some time tonight and Friday to work on finishing this up; will probably rebase/merge #1344 into it to see the Benchmarks for the changes.

StephenBrown2:
I've got this set up so far, just in case I don't get to this tonight like I hope:
```py
import datetime
import random
import string
import uuid

from devtools import debug
from pydantic.color import COLORS_BY_NAME, COLORS_BY_VALUE, Color

class SubStr(str):
    pass


class HexColor(Color):
    def __str__(self) -> str:
        return self.as_hex()


class ColorOriginal(Color):
    def __str__(self) -> str:
        return str(self.original())


class DateTime(datetime.datetime):
    def __str__(self):
        return self.isoformat("T")

    def __repr__(self):
        us = ""
        if self.microsecond:
            us = ", {}".format(self.microsecond)

        repr_ = "{klass}({year}, {month}, {day}, {hour}, {minute}, {second}{us}"

        if self.tzinfo is not None:
            repr_ += ", tzinfo={tzinfo}"

        repr_ += ")"

        return repr_.format(
            klass=self.__class__.__name__,
            year=self.year,
            month=self.month,
            day=self.day,
            hour=self.hour,
            minute=self.minute,
            second=self.second,
            us=us,
            tzinfo=self.tzinfo,
        )


# https://github.com/tiangolo/fastapi/pull/756#issuecomment-572251121
class MyUuid:
    def __init__(self, uuid_string: str):
        self.uuid = uuid_string

    def __str__(self):
        return self.uuid

    def __repr__(self):
        return f"{self.__class__.__name__}({self.uuid})"

    @property
    def __class__(self):
        return uuid.UUID

    @property
    def __dict__(self):
        """Spoof a missing __dict__ by raising TypeError, this is how
        asyncpg.pgroto.pgproto.UUID behaves"""
        raise TypeError("vars() argument must have __dict__ attribute")


class ValidatedUuid(str):
    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, v):
        if not isinstance(v, str):
            raise TypeError('string required')
        return MyUuid(v)

class SomeCustomClass(BaseModel):
    a_uuid: ValidatedUuid
    sub_str: SubStr
    hex_color: HexColor
    orig_color: ColorOriginal
    date_time: DateTime

    class Config:
        json_encoders = {
            MyUuid: str,
            SubStr: str,
            HexColor: str,
            ColorOriginal: str,
            DateTime: str,
        }


PUNCTUATION = " \t\n!\"#$%&'()*+,-./"
LETTERS = string.ascii_letters
UNICODE = "\xa0\xad¡¢£¤¥¦§¨©ª«¬ ®¯°±²³´µ¶·¸¹º»¼½¾¿ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ"
ALL = PUNCTUATION * 5 + LETTERS * 20 + UNICODE


def rand_string(min_length, max_length, corpus=ALL):
    return "".join(random.choices(corpus, k=random.randrange(min_length, max_length)))


def rand_date():
    r = random.randrange
    return f"{r(1900, 2020)}-{r(0, 12)}-{r(0, 32)}T{r(0, 24)}:{r(0, 60)}:{r(0, 60)}"


def generate_case():
    return dict(
        a_uuid=str(uuid.uuid4()),
        sub_str=rand_string(5, 20),
        hex_color=random.choice(list(COLORS_BY_NAME.keys()) + list(COLORS_BY_VALUE.keys())),
        orig_color=random.choice(list(COLORS_BY_NAME.keys()) + list(COLORS_BY_VALUE.keys())),
        date_time=rand_date(),
    )


cases = [generate_case() for _ in range(20)]

debug(cases)

models = []
for case in cases:
    try:
        models.append(SomeCustomClass(**case))
    except ValidationError:
        pass

debug(models)

for model in models:
    debug(model.json())
```

pydantic/json.py
```diff
@@ -18,25 +18,27 @@ def isoformat(o: Union[datetime.date, datetime.time]) -> str:
 
 
 ENCODERS_BY_TYPE: Dict[Type[Any], Callable[[Any], Any]] = {
+    bytes: lambda o: o.decode(),
```
samuelcolvin, line 21: I'm not clear why we're bothering to re-order this dict?

samuelcolvin:
this is awesome. Thanks so much and sorry again for the delay.

##🥳 🍾 🎉

pydantic/json.py
StephenBrown2: It's a dict, so order doesn't matter, especially since we're `KeyError`ing, and sorted items make it easier to add more later. I sorted it when I moved `Path` and `Enum`.

StephenBrown2:
Thanks to you, for taking it over the finish line. Life has taken a bit of a busy turn so I haven't had as much time for open-source contributions like I would like.

samuelcolvin:
No problem, thank you.

PR #1293, opened 2020-03-05
Add MacOS and Windows build targets

### LINKED ISSUE #555: MacOS and windows pre-compiles binaries
pydantic will soon be compiled with cython, however pre-compiled binaries are only available on linux at the moment.

It should be possible for macos on travis and windows on azure pipeline.

However since I only use linux, it's not my primary concern.

I'd be happy to accept a PR if someone is prepared to implement one or both.

StephenBrown2:
## Change Summary

As of about a week ago, the cibuildwheel repository has new examples for Travis on Mac and Windows: https://github.com/joerick/cibuildwheel/blob/master/examples/travis-ci-test-and-deploy.yml

This adds those targets to .travis.yml to hopefully resolve #555 .

If this passes (which I can't test myself, since you should handle the releases), I can edit https://github.com/samuelcolvin/pydantic/blob/8727914ced9d4546f0f95a453aff78f03b778275/docs/install.md#L19-L21 to reflect it.

## Related issue number

#555 

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,18 @@
 import os
 import re
 import sys
+from distutils.command import build_ext
 from importlib.machinery import SourceFileLoader
 from pathlib import Path
 
 from setuptools import setup
+
+if os.name == 'nt':
+
+    def pass_ges(self, ext):
+        pass
+
+    build_ext.get_export_symbols = pass_ges
 
 
 class ReplaceLinks:
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ if os.name == 'nt':
         except UnicodeEncodeError:
             suffix = 'U' + suffix.encode('punycode').replace(b'-', b'_').decode('ascii')
 
-        initfunc_name = "PyInit_" + suffix
+        initfunc_name = 'PyInit_' + suffix
         if initfunc_name not in ext.export_symbols:
             ext.export_symbols.append(initfunc_name)
         return ext.export_symbols
--- a/setup.py
+++ b/setup.py
@@ -1,19 +1,46 @@
 import os
 import re
 import sys
-from distutils.command import build_ext
 from importlib.machinery import SourceFileLoader
 from pathlib import Path
 
 from setuptools import setup
 
 if os.name == 'nt':
+    from distutils.command import build_ext
 
-    def pass_ges(self, ext):
-        pass
+    def get_export_symbols(self, ext):
+        """Return the list of symbols that a shared extension has to
+        export.  This either uses 'ext.export_symbols' or, if it's not
+        provided, "PyInit_" + module_name.  Only relevant on Windows, where
+        the .pyd file (DLL) must export the module "PyInit_" function.
 
-    build_ext.get_export_symbols = pass_ges
+        Source: https://github.com/python/cpython/blob/8849e5962ba481d5d414b3467a256aba2134b4da\
+                /Lib/distutils/command/build_ext.py#L686-L703
+        """
+        # Patch from: https://bugs.python.org/issue35893
+        parts = ext.name.split('.')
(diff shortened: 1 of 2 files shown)


samuelcolvin:
Humm, I'd rather move all this to github actions, see [rtoml](https://github.com/samuelcolvin/rtoml) for a more complex but similar case.

Still until we move to github actions completely, maybe this would be a good stop gap.

samuelcolvin:
please confirm that these steps are working by commenting out the `if: type = push AND (branch = master OR tag IS present)` and `deploy:` chunks.

StephenBrown2:
Well, MacOS is building, but I can't figure out the Windows errors at the moment.  I will happily leave those commented out for someone else to tackle, if you wish.

demospace:
> Well, MacOS is building, but I can't figure out the Windows errors at the moment. I will happily leave those commented out for someone else to tackle, if you wish.

I patched pydantic in a few places recently and tried to compile in Windows and ran into the following:

`\temp.win-amd64-3.7\Release\pydantic\__init__.cp37-win_amd64.lib
LINK : error LNK2001: unresolved external symbol PyInit___init__
build\temp.win-amd64-3.7\Release\pydantic\__init__.cp37-win_amd64.lib : fatal error LNK1120: 1 unresolved externals
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.16.27023\\bin\\HostX86\\x64\\link.exe' failed with exit status 1120
`
The issue relates to distutils/setuptools and you can read more about it here:
 https://bugs.python.org/issue35893

The last comment in that thread, when added to pydantic's setup.py does resolve the issue:
`from distutils.command import build_ext

def get_export_symbols(self, ext):
    parts = ext.name.split(".")
    print('parts', parts)
    if parts[-1] == "__init__":
        initfunc_name = "PyInit_" + parts[-2]
    else:
        initfunc_name = "PyInit_" + parts[-1]`

StephenBrown2:
In theory this could break non-windows? So best to include it in an `os.name` check, like:
```py
import os
from distutils.command import build_ext

if os.name == 'nt':

    def pass_ges(self, ext):
        pass

    build_ext.get_export_symbols = pass_ges
```

samuelcolvin:
> In theory this could break non-windows? So best to include it in an `os.name` check, like:
> 
> ```python
> import os
> from distutils.command import build_ext
> 
> if os.name == 'nt':
> 
>     def pass_ges(self, ext):
>         pass
> 
>     build_ext.get_export_symbols = pass_ges
> ```

agreed, let's try this.

StephenBrown2:
Doesn't looke like it's fixed that issue:
```
  LINK : error LNK2001: unresolved external symbol PyInit___init__
```
still happening.

setup.py
```diff
@@ -1,11 +1,19 @@
...
 from pathlib import Path
 
 from setuptools import setup
 
+if os.name == 'nt':
+
+    def pass_ges(self, ext):
+        pass
+
+    build_ext.get_export_symbols = pass_ges
```
samuelcolvin: according to https://bugs.python.org/issue35893 this should be 

```py
build_ext.build_ext.get_export_symbols = ...
```

Also they have a different function for `get_export_symbols`.

StephenBrown2: Good point, I didn't look at the bug. The Stackoverflow answer is what suggested the pass (as the "simplest" option), but I'll try the solution in the bug.

StephenBrown2:
Well it looks like that indeed worked. I took the function from cpython master and the patch from the bug report, and Windows builds are now succeeding!

samuelcolvin:
incorporated into #1326, thank you so much.

PR #1301, opened 2020-03-11
Add datetime field to default_factory example

StephenBrown2:
## Change Summary

While looking through the docs, I was happy to find that `default_factory` will be added to 1.5 provisionally, however, I was specifically searching for an example where `datetime` was being used, and my searches came up fruitless until I checked the closed issues. This adds an example for searchability purposes (and works as expected in my tests)

## Related issue number

None

## Checklist

### DIFF
--- a/docs/examples/models_default_factory.py
+++ b/docs/examples/models_default_factory.py
@@ -1,9 +1,12 @@
-from pydantic import BaseModel, Field
+from datetime import datetime
 from uuid import UUID, uuid4
+from pydantic import BaseModel, Field
 
 class Model(BaseModel):
     uid: UUID = Field(default_factory=uuid4)
+    updated: datetime = Field(default_factory=datetime.utcnow)
 
 m1 = Model()
 m2 = Model()
-assert m1.uid != m2.uid
+print(f'{m1.uid} != {m2.uid}')
+print(f'{m1.updated} != {m2.updated}')
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,8 +8,9 @@ filterwarnings =
 [flake8]
 max-line-length = 120
 max-complexity = 14
-inline-quotes = '
-multiline-quotes = """
+inline-quotes = single
+multiline-quotes = double
+avoid-escape = False
 ignore = E203, W503
 
 [bdist_wheel]
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,7 +10,6 @@ max-line-length = 120
 max-complexity = 14
 inline-quotes = single
 multiline-quotes = double
-avoid-escape = False
 ignore = E203, W503
 
 [bdist_wheel]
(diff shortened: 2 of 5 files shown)


StephenBrown2:
Updated page: https://pydantic.helpmanual.io/dev/pr-1301/usage/models/#field-with-dynamic-default-value

docs/examples/models_default_factory.py
```diff
@@ -1,9 +1,12 @@
...
+from pydantic import BaseModel, Field
 
 class Model(BaseModel):
     uid: UUID = Field(default_factory=uuid4)
+    updated: datetime = Field(default_factory=datetime.utcnow)
 
 m1 = Model()
 m2 = Model()
 assert m1.uid != m2.uid
+assert m1.updated != m2.updated
```
samuelcolvin: better to print here than do an assert, could you change the case above too?

StephenBrown2: Sure thing.

StephenBrown2:
Changed the asserts to prints, left the `!=` in the string to show them on the same line in a self-explanatory manner. Let me know if you'd prefer them on separate lines or otherwise.

samuelcolvin:
single quotes please :smile:

StephenBrown2:
Removed bad quotes, but I noticed that it would have been checked if the lint step included the docs/ folder.

I did so and found that there are a bunch of lints (including other double quote usage) in the examples, shall I open another PR to fix those?

samuelcolvin:
up to you, if it's easy you could fix it here, otherwise leave the linting unchanged, and I'll do it elsewhere.

StephenBrown2:
Fairly easy. I added a separate isort/black_docs command in the makefile, to apply an 80-character limit to docs only, matching most of the current formatting, and better for the view inside mkdocs.

I did them in separate commits for easy reversion if desired, and it looks like I needed to allow two newlines in the exec_examples script (Fabulous bit of code, that) to let it build properly.

StephenBrown2:
Not sure why this is failing on Windows 3.7, when it says python 3.8 only and has passed before...

StephenBrown2:
Looks like it's a bit more tricky than I anticipated, so I'll revert it for now and leave it for later.

samuelcolvin:
Great, I think the reason I didn't do linting was that it wasn't trivial. I'll fix soon.

setup.cfg
```diff
@@ -8,8 +8,9 @@ filterwarnings =
 [flake8]
 max-line-length = 120
 max-complexity = 14
-inline-quotes = '
-multiline-quotes = """
+inline-quotes = single
+multiline-quotes = double
```
StephenBrown2: Changed these to words to be more clear, and to not break syntax highlighting.

setup.cfg
```diff
@@ -8,8 +8,9 @@ filterwarnings =
...
+avoid-escape = False
```
StephenBrown2: Disable avoiding escaping same quotes in inline strings.
https://github.com/zheller/flake8-quotes#caveats

samuelcolvin: I think this is the default, so not necessary. I generally try to avoid unnecessary settings.

PR #1302, opened 2020-03-11
Made user info not required for RedisDSN.  This will fix the issue #1275

### LINKED ISSUE #1275: RedisDSN with out username
# Question

```
             pydantic version: 1.4
            pydantic compiled: True
                 install path: /home/kesav/.virtualenvs/smtp/lib/python3.8/site-packages/pydantic
               python version: 3.8.0 (default, Oct 28 2019, 16:14:01)  [GCC 9.2.1 20191008]
                     platform: Linux-5.3.0-40-generic-x86_64-with-glibc2.29
     optional deps. installed: []
```

```py
import pydantic

class AppSettings(pydantic.BaseSettings):
   redis_dsn: pydantic.RedisDsn = 'redis://:password@integration.com'

```
My redis server has no user name only password it has.  If I use above setting I get the following error:

```
  File "pydantic/env_settings.py", line 28, in pydantic.env_settings.BaseSettings.__init__
  File "pydantic/main.py", line 283, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for Test
redis_dsn
  userinfo required in URL but missing (type=value_error.url.userinfo)
```

How to use RedisDsn without user name?

kesavkolla:
## Change Summary

Making user info optional for RedisDSN.  Most of the redis instances will have either no authentication or only password is needed.

## Related issue number

Fix #1275 

## Checklist

* [x] `changes/1275-kesavkolla.md` file added describing change


changes/1275-kesavkolla.md
```diff
@@ -0,0 +1 @@
+Made user info optional for RedisDSN
```
samuelcolvin:
```suggestion
Made user info optional for `RedisDSN`
```

samuelcolvin:
@kesavkolla any chance of this getting fixed?

kesavkolla:
sorry I didn't understand the comment.  I thought it's waiting on you to merge

samuelcolvin:
no, I've made suggestions, and asked for some more changes.

kesavkolla:
> should we modify `user_required` to be satisfied by `user` or `password` rather than just `user`?

I think that is a good idea of chanding the behaviour of `user_required`.  Sometimes user is is there password is not there other times both exist.  So we should really make the interpreation of user_required to any of user or password is present.

kesavkolla:
@samuelcolvin anything else to do for this PR?

pydantic/networks.py
```diff
@@ -189,7 +190,8 @@ def validate(cls, value: Any, field: 'ModelField', config: 'BaseConfig') -> 'Any
             raise errors.UrlSchemePermittedError(cls.allowed_schemes)
 
         user = parts['user']
-        if cls.user_required and user is None:
+        password = parts['password']
+        if cls.user_required and user is None and password is None:
```
samuelcolvin, line 194: Humm, this is a breaking change (and arguably a bug) since now user is no longer required when `user_required = True` - just password will suffice .

To fix this, either we need to:
1. add `password_required` (and then probably `user_or_password_required`, maybe `credentials_required` would be a better name)
2. or, leave this logic unchanged and just change `user_required = False` on `RedisDsn`

I think I favour option 1., but on `RedisDsn` it looks like neither user or password are required, so it doesn't make much difference in this case.

pydantic/networks.py
```diff
@@ -53,7 +53,7 @@ def url_regex() -> Pattern[str]:
     if _url_regex_cache is None:
         _url_regex_cache = re.compile(
             r'(?:(?P<scheme>[a-z][a-z0-9+\-.]+)://)?'  # scheme https://tools.ietf.org/html/rfc3986#appendix-A
-            r'(?:(?P<user>[^\s:/]+)(?::(?P<password>[^\s/]*))?@)?'  # user info
+            r'(?:(?P<user>[^\s:/]+)?(?::(?P<password>[^\s/]*))?@)?'  # user info
```
samuelcolvin, line 56:
```suggestion
            r'(?:(?P<user>[^\s:/]+)?'  # user
            r'(?::(?P<password>[^\s/]*))?@)?'  # password
```
as i understand it user and password are now independent, so should go on separate lines.

changes/1275-kesavkolla.md
```diff
@@ -0,0 +1 @@
+Made user info optional for `RedisDSN`
```
samuelcolvin, line 1: If we're adding `password_required` and `credentials_required` we should put that in the change description.

samuelcolvin:
Thanks so much for working on this.

In the end to get something out, I've implemented a simple fix in #1658.

PR #1306, opened 2020-03-12
Allow schema_extra to be a classmethod

### LINKED ISSUE #1305: schema_extra should be allowed to be a classmethod
In my opinion, this check is unnecessarily restrictive:
https://github.com/samuelcolvin/pydantic/blob/938335c46f7e31da71e5c2a45eb3cf1ad51030b5/pydantic/schema.py#L470-L471
Also, it raises `TypeError` only for classmethods, but not for regular methods.

laevilgenius:
## Change Summary
Allow `schema_extra` to be a classmethod.

## Related issue number
Resolves #1305

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -21,6 +21,7 @@ from typing import (
     Union,
     cast,
     no_type_check,
+    overload,
 )
 
 from .class_validators import ROOT_KEY, ValidatorGroup, extract_root_validators, extract_validators, inherit_validators
@@ -44,6 +45,7 @@ from .utils import (
 
 if TYPE_CHECKING:
     from inspect import Signature
+    import typing_extensions
     from .class_validators import ValidatorListDict
     from .types import ModelOrDc
     from .typing import CallableGenerator, TupleGenerator, DictStrAny, DictAny, SetStr
@@ -51,6 +53,19 @@ if TYPE_CHECKING:
 
     ConfigType = Type['BaseConfig']
     Model = TypeVar('Model', bound='BaseModel')
+
+    class SchemaExtraCallable(typing_extensions.Protocol):
+        @overload
+        def __call__(self, schema: Dict[str, Any]) -> None:
+            pass
+
+        @overload
+        def __call__(self, schema: Dict[str, Any], model: Type['Model']) -> None:
+            pass
+
+        def __call__(self, schema: Dict[str, Any], model: Type['Model'] = None) -> None:
+            pass
+
 
 try:
     import cython  # type: ignore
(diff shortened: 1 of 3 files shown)


laevilgenius:
I find it far easier to remove the check altogether, than tinker with method inspection on class object. Not sure what is the best way to make mypy happy though.
@samuelcolvin wdyt?

tests/test_schema.py
```diff
@@ -1534,20 +1534,6 @@ def schema_extra(schema):
     assert Model.schema() == {'title': 'Model', 'type': 'override'}
 
 
-def test_model_with_schema_extra_callable_classmethod_asserts():
```
samuelcolvin: let's keep this test but demonstrate that it works.

laevilgenius: Done.

laevilgenius:
So, I've added a test for instance methods, but I'm not sure how useful it is.

Lint is failing due to obsolete pyflakes version (fixed in master, but not yet released on pypi). How should we proceed?

pydantic/main.py
```diff
@@ -52,6 +54,19 @@
     ConfigType = Type['BaseConfig']
     Model = TypeVar('Model', bound='BaseModel')
 
+    class SchemaExtraCallable(typing_extensions.Protocol):
+        @overload
+        def __call__(self, schema: Dict[str, Any]) -> None:
+            pass
+
+        @overload
+        def __call__(self, schema: Dict[str, Any], model: Type['Model']) -> None:
```
samuelcolvin: Surely you don't need both this case and, the one below?

Better to use `model_class` than `model` to be clear.

You need to add `. # noqa F811` to get linting to pass, see other examples in the code base

pydantic/main.py
```diff
@@ -89,7 +104,7 @@ class BaseConfig:
     getter_dict: Type[GetterDict] = GetterDict
     alias_generator: Optional[Callable[[str], str]] = None
     keep_untouched: Tuple[type, ...] = ()
-    schema_extra: Union[Dict[str, Any], Callable[[Dict[str, Any]], None]] = {}
+    schema_extra: Union[Dict[str, Any], 'SchemaExtraCallable'] = {}
```
samuelcolvin: maybe you could test this works by adding a to `tests/mypy/modules/success.py`?

laevilgenius: I could use some help here. I've been trying something like this to pass
```py
class ConfigSchemaExtraCallable(BaseConfig['MyModel']):
    @staticmethod
    def schema_extra(schema: Dict[str, Any], model_class: Type['MyModel']) -> None:
        pass
```
to no avail.
Even when I made `BaseConfig` generic, mypy still complained about signature incompatibility.

pydantic/main.py
laevilgenius: Done.

samuelcolvin:
spent ages trying to improve on this, but nothing seems to work.

Let's go with this.

PR #1307, opened 2020-03-13
accept None as value for all optional fields

### LINKED ISSUE #1295: When using `min_items` and `max_items` pydantic cannot accept `None`
# Bug

```
             pydantic version: 1.4
            pydantic compiled: False
                 install path: /Users/spike/.pyenv/versions/3.8.2/lib/python3.8/site-packages/pydantic
               python version: 3.8.2 (default, Feb 26 2020, 16:45:04)  [Clang 11.0.0 (clang-1100.0.33.17)]
                     platform: macOS-10.15.3-x86_64-i386-64bit
     optional deps. installed: []

```

```py
>>> class Foo(BaseModel):
...     bar: List = Field(None)
...     baz: List = Field(None, max_items=10)
...
>>> Foo()
Foo(bar=None, baz=None)
>>> Foo(bar=None)
Foo(bar=None, baz=None)
>>> Foo(baz=None)
  File "/Users/spike/.pyenv/versions/3.8.2/lib/python3.8/site-packages/pydantic/main.py", line 283, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for Foo
baz
  value is not a valid list (type=type_error.list)
```

It is not what I expected.

Also when using `min_items`

```py
>>> class Foo(BaseModel):
...     bar: List = Field(None, min_items=0)
...
>>> Foo()
Foo(bar=None)
>>> Foo(bar=None)
  File "/Users/spike/.pyenv/versions/3.8.2/lib/python3.8/site-packages/pydantic/main.py", line 283, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for Foo
bar
  value is not a valid list (type=type_error.list)
```

I think both `min_items` and `max_items` expect that an attribute must iterable. It is the same when using `Optional`.

PrettyWood:
## Change Summary

All optional list fields with `None` as value should be valid except if there is a custom validation
to handle this case.

ℹ️  **Test skipped on python 3.6**
```
        if unused_constraints:
            raise ValueError(
>               f'On field "{field_name}" the following field constraints are set but not enforced: '
                f'{", ".join(unused_constraints)}. '
                f'\nFor more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints'
            )
E           ValueError: On field "bar" the following field constraints are set but not enforced: min_items. 
E           For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints
```

## Related issue number

closes #1295

## Checklist

### DIFF
--- a/changes/1307-prettywood.md
+++ b/changes/1307-prettywood.md
@@ -1,0 +1,1 @@
+Allow `None` as input to all optional list fields
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -81,6 +81,7 @@ if TYPE_CHECKING:
     from .dataclasses import DataclassType  # noqa: F401
     from .main import BaseModel, BaseConfig  # noqa: F401
     from .typing import CallableGenerator
+    from .fields import ModelField
 
     ModelOrDc = Type[Union['BaseModel', 'DataclassType']]
 
@@ -129,7 +130,7 @@ class ConstrainedList(list):  # type: ignore
         update_not_none(field_schema, minItems=cls.min_items, maxItems=cls.max_items)
 
     @classmethod
-    def list_length_validator(cls, v: 'Optional[List[T]]', field: 'ModelField') -> 'Optional[List[T]]':  # type: ignore[name-defined] # noqa: E501,F821
+    def list_length_validator(cls, v: 'Optional[List[T]]', field: 'ModelField') -> 'Optional[List[T]]':
         if v is None and not field.required:
             return None
 
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -122,7 +122,6 @@ class ConstrainedList(list):  # type: ignore
 
     @classmethod
     def __get_validators__(cls) -> 'CallableGenerator':
-        yield list_validator
         yield cls.list_length_validator
 
     @classmethod
@@ -130,7 +129,11 @@ class ConstrainedList(list):  # type: ignore
         update_not_none(field_schema, minItems=cls.min_items, maxItems=cls.max_items)
 
     @classmethod
-    def list_length_validator(cls, v: 'List[T]') -> 'List[T]':
+    def list_length_validator(cls, v: 'Optional[List[T]]', field: 'ModelField') -> 'Optional[List[T]]':  # type: ignore[name-defined] # noqa: E501,F821
+        if v is None and not field.required:
+            return None
+
+        v = list_validator(v)
         v_len = len(v)
 
         if cls.min_items is not None and v_len < cls.min_items:
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,3 +1,4 @@
+import sys
 from enum import Enum
 from typing import Any, Callable, ClassVar, List, Mapping, Type
 from uuid import UUID, uuid4
@@ -1027,6 +1028,7 @@ def test_default_factory():
     assert m.uid is uuid4
 
 
+@pytest.mark.skipif(sys.version_info < (3, 7), reason='field constraints are set but not enforced with python 3.6')
 def test_none_min_max_items():
     # None default
     class Foo(BaseModel):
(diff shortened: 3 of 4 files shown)


pydantic/fields.py
```diff
@@ -705,6 +705,10 @@ def _apply_validators(
     ) -> 'ValidateReturn':
         for validator in validators:
             try:
+                # optional field set to None should be valid unless a custom validator handles this case
+                if validator.__name__ in DEFAULT_VALIDATORS_NAMES and v is None and not self.required:
```
PrettyWood: I'm not really satisfied with the check on name. But the validators are wrapped so no easy way to directly check that. I'm open to proposals

samuelcolvin: Instead modify `ConstrainedList` to something like

```py
# This types superclass should be List[T], but cython chokes on that...
class ConstrainedList(list):  # type: ignore
    # Needed for pydantic to detect that this is a list
    __origin__ = list
    __args__: List[Type[T]]  # type: ignore

    min_items: Optional[int] = None
    max_items: Optional[int] = None
    item_type: Type[T]  # type: ignore

    @classmethod
        yield cls.list_length_validator

    @classmethod
        update_not_none(field_schema, minItems=cls.min_items, maxItems=cls.max_items)

    @classmethod
    def list_length_validator(cls, v: 'List[T]', field) -> Optional['List[T]']:
        if v is None and not field.required:
            return None

        v = list_validator(v)
        v_len = len(v)

        if cls.min_items is not None and v_len < cls.min_items:
            raise errors.ListMinLengthError(limit_value=cls.min_items)

        if cls.max_items is not None and v_len > cls.max_items:
            raise errors.ListMaxLengthError(limit_value=cls.max_items)

        return v
```

tests/test_main.py
```diff
@@ -1025,3 +1026,21 @@ class FunctionModel(BaseModel):
 
     m = FunctionModel()
     assert m.uid is uuid4
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason='field constraints are set but not enforced with 3.6 and above')
```
samuelcolvin: surely this can be removed then?

pydantic/fields.py
PrettyWood: > I'm afraid I'm not happy with this. This is the most performance critical part of the whole of pydantic.
> 
> I think we can fix this bug without modifying this funciton.

I definitely understand ! I was not satisfied either. But I wanted to tackle the whole problem instead of just the ConstrainList one. As a matter of fact it seems it's not needed! Sorry for the overkill

changes/1307-prettywood.md
```diff
@@ -0,0 +1 @@
+fix : value None should be valid for all optional fields by default
```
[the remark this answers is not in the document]
PrettyWood: Just added `list fields` as the fix is scoped to ConstrainedList

pydantic/types.py
```diff
@@ -122,15 +122,18 @@ class ConstrainedList(list):  # type: ignore
...
-        yield list_validator
         yield cls.list_length_validator
 
     @classmethod
     def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
         update_not_none(field_schema, minItems=cls.min_items, maxItems=cls.max_items)
 
     @classmethod
-    def list_length_validator(cls, v: 'List[T]') -> 'List[T]':
+    def list_length_validator(cls, v: 'Optional[List[T]]', field: 'ModelField') -> 'Optional[List[T]]':  # type: ignore[name-defined] # noqa: E501,F821
```
PrettyWood: @samuelcolvin Thank you for the review ! Just a question on the mypy policy as I could
- not set the type of `field` and just add `# type: ignore`
- same but explaining the ignored error but then add a `noqa` for length
- do some refactoring to avoid circular imports (overkill probably)
- let it like that

tests/test_main.py
PrettyWood: As a matter of fact it seems it can't :/

pydantic/types.py
samuelcolvin: you just needed to import `ModelField`, imports inside `if TYPE_CHECKING:` don't cause circular import problems.

Issue #596, opened 2019-06-17
Configurable SecretStr


jasonkuhrt:
This issue revisits this comment https://github.com/samuelcolvin/pydantic/issues/462#issuecomment-480326378 by @tiangolo.

I think it would be a good idea that there is a standard way of getting secrets exported for propagation to other services.

This is not something that tends to happen in three tier apps where e.g. the db creds are secret but rather a case that happens _a lot_ when dealing with microservice architectures wherein request payloads may serialize and deserialize multiple times through the end to end request lifecycle.

`.json()` to me is semantically like `.export` and as such defaulting to revealing secret makes sense. But that would be a breaking change.

Other approaches:

- `.json(reveal_secrets=True)`
- `.export()`

But maybe we can take the breaking change path via https://github.com/samuelcolvin/pydantic/issues/576 and then:

- `.json(keep_secrets=True)`

To be clear I don't see `.json` as being something used for logging. Something like `structlog` would work with `pydantic.dict()` instead:

```
log.info('something', data=model.dict())
```

I _think_ `.dict` defaulting to maintaining secrets seems right. But we could have, too:

```
log.info('something', data=model.dict(reveal_secrets=True))
```

But than we should make considerations around API consistency across methods and ensure usability is good overall, not just per case.

samuelcolvin:
I think this is unnecessary.

The purpose of `SecretStr` is questionable anyway, or rather, it has relatively few useful applications. The environment in which you're processing "secret" values has to be trusted, so why not just use `str`?

Given this I think it's overkill to add another argument to `dict()` (and associated methods) and then pass the setting around recursively.

You can accomplish the same thing with your own custom encoder very easily:

```py
from pydantic import BaseModel, SecretStr
from pydantic.json import pydantic_encoder

def custom_encoder(obj):
    if isinstance(obj, SecretStr):
        return obj.get_secret_value()
    else:
        return pydantic_encoder(obj)

class Model(BaseModel):
    password: SecretStr = None

    def json(self) -> str:
        return super().json(encoder=custom_encoder)

print(Model(password='testing').json())
```

If you really wanted you could use your own custom `BaseModel` that adds another keyword argument `reveal_secrets=False` to `json()` and uses the above encoder if `reveal_secrets=True`.

jasonkuhrt:
> The environment in which you're processing "secret" values has to be trusted, so why not just use str

[...]

PR #1313, opened 2020-03-15
add SecretStr and SecretBytes to the default json encoder.

Atheuz:
## Change Summary

* add SecretStr and SecretBytes to the default json encoder.

#596 discusses making SecretStr, SecretBytes .json() dumpable such that we get the proper value out of it when calling it.
This PR changes the default json encoder such that it properly output SecretStr and SecretBytes.
Further also updates the documentation.

## Related issue number

#596 

## Checklist

* [X] `changes/1313-atheuz.md` file added describing change

### DIFF
--- a/changes/1313-atheuz.md
+++ b/changes/1313-atheuz.md
@@ -1,0 +1,1 @@
+add SecretStr and SecretBytes to the default json encoder.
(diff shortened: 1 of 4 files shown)


samuelcolvin:
Sorry if I wasn't clear or contradicted myself on the issue. `SecretStr` and `SecretBytes` definitely shouldn't be shown by default in JSON.

I think we should stick to the current status quo:
* either, use `json_encoders`
* or, use a `custom_encoder`
* or, wait for better custom serialisation when I get around to it
* or, allow some way of customising `SecretStr` on the field definition - this needs more discussion since none of the existing solutions look good

PR #1319, opened 2020-03-17
Have BaseModel inherit from Representation

FuegoFro:
This changes the inheritance so that `BaseModel` *actually* inherits from `Representation`, rather than just having the same methods as it. While this doesn't actually affect the runtime, it does make mypy happier.

Fixes https://github.com/samuelcolvin/pydantic/issues/1310

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -172,6 +172,12 @@ def validate_custom_root_type(fields: Dict[str, ModelField]) -> None:
 
 UNTOUCHED_TYPES = FunctionType, property, type, classmethod, staticmethod
 
+# Note `ModelMetaclass` refers to `BaseModel`, but is also used to *create* `BaseModel`, so we need to add this extra
+# (somewhat hacky) boolean to keep track of whether we've created the `BaseModel` class yet, and therefore whether it's
+# safe to refer to it. If it *hasn't* been created, we assume that the `__new__` call we're in the middle of is for
+# the `BaseModel` class, since that's defined immediately after the metaclass.
+_is_base_model_class_defined = False
+
 
 class ModelMetaclass(ABCMeta):
     @no_type_check  # noqa C901
@@ -183,7 +189,7 @@ class ModelMetaclass(ABCMeta):
 
         pre_root_validators, post_root_validators = [], []
         for base in reversed(bases):
-            if issubclass(base, BaseModel) and base != BaseModel:
+            if _is_base_model_class_defined and issubclass(base, BaseModel) and base != BaseModel:
                 fields.update(deepcopy(base.__fields__))
                 config = inherit_config(base.__config__, config)
                 validators = inherit_validators(base.__validators__, validators)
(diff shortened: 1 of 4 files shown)


FuegoFro:
For

> Documentation reflects the changes where applicable

I wasn't sure if this required any documentation changes. I didn't think so, but figured it'd be better to check in before assuming so 🙂

pydantic/main.py
```diff
@@ -301,13 +307,8 @@ class BaseModel(metaclass=ModelMetaclass):
...
 
     Config = BaseConfig
-    __slots__ = ('__dict__', '__fields_set__')
-    # equivalent of inheriting from Representation
-    __repr_name__ = Representation.__repr_name__
-    __repr_str__ = Representation.__repr_str__
-    __pretty__ = Representation.__pretty__
-    __str__ = Representation.__str__
-    __repr__ = Representation.__repr__
+    __slots__ = ('__fields_set__',)
```
samuelcolvin: please don't change this, if you think it needs changing, please create another issue.

pydantic/main.py
```diff
@@ -301,13 +307,8 @@ class BaseModel(metaclass=ModelMetaclass):
...
+    __doc__ = ''  # Null out the Representation docstring
```
samuelcolvin: can we have a proper docstring, maybe as simple as "Base class for creating pydantic models.".

Also, can this be done via a proper docstring rather than `__doc__ = '..`?

FuegoFro: It's unclear that we want a real docstring here, since it gets inherited by subclasses. In particular, this line was added because tests were failing after classes started getting errant descriptions on them. Eg, [`test_schema`](http://github.com/FuegoFro/pydantic/blob/e67e1fe3a24aaea0a158557529711540a8061252/tests/test_dataclasses.py#L431) would need an added `'description': 'Base class for creating pydantic models.',` key, and that seems... incorrect?

I used `__doc__` rather than a normal docstring because it seemed clearer given that we're not setting a normal docstring, but rather removing one (which is a fairly unusual operation). If we did want to null out `Representation`'s docstring (for above reasons) without using `__doc__` we could do something like

```python
class BaseModel(Representation, metaclass=ModelMetaclass):
    """"""  # Empty docstring to null out the docstring from Representation
    ...
```
but I thought that might be more weird? Up to you though 🙂

pydantic/main.py
FuegoFro: It was complaining because `Representation` already had `'__dict__'` in its `__slots__` (not allowed to have the same slot twice in the inheritance hierarchy), but now that you point this out I think the better solution is to set `__slots__ = tuple()` in `Representation`, then we can keep `'__dict__'` here 🙂

FuegoFro:
Hey! Did you have thoughts on the `BaseModel` docstring? Was there anything else you wanted to see changed before this is ready? 🙂

FuegoFro:
Thank you for the merge! 😀

PR #1321, opened 2020-03-17
fix card number length check in PaymentCardNumber

### LINKED ISSUE #1317: Invalid `is` check in credit card validation
https://github.com/samuelcolvin/pydantic/blob/master/pydantic/types.py#L669

```python
if card_number.brand is (PaymentCardBrand.visa or PaymentCardBrand.mastercard)
```
Assuming brand was reduced to the integers, and we had 3 brands, then this is the equivalent of:
`3 is (1 or 2)`, which you'd hope were false, `1 is (1 or 2)` which you'd hope were true, and `2 is (1 or 2)` which you'd also hope were true. However, `or` short-circuits, so these reduce to `3 is 1`, `1 is 1`, and `2 is 1`. This means the preferred outcome for the first two cases works, but it would fail for mastercard.

Perhaps this should be
```python
if card_number.brand is PaymentCardBrand.visa or card_number.brand is PaymentCardBrand.mastercard:
```

I think this was added relatively recently - https://github.com/samuelcolvin/pydantic/commit/0c5793770be1de8b1a28c3fbaca669cf3bab1e34#diff-470d0730db2e82eaaf27ad3655a667b3R582 - so hopefully it hasn't affected too many people.

samuelcolvin:
## Change Summary

fix card number length check in `PaymentCardNumber`, `PaymentCardBrand` now inherits from `str`

## Related issue number

fix #1317

## Checklist

### DIFF
--- a/changes/1317-samuelcolvin.md
+++ b/changes/1317-samuelcolvin.md
@@ -1,0 +1,1 @@
+fix card number length check in `PaymentCardNumber`, `PaymentCardBrand` now inherits from `str`.
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -591,7 +591,7 @@ class SecretBytes:
         return self._secret_value
 
 
-class PaymentCardBrand(Enum):
+class PaymentCardBrand(str, Enum):
     amex = 'American Express'
     mastercard = 'Mastercard'
     visa = 'Visa'
@@ -666,7 +666,7 @@ class PaymentCardNumber(str):
         https://en.wikipedia.org/wiki/Payment_card_number#Issuer_identification_number_(IIN)
         """
         required_length: Optional[int] = None
-        if card_number.brand is (PaymentCardBrand.visa or PaymentCardBrand.mastercard):
+        if card_number.brand in {PaymentCardBrand.visa, PaymentCardBrand.mastercard}:
             required_length = 16
             valid = len(card_number) == required_length
         elif card_number.brand is PaymentCardBrand.amex:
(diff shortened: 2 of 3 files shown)


mike-hart:
My only concerns with this approach are the following:
* The change to the tests are _just_ enough to trigger things, it might be worth being a little more aggressive (but maybe that's just my approach here)
* The use of a set for two items is more costly than `in` on a tuple
* The use of equality relaxes the requirement that someone directly assign from the enum, now they can just use a similar string (this might be considered a good thing, I don't know).

samuelcolvin:
> My only concerns with this approach are the following:
> 
> * The change to the tests are _just_ enough to trigger things, it might be worth being a little more aggressive (but maybe that's just my approach here)

What would you consider "more aggressive"? Feel free to add those tests as a separate PR.

> * The use of a set for two items is more costly than `in` on a tuple

barely, also I believe that cython rewrites simple containment tests like this as 'if x == a or x == b or ...`

> * The use of equality relaxes the requirement that someone directly assign from the enum, now they can just use a similar string (this might be considered a good thing, I don't know).

It doesn't relax anything, it makes it more likely to be correct, if anything it makes this test stricter, before if `card_number.brand` was a string it was assumed to be valid without a test.

In general pydantic leans towards considering tan enum value to be equivalent to the enum member with that value.

mike-hart:
Given the move from identity to equality, should https://github.com/samuelcolvin/pydantic/blob/master/pydantic/types.py#L672 also change from `is` to `==`?

samuelcolvin:
Yes, my mistake. Pr welcome.

PR #1328, opened 2020-03-20
document making secret types dumpable using the json method

### LINKED ISSUE #596: Configurable SecretStr

Atheuz:
## Change Summary

document making secret types dumpable using the json method, 
using json_encoders in the Config class.

## Related issue number
It will partially resolve #596

## Checklist

* [X] `changes/1328-atheuz.md` file added describing change

### DIFF
--- a/changes/1328-atheuz.md
+++ b/changes/1328-atheuz.md
@@ -1,0 +1,1 @@
+document making secret types dumpable using the json method
--- a/docs/examples/types_secret_types.py
+++ b/docs/examples/types_secret_types.py
@@ -9,6 +9,7 @@ sm = SimpleModel(password='IAmSensitive', password_bytes=b'IAmSensitiveBytes')
 # Standard access methods will not display the secret
 print(sm)
 print(sm.password)
+print(sm.dict())
 print(sm.json())
 
 # Use get_secret_value method to see the secret's content.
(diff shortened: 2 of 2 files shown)


samuelcolvin:
thank you so much, sorry for the very slow response.

Issue #1044, opened 2019-11-28
Should `SecretStr` be idempotent?


dmontagu:
# Question

* Pydantic version 1.2

Currently, a `SecretStr` can't be validated as a `SecretStr`:
```python
from pydantic import BaseModel, SecretStr


class M(BaseModel):
    a: SecretStr

M(a=SecretStr("s1"))
"""
pydantic.error_wrappers.ValidationError: 1 validation error for M
a
  str type expected (type=type_error.str)
"""

# Would raise the same error:
m = M(a="s1")
M(**dict(m))
```

Is this desirable? It seems to me it would be better if instances of `SecretStr` would pass validation as `SecretStr`.

I'm not sure whether "fixing" this would count as a bug fix, or would rise to the level of a breaking change (and require waiting until v2). I'd personally prefer to consider it a bug, but could be convinced either way. In any case, I think it would be a relatively easy fix.

(Note that this currently prevents the use of `SecretStr` as a field of a `response_model` in FastAPI due to the reparsing, as described in https://github.com/tiangolo/fastapi/issues/744)

-------

More generally, I think a goal for v2 should be for all pydantic-provided types (or at least, as many as possible) to be idempotent in this way. Does anyone disagree?

samuelcolvin:
> Is this desirable? It seems to me it would be better if instances of `SecretStr` would pass validation as `SecretStr`.

Agreed. And `SecretBytes` too of course.

MrMrRobat:
Is there acceptable way we inherit it from `str`?

samuelcolvin:
Yes, it's possible to inherit from `str`, see `AnyUrl`. But that's not required (and wouldn't work) in this case, since `str(SecretStr('fluffy123'))` is not idempotent.

Atheuz:
@samuelcolvin do you have any advice on how to resolve this? I would like to do so but I'm not sure how.

Atheuz:
@MrMrRobat @dmontagu @samuelcolvin I've created a tentative PR at #1330. 

I took inspiration from the `NameEmail` type defined here: https://github.com/samuelcolvin/pydantic/blob/master/pydantic/networks.py#L325-L350

Atheuz:
This was resolved in #1330. I believe this issue can be closed now.

samuelcolvin:
Thanks so much.

PR #1330, opened 2020-03-21
make secret types idempotent

Atheuz:
## Change Summary

make secret types idempotent

## Related issue number

#1044

## Checklist

* [X] `changes/1330-atheuz.md`

### DIFF
--- a/docs/examples/types_secret_types.py
+++ b/docs/examples/types_secret_types.py
@@ -16,9 +16,7 @@ print(sm.json())
 print(sm.password.get_secret_value())
 print(sm.password_bytes.get_secret_value())
 
-# Validation works.
 try:
-    SimpleModel(password=[1, 2, 3], 
-                password_bytes=[1, 2, 3])
+    SimpleModel(password=[1, 2, 3], password_bytes=[1, 2, 3])
 except ValidationError as e:
     print(e)
--- a/docs/examples/types_secret_types.py
+++ b/docs/examples/types_secret_types.py
@@ -3,19 +3,27 @@ from pydantic import BaseModel, SecretStr, SecretBytes, ValidationError
 class SimpleModel(BaseModel):
     password: SecretStr
     password_bytes: SecretBytes
+    password_idempotent: SecretStr
 
-sm = SimpleModel(password='IAmSensitive', password_bytes=b'IAmSensitiveBytes')
+sm = SimpleModel(password='IAmSensitive', 
+                 password_bytes=b'IAmSensitiveBytes', 
+                 password_idempotent=SecretStr('IAmIdempotent'))
 
 # Standard access methods will not display the secret
 print(sm)
 print(sm.password)
+print(sm.dict())
 print(sm.json())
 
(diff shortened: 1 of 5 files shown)


pydantic/types.py
```diff
@@ -530,11 +530,13 @@ def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
 
     @classmethod
     def __get_validators__(cls) -> 'CallableGenerator':
-        yield str_validator
         yield cls.validate
 
     @classmethod
-    def validate(cls, value: str) -> 'SecretStr':
+    def validate(cls, value: Any) -> 'SecretStr':
+        if value.__class__ == cls:
```
samuelcolvin, line 537: why not isinstance?

docs/examples/types_secret_types.py
```diff
@@ -3,19 +3,27 @@
 class SimpleModel(BaseModel):
     password: SecretStr
     password_bytes: SecretBytes
+    password_idempotent: SecretStr
```
samuelcolvin, line 6: do you need another field to cover the idempotent case?

Surely just a note in the docs should be enough?

pydantic/types.py
samuelcolvin: in case someone uses some subclass in future.

tests/test_types.py
```diff
@@ -1888,6 +1888,14 @@ def test_secretstr_equality():
     assert SecretStr('123') is not SecretStr('123')
 
 
+def test_secretstr_idempotent():
+    class Foobar(BaseModel):
+        password: SecretStr
+
+    # Should not raise an exception
+    _ = Foobar(password=SecretStr('1234'))
```
samuelcolvin, line 1896:
```suggestion
    m = Foobar(password=SecretStr('1234'))
    assert m.password. get_secret_value == '1234'
```

docs/examples/types_secret_types.py
```diff
@@ -3,19 +3,27 @@
...
 
-sm = SimpleModel(password='IAmSensitive', password_bytes=b'IAmSensitiveBytes')
+sm = SimpleModel(password='IAmSensitive', 
+                 password_bytes=b'IAmSensitiveBytes',
```
samuelcolvin, line 9: I think this is not required anymore, but I don't like this format, better to do it the way black suggests if you have to.

pydantic/types.py
Atheuz: I was sure I saw this used somewhere, but I can't find it now, I'll change it to use isinstance instead.

Atheuz: This is fixed now.

tests/test_types.py
Atheuz: This is fixed.

docs/examples/types_secret_types.py
Atheuz: This is fixed now.

docs/examples/types_secret_types.py
Atheuz: I added a note in the latest commit and removed this field.

PR #1343, opened 2020-03-26
Add `many` argument to `parse_*` methods to serialize lists easily

### LINKED ISSUE #1225: Proper way to serialize list object
# Question

In marshamllow or drf serializer has `many=True` options for list object.

But pydantic seems does not have this option.

So after searched about it, I found ways through `__root__`

```
class UserList(BaseModel):
    __root__: List[User]

UserList.parse_obj([
    {'id': '123', 'signup_ts': '2017-06-01 12:22', 'friends': [1, '2', b'3']},
    {'id': '456', 'signup_ts': '2017-06-02 12:22', 'friends': ['you']},
])
```

I wonder that, is it a official way to serialize list object?

PrettyWood:
## Change Summary
Just add a `many` argument to the parsers to serialize lists

## Related issue number

closes #1225

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -419,7 +419,7 @@ class BaseModel(metaclass=ModelMetaclass):
         return self.__config__.json_dumps(data, default=encoder, **dumps_kwargs)
 
     @classmethod
-    def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
+    def _parse_single_obj(cls: Type['Model'], obj: Any) -> 'Model':
         if cls.__custom_root_type__ and (
             not (isinstance(obj, dict) and obj.keys() == {ROOT_KEY}) or cls.__fields__[ROOT_KEY].shape == SHAPE_MAPPING
         ):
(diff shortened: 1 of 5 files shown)


samuelcolvin:
see my comment on #1225, let's wait to resolve this before doing anything more here.

pydantic/main.py
```diff
@@ -428,10 +428,19 @@ def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
...
-                exc = TypeError(f'{cls.__name__} expected dict not {obj.__class__.__name__}')
-                raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
+                exc_msg = f'{cls.__name__} expected dict not {obj.__class__.__name__}'
+                if isinstance(obj, list):
+                    exc_msg += ' (did you want to parse multiple objects with `many=True` ?)'
+                raise ValidationError([ErrorWrapper(TypeError(exc_msg), loc=ROOT_KEY)], cls) from e
         return cls(**obj)
 
+    @classmethod
+    def parse_obj(cls: Type['Model'], obj: Any, *, many: bool = False) -> Union['Model', List['Model']]:
```
PrettyWood: I decided to go for a `_parse_single_obj` for `mypy` instead of using `typing.overload` decorator as it introduced a regression on a test.
By writing something like this:
```python
@classmethod
@overload
def parse_obj(cls: Type['Model'], obj: Any, many: Literal[False]) -> 'Model': ...

@classmethod
@overload
def parse_obj(cls: Type['Model'], obj: Any, many: Literal[True]) -> List['Model']: ...

@classmethod
@overload
def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
    """The last overload is a fallback for mypy (https://mypy.readthedocs.io/en/stable/literal_types.html)"""

@classmethod
def parse_obj(cls: Type['Model'], obj: Any, many: bool = False) -> Union['Model', List['Model']]:  # noqa: F811
    if many:
        return [cls.parse_obj(x) for x in obj]
    ...
```

the test `tests/test_schema.py::test_literal_schema` started to fail because of the `Literal[1]` that didn't have the same `__values__`.

In fact it showed a bug that I didn't know:
```python
In [1]: from typing import Literal

In [2]: Literal[1]
Out[2]: typing.Literal[1]

In [3]: Literal[True]
Out[3]: typing.Literal[1]
```

```python
In [1]: from typing import Literal

In [2]: Literal[True]
Out[2]: typing.Literal[True]

In [3]: Literal[1]
Out[3]: typing.Literal[True]
```

So the annotations that were just added had in fact some side effect ! (WTF...)
As this bug is directly in python, I decided to just forget the `@overload` stuff and just go with this one instead

PrettyWood: @samuelcolvin As a matter of fact I think `parse_obj_as` is probably the best. I let the PR open but don't hesitate to close it as I'm not convinced it's pushing pydantic in the right direction on this topic

PR #1348, opened 2020-03-28
Replace exception to silent passing in mypy plugin for non-Var attributes

b0g3r:
## Change Summary

Mypy plugin breaks with an exception if I store TypeAlias in model attributes, e.g.:

```python
from typing import Union

class NotTrivialAttrubute(BaseModel):
    trivial: str
    __internal__ = Union[str, bytes]
```

## Checklist

### DIFF
--- a/tests/mypy/outputs/plugin-success-strict.txt
+++ b/tests/mypy/outputs/plugin-success-strict.txt
@@ -1,3 +1,4 @@
 29: error: Unexpected keyword argument "z" for "Model"  [call-arg]
 64: error: Untyped fields disallowed  [pydantic-field]
 79: error: Argument "x" to "OverrideModel" has incompatible type "float"; expected "int"  [arg-type]
+125: error: Untyped fields disallowed  [pydantic-field]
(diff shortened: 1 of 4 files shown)


tests/mypy/outputs/plugin-success-strict.txt
```diff
@@ -1,3 +1,4 @@
 29: error: Unexpected keyword argument "z" for "Model"  [call-arg]
 64: error: Untyped fields disallowed  [pydantic-field]
-79: error: Argument "x" to "OverrideModel" has incompatible type "float"; expected "int"  [arg-type]
\ No newline at end of file
+79: error: Argument "x" to "OverrideModel" has incompatible type "float"; expected "int"  [arg-type]
+125: error: Untyped fields disallowed  [pydantic-field]
```
b0g3r: In current test-case, I write it without explicit type, because it is the only way when mypy's semanal thinks that it is TypeAlias

In my IRL case, we use it with strict mypy and inheritance, so mypy's semanal also thinks it is a TypeAlias, like this:

```python
class Parent(BaseModel):
    __type_alias_attribute__: type

class TypeAliasAsAttribute(Parent):
    __type_alias_attribute__ = Union[str, bytes]
```

samuelcolvin:
thanks so much.

PR #1381, opened 2020-04-08
allow to pass context for orm_mode

aCLr:
Possibility to pass "context" to orm_mode method.


samuelcolvin:
please create an issue to discuss this.

Also, best to respect the template for pull requests, not just delete it.

Issue #1417, opened 2020-04-22
'exclusiveMaximum': inf probably shouldn't appear in model schema


vdwees:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
pydantic version: 1.5
pydantic compiled: True
python version: 3.7.5 (default, Nov  7 2019, 10:50:52)  [GCC 8.3.0]
platform: Linux-4.4.0-18362-Microsoft-x86_64-with-Ubuntu-18.04-bionic
optional deps. installed: ['typing-extensions']
```

Example code:
```py
import math
from pydantic import BaseModel, Field

class MyModel(BaseModel):
    myfloat: float = Field(..., lt=math.inf)
print(MyModel.schema())
# {'title': 'MyModel', 'type': 'object', 'properties': {'myfloat': {'title': 'Myfloat', 'exclusiveMaximum': inf, 'type': 'number'}}, 'required': ['myfloat']}
```

The `inf` is not json-compliant:

```json
{
  "properties": {
    "myfloat": {
      "exclusiveMaximum": inf,
      "title": "Myfloat",
      "type": "number"
    }
  },
  "required": [
    "myfloat"
  ],
  "title": "MyModel",
  "type": "object"
}
```
However, json-compliant strings cannot include `inf` anyway, so the presence of `'exclusiveMaximum': inf` (or conversely, `'exclusiveMinimum': -inf`) in the schema seems to be a bug.

Note that the reason that I have `lt=math.inf` in the first place is because I am trying to limit myfloat to real-valued floats. Perhaps pydantic provides a better way to do this?

samuelcolvin:
Happy to accept a pr to fix this. I agree it should be omitted.

> Note that the reason that I have lt=math.inf in the first place is because I am trying to limit myfloat to real-valued floats. Perhaps pydantic provides a better way to do this?

You could probably achieve the same thing with a validator and avoid anything begging added to the schema, though it might be less elegant.

vdwees:
This issue is really the definition of an edge case haha. Happy to make any changes you suggest.

chris-allnutt:
@samuelcolvin any reason not to use the json equivalent value Infinity ? 
```
{ "valid_infinity" : Infinity }
```

samuelcolvin:
very good point, we should.

vdwees:
[...]

PR #1422, opened 2020-04-23
ConstrainedFloat schema: differences between IEEE floats and json

vdwees:
## Change Summary

Modify constraints on ConstrainedFloat schema to account for differences between IEEE floats and json

Transformations applied:
- remove field exclusiveMinimum if it is equal to `-math.inf`
- remove field minimum if it is equal to `-math.inf`
- remove field exclusiveMaximum if it is equal to `math.inf`
- remove field maximum if it is equal to `math.inf`

## Related issue number

#1417 

## Checklist

### DIFF
--- a/changes/1417-vdwees.md
+++ b/changes/1417-vdwees.md
@@ -1,0 +1,4 @@
+
+Modify schema constraints on ConstrainedFloat so that exclusiveMinimum and
+minimum are not included in the schema if they are equal to `-math.inf` and
+exclusiveMaximum and maximum are not included if they are equal to `math.inf.
--- a/changes/1417-vdwees.md
+++ b/changes/1417-vdwees.md
@@ -1,3 +1,3 @@
 Modify schema constraints on `ConstrainedFloat` so that `exclusiveMinimum` and
 minimum are not included in the schema if they are equal to `-math.inf` and
-exclusiveMaximum and maximum are not included if they are equal to `math.inf.
+`exclusiveMaximum` and `maximum` are not included if they are equal to `math.inf`.
--- a/changes/1417-vdwees.md
+++ b/changes/1417-vdwees.md
@@ -1,3 +1,3 @@
-Modify schema constraints on ConstrainedFloat so that exclusiveMinimum and
+Modify schema constraints on `ConstrainedFloat` so that `exclusiveMinimum` and
 minimum are not included in the schema if they are equal to `-math.inf` and
 exclusiveMaximum and maximum are not included if they are equal to `math.inf.
--- a/changes/1417-vdwees.md
+++ b/changes/1417-vdwees.md
@@ -1,4 +1,3 @@
-
 Modify schema constraints on ConstrainedFloat so that exclusiveMinimum and
 minimum are not included in the schema if they are equal to `-math.inf` and
 exclusiveMaximum and maximum are not included if they are equal to `math.inf.
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -234,7 +234,7 @@ def get_field_schema_validations(field: ModelField) -> Dict[str, Any]:
         f_schema['const'] = field.default
     if field.field_info.extra:
         f_schema.update(field.field_info.extra)
-    modify_schema = getattr(field.outer_type_, '__modify_schema__', None)
+    modify_schema = getattr(field.type_, '__modify_schema__', None)
     if modify_schema:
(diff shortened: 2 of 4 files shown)


changes/1417-vdwees.md
```diff
@@ -0,0 +1,4 @@
+
```
samuelcolvin, line 1:
```suggestion
```

changes/1417-vdwees.md
```diff
@@ -0,0 +1,4 @@
...
+Modify schema constraints on ConstrainedFloat so that exclusiveMinimum and
```
samuelcolvin, line 2:
```suggestion
Modify schema constraints on `ConstrainedFloat` so that `exclusiveMinimum` and
```
and below

pydantic/types.py
```diff
@@ -339,13 +340,33 @@ def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
...
 
     @classmethod
     def __get_validators__(cls) -> 'CallableGenerator':
         yield strict_float_validator if cls.strict else float_validator
         yield number_size_validator
         yield number_multiple_validator
 
+    @classmethod
+    def _schema_ieee_compatibility_transform(cls, field_schema: Dict[Any, Any]) -> None:
+        """Modify constraints to account for differences between IEEE floats and json
```
samuelcolvin, line 353:
```suggestion
        """
        Modify constraints to account for differences between IEEE floats and json
```

pydantic/types.py
```diff
@@ -339,13 +340,33 @@ def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
...
+        cls._schema_ieee_compatibility_transform(field_schema)
 
     @classmethod
     def __get_validators__(cls) -> 'CallableGenerator':
         yield strict_float_validator if cls.strict else float_validator
         yield number_size_validator
         yield number_multiple_validator
 
+    @classmethod
+    def _schema_ieee_compatibility_transform(cls, field_schema: Dict[Any, Any]) -> None:
```
samuelcolvin, line 352: is there a good reason this is a separate method?

pydantic/schema.py
```diff
@@ -234,6 +234,9 @@ def get_field_schema_validations(field: ModelField) -> Dict[str, Any]:
         f_schema['const'] = field.default
     if field.field_info.extra:
         f_schema.update(field.field_info.extra)
+    modify_schema = getattr(field.outer_type_, '__modify_schema__', None)
```
samuelcolvin, line 237: maybe I'm being dumb, but can you explain why this is required?

samuelcolvin:
good point from @chris-allnutt, we should use the JSON values `Infinity` and `-Infinity`.

vdwees:
> 
> 
> good point from @chris-allnutt, we should use the JSON values `Infinity` and `-Infinity`.

Its tricky because NaN, Infinity, and -Infinity are not valid JSON, but a lot of python json libs include them.

samuelcolvin:
The acid test for me (both semantically and practically) is: what happens in js with `JSON.parse(...)`? I'd suggest we follow that.

(I'm on my phone atm., so don't actually know the answer)

pydantic/types.py
vdwees: It does not need to be. I'll merge it with `__modify_schema__`.

vdwees:
Here is the javascript JSON behaviour (tested in Firefox, but this is a pretty fundamental language feature):
```js
>>JSON.stringify({var:Infinity})
"{\"var\":null}"
>>JSON.parse("{\"var\":Infinity}")
SyntaxError: JSON.parse: unexpected character at line 1 column 8 of the JSON data
```

pydantic/schema.py
vdwees: I didn't include it at first, but I found that the type schema and field schema are both setting the constraint properties. The dict update() method merges the output of the two, so to get rid of the props you need to call `__modify_schema__` in two places. Maybe it makes sense to call `__modify_schema__` only once, at the end of `field_schema`?

samuelcolvin: if it's trivial to change it to only call this in one place and all tests pass, do that. Otherwise this is fine.

vdwees: Perhaps I am not really familiar with the codebase yet, but I don't think its trivial.

samuelcolvin:
thanks a lot.

PR #1429, opened 2020-04-24
Issue1414: wrong behavior of json() method with nested models containing custom root

silversurfer34:
## 1414
https://github.com/samuelcolvin/pydantic/issues/1414
Fix proposal for the issue 1414, wrong behavior of json method when there are nested models with custom root.
Once the dict is generated, it is parsed and each time an item has the __root__ key, its associated value is attached to the parent key
(Maybe this can be done in a more pythonic way, but with 15 years of C++ I still have bad habits)

## Checklist

### DIFF
--- a/changes/1414-silversurfer34.md
+++ b/changes/1414-silversurfer34.md
@@ -1,2 +1,1 @@
-Fix proposal for the issue 1414, wrong behavior of json method when there are nested models with custom root.
-Once the dict is generated, it is parsed and each time an item has the __root__ key, its associated value is attached to the parent key
+Change the behavior of json() method to remove the extra `__root__` keys in the result when using nested models with custom root types
--- a/changes/1414-silversurfer34.md
+++ b/changes/1414-silversurfer34.md
@@ -1,0 +1,2 @@
+Fix proposal for the issue 1414, wrong behavior of json method when there are nested models with custom root.
+Once the dict is generated, it is parsed and each time an item has the __root__ key, its associated value is attached to the parent key
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -428,8 +428,29 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
             exclude_defaults=exclude_defaults,
             exclude_none=exclude_none,
         )
-        if self.__custom_root_type__:
-            data = data[ROOT_KEY]
+
+        def reparent_root(d: 'DictStrAny') -> 'DictStrAny':
+            """
+            Parse the input dictionary, search for the keys "__root__" and take the associated values and reparent them
+            to "__root__" parent
+            """
+            result = dict()
+            for k, v in d.items():
+                if isinstance(v, dict):
+                    if k is not ROOT_KEY:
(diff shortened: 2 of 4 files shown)


pydantic/main.py
```diff
@@ -428,8 +428,28 @@ def json(
             exclude_defaults=exclude_defaults,
             exclude_none=exclude_none,
         )
-        if self.__custom_root_type__:
-            data = data[ROOT_KEY]
+
+        def reparent_root(d: 'DictStrAny') -> 'DictStrAny':
```
samuelcolvin, line 432: I think this can be done more easily with a change to `_get_value`, though I haven't tried myself.

changes/1414-silversurfer34.md
```diff
@@ -0,0 +1,2 @@
+Fix proposal for the issue 1414, wrong behavior of json method when there are nested models with custom root.
```
samuelcolvin, line 1: the change summary should be one line (or paragraph). Please remove "Fix proposal for the issue 1414", this for the change summary, not a description of the PR.

You can also make this much more concise.

changes/1414-silversurfer34.md
```diff
@@ -0,0 +1,2 @@
+Fix proposal for the issue 1414, wrong behavior of json method when there are nested models with custom root.
+Once the dict is generated, it is parsed and each time an item has the __root__ key, its associated value is attached to the parent key
```
samuelcolvin, line 2: please use back quotes around code or code like values in markdown, here you'll get "has the __root__ key" not "has the `__root__` key"

tests/test_json.py
```diff
@@ -204,6 +204,37 @@ class Model(BaseModel):
     assert Model(__root__=['a', 'b']).json() == '["a", "b"]'
 
 
+def test_encode_nested_root():
```
samuelcolvin, line 207: this has nothing to do with `.json()` in particular, it should be true of `.dict()` too, so tests shouldn't be in this file.

silversurfer34: I am not sure to understand your comment. Those tests build models with nested root as child elements, then calls the json method and the expected result is a json without the `__root__` intermediate key.
(The behavior change concerns only the json method, doing it in dict method is breaking other tests)

changes/1414-silversurfer34.md
silversurfer34: done

changes/1414-silversurfer34.md
silversurfer34: done, I did not knew the back quote trick thank you

pydantic/main.py
silversurfer34: I tried to look at _get_value but it would spread the change also to dict() method and make some other tests failing (and might introduce new issues?).

samuelcolvin:
I think the point here is that this **should** effect `dict()` as well as `json()`, in this regard I don't see that there's anything special about `json()`.

samuelcolvin:
replaced by #1607

Issue #1339, opened 2020-03-25
Document the API


charlax:
# Feature Request

It would be great to include API documentation. The Usage docs are great as a walkthrough, but when you want to go fast ("what is the name for the `Field` parameter for a constant value?"), API docs provide a faster answer.

Another use case: I want to inspect the `__fields__` on a model, and had to read the code to get the attribute name I was looking for.

API docs would be such a great addition to this amazing library! Super useful!

samuelcolvin:
I disagree, documenting everything would be extremely difficult. In the end, the only real source of truth is the code itself. One of the reasons we're all writing python not c is that it's easily readable.

I'd be happy to accept a PR to add info on `__fields__` to the model documentation. And many more PRs to extend the documentation in other ways. But I don't think "API docs" as a separate section either auto generated, or hand written are a good idea.

charlax:
I'm not asking about documenting absolutely everything :) but `Field` is a good candidate.

The API documentation is usually auto-generated (meaning no inconsistency possible between the code and the doc) and makes it possible for advanced users to go straight to the answers, whereas guides provide a more narrated way of discovering the library.

Here are some good examples of API reference from some major packages:

* https://reactjs.org/docs/react-api.html
* https://flask.palletsprojects.com/en/1.1.x/api/
* https://docs.djangoproject.com/en/3.0/ref/
* https://www.attrs.org/en/stable/api.html
* https://requests.readthedocs.io/en/master/api/
* https://marshmallow.readthedocs.io/en/stable/api_reference.html
* https://jinja.palletsprojects.com/en/2.11.x/api/
* ...

Here are some good arguments in favor of having an API reference:

* It provides immediate access to a comprehensive interface definition, instead of having to follow a narrated guide that only focuses on certain elements.
* It lets the reader discover all the functionalities easily from a single place.

samuelcolvin:
See [this](https://github.com/aio-libs/aiohttp/issues/4078) fairly long discussion on the matter.

I'm strongly opposed to multiple different types of documentation, I think there should be one and only one section of documentation on a feature. That's impossible to achieve entirely, but adding a whole new "api" section to the docs makes it impossible.

> The API documentation is usually auto-generated

Is that really true? It's not for python itself, or aiohttp, I wonder if it is for many of the projects you list above. As mentioned on the issue above, it generally doesn't work that well.

There's also no widely used way to auto-generate docs in markdown - maybe that's a strong hint that it's not a populate approach anymore.

[...]

PR #1432, opened 2020-04-25
Enums as separate models

calvinwyoung:
## Change Summary

This PR updates the OpenAPI schema generation behavior so that all enums are outputted as separate models, following the recipe for [reusable enums](https://swagger.io/docs/specification/data-models/enums/). The primary benefit is to improve support for OpenAPI code generation tools — more on this below.

To illustrate the changes, suppose we have the following models:

```python
class Foo(str, enum.Enum):
    A = 'A'
    B = 'B'

class ModelA(pydantic.BaseModel):
    foo: Foo    

class ModelB(pydantic.BaseModel):
    foo: Foo
```

Prior to this PR, the generated schema would look like the following:

```json
{
    "definitions": {
        "ModelA": {
            "title": "ModelA",
            "type": "object",
            "properties": {
                "foo": {
                    "title": "Foo",
                    "enum": [
                        "A",
                        "B"
                    ],
                    "type": "string"
                }
            },
            "required": [
                "foo"
            ]
        },
        "ModelB": {
            "title": "ModelB",
            "type": "object",
            "properties": {
                "foo": {
                    "title": "Foo",
                    "enum": [
                        "A",
                        "B"
                    ],
                    "type": "string"
                }
            },
            "required": [
                "foo"
            ]
        }
    }
}
```

Notice how the enum values are duplicated in both the `Foo` and `Bar` definitions. As a result, an OpenAPI code generation tool will create 2 definitions of the enum — one for `ModelAFoo` and one for `ModelBFoo`. This is obviously not ideal as it makes it harder to reuse the enum definition in the client code.

With this PR, the generated schema will instead look like this:

```json
{
    "definitions": {
        "Foo": {
            "title": "Foo",
            "enum": [
                "A",
                "B"
            ],
            "type": "string"
        },
        "ModelA": {
            "title": "ModelA",
            "type": "object",
            "properties": {
                "foo": {
                    "$ref": "#/definitions/Foo"
                }
            },
            "required": [
                "foo"
            ]
        },
        "ModelB": {
            "title": "ModelB",
            "type": "object",
            "properties": {
                "foo": {
                    "$ref": "#/definitions/Foo"
                }
            },
            "required": [
                "foo"
            ]
        }
    }
}
```

As a result, an OpenAPI code generation tool will output a single `Foo` enum definition that's referenced by both `ModelA` and `ModelB`.

This PR works seamlessly with fastapi v0.54.1 and with openapi-generator-cli via the `openapitools/openapi-generator-cli` docker image v4.3.0.

## Related issue number

https://github.com/samuelcolvin/pydantic/issues/1173

## Checklist

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -251,7 +251,7 @@ def get_model_name_map(unique_models: Set[Type['BaseModel']]) -> Dict[Type['Base
     name_model_map = {}
     conflicting_names: Set[str] = set()
     for model in unique_models:
-        model_name = normalize_model_name(model.__name__)
+        model_name = normalize_name(model.__name__)
         if model_name in conflicting_names:
             model_name = get_long_model_name(model)
             name_model_map[model_name] = model
@@ -518,20 +518,21 @@ def model_type_schema(
 
 def enum_process_schema(enum: EnumMeta) -> Dict[str, Any]:
     """
-    Take a single ``enum`` and generate its schema.
+    Take a single `enum` and generate its schema.
 
-    This is similar to the ``model_process_schema`` function, but applies to ``Enum`` objects.
+    This is similar to the `model_process_schema` function, but applies to ``Enum`` objects.
     """
     from inspect import getdoc
 
-    schema: Dict[str, Any] = {'title': enum.__name__}
+    schema: Dict[str, Any] = {
+        'title': enum.__name__,
+        # Python assigns all enums a default docstring value of 'An enumeration', so
+        # all enums will have a description field even if not explicitly provided.
+        'description': getdoc(enum),
+        # Add enum values and the enum field type to the schema.
+        'enum': [item.value for item in cast(Iterable[Enum], enum)],
+    }
 
-    # Python assigns all enums a default docstring value of 'An enumeration', so
-    # all enums will have a description field even if not explicitly provided.
-    schema['description'] = getdoc(enum)
-
-    # Add enum values and the enum field type to the schema.
-    schema.update({'enum': [item.value for item in cast(Iterable[Enum], enum)]})
     add_field_type_to_schema(enum, schema)
 
     return schema
@@ -614,9 +615,9 @@ json_scheme = {'type': 'string', 'format': 'json-string'}
 
 def add_field_type_to_schema(field_type: Any, schema: Dict[str, Any]) -> None:
     """
-    Update the given ``schema`` with the type-specific metadata for the given ``field_type``.
+    Update the given `schema` with the type-specific metadata for the given `field_type`.
 
-    This function looks through ``field_class_to_schema`` for a class that matches the given ``field_type``,
+    This function looks through `field_class_to_schema` for a class that matches the given `field_type`,
     and then modifies the given `schema` with the information from that type.
     """
     for type_, t_schema in field_class_to_schema:
@@ -676,9 +677,9 @@ def field_singleton_schema(  # noqa: C901 (ignore complexity)
         f_schema['const'] = literal_value
 
     if isinstance(field_type, EnumMeta):
-        model_name = normalize_model_name(field_type.__name__)
-        f_schema = {'$ref': ref_prefix + model_name}
-        definitions[model_name] = enum_process_schema(field_type)
+        enum_name = normalize_name(field_type.__name__)
+        f_schema = {'$ref': ref_prefix + enum_name}
+        definitions[enum_name] = enum_process_schema(field_type)
     else:
         add_field_type_to_schema(field_type, f_schema)
 
@@ -839,8 +840,10 @@ def get_annotation_from_field_info(annotation: Any, field_info: FieldInfo, field
     return ans
 
 
-def normalize_model_name(name: str) -> str:
-    """Normalizes the given model name."""
+def normalize_name(name: str) -> str:
+    """
+    Normalizes the given name. This can be applied to either a model *or* enum.
+    """
     return re.sub(r'[^a-zA-Z0-9.\-_]', '_', name)
 
 
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -240,8 +240,8 @@ def get_field_schema_validations(field: ModelField) -> Dict[str, Any]:
 
 def get_model_name_map(unique_models: Set[Type['BaseModel']]) -> Dict[Type['BaseModel'], str]:
     """
-    Process a set of models or enums and generate unique names for them to be used as keys in the JSON Schema
-    definitions. By default the names are the same as the class name. But if two models / enums in different Python
+    Process a set of models and generate unique names for them to be used as keys in the JSON Schema
+    definitions. By default the names are the same as the class name. But if two models in different Python
     modules have the same name (e.g. "users.Model" and "items.Model"), the generated names will be
     based on the Python module path for those conflicting models to prevent name collisions.
 
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -2,7 +2,7 @@ import re
 import warnings
 from datetime import date, datetime, time, timedelta
 from decimal import Decimal
-from enum import Enum, EnumMeta
+from enum import Enum
 from ipaddress import IPv4Address, IPv4Interface, IPv4Network, IPv6Address, IPv6Interface, IPv6Network
 from pathlib import Path
 from typing import (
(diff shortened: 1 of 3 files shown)


calvinwyoung:
@samuelcolvin One of the fastapi tests is currently failing since it's expecting an OpenAPI schema for the old enum behavior:
https://github.com/tiangolo/fastapi/blob/master/tests/test_tutorial/test_path_params/test_tutorial005.py#L37

What's the best way to handle this? Should I open another PR updating the fastapi test now, or should I do it once this PR gets merged in?

pydantic/schema.py
```diff
@@ -516,6 +516,27 @@ def model_type_schema(
     return out_schema, definitions, nested_models
 
 
+def enum_process_schema(enum: EnumMeta) -> Dict[str, Any]:
+    """
+    Take a single ``enum`` and generate its schema.
```
samuelcolvin, line 521:
```suggestion
    Take a single `enum` and generate its schema.
```

docstrings currently aren't rendered as rst or md, but if they ever are, it'll be as markdown, see #1339.  

And below.

pydantic/schema.py
```diff
@@ -516,6 +516,27 @@ def model_type_schema(
...
+    from inspect import getdoc
+
+    schema: Dict[str, Any] = {'title': enum.__name__}
+
+    # Python assigns all enums a default docstring value of 'An enumeration', so
+    # all enums will have a description field even if not explicitly provided.
+    schema['description'] = getdoc(enum)
+
+    # Add enum values and the enum field type to the schema.
+    schema.update({'enum': [item.value for item in cast(Iterable[Enum], enum)]})
```
samuelcolvin, line 534:
```suggestion
    schema['enum'] = [item.value for item in cast(Iterable[Enum], enum)]
```
or you could writing the entire thing to:

```py
schema = {
    'title': enum.__name__,
    # comment ...
    'description': getdoc(enum),
    # comment ...
    'enum': ...,
}
```

pydantic/schema.py
```diff
@@ -641,14 +675,12 @@ def field_singleton_schema(  # noqa: C901 (ignore complexity)
...
-    if issubclass(field_type, Enum):
-        f_schema.update({'enum': [item.value for item in field_type]})
-        # Don't return immediately, to allow adding specific types
-
-    for type_, t_schema in field_class_to_schema:
-        if issubclass(field_type, type_):
-            f_schema.update(t_schema)
-            break
+    if isinstance(field_type, EnumMeta):
+        model_name = normalize_model_name(field_type.__name__)
```
samuelcolvin, line 679:
```suggestion
        enum_name = normalize_model_name(field_type.__name__)
```
it's not a model

pydantic/schema.py
```diff
@@ -807,6 +839,11 @@ def go(type_: Any) -> Type[Any]:
     return ans
 
 
+def normalize_model_name(name: str) -> str:
+    """Normalizes the given model name."""
```
samuelcolvin, line 843:
```suggestion
    """
    Normalizes the given model name.
    """
```
Sorry, just how I like docstrings.

pydantic/schema.py
```diff
@@ -807,6 +839,11 @@ def go(type_: Any) -> Type[Any]:
     return ans
 
 
+def normalize_model_name(name: str) -> str:
```
samuelcolvin, line 842:
```suggestion
def normalize_name(name: str) -> str:
```
? since it applies to both models and enums.

pydantic/schema.py
calvinwyoung: That sounds awesome. Would it be helpful if I replaced all of the `` characters with ` in this file? Or would you prefer that I only update `` characters that were added by this PR to keep it cleaner?

pydantic/schema.py
calvinwyoung: Fixed!

pydantic/schema.py
calvinwyoung: Haha sounds great, fixed!

pydantic/schema.py
calvinwyoung: Fixed!

pydantic/schema.py
calvinwyoung: Well that was silly. Fixed!

samuelcolvin:
> @samuelcolvin One of the fastapi tests is currently failing since it's expecting an OpenAPI schema for the old enum behavior:
> https://github.com/tiangolo/fastapi/blob/master/tests/test_tutorial/test_path_params/test_tutorial005.py#L37
> 
> What's the best way to handle this? Should I open another PR updating the fastapi test now, or should I do it once this PR gets merged in?

Should be fixed if you rebase.

calvinwyoung:
> > @samuelcolvin One of the fastapi tests is currently failing since it's expecting an OpenAPI schema for the old enum behavior:
> > https://github.com/tiangolo/fastapi/blob/master/tests/test_tutorial/test_path_params/test_tutorial005.py#L37
> > What's the best way to handle this? Should I open another PR updating the fastapi test now, or should I do it once this PR gets merged in?
> 
> Should be fixed if you rebase.

@samuelcolvin Got it. Also thanks for the comments — I just pushed the suggested fixes.

I'll defer to you and @tiangolo with regard to the release schedule. In theory this *shouldn't* be a breaking change so long as the end user is following the Open API spec. But I also understand these things don't always work out so cleanly.

samuelcolvin:
@calvinwyoung can you rebase to master.

@tiangolo would love your feedback on the V2 decision.

tiangolo:
Sorry for the delay on this... I think I finally found a way to filter and label GitHub notifications with mentions :sweat_smile: 

I like this! :rocket: :tada: 

I've been trying to think of a way that this could break something badly enough as to be considered a "breaking change", but I haven't found anything. I don't expect FastAPI users to test their OpenAPI schemas as FastAPI already tests the generation of valid OpenAPI.

The only concern I have is, @calvinwyoung , have you tested how this looks like in the `/docs` Swagger UI?

But apart from that I think this should be safe to release in a MINOR bump.

I can update the tests in FastAPI to accept the schemas generated by this and release a minor version, that should make tests pass here. :heavy_check_mark:

samuelcolvin:
Oh so the failing FastAPI tests are real - catching an actual change with a test! 😆 

I guess you'll have to get the test to pass with both the old and new schema for a while. Let us know when you've updated FastAPI and we'll get tests passing here.

tiangolo:
So, I rushed to add the extra check in the tests and release a new version. But now I realized that it break Swagger UI :facepalm: .

I'm now checking what needs to be changed in this PR and in FastAPI, I'm getting deeper and deeper into this... :sweat_smile: 

I'll keep reporting here what I find.

Would it be fine if I commit on top of this same branch? Or would it be better if I submit a new PR with extra changes on top of this?

calvinwyoung:
@tiangolo Thanks so much for taking a look at this. I had thought I tested the Swagger UI previously, and it seemed to work okay. Before this PR, the `Schemas` section rendered something like this:

Screen Shot 2020-05-23 at 9 12 18 AM

After this PR, the schemas render like this:
Screen Shot 2020-05-23 at 9 26 32 AM

And we also have a new `Foo` model:
Screen Shot 2020-05-23 at 9 26 44 AM

Where else are you seeing breaking changes? And yes, feel free to commit directly on top of the same branch! Thanks again — really appreciate the work you and @samuelcolvin have put into maintaining these two incredible projects 🥇

samuelcolvin:
> Would it be fine if I commit on top of this same branch? Or would it be better if I submit a new PR with extra changes on top of this?

Fine by me if you commit directly to the branch for this PR, it might be simpler than a PR into this PR. I hope @calvinwyoung doesn't mind.

tiangolo:
I think that should work :rocket: 

I updated FastAPI, it ended up touching a bunch of places :sweat_smile: The main issue was that enums are supported for parameters in path, cookie, query, etc. And those didn't have a way to collect additional schema definitions. But this is supported now in 0.55.1 (just released).

I updated it here to include enums when collecting flat models and treat them as sort of top-level models for schema generation, as they get their own independent definitions. In short, this works correctly with the current FastAPI (and any other libraries that use this).

I added a couple of alias types to reduce duplication, but that can be changed if you prefer @samuelcolvin .

If you think this is fine, it's also ready from my side :slightly_smiling_face: :rocket:

samuelcolvin:
Awesome, thank you both.

calvinwyoung:
@tiangolo Ah yes that makes complete sense. Thank you!

Issue #1421, opened 2020-04-23
Validation: preserve invalid data


olexandr-klymenko:
# Feature Request

In order to be able to find broken records in the database, it would be great to be able to get full data of invalid item in the ValidationError traceback output.

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
pydantic version: 1.4
            pydantic compiled: True
                 install path: /home/oklymenko/.virtualenvs/zakaz-commerceman-api/lib/python3.8/site-packages/pydantic
               python version: 3.8.0 (default, Oct 21 2019, 13:32:28)  [GCC 7.4.0]
                     platform: Linux-5.3.0-46-generic-x86_64-with-glibc2.27
     optional deps. installed: []
```

```py
from datetime import datetime

from pydantic import BaseModel, ValidationError


class SomeResponse(BaseModel):
    class Config:
        invalid_model_full_data = True

    id: str
    name: str
    start_datetime: datetime
    end_datetime: datetime


imvalid_data = {
    "id": "123123",
    "name": ["some_name"],
    "start_datetime": 4,
    "end_datetime": 2,
}

try:
    SomeResponse(**imvalid_data)
except ValidationError as e:
    print(e)

"""
1 validation error for SomeResponse
name
  str type expected (type=type_error.str)
full_data
  {
    "id": "123123",
    "name": ["some_name"],
    "start_datetime": 4,
    "end_datetime": 2,
}
"""
...
```

samuelcolvin:
I think that's reasonable if it's easy to implement.

Let's do it as part of v2.

StephenBrown2:
At the risk of bikeshedding, I'd like to see this as an attribute on the ValidationError raised, rather than by default in the message, as I may be validating large objects and don't want to see the whole thing in every case.

If it's adding anything to the default message string, I would presume it would be under a "received" or "got" key and displayed next to/at the end of the `expected` line.

samuelcolvin:
total agree, that's why I didn't add it earlier, but I can see many scenarios where access to that data would be useful.

olexandr-klymenko:
@samuelcolvin @StephenBrown2 guys, please, review

samuelcolvin:
# Big warning to future me:

This is not trivial, we need to worry about XSS, ansi injection (think uvicorn, does ansi injection have a trendy acronym?), maybe even SQL injection.

[...]

PR #1436, opened 2020-04-25
1421 Validation: preserve invalid data

olexandr-klymenko:
## Change Summary

We introduce a new Model config attribute `show_input_data` (`False` by default) which is the case of being set to True enables invalid data output, i.e.:
```
2 validation errors for TestModel
id
  value is not a valid integer (type=type_error.integer)
age
  value is not a valid integer (type=type_error.integer)
input data
  {'age': 'QQQ', 'id': 'AAA', 'name': 'John'}
```
```
"""
>>> check_exception()
"{'age': 'QQQ', 'id': 'AAA', 'name': 'John'}"
"""

import pytest


class TestModel(BaseModel):
    id: int
    name: str
    age: int

    class Config:
        show_input_data = True


invalid_data = {
    "id": "AAA",
    "name": "John",
    "age": "QQQ"
}


def check_exception():
        TestModel(**invalid_data)
    return str(exc_info.value).split("input data")[1].strip()
```
## Related issue number
#1421 

## Checklist

### DIFF
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -81,6 +81,9 @@ not be included in the model schemas. **Note**: this means that attributes on th
 **`json_encoders`**
 : a `dict` used to customise the way types are encoded to JSON; see [JSON Serialisation](exporting_models.md#modeljson)
 
+**`show_input_data`**
+: whether to display invalid input data in case of validation error (default: `False`)
+
 ```py
 {!.tmp_examples/model_config_main.py!}
 ```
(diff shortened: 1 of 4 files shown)


samuelcolvin:
I think this needs to work differently, the input data needs to be kept with each error, not kept on the outer `ValidationError`.

There are some other things to fix, but probably worth getting this fixed before going further into a review.

samuelcolvin:
I think that the way validators work will change significantly on v2 and might require repeating some of this work.

I'm happy for you to carry on working on this if you're interested but it might be better to pause this until we get into v2.

StephenBrown2:
As a note, I too would prefer to see a 'value' key in the error_dict, though not entirely sure how much reworking that would need, possibly for every error type or `PydanticErrorMixin`...

samuelcolvin:
This needs implementing differently, is out of date and no reply, so closing for now.

PR #1453, opened 2020-04-29
Avoid unintended type coercion of validation

### LINKED ISSUE #1447: Type coercion on TypeVar fields even though type is known
# Bug

```
             pydantic version: 1.3
            pydantic compiled: True
                 install path: /usr/local/lib/python3.7/site-packages/pydantic
               python version: 3.7.4 (default, Oct 17 2019, 05:59:21)  [GCC 8.3.0]
                     platform: Linux-4.15.0-96-generic-x86_64-with-debian-10.1
     optional deps. installed: ['email-validator']
```

Pydantic appears to coerce the value of a  type variable based on the order of types defined for that type variable instead of the know type of that variable. The following code exhibits the behavior: 

```py
from pydantic import BaseModel, StrictBool, StrictInt, StrictStr
from typing import TypeVar

BoolFirst = TypeVar('BoolFirst', bool, int, str)
BoolLast = TypeVar('BoolLast', int, str, bool)
StrictType = TypeVar('StrictType', StrictBool, StrictInt, StrictStr)


class TypeTest(BaseModel):
    first: BoolFirst
    last: BoolLast
    strict: StrictType


def show_type(value: BoolFirst):
    print(f"Parameter type: {type(value)} value: {value}")
    model = TypeTest(first=value, last=value, strict=value)

    print(f"First type: {type(model.first)} : value {model.first}")
    print(f"Last type: {type(model.last)} : value {model.last}")
    print(f"Strict type: {type(model.strict)} : value {model.strict}")
    print("")


[show_type(v) for v in ['true', 1, True]]

```

Output:
```
Parameter type: <class 'str'> value: true
First type: <class 'bool'> : value True
Last type: <class 'str'> : value true
Strict type: <class 'str'> : value true

Parameter type: <class 'int'> value: 1
First type: <class 'bool'> : value True
Last type: <class 'int'> : value 1
Strict type: <class 'int'> : value 1

Parameter type: <class 'bool'> value: True
First type: <class 'bool'> : value True
Last type: <class 'int'> : value 1
Strict type: <class 'bool'> : value True
```

As can be seen from the output, the type and value of the show_type parameter is known to the function. When this known type/value is provided to a pydantic model, however, that value is type coerced to a value based on the order of the generic types allowed by the type variable, as can be seen by the First type:, Last type: output where the order of types within the type variable results in a different value for the model field with the same input value. The expected behaviour would be that pydantic would accept the type of the inbound value if it was any of the accepted types of the type variable instead of selecting the first type that was successfully coerced.

Currently, a workaround to this issue is to declare the types of the type variable using the equivalent pydantic strict types which results in the correct behaviour as shown in the output.

PrettyWood:
## Change Summary
When a `Field` value was validated, it would follow the order of the allowed types
(e.g with `Union[str, int]`, it would first check with `str` and then with `int`).
But as some validators coerce the value (for example `1` can be converted to `"1"` if a `str` is expected), a valid value could be changed because of the order of the validators.
To avoid this, we prioritize the right type if found before checking with the others

## Related issue number

fix #1447

## Checklist

### DIFF
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -30,7 +30,7 @@ def test_str_bytes():
     assert repr(m.__fields__['v']) == "ModelField(name='v', type=Union[str, bytes], required=True)"
 
     m = Model(v=b'b')
-    assert m.v == 'b'
+    assert m.v == b'b'
 
     with pytest.raises(ValidationError) as exc_info:
         Model(v=None)
@@ -47,7 +47,7 @@ def test_str_bytes_none():
     assert m.v == 's'
 
     m = Model(v=b'b')
-    assert m.v == 'b'
+    assert m.v == b'b'
 
     m = Model(v=None)
     assert m.v is None
@@ -61,7 +61,7 @@ def test_union_int_str():
     assert m.v == 123
 
     m = Model(v='123')
-    assert m.v == 123
+    assert m.v == '123'
 
     m = Model(v=b'foobar')
     assert m.v == 'foobar'
@@ -84,7 +84,7 @@ def test_union_priority():
     class ModelTwo(BaseModel):
         v: Union[str, int] = ...
 
-    assert ModelOne(v='123').v == 123
+    assert ModelOne(v='123').v == '123'
     assert ModelTwo(v='123').v == '123'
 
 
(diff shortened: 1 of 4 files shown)


tests/test_edge_cases.py
```diff
@@ -30,7 +30,7 @@ class Model(BaseModel):
     assert repr(m.__fields__['v']) == "ModelField(name='v', type=Union[str, bytes], required=True)"
 
     m = Model(v=b'b')
-    assert m.v == 'b'
+    assert m.v == b'b'
```
PrettyWood, line 33: This fixes also some edge cases. I just fixed the tests here but I guess we could remove them to avoid unnecessary tests

samuelcolvin:
I think this will slow pydantic down too much. For what's basically an edge case, also I've we're going to fix this, we should provide a more comprehensive fix.

PR #1474, opened 2020-05-04
Add an example to documentation for reserved ORM field names

lukapeschke:
Adding a documentation example as discussed in https://github.com/samuelcolvin/pydantic/issues/1463

### DIFF
--- a/docs/examples/models_orm_mode_reserved_name.py
+++ b/docs/examples/models_orm_mode_reserved_name.py
@@ -1,0 +1,27 @@
+import typing
+
+from pydantic import BaseModel, Field
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+
+class MyModel(BaseModel):
+    metadata: typing.Dict[str, str] = Field(alias='metadata_')
+
+    class Config:
+        orm_mode = True
+
+BaseModel = declarative_base()
+
+class SQLModel(BaseModel):
+
+    __tablename__ = 'my_table'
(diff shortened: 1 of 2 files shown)


docs/examples/models_orm_mode_reserved_name.py
```diff
@@ -0,0 +1,27 @@
+import typing
+
+from pydantic import BaseModel, Field
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+
+class MyModel(BaseModel):
+    metadata: typing.Dict[str, str] = Field(alias="metadata_")
```
samuelcolvin, line 8: single quotes please.

StephenBrown2: Would you mind a PR to add flake8 linting to `docs/examples`, so you don't need to make comments like these?

samuelcolvin: That would be great, but I think you might need to disable many error codes. The many one is that currently docs examples use one blank line everywhere, not two. I'd be happy to change that in all the examples.

StephenBrown2: Here ya go! https://github.com/samuelcolvin/pydantic/pull/1530

lukapeschke:
@samuelcolvin thanks for the review! I've updated the example and rebased on master

samuelcolvin:
awesome, thank you.

Issue #1479, opened 2020-05-05
Support more json schema `$ref` patterns


Kilo59:
# Feature Request

I would like to be able to generate more types of references than can be created via a simple `ref_prefix`. For example, referencing a schema that resides in another file.

`{ "$ref": "/schemas/Foo.json#/}`

https://json-schema.org/understanding-json-schema/structuring.html#reuse
https://docs.opis.io/json-schema/1.x/pointers.html

## Example

```py
import json
from pydantic import BaseModel
from pydantic.schema import schema

class Foo(BaseModel):
    a: int

class Model(BaseModel):
    a: Foo

# Default location for OpenAPI
model_schema = Model.schema(ref_template='/schemas/{name}.json#/')
print(json.dumps(model_schema , indent=2))
```
```json
{
  "Model": {
    "title": "Model",
    "type": "object",
    "properties": {
      "a": {
        "$ref": "/schemas/Foo.json#/"
      }
    },
    "required": ["a"]
  }
}
```

I believe python's [`string.Template`](https://docs.python.org/3/library/string.html#template-strings) could be a good way to implement this (which I have attempted). 
There are still some details that need to be worked out.

[Pull request](https://github.com/samuelcolvin/pydantic/pull/1480)

samuelcolvin:
Sorry for the slow reply and thanks for the feature request and PR.

I haven't personally used `$ref` so I'm personally kind of neutral on the feature.

To accept it I guess I would hope for some other interest here via some 👍s.

Regarding how it's implemented, I don't think it needs `string.Template`. `.format()` should work find, e.g. something like

```py
ref_template = ref_template or '#/definitions/{model}'
m_schema = {'$ref': ref_template.format(model=model_name)}
```

I guess to be clean, if this is accepted, we should depreciate `ref_prefix` in v2 and replaced it with `ref_template`

Kilo59:
@samuelcolvin just updated the PR to use `.format`

PR #1480, opened 2020-05-05
schema ref_template

Kilo59:
## Change Summary

Adds a `ref_template` option to `schema` generation that takes a [`string.Template`](https://docs.python.org/3/library/string.html#template-strings) and substitutes the `model_name` when generating `$ref`s to other schemas.

`my_model.schema(ref_template=string.Template("/foo/${model_name}.json#/bar"))`

### Edit:
Updated to just use a regular `.format()` template.
https://github.com/samuelcolvin/pydantic/issues/1479#issuecomment-630087550
```python
my_model.schema(ref_template="/foo/${model}.json#/bar")
```

## Related issue number

#1479

## Checklist

- [x] change to regular string.format() template


pydantic/schema.py
```diff
@@ -135,7 +146,10 @@ def model_schema(
     if model_name in nested_models:
         # model_name is in Nested models, it has circular references
         m_definitions[model_name] = m_schema
-        m_schema = {'$ref': ref_prefix + model_name}
+        if ref_template:
+            m_schema = {'$ref': ref_template.substitute(model_name=model_name)}
+        else:
+            m_schema = {'$ref': ref_prefix + model_name}
```
Kilo59, line 152: This could probably be done more elegantly.

Some additional things I could do.
1. regular `ref_prefix`s could also be resolved using the same `string.Template` method. `"#/components/schemas/${model_name}"`
2. better error message if a bad template is passed.
3. coerce a regular string into a template instead of taking an actual `string.Template` objecty.

pydantic/schema.py
```diff
@@ -69,6 +70,7 @@ def schema(
     title: Optional[str] = None,
     description: Optional[str] = None,
     ref_prefix: Optional[str] = None,
+    ref_template: Optional[string.Template] = None,
```
Kilo59, line 73: With no extra handling...
1. If `str` is passed instead of a `string.Template` an `AttributeError` will be raised.
   `AttributeError: 'str' object has no attribute 'substitute'`
2. If `string.Template` is passed that doesn't have the `$model_name` identifier a `KeyError` will be raised.
     `KeyError: 'foo_bar'`

https://docs.python.org/3/library/string.html#template-strings

tests/test_schema.py
```diff
@@ -1088,6 +1089,41 @@ class Baz(BaseModel):
...
+            'Foo': {
+                'title': 'Foo',
+                'type': 'object',
+                'properties': {'a': {'title': 'A', 'type': 'string'}},
+                'required': ['a'],
+            },
+        }
+    }
+
+
```
Kilo59, line 1126: I'm happy to add more tests.

samuelcolvin:
sorry for the slow reply, as I mentioned on the issue #1479, I think a normal `str` should be used not `Template`.

It's not necessary, and `Template` is very rarely used so will confuse many people.

Kilo59:
> sorry for the slow reply, as I mentioned on the issue #1479, I think a normal `str` should be used not `Template`.
> 
> It's not necessary, and `Template` is very rarely used so will confuse many people.

Makes sense and thanks for the reminder. I'll try to update this over the weekend.

pydantic/schema.py
```diff
@@ -76,6 +77,7 @@ def schema(
     title: Optional[str] = None,
     description: Optional[str] = None,
     ref_prefix: Optional[str] = None,
+    ref_template: Optional[str] = None,
```
Kilo59, line 80: @samuelcolvin I was just following the pattern here.
But should I set the `default_template` as the default argument for `ref_template`?

```python
 ref_template: Optional[str] = '#/definitions/{model}',
```

image

Kilo59:
Getting a failure on `test_constrained_set_default` that I don't understand.
https://github.com/samuelcolvin/pydantic/runs/837388454
```python
E           pydantic.error_wrappers.ValidationError: 1 validation error for ConSetModelMax
E           v
E             value is not a valid set (type=type_error.set)

pydantic/main.py:346: ValidationError
```

samuelcolvin:
See #1682

Kilo59:
@samuelcolvin should I also update the schema-customization section docs?
If so should I say anything about `ref_prefix` being deprecated in the future etc?
https://pydantic-docs.helpmanual.io/usage/schema/#schema-customization

pydantic/schema.py
samuelcolvin: I think better to switch to `ref_template: str = '#/definitions/{model}',`, the others are only optional because they really can be optional.

The only advantage of making this optional would be that `ref_template` could take priority over `ref_prefix` if it's not `None`.

samuelcolvin (commented):
Just need documentation and the tweak discussed here, otherwise looks good.

pydantic/schema.py
Kilo59: Done. 
I don't know if it would have been better to keep a `REF_TEMPLATE` constant and use that as the default for all the `ref_template` constant, but it looked like doing that prevented the editor help text from showing what the actual value was.

samuelcolvin:
This is looking great, but you still need to:
* [ ] fix coverage (there are two cases) which are currently not covered
* [ ] add something to docs.

Kilo59:
> This is looking great, but you still need to:
> 
> * [ ]  fix coverage (there are two cases) which are currently not covered

What are the cases that are not covered?

samuelcolvin:
look at the codecov report above.

samuelcolvin:
thanks so much.

PR #1492, opened 2020-05-08
document ability to use `Field()` in `create_model()`

xlotlu:
## Change Summary

Docstring update of `create_model()`.


samuelcolvin:
please could you add this to the docs and also test(s) for it.

samuelcolvin:
thanks so much.

PR #1504, opened 2020-05-11
fix(field): remove some side effects of `default_factory`

### LINKED ISSUE #1491: `default_factory` being run prior to object init
# Bug

```
             pydantic version: 1.5.1
```

`default_factory` gets run twice before any object instantiation. This is obvious for callables with side-effects:

```py
>>> from pydantic import BaseModel, Field
>>> class Seq():
...     def __init__(self):
...         self.v = 0 
...     def __call__(self):
...         self.v += 1
...         return self.v
...   
>>> class MyModel(BaseModel):
...     id: int = Field(default_factory=Seq())
...   
>>> MyModel()
MyModel(id=3)
```

PrettyWood:
## Change Summary
Fix some side effects of `default_factory`:
- call it only once when no validation is needed
- remove default value of generated schema

## Related issue number

fix #1491 and #1520

## Checklist

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -296,7 +296,7 @@ class ModelField(Representation):
 
         if isinstance(value, FieldInfo):
             field_info = value
-            value = field_info.default_factory() if field_info.default_factory is not None else field_info.default
+            value = None if field_info.default_factory is not None else field_info.default
         else:
             field_info = FieldInfo(value, **field_info_from_config)
         required: 'BoolUndefined' = Undefined
@@ -341,7 +341,14 @@ class ModelField(Representation):
         Note: this method is **not** idempotent (because _type_analysis is not idempotent),
         e.g. calling it it multiple times may modify the field and configure it incorrectly.
         """
+
+        # to prevent side effects by calling the default_factory for nothing, we only call it
+        # when we want to validate the default value i.e. when `validate_all` is set to True
+        if self.default_factory is not None and self.type_ is not None and not self.model_config.validate_all:
+            return
+
         default_value = self.get_default()
+
         if default_value is not None and self.type_ is None:
             self.type_ = default_value.__class__
             self.outer_type_ = self.type_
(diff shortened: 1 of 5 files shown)


samuelcolvin:
The problem with this is that it doesn't work if someone does something like:

```py
v = 0


def factory():
    global v
    v += 1
    return v


class MyModel(BaseModel):
    id: int = Field(default_factory=factory)
```

I worry that copying the factory in one place complicates things more than it solve problems.

PrettyWood:
@samuelcolvin I agree it doesn't solve all the problems but handling all the cases would mean rethink the workflow with `prepare`. I thought that it would at least solve most edge cases and that the example of the linked issue would be way more common than the one you put. But if you think otherwise, I'll remove the code and just keep the `default = None` for the other issue

PrettyWood:
@samuelcolvin I updated the content of the PR and the description. I also added some documentation to keep that in mind.

pydantic/fields.py
```diff
@@ -341,7 +341,14 @@ def prepare(self) -> None:
         Note: this method is **not** idempotent (because _type_analysis is not idempotent),
         e.g. calling it it multiple times may modify the field and configure it incorrectly.
         """
+
+        # to prevent side effects by calling the default_factory for nothing, we only call it
+        # when we want to validate the default value i.e. when `validate_all` is set to True
+        if self.default_factory is not None and self.type_ is not None and not self.model_config.validate_all:
```
PrettyWood, line 347: I added a check of `self.type_` to avoid regressions but as it's a beta feature I don't know if we could actually just force people to set a type when using it. I'm not sure it makes sense to use `default_factory` without one

samuelcolvin: I think requiring a type with `default_factory` would be a better solution to this.

PrettyWood: Alright glad we agree on this ! Will remove the extra check then

PrettyWood: @samuelcolvin Change done in fe533a0

aviramha:
Waiting for this!

samuelcolvin:
Thanks so much.

PR #1525, opened 2020-05-17
Benchmark apischema

wyfo:
## Change Summary

Add a the library [apischema](https://github.com/wyfo/apischema/) to benchmark.
Actually, I adds 2 tests, one for the library in its current state, and one for an optimized version (because it's a benchmark) coded in place ; this version drops some features of the library that are not handled by *Pydantic* `BaseModel` (`__post_init__`, `patternProperties` at field level) and adds some optimizations (caching). 

Also update `rand_date` format to conform to `datetime.fromisoformat`.

### DIFF
--- a/benchmarks/requirements.txt
+++ b/benchmarks/requirements.txt
@@ -10,3 +10,4 @@ attr
 cattrs
 cerberus
 voluptuous
+apischema
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -106,7 +106,7 @@ def null_missing_email():
 
 def rand_date():
     r = random.randrange
-    return f'{r(1900, 2020)}-{r(0, 12)}-{r(0, 32)}T{r(0, 24)}:{r(0, 60)}:{r(0, 60)}'
+    return f'{r(1900, 2020)}-{r(0, 12):02}-{r(0, 32):02}T{r(0, 24):02}:{r(0, 60):02}:{r(0, 60):02}'
 
 
 def remove_missing(d):
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -51,6 +51,11 @@ try:
 except Exception as e:
     TestVoluptuous = None
 
+try:
+    from test_apischema import TestApischema, TestApischemaOptimized
+except ImportError:
+    TestApischema, TestApischemaOptimized = None, None
+
 PUNCTUATION = ' \t\n!"#$%&\'()*+,-./'
 LETTERS = string.ascii_letters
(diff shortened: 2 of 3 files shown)


StephenBrown2:
I believe `rand_date` is the way it is to specifically catch badly (i.e. non-ISO) formatted dates and raise a `ValidationError`.

Also, https://github.com/samuelcolvin/pydantic/pull/1525/files#diff-6d758509aa91a59278e79ec61a427c99R55-R58 states that you cannot use PositiveInt because of coercion, but then end up coercing the value anyways? https://github.com/samuelcolvin/pydantic/pull/1525/files#diff-6d758509aa91a59278e79ec61a427c99R14-R22

samuelcolvin:
I'm afraid apischema doesn't have enough usage to be warrant being included in the benchmarks - the idea behind the benchmarks was to give a rough estimate of speed relative to "household name" libraries like DRF, Marshmallow and Cerberus.

I know this is annoying, but I've refused other libraries, so I can't now include apischema.

In future, best to submit an issue suggesting something like this before working on a PR.

wyfo:
> I'm afraid apischema doesn't have enough usage to be warrant being included in the benchmarks - the idea behind the benchmarks was to give a rough estimate of speed relative to "household name" libraries like DRF, Marshmallow and Cerberus.
> 
> I know this is annoying, but I've refused other libraries, so I can't now include apischema.
> 
> In future, best to submit an issue suggesting something like this before working on a PR.

Thank you for your response. I understand your refusal (but I'm wondering why 2 no-longer-maintained libraries have been added for the beginning of my work ; in fact, it was the presence of this 2 libraries that makes me think I could add mine in a PR without the need of an issue).

samuelcolvin:
Humm, I see valideer is no longer maintained, but which other library would you say is not maintained?

wyfo:
> I no longer use Voluptuous personally (in fact I no longer regularly write Python code). Rather than leave the project in a limbo of people filing issues and wondering why they're note being worked on, I believe this notice will more clearly set expectations.
From the author of *Voluptuous*

samuelcolvin:
Good point, yes, it should be removed. 

I hadn't seen it because it's not included in the benchmarks on docs site (I guess because import failed).

wyfo:
Ok, so my interrogation as been answered, and consistency is preserved. That's good for me.

samuelcolvin:
Voluptuous has over a million downloads a month: https://pypistats.org/packages/voluptuous regardless of how it's maintained, I think that represents a considerable amount of usage.

StephenBrown2:
And it's also in active use in many projects. Valdeer as well has over 13k downloads last month. If it's mature, it doesn't necessarily need active maintenance.

wpk-:
Some time has passed now. If the valideer download stats are anything to go by the time has come to include apischema, too, won't you think?

ltworf:
No can do, apischema is actually faster than pydantic :)

samuelcolvin:
We removed the benchmarks from our documentation some time ago precisely because we didn't want to take part in a flame war. 

We recognize that pydantic's performance advantage has diminished as other libraries have improved, but we are looking forward to the release of v2 which will be a 5-50x improvement over v1.

PR #1527, opened 2020-05-18
Fix NameEmail equality comparison

### LINKED ISSUE #1514: Same NameEmail field instances are not equal
# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /Users/***/.local/share/virtualenvs/tempenv-42b2286175873/lib/python3.7/site-packages/pydantic
               python version: 3.7.4 (default, Jul 22 2019, 11:21:52)  [Clang 10.0.1 (clang-1001.0.46.4)]
     optional deps. installed: ['email-validator']
```

I'm assuming that equality should be enforced on field types such as `NameEmail`. At the moment it seems that instances that appear to be equal are generating different hashes. Unsure if this is desired, but I would argue it shouldn't be.

```py
import pydantic

# will raise an AssertionError
assert pydantic.NameEmail("test", "test@test.com") == pydantic.NameEmail("test", "test@test.com")
```

stephen-bunn:
## Change Summary

Introducing an `__eq__` for `networks.NameEmail` as two similar instances were not being correctly evaluated as equal.

```python
import pydantic

# previously this would raise an AssertionError
assert pydantic.NameEmail("test", "test@example.com") == pydantic.NameEmail("test", "test@example.com")1
```

## Related issue number

closes #1514

## Checklist

### DIFF
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -330,10 +330,7 @@ class NameEmail(Representation):
         self.email = email
 
     def __eq__(self, other: Any) -> bool:
-        if isinstance(other, NameEmail):
-            return (self.name, self.email,) == (other.name, other.email,)
-        else:
-            return other == self
+        return isinstance(other, NameEmail) and (self.name, self.email) == (other.name, other.email)
 
     @classmethod
     def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -329,6 +329,12 @@ class NameEmail(Representation):
         self.name = name
         self.email = email
 
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, NameEmail):
+            return (self.name, self.email,) == (other.name, other.email,)
+        else:
+            return other == self
+
     @classmethod
     def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
         field_schema.update(type='string', format='name-email')
--- a/tests/test_networks.py
+++ b/tests/test_networks.py
@@ -453,5 +453,5 @@ def test_name_email():
 
     assert str(Model(v=NameEmail('foo bar', 'foobaR@example.com')).v) == 'foo bar <foobaR@example.com>'
     assert str(Model(v='foo bar  <foobaR@example.com>').v) == 'foo bar <foobaR@example.com>'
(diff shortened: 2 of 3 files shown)


pydantic/networks.py
```diff
@@ -329,6 +329,12 @@ def __init__(self, name: str, email: str):
         self.name = name
         self.email = email
 
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, NameEmail):
+            return (self.name, self.email,) == (other.name, other.email,)
+        else:
+            return other == self
```
samuelcolvin, line 336:
```suggestion
        return isinstance(other, NameEmail) and (self.name, self.email) == (other.name, other.email)
```
?

tests/test_networks.py
```diff
@@ -453,3 +453,5 @@ class Model(BaseModel):
 
     assert str(Model(v=NameEmail('foo bar', 'foobaR@example.com')).v) == 'foo bar <foobaR@example.com>'
     assert str(Model(v='foo bar  <foobaR@example.com>').v) == 'foo bar <foobaR@example.com>'
+    assert Model(v=NameEmail('foo bar', 'foobaR@example.com')) == Model(v=NameEmail('foo bar', 'foobaR@example.com'))
+    assert Model(v=NameEmail('foo bar', 'foobaR@example.com')) != 'foo bar <foobaR@example.com>'
```
samuelcolvin, line 457:
```suggestion
        assert NameEmail('foo bar', 'foobaR@example.com') == NameEmail('foo bar', 'foobaR@example.com')
        assert NameEmail('foo bar', 'foobaR@example.com') != NameEmail('foo bar', 'different@example.com')
```
Your first assertion was more complicated than necessary, the second I'm afraid didn't make any sense since you're comparing a model to a string.

pydantic/networks.py
StephenBrown2: Or should this also compare to the string representation?
```py
return (isinstance(other, NameEmail) and (self.name, self.email) == (other.name, other.email)) or str(self) == other
```
?

samuelcolvin: surely too much magic?

stephen-bunn: Yeah, not a fan of that magic. Changes applied.

tests/test_networks.py
stephen-bunn: Originally wanted a test case to hit that alternate branch, but with the above suggestions not really necessary. Changes applied

samuelcolvin:
Awesome, thank you.

PR #1568, opened 2020-05-27
Update benchmarks

fbergroth:
## Change Summary

Seeing that `docs/.benchmarks_table.md` hadn't been updated for quite a while, I ran it along with a profiler to compare against attrs.

I noticed that the results were quite skewed due to attrs using dateutil vs the custom date parser in pydantic. It seemed only fair to give attrs a better one, since it's not dateutil we're interested in here. So I replaced dateutil with ciso8601 in the benchmark.

This changed things drastically! Now attrs and valideer beats pydantic by a huge margin. Do others get similar results?

Cheers!

## Related issue number

None

## Checklist

### DIFF
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -105,8 +105,11 @@ def null_missing_email():
 
 
 def rand_date():
-    r = random.randrange
-    return f'{r(1900, 2020)}-{r(0, 12)}-{r(0, 32)}T{r(0, 24)}:{r(0, 60)}:{r(0, 60)}'
+    ts = random.randrange(
+        datetime(1900, 1, 1).timestamp(),
+        datetime(2021, 1, 1).timestamp(),
+    )
+    return datetime.fromtimestamp(ts).isoformat()
 
 
 def remove_missing(d):
--- a/benchmarks/test_cattrs.py
+++ b/benchmarks/test_cattrs.py
@@ -3,7 +3,7 @@ from typing import List, Optional
 
 import attr
 import cattr
-from dateutil.parser import parse
+from ciso8601 import parse_datetime as parse
 
 
 class TestCAttrs:
--- a/benchmarks/test_cattrs.py
+++ b/benchmarks/test_cattrs.py
@@ -3,7 +3,6 @@ from typing import List, Optional
 
 import attr
 import cattr
-from ciso8601 import parse_datetime as parse
 
 
 class TestCAttrs:
@@ -33,7 +32,7 @@ class TestCAttrs:
                 raise ValueError()
             return i
 
-        cattr.register_structure_hook(datetime, lambda isostring, _: parse(isostring))
+        cattr.register_structure_hook(datetime, lambda isostring, _: datetime.fromisoformat(isostring))
         cattr.register_structure_hook(str, structure_str)
         cattr.register_structure_hook(int, structure_int)
         cattr.register_structure_hook(PositiveInt, structure_posint)
--- a/docs/.benchmarks_table.md
+++ b/docs/.benchmarks_table.md
@@ -2,11 +2,11 @@
 
 Package | Version | Relative Performance | Mean validation time
 --- | --- | --- | ---
-attrs + cattrs | `19.3.0` |  | 9.4μs
-valideer | `0.4.2` | 1.4x slower | 12.9μs
-trafaret | `2.0.2` | 7.1x slower | 66.5μs
-pydantic | `1.5.1` | 8.7x slower | 81.3μs
-voluptuous | `0.11.7` | 9.4x slower | 88.1μs
-marshmallow | `3.6.0` | 15.3x slower | 143.0μs
-django-rest-framework | `3.11.0` | 75.1x slower | 703.4μs
-cerberus | `1.3.2` | 146.2x slower | 1368.9μs
+attrs + cattrs | `19.3.0` |  | 16.1μs
+valideer | `0.4.2` | 1.3x slower | 21.1μs
+pydantic | `1.5.1` | 4.3x slower | 69.8μs
+voluptuous | `0.11.7` | 4.5x slower | 73.0μs
+trafaret | `2.0.2` | 6.9x slower | 111.6μs
+marshmallow | `3.6.0` | 7.2x slower | 115.9μs
+django-rest-framework | `3.11.0` | 39.6x slower | 638.3μs
+cerberus | `1.3.2` | 77.5x slower | 1248.8μs
--- a/docs/.benchmarks_table.md
+++ b/docs/.benchmarks_table.md
@@ -2,10 +2,11 @@
 
 Package | Version | Relative Performance | Mean validation time
 --- | --- | --- | ---
-pydantic | `1.1` |  | 43.6μs
-attrs + cattrs | `19.3.0` | 1.4x slower | 59.9μs
-valideer | `0.4.2` | 1.4x slower | 61.8μs
-marshmallow | `3.2.2` | 2.5x slower | 107.6μs
-trafaret | `2.0.0` | 3.4x slower | 148.7μs
-django-rest-framework | `3.10.3` | 12.6x slower | 551.2μs
-cerberus | `1.3.2` | 26.3x slower | 1146.3μs
+attrs + cattrs | `19.3.0` |  | 11.0μs
+valideer | `0.4.2` | 1.5x slower | 16.9μs
+trafaret | `2.0.2` | 5.7x slower | 63.3μs
+pydantic | `1.5.1` | 6.7x slower | 73.3μs
+voluptuous | `0.11.7` | 7.7x slower | 85.0μs
+marshmallow | `3.6.0` | 11.2x slower | 123.2μs
+django-rest-framework | `3.11.0` | 59.7x slower | 657.4μs
+cerberus | `1.3.2` | 129.8x slower | 1429.2μs
--- a/docs/.benchmarks_table.md
+++ b/docs/.benchmarks_table.md
@@ -2,11 +2,11 @@
 
 Package | Version | Relative Performance | Mean validation time
 --- | --- | --- | ---
-attrs + cattrs | `19.3.0` |  | 11.0μs
-valideer | `0.4.2` | 1.5x slower | 16.9μs
-trafaret | `2.0.2` | 5.7x slower | 63.3μs
-pydantic | `1.5.1` | 6.7x slower | 73.3μs
-voluptuous | `0.11.7` | 7.7x slower | 85.0μs
-marshmallow | `3.6.0` | 11.2x slower | 123.2μs
-django-rest-framework | `3.11.0` | 59.7x slower | 657.4μs
-cerberus | `1.3.2` | 129.8x slower | 1429.2μs
+attrs + cattrs | `19.3.0` |  | 9.4μs
+valideer | `0.4.2` | 1.4x slower | 12.9μs
+trafaret | `2.0.2` | 7.1x slower | 66.5μs
+pydantic | `1.5.1` | 8.7x slower | 81.3μs
+voluptuous | `0.11.7` | 9.4x slower | 88.1μs
+marshmallow | `3.6.0` | 15.3x slower | 143.0μs
+django-rest-framework | `3.11.0` | 75.1x slower | 703.4μs
+cerberus | `1.3.2` | 146.2x slower | 1368.9μs
--- a/benchmarks/requirements.txt
+++ b/benchmarks/requirements.txt
@@ -1,6 +1,5 @@
 # no versions since benchmarks should always run with the most recent version of each package
 
-ciso8601
 trafaret
 django
 djangorestframework
--- a/benchmarks/requirements.txt
+++ b/benchmarks/requirements.txt
@@ -1,6 +1,6 @@
 # no versions since benchmarks should always run with the most recent version of each package
 
-python-dateutil
+ciso8601
 trafaret
 django
 djangorestframework
--- a/benchmarks/test_cerberus.py
+++ b/benchmarks/test_cerberus.py
@@ -1,6 +1,5 @@
 from cerberus import Validator, __version__
-from dateutil.parser import parse as datetime_parse
-
+from ciso8601 import parse_datetime
 
 class TestCerberus:
     package = 'cerberus'
@@ -20,7 +19,7 @@ class TestCerberus:
             'contractor': {'type': 'integer', 'min': 0, 'nullable': True, 'coerce': int},
             'upstream_http_referrer': {'type': 'string', 'maxlength': 1023, 'nullable': True},
             'grecaptcha_response': {'type': 'string', 'minlength': 20, 'maxlength': 1000, 'required': True},
-            'last_updated': {'type': 'datetime', 'nullable': True, 'coerce': datetime_parse},
+            'last_updated': {'type': 'datetime', 'nullable': True, 'coerce': parse_datetime},
             'skills': {
                 'type': 'list',
                 'default': [],
--- a/benchmarks/test_cerberus.py
+++ b/benchmarks/test_cerberus.py
@@ -1,5 +1,6 @@
+from datetime import datetime
 from cerberus import Validator, __version__
-from ciso8601 import parse_datetime
+
 
 class TestCerberus:
     package = 'cerberus'
@@ -19,7 +20,7 @@ class TestCerberus:
             'contractor': {'type': 'integer', 'min': 0, 'nullable': True, 'coerce': int},
             'upstream_http_referrer': {'type': 'string', 'maxlength': 1023, 'nullable': True},
             'grecaptcha_response': {'type': 'string', 'minlength': 20, 'maxlength': 1000, 'required': True},
-            'last_updated': {'type': 'datetime', 'nullable': True, 'coerce': parse_datetime},
+            'last_updated': {'type': 'datetime', 'nullable': True, 'coerce': datetime.fromisoformat},
             'skills': {
                 'type': 'list',
                 'default': [],
--- a/benchmarks/test_trafaret.py
+++ b/benchmarks/test_trafaret.py
@@ -1,4 +1,4 @@
-from ciso8601 import parse_datetime as parse
+from datetime import datetime
 import trafaret as t
 
 
@@ -23,7 +23,7 @@ class TestTrafaret:
             t.Key('upstream_http_referrer', optional=True): t.Or(t.Null | t.String(max_length=1023)),
             t.Key('grecaptcha_response'): t.String(min_length=20, max_length=1000),
 
-            t.Key('last_updated', optional=True): t.Or(t.Null | t.String >> parse),
+            t.Key('last_updated', optional=True): t.Or(t.Null | t.String >> datetime.fromisoformat),
 
             t.Key('skills', default=[]): t.List(t.Dict({
                 'subject': t.String,
--- a/benchmarks/test_trafaret.py
+++ b/benchmarks/test_trafaret.py
@@ -1,4 +1,4 @@
-from dateutil.parser import parse
+from ciso8601 import parse_datetime as parse
 import trafaret as t
 
 
--- a/benchmarks/test_valideer.py
+++ b/benchmarks/test_valideer.py
@@ -1,7 +1,6 @@
 import re
 import subprocess
-
-from ciso8601 import parse_datetime
+from datetime import datetime
 
 import valideer as V
 
@@ -24,7 +23,7 @@ class TestValideer:
             'contractor': V.Range(V.AdaptTo(int), min_value=1),
             'upstream_http_referrer': V.Nullable(V.String(max_length=1023)),
             '+grecaptcha_response': V.String(min_length=20, max_length=1000),
-            'last_updated': V.AdaptBy(parse_datetime),
+            'last_updated': V.AdaptBy(datetime.fromisoformat),
             'skills': V.Nullable(
                 [
                     {
--- a/benchmarks/test_valideer.py
+++ b/benchmarks/test_valideer.py
@@ -1,7 +1,8 @@
 import re
 import subprocess
 
-import dateutil.parser
+from ciso8601 import parse_datetime
+
 import valideer as V
 
 # valideer appears to provide no way of getting the installed version
@@ -23,7 +24,7 @@ class TestValideer:
             'contractor': V.Range(V.AdaptTo(int), min_value=1),
             'upstream_http_referrer': V.Nullable(V.String(max_length=1023)),
             '+grecaptcha_response': V.String(min_length=20, max_length=1000),
-            'last_updated': V.AdaptBy(dateutil.parser.parse),
+            'last_updated': V.AdaptBy(parse_datetime),
             'skills': V.Nullable(
                 [
                     {
--- a/benchmarks/test_voluptuous.py
+++ b/benchmarks/test_voluptuous.py
@@ -1,4 +1,4 @@
-from ciso8601 import parse_datetime
+from datetime import datetime
 import voluptuous as v
 from voluptuous.humanize import humanize_error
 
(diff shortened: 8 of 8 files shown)


benchmarks/test_cattrs.py
```diff
@@ -3,7 +3,7 @@
 
 import attr
 import cattr
-from dateutil.parser import parse
+from ciso8601 import parse_datetime as parse
```
pganssle, line 6: Not sure what the full scope of the parsing and the support matrix of these benchmarks is, but if you are willing to scope down from "parses all date-like strings" to "parses the subset of ISO 8601 supported by `ciso8601`", then you may also be willing to scope it to "parses the results of `datetime.isoformat()`", in which case in Python 3.7+ there's the `datetime.fromisoformat` alternate constructor, which is very fast (likely same speed as `ciso8601`).

It appears [someone has backported](https://pypi.org/project/backports-datetime-fromisoformat/) it to earlier Python 3 versions as well.

pganssle:
As the maintainer of `dateutil`, I'm +1 on this change — `dateutil.parser.parse` is not really intended to be used for validation, it assumes the string you have is a valid date and tries to squint and change its perspective until it hallucinates something that might be construed as a date, like so:

```python
>>> from dateutil.parser import parse
>>> parse("1")
datetime.datetime(2020, 5, 1, 0, 0)
```

So even if it were very fast I don't think it's an appropriate validator. `dateutil.parser.isoparse` *will* validate that something is a valid ISO 8601 datetime and it's much faster than `dateutil.parser.parse` (even for ISO 8601 datetimes), but it's considerably slower than one of the C-accelerated versions like `datetime.fromisoformat`.

samuelcolvin:
This is definitely not a fair comparison to pydantic datetime parsing.

Maybe we have to implement a datetime parsing method similar to pydantics for libraries that don't come with their own.

I know it sounds weird, but it might be fairer than either what we have now or this.

pganssle:
@samuelcolvin What specifically seems unfair about it? It seems like `pydantic` supports these formats:

-  `YYYY-MM-DD[T]HH:MM[:SS[.ffffff]][Z[±]HH[:]MM]]]`
- int or float as a string (assumed as Unix time)

The first one is an ISO 8601 time, so `ciso8601` will parse all of those formats, plus other valid ISO 8601 formats. `fromisoformat` will also parse all of those, except for the `Z` at the end. The "string can be interpreted as an integer" sounds to me like an anti-feature to me, TBH.

In any case, it's way more fair than what's happening now, which is basically just incorrectly using a slow function as a validator. If you insist on having essentially identical coverage to what `pydantic` provides, you can do this:

```python
def validate_exactly_like_pydantic(s):
    try:
        datetime.fromisoformat(s)
        return True
    except ValueError:
        pass
    if s.endswith("Z"):
        try:
            datetime.fromisoformat(s[:-1])
            return True
         except ValueError:
             pass
    try:
         return -62135492638.0 <= float(s) <= 253402318800.0
    except ValueError:
        return False
```

Or with `ciso8601`:

```python
def validate_exactly_like_pydantic(s):
    try:
        ciso8601.parse_datetime(s)
    except ValueError:
        pass

    try:
         return -62135492638.0 <= float(s) <= 253402318800.0
    except ValueError:
        return False
```

Still, I think you should just go with bare `fromisoformat` or `ciso8601.parse_datetime`. They're conceptually equivalent.

pganssle:
> @samuelcolvin What specifically seems unfair about it? It seems like `pydantic` supports these formats:
> 
>     * `YYYY-MM-DD[T]HH:MM[:SS[.ffffff]][Z[±]HH[:]MM]]]`
> 
>     * int or float as a string (assumed as Unix time)

It seems that the documentation on this is not accurate (I've opened #1573), so most of the examples in the benchmark are _failing_ validation when you switch to a pure-ISO 8601 parser, because it includes a bunch of weird test cases. Updating those to conform to the format in the documentation, I get:

```
pydantic best=64.159μs/iter avg=65.664μs/iter stdev=1.218μs/iter version=1.5.1
attrs + cattrs best=23.091μs/iter avg=23.731μs/iter stdev=0.662μs/iter version=19.3.0
```

benchmarks/test_cattrs.py
fbergroth: Thanks, I wasn't aware of this. I updated the PR to use `datetime.fromisoformat`, and also gave you write access to my fork, in case you'd like to push to this PR.

docs/.benchmarks_table.md
```diff
@@ -2,11 +2,11 @@
...
-trafaret | `2.0.2` | 5.7x slower | 63.3μs
-pydantic | `1.5.1` | 6.7x slower | 73.3μs
-voluptuous | `0.11.7` | 7.7x slower | 85.0μs
-marshmallow | `3.6.0` | 11.2x slower | 123.2μs
-django-rest-framework | `3.11.0` | 59.7x slower | 657.4μs
-cerberus | `1.3.2` | 129.8x slower | 1429.2μs
+attrs + cattrs | `19.3.0` |  | 9.4μs
+valideer | `0.4.2` | 1.4x slower | 12.9μs
+trafaret | `2.0.2` | 7.1x slower | 66.5μs
+pydantic | `1.5.1` | 8.7x slower | 81.3μs
```
pganssle, line 8: I'm assuming the benchmarks are meant to be relative to `pydantic` no? Since it's about comparing `pydantic` to other things?

StephenBrown2: They're meant to be relative to the fastest, which up until now has been `pydantic`.

fbergroth:
> This is definitely not a fair comparison to pydantic datetime parsing.

Can you elaborate? It is a minimal change that fixes a rather poor choice of library unrelated to attrs.

> It seems that the documentation on this is not accurate (I've opened #1573), so most of the examples in the benchmark are _failing_ validation when you switch to a pure-ISO 8601 parser, because it includes a bunch of weird test cases.

Great catch! I must have missed the low success rates in the output.

Date parsing aside. If I drop the `last_updated` field in the benchmark, which skips all the date parsing, pydantic is about 3.5x slower than attrs.

fbergroth:
I've updated `rand_date` to only generate proper isodates. Now all libraries have the same success rate.

benchmarks/run.py
```diff
@@ -105,8 +105,11 @@ def null_missing_email():
 
 
 def rand_date():
-    r = random.randrange
-    return f'{r(1900, 2020)}-{r(0, 12)}-{r(0, 32)}T{r(0, 24)}:{r(0, 60)}:{r(0, 60)}'
+    ts = random.randrange(
```
samuelcolvin, line 108: I don't agree with this change:
* I've used the same data (good bad or ugly) to do the benchmarks since their inception, I don't want to change it unless I have to
* the point is that some of these should be invalid, os check that the validation library catches them

docs/.benchmarks_table.md
```diff
@@ -2,10 +2,11 @@
...
-pydantic | `1.1` |  | 43.6μs
-attrs + cattrs | `19.3.0` | 1.4x slower | 59.9μs
-valideer | `0.4.2` | 1.4x slower | 61.8μs
-marshmallow | `3.2.2` | 2.5x slower | 107.6μs
-trafaret | `2.0.0` | 3.4x slower | 148.7μs
-django-rest-framework | `3.10.3` | 12.6x slower | 551.2μs
-cerberus | `1.3.2` | 26.3x slower | 1146.3μs
+attrs + cattrs | `19.3.0` |  | 16.1μs
+valideer | `0.4.2` | 1.3x slower | 21.1μs
+pydantic | `1.5.1` | 4.3x slower | 69.8μs
```
samuelcolvin, line 7: are you using the compiled version of pydantic? Does you machine match that's in the docs?

benchmarks/run.py
fbergroth: > I've used the same data (good bad or ugly) to do the benchmarks since their inception, I don't want to change it unless I have to

Then it would make sense to seed the rng, so that everyone gets the same cases.json?

> the point is that some of these should be invalid, os check that the validation library catches them

Alright, I thought the invalid dates was an oversight and not intended.

docs/.benchmarks_table.md
fbergroth: Apparantly I was not using a compiled version.

Running the following outputs `False` on my machine:
```
python3.7 -m venv .venv
source .venv/bin/activate
pip install pydantic
python -c 'import pydantic; print(pydantic.compiled)'                                                                                                                                                                                          
```
I'm running Ubuntu 18.04.4 LTS and Python 3.7.5.

When compiling with cython locally, pydantic goes from 4.3x to 2.7x slower.

fbergroth:
> `datetime.fromisoformat` is not a valid comparison with pydantic's datetime parsing because it fails for 21 out of the 25 test parameters in `tests/test_datetime_parse.py::test_datetime_parsing`.
> 
> If there were a few cases with different parsing behaviour, that would be fine. But this is just a completely different function.

@samuelcolvin what do you consider acceptable? Should the date parsing function behave the same as pydantic for the benchmark cases, or for the test suite?

samuelcolvin:
> what do you consider acceptable

Let's just copy [this](https://github.com/samuelcolvin/pydantic/blob/2eb62a3b2f2d345aa7b4625306331f68a7bc0fd5/pydantic/datetime_parse.py#L154-L200) function and use it for those libraries that don't come with datetime parsing, seems simple and as reasonably fair.

> Then it would make sense to seed the rng, so that everyone gets the same cases.json?

Best to use [this `cases.json`](https://gist.github.com/samuelcolvin/b99b6c1bea779f31197382ffab367bb8) file as that's what I always use to run benchmarks.

> When compiling with cython locally, pydantic goes from 4.3x to 2.7x slower.

Best to just use what's installed with `pip install -U pydantic` that way you've definitely got the same version others get and it's a fair comparison to the other libraries.

fbergroth:
> Let's just copy [this](https://github.com/samuelcolvin/pydantic/blob/2eb62a3b2f2d345aa7b4625306331f68a7bc0fd5/pydantic/datetime_parse.py#L154-L200) function and use it for those libraries that don't come with datetime parsing, seems simple and as reasonably fair.

Should I copy it verbatim or just import it from pydantic?

> Best to use [this `cases.json`](https://gist.github.com/samuelcolvin/b99b6c1bea779f31197382ffab367bb8) file as that's what I always use to run benchmarks.

If benchmarks are supposed to be ran with that file, I think you (or I?) should either commit it, or we can agree to seed the rng?

pganssle:
> `datetime.fromisoformat` is not a valid comparison with pydantic's datetime parsing because it fails for 21 out of the 25 test parameters in `tests/test_datetime_parse.py::test_datetime_parsing`.

This is just a recipe for a misleading benchmark, TBH. Right now the `attrs` and `valideer` measurements are basically just "how long does an unrelated function take to parse a datetime", which is nothing to do with the speed of either library. A *lot* of people are going to want to validate against ISO 8601 or a subset thereof. I suspect comparatively few people will want to validate against formats like `2019-1-1T3:5:3` (and as I've pointed out *the documentation of pydantic doesn't even suggest that these formats are supported*).

So a benchmark comparing how fast `pydantic` is compared to other libraries attempting to replicate `pydantic`'s exact behavior is a bit contrived and inherently biased towards `pydantic`. I think you should do one or more of the following things:

1. Adjust the `datetime` format in benchmark so that they are doing the more common task of validating ISO 8601 datetimes (which, conveniently, will make the datetime parsing not the major bottleneck anymore).
2. Drop the `datetime` field entirely, since it's not well-agreed what it means to "validate a datetime".
3. Split this out into separate benchmarks that focus on different aspects of the performance instead of one omnibus task.
4. Don't publish benchmarks at all because it's very easy for them to be misleading.

samuelcolvin:
@pganssle thanks for your input, but I don't think the negativity and criticism is necessary.

I'm not a company selling something based on these benchmarks, they're based on open source code and easy to read if you so wish, I'm explicit about where the benchmark code is and I've accepted numerous PRs to alter them.

When I first build pydantic I wanted to get a ball park figure of its performance compared to other libraries I had used like DRF and trafaret (the benchmarks are even adapted from some code we had at TutorCruncher that validated enquires using trafaret). Since then others have extended them and I've kept them passing.

In the nearly 3 years the benchmarks have been public, no one else that I can remember has complained about this.

I'm a little mystified and disappointed by how much ire one particular field has drawn in the last few days.

dateutil is great, I've used it a lot, so when some libraries didn't have datetime parsing support, I just reached for that to get something working. Now I'm making a suggestion of an easy fix since there are much more important things I should be responding to on pydantic, let alone the rest of my life.

I thought (and frankly still think) it was a good choice since using it would be reasonable choice if you were using a library that didn't have datetime parsing and you needed datetime parsing.

---

> So a benchmark comparing how fast pydantic is compared to other libraries attempting to replicate pydantic's exact behavior is a bit contrived and inherently biased towards pydantic.

I agree that's why it's not how I originally built the benchmarks, I was just suggestion a quick solution that removed any time differences for this field.

> Right now the attrs and valideer measurements are basically just "how long does an unrelated function take to parse a datetime"

Is that really true? If you're going to make assertions of that nature (particularly in a thread like this), please provide some numbers. How long does `dateutil.parse` take relative to `pydantic.datetime_parse`? What percentage of validation time is spent on datetimes? I suspect it's significant but not all, thus your statement may be inaccurate.

> I suspect comparatively few people will want to validate against formats like 2019-1-1T3:5:3

I've acknowledge a bug and agreed to fix it - this is therefore an unnecessarily pejorative comparison.

> Adjust the datetime format in benchmark so that they are doing the more common task of validating ISO 8601 datetimes

My experience is that many people appreciate the relatively permissive way pydantic parsing datetimes, I certainly do; hence why I implemented it that way. 

> Drop the datetime field entirely, since it's not well-agreed what it means to "validate a datetime"

Removing complex datetime parsing because it's not supported by some of the libraries included in the benchmarks seems odd given these benchmarks are in pydantic's docs.

> Split this out into separate benchmarks that focus on different aspects of the performance instead of one omnibus task.

I'd love this, like the [web framework benchmarks](https://www.techempower.com/benchmarks/).

You'll need about 200 cases once you've tested the performance of all common field types, custom field types, validation vs. parsing, passing vs. failing data.

Once you've built this, let me know I'll happily remove the benchmarks from pydantic and link to your site.

> Don't publish benchmarks at all because it's very easy for them to be misleading.

Or you could [not look at them if you don't like them](https://m.xkcd.com/386/).

---

## In Summary

* I'll happily change this if we can come up with a "better" solution
* I'll happily accept a PR to add a disclaimer to the benchmarks
* I'll happily close this and do something more useful with my life
* Please remain positive and constructive, I'm here for fun (I guess you are too), if it stops being fun to be here, we'll leave and the world will be a poorer place

pganssle:
> @pganssle thanks for your input, but I don't think the negativity and criticism is necessary.

I'm sorry that you've taken my constructive criticism as "negativity". I am saying that the benchmark is misleading, because it is mostly just measuring a slow function that is not intended as a validator *anyway*. Currently it's giving pydantic a hugely unfair advantage in the benchmarks.

> Is that really true? If you're going to make assertions of that nature (particularly in a thread like this), please provide some numbers. How long does dateutil.parse take relative to pydantic.datetime_parse? What percentage of validation time is spent on datetimes? I suspect it's significant but not all, thus your statement may be inaccurate.

Yes. Both I and @fbergroth have given you numbers on this to very clearly indicate that this is the case.

Using `pydantic` as installed from `pip install -U`, but removing the `last_updated` field, I get:

```
pydantic best=55.854μs/iter avg=58.045μs/iter stdev=2.752μs/iter version=1.5.1
attrs + cattrs best=22.321μs/iter avg=22.655μs/iter stdev=0.195μs/iter version=19.3.0
valideer best=25.020μs/iter avg=25.499μs/iter stdev=0.413μs/iter version=0.4.2
marshmallow best=150.253μs/iter avg=152.754μs/iter stdev=1.850μs/iter version=3.6.0
voluptuous best=91.668μs/iter avg=93.607μs/iter stdev=2.900μs/iter version=0.11.7
trafaret best=111.797μs/iter avg=112.847μs/iter stdev=0.933μs/iter version=2.0.2
django-rest-framework best=750.342μs/iter avg=775.890μs/iter stdev=25.370μs/iter version=3.11.0
cerberus best=1423.815μs/iter avg=1473.033μs/iter stdev=41.757μs/iter version=1.3.2
```

When running it *with* `last_updated` enabled, I get the following:

```
pydantic best=63.883μs/iter avg=67.589μs/iter stdev=4.754μs/iter version=1.5.1
attrs + cattrs best=84.108μs/iter avg=86.639μs/iter stdev=3.763μs/iter version=19.3.0
valideer best=89.428μs/iter avg=103.729μs/iter stdev=17.651μs/iter version=0.4.2
marshmallow best=179.458μs/iter avg=201.214μs/iter stdev=31.996μs/iter version=3.6.0
voluptuous best=191.930μs/iter avg=193.918μs/iter stdev=1.879μs/iter version=0.11.7
trafaret best=212.664μs/iter avg=218.940μs/iter stdev=8.965μs/iter version=2.0.2
django-rest-framework best=820.654μs/iter avg=832.587μs/iter stdev=13.580μs/iter version=3.11.0
cerberus best=1564.833μs/iter avg=1591.593μs/iter stdev=44.106μs/iter version=1.3.2
```

As you can see, `attrs` spends 80% of the benchmark on the datetime parsing part of it! For the common workflow of "validate that it is an ISO 8601 datetime", the datetime parsing is no longer the bottleneck. This is consistent with the speed of `dateutil`, which on my machine takes about 70µs to parse a datetime in one of the specified formats:

```python
>>> from dateutil.parser import parse                      
>>> %timeit parse("2020-01-5T0:1:3")                       
69.6 µs ± 469 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

> I've acknowledge a bug and agreed to fix it - this is therefore an unnecessarily pejorative comparison.

I don't understand why you think this was pejorative. I was saying that a benchmark is intended to give you an indication of what is faster for given workflows. I suspect that "ISO 8601 datetime" or even the specific subset of ISO 8601 that `datetime.isoformat()` generates is a very common workflow — many data sources use ISO 8601, and it is a logical choice for anyone who controls both the data generation and the data parsing sides of the datetime. The reason I bring up the mismatch between documentation and implementation is that it means that anyone reading the documentation, and presumably the person who *wrote* the documentation, felt that ISO 8601 should be enough.

fbergroth:
@samuelcolvin I don't see the negativity. It's not pydantic being criticized, but the benchmark, which really is misleading in its current state.

> > I suspect comparatively few people will want to validate against formats like 2019-1-1T3:5:3
> 
> I've acknowledge a bug and agreed to fix it - this is therefore an unnecessarily pejorative comparison.

But it's the only case where pydantics `parse_datetime` will differ from `datetime.fromisoformat` in the benchmark data, no?

Regarding the other libraries in this benchmark that _do_ parse datetimes, I'm sure they don't pass the tests for `pydantic.datetime_parse.parse_datetime` either. Coercing a string to a float and interpreting it as a timestamp is not a common scenario, and arguably bad to do by default. The format of the datetime in the benchmark must be specified so that every library tries to parse the same thing. Then the invalid cases will be obvious, and we won't have different success rates for the various libraries.

For the record this is by no means an attempt to shed bad light on pydantic. I'm gladly using it in a couple of my projects.

StephenBrown2:
Just popping in to comment on two things based on my understanding of how pydantic is intended to work:

> Coercing a string to a float and interpreting it as a timestamp is not a common scenario, and arguably bad to do by default.

Pydantic does coersion for everything by default. It is a parsing library that can also be used for validation, so it is more flexible than others in that respect. Also, I have indeed come across timestamps passed as strings in my work, so while not common, it's not unreasonable, especially since when setting the type, one would know "this timestamp should be a datetime" ahead of time.

> The format of the datetime in the benchmark must be specified so that every library tries to parse the same thing.

The format is a rough approximation of ISO8601 with some invalid data, but every library is parsing the same data and so should have a similar success rate.

fbergroth:
> Pydantic does coersion for everything by default.

I'm aware, and I would prefer strict by default, but that is another discussion. It was meant as a reply to the following:



> `datetime.fromisoformat` is not a valid comparison with pydantic's datetime parsing because it fails for 21 out of the 25 test parameters in `tests/test_datetime_parse.py::test_datetime_parsing`.



Anyway,

> The format is a rough approximation of ISO8601 with some invalid data, but every library is parsing the same data and so should have a similar success rate.

Apparantly the other libraries parses `2019-1-1T3:5:3` as well. Oh well.

samuelcolvin:
Okay, after stepping back, perhaps I overreacted a little above. But I still think (entirely understandably) @fbergroth and @pganssle have a degree of tunnel vision on this.

That's probably my fault for not explaining what I meant clearly enough, I'll try to do that now. Then propose a choice of solutions that hopefully satisfy everyone.

### Apples and Oranges

We're comparing two very different libraries when we put DRF and attrs in the same benchmarks. 

DRF is designed to validate/parse data for a (restful) http API, attrs is designed to provide some constraints at a code boundary (among other things, as far as I understand it).

Comparing these very two very different libraries is perhaps understandable since pydantic is used in both contexts, but it still might be a mistake.

To get around the fact that attr + cattr, valideer and others don't provide all the same functions for validation as pydantic, DRF, cerberus, marshmallow etc., there are a few things we could do (none of which are ideal). My solution was to ask the question:

> If I was writing some code to validate this data using X, what would I do to "just get something working"?

I didn't give any particular consideration to performance when answer this question because:
1. I was in a hurry to get some benchmarks working.
2. I felt that poor performance was a fair price for that library to pay for not having builtin support for that data type.
3. (probably) subconsciously I wanted pydantic to "win" the benchmarks, so it suited my purpose for the hack to slow down the library in question. To the extend this was true, it was a mistake. Sorry.

I still think in the beginning my solution was reasonable - if I was parsing some data using one of the libraries which doesn't come with datetime coercion/parsing I would still reach for dateutil's `parse` function. But I also see that in the benchmarks as they stand now, using that method is severely distorting the results.

---

### Resolution Options 

I think just removing the datetime field is not a good solution, since being able to parse datetimes is very often useful in many of the contexts where pydantic (and its competitors) is used.

Using a hand written function to parse the particular cases in the benchmarks data is also not a useful data point for most people looking at the benchmarks results IMHO.

I'm happy to do any of the following to resolve this (just answer below with your preferences and I'll accept a PR for the most popular selection):
1. Remove libraries from the benchmarks that don't come with built in support for parsing all the types in the benchmarks.
2. Keep the benchmarks as the are but explain prominently in the disclaimer (see below) that attrs etc. use `dateutil.parser.parse` for parsing datetimes which damage their performance
3. swap out `dateutil.parser.parse` for some other parsing method from another library (pendulum, arrow, maya etc.) which is more performant
4. swap out `dateutil.parser.parse` for a method similar to pydantic's `parse_datetime` - I still don't think this is unreasonable for libraries that don't support datetime parsing.

I'll also (regardless of which of the above is chosen) add a prominent disclaimer to the benchmarks giving a better description of what's going on and the trade-offs chosen.

Is that reasonable?

samuelcolvin:
Based on a comment above and this discussion, I've implemented #1583 (sadly it'll need to wait for v2 to get merged).

This should make pydantic more correct and more performant and also make it easier to implement a datetime parsing function for libraries that don't ship with one which is a fair alternative to what pydantic ships with.

fbergroth:
Sorry for the delay, I've been swamped with other things.

> I think just removing the datetime field is not a good solution, since being able to parse datetimes is very often useful in many of the contexts where pydantic (and its competitors) is used.

I agree, I wouldn't make sense to leave datetimes out. 

> Using a hand written function to parse the particular cases in the benchmarks data is also not a useful data point for most people looking at the benchmarks results IMHO.

I don't think using `datetime.fromisoformat` would qualify as a hand written function. With the PR you just referenced, using fromisoformat would behave just as pydantic in the benchmark? I think it's great for the benchmark if you can just use stdlib rather than involving yet another library. Then you would compare, say, attrs + pendulum to pydantic, rather than just attrs.

I would also suggest some more improvements to the benchmark:
- define the format and post it on the website (what is being benchmarked)
- define the valid and invalid inputs, and base the success rate on that
- update the benchmark on every release
- make the official numbers reproducible (either using a seed or committing the test data)

Anyway, I'll close this PR since you clearly don't want this implementation, and I think I'm just reiterating my previous posts or what @pganssle already argued better than me.

PR #1574, opened 2020-05-27
add colander to benchmarks

slav0nic:
https://docs.pylonsproject.org/projects/colander/en/latest/index.html

### DIFF
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -65,7 +65,7 @@ random = random.SystemRandom()
 # in order of performance for csv
 other_tests = [
     t for t in
-    [TestCAttrs, TestValideer, TestMarshmallow, TestVoluptuous, TestTrafaret, TestDRF, TestCerberus, TestColander]
+    [TestColander, TestCAttrs, TestValideer, TestMarshmallow, TestVoluptuous, TestTrafaret, TestDRF, TestCerberus]
     if t is not None
 ]
 
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -51,6 +51,11 @@ try:
 except Exception as e:
     TestVoluptuous = None
 
+try:
+    from test_colander import TestColander
+except Exception:
+    TestColander = None
+
 PUNCTUATION = ' \t\n!"#$%&\'()*+,-./'
 LETTERS = string.ascii_letters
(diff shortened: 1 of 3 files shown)


benchmarks/run.py
```diff
@@ -60,7 +65,7 @@
 # in order of performance for csv
 other_tests = [
     t for t in
-    [TestCAttrs, TestValideer, TestMarshmallow, TestVoluptuous, TestTrafaret, TestDRF, TestCerberus]
+    [TestCAttrs, TestValideer, TestMarshmallow, TestVoluptuous, TestTrafaret, TestDRF, TestCerberus, TestColander]
```
samuelcolvin, line 68: it's a bit hacky, but the benchmarks currently use this order to decide how the order results are shown in, therefore could you changed this to 
```suggestion
    [TestColander, TestCAttrs, TestValideer, TestMarshmallow, TestVoluptuous, TestTrafaret, TestDRF, TestCerberus]
```

benchmarks/test_colander.py
```diff
@@ -0,0 +1,55 @@
+import pkg_resources
+import colander as c
+
+
+class TestColander:
+    package = "colander"
```
samuelcolvin, line 6: single quotes please. This code isn't linted yet, but pydantic prefers single quotes everywhere else.

benchmarks/test_colander.py
```diff
@@ -0,0 +1,55 @@
...
+            location = LocationSchema()
+            contractor = c.SchemaNode(c.Int(), validator=c.Range(min=0), missing=None)
+            upstream_http_referrer = c.SchemaNode(
+                c.Str(), validator=c.Length(max=1023), missing=None
+            )
+            grecaptcha_response = c.SchemaNode(
+                c.Str(), validator=c.Length(min=20, max=1000)
+            )
+            last_updated = c.SchemaNode(
+                c.DateTime(format="%Y-%m-%dT%H:%M:%S"), missing=None
```
samuelcolvin, line 42: see #1568 - in summary we need to modify the benchmarks but all libraries should be able to accept ints as unix time stamps as well as strings and strings should work with timezones like `fromisoformat()`.

benchmarks/test_colander.py
samuelcolvin: and in other places.

benchmarks/test_colander.py
slav0nic: internally `colander` use legacy `iso8601` lib, that not fully follow ISO format, so this will be required fully custom DateTime field implementation

samuelcolvin: maybe simplest to wait until this issues raised in #1568 are resolved, then complete this.

I know the current situation is annoying and I agree completely custom datetime parsing doesn't make sense, but I don't see another way until the problems in pydantic are resolved in v2.

samuelcolvin:
see #3973

PR #1583, opened 2020-05-31
more correct and performant datetime parsing

### LINKED ISSUE #1573: datetime validation is less strict than the documentation suggests
# Bug
The documentation for [datetime types](https://pydantic-docs.helpmanual.io/usage/types/#datetime-types) suggests the following string types are acceptable:

- `YYYY-MM-DD[T]HH:MM[:SS[.ffffff]][Z[±]HH[:]MM]]]`
- `int` or `float` as a string (assumed as Unix time)

But the actual parser accepts much looser formats, including some pretty funky ones, because month, day, hour, minute and second are all allowed to be 1 digit:

```python
>>> from pydantic.datetime_parse import parse_datetime
>>> parse_datetime('1915-5-9T14:1:2')
datetime.datetime(1915, 5, 9, 14, 1, 2)
```

I suppose the easy thing to do would be to update the documentation to reflect this, but I think the *right* thing to do would be to deprecate those unusual formats or put them behind a flag of some sort. They're not formats that are in common use.

samuelcolvin:
## Change Summary

Use `fromisoformat` for datetime parsing, this should be more correct and more performant.

This is not finished yet:
* docstrings need updated
* might be possible to use similar logic for dates
* docs needs updating
* probably needs to wait for v2 since it's a subtle change in behaviour for a very commonly used field

## Related issue number

fix #1573

## Checklist


samuelcolvin:
This is replaced by logic in pydantic-core

PR #1585, opened 2020-05-31
Fix #1561 - apply Config env names for fields defined in the parent c…

### LINKED ISSUE #1561: Env settings in the config for a parent field apparently don`t work
# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /home/yury/.cache/pypoetry/virtualenvs/chatbot-py3.6/lib/python3.6/site-packages/pydantic
               python version: 3.6.9 (default, Apr 18 2020, 01:56:04)  [GCC 8.4.0]
                     platform: Linux-5.3.0-51-generic-x86_64-with-Ubuntu-18.04-bionic
```
If I configure an alternative prefix for a field `foo` in the same class it gets into the field configuration, but whenever I set that in the descendant it doesn't seem to have effect
```py
In [2]: from pydantic import AnyUrl, BaseSettings 
In [4]: %set_env FOO=7                                                                                                                                                                                      
env: FOO=7

In [9]: class A(BaseSettings): 
   ...:     foo : int 
   ...:     class Config: 
   ...:         env_prefix = 'PREFIX_' 
   ...:         fields = { 
   ...:             'foo': {'env': ['FOO', 'PREFIX_FOO'],}, 
   ...:         } 
   ...:          
   ...:                                                                                                                                                                                                     
In [10]: A().foo                                                                                                                                                                                            
Out[10]: 7

In [11]: A.__fields__['foo'].field_info                                                                                                                                                                     
Out[11]: FieldInfo(default=Ellipsis, extra={'env': ['FOO', 'PREFIX_FOO'], 'env_names': ['foo', 'prefix_foo']})

In [12]: class A(BaseSettings): 
    ...:     foo : int 
    ...:      
                                                                                                                                                                                              
In [13]: class B(A): 
    ...:     class Config: 
    ...:         env_prefix = 'PREFIX_' 
    ...:         fields = { 
    ...:             'foo': {'env': ['FOO', 'PREFIX_FOO'],}, 
    ...:         } 
    ...:          
    ...:          
    ...:                                                                                                                                                                                                    

In [14]: B().foo                                                                                                                                                                                            
ValidationError                           Traceback (most recent call last)
<ipython-input-14-40329edc0228> in <module>
----> 1 B().foo
~/.cache/pypoetry/virtualenvs/chatbot-py3.6/lib/python3.6/site-packages/pydantic/env_settings.cpython-36m-x86_64-linux-gnu.so in pydantic.env_settings.BaseSettings.__init__()
~/.cache/pypoetry/virtualenvs/chatbot-py3.6/lib/python3.6/site-packages/pydantic/main.cpython-36m-x86_64-linux-gnu.so in pydantic.main.BaseModel.__init__()
ValidationError: 1 validation error for B
foo

In [15]: A.__fields__['foo']                                                                                                                                                                                
Out[15]: ModelField(name='foo', type=int, required=True)

In [16]: A.__fields__['foo'].field_info                                                                                                                                                                     
Out[16]: FieldInfo(default=Ellipsis, extra={'env_names': {'foo'}})

In [17]: B.__fields__['foo'].field_info                                                                                                                                                                     
Out[17]: FieldInfo(default=Ellipsis, extra={'env_names': {'prefix_foo'}})
```

ojomio:
…lass

## Change Summary

Allow descendant Settings to override env variable names for fields defined in parent Settings

#1561 

## Checklist

### DIFF
--- a/changes/1561-ojomio.md
+++ b/changes/1561-ojomio.md
@@ -1,0 +1,2 @@
+Allow descendant Settings models to override env variable names for the fields defined in parent Settings models with `env` in their `Config`.
+Previously only `env_prefix` configuration option was applicable
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import AbstractSet, Any, Dict, List, Mapping, Optional, Union
 
 from .fields import ModelField
-from .main import BaseModel, Extra
+from .main import BaseConfig, BaseModel, Extra
 from .typing import display_as_type
 from .utils import deep_update, sequence_like
 
@@ -65,7 +65,7 @@ class BaseSettings(BaseModel):
             d[field.alias] = env_val
         return d
 
-    class Config(BaseModel.Config):
+    class Config(BaseConfig):
         env_prefix = ''
         env_file = None
         validate_all = True
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -3,7 +3,7 @@ import warnings
 from pathlib import Path
 from typing import AbstractSet, Any, Dict, List, Mapping, Optional, Union
 
-from .fields import FieldInfo, ModelField
+from .fields import ModelField
 from .main import BaseModel, Extra
 from .typing import display_as_type
 from .utils import deep_update, sequence_like
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -3,7 +3,7 @@ import warnings
 from pathlib import Path
 from typing import AbstractSet, Any, Dict, List, Mapping, Optional, Union
 
-from .fields import ModelField
+from .fields import FieldInfo, ModelField
 from .main import BaseModel, Extra
 from .typing import display_as_type
 from .utils import deep_update, sequence_like
@@ -65,7 +65,7 @@ class BaseSettings(BaseModel):
             d[field.alias] = env_val
         return d
 
-    class Config:
+    class Config(BaseModel.Config):
         env_prefix = ''
         env_file = None
         validate_all = True
@@ -76,7 +76,9 @@ class BaseSettings(BaseModel):
         @classmethod
         def prepare_field(cls, field: ModelField) -> None:
             env_names: Union[List[str], AbstractSet[str]]
-            env = field.field_info.extra.get('env')
+            field_info_from_config = cls.get_field_info(field.name)
+
+            env = field_info_from_config.get('env') or field.field_info.extra.get('env')
             if env is None:
                 if field.has_alias:
                     warnings.warn(
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -1,5 +1,5 @@
 import os
-from typing import Dict, List, Set
+from typing import Dict, List, Optional, Set
 
 import pytest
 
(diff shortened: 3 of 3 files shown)


ojomio:
Could somebody take a look please?

pydantic/env_settings.py
```diff
@@ -76,7 +76,9 @@ class Config:
         @classmethod
         def prepare_field(cls, field: ModelField) -> None:
             env_names: Union[List[str], AbstractSet[str]]
-            env = field.field_info.extra.get('env')
+            field_info_from_config = cls.get_field_info(field.name)
+
+            env = field_info_from_config.get('env') or field.field_info.extra.get('env')
```
samuelcolvin, line 81: I'm not entirely clear what this order of priority means in practice.

E.g. if we set `env` on the parent and the child, does the child take priority? Does the same apply both when done via config and done using `Field(..., env='my_api_key')`?

Do we need to add documentation about this? We definitely need to add tests to confirm and lock the behaviour is as expected.

ojomio: So, child should definitely override the parent with the child config, and as far as I understand, parent declaration method is irrelevant, because, by the time of this statement, the configuration in the parent is already applied and expressed in `field.field_info.extra.get('env')
`

> Does the same apply both when done via config and done using Field(..., env='my_api_key')?

When declaring the same field in the child with `Field(`, you effectively override the whole field. 

I added test to confirm the behavior I expected, could you specify what other scenarios you want to get tested?

changes/1561-ojomio.md
```diff
@@ -0,0 +1,2 @@
+Allow descendant Settings models to override env variable names for the fields defined in parent Settings models with `env` in their `Config`.
```
samuelcolvin, line 1: this needs to be one line and one sentence.

pydantic/env_settings.py
samuelcolvin: can we add a test for something like

```py
def test_env_inheritance_config(env):
    class Parent(BaseSettings):
        a: str = Field(None, env='a_in_field')
        b: str = None
        c: str = None
        d: str = None

        class Config:
            fields = {
                'b': {'env': ['b_in_config']},
                'c': {'env': ['c_in_config']},
            }

    class Child(Parent):
        class Config:
            env_prefix = 'prefix_'
            fields = {
                'a': {'env': ['a_child_config']},
                'b': {'env': ['b_child_config']},
            }

    env.set('a_in_field', 'a_in_field')
    env.set('b_in_config', 'b_in_config')
    env.set('c_in_config', 'c_in_config')
    env.set('d', 'd')
    env.set('a_child_config', 'a_child_config')
    env.set('b_child_config', 'b_child_config')
    env.set('prefix_d', 'prefix_d')

    assert Child().dict() == ...
```
I'm just not 100% sure what would happen in each case, so best to add a test to be explicit.

samuelcolvin: maybe it's obvious and the test does exactly as expected, in which case you could consider it unnecessary, but since even I'm not exactly clear what would happen, better to add the test.

ojomio: Considering your snippet, I'd rather break it down into individual cases and name them correspondingly, this way it will be more readable and it is less likely that we 've missed something

samuelcolvin:
This is great, thank you.

Issue #1571, opened 2020-05-27
Allow aggregation of errors from multiple root validators


beezee:
# Allow aggregation of errors from multiple root validators

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /usr/local/lib/python3.7/site-packages/pydantic
               python version: 3.7.6 (default, Dec 30 2019, 19:38:28)  [Clang 11.0.0 (clang-1100.0.33.16)]
                     platform: Darwin-18.2.0-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']

```

It looks like this line of code is the only thing forcing the first root validator failure on a model to be the _only_ root validator failure to be reported - https://github.com/samuelcolvin/pydantic/blob/master/pydantic/main.py#L918

I know it's subjectively arguable that one root validator is all you need, but a) the fact that in this code we are iterating over a collection of root validators seems to conflict with the idea of limiting to a single root validator and b) this seems like an unnecessary limitation that would only serve to make the library less flexible for users.

Is this a deliberate choice? Are there dire consequences I'm not anticipating that would arise from aggregating _all_ failures for _all_ root validators, same as we do for those of the non-root variety?

I'd be happy to open a PR and update/add any tests/docs whatever other auxiliary changes that would accompany this change in behavior, but I wanted to ask about the motivation and whether such a PR would be welcome before putting in the work.

Thanks.

samuelcolvin:
PR, welcome. I guess it was just done for performance but was perhaps a mistake.

This will need to be prominently noted in the docs and changelog.

beezee:
Thanks @samuelcolvin happy to hear it.

I've opened the PR, would welcome your guidance on where you think this is worth calling out in the docs. My first impression was that this was not really a documented behavior prior, and since the change brings root validators in line with attribute validators, I'm having a hard time coming up with an intuitive way to highlight the new behavior.

beezee:
Linking to the PR https://github.com/samuelcolvin/pydantic/pull/1586

@samuelcolvin let me know if there's anything else I can do to help expedite approval and merge.

fsramalho:
I was exploring pydantic and just found that. Thanks @beezee for the PR to fix that 😃

PR #1586, opened 2020-05-31
aggregate root validation errors

beezee:
## Change Summary

This change enables aggregation of multiple root validation errors. 

It turns out deduplication of root validators is necessary in the metaclass, since it appears `__new__` is called twice for generics, once at class definition, and another when type arguments are applied (seems like semantically python is creating an on-the-fly subclass with the type variables unified,) the second of which results in all root validators defined on the class itself appearing in a base for the type-arg-unified subclass as well as on the type-arg-unified subclass itself, hence duplicating all of them. The issue is caught by current generic tests, but was masked by the short-circuiting (duplication of root validators was not apparent because root validation was halted after the first failure.)

Added test coverage to make sure multiple root validators can fail simultaneously.

## Related issue number

#1571 

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -42,6 +42,7 @@ from .utils import (
     generate_model_signature,
     lenient_issubclass,
     sequence_like,
+    unique_list,
     validate_field_name,
 )
 
@@ -288,27 +289,13 @@ class ModelMetaclass(ABCMeta):
             json_encoder = pydantic_encoder
         pre_rv_new, post_rv_new = extract_root_validators(namespace)
 
-        dedupe_pre_rv = set([])
-        pre_rv_unique = []
-        for rv in pre_root_validators + pre_rv_new:
-            if rv not in dedupe_pre_rv:
-                dedupe_pre_rv.add(rv)
-                pre_rv_unique.append(rv)
-
-        dedupe_post_rv = set([])
-        post_rv_unique = []
-        for rv in post_root_validators + post_rv_new:
-            if rv not in dedupe_post_rv:
-                dedupe_post_rv.add(rv)
-                post_rv_unique.append(rv)
-
         new_namespace = {
             '__config__': config,
             '__fields__': fields,
             '__field_defaults__': fields_defaults,
             '__validators__': vg.validators,
-            '__pre_root_validators__': pre_rv_unique,
-            '__post_root_validators__': post_rv_unique,
+            '__pre_root_validators__': unique_list(pre_root_validators + pre_rv_new),
+            '__post_root_validators__': unique_list(post_root_validators + post_rv_new),
             '__schema_cache__': {},
             '__json_encoder__': staticmethod(json_encoder),
             '__custom_root_type__': _custom_root_type,
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -287,13 +287,28 @@ class ModelMetaclass(ABCMeta):
         else:
             json_encoder = pydantic_encoder
         pre_rv_new, post_rv_new = extract_root_validators(namespace)
+
+        dedupe_pre_rv = set([])
+        pre_rv_unique = []
+        for rv in pre_root_validators + pre_rv_new:
+            if rv not in dedupe_pre_rv:
+                dedupe_pre_rv.add(rv)
+                pre_rv_unique.append(rv)
+
+        dedupe_post_rv = set([])
+        post_rv_unique = []
+        for rv in post_root_validators + post_rv_new:
+            if rv not in dedupe_post_rv:
+                dedupe_post_rv.add(rv)
+                post_rv_unique.append(rv)
+
         new_namespace = {
             '__config__': config,
             '__fields__': fields,
             '__field_defaults__': fields_defaults,
             '__validators__': vg.validators,
-            '__pre_root_validators__': pre_root_validators + pre_rv_new,
-            '__post_root_validators__': post_root_validators + post_rv_new,
+            '__pre_root_validators__': pre_rv_unique,
+            '__post_root_validators__': post_rv_unique,
             '__schema_cache__': {},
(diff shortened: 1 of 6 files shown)


beezee:
Appreciate any suggestions about what documentation to update.

It seems to me the behavior at present is not currently documented, and since the change brings root validators in line with the behavior of attribute validators, I'm having a hard time coming up with an intuitive way to highlight the new behavior.

Is there a place where we call this behavior (aggregation of errors) out for attribute validators explicitly? Perhaps changing the wording to reflect _all_ validators there would be the most impactful way to communicate the new behavior.

samuelcolvin:
I would add something to https://pydantic-docs.helpmanual.io/usage/validators/#root-validators

Just a note (`!!! note`) saying "all will be call regardless" would be useful. Also worth saying root validators can't know whether a previous validator failed (that's the case as far as i know right?).

beezee:
I think this might be captured by changing the language " root validators by default will be called even if field validation fails; this behaviour can be changed by setting the skip_on_failure=True keyword argument to the validator." to " root validators by default will be called even if _prior_ validation fails; this behaviour can be changed by setting the skip_on_failure=True keyword argument to the validator."

The new language expands the scope of "things that can fail prior to a given root validator without preventing that root validator from running", from an original scope of field validators, to a new scope of field validators _plus_ any other root validators that run prior to the root validator in question. It also correctly captures that `skip_on_failure` will be honored regardless which of these sources the prior failure comes from.

That said if you think the note is worthwhile just for the sake of higher visibility, I'm happy to put one in addition, I can see that the language change I'm suggesting is subtle despite accuracy.

To your last point, while the library doesn't do anything for a user in terms of tracking success or failure of prior root validators, there's technically nothing stopping someone from attaching extra data to the underlying values dictionary as they return updated dictionaries from their root validators, allowing subsequent validators to check for presence of these values as signal of success, or absence as signal of failure. From the view of pydantic, I'd say we're un-opinionated in this respect, it's not impossible but we're not helping you do it in any meaningful way. 

Open to your thoughts there and happy to include whatever language around this you might feel worthwhile, fwiw I wouldn't find it problematic to not mention at all given the passive role the library has with respect to this nuance.

pydantic/main.py
```diff
@@ -287,13 +287,28 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
         else:
             json_encoder = pydantic_encoder
         pre_rv_new, post_rv_new = extract_root_validators(namespace)
+
+        dedupe_pre_rv = set([])
```
samuelcolvin, line 291: why is this required? Maybe it's a separate issue.

beezee: I explained this in my original "essay" - "It turns out deduplication of root validators is necessary in the metaclass, since it appears __new__ is called twice for generics, once at class definition, and another when type arguments are applied (seems like semantically python is creating an on-the-fly subclass with the type variables unified,)"

The rest of the paragraph goes into further detail describing that this is a current bug in pydantic which is hidden by the current behavior of stopping on the first root validator to fail. In short, you are currently duplicating root validators on generic models.

I gotta say, I can appreciate you maintain this project in addition to your whole real life, and that you do it probably for free, but please consider that my contribution here is under _exactly_ those same circumstances. Your last comment comes off hostile, I'm not really sure why that's warranted.

beezee:
Honestly not a response I was expecting - I don't know if you intend hostility here, but this doesn't come off as "contributions welcome" to me

samuelcolvin:
Sorry it if it seemed hostile, it wasn't meant as such. Just that I'm dyslexic and long responses take me a long time to parse, as you can see from the volume of issues on pydantic, avoiding such time sinks is helpful.

I actually skim-read your message above, but missed the part about duplication. That's my fault, but also kind of an example of what I was saying.

I'll look into this more soon and let you know what I think.

beezee:
Thanks and sorry if I read wrong. I'll take your feedback to heart and continue striving for brevity without sacrificing precision.

pydantic/main.py
samuelcolvin, line 291:
```suggestion
        dedupe_pre_rv = set()
```

samuelcolvin: might be worth adding a method for this to utils, something like:

```py
def unique_list(input_list: Union[List[T], Tuple[T, ...]]) -> List[T]:
    """
    Make a list unique while maintaining order.
    """
    result = []
    unique_set = set()
    for v in input_list:
        if v not in unique_set:
            unique_set.add(v)
            result.append(v)

    return result
```

Then below you can just use

```py
            '__pre_root_validators__': unique_list(pre_root_validators + pre_rv_new)
```

fsramalho: Perhaps the `more_itertools.unique_everseen` works for that job and should be more performatic as well.
https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.unique_everseen

samuelcolvin: not sure why it should be more performant? I don't think we can add another dependency for two function calls.

StephenBrown2: Especially since the code for that function is nearly identical to what @samuelcolvin posted above:
```
    seenset = set()
    seenset_add = seenset.add
    seenlist = []
    seenlist_add = seenlist.append
    iterable, keys = tee(iterable)
    for element, k in zip(iterable, map(key, keys) if key else keys):
        try:
            if k not in seenset:
                seenset_add(k)
                yield element
        except TypeError:
            if k not in seenlist:
                seenlist_add(k)
                yield element
```
Could just copy it in as a util: https://more-itertools.readthedocs.io/en/stable/_modules/more_itertools/recipes.html#unique_everseen

beezee: Updated to include util function and use in place of previously inlined logic.

samuelcolvin:
awesome, thank you.

PR #1619, opened 2020-06-09
Update model_config.md

Kludex:
Fix single typo.


StephenBrown2:
As this is a UK vs US-ism, I don't think it's a typo:
> **Behavior** ([American English](https://en.wikipedia.org/wiki/American_English)) or **behaviour** ([British English](https://en.wikipedia.org/wiki/British_English); see [spelling differences](https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-our,_-or))

Ref: https://en.wikipedia.org/wiki/Behavior

Source: I'm an American. 🇺🇸

Issue #1623, opened 2020-06-10
constrained set type "conset"


patrickkwang:
# Feature Request

> Is there a way to benefit from the best of both worlds: [uniqueness] from set and length constraints from conlist()? - _@neimad1985 in https://github.com/samuelcolvin/pydantic/issues/1472#issue-611410387_

Also important to me is that sets are unordered.

---

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: .../venv/lib/python3.8/site-packages/pydantic
               python version: 3.8.1 (default, Feb 19 2020, 09:57:06)  [Clang 11.0.0 (clang-1100.0.33.17)]
                     platform: macOS-10.15.5-x86_64-i386-64bit
     optional deps. installed: []
```

Example behavior:

```py
from pydantic import conset

class ConSetModel(BaseModel):
    v: conset(int, max_items=3)

print(ConSetModel(v=[1, 2, 2]).v)     # {1, 2}
print(ConSetModel(v=[1, 2, 2, 3]).v)  # {1, 2, 3}
print(ConSetModel(v=[1, 2, 3, 4]).v)  # raises ValidationError
```

samuelcolvin:
👍 PR welcome.

samuelcolvin:
fixed by #1627

PR #1627, opened 2020-06-11
Add conset

patrickkwang:
## Change Summary

This adds `conset()`. The functional additions are exactly analogous to the machinery for `conlist()`. The tests are also exactly analogous to those for `conlist()`, with the addition of [a single assertion](https://github.com/samuelcolvin/pydantic/pull/1627/files#diff-e1bc6e780c4b78fd3a4a85bf7b8ae0acR1742) ensuring that length checks are occurring _after_ casting the input to a set, i.e. `set([1, 1, 1, 2, 2])` satisfieds `max_items=4`.

## Related issue number

#1623

## Checklist


patrickkwang:
I'm puzzled. The checks that I fixed in https://github.com/samuelcolvin/pydantic/pull/1627/commits/7dcc248c5e34098ec955f41935dced442ad268df are still failing, as if nothing changed. Any ideas?

samuelcolvin:
It looks like linting is failing because lines are too long.

Nothing stranger I can see.

patrickkwang:
If I'm reading things correctly, the latest checks indicate that types.py:8 has 121 characters:
https://github.com/samuelcolvin/pydantic/pull/1627/checks?check_run_id=763653036#step:5:9

But the current diff shows that types.py:8 has 20 characters:
https://github.com/samuelcolvin/pydantic/pull/1627/files#diff-470d0730db2e82eaaf27ad3655a667b3R8

I must be misinterpreting something.

samuelcolvin:
not sure what was wrong with linting, it's now failing on something else.

changes/1623-patrickkwang.md
```diff
@@ -0,0 +1 @@
+Add conset(), analogous to conlist()
```
samuelcolvin, line 1:
```suggestion
Add `conset()`, analogous to `conlist()`
```

tests/test_schema.py
```diff
@@ -1731,6 +1732,57 @@ class Model(BaseModel):
...
+                'items': {'type': 'string'},
+                'uniqueItems': True,
+                'minItems': 1,
+                'maxItems': 4,
+            },
+        },
+        'required': ['foo'],
+    }
+
+    with pytest.raises(ValidationError) as exc_info:
```
samuelcolvin, line 1774: this test should not be part of `test_schema.py`

tests/test_schema.py
```diff
@@ -1731,6 +1732,57 @@ class Model(BaseModel):
...
+        Model(foo=[1, 'x', 'y'])
+    assert exc_info.value.errors() == [
```
samuelcolvin, line 1776: I think this is brittle, since converting to a set is performed before the `int` check, thus the indexes might not be 1 and 2.

Hence the test failure.

tests/test_schema.py
patrickkwang: I put this here because this is where the analogous `test_conlist()` is (just above). Should they both go in `test_types.py`?

samuelcolvin: I guess so, yes.

samuelcolvin:
awesome, thank you.

tests/test_types.py
```diff
@@ -202,6 +204,189 @@ class ConListModel(BaseModel):
...
+    class ConSetModelMax(BaseModel):
+        v: conset(int) = []
+
+    m = ConSetModelMax(v=[1, 2, 3])
+    assert m.v == {1, 2, 3}
+
+
+def test_constrained_set_default():
+    class ConSetModelMax(BaseModel):
+        v: conset(int) = {}
```
MrMrRobat, line 252: This test falils on my side indicating that is not `set`, but rather a `dict`.

samuelcolvin: Matter is failing, maybe that's why?

MrMrRobat: Matter?

samuelcolvin: Sorry, master.

samuelcolvin: This is very weird, as per #1682 this can be fixed by setting the default to a set.

However this should still pass since we don't have `validate_always=True` - the default value (an empty dict) should be okay.

The question is: what has changed on master since this was started that's causing this to fail when merged, and could that be a smoke signal to a bigger problem somewhere?

MrMrRobat: validate_always is getting set to `True` as per 8aebba2c35007b848aceed2f8ae55fd7b4888ab6

samuelcolvin: Yes I just saw that. I was investigating, then had to stop. I'll try to look later.

PR #1630, opened 2020-06-13
fix: make pydantic errors (un)pickable

### LINKED ISSUE #1616: Make ValidationError unpicklable
# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: False
                 install path: /home/badger/.cache/pypoetry/virtualenvs/antsibull-F3umzYPl-py3.8/lib/python3.8/site-packages/pydantic
               python version: 3.8.3 (default, Feb 26 2020, 00:00:00)  [GCC 9.3.1 20200408 (Red Hat 9.3.1-2)]
                     platform: Linux-5.6.15-200.fc31.x86_64-x86_64-with-glibc2.2.5
...
```

## Use case
I'm going to give you two code snippets because it might not be obvious from the simplest case why I would want to do it.

A simple approximation of my use case is here: https://gist.github.com/abadger/bfd55741c281ccb534f7bbc8fe9b6202

I am trying to use pydantic to validate and normalize data from a large number of data sources I need to run each validation separately so that I can know which data sources are providing invalid data. I decided to split it up amongst multiple CPUs by using asyncio's run_in_executor with a ProcessPoolExecutor.  However, when the pydantic.constr validation failed, I would get a BrokenProcessPool error on everything that had been queued but not run rather than a pydantic ValidationError on the specific task which failed.

## Root cause

I was able to workaround the problem by catching the pydantic exception and raising a ValueError with all of the information I needed.  This lead me to the root cause: pydantic errors are not **un**picklable.  Because of that, the exception raised in the worker process is pickled there and sent back to the parent process.  The parent process attempts to unpickle it, encounters the error, and then gives the generic, unhelpful BrokenProcessPool error and cancels the other pending tasks.

Here's a reproducer for the root cause:
```python
import pickle
from pydantic.errors import StrRegexError

p = pickle.dumps(StrRegexError(pattern='test'))
print(pickle.loads(p))

#   File "<stdin>", line 1, in <module>
# TypeError: __init__() missing 1 required positional argument: 'pattern'
```

Looking at the python stdlib bugtracker there are many open bugs with interactions between pickle and exceptions.  I didn't see this one so I added this: https://bugs.python.org/issue40917  Some others that might cause different bugs with pydantics exceptions:

* https://bugs.python.org/issue32696
* https://bugs.python.org/issue30005

Given so many potential bugs, I'm not sure if this is solvable in pydantic code or has to wait for pickle fixes.  However, if it's not solvable, adding my workaround and an explanation of what's happening to the docs would be nice.  That way searching for pydantic, ProcessPoolExecutor, pickle, multiprocessing might save the next person some time wondering why only a portion of their data was being converted.

PrettyWood:
## Change Summary

Custom *pydantic* errors (subclasses of `PydanticTypeError` and `PydanticValueError`) were not (un)pickable. Now they are!

## Related issue number

closes #1616

## Checklist


pydantic/errors.py
```diff
@@ -101,6 +112,9 @@ def __init__(self, **ctx: Any) -> None:
     def __str__(self) -> str:
         return self.msg_template.format(**self.__dict__)
 
+    def __reduce__(self):  # type: ignore
```
PrettyWood, line 115: @samuelcolvin I thought ignore the typing made more sense as it made code more complicated for nothing.
But here is the commit to add types https://github.com/PrettyWood/pydantic/commit/753abddfefd59d5d8eda7bbdeaef48d973157d36
If you want me to, I'll add it to this PR

abadger:
I have verified that this PR fixes the bug I reported in #1616  Thanks @PrettyWood !

abadger:
Oh, for the record.... I did not test the Cython version of pydantic with this PR applied.  I couldn't figure out how that would be installed locally.

pydantic/errors.py
```diff
@@ -91,6 +91,17 @@
 )
 
 
+def cls_kwargs(cls, kwargs):  # type: ignore
```
samuelcolvin, line 94:
```suggestion
def cls_kwargs(cls: Type[T], kwargs: Dict[str, Any) -> T:  # type: ignore
```
Or something?

Even if mypy can't handle it should still add type hints.

pydantic/errors.py
samuelcolvin: I think we should include type hints if at all possible

pydantic/errors.py
samuelcolvin: also, shouldn't this be `cls_kwargs(cls, ctx)`?

PrettyWood: You would like this method to be more generic with `Type[T]` and `ctx` for `args` and `kwargs` ? Sorry if I misunderstood

samuelcolvin: I mean:
1. we should have type annotations
2. i think it would make more sense to name the second argument `ctx`

PrettyWood: Alright ! I added the annotations that I already linked and renamed `kwargs` into `ctx`.

PR #1642, opened 2020-06-17
Use chain.from_iterable in class_validators.py

cool-RR:
This is a faster and more idiomatic way of using `itertools.chain`. Instead of computing all the items in the iterable and storing them in memory, they are computed one-by-one and never stored as a huge list. This can save on both runtime and memory space.


samuelcolvin:
Seems reasonable, please could you add a change description, also are we calling `chain()` anywhere else?

cool-RR:
Added change description. No, this is the only instance of this issue.

cool-RR:
@samuelcolvin Are we good for merging?

samuelcolvin:
Yes, I just have a lot on at the moment. I'll get to this PR when I have time.

PR #1648, opened 2020-06-21
Document default `regex` anchoring semantics (#1631)

yurikhan:
## Change Summary

Document that `regex` is checked using `re.match`. Suggest a way to write regular expressions that have the same meaning in Pydantic and JSON Schema.

## Related issue number

#1631 

## Checklist

* [x] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)

Issue #1631, opened 2020-06-14
change regex to use re.search or re.fullmatch


yurikhan:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.5.1
            pydantic compiled: False
                 install path: /home/khan/.local/lib/python3.6/site-packages/pydantic
               python version: 3.6.9 (default, Nov  7 2019, 10:44:02)  [GCC 8.3.0]
                     platform: Linux-4.15.0-88-generic-x86_64-with-Ubuntu-18.04-bionic
     optional deps. installed: ['typing-extensions']
```

Related: #1396 

```py
import jsonschema
import pydantic

class Foo(pydantic.BaseModel):
    bar: str = pydantic.Field(..., regex='baz')

try:
    Foo(bar='bar baz quux')
except pydantic.ValidationError as e:
    print(e)
    # ValidationError: 1 validation error for Foo
    # bar
    #   string does not match regex "baz" (type=value_error.str.regex; pattern=baz)
else:
    print('Valid')

try:
    jsonschema.validate({'bar': 'bar baz quux'}, Foo.schema())
except jsonschema.ValidationError as e:
    print(e)
else:
    print('Valid')
    # Valid
```

In JSON Schema, all versions so far, the regular expression in the `pattern` keyword is treated as [unanchored at both ends][1], i.e. `re.search` behavior rather than `re.match` or `re.fullmatch`.

[1]: https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-4.3

Pydantic uses `re.match` to validate strings with `Field` `regex` argument, as explained in #1396.

When constructing a JSON Schema from a model, Pydantic generates a `pattern` keyword from a `Field` `regex` argument, without any regex postprocessing. Thus, the resulting JSON Schema validates differently from the original Pydantic model.

I strongly feel Pydantic should follow suit and use `re.search` to validate fields with `regex` argument, and explicitly call this out in the documentation. (Anchors in example usage are not sufficient.)

Alternatively, if you are concerned with backward compatibility, I propose the following long-term solution:

1. Add a new `Field` (and `constr`) argument, perhaps named `pattern` after the JSON Schema keyword, mutually exclusive with the existing `regex` argument. This is API extension, thus, backward compatible.
2. Implement validation for `pattern` using `re.search`.
[...]

PR #1674, opened 2020-07-01
Fix typo (#1631)

yurikhan:
Fix typo in #1648.


StephenBrown2:
Thanks! Was a bit late on the review, but it stood out to me when I read it over.

PR #1679, opened 2020-07-02
Add private attributes support

### LINKED ISSUE #655: Question: add private attribute

MrMrRobat:
## Change Summary
Private attributes declared as regular fields, but always start with underscore and `PrivateAttr` is used instead of `Field`. Upon class creation they added in `__slots__` and `Model.__private_attributes__` for faster lookups and support for default values
Default values declared just like fields: `_attr[: annotation] = default_value` or `_attr = PrivateAttr(default_value)`

## Example from docs
```py
from datetime import datetime
from random import randint

class TimeAwareModel(BaseModel):
    _processed_at: datetime = PrivateAttr(default_factory=datetime.now)
    _secret_value: str = PrivateAttr()

    def __init__(self, **data):
        super().__init__(**data)
        self._secret_value = randint(1, 5)


m = TimeAwareModel()
print(m._processed_at)
print(m._secret_value)

```

## Other changes
Replaced `copy.deepcopy()` usage with new `pydantic.utils.smart_deepcopy()` for faster immutable values handling

## Related issue number
Closes #655

## Checklist


MrMrRobat:
@samuelcolvin, seems like tests merged from `master` were broken

Benoss:
I quite like this way of doing it

pydantic/main.py
```diff
@@ -363,11 +373,23 @@ def __setattr__(self, name, value):
...
+        }
 
     def __setstate__(self, state: 'DictAny') -> None:
-        object.__setattr__(self, '__dict__', state['__dict__'])
-        object.__setattr__(self, '__fields_set__', state['__fields_set__'])
+        object_setattr(self, '__dict__', state['__dict__'])
+        object_setattr(self, '__fields_set__', state['__fields_set__'])
+        self._set_private_attributes(state['__private_attributes_values__'], need_copy=False, check=False)
+
+    def _set_private_attributes(self, source: 'DictStrAny', need_copy: bool = True, check: bool = True) -> None:
```
samuelcolvin, line 387: Looks like the two kwargs are always used together, maybe they can be combined into `check_copy` or `from_pickle=False`?

This method might need a docstring too.

pydantic/main.py
```diff
@@ -363,11 +373,23 @@ def __setattr__(self, name, value):
...
+        for name, value in source.items():
+            if value is not Undefined:
+                object_setattr(self, name, smart_deepcopy(value) if need_copy else value)
+            elif check and not hasattr(self, name):
+                raise AttributeError(f'private attribute "{name}" is unset')
```
samuelcolvin, line 392: maybe if check is `True` we should do the check first so the exception message is always the same?

samuelcolvin: or make it identical here.

docs/examples/private_attributes.py
```diff
@@ -0,0 +1,47 @@
...
+from pydantic import BaseModel, Field
+
+storage = {}
+
+
+def get_new_id():
+    return max(storage) + 1 if storage else 0
+
+
+class BaseStorageModel(BaseModel):
```
samuelcolvin, line 12: maybe the `_processed_at` example at the top of #655 is simpler and clearer?

samuelcolvin (commented):
sorry it's been so long, I've been really busy.

This looks amazing, but might need to be split into multiple PRs.

pydantic/utils.py
```diff
@@ -477,3 +495,36 @@ def __get__(self, instance: Any, owner: Type[Any]) -> None:
...
+            raise NameError(
+                f'Private attributes "{name}" must not be a valid field name; '
+                f'Use sunder or dunder names, e. g. "_{name}" or "__{name}__"'
+            )
+
+
+Obj = TypeVar('Obj')
+
+
+def smart_deepcopy(obj: Obj) -> Obj:
```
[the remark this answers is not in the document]
MrMrRobat: You're right, separate PR is more sutable for this change.

> Or just works in all the cases we know of? 

Good catch. I think implementing faster behaviour only for built-in collections and immutable types is a way to go.

pydantic/main.py
```diff
@@ -363,11 +373,23 @@ def __setattr__(self, name, value):
...
+            '__fields_set__': self.__fields_set__,
+            '__private_attributes_values__': {k: getattr(self, k, Undefined) for k in self.__private_attributes__},
+        }
 
     def __setstate__(self, state: 'DictAny') -> None:
-        object.__setattr__(self, '__dict__', state['__dict__'])
-        object.__setattr__(self, '__fields_set__', state['__fields_set__'])
+        object_setattr(self, '__dict__', state['__dict__'])
+        object_setattr(self, '__fields_set__', state['__fields_set__'])
+        self._set_private_attributes(state['__private_attributes_values__'], need_copy=False, check=False)
```
MrMrRobat, line 385: @samuelcolvin, maybe we should use `dict.get()` here, so unpickling models dumped with older versions of `pydantic` would work?
```suggestion
        self._set_private_attributes(state.get('__private_attributes_values__', {}), need_copy=False, check=False)
```

pydantic/main.py
```diff
@@ -205,15 +202,19 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
...
-
         pre_root_validators, post_root_validators = [], []
-        for base in reversed(bases):
-            if _is_base_model_class_defined and issubclass(base, BaseModel) and base != BaseModel:
-                fields.update(deepcopy(base.__fields__))
+        slots: Union[str, Tuple[str, ...]] = namespace.get('__slots__', ())
+        private_attributes: Dict[str, Any] = dict.fromkeys((slots,) if isinstance(slots, str) else slots, Undefined)
+        validate_private_attributes(private_attributes)
+
+        for base in reversed(bases) if _is_base_model_class_defined else ():
```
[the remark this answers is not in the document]
MrMrRobat: Oof, unfortunately I don't remember. Should've commented it out 🤦 . But from what I see, now it includes `BaseModel` and it didn't before. I'll try to remember the reason I did it this way and either put explaning comment, or change the line back.

MrMrRobat (commented):
@samuelcolvin, thanks for review, I answered some comments and think I'll be able to split PR on the next weekend.

pydantic/main.py
samuelcolvin: Yes makes sense.

pydantic/utils.py
```diff
@@ -40,14 +41,31 @@
...
     'Representation',
     'GetterDict',
     'ValueItems',
     'version_info',  # required here to match behaviour in v1.3
     'ClassAttribute',
+    'ROOT_KEY',
 )
 
+ROOT_KEY = '__root__'
+IMMUTABLE_NON_COLLECTIONS_TYPES: AbstractSet[Type[Any]] = {
```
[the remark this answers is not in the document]
samuelcolvin: ok

samuelcolvin (commented):
Most things look good here, met me know if I've missed any questions.

samuelcolvin:
automatically generating `__slots__` would be great, I think this would make it much easier to use.

MrMrRobat:
What do you think about `PrivateField`? Should it subclass regular field class, be an independent class, or it's better to use `Field(private=True)`?

samuelcolvin:
I think probably easiest if it's a regular class.

I think we should keep it separate from `Field` because the point is that these attributes are NOT fields.

pydantic/main.py
MrMrRobat: Didn't get this one, could you explain?

pydantic/main.py
```diff
@@ -339,17 +357,18 @@ def __init__(__pydantic_self__, **data: Any) -> None:
...
             raise validation_error
-        object.__setattr__(__pydantic_self__, '__dict__', values)
-        object.__setattr__(__pydantic_self__, '__fields_set__', fields_set)
+        object_setattr(__pydantic_self__, '__dict__', values)
+        object_setattr(__pydantic_self__, '__fields_set__', fields_set)
+        __pydantic_self__._set_default_private_attributes(__pydantic_self__.__private_attributes__)
 
     @no_type_check
-    def __setattr__(self, name, value):
+    def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
```
MrMrRobat, line 368: had to set `# noqa: C901 (ignore complexity)` here because of today's changes in #1972. Is this the way to go, or some recent changes have to go outside of this method?

samuelcolvin: i think this is fine.

samuelcolvin:
@MrMrRobat this is looking good, do you think it will be ready to merge today?

I want to get v1.7 released, but I'm happy wait until tomorrow if we can include this?

MrMrRobat:
> do you think it will be ready to merge today?

Yeah, if don't have any comments or suggestions, it's ready :)

pydantic/main.py
```diff
@@ -339,17 +357,18 @@ def __init__(__pydantic_self__, **data: Any) -> None:
...
-            __pydantic_self__.__dict__: Dict[str, Any] = {}
-            __pydantic_self__.__fields_set__: 'SetStr' = set()
         values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
         if validation_error:
             raise validation_error
-        object.__setattr__(__pydantic_self__, '__dict__', values)
-        object.__setattr__(__pydantic_self__, '__fields_set__', fields_set)
+        object_setattr(__pydantic_self__, '__dict__', values)
+        object_setattr(__pydantic_self__, '__fields_set__', fields_set)
+        __pydantic_self__._set_default_private_attributes(__pydantic_self__.__private_attributes__)
```
samuelcolvin, line 365: why is this required?

docs/examples/private_attributes_underscore_attrs_are_private.py
```diff
@@ -0,0 +1,16 @@
...
+
+from pydantic import BaseModel
+
+
+class Model(BaseModel):
+    __class_var__: ClassVar[str] = 'class var value'
+    __private_attr__: str = 'private attr value'
+
+    class Config:
+        underscore_attrs_are_private = True
```
samuelcolvin, line 11: I this needs documenting in `model_config.md`. Also we need to be clear exactly what form names need to be in.

It looks from this like if you use `PrivateAttr` you can use `_whatever`, but with `underscore_attrs_are_private` you have to use `__whatever__`, but reading the code, I think that's not the case.

docs/examples/private_attributes.py
```diff
@@ -0,0 +1,18 @@
...
+from pydantic import BaseModel, PrivateAttr
+
+
+class TimeAwareModel(BaseModel):
+    _processed_at: datetime = PrivateAttr(default_factory=datetime.now)
+    _secret_value: str = PrivateAttr()
+
+    def __init__(self, **data):
+        super().__init__(**data)
+        self._secret_value = randint(1, 5)
```
samuelcolvin, line 13: maybe add a commit that this could also be done with `default_factory`

docs/examples/private_attributes_underscore_attrs_are_private.py
samuelcolvin: actually you've written it clear in the copy, but still better to use the same format in both code examples.

Is there anything special about `__whatever__` or is it just that it starts with an underscore, like `_whatever`?

samuelcolvin: Since I think according to some PEP we're not supposed to use dunder methods, unless they're official to python, maybe leave them out of the example and just mention them in the text?

pydantic/main.py
```diff
@@ -123,6 +126,7 @@ class BaseConfig:
     json_loads: Callable[[str], Any] = json.loads
     json_dumps: Callable[..., str] = json.dumps
     json_encoders: Dict[Type[Any], AnyCallable] = {}
+    underscore_attrs_are_private = False
```
samuelcolvin, line 129:
```suggestion
    underscore_attrs_are_private: bool = False
```
I know it's obvious, but best to be explicit.

samuelcolvin (commented):
looking good, just a few things to clarify.

docs/examples/private_attributes.py
MrMrRobat: you mean comment?

pydantic/main.py
MrMrRobat: what do you mean exactly?

docs/examples/private_attributes.py
samuelcolvin: _yes_

pydantic/main.py
samuelcolvin: Can we avoid this line since this is the critical path in terms of performance?

MrMrRobat: It sets defaults of `__private_attributes__`, without this line `default` not `default_factoru` wouldn't work.
What's critical here in terms of performance? Function call itself or executing code of the function?

samuelcolvin: Ok, I guess we need it.

Initialising/validating models is the critical path, I don't want to slow that down unless we absolutely have to.

pydantic/main.py
```diff
@@ -385,11 +404,24 @@ def __setattr__(self, name, value):
...
     def __setstate__(self, state: 'DictAny') -> None:
-        object.__setattr__(self, '__dict__', state['__dict__'])
-        object.__setattr__(self, '__fields_set__', state['__fields_set__'])
+        object_setattr(self, '__dict__', state['__dict__'])
+        object_setattr(self, '__fields_set__', state['__fields_set__'])
+        for name, value in state.get('__private_attribute_values__', {}).items():
+            if value is not Undefined:
+                object_setattr(self, name, value)
+
+    def _set_default_private_attributes(self, source: Dict[str, PrivateAttr]) -> None:
```
samuelcolvin, line 420:
```suggestion
    def _init_private_attributes(self, source: Dict[str, PrivateAttr]) -> None:
```
maybe?

More concise and explains it well I think.

samuelcolvin (commented):
I think this is looking amazing, thank you @MrMrRobat.

I've tried it myself and i think it'll be extremely useful.

I have one small proposed change in name, but only other question is on `model.copy()` which currently doesn't copy private attributes, but instead initialises from the default or default factory. I think I would expect `copy()` to (deep?)copy the private attributes, a bit like `__setstate__`.

What do you think?

samuelcolvin:
you need to reference the new name, sorry I should have made that clear in my change proposal.

Will you change it or will I?

MrMrRobat:
>  I think I would expect copy() to (deep?)copy the private attributes, a bit like __setstate__.

Yeah, you right. Working on it.

MrMrRobat:
> Working on it.

Done

samuelcolvin:
awesome, thank you.

I'm really looking forward to using this.

samuelcolvin:
Question on this in https://github.com/pydantic/pydantic/discussions/4518.

PR #1545, opened 2020-05-21
check ModelField().validate_always when inheriting

dcHHH:
## Change Summary

Indeed, the bug is originated from the change main.py:169. The fields of C is a deepcopy of it's parent class, so C.value_b.validate_always = B.value_b.validate_always = False
https://github.com/samuelcolvin/pydantic/blob/1eeb225c67f1131f948c5f129f5c2e56a4c2c6b7/pydantic/main.py#L881-L883
So value_b is not validated. Just move the assignment of validate_always from self.prepare() to self.populate_validators(), then the bug is solved. 

## Related issue number

https://github.com/samuelcolvin/pydantic/issues/1155

## Checklist

* [x] Unit tests for the changes exist
* [x] Tests pass on CI and coverage remains at 100%
* [ ] Documentation reflects the changes where applicable
* [ ] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)

PR #1682, opened 2020-07-04
Fix master

samuelcolvin:
Trying to fix master.

### DIFF
--- a/changes/1682-samuelcolvin.md
+++ b/changes/1682-samuelcolvin.md
@@ -1,0 +1,1 @@
+Modify validators for `conlist` and `conset` to not have `always=True`.
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -129,7 +129,7 @@ def read_env_file(file_path: Path, *, encoding: str = None, case_sensitive: bool
     except ImportError as e:
         raise ImportError('python-dotenv is not installed, run `pip install pydantic[dotenv]`') from e
 
-    file_vars: Dict[str, Optional[str]] = dotenv_values(file_path, encoding=encoding)
+    file_vars: Dict[str, Optional[str]] = dotenv_values(file_path, encoding=encoding or 'utf8')
     if not case_sensitive:
         return {k.lower(): v for k, v in file_vars.items()}
     else:
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -451,10 +451,7 @@ class ModelField(Representation):
             get_validators = getattr(self.type_, '__get_validators__', None)
             if get_validators:
                 self.class_validators.update(
-                    {
-                        f'list_{i}': Validator(validator, pre=True, always=True)
-                        for i, validator in enumerate(get_validators())
-                    }
+                    {f'list_{i}': Validator(validator, pre=True) for i, validator in enumerate(get_validators())}
                 )
 
             self.type_ = self.type_.__args__[0]
@@ -464,10 +461,7 @@ class ModelField(Representation):
             get_validators = getattr(self.type_, '__get_validators__', None)
             if get_validators:
                 self.class_validators.update(
-                    {
-                        f'set_{i}': Validator(validator, pre=True, always=True)
-                        for i, validator in enumerate(get_validators())
-                    }
+                    {f'set_{i}': Validator(validator, pre=True) for i, validator in enumerate(get_validators())}
                 )
 
             self.type_ = self.type_.__args__[0]
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -239,6 +239,16 @@ def test_conlist():
     assert exc_info.value.errors() == [{'loc': ('foo',), 'msg': 'value is not a valid list', 'type': 'type_error.list'}]
 
 
+def test_conlist_wrong_type_default():
+    """It should not validate default value by default"""
+
+    class Model(BaseModel):
+        v: conlist(int) = 'a'
+
+    m = Model()
+    assert m.v == 'a'
+
+
 def test_constrained_set_good():
     class ConSetModelMax(BaseModel):
         v: conset(int) = []
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -250,19 +250,27 @@ def test_conlist_wrong_type_default():
 
 
 def test_constrained_set_good():
-    class ConSetModelMax(BaseModel):
+    class Model(BaseModel):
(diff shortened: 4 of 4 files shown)


MrMrRobat:
`validate_always` is getting set to `True` because of https://github.com/samuelcolvin/pydantic/commit/8aebba2c35007b848aceed2f8ae55fd7b4888ab6#diff-1ef405a6b5a2fa9f20ea7e1f92e4c38eR510 combined with https://github.com/samuelcolvin/pydantic/commit/dca98559d58b24627a8da543a688549b7c91b459#diff-1ef405a6b5a2fa9f20ea7e1f92e4c38eR468

MrMrRobat:
The strange part is that tests were passing on windows somehow. This potentially means that pydantic could behave differently on this platform

PrettyWood:
@samuelcolvin Do you want some help to investigate this matter?

samuelcolvin:
This is preventing v1.6 from being released.

Any help on what's going wrong would be useful, I'll try and look later today.

MrMrRobat:
I see this as a plan of investigating and fixing the issue:
* [X] ~Find why tests were broken~ — done in https://github.com/samuelcolvin/pydantic/pull/1682#issuecomment-653760235
* [ ] Find out why this changes didn't break tests on windows, fix if needed
* [ ] Restore expected behaviour while maintaining features introduced in #1545 and #1627

Unfortunately, I don't have IDE installed on windows to investigate, if someone could help with that, it would be great.

PrettyWood:
So the real problem is that we have a "regression" with https://github.com/samuelcolvin/pydantic/pull/1545 where
```python
def test_conlist():
    class Model(BaseModel):
        v: conlist(int) = 'a'

    m = Model()
    assert m.v == 'a'
```
used to be valid with `1.5.1` because `validate_always` would not be set to `True` but now it does.
Do we actually need the `always=True` in `conlist` and recently added `conset` ?

PrettyWood:
Meh I just tried on a windows 10 and python 3.8.2:
```python
tests\test_types.py:249 (test_constrained_set_default)
def test_constrained_set_default():
        class ConSetModelMax(BaseModel):
            v: conset(int) = {}
    
>       m = ConSetModelMax()
```
Would be nice to be able to connect via SSH on the windows 2019 server machine of github

MrMrRobat:
Well, I looked at CI output closer and it noticed that tests actually failed, but CI did not detect it for some reason:
(and yeah, `test_env_file_custom_encoding` fails on windows ever since it was introduced: https://github.com/samuelcolvin/pydantic/runs/761260990)
```py
================================== FAILURES ===================================
________________________ test_env_file_custom_encoding ________________________

tmp_path = WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_env_file_custom_encoding0')

    @pytest.mark.skipif(not dotenv, reason='python-dotenv not installed')
    def test_env_file_custom_encoding(tmp_path):
        p = tmp_path / '.env'
        p.write_text('pika=p!�@', encoding='latin-1')
    
        class Settings(BaseSettings):
            pika: str
    
        with pytest.raises(UnicodeDecodeError):
>           Settings(_env_file=str(p))
E           Failed: DID NOT RAISE <class 'UnicodeDecodeError'>

tests\test_settings.py:557: Failed

________________________ test_constrained_set_default _________________________

    def test_constrained_set_default():
        class ConSetModelMax(BaseModel):
            v: conset(int) = {}
    
>       m = ConSetModelMax()

tests\test_types.py:254: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

__pydantic_self__ = ConSetModelMax(), data = {}, values = {}, fields_set = set()
validation_error = ValidationError(model='ConSetModelMax', errors=[{'loc': ('v',), 'msg': 'value is not a valid set', 'type': 'type_error.set'}])

        """
        Create a new model by parsing and validating input data from keyword arguments.
    
        Raises ValidationError if the input data cannot be parsed to form a valid model.
        """
        if TYPE_CHECKING:
            __pydantic_self__.__dict__: Dict[str, Any] = {}
            __pydantic_self__.__fields_set__: 'SetStr' = set()
        if validation_error:
E           pydantic.error_wrappers.ValidationError: 1 validation error for ConSetModelMax
E           v
E             value is not a valid set (type=type_error.set)

pydantic\main.py:346: ValidationError
=========================== short test summary info ===========================
FAILED tests/test_settings.py::test_env_file_custom_encoding - Failed: DID NO...
FAILED tests/test_types.py::test_constrained_set_default - pydantic.error_wra...
================= 2 failed, 1707 passed, 9 skipped in 27.38s ==================
mingw32-make: *** [Makefile:42: test] Error 1

PrettyWood:
Thanks for checking @MrMrRobat. It's reassuring! I was afraid it was ok only on some windows systems.
So the next question is "How do we fix the CI on windows to **actually** fail?"

samuelcolvin:
this is really weird, I've done quite a lot of digging but can't find anyone else having this problem with steps passing when `make` commands fail.

I think it must be some weird way gnu make and powershell are not working correctly together, but I can't find any clues about what's wrong.

PrettyWood:
@samuelcolvin Don't you think we could ask someone from github (actions) to have a look ? Nothing has changed on pydantic side regarding the CI. The exit code is 1 and fails on local. And we are talking about `make`, which is used like in every CI and has been stable for so long. I don't know if something has changed on github action side though. Maybe a new windows OS or a new binary for gnu make ? I don't think you can connect via SSH on the machine to have a deeper look (like on CircleCI) but if you can maybe it's worth a try too?

Kilo59:
@samuelcolvin @PrettyWood 
Could pydantic move away from `make` and start using `invoke`?
Then we don't have to rely on gnu make at all.
http://www.pyinvoke.org/

PrettyWood:
Thanks for sharing @Kilo59. But I don't think the issue is to know if `make` should be used or not but first to understand why the CI fails. Samuel said it **could** come from `make` but nothing is sure! And anyway if there is an issue here, it might  well happen on other projects so it could be nice to dig into it and understand it. There are indeed **a lot** of projects that use `make` (personally I use it in all my python projects!)

samuelcolvin:
@MrMrRobat @PrettyWood thank you both for your help. I'm working on the windows problem now on #1700

PR #1686, opened 2020-07-05
Add pickle support to dynamically created models and generics 

### LINKED ISSUE #1667: How to pickle a generic pydantic class?
Hi,

I am trying to pickle a pydantic generic model as illustrated below, using Python 3.7.6 and Pydantic 1.5.1:

```
import pickle

T = TypeVar('T')


class MyClass(BaseModel):
    pass


class GenericClass(GenericModel, Generic[T]):
    value: T


obj = MyClass()
instance = GenericClass[MyClass](value=obj)

pickle.dumps(instance)

```

This piece of code generates this error:
```
_pickle.PicklingError: Can't pickle <class '__main__.GenericClass[MyClass]'>: 
attribute lookup GenericClass[MyClass] on __main__ failed
```

It seems to be an error related specifically to pydantic generic models, as I am able to pickle non generic pydantic objects and generic objects that are not pydantic.

Am I doing something wrong or is this not supported yet? 
Thanks for your help!

### LINKED ISSUE #1734: Dynamically create_model that contains forward refs
# Question

Hi, I am asking this as a question but I feel it could be a bug. 

In my usecase I actually get type declarations (potentially with cycle deplendencies, optional, ...) dynamically at runtime.
I am building a static dict of those types and when it is ready, I will create the models with `create_model()` and apply `update_froward_refs()`.

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /Users/A492YC/.virtualenvs/sam-quote-ujUy7lE3/lib/python3.7/site-packages/pydantic
               python version: 3.7.4 (default, Feb 17 2020, 12:06:24)  [Clang 11.0.0 (clang-1100.0.33.17)]
```

I think that this type dict would be enough to play the role of localns.
But apparently, the function call `update_froward_refs()` fails because such models do not have `cls.__module__`

A simplified implementation:

```py
import pydantic

test_types = {
    'Sub': {
        "name": (str, ...),
        "param": (str, ...)
    },

    'Main': {
        "name": (str, ...),
        "male": (bool, ...),
        "sub": ("Sub", ...)
    }
}


sub = test_types["Sub"]
sub_model = pydantic.create_model("Sub", **sub)
sub_instance = sub_model(**{
    "name": "hello",
    "param": "there"
})

main = test_types["Main"]
main_model = pydantic.create_model("Main", **main)
main_instance = main_model(**{
    "name": "world",
    "male": True,
    "sub": {
        "name": "hello",
        "param": "there"
    }
})

#  File "<input>", line 31, in <module>
#   File "pydantic/main.py", line 336, in pydantic.main.BaseModel.__init__
#   File "pydantic/main.py", line 861, in pydantic.main.validate_model
# pydantic.errors.ConfigError: field "sub" not yet prepared so type is still a ForwardRef, you might need to call Main.update_forward_refs().
```

Because of the error above, I instead try to update_forward_refs on my dynamic model

```py
main_model.update_forward_refs(Sub=sub_model)

#   File "<input>", line 1, in <module>
#   File "pydantic/main.py", line 663, in pydantic.main.BaseModel.update_forward_refs
# KeyError: None
```

Could you provide a way to implement this if I am missing something?

MrMrRobat:
## Change Summary
- Always set `__module__` to created class (try to find the actual one, in case of failure fallback to `pydantic.main`)
- Set reference of created concrete model to it's module to allow pickling (not applied to models created in functions)
- ~Add `pydantic.utils.ensure_picklable` which will allow to make model, created inside a functions or/and referenced with a different name from class name pickable~

## Related issue number
Closes #1667
Closes #1734

## Checklist


PrettyWood:
I reckon your PR also fixes #1734 so #1757 can probably be closed. I let it open in the meantime

tests/test_utils.py
```diff
@@ -347,3 +349,18 @@ def test_all_literal_values():
...
+    def get_current_module_name():
+        return get_caller_module_name()
+
+    assert get_current_module_name() == __name__
+
+
+def test_get_caller_module_name_error():
+    with mock.patch('inspect.getmodule', return_value=None):
+        with pytest.raises(LookupError, match='Could not find caller module'):
+            get_caller_module_name()
```
PrettyWood, line 366:
```suggestion
def test_get_caller_module_name_error(mocker):
    mocker.patch('inspect.getmodule', return_value=None)
    with pytest.raises(LookupError, match='Could not find caller module'):
        get_caller_module_name()
```
Imo it's better to use directly `mocker` fixture and not import `unittest.mock`

MrMrRobat:
@PrettyWood, test I found in #1757 seems to work regardless of changes I made 🤔.
I assume example from #1734 will work as a test, as it reproduce the issue

PrettyWood:
@MrMrRobat weird 🤔 I just tried with [the test](https://github.com/samuelcolvin/pydantic/pull/1757/files#diff-eda3a99b42de19e6ab13f7b18b37dbb4R444) on `master` and it actually raises
```bash
>       globalns = sys.modules[cls.__module__].__dict__.copy()
E       KeyError: None

pydantic/main.py:684: KeyError
```

MrMrRobat:
Ah, sorry, now I get it. I just removed setting model attribute to a module, but kept setting of a module to a model

MrMrRobat:
So, update pickling can't be done if model has no global reference somewhere in code. Pickle threats path in  `__qualname__` of as such reference.

Dynamically created models don't have global references therefore can not be pickled, unless referenced manually. 

We can't set global references to each dynamically created model, as it will either override previous created model, or/and lead to memory leaks.

But turns out, it's not trivial task to determine if model was created globally or locally.

It can be done like this in normal conditions:
```py
def is_call_from_module():
    import inspect

    return inspect.stack()[2].function == '<module>'
```

However,  if code is executed in debugger, value of a `inspect.stack()[2].function` would be `evaluate_expression`. Obviously, we can't predict any situation like this.

So, we either need a stable way to detect such differences, or think about other implementations.

MrMrRobat:
`typing.Generic`, for example, doesn't create new classes for concrete types, but instead uses special class instances and therefore can be pickled. However I'm not sure how it can be used for pydantic models.

MrMrRobat:
I finally found the way to determine call from module, but it won't work with compiled code (as won't work `get_caller_module_name`)

The question is: do we need that functionality for non-compiled code, or we better stick with the way it will work the same on both versions? 

What will work in compiled and non-compiled code:
- Add default value to `create_model` argument `__module__=__name__` (`'pydantic.main'`)
- Add `ensure_picklable` function wicth will help to ensure that dynamically created models are pickable (all it actually does is create reference in `model.__module__` by `model.__name__`, if it's not used by another object)
```py
import pickle
from pydantic.utils import ensure_picklable

model = ensure_picklable(create_model('FooModel'))
pickle.dumps(model)  # works

BarModel = create_model('BarModel', __module__=__name__)
pickle.dumps(BarModel)  # works
```


Additional things that will work in non-compiled code:
- Determine in what module model was created, so this will work without specifying module:
```py
BarModel = create_model('BarModel')
pickle.dumps(BarModel) 
```

remidebette:
Hi! 

Is this blocked? Can we help?

Thank you

MrMrRobat:
@remidebette, I'm trying to figure out consistent way of solving the problem acroas all platforms. Hacky staff with frames seems to not do the job. Hope, I'll have a success with implementing this with `__reduce__` or `__reduce_ex__` method, but need more time for that.

Until issue is resolved, you can still fix such issues by using `create_model` only on module level, declaring variable with the name of the model and using `__module__` arg like this:
```py

MyModel = create_model('MyModel', __module__=__name__)
```
Such models won't have any problems with pickling/forward refs/etc.

For generics you have to manually set reference to base model module, like so:
```py

T = TypeVar('T')
class MyGenericModel(Generic[T]. GenericModel):
    t: T

IntModel = MyGenericModel[int]
setattr(MyGenericModel.__module__, IntModel .__name__, IntModel)
```
Code above might be implemented in your base classe's `__class_getitem__` like so:
```py
class MyAutoGenericModel(Generic[T]. GenericModel):
    t: T

    def __class_getitem__(cls, *args, **kwargs):
        created_model = super().__class_getitem__(*args, **kwargs)
        setattr(cls.__module__, created_model .__name__, created_model)
        return created_model

**Comment by remidebette** [comment]
Thanks for the detailed explanation!

File: tests/test_utils.py [hunk 2]
```diff
@@ -347,3 +348,58 @@ def test_all_literal_values():
...
+
+
+def test_get_caller_module_name_from_module(create_module):
+    create_module(
+        """
+from pydantic.utils import get_caller_module_name
+def get_current_module_name():
+    return get_caller_module_name()
+
+assert get_current_module_name() == __name__
```
samuelcolvin, line 367:
```suggestion
module_name = get_current_module_name()
assert module_name == __name__, (module_name, __name__)
```
so we can better understand the error?

samuelcolvin (commented):
This is looking good, let me know if I can do anything to help fix it.

I assume it's not as simple as moving `is_call_from_module` to `genetics.py` which is never compiled?

tests/test_utils.py
```diff
@@ -347,3 +348,58 @@ def test_all_literal_values():
...
+    assert get_current_module_name() == __name__
+
+
+def test_get_caller_module_name_not_found(mocker):
+    mocker.patch('inspect.getmodule', return_value=None)
+    assert get_caller_module_name() is None
+
+
+@pytest.mark.skipif(
+    compiled,
```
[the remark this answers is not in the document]
MrMrRobat: Wow, somehow I never thought about this. I did so and CI just passed! 

Just to be clear: compiled code may not contain any use of `GenericModel`, right?

tests/test_utils.py
MrMrRobat: doesn't pytest provide such introspection in case of error? 🤔

pydantic/utils.py
```diff
@@ -241,7 +243,8 @@ def unique_list(input_list: Union[List[T], Tuple[T, ...]]) -> List[T]:
 
 
 def update_normalized_all(
-    item: Union['AbstractSetIntStr', 'MappingIntStrAny'], all_items: Union['AbstractSetIntStr', 'MappingIntStrAny'],
+    item: Union['AbstractSetIntStr', 'MappingIntStrAny'],
```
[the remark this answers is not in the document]
MrMrRobat: seems like inconsistency between black versions. Which one should I use?

tests/test_utils.py
MrMrRobat: I forgot to apply suggestions before moving functions and tests to generics, so have to make changes manually

samuelcolvin:
this looks great, please could you:
* add a change description
* fix the coverage

MrMrRobat:
@samuelcolvin, master CI is failing again 😕. I guess PR fromatted with different black version was merged and broke CI.

samuelcolvin:
master should now be fixed.

MrMrRobat:
@samuelcolvin, FYI: `test_module` fixture could be used to test `try_assert.py` without having a separate command for that:

```py
def test_assert_in_validators(test_module):
    test_module("try_assert.py")
```

pydantic/main.py
```diff
@@ -812,8 +812,9 @@ def create_model(
     :param __model_name: name of the created model
     :param __config__: config class to use for the new model
     :param __base__: base class for the new model to inherit from
+    :param __module__: module for the new model, if None caller module will be used
```
MrMrRobat, line 815:
```suggestion
    :param __module__: module of the created model
```

PR #1695, opened 2020-07-07
feat(field): add dump_alias and load_alias options

### LINKED ISSUE #624: load_alias and dump_alias
Just realised I've referenced this on a few different issues #477, #576, and #585, but there's no issue dedicated to it.

I suggest we have 3 alias options on a field:
* `alias` used for loading and dumping
* `load_alias` only used for loading, takes precedence over `alias`
* `dump_alias` only used for dumping (`.dict()` and `.json()`), takes precedence over `alias`

PrettyWood:
## Change Summary

* [x] add dump_alias option
* [x] add load_alias option
* [x] generate schema for both serialization and deserialization 
* [ ] add documentation

## Related issue number

closes #624

## Checklist

### DIFF
--- a/changes/624-PrettyWood.md
+++ b/changes/624-PrettyWood.md
@@ -1,0 +1,1 @@
+add `dump_alias` and `load_alias` options for `Field` to be able to differentiate the field name used for serialization and deserialization
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -93,6 +93,12 @@ for validation, for use with `orm_mode`; see [Data binding](models.md#data-bindi
 **`alias_generator`**
 : a callable that takes a field name and returns an alias for it (see [the dedicated section](#alias-generator))
 
+**`dump_alias_generator`**
+: Same as `alias_generator` but only for `dump_alias`, which takes precedence over `alias` for serialization
+
+**`load_alias_generator`**
+: Same as `alias_generator` but only for `load_alias`, which takes precedence over `alias` for deserialization
+
 **`keep_untouched`**
 : a tuple of types (e.g. descriptors) for a model's default values that should not be changed during model creation and will
 not be included in the model schemas. **Note**: this means that attributes on the model with *defaults of this type*, not *annotations of this type*, will be left alone.
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -48,7 +48,9 @@ It has the following arguments:
 * `default_factory`: a zero-argument callable that will be called when a default value is needed for this field.
     Among other purposes, this can be used to set dynamic default values.
     It is forbidden to set both `default` and `default_factory`.
-* `alias`: the public name of the field
+* `alias`: the public name of the field (used for serialization and deserialization)
+* `dump_alias`: only used when dumping the model with `.dict()` and `.json()` (serialization), takes precedence over `alias`
+* `load_alias`: only used when loading data (deserialization), takes precedence over `alias`
 * `title`: if omitted, `field_name.title()` is used
 * `description`: if omitted and the annotation is a sub-model,
     the docstring of the sub-model will be used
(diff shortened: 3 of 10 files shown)


samuelcolvin (commented):
This is looking good.

Questions:
* are we satisfied it'll work correctly with inheritance? The inheritance landscape with aliases is already mind numbingly complicated, just this just slot in or confuse matters more?
* How do we deal with skipping/excluding from `dict()` etc.? I guess the simplest solution might be an object `Exclude` which can only be used on `dump_alias`.

PrettyWood:
@samuelcolvin This is the whole question and honestly I need to think it over. This is why it's just a draft I feel like we still need some feedback on the desired API. The best would be to have a bunch of tests with nested models and load/dump + exclude fields to have the expected behaviour for each one

daviskirk:
Just saw this while working on https://github.com/samuelcolvin/pydantic/issues/660 ... not sure if I'm out of lying trying to create a pull requests for that or if the discussion here has changed the view on it.
Especially the idea of having an "Exclude" object on the dump alias is a bit different compared to https://github.com/samuelcolvin/pydantic/issues/660#issuecomment-515406193.

That said, there are a few arguments for separating the two:

1. having separation of concerns: the "exclude" functionality with an explicit "exclude" option is probably needed anyway because we probably want to support the "advanced" exclude syntax: `t.dict(exclude={'user': {'username', 'password'}, 'value': ...})`. Having another way of excluding a field using the dump_alias might be confusing as there would now be two ways of doing this.

2. We might want to toggle exclusion separately from the dump alias, assuming we want to allow a case like this: Let's say we have a model

```
class TestModel(BaseModel):
    a: int = Field(0, dump_alias="a_dump", exclude=...)
    b: str = "foo"
    c: str = "bar"
```

would we want to allow the fields to be "included" by the `.dict` override??

```python
TestModel(a=3).dict(include={"a"})  # {"a_dump": 3}
```

There are cases I know where this might be the desired behavior.
  
There are also arguments against the second point: I can just as well think of cases where this might be confusing because the author of the class wants the field to be hidden permanently and not being randomly resurrected by a overriding using dict or json.  Especially confusing would might be a case like this:

```python
TestModel(a=3).dict()  # {"b": "foo", "c": "bar"}
# user decides they don't want to see "b"
TestModel(a=3).dict(exclude={"b"})  # {"a_dump": 3, "c": "bar"} ... what?? why did a_dump show up, it wasn't there before.  
```

Were there any thoughts on "dump_alias" vs. "exclude" in class config / field or am I just confusing a bunch of issues here?

samuelcolvin:
What if we didn't allow excluding via `dump_alias` and `load_alias` - just allowed the alias to be changed?

That way we get clear separation of concerns and a simple interface.

PrettyWood:
> What if we didn't allow excluding via `dump_alias` and `load_alias` - just allowed the alias to be changed?
> 
> That way we get clear separation of concerns and a simple interface.

I'm really in favor of this. I'll be back on this PR in the next few days. Thanks a lot @daviskirk for the suggestion. Sorry I missed your message before

daviskirk:
> I'm really in favor of this. I'll be back on this PR in the next few days. Thanks a lot @daviskirk for the suggestion. Sorry I missed your message before

https://github.com/samuelcolvin/pydantic/pull/2231 now has a working candidate (see: https://github.com/samuelcolvin/pydantic/pull/2231#issuecomment-782196223 in particular, in case that influences anything here)

PrettyWood:
@samuelcolvin I went back on this.
I added `load_alias` and `dump_alias` and also changed schema generation to decide whether serialization ou deserialization schema should be generated.
I'll first wait for your input before writing documentation.

PS: _fastapi_ tests fail because they use `ModelField.alias` but I reckon code can probably be updated on _fastapi_ side.

samuelcolvin:
Closing this.

Although we'll definitely support `load_alias` and `dump_alias` in V2, I doubt we'll use any of this logic.

PR #1700, opened 2020-07-09
windows CI wrongly passing

samuelcolvin:
Fix error with CI which caused failing windows tests to pass.

The problem here is that for some reason (presumably due to the way github actions pass commands to powershell on windows) multiline `run` steps can't fail. We had

```yaml
    - name: test
      run: |
        make test
        coverage xml
```

needed changing to

```yaml
    - name: test
      run: make test

    - run: coverage xml
```

I can't find this documented anywhere obvious, I bet lots of people have this problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,11 +111,11 @@ jobs:
       run: make mypy
 
     - name: test
-      run: |
-        make test
-        coverage xml
+      run: make test
 
-    - uses: samuelcolvin/codecov-action@env-vars
+    - run: coverage xml
+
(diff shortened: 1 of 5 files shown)


PrettyWood:
Seems like splitting steps fixes the issue. That's good to know.
Just out of curiosity did you figure out why exactly @samuelcolvin ?

samuelcolvin:
> Just out of curiosity did you figure out why exactly @samuelcolvin ?

I tried breaking linting, which failed correctly. The multi-line statement for the `test` step was the obvious difference.

PR #1712, opened 2020-07-12
fix: validate and parse nested models properly with `default_factory`

### LINKED ISSUE #1710: Pydantic 1.6 does not validate nested models when using a default_factory
# Bug

```
             pydantic version: 1.6
            pydantic compiled: True
                 install path: /usr/local/Caskroom/miniconda/base/envs/server/lib/python3.8/site-packages/pydantic
               python version: 3.8.3 (default, Jul  2 2020, 11:26:31)  [Clang 10.0.0 ]
```

I'm aware that [validation does not apply to default values](https://github.com/samuelcolvin/pydantic/issues/1488), but it appears that it is no longer applied to *user-provided* values that have a default factory value, either. If this is expected, I'm not sure if I saw it in the documentation or changelog. 

```py
from typing import List

class Child(BaseModel):
    x: int

class Parent(BaseModel):
    children: List[Child] = Field(default_factory=list)


p = Parent(children=[{'x':1}, {'y':2}])
```

In Pydantic 1.5.1, an error is raised because the second child fails to validate (it is missing required value 'x')

In Pydantic 1.6, the Parent object is instantiated successfully and contains two dicts:

```py
print(p)
# children=[{'x': 1}, {'y': 2}]
```

To avoid this and recover the expected behavior, `validate_all` must be set on the `Parent` class. However, since I'm providing the child value, I'm surprised that I had to ask for it to be validated.

### LINKED ISSUE #1717: Pydantic 1.6: fields with default_factory cause parse_obj() to set fields to unparsed values
# Bug

This is likely related to #1710 but does not explicitly involve validators. This would be a good candidate for a unit test to prevent future regressions.

```
pydantic version: 1.6
pydantic compiled: True
python version: 3.8.3 (default, May 16 2020, 07:08:28)  [GCC 8.3.0]
platform: Linux-4.15.0-109-generic-x86_64-with-glibc2.2.5
optional deps. installed: []
```
Here is a self-contained example illustrating the problem:
```py

class Inner(BaseModel):
    val: int = Field(0)

class Outer(BaseModel):
    inner_1: Inner = Field(default_factory=Inner)
    inner_2: Inner = Field(Inner())

default = Outer().dict() # default == {'inner_1': {'val': 0}, 'inner_2': {'val': 0}}
parsed = Outer.parse_obj(default)

print(parsed)
```
In 1.6, this prints:
`inner_1={'value': 'default'} inner_2=Inner(value='default')`
Note that `inner_1` is a dict, not an `Inner` object!

In 1.5.1 this prints:
`inner_1=Inner(val=0) inner_2=Inner(val=0)`

### LINKED ISSUE #1722: Error in find_validators (issubclass() arg 1 must be a class)
# Bug

```
             pydantic version: 1.6
            pydantic compiled: True
                 install path: <prefix>/env/lib/python3.8/site-packages/pydantic
               python version: 3.8.2 (default, Apr 27 2020, 15:53:34)  [GCC 9.3.0]
                     platform: Linux-5.4.0-39-generic-x86_64-with-glibc2.29

```

```py
from typing import List

from pydantic import BaseModel, Field, validator


class Schema(BaseModel):
    foo: List[str] = Field(default_factory=list)

    @validator("foo", pre=True, each_item=True)
    def mutate_foo(cls, v):
        return f"{v}-1"


class Derived(Schema):
    pass


print(Derived(foo=['bar']).foo)
```

This results in
```
  File "pydantic/validators.py", line 579, in pydantic.validators.find_validators

  File "ptest.py", line 14, in <module>
    class Derived(Schema):
  File "pydantic/main.py", line 231, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/fields.py", line 523, in pydantic.fields.ModelField.populate_validators
  File "pydantic/validators.py", line 588, in find_validators
RuntimeError: error checking inheritance of typing.List[str] (type: List[str])
```

This code sample works in pydantic 1.5.1.  The bug seems to rely on the combination of `default_factory`, the `@validator` method, and the subclass.  Removing any one of these causes the code to run correctly.

Note: there are a few other issues with the similar error message, but they appear to be distinct from this because they apply to versions <= 1.5.1 and have slightly different stack traces.

PrettyWood:
## Change Summary

PR #1504 introduced a regression by bypassing `populate_validators()`,
which would skip the validation and mess with parsing of nested models with `default_factory`

## Related issue number

closes #1710
closes #1717
closes #1722

## Checklist

### DIFF
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -274,6 +274,6 @@ if __name__ == '__main__':
     else:
         main()
 
-    if None in other_tests:
-        print('not all libraries could be imported!')
-        sys.exit(1)
+    # if None in other_tests:
+    #     print('not all libraries could be imported!')
+    #     sys.exit(1)
--- a/changes/1710-PrettyWood.md
+++ b/changes/1710-PrettyWood.md
@@ -1,0 +1,1 @@
+fix validation of nested models with `default_factory`
--- a/changes/1710-PrettyWood.md
+++ b/changes/1710-PrettyWood.md
@@ -1,1 +1,1 @@
-fix validation of nested models with `default_factory`
+fix validation and parsing of nested models with `default_factory`
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -341,8 +341,22 @@ class ModelField(Representation):
         e.g. calling it it multiple times may modify the field and configure it incorrectly.
         """
 
-        # To prevent side effects by calling the `default_factory` for nothing, we only call it
-        # when we want to validate the default value i.e. when `validate_all` is set to True.
+        self._set_default_and_type()
+        self._type_analysis()
+        if self.required is Undefined:
+            self.required = True
+            self.field_info.default = Required
+        if self.default is Undefined and self.default_factory is None:
+            self.default = None
+        self.populate_validators()
+
+    def _set_default_and_type(self) -> None:
+        """
+        Set the default value, infer the type if needed and check if `None` value is valid.
+
+        Note: to prevent side effects by calling the `default_factory` for nothing, we only call it
+        when we want to validate the default value i.e. when `validate_all` is set to True.
+        """
         if self.default_factory is not None:
             if self.type_ is None:
                 raise errors_.ConfigError(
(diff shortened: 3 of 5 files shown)


dgasmith:
Thanks for finding and fixing this so quickly! Is there anything we can do to expedite a 1.6.1 patch release? This issue has some pretty far reaching complications for our software stack.

samuelcolvin:
thanks so much. This looks good, I've tweaked some of the tests and added another one, otherwise I think this is good to go, will merge once it passes and release a patch.

samuelcolvin:
Thanks so much @PrettyWood for solving this so quickly and to all those who reported it.

I've just tagged a release, it should be available in pypi in a few minutes.

Sorry for the slightly delay.

PR #1737, opened 2020-07-17
feat(field): add `Enum` and `IntEnum` as valid types for fields

### LINKED ISSUE #1735: Doesn't correctly validate when type is Enum
# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /Users/<redacted>/.venvs/<redacted>/lib/python3.6/site-packages/pydantic
               python version: 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 03:03:55)  [GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
     optional deps. installed: []
```
Type validation works as expected when using children of Enums (e.g., [here](https://pydantic-docs.helpmanual.io/usage/types/#enums-and-choices)) but not when using Enums themselves.

```py
from enum import Enum

class FruitEnum(Enum):
    pear = 'pear'
    banana = 'banana'


class CookingModelA(BaseModel):
    fruit: FruitEnum


print(CookingModelA(fruit=FruitEnum.banana)) # works as expected


class CookingModelB(BaseModel):
    fruit: Enum


print(CookingModelB(fruit=FruitEnum.banana)) # fails validation with error
    """
    ValidationError: 1 validation error for CookingModelB
    fruit
      value is not a valid enumeration member; permitted:  (type=type_error.enum; enum_values=[])
    """
```
My expectation is that both should validate correctly, since `FruitEnum` is an instance of `Enum`.

PrettyWood:
_This is a proposal as I agree with the author of the issue. Feel free to close it if need be_

## Change Summary
Support `Enum` type as follows

```python
class Model(BaseModel):
    fruit: Enum
```

## Related issue number

closes #1735

## Checklist

### DIFF
--- a/pydantic/errors.py
+++ b/pydantic/errors.py
@@ -29,7 +29,7 @@ __all__ = (
     'UrlHostTldError',
     'UrlPortError',
     'UrlExtraError',
-    'EnumError',
+    'EnumMemberError',
     'IntegerError',
     'FloatError',
     'PathError',
@@ -216,7 +216,9 @@ class UrlExtraError(UrlError):
     msg_template = 'URL invalid, extra characters found after valid URL: {extra!r}'
 
 
-class EnumError(PydanticTypeError):
+class EnumMemberError(PydanticTypeError):
+    code = 'enum'
+
     def __str__(self) -> str:
         permitted = ', '.join(repr(v.value) for v in self.enum_values)  # type: ignore
         return f'value is not a valid enumeration member; permitted: {permitted}'
--- a/pydantic/errors.py
+++ b/pydantic/errors.py
@@ -29,6 +29,8 @@ __all__ = (
     'UrlHostTldError',
     'UrlPortError',
     'UrlExtraError',
+    'EnumError',
+    'IntEnumError',
     'EnumMemberError',
     'IntegerError',
     'FloatError',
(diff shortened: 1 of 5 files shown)


PrettyWood:
@zabzug-pfpt You're absolutely right! I even changed the attribute name to be more explicit. Thanks!

samuelcolvin (commented):
I agree that this is a good idea, but I would suggest implementing it in a different way to be more performant:

We should have a different validator specifically for `Enum` itself (and perhaps `IntEnum`?), that way the performance of the standard `enum_validator` shouldn't be effected at all?

pydantic/errors.py
```diff
@@ -216,7 +216,9 @@ class UrlExtraError(UrlError):
     msg_template = 'URL invalid, extra characters found after valid URL: {extra!r}'
 
 
-class EnumError(PydanticTypeError):
+class EnumValueError(PydanticTypeError):
+    code = 'enum'
```
PrettyWood, line 222: Kept the same code to avoid breaking change with `fastAPI`

PrettyWood:
Thanks @samuelcolvin! PR updated. Feedback welcome!

PrettyWood:
Thanks @samuelcolvin! PR updated. Feedback welcome!

pydantic/errors.py
```diff
@@ -216,7 +218,9 @@ class UrlExtraError(UrlError):
     msg_template = 'URL invalid, extra characters found after valid URL: {extra!r}'
 
 
-class EnumError(PydanticTypeError):
+class EnumMemberError(PydanticTypeError):
+    code = 'enum'
```
PrettyWood, line 222: I kept the code `enum` to avoid breaking the tests of *fastAPI*.
I first named it `enum_member` to use `enum` and `int_enum` for added errors but I'm not sure it's worth a change. WDYT?

samuelcolvin:
👍

PR #1739, opened 2020-07-17
Add init for forward ref stub

SirTelemak:
## Change Summary

added `__init__` with python's 3.6 ForwardRef signature  

## Related issue number

In order to fix https://github.com/samuelcolvin/pydantic/issues/1738#issue-659667704 

## Checklist

* [+] Unit tests for the changes exist
* [+] Tests pass on CI and coverage remains at 100%
* [+] Documentation reflects the changes where applicable
* [+] `changes/<pull request or issue id>-<github username>.md` file added describing change


changes/1738-sirtelemak.md
```diff
@@ -0,0 +1 @@
+Fix: Add stub `__init__` with python 3.6 signature for ForwardRef
```
samuelcolvin, line 1:
```suggestion
Add stub `__init__` with python 3.6 signature for `ForwardRef`
```

SirTelemak:
> looking good, but please could you add a test too.

Sure

SirTelemak:
@samuelcolvin do you have any ideas how we can test this? Can't find how I can set `TYPE_CHECKING` as True for test using  `create_module` fixture

tests/mypy/modules/success.py
```diff
@@ -127,3 +128,9 @@ def foo(a: int, *, c: str = 'x') -> str:
 
 foo(1, c='thing')
 foo(1)
+
+if sys.version_info < (3, 7):
```
SirTelemak, line 132:
```suggestion
if sys.version_info < (3, 7):

```

tests/mypy/modules/success.py
```diff
@@ -127,3 +128,10 @@ def foo(a: int, *, c: str = 'x') -> str:
...
```
PrettyWood, line 132: I think you can just remove this check and test the `ForwardRef` for all python versions. Otherwise the test looks good

SirTelemak: Yea, I thought about it, but I think we have no need in testing build-ins so I made test only for implemented functionality

PrettyWood: I completely agree in general but I feel like this is one of the cases where it's worth testing it as the implemented code is here to mimic the builtin. But I feel like it's more a matter of taste than anything else ;)

SirTelemak: Yup, matter of taste, let's wait for anyone's else opinion then ;)

SirTelemak: @PrettyWood ForwardRef `__init__` signature different between versions: 
```python
def __init__(self, arg, is_argument=True):
```
for python 3.8, so we need make more complex test if we want cover all cases :)

samuelcolvin: I think best to remove this version check.

from what I can see the test here should still work with 3.8?

PR #1749, opened 2020-07-21
feat(schema): support custom title, description and default for enums

### LINKED ISSUE #1748: Field title and description being overridden for Enums.
# Bug

When using the Field() function with an Enum, I can set an alias, but if I try to set a title or description they are ignored for default values of both (the name of the Enum subclass for the title, and 'An enumeration' for the description).

```
             pydantic version: 1.6.1
            pydantic compiled: True
                 install path: /Users/joshourisman/.local/share/virtualenvs/forms-test-3oF3wXzB/lib/python3.8/site-packages/pydantic
               python version: 3.8.3 (default, Jun 30 2020, 11:16:06)  [Clang 11.0.0 (clang-1100.0.33.8)]
     optional deps. installed: []

```

```py
>>> from enum import Enum                          
>>> from pydantic import BaseModel, Field          
>>> class TestEnum(str, Enum): 
...     foo = 'foo' 
...     bar = 'bar'                                
>>> class TestModel(BaseModel): 
...     test: TestEnum = Field(alias='functioning_alias', title='I Will Be Ignored')                      
>>> TestModel.schema()                             
{'title': 'TestModel', 'type': 'object', 'properties': {'functioning_alias': {'$ref': '#/definitions/TestEnum'}}, 'required': ['functioning_alias'], 'definitions': {'TestEnum': {'title': 'TestEnum', 'description': 'An enumeration.', 'enum': ['foo', 'bar'], 'type': 'string'}}}
```

### LINKED ISSUE #1856: Field with Enum type does not respect schema override
# Bug

```
D:\Develop\pydantic>python -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.6.1
            pydantic compiled: False
                 install path: D:\Develop\pydantic\pydantic
               python version: 3.8.3 (tags/v3.8.3:6f8c832, May 13 2020, 22:37:02) [MSC v.1924 64 bit (AMD64)]
     optional deps. installed: []

```

```py
from enum import Enum
from typing import Optional

class Operation(Enum):
    """
    Enum that represents comparison operations.
    """
    MORE = ">"
    LESS = "<"


class Test(BaseModel):
    operation: Operation = Field(default=Operation.LESS,
                                 description="Operation used to compare something with something.")


print(Test.schema_json(indent=2))
"""
prints:

{
  "title": "Test",
  "type": "object",
  "properties": {
    "operation": {
      "$ref": "#/definitions/Operation"
    }
  },
  "definitions": {
    "Operation": {
      "title": "Operation",
      "description": "Enum that represents comparison operations.",
      "enum": [
        ">",
        "<"
      ]
    }
  }
}
"""
```
Its expected to have default value and description alongside with $ref.

Fix is pretty straightforward:
```diff
diff --git a/pydantic/schema.py b/pydantic/schema.py
index 27c66b2..ee8b331 100644
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -702,6 +702,10 @@ def field_singleton_schema(  # noqa: C901 (ignore complexity)
         enum_name = normalize_name(field_type.__name__)
         f_schema = {'$ref': ref_prefix + enum_name}
         definitions[enum_name] = enum_process_schema(field_type)
+        if not schema_overrides:
+            return f_schema, definitions, nested_models
+        else:
+            return {'allOf': [f_schema]}, definitions, nested_models
     else:
         add_field_type_to_schema(field_type, f_schema)

```

PrettyWood:
## Change Summary
We now have reusable enums thanks to #1432 but default title and description were not customizable. Now they are!
The behaviour has been homogenized by setting the `title` by default (as it is for all the other types).

ℹ️ Doc doesn't need any update as the schema is generated from `docs/examples/schema_main.py`

## Related issue number
closes #1748
closes #1856

## Checklist


PrettyWood:
@samuelcolvin @tiangolo I added the title on schema of enums since it's already added for all other types. It's not really a breaking change but still deserves an update on *fastapi* (see https://github.com/tiangolo/fastapi/pull/1765).
I could just add the title when explicitly set by the user but we would have different behaviours, which is imo a bad idea.
What are your thoughts? If this is approved how do you proceed to coordinate?
Thanks

mostaphaRoudsari:
Hi @PrettyWood, this is a great PR and I hope to get merged sooner than later.

I think you need to make a small adjustment for the schema to be easier to use with openapi generators and redocly. Line 711 should change from `f_schema['$ref'] = ref_prefix + enum_name` to `f_schema['allOf'] = [{'$ref': ref_prefix + enum_name}]`.

https://github.com/samuelcolvin/pydantic/pull/1749/files#diff-7d11a8c77c4167b2958bcdb27b18cf2aR711

This will ensure that the reference object will be extended with new default key.

```py
{'definitions': {'Foo': {'description': 'An enumeration.',
                         'enum': ['A', 'B'],
                         'title': 'Foo',
                         'type': 'string'}},
 'properties': {'foo': {'allOf': [{'$ref': '#/definitions/Foo'}],
                        'default': 'A',
                        'title': 'Foo'}},
 'title': 'ModelA',
 'type': 'object'}
```

Compare the live documentation for [the modified version](https://redocly.github.io/redoc/?url=https://gist.githubusercontent.com/mostaphaRoudsari/29a9fc4a6290a2d10ff5c79ca8b5642b/raw/528feef16e253075083b56c715fab90daa3e278a/enum_modified.json#tag/modela_model) with [the original version](https://redocly.github.io/redoc/?url=https://gist.githubusercontent.com/mostaphaRoudsari/29a9fc4a6290a2d10ff5c79ca8b5642b/raw/76e87aa95b3718aa1c1db6f0186a6630635cc033/enum_org.json#tag/modela_model).

## before

image


## after

image

PrettyWood:
Hi @mostaphaRoudsari 
Thank you for your feedback and for checking the behaviour with the doc generation tools. Very much appreciated!
I'll add your suggestion as it doesn't change anything in term of schema validation.
Out of curiosity do you know why it doesn't work directly with `$ref` ?

mostaphaRoudsari:
Hi @PrettyWood, thanks!

> Out of curiosity do you know why it doesn't work directly with $ref ?

I don't really! ¯\_(ツ)_/¯ My understanding is that the valid approach for extending an object in OpenAPI 3 is to use `allOf` key. There are long discussions on GitHub and not all the code-generators agree on the details of reusing a referenced object but I have found this approach to work the best both with re-doc and dotnet and JS code generators that we use. Sorry! I wish I could be more helpful. Maybe @tiangolo can help!

tiangolo:
Hmm, I think this PR is not necessary. I think Pydantic is already working as it should.

---

If you see how it behaves for models, the title is declared in the definition of the model itself, and then only the reference is used. A title would be overridden only if it was declared in a `= Field(..., title="Some title")`.

Here's a complete example showing both cases:

```Python


class Simple(BaseModel):
    count: int = 0
    size: float = None


class Bar(BaseModel):
    pure: Simple
    titled: Simple = Field(..., title="Not so simple")


print(Bar.schema_json(indent=2))

# {
#   "title": "Bar",
#   "type": "object",
#   "properties": {
#     "pure": {
#       "$ref": "#/definitions/Simple"
#     },
#     "titled": {
#       "title": "Not so simple",
#       "allOf": [
#         {
#           "$ref": "#/definitions/Simple"
#         }
#       ]
#     }
#   },
#   "required": [
#     "pure",
#     "titled"
#   ],
#   "definitions": {
#     "Simple": {
#       "title": "Simple",
#       "type": "object",
#       "properties": {
#         "count": {
#           "title": "Count",
#           "default": 0,
#           "type": "integer"
#         },
#         "size": {
#           "title": "Size",
#           "type": "number"
#         }
#       }
#     }
#   }
# }
```

Then, for enums, it should work the same way as for models, as described above.

And it is indeed already working that way :heavy_check_mark: :

```Python
from enum import Enum


class Names(str, Enum):
    rick = "Rick"
    morty = "Morty"
    summer = "Summer"


class Foo(BaseModel):
    pure: Names
    titled: Names = Field(
        ...,
        title="Titled Names",
        description="Only one of the available names is valid",
    )


print(Foo.schema_json(indent=2))

# {
#   "title": "Bar",
#   "type": "object",
#   "properties": {
#     "pure": {
#       "$ref": "#/definitions/Simple"
#     },
#     "titled": {
#       "title": "Not so simple",
#       "allOf": [
#         {
#           "$ref": "#/definitions/Simple"
#         }
#       ]
#     }
#   },
#   "required": [
#     "pure",
#     "titled"
#   ],
#   "definitions": {
#     "Simple": {
#       "title": "Simple",
#       "type": "object",
#       "properties": {
#         "count": {
#           "title": "Count",
#           "default": 0,
#           "type": "integer"
#         },
#         "size": {
#           "title": "Size",
#           "type": "number"
#         }
#       }
#     }
#   }
# }
```

PrettyWood:
@tiangolo Thanks for your answer but your example actually shows title and description (and default value) are not supported for enums.
With your last example this is what we currently have
```
{
  "title": "Foo",
  "type": "object",
  "properties": {
    "pure": {
      "$ref": "#/definitions/Names"
    },
    "titled": {
      "$ref": "#/definitions/Names"
    }
  },
  "required": [
    "pure",
    "titled"
  ],
  "definitions": {
    "Names": {
      "title": "Names",
      "description": "An enumeration.",
      "enum": [
        "Rick",
        "Morty",
        "Summer"
      ],
      "type": "string"
    }
  }
}
```
and this is what we have now with this PR
```
{
  "title": "Foo",
  "type": "object",
  "properties": {
    "pure": {
      "title": "Pure",
      "allOf": [
        {
          "$ref": "#/definitions/Names"
        }
      ]
    },
    "titled": {
      "title": "Titled Names",
      "description": "Only one of the available names is valid",
      "allOf": [
        {
          "$ref": "#/definitions/Names"
        }
      ]
    }
  },
  "required": [
    "pure",
    "titled"
  ],
  "definitions": {
    "Names": {
      "title": "Names",
      "description": "An enumeration.",
      "enum": [
        "Rick",
        "Morty",
        "Summer"
      ],
      "type": "string"
    }
  }
}
```
which is as you say the exact same behaviour as we do with a model reference.

mostaphaRoudsari:
@tiangolo, I see that the title of the PR might have been confusing but this PR does more than only fixing the title. I think the Enums description and title are secondary to *default values* which is currently missing as you can see in the above example.

tiangolo:
You're right, I must have been doing something wrong :sweat_smile:  . I just tried again re-installing and it's behaving as you describe. I probably ran the same example twice :man_facepalming: .

Seeing that, I think it's great to support this feature. :tada: 

But I think there shouldn't be a `title` override for an enum field that doesn't declare a custom title, as it works with models.

So, in short, I think a test like this should pass:

```Python
from enum import Enum

def test_enum_title_override():
    class Names(str, Enum):
        rick = 'Rick'
        morty = 'Morty'
        summer = 'Summer'

    class Foo(BaseModel):
        pure: Names
        titled: Names = Field(
            ..., title='Titled Names', description='Only one of the available names is valid',
        )

    assert Foo.schema() == {
        'title': 'Foo',
        'type': 'object',
        'properties': {
            'pure': {'$ref': '#/definitions/Names'},
            'titled': {
                'title': 'Titled Names',
                'description': 'Only one of the available names is valid',
                'allOf': [{'$ref': '#/definitions/Names'}],
            },
        },
        'required': ['pure', 'titled'],
        'definitions': {
            'Names': {
                'title': 'Names',
                'description': 'An enumeration.',
                'enum': ['Rick', 'Morty', 'Summer'],
                'type': 'string',
            }
        },
    }
```

tiangolo:
Also, about FastAPI, once this is ready, I'll check out this branch locally, and update the FastAPI tests to support it (both versions, the current and after this PR). That should then make the tests pass here. :heavy_check_mark:

tests/test_schema.py
```diff
@@ -280,13 +281,68 @@ class Model(BaseModel):
...
+            'Names': {
+                'description': 'An enumeration.',
+                'enum': ['Rick', 'Morty', 'Summer'],
+                'title': 'Names',
+                'type': 'string',
+            },
+        },
+        'properties': {
+            'enum': {'allOf': [{'$ref': '#/definitions/Names'}]},
+            'model': {'$ref': '#/definitions/Pika'},
```
PrettyWood, line 328: @tiangolo @samuelcolvin I updated the PR as discussed.
Looking at the schema generation for enums and models, I reckon we should (probably in a future PR) try to have the exact same behaviour and use allOf + $ref for models too.

tiangolo: Thanks for the attention and double-checking this! :mag: :detective: 

It's because in previous versions of JSON Schema (including the one that the current OpenAPI spec is based on) `$ref` was not allowed to have other keywords alongside of it. So, whenever a `$ref` was used, it had to be the only key in that dictionary/object.

That is allowed in the latest version (I just checked, I actually didn't notice it before). But for compatibility with OpenAPI, which is still based on the previous versions, I think this should work better.

Ref: https://json-schema.org/draft/2019-09/release-notes.html#keyword-changes

MingboPeng:
Hope this fix will be merged soon. Thanks all for making this fixed.

TenOs:
👍 can we get this merged please?

pydantic/schema.py
```diff
@@ -648,6 +667,11 @@ def add_field_type_to_schema(field_type: Any, schema: Dict[str, Any]) -> None:
             break
 
 
+def get_schema_ref(ref_name: str, schema_overrides: bool) -> Dict[str, Any]:
```
samuelcolvin, line 670: fix this going to conflict with #1480?

pydantic/schema.py
```diff
@@ -148,6 +148,31 @@ def model_schema(
...
+def get_field_info_schema(field: ModelField) -> Tuple[Dict[str, Any], bool]:
+    schema_overrides = False
+
+    # If no title is explicitly set, we don't set title in the schema for enums.
+    # The behaviour is the same as `BaseModel` reference, where the default title
+    # is in the definitions part of the schema.
+    is_enum_with_no_title_set = lenient_issubclass(field.type_, Enum) and not field.field_info.title
+    schema: Dict[str, Any] = {} if is_enum_with_no_title_set else {
+        'title': field.field_info.title or field.alias.title().replace('_', ' ')
+    }
```
samuelcolvin, line 160:
```suggestion
    schema: Dict[str, Any] = {}
    if lenient_issubclass(field.type_, Enum) and not field.field_info.title:
        schema['title'] = field.field_info.title or field.alias.title().replace('_', ' ')
```
or something would be clearer.

samuelcolvin (commented):
otherwise LGTM

pydantic/schema.py
PrettyWood: Hmm I'll update this PR is #1480 is merged first. But imo it should be quite easy to resolve conflicts

PrettyWood:
@samuelcolvin Feedback taken. Thanks for all the reviews!
Imo it's "safer" to merge this one first and update #1480 but it's your call. @Kilo59 WDYT ?

pydantic/schema.py
Kilo59: What's the benefit of pulling this logic into this function?

samuelcolvin:
Thanks so much.

PR #1755, opened 2020-07-22
feat: support `default_factory` with `BaseModel.construct`

### LINKED ISSUE #1732: BaseModel.construct returns instances with None assigned to fields with a default_factory
# Bug

```
             pydantic version: 1.6.1
            pydantic compiled: True
                 install path: /home/mihai/.virtualenvs/pydantic/lib/python3.8/site-packages/pydantic
               python version: 3.8.3 (default, May 17 2020, 18:15:42)  [GCC 10.1.0]
                     platform: Linux-5.7.7-1-ck-x86_64-with-glibc2.2.5
     optional deps. installed: ['devtools']
```

```py
from typing import List

class Model(BaseModel):
    foo: List[int] = Field(default_factory=list)
    bar: str = 'Baz'

print(Model.construct()) 

# in pydantic   1.5.1 the result is: foo=[] bar='Baz'
# in pydantic  ^1.6.0 the result is: foo=None bar='Baz'
```

Hello, I'm using pydantic in an application and after updating to v1.6.1 from v1.5.1 I came across this situation which results in a couple of failures.

From the changelog and documentation it's not clear if this change was intended or not.

PrettyWood:
# Change Summary
Some changes have been made on `default_factory` to avoid calling it for nothing and avoid some side effects in #1504 and #1712. And [this line](https://github.com/samuelcolvin/pydantic/pull/1504/files#diff-1ef405a6b5a2fa9f20ea7e1f92e4c38eR299) changed the default value that is used in `BaseModel.construct` to `None`.
Instead of changing `__field_defaults__`, which would be a mess, let's just use `ModelField.get_default` in `construct`.
The feature is hence scoped to `BaseModel.construct` and has the exact same behaviour as `BaseModel.__init__`

## Related issue number
closes #1732

## Checklist


PrettyWood:
Hello @zoopp 
To be honest I checked the `__field_defaults__` while doing the PR and I agree it could be removed.
Right now we don't support `dict(exclude_defaults)` with `default_factory` and I wanted to remove `__field_defaults__` while working on it. But maybe we could remove `__field_defaults__` in this PR by just calling `BaseField.default`

zoopp:
Sure, I was only inquiring about it out of curiosity.

zoopp:
I'm wondering if there anything I can do expedite a potential merge of this pull request. Is there any specific feedback I can provide to help with it?

PrettyWood:
@zoopp As Samuel is busy working, I would just fork this repo and merge the PRs you need if you can't wait an extra month

samuelcolvin (commented):
This looks like it could be fine, but questions:
* could this be considered a breaking change, after all people use `__fields__` a lot, they probably use `__field_defaults__` too? at least I think we need a note in the change description
* this is currently conflicting with master, could you fix
* does this have a measurable effect on performance.

PrettyWood:
There is an effect on performance for sure

##### with `default_factory`
```
# 3 loops on `master` branch with 10 ** 6 loops
5704059017ns
5680524864ns
5950163218ns

# 3 loops on this branch with 10 ** 6 loops
5811149041ns
6014411367ns
5877212718ns
```

#### without `default_factory` (was not supported)
```
# 3 loops on `master` branch with 10 ** 6 loops
1517463041ns
1480426577ns
1481139567ns

# 3 loops on this branch with 10 ** 6 loops
1648662578ns
1656794397ns
1632753169ns
```

Don't know if it's a big deal.
We could maybe change the approach by keeping  `__field_defaults__` but not having a regular `dict` but something like
```python
class FieldDefaultsDict(Dict[str, Any]):
    def add_field_default(self, key: str, field: ModelField) -> None:
        if field.default_factory:
            default_value = field.default_factory
            default_value.__default_factory__ = True  # this would need some hack as `default_value` can be a builtin
            self[key] = default_value
        else:
            self[key] = field.default

    def __getitem__(self, item: str) -> Any:
        default_value = super().__getitem__(item)
        return default_value() if default_value.__default_factory__ else default_value
```

samuelcolvin:
thanks so much. I think this is definitely better than what went before.

PR #1759, opened 2020-07-23
fix: raise RuntimeError in schema.get_long_model_name() if Model.__module__ not set

### LINKED ISSUE #1758: schema.get_long_model_name() generates non-unique names if model.__module__ not set
# Bug

```
             pydantic version: 1.6.1
            pydantic compiled: False
                 install path: /home/lauris/git/pydantic/pydantic
               python version: 3.7.3 (default, Dec 20 2019, 18:57:59)  [GCC 8.3.0]
                     platform: Linux-4.19.0-9-amd64-x86_64-with-debian-10.4
...
```
`schema.get_long_model_name()` generates non-unique model names if `Model.__module__` is not set.
```py
>>> from pydantic import create_model
>>> from pydantic.schema import get_long_model_name
>>>
>>> model = create_model('DynamicFoobarModel', foo=(str, ...), bar=123)
>>> get_long_model_name(model)
'None__DynamicFoobarModel'
```

alekna:
closes #1758

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -353,6 +353,8 @@ def get_flat_models_from_models(models: Sequence[Type['BaseModel']]) -> TypeMode
 
 
 def get_long_model_name(model: TypeModelOrEnum) -> str:
+    if not model.__module__:
+        raise RuntimeError(f'{model.__name__}.__module__ not set')
     return f'{model.__module__}__{model.__name__}'.replace('.', '__')
 
 
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -12,13 +12,14 @@ from uuid import UUID
 
 import pytest
 
-from pydantic import BaseModel, Extra, Field, ValidationError, conlist, conset, validator
+from pydantic import BaseModel, Extra, Field, ValidationError, conlist, conset, create_model, validator
 from pydantic.color import Color
 from pydantic.dataclasses import dataclass
 from pydantic.networks import AnyUrl, EmailStr, IPvAnyAddress, IPvAnyInterface, IPvAnyNetwork, NameEmail, stricturl
 from pydantic.schema import (
     get_flat_models_from_model,
     get_flat_models_from_models,
+    get_long_model_name,
     get_model_name_map,
     model_process_schema,
     model_schema,
@@ -1880,3 +1881,15 @@ def test_new_type():
         'properties': {'a': {'title': 'A', 'type': 'string'}},
         'required': ['a'],
     }
+
+
+def test_get_long_model_name():
+    model = create_model('DynamicFoobarModel', foo=(str, ...), bar=123)
+
+    with pytest.raises(RuntimeError):
+        get_long_model_name(model)
+
+    class Model(BaseModel):
+        foo: str
+
+    assert get_long_model_name(Model) == 'tests__test_schema__Model'


PrettyWood:
Hi @alekna,

I'm not convinced we should raise an error for that.
For example, this is valid at the moment but will break with your PR

```python
def test_model_name_maps_no_module():
    create_testing_submodules()
    from pydantic_schema_test.modulea.modela import Model as ModelA

    Model = create_model('Model', foo=(str, ...))

    model_name_map = get_model_name_map([Model, ModelA])
    assert model_name_map == {
        Model: 'None__Model',
        ModelA: 'pydantic_schema_test__modulea__modela__Model',
    }
```

alekna:
For the user who is not familiar with `pydantic` internals `could not generate unique name...`None_Model` already exists` wouldn't be a very informative message as to where the problem is.

PrettyWood:
I would really like a full example of the issue you're trying to solve because I can imagine issues with or without modules. I don't really see how it is related to the module being not set only

alekna:
Here's the example of a problem I had tonight (we're using `create_model` to generate models dynamically). Had to spend some time digging into `pydantic` to understand why this is happening. Proposed error message could have helped to reduce time I spent troubleshooting this issue.

**Code:**
```py
from fastapi import FastAPI

FirstRequestModel = create_model('Model', foo=(str, ...))
SecondRequestModel = create_model('Model', foo=(str, ...))

app = FastAPI(docs_url="/")


@app.get("/foo")
def first(req: FirstRequestModel):
    return


@app.get("/bar")
def second(req: SecondRequestModel):
    return
```

**Log:**
```
INFO:     127.0.0.1:44474 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:44474 - "GET /openapi.json HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 386, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/fastapi/applications.py", line 181, in __call__
    await super().__call__(scope, receive, send)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/applications.py", line 111, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/routing.py", line 566, in __call__
    await route.handle(scope, receive, send)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/routing.py", line 227, in handle
    await self.app(scope, receive, send)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/starlette/routing.py", line 41, in app
    response = await func(request)
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/fastapi/applications.py", line 130, in openapi
    return JSONResponse(self.openapi())
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/fastapi/applications.py", line 112, in openapi
    servers=self.servers,
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/fastapi/openapi/utils.py", line 350, in get_openapi
    flat_models=flat_models, model_name_map=model_name_map  # type: ignore
  File "/home/lauris/git/pydantic/venv/lib/python3.7/site-packages/fastapi/utils.py", line 64, in get_model_definitions
    model_name = model_name_map[model]
KeyError: <class 'Model'>
```

alekna:
I also don't agree that casting `NoneType` to `"None"` is a way go as per your code example.

PrettyWood:
I never said `None` as module made sense but it could be replaced by `__main__` or just ''. Anyway now I understand better the problem and I keep thinking it's not a module issue.
With your PR you remove some allowed behaviour to have a more explicit error but imo the real issue is that if you write
```python
FirstRequestModel = create_model('FirstRequestModel', foo=(str, ...))
SecondRequestModel = create_model('SecondRequestModel', bar=(str, ...))
```
it doesn't work even though it should not conflict at all. Apparently both models have same name `'Model'`.
It hence breaks at they have the same module (here `None` yes...) but it would still break if we set the same `__module__` arg

alekna:
The way I see it, the purpose of `get_long_model_name()` is to generate more unique model name whenever there is a collision. It doesn't serve it's intended purpose if `Model.__module__` is not required to be set. Hence the proposed change.

Additionally, when used in conjunction with `FastAPI`, it results in `KeyError` which is not very informative. The purpose of my PR was to make error more descriptive so that when somebody hits same edge case he or she could spent less time troubleshooting it. Does that make sense? Whether it's the most efficient way to solve this problem - that's another question.

PrettyWood:
@alekna I agree the error is not very explicit and I'm open to a PR to add a better error message. I just say the current implementation does not tackle the real problem and break some currently supported behaviour.

alekna:
I think I oversimplified my example by too much so it doesn't reflect how I got to this problem. Unfortunately I'm not in position to share more details on how we use `pydantic`.

With regard to breaking some currently supported behaviour, as far as I can see, all unit tests are still passing.

PrettyWood:
@alekna I gave you an example of currently working example that would break with your PR. It's not in the tests but still it works. Don't you think something like [this](https://github.com/samuelcolvin/pydantic/compare/master...PrettyWood:model-name-conflict?expand=1) would make more sense ?
It's even more explicit and actually tackle you problem.
Sorry to challenge you so much

PR #1814, opened 2020-08-09
Parse time with tz

### LINKED ISSUE #1744: time ignores timedelta without error
# Bug

```
pydantic version: 1.5.1
pydantic compiled: True
install path: C:\...\.venv37\Lib\site-packages\pydantic
python version: 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
platform: Windows-10-10.0.17134-SP0
```

In the documentation at https://pydantic-docs.helpmanual.io/usage/types/#datetime-types it is stated that

> - time fields can be:
>    - time, existing time object
>    - str, following formats work:
>        - HH:MM[:SS[.ffffff]]

but as shown in the example below strings like `11:05+05:00`
- are accepted and raise no error
- loose the timedelta information

Loosing information (especially timedelta) without recognition can e.g. lead to discrepancies when collecting data from different sources and a relationship between single pieces of data must be established based on a timestamp.

I tried to track it down to the root cause:
```py
>>> from pydantic import datetime_parse

>>> incorrect_time = datetime_parse.parse_time("11:05+05:00")
>>> print(incorrect_time)
datetime.time(11, 5)
# Here I would expect to 
# - either get an error because it is not stated that a format having a timedelta is accepted
# - or a datetime.time(11, 5, tzinfo=datetime.timezone(datetime.timedelta(seconds=18000)) as datetime.time.fromisoformat would return
```

noelevans:
## Change Summary

Permits timezones to be supplied with times

## Related issue number

 time ignores timedelta without error: closes #1744 

## Checklist


noelevans:
I'm a bit confused by the code-coverage report. I've breakpointed my change and believe I've seen each line I've changed being passed from the unittest I've added. Most of the files cited by the report have not been modified in this PR

PrettyWood:
@noelevans Don't worry it's only the first report where some tests are skipped ;)

tests/test_datetime_parse.py
```diff
@@ -242,3 +242,8 @@ class Model(BaseModel):
         'type': 'value_error.unicodedecode',
         'msg': "'utf-8' codec can't decode byte 0x81 in position 0: invalid start byte",
     }
+
+
+def test_time_with_timedelta():
```
PrettyWood, line 247: You should put your new test cases [here](https://github.com/samuelcolvin/pydantic/blob/95d6f613b223d532295892090a97952065c1b87a/tests/test_datetime_parse.py#L35)

noelevans: Done. I also added an extra try-except guard when a TZ > 24 hours is stated. 

(I checked parse_datetime which already copes with an error of this nature.)

noelevans:
I'd done a rebase to squash this branch's commits. To push to origin I had to do a merge again. This triggered the build checks to be run again.

Looking at the logs there seems to be an installation problem:
2020-08-13T08:36:16.3906514Z ##[section]Starting: Request a runner to run this job
2020-08-13T08:36:17.7544874Z Can't find any online and idle self-hosted runner in current repository that matches the required labels: 'macos-latest'
2020-08-13T08:36:17.7544976Z Can't find any online and idle self-hosted runner in current repository's account/organization that matches the required labels: 'macos-latest'
2020-08-13T08:36:17.7545356Z Found online and idle hosted runner in current repository's account/organization that matches the required labels: 'macos-latest'
2020-08-13T08:36:17.9009280Z ##[section]Finishing: Request a runner to run this job

PrettyWood:
@noelevans I relaunched the CI and it's good. Probably on github side

noelevans:
> I relaunched the CI and it's good.
Brilliant!  

> Do not hesitate to just git commit -m "retry" --allow-empty in those cases if you want
I'll keep that in mind. I'm still new-ish to public repository etiquette. Thanks for your help @PrettyWood

PrettyWood (approved):
LGTM thanks

pydantic/datetime_parse.py
```diff
@@ -143,7 +143,23 @@ def parse_time(value: Union[time, StrBytesIntFloat]) -> time:
     if kw['microsecond']:
         kw['microsecond'] = kw['microsecond'].ljust(6, '0')
 
-    kw_ = {k: int(v) for k, v in kw.items() if v is not None}
+    tzinfo_str = kw.pop('tzinfo')
+    if tzinfo_str == 'Z':
```
samuelcolvin, line 147: looks like this is duplicated from `parse_datetime`, maybe extract to a new file.

pydantic/datetime_parse.py
```diff
@@ -23,7 +23,9 @@
 date_re = re.compile(r'(?P<year>\d{4})-(?P<month>\d{1,2})-(?P<day>\d{1,2})$')
 
 time_re = re.compile(
-    r'(?P<hour>\d{1,2}):(?P<minute>\d{1,2})(?::(?P<second>\d{1,2})(?:\.(?P<microsecond>\d{1,6})\d{0,6})?)?'
+    r'(?P<hour>\d{1,2}):(?P<minute>\d{1,2})'
+    r'(?::(?P<second>\d{1,2})(?:\.(?P<microsecond>\d{1,6})\d{0,6})?)?'
+    r'(?P<tzinfo>Z|[+-]\d{2}(?::?\d{2})?)?$'
```
samuelcolvin, line 28:
```py
tz_suffix = r'(?P<tzinfo>Z|[+-]\d{2}(?::?\d{2})?)?$'

time_re = re.compile(... + tz_suffix)
...
datetime_re = re.compile(... + tz_suffix)
```

and reuse the reusable bit

samuelcolvin: in fact you might be able to reuse the whole of the time regex at the end of `datetime_re`?

samuelcolvin (commented):
otherwise this looks good.

pydantic/datetime_parse.py
noelevans: I think you meant to write function instead of file? - as updated?

noelevans:
In the associated ticket, this comment https://github.com/samuelcolvin/pydantic/issues/1744#issuecomment-661863986 speculates that the fix may turn out to be a breaking change. With this PR I don't think that is the case?

pydantic/datetime_parse.py
noelevans: In moving this code to another function I spotted that the `datetime` parsing is also able to raise a ValueError exception when a bad TZ is given. There is an new test case demonstrating this which is fixed by the last commit, instead raising a `DateTimeError`.

pydantic/datetime_parse.py
```diff
@@ -81,6 +79,23 @@ def from_unix_seconds(seconds: Union[int, float]) -> datetime:
     return dt.replace(tzinfo=timezone.utc)
 
 
+def _parse_timezone(value: str, error: Callable[[], errors.PydanticValueError]) -> timezone:
```
samuelcolvin, line 82:
```suggestion
def _parse_timezone(value: str, error: Type[Exception]) -> Optional[timezone]:
```

pydantic/datetime_parse.py
```diff
@@ -81,6 +79,23 @@ def from_unix_seconds(seconds: Union[int, float]) -> datetime:
...
+    if value == 'Z':
+        tzinfo = timezone.utc
```
samuelcolvin, line 84:
```suggestion
        return timezone.utc
```

pydantic/datetime_parse.py
```diff
@@ -81,6 +79,23 @@ def from_unix_seconds(seconds: Union[int, float]) -> datetime:
...
+    elif value is not None:
+        offset_mins = int(value[-2:]) if len(value) > 3 else 0
+        offset = 60 * int(value[1:3]) + offset_mins
+        if value[0] == '-':
+            offset = -offset
+        try:
+            tzinfo = timezone(timedelta(minutes=offset))
```
samuelcolvin, line 91:
```suggestion
            return timezone(timedelta(minutes=offset))
```

pydantic/datetime_parse.py
```diff
@@ -81,6 +79,23 @@ def from_unix_seconds(seconds: Union[int, float]) -> datetime:
...
+        except ValueError:
+            raise error()
+    else:
+        tzinfo = None
+    return tzinfo
```
samuelcolvin, line 96:
```suggestion
    return None
```

samuelcolvin (commented):
otherwise looking good.

pydantic/datetime_parse.py
noelevans: When the return annotation is 
-> Optional[timezone]:
The unittests all pass but calling mypy, I get this:
$ mypy pydantic/datetime_parse.py 
pydantic/datetime_parse.py:161: error: Incompatible types in assignment (expression has type "Optional[timezone]", target has type "Union[int, timezone]")  [assignment]
pydantic/datetime_parse.py:199: error: Incompatible types in assignment (expression has type "Optional[timezone]", target has type "Union[int, timezone]")  [assignment]
Found 2 errors in 1 file (checked 1 source file)

acnebs:
Is it possible to disable this behavior? I have timestamp strings with 'Z' in them, but I don't actually want the tzinfo to be included in the resulting `datetime` object.

Mainly because I'm using pydantic alongside asyncpg, which wants you to be using a `timestamptz` column, but all my dates are UTC so I don't need the tz...

PR #1817, opened 2020-08-11
feat: make pydantic `dataclass` decorator support built-in `dataclass` 

### LINKED ISSUE #744: converting normal dataclass to pydantic dataclass
# Feature Request

Please complete:
* OS: **Ubuntu**
* Python version `import sys; print(sys.version)`: **3.7.4**
* Pydantic version `import pydantic; print(pydantic.VERSION)`: **v0.32.1**

Would be great if `pydantic.dataclasses.dataclass` could convert an exsting dataclass to a validated/parsed dataclass.

This currently doesn't work and you just get back the normal dataclass.

```py
from typing import Optional
import pydantic.dataclasses

@dataclass
class File:
    hash: str
    name: Optional[str]
    size: int
    content: Optional[bytes] = None

FileChecked = pydantic.dataclasses.dataclass(File)
f = FileChecked(hash='xxx', name=b'whatever.txt', size='456')
debug(f)
# doesn't currently convert name to a str or size to an int
```

### LINKED ISSUE #1743: Validating nested dataclasses without class ownership
# Question

Hi! I want to validate a model whose definition I don't control from a package we depend on. That model consists of nested dataclasses. I thought I'd be able to subclass that model with pydantic's `dataclass` decorator and it would work just fine, and it does, but only if the model I am subclassing doesn't have any nested dataclasses. Here's what I mean:

```py
from pydantic.dataclasses import dataclass as pydataclass

# FROM AN EXTERNAL PACKAGE
@dataclass
class NavbarButton:
    href: str


@dataclass
class Navbar:
    button: NavbarButton
# END EXTERNAL PACKAGE

@pydataclass
class CustomNavbar(Navbar):
    pass

navbar = CustomNavbar("https://example.com",)
print(navbar)
```

This throws:

`RuntimeError: no validator found for <class '__main__.NavbarButton'>, see arbitrary_types_allowed in Config`

Is there any way to get Pydantic to play nice with these classes except by essentially copying code?

PrettyWood:
## Change Summary
It was not possible to use `pydantic.dataclasses.dataclass` decorator directly on a built-in `dataclass`.
Now it should work.
It allows to easily support nested dataclasses and add validation for them. This is why it's also part of this PR

ℹ️ I'll add some doc if the behaviour is validated and no extra use cases are added

## Related issue number
closes #744
closes #1743

## Checklist


dylrich:
This is great!

hawkaa:
This is amazing! Need a hand with the last details before this is merged to master?

PrettyWood:
@hawkaa I'll update the doc once the behaviour is fully validated. If you want to help, you could review the code and ensure the behaviour is the right one even for nested dataclasses. I'm not sure I thought about everything there may probably be some edge cases to handle.
Thanks for the feedback and for the help !

pydantic/dataclasses.py
```diff
@@ -128,18 +203,23 @@ def dataclass(
...
+    def wrap(cls: Type[Any]) -> Type['DataclassType']:
         return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen, config)
 
     if _cls is None:
         return wrap
 
     return wrap(_cls)
+
+
+def make_dataclass_validator(_cls: Type[Any], **kwargs: Any) -> 'CallableGenerator':
```
samuelcolvin, line 223: docstring please to explain to what this does.

pydantic/dataclasses.py
```diff
@@ -1,23 +1,32 @@
-from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, Optional, Type, TypeVar, Union
+import dataclasses
```
samuelcolvin, line 1: I have a suspicion that importing `dataclasses` globally had a big effect on overall import time.

I think easiest to move it back to imports in functions, we could benchmark to see the performance of making this global.

pydantic/dataclasses.py
```diff
@@ -128,18 +203,23 @@ def dataclass(
     unsafe_hash: bool = False,
     frozen: bool = False,
     config: Type[Any] = None,
-) -> Union[Callable[[Type[Any]], 'DataclassType'], 'DataclassType']:
+) -> Union[Callable[[Type[Any]], Type['DataclassType']], Type['DataclassType']]:
```
samuelcolvin, line 206: Wait, surely this signature is now wrong, you have double `Type` and the type of a type is always just `type`.

Maybe `DataclassType` is defined wrongly?

pydantic/dataclasses.py
```diff
@@ -54,6 +68,39 @@ def setattr_validate_assignment(self: 'DataclassType', name: str, value: Any) ->
...
 
+def is_builtin_dataclass(_cls: Type[Any]) -> bool:
+    """
+    `dataclasses.is_dataclass` is True if one of the class parents is a `dataclass`.
+    This is why we also add a class attribute `__processed__` to only consider 'direct' built-in dataclasses
+    """
+    return not hasattr(_cls, '__processed__') and dataclasses.is_dataclass(_cls)
+
+
+def _dataclass_with_validation(
```
samuelcolvin, line 79: not sure it's worth this being a separate function?

tests/test_dataclasses.py
```diff
@@ -633,3 +633,93 @@ class MyDataclass:
...
+                'required': ['modified_date', 'seen_count'],
+                'title': 'Meta',
+                'type': 'object',
+            }
+        },
+        'properties': {'filename': {'title': 'Filename', 'type': 'string'}, 'meta': {'$ref': '#/definitions/Meta'}},
+        'required': ['filename', 'meta'],
+        'title': 'File',
+        'type': 'object',
+    }
```
samuelcolvin, line 725: can we add a case like:

```py
@dataclasses.dataclass
class File:
    filename: str
    meta: Meta


class Foo(BaseModel):
    file: File

...
```

I know this should work, but this is what I often want to do so best to be explicit that it works.

samuelcolvin (commented):
This looks awesome, I really want it.

Mostly small questions or tweaks.

pydantic/dataclasses.py
PrettyWood: The name was missleading as `DataclassType` was refering to an instance of a pydantic dataclass. I renamed it to `Dataclass`.

pydantic/dataclasses.py
PrettyWood: Yes I know I remember when you changed some imports to be local in order to improve overall import time.
I forgot to import it back to all functions.

pydantic/dataclasses.py
PrettyWood: 939f990

PrettyWood:
@samuelcolvin Thanks for the review! I took all your remarks into account. I still need to write some documentation about this but first I really want to make sure the whole behaviour is right. If you think it is please let me know!

pydantic/dataclasses.py
```diff
@@ -107,17 +147,46 @@ def _pydantic_post_init(self: 'DataclassType', *initvars: Any) -> None:
...
         cls.__post_init_original__ = post_init_original
 
     if cls.__pydantic_model__.__config__.validate_assignment and not frozen:
-        cls.__setattr__ = setattr_validate_assignment
+        cls.__setattr__ = setattr_validate_assignment  # type: ignore[assignment]
 
     return cls
 
 
+@overload
```
samuelcolvin, line 161: No biggy, but make these should be in `TYPE_CHECKING`?

samuelcolvin (commented):
This looks great, just a small question, and documentation.

pydantic/dataclasses.py
PrettyWood: I tried but doesn't seem to work
And then I saw that it was not in `TYPE_CHECKING` for [root_validator](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/class_validators.py#L99)

PrettyWood:
@samuelcolvin Documentation has been added 👍 Do not hesitate to tweak it if need be

samuelcolvin:
this is great, thanks.

Issue #1279, opened 2020-03-02
Read values from secrets files when using BaseSettings


ricardogsilva:
# Feature Request

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
pydantic version: 1.4
pydantic compiled: True
install path: /home/likeno/.cache/pypoetry/virtualenvs/geospaca-yth2PwXS-py3.7/lib/python3.7/site-packages/pydantic
python version: 3.7.6 (default, Dec 28 2019, 05:50:43)  [GCC 8.3.0]
platform: Linux-5.3.0-40-generic-x86_64-with-debian-10.2
optional deps. installed: ['typing-extensions']
```

I'm using pydantic to provide an easy way to read my app's settings from the environment. Using `BaseSettings` it works great.

Now I'm moving to a setup where some sensitive settings are being managed as [docker secrets][] and I'd like to have support for those baked into `pydantic.BaseSettings` too.

Basically this means `BaseSettings` would need to be able to read values from individual files, getting the name of the variable from the filename and its value from the file's contents. 

I've actually implemented this is a derived class already, but would like to know if there is interest on incorporating it on pydantic.

Code looks like this:

```python
# this is my settings file (it's actually a django settings file so I'm 
# showing only relevant stuff)
import pydantic


class MyConfig(pydantic.BaseSettings):
    rabbitmq_url: pydantic.AnyUrl = "amqp://guest:guest@127.0.0.1:5672"

    class Config:
        env_prefix = "django_"

    def __init__(
            __pydantic_self__,
            *args,
            secrets_directory: typing.Optional[typing.Union[str, Path]] = None,
            **kwargs
    ):
        if secrets_directory is not None:
            extra_values = __pydantic_self__._scan_secrets_dir(
                Path(secrets_directory).expanduser())
            modified_values = __pydantic_self__._modify_secrets(extra_values)
            super().__init__(*args, **modified_values, **kwargs)
        else:
            super().__init__(*args, **kwargs)

    def _scan_secrets_dir(
            self,
            secrets_directory: Path
    ) -> typing.Dict[str, str]:
        result = {}
        if secrets_directory.is_dir():
            for item in secrets_directory.iterdir():
                if item.is_file():
                    contents = item.read_text().strip()
                    result[item.name] = contents
        return result

    def _modify_secrets(self, secrets: typing.Dict[str, str]):
        result = {}
        for name, value in secrets.items():
            if self.__config__.case_sensitive:
[...]

PR #1820, opened 2020-08-12
Add ability to read secrets from files

mdgilene:
## Change Summary

Added ability for `BaseSettings` to load `SecretStr` and `SecretBytes` from files. I opted to drop the ability to load complex types from secrets as there is no existing secret type for complex objects.

## Related issue number

#1279

## Checklist


pydantic/env_settings.py
```diff
@@ -27,20 +29,61 @@ def __init__(
...
+
+        secret_paths = {
+            (item.name if self.__config__.case_sensitive else item.name.lower()): item
+            for item in secrets_path.iterdir()
+            if item.is_file()
+        }
+
+        for field in self.__fields__.values():
+            if field.type_ not in (SecretStr, SecretBytes):
+                continue
```
FichteFoll, line 77: I was wondering whether a debug log would be useful here, but logging doesn't seem to be used much in pydantic.

Regardless, it is currently not documented that secrets are only loaded for `SecretStr` and `SecretBytes` types.

pydantic/env_settings.py
```diff
@@ -27,20 +29,61 @@ def __init__(
...
         init_kwargs: Dict[str, Any],
         _env_file: Union[Path, str, None] = None,
         _env_file_encoding: Optional[str] = None,
+        _secrets_dir: Union[Path, str, None] = None,
     ) -> Dict[str, Any]:
-        return deep_update(self._build_environ(_env_file, _env_file_encoding), init_kwargs)
+        return deep_update(
+            deep_update(self._build_secrets(_secrets_dir), self._build_environ(_env_file, _env_file_encoding)),
+            init_kwargs,
+        )
```
FichteFoll, line 52: Maybe change the signature of `deep_update` to support `*args`?

pydantic/env_settings.py
mdgilene: I agree a message of some kind would be nice here but like you said logging isn't used much. An alternative would be to raise an exception and make it a hard failure if you try to load a secret value for a field not defined as `SecretStr` or `SecretBytes`. Either way I will make sure to add a note to the docs stating that only those 2 types are loaded from secret files.

mdgilene: I decided to go ahead and include a warning message. There is some precedence for warning messages elsewhere in the code so seems fine to use it here as well.

pydantic/env_settings.py
```diff
@@ -47,8 +47,7 @@ def _build_values(
         _secrets_dir: Union[Path, str, None] = None,
     ) -> Dict[str, Any]:
         return deep_update(
-            deep_update(self._build_secrets(_secrets_dir), self._build_environ(_env_file, _env_file_encoding)),
-            init_kwargs,
+            self._build_secrets(_secrets_dir), self._build_environ(_env_file, _env_file_encoding), init_kwargs
```
FichteFoll, line 50: for better readability (and following PEP8, afaik), I'd split the parameters over multiple lines, one each.

mdgilene: I ran the configured code formatter so I would assume it would handle this if it was a problem no? Maybe not? I will wait for a review and if a maintainer would like it re-formatted I will do so.

samuelcolvin: this is great, in black we trust (mostly)

pydantic/env_settings.py
```diff
@@ -5,10 +5,12 @@
 
 from .fields import ModelField
 from .main import BaseConfig, BaseModel, Extra
+from .types import SecretBytes, SecretStr
 from .typing import display_as_type
 from .utils import deep_update, sequence_like
 
 env_file_sentinel = str(object())
+secrets_dir_sentinel = str(object())
```
samuelcolvin, line 13: this doesn't seem to be used.

pydantic/env_settings.py
```diff
@@ -27,20 +29,66 @@ def __init__(
...
+        """
+        d: Dict[str, Optional[str]] = {}
+
+        secrets_dir = _secrets_dir or self.__config__.secrets_dir
+        if secrets_dir is None:
+            return d
+
+        secrets_path = Path(secrets_dir)
+
+        if not secrets_path.is_dir():
```
samuelcolvin, line 65: surely to avoid a lot of confusion we should raise an error if `secrets_path.is_file()`?

samuelcolvin: perhaps we should also raise an error/warning if the directory is missing? i'm not so sure on that.

Perhaps only if it's provided via an argument, not config?

pydantic/env_settings.py
```diff
@@ -27,20 +29,66 @@ def __init__(
...
+            for item in secrets_path.iterdir()
+            if item.is_file()
+        }
+
+        for field in self.__fields__.values():
+            if field.type_ not in (SecretStr, SecretBytes):
+                warnings.warn(
+                    f'found secret file matching field "{field.name}" but the field was not '
+                    'of type "SecretStr" or "SecretBytes". This secret file was not loaded.',
+                    Warning,
```
samuelcolvin, line 79: perhaps we should remove this and use `UserWarning` so the warning is printed? (From memory only `UserWarning` is printed out by default which would be useful I imagine.

pydantic/env_settings.py
```diff
@@ -27,20 +29,66 @@ def __init__(
...
+        secret_paths = {
+            (item.name if self.__config__.case_sensitive else item.name.lower()): item
+            for item in secrets_path.iterdir()
+            if item.is_file()
+        }
+
+        for field in self.__fields__.values():
+            if field.type_ not in (SecretStr, SecretBytes):
+                warnings.warn(
+                    f'found secret file matching field "{field.name}" but the field was not '
```
samuelcolvin, line 77: wait, this is in the wrong place I think. No file has been found yet, we're just iterating through fields.

pydantic/env_settings.py
```diff
@@ -27,20 +29,66 @@ def __init__(
...
+        secrets_dir = _secrets_dir or self.__config__.secrets_dir
+        if secrets_dir is None:
+            return d
+
+        secrets_path = Path(secrets_dir)
+
+        if not secrets_path.is_dir():
+            return d
+
+        secret_paths = {
```
samuelcolvin, line 68: remove this see comment below

pydantic/env_settings.py
```diff
@@ -27,20 +29,66 @@ def __init__(
...
+                    f'found secret file matching field "{field.name}" but the field was not '
+                    'of type "SecretStr" or "SecretBytes". This secret file was not loaded.',
+                    Warning,
+                    stacklevel=4,
+                )
+                continue
+
+            value: Optional[str] = None
+            for env_name in field.field_info.extra['env_names']:
+                path = secret_paths.get(env_name, None)
```
samuelcolvin, line 86: Better to remove the `` dict above and just do

```py
path = secrets_path  / env_name
if path.is_file() and secret type: add to the main dict
elif path.is_dir() and secret type: warning that it's a directory
elif path.is_file() and not secret type: warning that the file exist but type is not secret
```

pydantic/env_settings.py
```diff
@@ -27,20 +29,66 @@ def __init__(
...
-        return deep_update(self._build_environ(_env_file, _env_file_encoding), init_kwargs)
+        return deep_update(
+            self._build_secrets(_secrets_dir), self._build_environ(_env_file, _env_file_encoding), init_kwargs
+        )
+
+    def _build_secrets(self, _secrets_dir: Union[Path, str, None] = None) -> Dict[str, Optional[str]]:
+        """
+        Build secrets suitable for passing to the Model.
+        """
+        d: Dict[str, Optional[str]] = {}
```
samuelcolvin, line 57: more descriptive name please.

pydantic/utils.py
```diff
@@ -122,13 +122,14 @@ def in_ipython() -> bool:
 KeyType = TypeVar('KeyType')
 
 
-def deep_update(mapping: Dict[KeyType, Any], updating_mapping: Dict[KeyType, Any]) -> Dict[KeyType, Any]:
+def deep_update(mapping: Dict[KeyType, Any], *updating_mappings: Dict[KeyType, Any]) -> Dict[KeyType, Any]:
```
samuelcolvin, line 125: :+1: looks good

docs/usage/settings.md
```diff
@@ -128,6 +128,50 @@ Because python-dotenv is used to parse the file, bash-like semantics such as `ex
...
+
+In either case, the value of the passed argument can be any valid directory, either absolute or relative to the
+current working directory. From there, *pydantic* will handle everything for you by loading in your variables and
+validating them.
+
+Even when using a secrets directory, *pydantic* will still read environment variables from a dotenv file or the environment,
+**a dotenv file and environment variables will always take priority over values loaded from the secrets directory**.
+
+Passing a file path via the `_secrets_dir` keyword argument on instantiation (method 2) will override
+the value (if any) set on the `Config` class.
```
samuelcolvin, line 173: Does this work with docker secrets and AWS Parameter Store, if so, please can we add a section to the docs saying  that and giving an example of how you'd use them.

samuelcolvin (commented):
this is looking great, a few things to change.

pydantic/env_settings.py
mdgilene: Good point, I think that since the default value for secrets_dir is None, that providing any value pointing to a directory that is missing can be an error.

pydantic/env_settings.py
mdgilene: Yep, leftover from copying the environment code. Will remove

pydantic/env_settings.py
mdgilene: Ok, this is the same name used by the environment loading code so may want to fix that at another time as well.

docs/usage/settings.md
mdgilene: I don't have any experience with AWS so I can't speak to it. However, Docker secrets as far as I'm aware simply mount the secrets as volumes to /run/secrets so yes this should handle that use case.

samuelcolvin:
I've made a few tweaks to the docs, this looks great and I'm ready to merge it except one thing.

I think these values from secrets files should apply to all fields, not just `SecretStr` and `SecretBytes`. This is because:
* it would make it more like other variable substitutions from arguments, environment variables and dot-env files, and thus less confusing
* I often want field types to be something like `HttpUrl` or even `int`, restricting to `SecretStr` and `SecretBytes` would prevent that

Sorry, I should have thought about this before, but it didn't occur to me until now.

@mdgilene and @FichteFoll what do you think? Unless you have strong objections, I can either wait for you to make the change, or make it myself.

FichteFoll:
> I think these values from secrets files should apply to all fields, not just `SecretStr` and `SecretBytes`.

I definitely agree with your points there, as they reflect my own concerns regarding gating this behind special classes. While the explicit declaration of loading from secrets makes it easier to see what values are different at a glance, it also limits the functionality compared to loading values from environment variable, where nothing needs to be configured.

mdgilene:
> I've made a few tweaks to the docs, this looks great and I'm ready to merge it except one thing.
> 
> I think these values from secrets files should apply to all fields, not just `SecretStr` and `SecretBytes`. This is because:
> 
> * it would make it more like other variable substitutions from arguments, environment variables and dot-env files, and thus less confusing
> * I often want field types to be something like `HttpUrl` or even `int`, restricting to `SecretStr` and `SecretBytes` would prevent that
> 
> Sorry, I should have thought about this before, but it didn't occur to me until now.
> 
> @mdgilene and @FichteFoll what do you think? Unless you have strong objections, I can either wait for you to make the change, or make it myself.

I don't have a strong preference one way or the other. So opening it up to all types works for me. I don't actually have much time at the moment, so if you are able to make the change that would be awesome. Otherwise it might take a few days till I get around to it.

samuelcolvin:
okay, thanks. I'll try and do it.

tests/test_settings.py
```diff
@@ -806,21 +806,23 @@ class Settings(BaseSettings):
         class Config:
             secrets_dir = p1
 
-    with pytest.raises(SettingsError, match=re.escape('secrets_dir must reference a directory, not a file')):
+    with pytest.raises(SettingsError, match='secrets_dir must reference a directory, not a file'):
         Settings()
 
 
+@pytest.mark.skipif(sys.platform.startswith('win'), reason='windows paths break regex')
```
FichteFoll, line 813: You could `.replace('/', r'[\\/]')`.

pydantic/env_settings.py
```diff
@@ -46,41 +45,34 @@ def _build_values(
...
+        Build fields from "secrets" files.
         """
         secrets: Dict[str, Optional[str]] = {}
 
         secrets_dir = _secrets_dir or self.__config__.secrets_dir
         if secrets_dir is None:
             return secrets
 
-        secrets_path = Path(secrets_dir)
+        secrets_path = Path(secrets_dir).expanduser()
```
FichteFoll, line 61: Is this documented?

tests/test_settings.py
samuelcolvin: I tried escaping, too boring without a Windows install.

pydantic/env_settings.py
samuelcolvin: No, but it's pretty obvious

PrettyWood: Related https://github.com/samuelcolvin/pydantic/pull/1804

PR #1830, opened 2020-08-13
documentation: remove strict_optional from sample mypy.ini

brianmaissy:
Really trivial suggested change, but strict_optional=True is the default for mypy since 0.600, so its presence in the sample mypy.ini in the docs seems redundant.

https://mypy.readthedocs.io/en/stable/config_file.html#none-and-optional-handling


PrettyWood (approved):
Indeed

samuelcolvin:
thanks.

Issue #1747, opened 2020-07-21
Literal enum values does not respect the use_enum_values config


eistaa:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.6.1
            pydantic compiled: True
                 install path: <path-to-virtualenv>/lib/python3.8/site-packages/pydantic
               python version: 3.8.1 (default, Dec 29 2019, 16:51:17)  [GCC 9.2.0]
                     platform: Linux-5.7.2-arch1-1-x86_64-with-glibc2.29
     optional deps. installed: ['typing-extensions']
```

When combining enum values with a `Literal` typing the `use_enum_values` config is not respected:

```py
from enum import Enum
from typing import Literal

from pydantic import BaseModel

class FruitEnum(Enum):
    pear = 'pear'
    banana = 'banana'

class CookingModel(BaseModel):
    fruit: FruitEnum

    class Config:
        use_enum_values = True

class LiteralCookingModel(BaseModel):
    fruit: Literal[FruitEnum.pear]

    class Config:
        use_enum_values = True

cm = CookingModel(fruit=FruitEnum.pear)
print(cm.dict())
# outputs: {'fruit': 'pear'}

lcm = LiteralCookingModel(fruit=FruitEnum.pear)
print(lcm.dict())
# outputs: {'fruit': <FruitEnum.pear: 'pear'>}
```

PrettyWood:
Closed by #1838

Issue #1832, opened 2020-08-13
Test case failures in Python 3.9


tirkarthi:
# Bug

`__args__` was removed in https://bugs.python.org/issue40397 and https://bugs.python.org/issue40398 . It seems a lot of places use it that causes test failures in Python 3.9. A quick grep of args shows many places. 

```
rg __args__ 
pydantic/fields.py
425:            for type_ in self.type_.__args__:
447:            for i, t in enumerate(self.type_.__args__):
449:                    self.type_ = self.type_.__args__[0]
463:            self.type_ = self.type_.__args__[0]
473:            self.type_ = self.type_.__args__[0]
476:            self.type_ = self.type_.__args__[0]
479:            self.type_ = self.type_.__args__[0]
482:            self.key_field = self._create_sub_type(self.type_.__args__[0], 'key_' + self.name, for_keys=True)
483:            self.type_ = self.type_.__args__[1]
488:            self.type_ = self.type_.__args__[0]
497:            self.sub_fields = [self._create_sub_type(t, f'{self.name}_{i}') for i, t in enumerate(self.type_.__args__)]

pydantic/types.py
135:    __args__: List[Type[T]]  # type: ignore
167:    # __args__ is needed to conform to typing generics api
168:    namespace = {'min_items': min_items, 'max_items': max_items, 'item_type': item_type, '__args__': [item_type]}
177:    __args__: Set[Type[T]]  # type: ignore
206:    # __args__ is needed to conform to typing generics api
207:    namespace = {'min_items': min_items, 'max_items': max_items, 'item_type': item_type, '__args__': [item_type]}

pydantic/schema.py
808:            args: Tuple[Any, ...] = type_.__args__

pydantic/typing.py
179:        return type_.__args__
250:            if type_.__args__ is None or not isinstance(type_.__args__[0], type):
252:            return type_.__args__[0]
```

**Sample test failure :** 

```
pytest -k test_list_bare  
=============================== test session starts ===============================
platform linux -- Python 3.9.0b5, pytest-5.3.1, py-1.5.3, pluggy-0.12.0
rootdir: /root/checked_repos/pydantic, inifile: setup.cfg, testpaths: tests
plugins: hypothesis-4.43.1, cov-2.8.1
collected 1754 items / 1753 deselected / 1 selected                               

tests/test_edge_cases.py F                                                  [100%]

==================================== FAILURES =====================================
_________________________________ test_list_bare __________________________________

    def test_list_bare():
[...]

PR #1838, opened 2020-08-16
Fix representation of literal enums (#1747)

noelevans:
## Change Summary

Fix for Literal enum values does not respect the use_enum_values config

## Related issue number

#1747

## Checklist


pydantic/main.py
```diff
@@ -14,13 +14,16 @@
...
+    Literal,
     Mapping,
     Optional,
     Tuple,
     Type,
     TypeVar,
     Union,
     cast,
+    get_origin,
+    get_type_hints,
```
PrettyWood, line 26: They don't exist in python 3.7-. We'll probably add `get_origin` and `get_type_hints` directly in `pydantic.typing` to support fallback for python 3.7-

pydantic/main.py
```diff
@@ -14,13 +14,16 @@
     Callable,
     Dict,
     List,
+    Literal,
```
PrettyWood, line 17: You should import it from `pydantic.typing`

pydantic/main.py
noelevans: Ok. I can look in to that

PrettyWood: You could have a look at #1832 and in particular this commit d86a184072655d93652d6ea941421005f34de962 that I quote in the end. But I feel like this should be done in a dedicated PR (I'll open one to add python 3.9 support soon) so maybe the best is just to wait a bit as you wish !

pydantic/main.py
noelevans: I'd started looking at the commit you had pointed to me and after a while realised the existing function `is_literal_type` could help out.

I'll investigate why this fails on Ubuntu 3.6 and 3.7 Python and not Windows / Mac. I'd written this on Ubuntu (testing with 3.6, 3.7 and 3.8) so this is a surprise.

changes/1747-noelevans.md
```diff
@@ -0,0 +1 @@
+Fix for issue 1747, properly representing Literal Enums when use\_enum\_values is True.
```
PrettyWood, line 1:
```suggestion
Properly represent `Literal` Enums when `use_enum_values` is True.
```
The issue will be linked automatically thanks to the file name and history generation

PrettyWood (commented):
I still don't know what is the expected behaviour with `Literal` and `use_enum_values`.
We could have simply changed `make_literal_validator` into
```python
def make_literal_validator(type_: Any, config: Type['BaseConfig']) -> Callable[[Any], Any]:
    allowed_choices_set = set(permitted_choices)

        if v not in allowed_choices_set:
        return v.value if isinstance(v, Enum) and config.use_enum_values else v

    return literal_validator
```
but it would have meant that both fields in model
```python
    class Model(BaseModel):
        baz: Literal[FooEnum.foo]
        boo: FooEnum = FooEnum.foo

        class Config:
            use_enum_values = True
```
would be considered the same.
With your implementation, they are equal only for dumping (`dict()`)

pydantic/main.py
```diff
@@ -673,6 +681,11 @@ def _get_value(
                 and (not value_include or value_include.is_included(i))
             )
 
+        elif getattr(cls.Config, 'use_enum_values', False) and any(
+            is_literal_type(v) for v in get_type_hints(cls).values()
```
PrettyWood, line 685: This will break if you have a second field in your model e.g.
```python
    class Model(BaseModel):
        baz: Literal[FooEnum.foo]
        boo: str
```
It will run `v.value` on `boo`
(And btw you should use another variable instead of a local `v`, which is confusing with the other one)

PrettyWood (changes requested):

noelevans:
> I still don't know what is the expected behaviour with `Literal` and `use_enum_values`.
> We could have simply changed `make_literal_validator` into
> 
> ```python
> def make_literal_validator(type_: Any, config: Type['BaseConfig']) -> Callable[[Any], Any]:
>     permitted_choices = all_literal_values(type_)
>     allowed_choices_set = set(permitted_choices)
> 
>     def literal_validator(v: Any) -> Any:
>         if v not in allowed_choices_set:
>             raise errors.WrongConstantError(given=v, permitted=permitted_choices)
>         return v.value if isinstance(v, Enum) and config.use_enum_values else v
> 
>     return literal_validator
> ```
> 
> but it would have meant that both fields in model
> 
> ```python
>     class Model(BaseModel):
>         baz: Literal[FooEnum.foo]
>         boo: FooEnum = FooEnum.foo
> 
>         class Config:
>             use_enum_values = True
> ```
> 
> would be considered the same.
> With your implementation, they are equal only for dumping (`dict()`)

Ok, I see. Perhaps @samuelcolvin or someone else could give an opinion on the best way to proceed?

pydantic/main.py
```diff
@@ -673,6 +681,13 @@ def _get_value(
                 and (not value_include or value_include.is_included(i))
             )
 
+        elif (
+            isinstance(v, Enum)
+            and getattr(cls.Config, 'use_enum_values', False)
+            and any(is_literal_type(hint) for hint in get_type_hints(cls).values())
```
samuelcolvin, line 687: I'm not quite sure what this line is doing?

Surely if `isinstance(v, Enum) and getattr(cls.Config, 'use_enum_values', False)` is enough?

tests/test_main.py
```diff
@@ -577,6 +578,22 @@ class Config:
     assert m.foo == 'foo'
 
 
+@pytest.mark.skipif(not Literal, reason='typing_extensions not installed')
+def test_literal_enum_values():
+    FooEnum = Enum('FooEnum', {'foo': 'foo', 'bar': 'bar'})
```
samuelcolvin, line 583:
```suggestion
    FooEnum = Enum('FooEnum', {'foo': 'foo_value', 'bar': 'bar_value'})
```

Makes the test clearer, otherwise we could be using `FooEnum.foo.name`

pydantic/main.py
noelevans: It seems so, yes

PR #1850, opened 2020-08-18
Fix mypy

hellocoldworld:
Using pydantic with mypy raises 

```python
venv/lib/python3.7/site-packages/pydantic/env_settings.py:128: error: Cannot find implementation or library stub for module named
'dotenv'  [import]
```

A simple `# type : ignore` in this line will do.


hellocoldworld:
Hmmm, CI fails saying this is unused but locally it fails for me - nevermind - must be an issue in my env! Sorry

Issue #1303, opened 2020-03-11
subclasses of BaseModel can be hashable


seansfkelley:
# Feature Request

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.4
            pydantic compiled: False
                 install path: /Users/username/Library/Caches/pypoetry/virtualenvs/virtualenvname/lib/python3.6/site-packages/pydantic
               python version: 3.6.4 (default, Mar  1 2018, 18:36:50)  [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]
                     platform: Darwin-18.7.0-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']
```

I use Pydantic extensively in place of dataclasses throughout my projects. It would be nice to be able to use some of the simpler types as dict keys, or to put into sets:

```py
import pydantic

class Foo(pydantic.BaseModel):
  foo: str = "foo"

d = { Foo(): "bar" }
```

I tried writing a superclass/mixin to selectively add this behavior to existing models:

```py
class HashableMixin:
    def __hash__(self):
        return hash(
            (type(self),) + tuple(getattr(self, f) for f in self.__fields__.keys())
        )
```

though this particular implementation has struggles, as it doesn't work when it's not the first in the list of inherited classes. I think this has something to do with Pydantic's initialization and maybe metaclasses, but I didn't dig too deep. So I wrote it as a decorator instead:

```py
def hashable(cls):
    def h(self):
        return hash(
            (type(self),) + tuple(getattr(self, f) for f in self.__fields__.keys())
        )
    setattr(cls, "__hash__", h)
    return cls
```

which seems to work more or less alright, though I haven't really run it through its paces so I don't know if I've missed anything.

Anyway, it would be great to have this baked in, even if it were default off. Maybe with something on `Config`?

samuelcolvin:
I would create your own `MyBaseModel` and use that in place of `BaseModel` to accomplish this.

```py
from pydantic import BaseModel

class MyBaseModel(BaseModel):
    def __hash__(self):
        return hash((type(self),) + tuple(self.__dict__.values()))

class Foo(MyBaseModel):
    foo: str = 'foo'

f = Foo()
d = {f: 'bar'}
print(d)
```

---

> Anyway, it would be great to have this baked in, even if it were default off. Maybe with something on `Config`?

I would argue that this is:
1. very easy to achieve as demonstrated above
[...]

PR #1881, opened 2020-08-30
Generate a hash function when `frozen` is `True`

### LINKED ISSUE #1880: Generate a hash function when `allow_mutation` is `False`
# Feature Request

```
             pydantic version: 1.6.1
            pydantic compiled: False
                 install path: /home/raphael/rhuille/pydantic/pydantic
               python version: 3.8.5 (default, Aug 18 2020, 10:43:36)  [GCC 7.5.0]
                     platform: Linux-5.0.0-1067-oem-osp1-x86_64-with-glibc2.27

```
Hi there,

I need the model to be hashable when I set `allow_mutation` to `False`. I could write myself a `__hash__` function, but I think it could be nice if pydantic generated it by default in the class `BaseModel`.

This would be the same behavior of the built-in `dataclass` which generate a hash function if the object is immutable (parameter `frozen` is `True`) and the `__eq__` function exists. 

Quoting the documentation https://docs.python.org/3/library/dataclasses.html :
```
If eq and frozen are both true, by default dataclass() will generate a __hash__() method for you.
```

We could generate a default `__hash__` function for `BaseModel` if `allow_mutation` is `False`, what do you think ?

### Examples

For example, now we have:
```py

class A(BaseModel):
    x: int
    
    class Config:
        allow_mutation = False

a = A(x=1)
d = {a: 2}

>>> TypeError: unhashable type: 'A'
```

After, we would have:
```py

class A(BaseModel):
    x: int
    
    class Config:
        allow_mutation = False

a = A(x=1)
d = {a: 2}
d[a]
>>> 2
```

(Note that if `allow_mutation = True`, the hash function would be `None` and the behavior would not be changed)

rhuille:
## Change Summary

**Before:**

```

class A(BaseModel):
    x: int
    
    class Config:
        allow_mutation = False

a = A(x=1)
d = {a: 2}

>>> TypeError: unhashable type: 'A'
```

**After:**

```py

class A(BaseModel):
    x: int
    
    class Config:
        frozen = True

a = A(x=1)
d = {a: 2}
d[a]
>>> 2
```

Note that we still have:

```py

class A(BaseModel):
    x: int
    
    class Config():
        frozen = True

a = A(x=1)
d = {a: 2}
d[a]
>>> TypeError: unhashable type: 'A'
```

Screenshot of the updated documentation:

image

## Related issue number
closes #1880

linked issue: #1303

## Checklist


ccharlesgb:
I think this is a good idea but I don't think this will work if the fields in the model cannot be hashed. For example what happens if you do the following:

```python

class A(BaseModel):
    x: int
    y: List[int]
    
    
    class Config:
        allow_mutation = False

a = A(x=1, y=[1,2,3])
d = {a: 2}
d[a]
>>> ?
```

Might be worth putting in a validation step to ensure that the fields are hashable?

Or alternatively `dataclasses` achieves this by assigning a `hash` attribute to each field and the overall model's hash is the combination of al of these. See: https://github.com/ericvsmith/dataclasses/blob/master/dataclasses.py#L540

I think this would be a better approach as it should allow you to do nested immutability:

```python

class B(BaseModel):
    z: int
    class Config:
        allow_mutation = False


class A(BaseModel):
    x: int
    y: B
    
    
    class Config:
        allow_mutation = False

a = A(x=1, y=B(z=4))
d = {a: 2}
```

layday:
I've had to roll my own `__hash__` function for Pydantic models in the past, so I'm very excited at the prospect of natively 'frozen' models.  A few quick notes - you cannot hash `self.__dict__.values()` directly because the `dict_values` container is unique.  This is why `dataclasses` places them in a tuple before hashing.  You also cannot (should not) implement a `__hash__` which behaves differently from `__eq__`:

```

In [2]: class Foo(BaseModel):
   ...:     def __hash__(self):
   ...:         return hash(self.__dict__.values())
   ...:

In [3]: Foo() == Foo()
Out[3]: True

In [4]: hash(Foo())
Out[4]: 276660083

In [5]: hash(Foo())
Out[5]: 277242134

In [6]: d = {Foo(): True}

In [7]: d[Foo()] = False

In [8]: d.keys()
Out[8]: dict_keys([Foo(), Foo()])
```

The `dataclasses` implementation mentioned above is more sophisticated (e.g. it compares values which cannot be hashed) and might be worth investigating.

rhuille:
Hi @ccharlesgb and @layday , thanks for your answer.

I agree with your comments and I corrected my implementation of `__hash__` in: https://github.com/samuelcolvin/pydantic/pull/1881/commits/b1a84dc30091bc94f5457d5db0451885f0d7b7dd

What do you think ?

If I am correct, this has now the same behavior has in `dataclass`:

```
from typing import List


class A(BaseModel):
    x: int
    y: int
    
    class Config:
        allow_mutation = False


@dataclass(frozen=True)
class B:
    x: int
    y: int
 
hash(A(x=1, y=2)) == hash(B(x=1, y=2))
>>> True

hash(A(x=1, y=2)) == hash((1, 2))
>>> True

class C(BaseModel):
    x: int
    y: List[int]
    class Config:
        allow_mutation = False

@dataclass(frozen=True)
class D:
    x: int
    y: List[int]

hash(C(x=1, y=[1, 2, 3]))
>>>TypeError: unhashable type: 'list'

hash(D(x=1, y=[1, 2, 3]))
>>>TypeError: unhashable type: 'list'
```

pydantic/main.py
```diff
@@ -313,6 +313,7 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
             '__schema_cache__': {},
             '__json_encoder__': staticmethod(json_encoder),
             '__custom_root_type__': _custom_root_type,
+            '__hash__': (lambda self: hash(tuple(self.__dict__.values()))) if not config.allow_mutation else None,
```
wozniakty, line 316: Instead of `.values` we should probably have `.items` here.  The keys ought to be considered a part of the identity of the object.

Example:
```
x = {"a": 1, "b": True}
y = {"d": 1, "c": True}
hash(tuple(x.values())) == hash(tuple(y.values()))  # True
hash(tuple(x.items())) == hash(tuple(y.items()))  # False
```

tests/test_main.py
```diff
@@ -373,6 +386,32 @@ class Config:
...
+        a: int = 10
+        y: List[int] = [1, 2, 3]
+
+        class Config:
+            allow_mutation = False
+
+    m = TestModel()
+    with pytest.raises(TypeError) as exc_info:
+        hash(m)
+    assert "unhashable type: 'list'" in exc_info.value.args[0]
```
wozniakty, line 412: IMO would be nice for the exception message to be more like "unhashable type: 'TestModel' contains unhashable type: 'list'  but I wouldn't say it's a strong opinion.  Food for thought.

pydantic/main.py
rhuille: In that case, we diverge from behaviour of the built-in dataclass. But I do not know if it is a problem.
If it not a problem, then I agree with you ! :)

samuelcolvin: I don't think this needs to be a lambda, it can be proper method on `BaseModel` I imagine. It could raise an error if `allow_mutation is True` .

samuelcolvin (commented):
I think this is a good start, but my problem here is that I imagine in the following code `f` and `b` would have the same hash and therefore be "equal"

```py
class Foo(BaseModel):
    a: str
    b: int

    class Config:
        allow_mutation = False


class Bar(BaseModel):
    a: str
    b: int

    def dict(self, *args, **kwargs):
        d = super().dict(*args, **kwargs)
        d.pop('a')
        return d

    class Config:
        allow_mutation = False


f = Foo(a='xx', b=2)
b = Bar(a='xx', b=3)
```

Which is definitely not what people would expected.

I think we need to add some kind of reference to the qualname to the hash.

I think we also need to hide this behind another config option.

layday:
> I don't think this needs to be a lambda, it can be proper method on `BaseModel` I imagine. It could raise an error if `allow_mutation is True`.

The 'correct' way to make an object unhashable is by setting its `__hash__` attribute to `None`:

```
In [1]: class Foo:
   ...:     pass
   ...:

In [2]: class Bar:
   ...:     __hash__ = None
   ...:

In [3]: hash(Foo())
Out[3]: 284941123

In [4]: hash(Bar())
TypeError                                 Traceback (most recent call last)
<ipython-input-4-d413494abdfb> in <module>
----> 1 hash(Bar())

TypeError: unhashable type: 'Bar'
```

This would have to happen on the metaclass and is done automatically by the Python interpreter for classes which implement `__eq__` but not `__hash__`:

```
In [5]: from pydantic import BaseModel

In [6]: class Baz(BaseModel):
   ...:     pass
   ...:

In [7]: hash(Baz())
TypeError                                 Traceback (most recent call last)
<ipython-input-7-67f7cb492b68> in <module>
----> 1 hash(Baz())

TypeError: unhashable type: 'Baz'

In [8]: Baz.__hash__

In [9]: class Bee:
   ...:     def __eq__(self, other):
   ...:         ...
   ...:

In [10]: Bee.__hash__

In [11]: Foo.__hash__
Out[11]: <slot wrapper '__hash__' of 'object' objects>
```

If `__hash__` were implemented on `BaseModel`, we'd have to emulate the default behaviour.  In addition,  if`__hash__` is not `None`, `isinstance(obj, typing.Hashable)` will not have the desired effect:

```
In [12]: from typing import Hashable

In [13]: isinstance(Bee(), Hashable)
Out[13]: False

In [14]: class FauxUnhashable:
    ...:     def __hash__(self):
    ...:         raise TypeError
    ...:

In [51]: isinstance(FauxUnhashable(), Hashable)
Out[51]: True
```

pydantic/main.py
layday: > I don't think this needs to be a lambda, it can be proper method on `BaseModel` I imagine. It could raise an error if `allow_mutation is True`.

The 'correct' way to make an object unhashable is by setting its `__hash__` attribute to `None`:

```
In [1]: class Foo:
   ...:     pass
   ...:

In [2]: class Bar:
   ...:     __hash__ = None
   ...:

In [3]: hash(Foo())
Out[3]: 284941123

In [4]: hash(Bar())
TypeError                                 Traceback (most recent call last)
<ipython-input-4-d413494abdfb> in <module>
----> 1 hash(Bar())

TypeError: unhashable type: 'Bar'
```

This would have to happen on the metaclass and is done automatically by the Python interpreter for classes which implement `__eq__` but not `__hash__`:

```
In [5]: from pydantic import BaseModel

In [6]: class Baz(BaseModel):
   ...:     pass
   ...:

In [7]: hash(Baz())
TypeError                                 Traceback (most recent call last)
<ipython-input-7-67f7cb492b68> in <module>
----> 1 hash(Baz())

TypeError: unhashable type: 'Baz'

In [8]: Baz.__hash__

In [9]: class Bee:
   ...:     def __eq__(self, other):
   ...:         ...
   ...:

In [10]: Bee.__hash__

In [11]: Foo.__hash__
Out[11]: <slot wrapper '__hash__' of 'object' objects>
```

If `__hash__` were implemented on `BaseModel`, we'd have to emulate the default behaviour.  In addition,  if `__hash__` is not `None`, `isinstance(obj, typing.Hashable)` will not have the desired effect:

```
In [12]: from typing import Hashable

In [13]: isinstance(Bee(), Hashable)
Out[13]: False

In [14]: class FauxUnhashable:
    ...:     def __hash__(self):
    ...:         raise TypeError
    ...:

In [51]: isinstance(FauxUnhashable(), Hashable)
Out[51]: True
```

Reference: https://docs.python.org/3/reference/datamodel.html#object.__hash__

samuelcolvin: okay, so it can't just be a method on `BaseModel`, but let's make it a proper function, not a lambda.

layday:
> I think we need to add some kind of reference to the qualname to the hash.

Would hashing the model class not work, e.g. `hash((self.__class__,) + tuple(self.__dict__.items()))`?  I know attrs combines the `__module__` name with the `__qualname__` but that might be because of something specific to attrs.

samuelcolvin:
I guess it might work, attrs might do it that way to support older versions of python.

samuelcolvin:
as long as this feature is hidden behind a 'hashable` config parameter, I'm not too bothered. We can say it's new and in "beta" or something in the docs.

pydantic/main.py
layday: > Instead of `.values` we should probably have `.items` here. The keys ought to be considered a part of the identity of the object.

I think this would unnecessary if we include a reference to the model class.

layday:
hmm, would `hashable = True` imply `allow_mutation = False` or would they both need to be declared?  I don't think `hashable` is useful by itself or something that should be exposed - `hashable` without `allow_mutation` should be disallowed, in fact.  How about adding a new `frozen` setting which would (or could) supersede `allow_mutation` in 2.0?

rhuille:
So 2 things to modify:

- include the name of the class in the hash --> so we agree to diverge from the built in dataclass behavior ?

- add a boolean parameter `hashable` in BaseModel config. This parameter would have the following behavior:
   - if `hashable=True`: implement a hash function, 
            This hash function would raise an error if `allow mutation=True`. This is a way to "disallow hashable without allow_mutation" @layday  what do you think ?
            And to overcome this error, pydantic users should implement themself a hash function in the class model.
   - if `hashable=False` : the hash function is None.

layday:
* The class name is not unique.  We'd have to either combine the module name with the `__qualname__` or hash the class itself as suggested. 
* The hash function should not raise, a configuration error should be raised on model creation by the metaclass.  My preference is for `hashable` not being exposed.

samuelcolvin:
> How about adding a new `frozen` setting which would (or could) supersede allow_mutation in 2.0?

Sounds good to me.

> or hash the class itself as suggested

also sounds good to me.

samuelcolvin:
@rhuille are you going to finish this? otherwise let's lose it.

rhuille:
Hi @samuelcolvin yes I am doing this now. Thanks for the specifications.

rhuille:
Hi @samuelcolvin I have reworked the PR (and updated the description) to match the specifications specified by @layday and you. Thanks a lot for your feedback !

changes/1880-rhuille.md
```diff
@@ -0,0 +1 @@
+Add a new `frozen` boolean parameters. Setting `frozen=True` does everything that `allow_mutation=False` does, and also generate a hash function for the model i.e. `__hash__` is not `None`. This makes instances of the model potentially hashable if all the attributes are hashable. (default: `False`)
```
PrettyWood, line 1:
```suggestion
Add a new `frozen` boolean parameters. Setting `frozen=True` does everything that `allow_mutation=False` does, and also generates a `__hash__()` method for the model. This makes instances of the model potentially hashable if all the attributes are hashable. (default: `False`)
```

docs/usage/model_config.md
```diff
@@ -26,6 +26,13 @@ Options:
 **`allow_mutation`**
 : whether or not models are faux-immutable, i.e. whether `__setattr__` is allowed (default: `True`)
 
+**`frozen`**
+
+!!! warning
+    This parameters is beta
```
PrettyWood, line 32:
```suggestion
    This parameter is in beta
```

docs/usage/model_config.md
```diff
@@ -26,6 +26,13 @@ Options:
...
+
+: setting `frozen=True` does everything that `allow_mutation=False` does, and also generate a hash function for the model i.e. `__hash__` is not `None`. This makes instances of the model potentially hashable if all the attributes are hashable. (default: `False`)
```
PrettyWood, line 34: Same remark as change file

pydantic/main.py
```diff
@@ -321,6 +329,7 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
             '__custom_root_type__': _custom_root_type,
             '__private_attributes__': private_attributes,
             '__slots__': slots | private_attributes.keys(),
+            '__hash__': generate_hash_function(config.frozen, super()),
```
PrettyWood, line 332: `super()` is the metaclass here. I don't understand why you need to do that 🤔

pydantic/main.py
```diff
@@ -198,6 +199,13 @@ def validate_custom_root_type(fields: Dict[str, ModelField]) -> None:
         raise ValueError('__root__ cannot be mixed with other fields')
 
 
+def generate_hash_function(frozen: bool, super_class: Any) -> Optional[Callable[[Any], int]]:
+    def hash_function(self_: Any) -> int:
+        return hash(super_class) + hash(tuple(self_.__dict__.values()))
```
PrettyWood, line 204: IMO you could just do `hash(self_.__class__)` instead of using `super_class`

PrettyWood (dismissed):

pydantic/main.py
rhuille: Yes your are right ! https://github.com/samuelcolvin/pydantic/pull/1881/commits/12cff48de640035718d1308c04eb007fe5e1ef05 (I was overthinked this...)

changes/1880-rhuille.md
rhuille: https://github.com/samuelcolvin/pydantic/pull/1881/commits/0d6eaec4bebb9e06fc2042f175a9c6fbf30deb18

docs/usage/model_config.md
rhuille: https://github.com/samuelcolvin/pydantic/pull/1881/commits/0d6eaec4bebb9e06fc2042f175a9c6fbf30deb18

docs/usage/model_config.md
rhuille: https://github.com/samuelcolvin/pydantic/pull/1881/commits/0d6eaec4bebb9e06fc2042f175a9c6fbf30deb18

changes/1880-rhuille.md
```diff
@@ -0,0 +1 @@
+Add a new `frozen` boolean parameters. Setting `frozen=True` does everything that `allow_mutation=False` does, and also generates a `__hash__()` method for the model. This makes instances of the model potentially hashable if all the attributes are hashable. (default: `False`)
```
PrettyWood, line 1:
```suggestion
Add a new `frozen` boolean parameter to `Config` (default: `False`).
Setting `frozen=True` does everything that `allow_mutation=False` does, and also generates a `__hash__()` method for the model. This makes instances of the model potentially hashable if all the attributes are hashable.
```

PrettyWood (commented):
Looks quite good.
Do we want to tackle the case `frozen=True` and `allow_mutation=True` and raise an error in this case ?

changes/1880-rhuille.md
rhuille: https://github.com/samuelcolvin/pydantic/pull/1881/commits/7397b678af3ef098d7e038a185e22f4edd973dfc

rhuille:
Good question ! For now, if `frozen=True` and `allow_mutation=True` the behavior is: "`frozen=True` is applied and `allow_mutation=True` is implicitly ignored". I guess I could add a warning message explaining that.

tests/mypy/modules/plugin_fail.py
```diff
@@ -202,3 +202,37 @@ class AddProject:
...
+
+    def method(self) -> None:
+        pass
+
+    class Config:
+        alias_generator = None
+        frozen = True
+        extra = Extra.forbid
+
+        def config_method(self) -> None:
```
samuelcolvin, line 220: do you need this?

tests/mypy/modules/plugin_fail.py
```diff
@@ -202,3 +202,37 @@ class AddProject:
...
 
 p = AddProject(name='x', slug='y', description='z')
+
+
+# Same as Model, but with frozen = True
+class FrozenModel(BaseModel):
+    x: int
+    y: str
+
+    def method(self) -> None:
```
samuelcolvin, line 212: can this be removed too?

tests/mypy/modules/plugin_fail.py
```diff
@@ -202,3 +202,37 @@ class AddProject:
...
+    class Config:
+        alias_generator = None
+        frozen = True
+        extra = Extra.forbid
+
+        def config_method(self) -> None:
+            ...
+
+
+frozenmodel = FrozenModel(x=1, y='y', z='z')
```
samuelcolvin, line 224: this isn't testing frozen behaviour I think, again can be removed.

tests/mypy/modules/plugin_fail.py
```diff
@@ -202,3 +202,37 @@ class AddProject:
...
+frozenmodel = FrozenModel(x=1)
+frozenmodel.y = 'a'
+FrozenModel.from_orm({})
+FrozenModel.from_orm({})  # type: ignore[pydantic-orm]  # noqa F821
```
samuelcolvin, line 228: again, are these lines testing frozen behaviour?

tests/mypy/modules/plugin_success.py
```diff
@@ -139,3 +139,22 @@ class Model(BaseModel):
...
+
+class FrozenModel(BaseModel):
+    x: int
+
+    class Config:
+        frozen = True
+
+
+class NotFrozenModel(FrozenModel):
+    a = 1
```
samuelcolvin, line 152: add a type here to avoid extra errors in `plugin-success-strict.txt `

tests/test_main.py
```diff
@@ -351,39 +351,96 @@ class Model(BaseModel):
...
-    assert m.a == 10
-    m.a = 11
-    assert m.a == 11
-    with pytest.raises(ValueError) as exc_info:
-        m.b = 11
-    assert '"TestModel" object has no field "b"' in exc_info.value.args[0]
+
+    immutable = allow_mutation_ is False or frozen_ is True
+
+    if immutable:
```
samuelcolvin, line 368: rather than this if clause, better to setup two separate tests.

samuelcolvin (commented):
otherwise I think this looks great.

rhuille:
Hi @samuelcolvin , thanks for your review !

Since your review:

- I resolved the conflict with master: https://github.com/samuelcolvin/pydantic/pull/1881/commits/4815e4c42c3a4dd330da6c56c544f28c1af24897
- Following your comments:
    - I reworked the mypy tests so it test only the frozen behavior: https://github.com/samuelcolvin/pydantic/pull/1881/commits/7c3f4debf2ba2e6ad6c5f0659c2e1e89d74712a7 and 
    - I split the test function in 2: https://github.com/samuelcolvin/pydantic/pull/1881/commits/a6943a81bac2f0faf95bf33985eca3763808ab88 & https://github.com/samuelcolvin/pydantic/pull/1881/commits/3e98d572173c6ac46fc31a16298da2210a75fdd1 & https://github.com/samuelcolvin/pydantic/pull/1881/commits/ad82bf0484be65980cc6e9fdba2b5df1b91e7b38

Also, if you want me to, I can clean the commits history.


Thanks a lot, I am very proud of this work :D

samuelcolvin:
This is great, thank you so much.

davidhyman:
Hi, I've been playing with master (at 90df33cf1e01e0235e6069715a251e718bc07753) and am concerned that this eliminates the ability for a user-defined `__hash__`;

I'm all in favour of auto-generating it upon request, but I wonder if it would be less intrusive to user code to avoid setting `__hash__` in `frozen=False`? And even if it is `frozen=True`, perhaps a `__hash__` doesn't need generating if it's user-defined (less sure on that)?

propose changing
```
    '__hash__': generate_hash_function(config.frozen),
```
to
```
if config.frozen:
    new_namespace['__hash__'] = generate_hash_function()
```
(or some other mechanism e.g. popping again from the new namespace)

I might be missing something though - is there a suggested approach that would let me define the hash for my model? Without it, I can't put my model instances in a set, dictionary etc.

Motivating example: having a model which the `generate_hash_function` presently can't handle, such as nested complex generic types; or would be entirely unhashable in the usual sense due to mutable containers, but using a unique database id for reference - so as far as the application is concerned, the db ID is all that's needed for hashing?

PrettyWood:
I don't understand @davidhyman
For me the behaviour is right. When you write a custom `__hash__` it actually overwrites the default `__hash__` whether or not `frozen` is set.
Could you please share a code snippet of the problem?

davidhyman:
@PrettyWood I think you have fixed it in https://github.com/samuelcolvin/pydantic/pull/2423
and it was reported as
https://github.com/samuelcolvin/pydantic/issues/2422

Thank you! (and thanks to all the other contributors)

Sorry I was late to the party. I spent a bit of time on a repro today, originally thought I couldn't reproduce it and then realised that it's fixed (at least, in 1.8.1). I don't think I'd realised it was an inheritance thing; recalled commenting out the __hash__ assignment, so was sure it was something that had originally changed in this MR :thinking: .

```python
import pydantic

class P(pydantic.BaseModel):
    a: int

    def __hash__(self):
        """custom hash function"""
        return self.a

class M(P):
    """our hash function is still overwritten."""

    b: dict = dict()

    class Config:
        """pydantic config."""

        allow_mutation = False
        frozen = True

m = M(a=5)
assert m.__hash__() == 5
```

Daniel-Chin:
has this behavior been documented?

PR #1920, opened 2020-09-13
add smart_deepcopy (originaly from #1679)

MrMrRobat:
## Change Summary
add `smart_deepcopy()`:
```py
smart_deepcopy(obj: ~Obj) -> ~Obj
    Return type as is for immutable built-in types
    Use obj.copy() for built-in empty collections
    Use copy.deepcopy() only on non-empty collections and unknown objects
```

It's primarily needed for faster copying of default values, since they're often are immutable types or empty collections

Here's benchmark showing the difference in speed between different types of values:
```py
from copy import deepcopy
from timeit import timeit

from pydantic.utils import smart_deepcopy, BUILTIN_COLLECTIONS

cases = {
    "IMMUTABLE":
        (1, 1.0, '1', b'1', int, None, smart_deepcopy, len, smart_deepcopy.__code__, lambda: _, ...),
    "EMPTY_COLLECTIONS":
        map(lambda collection: collection(), BUILTIN_COLLECTIONS),
    "NON_EMPTY_COLLECTIONS":
        map(lambda c: c.fromkeys([1]) if issubclass(c, dict) else c([1]), BUILTIN_COLLECTIONS),
}

for name, values in cases.items():
    deepcopy_results = []
    smart_deepcopy_results = []
    for value in values:
        deepcopy_results.append(timeit(lambda: deepcopy(value), number=100000))
        smart_deepcopy_results.append(timeit(lambda: smart_deepcopy(value), number=100000))

    deepcopy_result = sum(deepcopy_results)
    smart_deepcopy_result = sum(smart_deepcopy_results)
    faster_by = deepcopy_result / smart_deepcopy_result
    print(f"{name}: {deepcopy_result=:.3}s, {smart_deepcopy_result=:.3}s, {faster_by=:.3} times")

```
```py
IMMUTABLE: deepcopy_result=0.658s, smart_deepcopy_result=0.235s, faster_by=2.8 times
EMPTY_COLLECTIONS: deepcopy_result=1.73s, smart_deepcopy_result=0.261s, faster_by=6.62 times
NON_EMPTY_COLLECTIONS: deepcopy_result=2.33s, smart_deepcopy_result=2.5s, faster_by=0.933 times
```

## Checklist

### DIFF
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -53,7 +53,7 @@ __all__ = (
     'ClassAttribute',
 )
 
-# these are types that returned by deepcopy unchanged
+# these are types that are returned unchanged by deepcopy
 IMMUTABLE_NON_COLLECTIONS_TYPES: Set[Type[Any]] = {
     int,
     float,
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -1,6 +1,9 @@
 import warnings
+import weakref
+from collections import OrderedDict, defaultdict, deque
+from copy import deepcopy
 from itertools import islice
-from types import GeneratorType
+from types import BuiltinFunctionType, CodeType, FunctionType, GeneratorType, LambdaType, ModuleType
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
@@ -20,7 +23,7 @@ from typing import (
     no_type_check,
 )
 
-from .typing import display_as_type
+from .typing import NoneType, display_as_type
 from .version import version_info
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


pydantic/utils.py
```diff
@@ -50,6 +53,41 @@
     'ClassAttribute',
 )
 
+# these are types that returned by deepcopy unchanged
```
PrettyWood, line 56:
```suggestion
# these are types that are returned unchanged by deepcopy
```

PrettyWood (commented):
Nice !

pydantic/utils.py
```diff
@@ -534,3 +572,22 @@ def __get__(self, instance: Any, owner: Type[Any]) -> None:
         if instance is None:
             return self.value
         raise AttributeError(f'{self.name!r} attribute of {owner.__name__!r} is class-only')
+
+
+Obj = TypeVar('Obj')
+
+
+def smart_deepcopy(obj: Obj) -> Obj:
```
wozniakty, line 580: Small food for thought that `smart_deepcopy` seems not particularly descriptive about what this function does, but I cannot come up with a better name so just offering food for thought in case someone else has an idea 😅

MrMrRobat: I also thought about «fast_deepcopy», but function is actually a bit slower in case it has to deepcopy value, so «smart» seemed a better choice.

I agree, it’s kinda vague word, so would love hear any better suggestions :)

samuelcolvin: I agree with both, it's not a perfect name, but I can't think of anything better.

samuelcolvin:
thanks so much, sorry I've taken so long to review this.

samuelcolvin:
I guess you can now update #1679

PR #1944, opened 2020-09-22
ensure cythonized functions are left untouched

### LINKED ISSUE #1943: Impl. of BaseModel with method crashes when Cythonized
# Bug

```
             pydantic version: 1.6.1
            pydantic compiled: True
                 install path: /Users/mordaren/Source/pydantic/pydantic
               python version: 3.7.7 (default, Mar 10 2020, 15:43:27)  [Clang 10.0.0 (clang-1000.11.45.5)]
                     platform: Darwin-17.2.0-x86_64-i386-64bit
```

Here's a minimal example:

main.py:
```py

class MyModel(BaseModel):
    x: float = 123

    def x2(self):
        return self.x**2

def main():
    model = MyModel(x=10)
    print("x^2 = ", model.x2())
```

setup.py:
```
from setuptools import setup
from Cython.Build import cythonize

setup(
    name='testapp',
    ext_modules=cythonize("main.py"),
)
```

Cythonize the script using the following command (mac OS): 

``` python setup.py build_ext --inplace```

When trying to import main, the following crash occurs (ensure that you're importing the actual Cythonized file):
```
  File "main.py", line 5, in init main
    class MyModel(BaseModel):
  File "pydantic/main.py", line 285, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/fields.py", line 309, in pydantic.fields.ModelField.infer
  File "pydantic/fields.py", line 271, in pydantic.fields.ModelField.__init__
  File "pydantic/fields.py", line 351, in pydantic.fields.ModelField.prepare
  File "pydantic/fields.py", line 529, in pydantic.fields.ModelField.populate_validators
  File "pydantic/validators.py", line 593, in find_validators
RuntimeError: no validator found for <class 'cython_function_or_method'>, see `arbitrary_types_allowed` in Config
```

kollmats:
## Change Summary

Extending the check for untouched types to include a check for cython functions or methods.

## Related issue number
Resolves #1943

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -245,7 +245,12 @@ class ModelMetaclass(ABCMeta):
         class_vars = set()
         if (namespace.get('__module__'), namespace.get('__qualname__')) != ('pydantic.main', 'BaseModel'):
             annotations = resolve_annotations(namespace.get('__annotations__', {}), namespace.get('__module__', None))
-            untouched_types = UNTOUCHED_TYPES + config.keep_untouched
+
+            def is_untouched(value):
+                untouched_types = UNTOUCHED_TYPES + config.keep_untouched
+
+                return isinstance(value, untouched_types) or value.__class__.__name__ == 'cython_function_or_method'
+
             # annotation only fields need to come first in fields
             for ann_name, ann_type in annotations.items():
                 if is_classvar(ann_type):
@@ -254,7 +259,7 @@ class ModelMetaclass(ABCMeta):
                     validate_field_name(bases, ann_name)
                     value = namespace.get(ann_name, Undefined)
                     if (
-                        isinstance(value, untouched_types)
+                        is_untouched(value)
                         and ann_type != PyObject
                         and not lenient_issubclass(getattr(ann_type, '__origin__', None), Type)
                     ):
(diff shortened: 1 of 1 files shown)


samuelcolvin:
please can you add tests and a change description.

samuelcolvin:
bump

samuelcolvin:
@kollmats please could you fixthis.

kollmats:
Hi @samuelcolvin, 

I'm sorry for this question since it may already be stated clearly somewhere, but where would be a good place to test this in your setup given that the a proper test would require a cythonized application? Thanks.

samuelcolvin:
all tests are run on pydantic when compiled with cython.

Just choose the file in `tests/` that makes most sense.

PrettyWood:
Made a test on https://github.com/samuelcolvin/pydantic/pull/2208 just to see the CI fail with a test similar to the one in the issue and it didn't fail. I probably miss something

samuelcolvin:
I see the problem, the point is that the code where `get_double_a` is defined (using the example from #2208) needs to be compiled.

It doesn't matter one way or the other whether or not pydantic itself is compiled.

I'm just seeing if I can add a simple test for this.

samuelcolvin:
in the end I created another PR to fix this, see #2228. @kollmats I've thanked you in the change description, hope that is sufficient.

Thanks for this.

Issue #1423, opened 2020-04-23
More sophisticated validation for Unions to avoid coercion when not needed


jvolkman:
# Bug

I've run into an issue in which an object's `__field_set__` is fully populated depending on a containing Union's order, despite no fields being set.

The example below defines a `Parent` class and a `Child` subclass. If `Parent()` is set on a `Union[Child, Parent]` field, all of its fields will be listed in `__fields_set__`.

Hopefully the example and output below do a better job of explaining.

```py
from typing import Union
import pydantic


class Parent(pydantic.BaseModel):
    field_that_should_be_unset: str = None


class Child(Parent):
    pass


class Container(pydantic.BaseModel):
    parent_first: Union[Parent, Child] = None
    child_first: Union[Child, Parent] = None


print(
    Container(
        parent_first=Parent(),
        child_first=Parent(),
    ).dict(exclude_unset=True)
)
print(
    Container(
        parent_first=Child(),
        child_first=Child(),
    ).dict(exclude_unset=True)
)
```

Output:
```
{'parent_first': {}, 'child_first': {'field_that_should_be_unset': None}}
{'parent_first': {}, 'child_first': {}}
```

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.4
            pydantic compiled: True
                 install path: /home/jvolkman/repos/tms/env/lib/python3.8/site-packages/pydantic
               python version: 3.8.2 (default, Mar 13 2020, 10:14:16)  [GCC 9.3.0]
                     platform: Linux-5.4.0-26-generic-x86_64-with-glibc2.29
     optional deps. installed: ['typing-extensions']

and

             pydantic version: 1.5
            pydantic compiled: True
                 install path: /home/jvolkman/repos/tms/env/lib/python3.8/site-packages/pydantic
               python version: 3.8.2 (default, Mar 13 2020, 10:14:16)  [GCC 9.3.0]
                     platform: Linux-5.4.0-26-generic-x86_64-with-glibc2.29
     optional deps. installed: ['typing-extensions']

```

samuelcolvin:
We might one day have more sophisticated logic to deal with unions, perhaps linked to #619.

Currently there's no work around for this:
* `Parent` is not a subclass of `Child`
* however `Parent` can be coerced to a child via effectively `Child(**parent.dict())`
* doing `Child(**parent.dict())` obviously sets `child. field_that_should_be_unset`, hence the confusing behaviour

[...]

Issue #619, opened 2019-06-24
[Feature Request] Provide a discriminated union type (OpenAPI 3)


sm-Fifteen:
# Feature Request

Pydantic currently has a decent support for union types through the `typing.Union` type from PEP484, but it does not currently cover all the cases covered by the JSONSchema and OpenAPI specifications, most likely because the two specifications diverge on those points.

OpenAPI supports something similar to tagged unions where a certain field is designated to serve as a "discriminator", which is then matched against literal values to determine which of multiple schemas to use for payload validation. In order to allow Pydantic to support those, I suppose there would have to be a specific type similar to `typing.Union` in order to specify what discriminator field to use and how to match it. Such a type would then be rendered into a schema object (`oneOf`) with an [OpenAPI discriminator object][2] built into it, as well as correctly validate incoming JSON into the correct type based on the value or the discriminator field. This change would only impact OpenAPI, as JSON schema (draft 7 onwards) uses [conditional types][3] instead, which would probably need to be the topic of a different feature request, as both methods appear mutually incompatible.

## Implementation ideas
I'd imagine the final result to be something like this.

```py
MyUnion = Union[Foo, Bar]
MyTaggedUnion = TaggedUnion(Union[Foo, Bar], discriminator='type', mapping={'foo': Foo, 'bar': Bar}))
```

Python doesn't have a feature like TypeScript to let you statically ensure that `discriminator` [exists as a field for all variants of that union][4], though that shouldn't be a problem since this is going to be raised during validation regardless.

`discriminator` and `mapping` could also simply be added to `Schema`, though I'm not sure about whether it's a good idea to add OpenAPI-specific extensions there.

[PEP 593][1] would also have been a nice alternative, since it would hypothetically allow tagged unions to be implemented as a regular union with annotations specific to Pydantic for that purpose, however it is only still a draft and most likely won't make it until Python 3.9 (if at all).

[1]: https://www.python.org/dev/peps/pep-0593/
[2]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#discriminatorObject
[3]: https://json-schema.org/understanding-json-schema/reference/conditionals.html
[4]: https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types

samuelcolvin:
[...]

PR #1951, opened 2020-09-26
Add strict_union Field feature for non-order affected unions.

aviramha:
## Change Summary

Add strict_union parameter to Field function - defaults to False. 
In case it's set to True and type is of Union, when validating it'll use the first subclass of a type, instead of just trying to validate against all. Unlike the regular validation, it wont accept int -> str or something similar.

## Related issue number

https://github.com/samuelcolvin/pydantic/issues/453

## Checklist

* [Will do after initial approval] Documentation reflects the changes where applicable
* [Will do after initial approval] `changes/<pull request or issue id>-<github username>.md` file added describing change


aviramha:
I'm wondering what to do in case of typing hints inside a Union, for example:
`a: Union[Set[str], List[str]]`
if I do model(a=["123"] I'll get back {"123"}...

samuelcolvin:
I'm not sure about this, I think #619 and #1423 would be better solutions.

PrettyWood:
@samuelcolvin I'm really torn with this issue.
On the one hand I completely agree the whole union behaviour should be improved and the linked issues are better solutions. But on the other hand we got so many issues about unions! (I think it's the most common issue on *pydantic*)
And every time we just answer "yes we know and we'll do something about it one day".
This will probably be tackled in v2, which won't be released anytime soon unfortunately (I hope 2021 🙏)
So I guess having a workaround behind a feature could be a nice spare wheel.

samuelcolvin:
I think we have to wait for at least #619.

samuelcolvin:
it's worth noting that #619 could be added without waiting for v2.

Also we could create a `StrictUnion` or `SmartUnion` which behaved more like #1423 without waiting for v2, then make it the default logic for v2.

PR #1957, opened 2020-09-29
Fix const validator not running when class validators are present

### LINKED ISSUE #1956: BUG: Const validator is not called when there are class validators
The issue is on this line:
https://github.com/samuelcolvin/pydantic/blob/bf9cc4a5e7903ada2e819a63fe2da011f292a143/pydantic/fields.py#L542

Unlike the line above, post_validators are not appended to but overwritten. This removes the const validator...

hmvp:
## Change Summary

Fix issue where const validator was not used

## Related issue number

closes #1956 

## Checklist

### DIFF
--- a/changes/1957-hmvp.md
+++ b/changes/1957-hmvp.md
@@ -1,0 +1,1 @@
+Fix const validators not running when custom validators are present
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -539,7 +539,7 @@ class ModelField(Representation):
 
         if class_validators_:
             self.pre_validators += prep_validators(v.func for v in class_validators_ if not v.each_item and v.pre)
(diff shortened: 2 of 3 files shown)


hmvp:
@samuelcolvin If you find the time, would you consider reviewing and merging this bugfix?

samuelcolvin:
This looks fine to me, but I'm worried there must have been a good reason the `+=` was omitted.

Is there any case you can think of that this could break?

hmvp:
Well it breaks code that relies on const validators not running when custom validators are present which I find odd behaviour to want.. The added test is an example of that. 

BTW in 1.5.x this did work and the added test would not fail (that is why I found the issue in the first place, when upgrading to 1.6.1 and having a test fail)

I suspect the plus was left out because this code moved around a bit and that line use to be the first place the post_validators where filled..

samuelcolvin:
makes sense, thank you.

PR #1972, opened 2020-10-06
fix(validation): also check `root_validators` when `validate_assignment` is on

### LINKED ISSUE #1971: validate_assignment does not work with root_validator
```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /Users/shane/repos/domain-model-specification/Python/.venv/lib/python3.7/site-packages/pydantic
               python version: 3.7.7 (default, Jul 18 2020, 22:28:30)  [Clang 11.0.3 (clang-1103.0.32.62)]
```

# Minimal Example
I'd like to enforce a condition on a whole model during arbitrary changes in its state:
```py

class Example(BaseModel):
    current_value: float
    max_value: float

    class Config:
        validate_assignment = True

    @root_validator
    def current_lessequal_max(cls, values):
        current_value = values.get("current_value")
        max_value = values.get("max_value")
        assert abs(current_value) <= max_value, f"current_value cannot be greater than max_value"
        return values
```

Based on my reading of the docs I'd expect this to fail validation if I set `current_value` above `max_value`, or set `max_value` below `current_value`, on an existing instance. Is that not the intended behavior?

Thanks for an excellent library!

PrettyWood:
## Change Summary
This is a proposal but I agree with @caldwellshane that `root_validator` should be considered as a regular validator. It should hence be triggered when `validate_assignment` is on.
This is not the case at the moment

## Related issue number

closes #1971

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -365,11 +365,32 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         elif not self.__config__.allow_mutation:
             raise TypeError(f'"{self.__class__.__name__}" is immutable and does not support item assignment')
         elif self.__config__.validate_assignment:
+            new_values = {**self.__dict__, name: value}
+
+            for validator in self.__pre_root_validators__:
+                try:
+                    new_values = validator(self.__class__, new_values)
+                except (ValueError, TypeError, AssertionError) as exc:
+                    raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], self.__class__)
+
             known_field = self.__fields__.get(name, None)
             if known_field:
                 value, error_ = known_field.validate(value, self.dict(exclude={name}), loc=name, cls=self.__class__)
                 if error_:
                     raise ValidationError([error_], self.__class__)
+                new_values[name] = value
+
+            errors = []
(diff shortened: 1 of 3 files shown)


pydantic/main.py
```diff
@@ -366,6 +366,16 @@ def __setattr__(self, name, value):
         elif self.__config__.validate_assignment:
             known_field = self.__fields__.get(name, None)
             if known_field:
+                new_values = {**self.__dict__, name: value}
+
+                for skip_on_failure, validator in self.__post_root_validators__:
```
samuelcolvin, line 371: this should be called after validating the field and not all `skip_on_failure` validators should be skipped.

samuelcolvin: so it's more like standard validation.

samuelcolvin (commented):
otherwise I'm good with this.

PrettyWood:
@samuelcolvin You were right sorry I went too fast. I added both pre- and post-rootvalidators. The code is quite similar with `validate_model` but I feel like factorize all this code makes things less explicit so I didn't go for a full DRY approach.

samuelcolvin:
👍 thank you

caldwellshane:
🙏 thank you @PrettyWood and @samuelcolvin

PR #1974, opened 2020-10-07
Add ability to use min_length and max_length constraints with secret types

### LINKED ISSUE #1461: SecretStr min_length, max_length
# Bug

```
             pydantic version: 1.4
            pydantic compiled: True
                 install path: /app/venv/lib/python3.7/site-packages/pydantic
               python version: 3.7.5 (default, Apr 19 2020, 20:18:17)  [GCC 9.2.1 20191008]
                     platform: Linux-5.3.0-51-generic-x86_64-with-Ubuntu-19.10-eoan
```
When the constraints min_length or max_length are set for a field of type SecretStr the erorr appear:
e.g.

```
class SecrTest(BaseModel):
    b: SecretStr = Field(None, max_length= 4)

On field "b" the following field constraints are set but not enforced: max_length. 
For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints
```

uriyyo:
## Change Summary
Add ability to use min_length and max_length constraints with `SecretStr` and `SecretBytes` types.

For instance, this model is totally valid:
```python
class Form(BaseModel):
    username: str
    password: SecretStr = Field(min_length=6, max_length=10)
```

## Related issue number
Fix #1461

## Checklist


pydantic/schema.py
```diff
@@ -831,10 +833,10 @@ def go(type_: Any) -> Type[Any]:
         attrs: Optional[Tuple[str, ...]] = None
         constraint_func: Optional[Callable[..., type]] = None
         if isinstance(type_, type):
-            if issubclass(type_, str) and not issubclass(type_, (EmailStr, AnyUrl, ConstrainedStr)):
+            if issubclass(type_, (str, StrLike)) and not issubclass(type_, (EmailStr, AnyUrl, ConstrainedStr)):
```
PrettyWood, line 836: I'm not really in favor of those two new classes `BytesLike` and `StrLike` because I would expect `EmailStr`, `ConstrainedStr` and so on to inherit from `StrLike`.
I would simply write `if issubclass(type_, (str, SecretStr))` and `elif issubclass(type_, (bytes, SecretBytes))`
and remove the two new classes.
It's just my opinion of course but I reckon it's confusing

uriyyo: I added those classes for the future extending.

For instance, if there will be some new classes which would not be str, but should be validated as str, like `SecretStr` and `SecretBytes`, you can simply inherit from the marker class. 

`EmailStr`, `AnyUrl`, `ConstrainedStr` those classes actually inherits from `str` so they are strings. But `secret` classes they aren't strings but they should be validated as strings, that's why I don't inherit `EmailStr`, `AnyUrl`, `ConstrainedStr` classes from `StrLike`.

@patrickkwang what is your opinion, should I remove `StrLike` and `BytesLike` classes?

PrettyWood: I tend to prefer adding "genericity" later on. Since we only have one class and I'm not convinced we'll add new ones for a while, I would remove them yes. If some day we have something like `issubclass(type_, (str, SecretStr, C1, C2))` then it won't cost much to do what you did here and add a dummy parent class. WDYT ?

uriyyo: I totally agree with you, for this moment it adds more complexity to solve a simple problem.

But in the future, we should definitely think about some "genericity"😊.

I will remove those classes and update PR.
Thanks for a review.

PrettyWood (approved):
LGTM. Thanks a lot!

samuelcolvin (commented):
looks good, please also check that `schema()` is correct.

uriyyo:
@samuelcolvin Looks like I found a bug, correct me in a case when I missed smth.

For instance, I define a custom type which inherits from `str`:
```python
class CustomStr(str):
   ...
```

And I have to models:
```python
class Foo(BaseModel):
    field: CustomStr # without constrain

class Bar(BaseModel):
    field: CustomStr = Field(..., min_length=10) # with constrain
```

Example of usage and results:
```python
foo = Foo(field='foo')
bar = Bar(field='bar')

print(type(foo.field))
print(type(bar.field))
```
```
<class 'CustomStr'>
<class 'str'>
```

That's because of this code:
```python
# schema.py L835
if issubclass(type_, (str, SecretStr)) and not issubclass(type_, (EmailStr, AnyUrl, ConstrainedStr)):
    attrs = ('max_length', 'min_length', 'regex')
    constraint_func = constr
```

So with my current implementation, `SectertStr` will be converted into `str` 😬


I have a future request and want to hear your opinion.

I propose to add `__constraints__` custom attribute for custom-defined classes so the user can specify what kind of constraints supported by the class.

So, for instance, `SecretStr` class will look like:
```python
class SecretStr:
    min_length: OptionalInt = None
    max_length: OptionalInt = None

    __constraints__: Set[str] = {'min_length', 'max_length'}

    @classmethod
        update_not_none(
            field_schema,
            type='string',
            writeOnly=True,
            format='password',
            minLength=cls.min_length,
            maxLength=cls.max_length,
        )
```

With this approach, we can simplify `schema.get_annotation_from_field_info` function and allows us to write more extendable code.

For instance, we can define mixins:
```python
class LengthValidation:
    min_length: OptionalInt = None
    max_length: OptionalInt = None

    __constraints__: Set[str] = {'min_length', 'max_length'}

    @classmethod
    def __modify_schema__(cls: Type[Sized], field_schema: Dict[str, Any]) -> None:
        update_not_none(field_schema, minLength=cls.min_length, maxLength=cls.max_length)

    @classmethod
        yield constr_length_validator

class SecretStr(LengthValidation):
    @classmethod
        super().__modify_schema__(field_schema)
        field_schema.update(type='string', writeOnly=True, format='password')

    def __len__(self) -> int:
        return len(self._secret_value)
```

@samuelcolvin What is your opinion regarding this feature?

Also, I thought about API like this:
```python
class SecretBytes:
    min_length: Constraint[OptionalInt] = None
    max_length: Constraint[OptionalInt] = None
```

I have already updated my PR using this feature.

samuelcolvin:
I think better to modify `schema.py` to add something like

```py
if issubclass(type_, SecretStr) and (field_info.min_items is not None or field_info.max_items is not None):
    type_.min_length = field_info.min_items
    type_.max_length = field_info.max_items
```

Then allow yield the length validator from `__get_validators__` if either of those properties is set.

I think that would require a smaller change?

uriyyo:
@samuelcolvin I have updated PR, thanks for your suggestions.

And what about the bug and feature request which I described above?
Should I create issues regarding them?

tests/test_types.py
```diff
@@ -2124,6 +2124,35 @@ class Foobar(BaseModel):
     assert exc_info.value.errors() == [{'loc': ('password',), 'msg': 'str type expected', 'type': 'type_error.str'}]
 
 
+def test_secretstr_min_max_length():
+    class Foobar(BaseModel):
+        password: SecretStr = Field(min_length=6, max_length=10)
+
+    with pytest.raises(ValidationError) as exc_info:
```
samuelcolvin, line 2131: you also need to test the case that the length is okay, e.g. string of length 8

tests/test_types.py
```diff
@@ -2180,6 +2209,60 @@ class Foobar(BaseModel):
     assert exc_info.value.errors() == [{'loc': ('password',), 'msg': 'byte type expected', 'type': 'type_error.bytes'}]
 
 
+def test_secretbytes_min_max_length():
+    class Foobar(BaseModel):
+        password: SecretBytes = Field(min_length=6, max_length=10)
+
```
samuelcolvin, line 2215: same.

pydantic/schema.py
```diff
@@ -831,7 +833,14 @@ def go(type_: Any) -> Type[Any]:
         attrs: Optional[Tuple[str, ...]] = None
         constraint_func: Optional[Callable[..., type]] = None
         if isinstance(type_, type):
-            if issubclass(type_, str) and not issubclass(type_, (EmailStr, AnyUrl, ConstrainedStr)):
+            if issubclass(type_, (SecretStr, SecretBytes)):
+                attrs = ('max_length', 'min_length')
+
+                def _secret_constraint(**kwargs: Any) -> Type[Any]:
```
samuelcolvin, line 839: can't you just do
```suggestion
                def constraint_func(**kwargs: Any) -> Type[Any]:
```
and remove `constraint_func = _secret_constraint` below?

samuelcolvin (commented):
otherwise this looks good.

samuelcolvin:
> And what about the bug and feature request which I described above?
> Should I create issues regarding them?

if you have a bug or feature request, please create an issue or issues. But in general I think the current approach works pretty well, I don't want to change the interface or add more features until it's absolutely necessary.

uriyyo:
I have updated PR.
I will create an issue regarding the bug which I found.

samuelcolvin:
thanks so much.

transfluxus:
nice, 

I am still getting the original error message with version 1.7:
```
ValueError: On field "password" the following field constraints are set but not enforced: min_length. 
For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints
```

uriyyo:
Hi @transfluxus, I just test it out, and it works as expected.

I have run this code:
```python
from pydantic import BaseModel, Field, SecretStr


class Form(BaseModel):
    password: SecretStr = Field(min_length=6)

form = Form(password="hello world")
```
And I didn't receive any error.

Please, make sure that you are using the latest available version of a `pydantic` by running:
```bash
pip install -U pydantic
```

transfluxus:
hey @uriyyo 
indeed, there was a problem with my ide not updating pydantic to 1.7.

Would be nice if the constructor wouldn't have a warning about expecting a "SecretStr" instead of str.
Since

`password=SecretStr("ceoofddffl")`

actually fails with

```
password
  str type expected (type=type_error.str)
```

but this concerns all Secret Fieldtypes and is another issue.

thanks!

jteppinette:
I am still receiving this error on Pydantic 1.9.0.

PR #1976, opened 2020-10-08
Include tests in source distributions

sbraz:
Hi,
We at Gentoo (and Debian too AFAIK) tend to use PyPI tarballs to run tests. Could you please include them in the next release?

## Change Summary
Include tests in source distributions.

## Checklist

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,6 @@
 include LICENSE
 include README.md
 include HISTORY.md
+graft tests
+global-exclude __pycache__
+global-exclude *.py[cod]
--- a/changes/1976-sbraz.md
+++ b/changes/1976-sbraz.md
@@ -1,0 +1,1 @@
+Include tests in source distributions.


sbraz:
I don't really understand how I could have changed the coverage, is that a bug?

samuelcolvin:
The coverage issue is something to do with codecov, it's using a different directory for windows tests and use showing lines missed in windows as misses, look in coverage. I'll try and fix it.

---

On the actual issue, I don't really wan to include tests in the `.whl` files if I can avoid it. Is there any way to include them in `.tar.gz` but not in `.whl`? Or is that already implemented somehow.

Have you build the binaries and `.tar.gz`and checked this is working? See https://github.com/samuelcolvin/pydantic/pull/1965/commits/55d1305ff633c73fa19d7fb4426a80a1fc0b88bd for an example of how to do this easily.

sbraz:
> On the actual issue, I don't really wan to include tests in the .whl files if I can avoid it. Is there any way to include them in .tar.gz but not in .whl? Or is that already implemented somehow.

I ran `bdist_wheel` and the test directory isn't included. It looks like wheel doesn't really care about `MANIFEST.in`.

> Have you build the binaries and .tar.gzand checked this is working?

Yes, the generated `sdist` does contain the test directory. This syntax in `MANIFEST.in` is more or less boilerplate that I add every times tests aren't included.

samuelcolvin:
looks good to me.

My only question would be around whether requirements from `tests/requirements.txt` are required somehow. Do we need to set `tests_require` in `setup.py`?

sbraz:
I think `tests_require` are only significant if you run `setup.py test` and I doubt many people still do that nowadays.

The file does get added to the sdist (at least on my setup) but I don't think it's a problem. It won't get installed anyway so it won't pollute the system.

samuelcolvin:
> The file does get added to the sdist (at least on my setup) but I don't think it's a problem. It won't get installed anyway so it won't pollute the system.

I realise that, but I thought if you wanted your platform to run tests, you might need a way to define requirements for those tests, hence suggesting `tests_require`. As per the following line, all that's actually required to run tests is pytest and pytest-mock

https://github.com/samuelcolvin/pydantic/blob/078e7090d01d81cbcb4dbc662624bd9a33b38a43/.github/workflows/ci.yml#L227

If you're happy with this I'll merge it.

samuelcolvin:
thanks

sbraz:
I don't think pytest-mock is even required, tests pass without it :)

Issue #1578, opened 2020-05-29
BaseModel.schema() fails when using generic types


adrianschneider94:
# Bug
## Version
Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
  pydantic version: 1.5.1
            pydantic compiled: True
                 install path: C:\Users\adrian\miniconda3\envs\sangl-butler\Lib\site-packages\pydantic
               python version: 3.8.1 (default, Mar  2 2020, 13:06:26) [MSC v.1916 64 bit (AMD64)]
                     platform: Windows-10-10.0.18362-SP0
     optional deps. installed: ['typing-extensions', 'email-validator']

```
## Description
Hi there!

I want to write a generic type, that behaves like the type it extends, but injects fields into the model with ``__modify_schema__``. So my approach was to use a generic class to use it like that:

```py
from pydantic import BaseModel

class MyModel(BaseModel):
   my_field: FieldThatModifiesModel[str] = "Default value"

````
The field should completely behave like the type var.

However, when I use a generic type in my model, I can't retrieve the model anymore.
See the code below (taken from the docs):
## Example
```py
from pydantic import BaseModel, ValidationError
from pydantic.fields import ModelField
from typing import TypeVar, Generic

AgedType = TypeVar('AgedType')
QualityType = TypeVar('QualityType')

# This is not a pydantic model, it's an arbitrary generic class
class TastingModel(Generic[AgedType, QualityType]):
    def __init__(self, name: str, aged: AgedType, quality: QualityType):
        self.name = name
        self.aged = aged
        self.quality = quality

    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    # You don't need to add the "ModelField", but it will help your
    # editor give you completion and catch errors
    def validate(cls, v, field: ModelField):
        if not isinstance(v, cls):
            # The value is not even a TastingModel
            raise TypeError('Invalid value')
        if not field.sub_fields:
            # Generic parameters were not provided so we don't try to validate
            # them and just return the value as is
            return v
        aged_f = field.sub_fields[0]
        quality_f = field.sub_fields[1]
        errors = []
        # Here we don't need the validated value, but we want the errors
        valid_value, error = aged_f.validate(v.aged, {}, loc='aged')
        if error:
            errors.append(error)
        # Here we don't need the validated value, but we want the errors
        valid_value, error = quality_f.validate(v.quality, {}, loc='quality')
        if error:
            errors.append(error)
[...]

PR #1981, opened 2020-10-09
Force fields.Undefined to be a singleton object

daviskirk:
## Change Summary

In various places of the code, we compare directly to `fields.Undefined`
since we assume it to be constant.
When new models get created however, the object is deepcopied and
is no longer identical with the original object.
We therefore add `__copy__` and `__deepcopy__` methods to ensure
that the copied objects are actually the same original object.

## Related issue number

While I did does not address the exact error message #1578 had, schema generation on inherited generics that had required generic fields did not work at all,  so it might also fix this.

## Checklist

### DIFF
--- a/changes/1981-daviskirk.md
+++ b/changes/1981-daviskirk.md
@@ -1,0 +1,1 @@
+Force `fields.Undefined` to be a singleton object, fixing inherited generic model schemas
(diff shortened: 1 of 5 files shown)


samuelcolvin:
great, thank you.

PR #1986, opened 2020-10-09
BaseModel dict serialize kwarg

### LINKED ISSUE #1409: JSON-like serialization on .dict()
# Feature Request

pydantic version: 1.5
pydantic compiled: True
install path: C:\Work\repos\data-purge\.venv\Lib\site-packages\pydantic
python version: 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 23:11:46) [MSC v.1916 64 bit (AMD64)]
optional deps. installed: []

I'm using a framework (lambda) that expects the lambdas to return a simple dict to be passed along to another step.  This is done through marshaling that I don't really have a lot of access to modifying.  I'm using pydantic models for all of my de/serialization, and it had been working great to just say something to the effect of `return result.dict()`

However, now that I've added a datetime field to the model, I get a failure on: `Unable to marshal response: Object of type datetime is not JSON serializable`

Now, I know this is a pretty standard python issue with anything Json related.  But given that pydantic is such a great library for managing all this stuff, and it's not uncommon to have _another_ library expect to receive dicts, not json strings, I think it would be nice if we could have some way for `.dict()` to return a serializable format by default.

Essentially, I want to still have my pydantic model define the fields as a datetime, but I want the `.dict` step to convert those into a string format, rather than the datetime objects themselves.  Is this at all possible/reasonable?

```py

In [2]: from datetime import datetime

In [3]: class Thing(BaseModel):
   ...:     x: datetime
   ...: 
In [5]: a = Thing(x="2020-01-01T00:00:00")

In [6]: a.dict()
Out[6]: {'x': datetime.datetime(2020, 1, 1, 0, 0)}

# preferred output:
Out[6]: {"x": "2020-01-01T00:00:00"}
...
```

Is this already possible?  Could this be added in some kind of way?  This isn't the first time I've hit this scenario, and I can't find an elegant way around it.  If only python actually was reasonable about datetimes, I wouldn't expect pydantic or something else to solve it :/

wozniakty:
## Change Summary

Added serialize kwarg to BaseModel.dict to create an immediately serializable dict format with configurable type conversion.

## Related issue number

Closes #1409 

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -383,8 +383,8 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
     def dict(
         self,
         *,
-        include: Union['AbstractSetIntStr', 'MappingIntStrAny'] = None,
-        exclude: Union['AbstractSetIntStr', 'MappingIntStrAny'] = None,
+        include: Optional[Union['AbstractSetIntStr', 'MappingIntStrAny']] = None,
+        exclude: Optional[Union['AbstractSetIntStr', 'MappingIntStrAny']] = None,
         by_alias: bool = False,
         skip_defaults: bool = None,
         exclude_unset: bool = False,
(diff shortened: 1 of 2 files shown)


wozniakty:
@samuelcolvin wanted to throw this implementation out there for feedback.  This has the structure you asked for in https://github.com/samuelcolvin/pydantic/issues/1409#issuecomment-650116116

Some things of note:
* The default serializer for True has been set to `self.__json_encoder__` so that it both uses `pydantic_encoder` by default, but also stays in sync with any custom encoding defined on the BaseModel
* The structure of a Bool | Callable as a kwarg creates for some wacky ternary arguments necessary to handle appropriate type behaviors

samuelcolvin:
I'll work on this when I have a chance.

There's a lot more to do than a few arguments to `dict()` etc. I think we need to significantly modify `json_encoder`;

Issue #1913, opened 2020-09-08
🙏 Sorry, I'm really busy at the moment


samuelcolvin:
Hi, you might have noticed that support on pydantic and particularly reviewing and merging pull requests is a bit slow at the moment. 

@PrettyWood and others have been amazing in their support recently, but I have been noticeably absent, and I know that slows things down or stops them.

**I'm really sorry about that**, I'm currently flat out with work at my new role with Nell Health and have reduced time to spend on pydantic.

I'm still committed to pydantic, I use it most days, I rely on it in production with multiple projects. I want the new features currently languishing in PRs and issue descriptions as much as you. I also have big ideas about v2 and how we can make pydantic both faster and easier to use at the same time.

**I'll do my best to review all open pull requests and issues over the month**.

## You can help!

If you're waiting on new features, please consider helping out with the pydantic project, you could:
* check open issues you've created, check if they're still relevant and either add more information or close them
* look through recent open issues and see if you can solve their problem, give feedback on a feature request or confirm a bug
* review, try and give feedback on open pull requests

xkortex:
I **really** appreciate this library. It has simplified an absolute ton of workflows for me. I know what it's like to be absolutely slammed at work. Try your best to delegate, maintain balance, and most of all....Keep up the awesome work!

ssbarnea:
That is wonderful library and letting it be affected by bit rot it would be a big loss for python community.

How about moving it under a github organization that can assure its future does not depend on a single person? It could be either one of the existing ones focused on python or a new one. The key here is to find few others that you can trust as not doing anything bad and give them access, so they can help when needed.

That is why none of the libraries I created of maintained with the goal of sharing them are under my personal github account, as I did not want to eventually become the bottleneck.

ssbarnea:
I did not want to imply that pydantic would be into a bad place. I only wanted to do a risk-mitigation measures. Do you have a designated successor configured on https://github.com/settings/admin ? -- that was the kind of issue I wanted to highlight: assuring that everything is set in place in case something bad happens with you, the current pandemy made us more aware of risks.

There is nothing stopping you from having an organization solely for pydantic if you want, they do not cost money.

[...]

PR #1989, opened 2020-10-10
Fix #947, #1483, #1247: allow inner type vars to be present in parent generic classes

### LINKED ISSUE #947: GenericModel does not support generic subclasses
# Bug

It seems the implementation of pydantic.generics.GenericModel does not support generic subclasses. Example:
```
import pydantic.generics
import typing

T = typing.TypeVar("T")

class ResponseModel(pydantic.generics.GenericModel, typing.Generic[T]):
    data: List[T]

class PaginatedResponseModel(ResponseModel[T]):
    total_count: int

ConcreteModel = PaginatedReponseModel[dict] # TypeError: Cannot parameterize a concrete instantiation of a generic model

```
`PaginatedReponseModel` is erroneously considered "concrete" even though type variables are used as parameters, and not concrete types.
It seems GenericModel.__class_getitem__ does not check if any parameters are `TypeVar`, like `typing.Generic` does. 

To support generic subclasses, `__class_getitem__` should check if all of the parameters are fully concrete type(i.e. any of them are `typing.TypeVar`s, or any of them are generic types that have non-concrete type parameters). If not, a concrete model should not be created.

I would consider this a bug since it does not follow the behavior of other `typing.Generic` usages, but it could also be a feature request.

Please complete:
* OS: Ubuntu 18.04.1
* Python version `import sys; print(sys.version)`: 3.7.1 (default, Oct 22 2018, 11:21:55) 
[GCC 8.2.0]
* Pydantic version `import pydantic; print(pydantic.VERSION)`: `1.0`

### LINKED ISSUE #1247: GenericModel - inheritance and Optional[TypeX] - fail to instantiate a subclass
# Bug

```
             pydantic version: 1.4
            pydantic compiled: False
                 install path: /Users/thaisdeboisfosse/envs/thenv/lib/python3.7/site-packages/pydantic
               python version: 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)  [Clang 6.0 (clang-600.0.57)]
                     platform: Darwin-19.3.0-x86_64-i386-64bit
```
### Summary

Thank you for the lib ! I found a weird bug while using GenericModel.

I create a BaseClass that inherits GenericModel and uses Generic. It contains a property X of type Optional[TypeX].

I create a subclass that fixes the Generic at runtime. The bug is that I cannot define the type of the Generic without an error.

If `X` is of type `TypeX` instead of `Optional[TypeX]` then there is no error (it is what I ended up doing).

### Code to reproduce the error

```py
import pydantic

TypeX = TypeVar('TypeX')

class BaseClass(GenericModel, Generic[TypeX]):
    X: Optional[TypeX]  # This causes the error

class ChildClass(BaseClass[TypeX], Generic[TypeX]):
    pass

ChildClass[int]
```

Error: 
```
cls = <class 'test_base_format.test_pydantic_failes.<locals>.ChildClass'>
params = <class 'int'>

    def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[Type[Any], ...]]) -> Type[Any]:
        cached = _generic_types_cache.get((cls, params))
        if cached is not None:
            return cached
        if cls.__concrete__:
>           raise TypeError('Cannot parameterize a concrete instantiation of a generic model')
E           TypeError: Cannot parameterize a concrete instantiation of a generic model
```

### LINKED ISSUE #1483: Generic type is not passed into items of a List
# Bug

```
pydantic version: 1.5.1
pydantic compiled: True
install path: /home/.../.venv/lib/python3.8/site-packages/pydantic
python version: 3.8.2 (default, Apr 22 2020, 13:26:47)  [GCC 7.5.0]
platform: Linux-5.3.0-45-generic-x86_64-with-glibc2.27
```

Using a modified example from the documentation on [Generic Models](https://pydantic-docs.helpmanual.io/usage/models/#generic-models): 

```py
from typing import Generic, TypeVar, List

T = TypeVar('T')

class InnerT(GenericModel, Generic[T]):
    inner: T

class OuterT(GenericModel, Generic[T]):
    outer: T
    nested: List[InnerT[T]]

print(OuterT[int](**{"outer": 1, "nested": [{"inner": "2"}]}))  # Output 1
print(OuterT[int](**{"outer": 1, "nested": [{"inner": "a"}]}))  # Output 2
```

### Expected output

1. `outer=1 nested=[InnerT[int](inner='2')]`
2. Validation error

### Actual output

1. `outer=1 nested=[InnerT[T](inner='2')]`
2. `outer=1 nested=[InnerT[T](inner='a')]`

### LINKED ISSUE #2191: Optional Generic nested field with a bound on a parent class does not resolve correctly
### Checks

# Bug

```
pydantic version: 1.6.1
            pydantic compiled: True
                 install path: /Users/antoniogusmao/.pyenv/versions/3.8.5/envs/cloud-app-server-sync-poc/lib/python3.8/site-packages/pydantic
               python version: 3.8.5 (default, Sep  8 2020, 12:09:39)  [Clang 11.0.3 (clang-1103.0.32.59)]
                     platform: macOS-10.15.6-x86_64-i386-64bit
     optional deps. installed: []
```

See the prints at the end of the code snippet for expected output and actual output.
`SuperEnvelopeB` does not resolve correctly and removes the fields specific to `MyDocument`. 
`SuperEnvelopeA` and `SuperEnvelopeC` work as expected.

```py
from pprint import pprint

from typing import Generic, Optional, TypeVar


class Document(BaseModel):
    id: str


class MyDocument(Document):
    my_var: str


A = TypeVar("A")


class DocEnvelopeA(GenericModel, Generic[A]):
    doc: Optional[A]


class SuperEnvelopeA(GenericModel, Generic[A]):
    wrapper: Optional[DocEnvelopeA[A]]


B = TypeVar("B", bound=Document)


class DocEnvelopeB(GenericModel, Generic[B]):
    doc: Optional[B]


class SuperEnvelopeB(GenericModel, Generic[B]):
    wrapper: Optional[DocEnvelopeB[B]]


C = TypeVar("C", bound=Document)


class DocEnvelopeC(GenericModel, Generic[C]):
    doc: C  # we removed the Optional


class SuperEnvelopeC(GenericModel, Generic[C]):
    wrapper: DocEnvelopeC[C]  # we removed the Optional


x = {
    'wrapper': {
        'doc': {
            'id': '1',
            'my_var': '2'
        }
    }
}

pprint(x)  # {'wrapper': {'doc': {'id': '1', 'my_var': '2'}}}
pprint(SuperEnvelopeA[MyDocument](**x).dict())  # {'wrapper': {'doc': {'id': '1', 'my_var': '2'}}}
pprint(SuperEnvelopeB[MyDocument](**x).dict())  # {'wrapper': {'doc': {'id': '1'}}}   -> expected my_var to be there
pprint(SuperEnvelopeC[MyDocument](**x).dict())  # {'wrapper': {'doc': {'id': '1', 'my_var': '2'}}}

```

### LINKED ISSUE #2227: Model Validation Issue for Nested, Optional, Generic Model
### Checks

# Bug

The code below outputs:

```
Caught A
Missed B
```

Although it should catch both validations. Generic model which is marked as optional seems to fail type checks, and produces "`Any`" type instead.

```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: /opt/virtualenvs/python3/lib/python3.8/site-packages/pydantic
               python version: 3.8.7 (default, Dec 21 2020, 20:10:35)  [GCC 7.5.0]
                     platform: Linux-5.4.0-1019-gcp-x86_64-with-glibc2.27
```

```py
from typing import TypeVar, Generic, Optional, List


DataT = TypeVar('DataT')


class Data(BaseModel):
  foo: str


# --- WORKS ---

class APIContentS(GenericModel, Generic[DataT]):
    """Second-level for the API response. i.e. the content"""
    data: DataT
    message: Optional[str] = None


class APIResponseS(GenericModel, Generic[DataT]):
    """Base API response model"""
    payload: APIContentS[DataT]
    is_successful: bool = True


try:
  model = APIResponseS[List[Data]](
    payload=APIContentS(
      data=[{'bar': True}]
    )
  )
  print("Missed A")
except:
  print("Caught A")


# --- FAILS ---

class APIContent(GenericModel, Generic[DataT]):
    """Second-level for the API response. i.e. the content"""
    data: Optional[DataT]
    message: Optional[str] = None


class APIResponse(GenericModel, Generic[DataT]):
    """Base API response model"""
    payload: APIContent[DataT]
    is_successful: bool = True


try:
  model = APIResponse[List[Data]](
    payload=APIContent(
      data=[{'bar': True}]
    )
  )
  print("Missed B")
except:
  print("Caught B")
```

daviskirk:
## Change Summary

Fix bugs where generic models with fields where the typevar is nested in another type `a: List[T]` are considered to be concrete. This allows these models to be subclassed as expected.

## Related issue number

Fixes #947, Fixes #1483, Fixes #1247
fix #2191
fix #2227

## Checklist


pydantic/generics.py
```diff
@@ -113,4 +123,34 @@ def _parameterize_generic_field(field_type: Type[Any], typevars_map: Dict[TypeVa
 
 
 def _is_typevar(v: Any) -> bool:
```
choogeboom, line 125: I propose changing the name of this function to something better, since it's no longer doing a simple check to see if something is a type variable. Maybe something like `_is_generic` or `_has_typevar`?

choogeboom (commented):
It's funny, I worked on this last Friday, but you beat me to the punch. I like your solution better than mine, but I think it's worth adding some test cases like the following:

https://github.com/wayfair-contribs/pydantic/blob/d2a1a5866283acf350338b1d33719f8b20598f5d/tests/test_generics.py#L204

https://github.com/wayfair-contribs/pydantic/blob/d2a1a5866283acf350338b1d33719f8b20598f5d/tests/test_generics.py#L220

pydantic/generics.py
daviskirk: Good idea, I renamed it to `_has_typevar` as suggested

daviskirk:
> It's funny, I worked on this last Friday, but you beat me to the punch. I like your solution better than mine, but I think it's worth adding some test cases like the following:
> 
> https://github.com/wayfair-contribs/pydantic/blob/d2a1a5866283acf350338b1d33719f8b20598f5d/tests/test_generics.py#L204
> 
> https://github.com/wayfair-contribs/pydantic/blob/d2a1a5866283acf350338b1d33719f8b20598f5d/tests/test_generics.py#L220

Well I'm glad I'm not alone on the generics code, there are some tricky problems and I'm really glad there are more eyes on this than just mine ;).  
I added the tests (as well as unifying the naming as I also realized there was already a test for subclasses just not testing the broken cases).  

However, I did change the "implicit parameters" test to not use implicit parameters as I think this would be more of a "new feature" (https://pydantic-docs.helpmanual.io/usage/models/#generic-models states that one has to inherit from typing.Generic in subclasses). 

I like the idea of implicit generic model inheritance but I think it should probably be discussed/handled in a separate pull request (I think your `_generate_class_parameters_from_bases` will probably work fine for that)

choogeboom (approved):
Changes look good! I agree that the implicit type hints is a new feature. I was delving into the mypy documentation on generics and forgot to cross reference the pydantic docs, so I missed the note about requiring all generic models to explicitly subclass `Generic`. I'll make an issue for it.

Not sure what the process is for getting a PR approved and merged, but #1913 mentions that @samuelcolvin is looking for people to review PRs. IMO this looks ready, and I hope it goes out quick!

choogeboom:
I created issue #2002 for the implicit generic parameters.

choogeboom:
Hmm... I found an edge case. Working to isolate it. I'll let you know once I figure it out.

pydantic/generics.py
```diff
@@ -79,9 +84,14 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
 def resolve_type_hint(type_: Any, typevars_map: Dict[Any, Any]) -> Type[Any]:
-    if hasattr(type_, '__origin__') and getattr(type_, '__parameters__', None):
-        concrete_type_args = tuple([typevars_map[x] for x in type_.__parameters__])
-        return type_[concrete_type_args]
+    type_args = _get_args(type_)
+    if type_args:
+        concrete_type_args = tuple(resolve_type_hint(arg, typevars_map) for arg in type_args)
+        origin_type = _get_origin(type_) or type_
+        origin_type = _builtin_type_map.get(origin_type, origin_type)
+        return origin_type[concrete_type_args]
```
choogeboom, line 92: If `type_` is `Type[T]`, then `origin_type=type`. Unfortunately `type` is not subscriptable, so this will fail.
```suggestion
        if origin_type is type:
            return type_[concrete_type_args]
        else:
            return origin_type[concrete_type_args]
```
I'm not sure if this is a problem for any other types, but here's a test case you can use for `Type[T]`:

```
@skip_36
def test_generic_with_generic_type():
    T = TypeVar('T')

    class ModelWithType(GenericModel, Generic[T]):
        some_type: Type[T]

    class ReferenceModel(GenericModel, Generic[T]):
        abstract_base_with_type: ModelWithType[T]

    ReferenceModel[int]
```

choogeboom (changes requested):

pydantic/generics.py
daviskirk: Thanks! Indeed there are other cases so my "_builtin_type_map" was way to specific.
I added the test and another variant.  python 3.9 will fix this with https://www.python.org/dev/peps/pep-0585/ but I worked around it for python < 3.9 / >= 3.6.

choogeboom (approved):
Nice

pydantic/generics.py
```diff
@@ -79,9 +83,28 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+                origin_type = getattr(typing, type_._name)
+                return origin_type[concrete_type_args]
+            except (TypeError, AttributeError):
+                pass
+        # this should not be reached
+        raise TypeError(f'Could not resolve type origin: {type_}')  # pragma: no cover
+
+    if lenient_issubclass(type_, GenericModel) and not type_.__concrete__:
+        return type_[tuple(resolve_type_hint(t, typevars_map) for t in type_.__parameters__)]
     return typevars_map.get(type_, type_)
```
choogeboom, line 108: Found another edge case, related to `Callable`. It turns out tat that the "input arguments" part of the type is a list of types, which means that `resolve_type_hints` fails. Here's a test case:

```py
@skip_36
def test_generic_with_callable():
    T = TypeVar('T')
    U = TypeVar('U')

    class VeryInnerModel(GenericModel, Generic[T, U]):
        t: T
        u: U
        some_callable: Callable[[Optional[int], str], None]

    VeryInnerModel[str, U]
```

This results in the following failure: `TypeError: unhashable type: 'list'` when `[Optional[int], str]` is used as a key in the `typevars_map`

choogeboom: Turns out there are a number of problems with `Callable`:
1. Need to handle it in `resolve_type_hint` properly
2. Need to handle it in `_has_typevar` properly

choogeboom:
While I was investigating the `Callable` problem above, I came across another failing case

```
@skip_36
def test_nested_with_partial():
    T = TypeVar('T')
    U = TypeVar('U')

    class Model(GenericModel, Generic[T, U]):
        t: List[T]
        u: List[U]

    alias = Model[int, U]
    assert alias.__concrete__ is False
    concrete = alias[int]
    assert concrete.__concrete__ is True
```

Basically, what's happening is that we're not assigning `__parameters__` properly. We should be extracting all the parameters, but instead we're just passing the outer type, so for example in the above case, `alias` has `List[U]` as a parameter instead of `U`.

pydantic/generics.py
```diff
@@ -79,9 +83,28 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+                # is a builtin type, see https://www.python.org/dev/peps/pep-0585
+                origin_type = getattr(typing, type_._name)
+                return origin_type[concrete_type_args]
+            except (TypeError, AttributeError):
+                pass
+        # this should not be reached
+        raise TypeError(f'Could not resolve type origin: {type_}')  # pragma: no cover
+
+    if lenient_issubclass(type_, GenericModel) and not type_.__concrete__:
+        return type_[tuple(resolve_type_hint(t, typevars_map) for t in type_.__parameters__)]
```
choogeboom, line 107:
```suggestion
        return type_[tuple(resolve_type_hint(t, typevars_map) for t in type_.__parameters__)]
    if isinstance(type_, list):
        return [resolve_type_hints(element, typevars_map) for element in type_]        
```

pydantic/generics.py
```diff
@@ -112,5 +135,35 @@ def _parameterize_generic_field(field_type: Type[Any], typevars_map: Dict[TypeVa
     return field_type
 
 
-def _is_typevar(v: Any) -> bool:
+def _has_typevar(v: Any) -> bool:
+    if lenient_issubclass(v, GenericModel):
+        return not v.__concrete__
+    args = _get_args(v)
```
choogeboom, line 141:
```suggestion
    if isinstance(v, list):
        return any(_has_typevar(element) for element in v)
    args = _get_args(v)
```

pydantic/generics.py
```diff
@@ -57,10 +61,10 @@ def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[T
             ),
         )
         created_model.Config = cls.Config
-        concrete = all(not _is_typevar(v) for v in concrete_type_hints.values())
+        concrete = all(not _has_typevar(v) for v in concrete_type_hints.values())
         created_model.__concrete__ = concrete
         if not concrete:
-            parameters = tuple(v for v in concrete_type_hints.values() if _is_typevar(v))
+            parameters = tuple(v for v in concrete_type_hints.values() if _has_typevar(v))
```
choogeboom, line 67: We need to extract the parameters from the concrete type hints rather than simply passing them along. Otherwise we end up with things like `Optional[T]` as a parameter.

choogeboom (changes requested):

pydantic/generics.py
```diff
@@ -57,10 +61,10 @@ def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[T
...
             parameters = tuple({k: None for k in parameters}.keys())  # get unique params while maintaining order
```
choogeboom, line 68:
```suggestion
            parameters = tuple({v: None for v in typevars_map.values() if isinstance(v, TypeVar)})
```
Turns out that we were complicating this step more than necessary. We don't need `_has_typevar` at all.

pydantic/generics.py
choogeboom: Nevermind. This is wrong.

pydantic/generics.py
choogeboom: Argh... Turns out I was wrong, This fails if one of the params is generic.

choogeboom: OK, I worked it out
```suggestion
            parameters = extract_parameters(typevars_map)
```
Where
```python
def extract_parameters(typevars_map: Dict[Any, Any]) -> Tuple[TypeVarType, ...]:
    extracted_parameters = {param: None for param in generate_parameters(list(typevars_map.values()))}
    return tuple(extracted_parameters)


def generate_parameters(v: Any) -> Iterator[TypeVarType]:
    if isinstance(v, TypeVar):
        yield v
    elif lenient_issubclass(v, GenericModel):
        yield from v.__parameters__
    elif isinstance(v, list):
        for var in v:
            yield from generate_parameters(var)
    else:
        args = _get_args(v)
        for arg in args:
            yield from generate_parameters(arg)
```

daviskirk:
Alright, great detective work!  I basically just copied your suggestion with small modification for mypy and linting.
I also removed _has_typevar and just used _generate_parameters everywhere since they were almost the same and added the tests: https://github.com/samuelcolvin/pydantic/pull/1989/commits/1b5d0f611f1992acc5d8e2722721faa67670068a

pydantic/generics.py
```diff
@@ -61,12 +75,10 @@ def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[T
...
         )
         created_model.Config = cls.Config
-        concrete = all(not _has_typevar(v) for v in concrete_type_hints.values())
+        concrete = not any(_generate_parameters(concrete_type_hints.values()))
         created_model.__concrete__ = concrete
         if not concrete:
-            parameters = tuple(v for v in concrete_type_hints.values() if _has_typevar(v))
-            parameters = tuple({k: None for k in parameters}.keys())  # get unique params while maintaining order
-            created_model.__parameters__ = parameters
+            created_model.__parameters__ = _extract_parameters(typevars_map)
```
choogeboom: Minor optimization:

```
new_params = _extract_parameters(typevars_map)
created_model.__concrete__ = bool(new_parameters)
created_model.__parameters__ = new_params
```

choogeboom (approved):

pydantic/generics.py
daviskirk: correct, also removed extract parameters as it is now only used once and moved a test to it's counterpart that already existed

daviskirk:
rebased on master

pydantic/generics.py
```diff
@@ -99,15 +107,48 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
         """
         This method can be overridden to achieve a custom naming scheme for GenericModels
         """
-        param_names = [param.__name__ if hasattr(param, '__name__') else str(param) for param in params]
+        try:
+            GenericAlias = typing.GenericAlias  # type: ignore
+        except AttributeError:
+            GenericAlias = ()  # for all python < 3.9
+        param_names = [
```
samuelcolvin, line 114: there's a lot going on in this comprehension. Might be more readable to pull it out into a boring old.

```py
param_names = []
for param in params:
    if ...:
        param_names.append(...)
```

pydantic/generics.py
```diff
@@ -99,15 +107,48 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
         """
         This method can be overridden to achieve a custom naming scheme for GenericModels
         """
-        param_names = [param.__name__ if hasattr(param, '__name__') else str(param) for param in params]
+        try:
```
samuelcolvin, line 110: is there a reason this needs to be in this function? Couldn't it be moved out to module level.

pydantic/generics.py
samuelcolvin: also please add comments explaining what's going on here.

pydantic/generics.py
```diff
@@ -99,15 +107,48 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+        ]
         params_component = ', '.join(param_names)
         return f'{cls.__name__}[{params_component}]'
 
 
-def resolve_type_hint(type_: Any, typevars_map: Dict[Any, Any]) -> Type[Any]:
-    if get_origin(type_) and getattr(type_, '__parameters__', None):
-        concrete_type_args = tuple([typevars_map[x] for x in type_.__parameters__])
-        return type_[concrete_type_args]
+def is_identity_typevars_map(typevars_map: Dict[TypeVarType, Type[Any]]) -> bool:
```
samuelcolvin, line 122: docstring please.

pydantic/generics.py
```diff
@@ -99,15 +107,48 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+    return bool(typevars_map) and all(k is v for k, v in typevars_map.items())
+
+
+def resolve_type_hint(type_: Any, typevars_map: Dict[Any, Any]) -> Any:
```
samuelcolvin, line 126: docstring please.

Also are you sure this doesn't overlap with any of the methods in `pydantic.typing`? Could it be moved into `pydantic.typing`? Or doe it need to be here.

pydantic/generics.py
```diff
@@ -119,6 +160,20 @@ def check_parameters_count(cls: Type[GenericModel], parameters: Tuple[Any, ...])
         raise TypeError(f'Too {description} parameters for {cls.__name__}; actual {actual}, expected {expected}')
 
 
+def iter_contained_typevars(v: Any) -> Iterator[TypeVarType]:
+    if isinstance(v, cast(type, TypeVar)):  # mypy < 0.790 needs this cast
```
samuelcolvin, line 164: we're now using `mypy` 0.790, so can this be removed?

Any docstring please.

pydantic/generics.py
```diff
@@ -132,16 +187,16 @@ def _build_generic_fields(
 
 
 def _parameterize_generic_field(field_type: Type[Any], typevars_map: Dict[TypeVarType, Type[Any]]) -> Type[Any]:
-    if lenient_issubclass(field_type, GenericModel) and not field_type.__concrete__:
+    try:
+        is_model = isinstance(field_type, GenericModel) and not field_type.__concrete__
```
samuelcolvin, line 191: I'm not clear why this changed is required?

pydantic/typing.py
```diff
@@ -80,8 +81,21 @@ def evaluate_forwardref(type_: ForwardRef, globalns: Any, localns: Any) -> Any:
...
-        return getattr(t, '__args__', ())
+    if sys.version_info < (3, 7):
+
+        def get_args(t: Type[Any]) -> Tuple[Any, ...]:
+            return getattr(t, '__args__', ())
+
+    else:
+        _GenericAlias = getattr(typing, '_GenericAlias')
+
+        def get_args(t: Type[Any]) -> Tuple[Any, ...]:
```
samuelcolvin, line 92: please rebase to master and check this is still required.

tests/test_generics.py
```diff
@@ -1,14 +1,21 @@
...
 import pytest
 
 from pydantic import BaseModel, Field, ValidationError, root_validator, validator
-from pydantic.generics import GenericModel, _generic_types_cache, get_caller_module_name
+from pydantic.generics import GenericModel, _generic_types_cache, get_caller_module_name, iter_contained_typevars
 
 skip_36 = pytest.mark.skipif(sys.version_info < (3, 7), reason='generics only supported for python 3.7 and above')
 
+if sys.version_info >= (3, 9):
+    CompatList = list
```
samuelcolvin, line 13: this shouldn't be necessary any more.

samuelcolvin (commented):
I know pydantic isn't that well commented (and docstringed (word?)) at the moment. But for new complex stuff like this, let's try and add docstrings wherever possible.

I really don't understand all of generics off the top of my head and @dmontagu who wrote most of it isn't contributing to pydantic at the moment.

pydantic/generics.py
daviskirk: Unfortunately `lenient_issubclass` fails if `field_type` is a generic alias.

```python
lenient_issubclass(list[int], GenericModel)
```
```
  File "/home/dek/Documents/Code/pydantic/pydantic/utils.py", line 145, in lenient_issubclass
    return isinstance(cls, type) and issubclass(cls, class_or_tuple)
  File "/home/dek/anaconda/envs/pydantic-39/lib/python3.9/abc.py", line 102, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
```

The exact logic is that:
```
isinstance(list[int], type)
```
returns `True` but then `issubclass` needs a `class` which `list[int]` is not.

I can add workaround code to lenient_issubclass but I wasn't sure that would be the right thing to do given how many places lenient_issubclass is used and this sort of handling of typehints seemed relatively specialized towards generics.

pydantic/typing.py
daviskirk: Yes, this is already rebased.  Unfortunately, typing changes a lot in between 3.6 and 3.9.
For python `< 3.8`, we need to vendor `get_args` and do it correctly https://github.com/python/cpython/blob/4f3c25043d651a13c41cffcee703f7d5cb677cc7/Lib/typing.py#L1497 (mainly for all the generics stuff), but to do this correctly we need `_GenericAlias` which doesn't exist in python 3.6.
Since generics don't support 3.6 anyway that's ok for us (we just use the simpler version of `return getattr(t, '__args__', ())`, and for python >= 3.8 we can just use the cpython version. 
So unfortunately we need all three cases.

I'll add comments though explaining this

pydantic/generics.py
daviskirk: I moved this to a more generic "all_identical" function (moved that to utils) and added some docs/examples/tests.  Hope it's clearer now.

pydantic/generics.py
daviskirk: I looked through `pydantic.typing` again and there's nothing in there that does this.  

The code is quite geared towards the `generics` module (i.e. I don't see anywhere else in pydantic where this would be used), but the code itself is relatively generic in the sense that it just recursively replaces types according to a a map/dictionary.

It does have to reference the `GenericModel` though so I think I would keep it in there until there is somewhere else that needs this sort of recursive type map.

I also though there may have been confusion regarding the `resolve_annotations` in the `pydantic.typing` module, so I renamed this to `replace_types` since that's what it does (not really specific to type hints)

pydantic/generics.py
daviskirk: Better?  There's also a test specific to this function.  If you'd like I can add an example in the docstring, I'm unsure about the policy of docstring examples vs. tests (I don't think doctests are being run in ci if I'm not mistaken)

tests/test_generics.py
daviskirk: Very true, took it out, thanks!

pydantic/generics.py
daviskirk: I used `pydantic.typing.display_as_type` and added a case to that instead, which cleans this whole function up quite a lot.  Hope it's clearer now.

pydantic/generics.py
daviskirk: See below, got rid of the weird stuff with `pydantic.typing.display_as_type`

daviskirk:
> I know pydantic isn't that well commented (and docstringed (word?)) at the moment. But for new complex stuff like this, let's try and add docstrings wherever possible.
> 
> I really don't understand all of generics off the top of my head and @dmontagu who wrote most of it isn't contributing to pydantic at the moment.

I tried to update a few of the docstrings and added some comments to the more complicated steps of the "creation" part of the generics model.
Let me know if you want more or less.

samuelcolvin:
Sorry for the delay.

I think this is looking good, though I will admit that some of the nuance is lost on me.

Please could you fix coverage, otherwise happy to merge.

daviskirk:
> I think this is looking good, though I will admit that some of the nuance is lost on me.

Lots of the trickier parts are dealing with the changes (across python versions) and edge cases of the python typing module. If there are more issues like this that I don't run into myself or see in the bug tracker feel free to ping me now that I'm a bit familiar with the code... 
Unfortunately there will probably be stuff related to the generic aliases like (`list[int], `dict[str, int]` and so on) and in in 3.10 features (like the union operator `int | str`) that will also have some sort of edge cases that aren't handled yet (not complaining, both are really awesome features for writing code, but handling them in libraries like pydantic is still tricky). 

> Please could you fix coverage, otherwise happy to merge.

Fixed it, is back at 100% again.

samuelcolvin:
thank you so much @daviskirk. I think this has to be single most impressive contribution to pydantic in a long time.

ofek:
Will this be released soon?

PR #1795, opened 2020-08-05
[fixes #1794][docs] clarify argument type to parse_file

mdavis-xyz:
## Change Summary

Clarify that `parse_file` takes a file *path* not a file-like object.

## Related issue number

closes #1794 

## Checklist

* [ ] Unit tests for the changes exist - `make docs` fails, but it fails on `master` with the same error
* [ ] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable
* [x] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)

PR #1990, opened 2020-10-10
chore: change example path in documentation to not be tracked by git

### LINKED ISSUE #1992: data.json file appears after `make docs`
# Step to reproduce

Follow the steps [here](https://pydantic-docs.helpmanual.io/contributing/).

When running `make docs`, a new file `data.json` appears at `/`. This should not appear there, or should already be committed there, or should be excluded by `.gitignore`.

For context, see [this comment](https://github.com/samuelcolvin/pydantic/pull/1987/files#r502764774).

PrettyWood:
To make `parse_file` more explicit in #1795, a file `data.json` is created
once we serve the documentation. But it's not ignored by git.
Instead of adding `data.json` in `.gitignore`, let's just change the path
to be with all the rest of the documentation

Screen Shot 2020-10-10 at 9 31 31 PM

Closes #1992

### DIFF
--- a/docs/examples/models_parse.py
+++ b/docs/examples/models_parse.py
@@ -1,5 +1,7 @@
 import pickle
 from datetime import datetime
+from pathlib import Path
+
 from pydantic import BaseModel, ValidationError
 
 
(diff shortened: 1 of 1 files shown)


samuelcolvin:
I decided #1994 was a better way of doing, not poluting the docs.

Thanks for this.

PR #1991, opened 2020-10-11
fix: check only first sublevel for validators with `each_item`

### LINKED ISSUE #1255: each_item is applied recursively
Setting `each_item=True` to a validator of a nested collection is apparently called recursively for each singleton of the collection, instead of the items of the collection:

```
pydantic version: 1.4
pydantic compiled: True
python version: 3.7.1 (default, Dec 14 2018, 19:46:24)  [GCC 7.3.0]
platform: Linux-4.4.0-173-generic-i686-with-debian-stretch-sid
```

```py
from pydantic import validator
from typing import Sequence

@dataclass
class DemoModel:
    square_numbers: Sequence[int] = ()
    square_number_lists: Sequence[Sequence[int]] = ()

    @validator('square_numbers', each_item=True)
    def check_squares(cls, v):
        assert v ** 0.5 % 1 == 0, f'{v} is not a square number'
        return v

    @validator('square_number_lists', each_item=True)
    def check_square_lists(cls, v):
        assert isinstance(v, Sequence), type(v)
        return v

print(DemoModel(square_numbers=[1, 4, 9]))
print(DemoModel(square_number_lists=[[1, 4, 9], [16, 25]]))
```

```
DemoModel(square_numbers=[1, 4, 9], square_number_lists=())
  File "foo.py", line 22, in <module>
    print(DemoModel(square_number_lists=[[1, 4, 9], [16, 25]]))
  File "<string>", line 4, in __init__
  File "pydantic/dataclasses.py", line 83, in pydantic.dataclasses._process_class._pydantic_post_init
    # Raise because not adding these methods would break the "frozen-ness"
pydantic.error_wrappers.ValidationError: 5 validation errors for DemoModel
square_number_lists -> 0 -> 0
  <class 'int'> (type=assertion_error)
square_number_lists -> 0 -> 1
  <class 'int'> (type=assertion_error)
square_number_lists -> 0 -> 2
  <class 'int'> (type=assertion_error)
square_number_lists -> 1 -> 0
  <class 'int'> (type=assertion_error)
square_number_lists -> 1 -> 1
  <class 'int'> (type=assertion_error)
```

### LINKED ISSUE #1933: `each_item` should iterate only at the top-level instead of going through nested items
```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: ...snip...
               python version: 3.6.9 (default, Jun 26 2020, 13:24:16)  [GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.16)]
```

I'm seeing what appears to be inconsistent behavior with `each_item=True` when I'm using a field that's a list of tuples.

For instance, a 'normal' model illustrates what seems to be logical behavior.

```py
from typing import List, Tuple

class Thing(BaseModel):
    things: List[str]

    @validator("things", each_item=False)
    def validate_all_things(cls, val):
        print(f"all vals: {val}")
        return val

    @validator("things", each_item=True)
    def validate_single_thing(cls, val):
        print(f"single val: {val}")
        return val


>>> Thing(things=["one", "two", "three,four"])

single val: one
single val: two
single val: three,four
all vals: ['one', 'two', 'three,four']
Thing(things=['one', 'two', 'three,four'])
```

But if the field is anything like `List[Tuple[str], str]]`, then behavior is absolutely different.

```py
class Thing2(BaseModel):
    things: List[Tuple[str, str]]

    @validator("things", each_item=False)
    def validate_all_things(cls, val):
        print(f"all vals: {val}")
        return val

    @validator("things", each_item=True)
    def validate_single_thing(cls, val):
        print(f"single val: {val}")
        return val


>>> Thing2(things=[("one", "two"), ("three", "four")])

single val: one
single val: two
single val: three
single val: four
all vals: [('one', 'two'), ('three', 'four')]
Thing2(things=[('one', 'two'), ('three', 'four')])
```

The output from `validate_all_things` makes perfect sense - it's just the entire list passed in.

I would expect, however, that `each_item` would do just what it implies, namely calling the validator on each item of the list **no matter what the item is**. Instead, it's operating on every value within each tuple within the list, and I cannot get it to perform in a way where `val == ("one", "two")`.

Is this expected behavior? If so, why?

I mean, strings are iterable just like tuples, so why isn't `each_item` iterating over the characters of each string in the first case?

PrettyWood:
## Change Summary
When using a validator with `each_item`, the items are all validated
one by one. But if an item is also iterable the subitems are then
validated because the validator is kept as it is.
Now the validator passed to the items is changed and won't be propagated

## Related issue number
closes #1255
closes #1933

## Checklist


samuelcolvin:
I'll reply on the issue.

samuelcolvin:
see https://github.com/samuelcolvin/pydantic/issues/1933#issuecomment-735418328, let's update this to add `recursive`.

PrettyWood:
@samuelcolvin Glad you accepted the proposal! Done here 8c336abdd8bd3de5cec97c319ed44028704d4144

PrettyWood:
@samuelcolvin Glad you accepted the proposal! I added the `recursive` kwarg to ensure the breaking change is easy to fix. But as you can see in [my test](https://github.com/samuelcolvin/pydantic/pull/1991/files#diff-6ab916975a15e776ca67e412324892450797a072021e31b5485e82172e167b04R590) the behaviour before was really weird. **Before it was not really recursive but more a validation for the "last sublevel" based on the field type!** if I'm not mistaken.
Should I rename the kwarg to `deep` or `last_sublevel`? `recursive` may be misleading. Should I try to explain the old behaviour?
Or should we actually have a real "recursive" way of validating in which case the old behaviour was buggy

samuelcolvin:
Sorry for the slow reply.

I agree `recusive` is wrong. Let's drop it and call this a breaking change.

PrettyWood:
Since we agree on the breaking change I removed the last few commits and just added the "breaking change" note with some basic explanation

Issue #1984, opened 2020-10-09
The big Onion


samuelcolvin:
I'm wondering about making a major change to pydantic's internals in v2.

Validators are currently a list of functions that are called one after another, here's the actual code

https://github.com/samuelcolvin/pydantic/blob/e8326f899e037904a8e47928e4eedec6a56cf5ef/pydantic/fields.py#L730-L735

I want to switch to an "Onion" in much the same way many web frameworks implement middleware, [here is a description from django's docs](https://docs.djangoproject.com/en/3.1/topics/http/middleware/#middleware-order-and-layering).

The idea is that validation would be done by one function, which when called, called the next layer down until you get to the inner-most function which just does some parsing/validation and returns the value.

### No more pre=True, pre=False

Your custom validators can now look like
```py
@validator('foobar')
def foobar_validator(cls, v, handler)
    v = v * 2
    v = handler(v)
    return v.upper()
```
(or something more sane)

Instead of either `pre=True` or `pre=False`, and the need for multiple validators if you want to do both.

### Skipping futher validation

The validator can also choose **not** to call `handler` and thereby skip heavy validation in some cases, or implement custom behaviour.

```py
@validator('ts')
def ts_validator(cls, v, handler)
    if v == 'now':
        return datetime.now()
    else:
        return handler(v)
```

This might be useful for the case of `None` which currently has lots of custom logic around not calling further validators when the value is `None`.

### Catching errors

You could also catch exceptions from inner validation and modify the error or continue with some default value.

### Maybe faster in simple cases

For very simple cases (which are very common) e.g. a plain string, validation would be as simple as calling one function.

This should be much faster than the current "iterate oven a list of one element". Edit: or perhaps not, hard to say.

### Much faster with lots of validators

Code like

https://github.com/samuelcolvin/pydantic/blob/e8326f899e037904a8e47928e4eedec6a56cf5ef/pydantic/types.py#L467-L471

can be modified to

```py
    @classmethod
    def __validate__(cls, value: Any, **kwargs) -> Decimal:
        value = decimal_validator(value, **kwargs)
        value = number_size_validator(value, **kwargs)
        value = number_multiple_validator(value, **kwargs)
        ... do custom decimal validation
```

Which cython should be able to compile to be much faster.

### simplify the interface to custom types

Instead of the current `__get_validators__` interface, custom types could simply have a `__validate__` method, which is called for validation. We could do this anyway I guess, but it would make more sense with the onion.

### Better stack trace

[...]

PR #1993, opened 2020-10-11
Merge _generic_validator_basic and _generic_validator_cls into one function

uriyyo:
## Change Summary
Merge `_generic_validator_basic` and `_generic_validator_cls` into one function.

They have similar functionality and can be merged into one function.

## Checklist

### DIFF
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -253,7 +253,7 @@ def make_generic_validator(validator: AnyCallable) -> 'ValidatorCallable':
             # assume the first argument was value which has already been removed
             skip_first = True
 
-        return wraps(validator)(_generic_validator(validator, sig, set(args), skip_first))
+        return _generic_validator(validator, sig, set(args), skip_first)
 
 
 def prep_validators(v_funcs: Iterable[AnyCallable]) -> 'ValidatorsList':
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -283,7 +283,11 @@ def _generic_validator(
 
     # Actual signature is (cls, value, values, field, config)
     # cls and value is required, other args is optional
-    return lambda *args_: validator(*args_[skip_first:required_args], **_fetch_args(args_[required_args:]))
+    @wraps(validator)
+    def validator_wrapper(*args):
+        return validator(*args_[skip_first:required_args], **_fetch_args(args_[required_args:]))
+        
+    return validator_wrapper
 
 
 def gather_all_validators(type_: 'ModelOrDc') -> Dict[str, classmethod]:
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -284,9 +284,9 @@ def _generic_validator(
     # Actual signature is (cls, value, values, field, config)
     # cls and value is required, other args is optional
     @wraps(validator)
-    def validator_wrapper(*args):
+    def validator_wrapper(*args_: Any) -> Any:
         return validator(*args_[skip_first:required_args], **_fetch_args(args_[required_args:]))
-        
+
     return validator_wrapper
 
 
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -244,86 +244,46 @@ def make_generic_validator(validator: AnyCallable) -> 'ValidatorCallable':
             f'Invalid signature for validator {validator}: {sig}, "self" not permitted as first argument, '
             f'should be: (cls, value, values, config, field), "values", "config" and "field" are all optional.'
         )
-    elif first_arg == 'cls':
-        # assume the second argument is value
-        return wraps(validator)(_generic_validator_cls(validator, sig, set(args[1:])))
     else:
-        # assume the first argument was value which has already been removed
-        return wraps(validator)(_generic_validator_basic(validator, sig, set(args)))
+        if first_arg == 'cls':
+            # assume the second argument is value
+            args = args[1:]
+            skip_first = False
+        else:
+            # assume the first argument was value which has already been removed
+            skip_first = True
+
+        return wraps(validator)(_generic_validator(validator, sig, set(args), skip_first))
 
 
 def prep_validators(v_funcs: Iterable[AnyCallable]) -> 'ValidatorsList':
     return [make_generic_validator(f) for f in v_funcs if f]
 
 
-all_kwargs = {'values', 'field', 'config'}
+args_order = ('values', 'field', 'config')
+all_kwargs = {*args_order}
+required_args = 2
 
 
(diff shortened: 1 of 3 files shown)


tests/test_validators.py
```diff
@@ -708,7 +708,7 @@ def test_validator(cls, v, foobar):
 
     with pytest.raises(ConfigError) as exc_info:
         make_generic_validator(test_validator)
-    assert ': (cls, v, foobar), should be: (cls, value, values, config, field)' in str(exc_info.value)
+    assert ': (cls, v, foobar), should be: (cls, value, values, field, config)' in str(exc_info.value)
```
daviskirk, line 711: Will this break existing code or is it just the error message that is ordered now?

And as a nice feature of pytest, if you use the "matches" argument in `raises` (https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions) the error message will be nicer when the test fails: 

```suggestion
    expected_str =  r',*: (cls, v, foobar), should be: (cls, value, values, field, config).*'
    with pytest.raises(ConfigError, match=expected_str) as exc_info:
        make_generic_validator(test_validator)
```

tests/test_validators.py
```diff
@@ -690,7 +690,7 @@ def test_validator(v, foobar):
 
     with pytest.raises(ConfigError) as exc_info:
         make_generic_validator(test_validator)
-    assert ': (v, foobar), should be: (value, values, config, field)' in str(exc_info.value)
+    assert ': (v, foobar), should be: (value, values, field, config)' in str(exc_info.value)
```
daviskirk, line 693: here same as below

pydantic/class_validators.py
```diff
@@ -244,86 +244,46 @@ def make_generic_validator(validator: AnyCallable) -> 'ValidatorCallable':
...
-    elif args == {'values', 'config'}:
-        return lambda cls, v, values, field, config: validator(v, values=values, config=config)
-    elif args == {'field', 'config'}:
-        return lambda cls, v, values, field, config: validator(v, field=field, config=config)
-    else:
-        # args == {'values', 'field', 'config'}
-        return lambda cls, v, values, field, config: validator(v, values=values, field=field, config=config)
+    # Actual signature is (cls, value, values, field, config)
+    # cls and value is required, other args is optional
+    return lambda *args_: validator(*args_[skip_first:required_args], **_fetch_args(args_[required_args:]))
```
daviskirk, line 286: Now that there's ony one left, perhaps it would be "nicer" if this were changed to a "def" instead of passing the lambda around or even just wrap it right away since we already have the validator we need in this function.

```suggestion
    @wraps(validator)
    def validator_wrapper(*args):
        return validator(*args_[skip_first:required_args], **_fetch_args(args_[required_args:]))
        
    return validator_wrapper
```

uriyyo: LGTM, it looks much better. I also thought about replacing `lambda` with a regular `def` function.

tests/test_validators.py
uriyyo: There was a bug with a validator error message with incorrect order of arguments. Correct order is `(cls, value, values, field, config)` but error message showed `(cls, value, values, config, field)`.

I will update PR to use `match` in `pytest.raises`.

Thanks, for review.

tests/test_validators.py
uriyyo: Updated

daviskirk (approved):
Looks great to me!

pydantic/class_validators.py
```diff
@@ -244,86 +244,50 @@ def make_generic_validator(validator: AnyCallable) -> 'ValidatorCallable':
...
-        return wraps(validator)(_generic_validator_basic(validator, sig, set(args)))
+        if first_arg == 'cls':
+            # assume the second argument is value
+            args = args[1:]
+            skip_first = False
+        else:
+            # assume the first argument was value which has already been removed
+            skip_first = True
+
+        return wraps(validator)(_generic_validator(validator, sig, set(args), skip_first))
```
daviskirk, line 256: just noticed that I think the wraps here is redundant now that it's in the _generic_validator function

daviskirk (changes requested):

pydantic/class_validators.py
```diff
@@ -244,86 +244,50 @@ def make_generic_validator(validator: AnyCallable) -> 'ValidatorCallable':
...
-    elif args == {'values', 'config'}:
-        return lambda cls, v, values, field, config: validator(cls, v, values=values, config=config)
-    elif args == {'field', 'config'}:
-        return lambda cls, v, values, field, config: validator(cls, v, field=field, config=config)
-    else:
-        # args == {'values', 'field', 'config'}
-        return lambda cls, v, values, field, config: validator(cls, v, values=values, field=field, config=config)
-
+    def _fetch_args(args_: Tuple[Any, ...]) -> Dict[str, Any]:
+        return {k: args_[args_order.index(k)] for k in args_order if k in args}
```
daviskirk, line 282: This is just a guess but could it be that @samuelcolvin used the big lambda list for performance reasons since fetch args is now run on every call of the validator?  

A side anecdote, while working on another pull request I have to say that this does make development MUCH easier... I was taking for ever to make all the changes in the correct places and since the validator code swallows the errors it's hard to track down the exact place where stuff fails, so if there aren't performance or editor/ide support reasons keeping the old versions I'm all for using your version here.

pydantic/class_validators.py
uriyyo: Totally agree with you, thanks.

pydantic/class_validators.py
uriyyo: @daviskirk I faced the same problem when was debugging an issue with `pydantic`, it's nice to have one entry point to call validator functions.

But I should say that code with lambdas was more explicit rather than my version.

It will be great to hear @samuelcolvin opinion about this PR🙂

daviskirk: Fyi @uriyyo I was using your code in https://github.com/samuelcolvin/pydantic/pull/2018 and went ahead and checked the benchmarks `python benchmarks/run.py` or equivalent make command and they were indeed worse than the baseline using master.
That said I might have slowed things down with the other changes I made, just wanted to say that the benchmarks are there and run if you want to do a comparison on this code as well.

uriyyo: @daviskirk Thanks for such important information.
Looks like I found places which lent for performance issues.

After update, if compare my branch and current master branch, benchmark results are:
my branch - pydantic best=50.963μs/iter avg=52.860μs/iter stdev=3.121μs/iter
master - pydantic best=49.436μs/iter avg=50.348μs/iter stdev=0.712μs/iter

samuelcolvin:
This was done for a very good reason to aid cython in efficiently compiling the validators.

I want to completely change how validators work, see #1984. But this is not the way.

PR #2006, opened 2020-10-14
Allow generics to extend concrete classes. fixes #2005

### LINKED ISSUE #2005: GenericModel - Allow subclasses of concrete generics to be generic
### Checks

# Feature Request

```
             pydantic version: 1.6.1
            pydantic compiled: False
                 install path: /wayfair/home/ch438l/pydantic/pydantic
               python version: 3.6.8 (default, Aug 10 2019, 06:54:07)  [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
                     platform: Linux-3.10.0-862.3.2.el7.x86_64-x86_64-with-centos-7.5.1804-Core
```
I'd like to be able to define a generic model, then create a subclass of that generic model that both fills in the type parameter of the superclass, while introducing new type parameters. Here's a currently failing test case that illustrates the issue:

```py
@skip_36
def test_generic_subclass_of_concrete_generic():
    T = TypeVar("T")
    U = TypeVar("U")

    class GenericBaseModel(GenericModel, Generic[T]):
        data: T

    class GenericSub(GenericBaseModel[int], Generic[U]):
        extra: U

    GenericSub[str]
```

Currently that test case fails with `TypeError: Cannot parameterize a concrete instantiation of a generic model`

This happens because `GenericBaseModel[int].__concrete__` gets set to `True`, and `GenericSub` inherits it.

This can easily be fixed with the following change to `generics.py`:

```py
        if cls.__concrete__:
            raise TypeError('Cannot parameterize a concrete instantiation of a generic model')
```

Should become

```py
        if cls.__concrete__ and typing.Generic not in cls.__bases__:
            raise TypeError('Cannot parameterize a concrete instantiation of a generic model')
```

choogeboom:
## Change Summary

Allows models to extend other generic models and add new type parameters, while completely filling out the superclass parameters.

## Related issue number

fixes #2005

## Checklist

### DIFF
--- a/docs/examples/models_generics_inheritance_extend.py
+++ b/docs/examples/models_generics_inheritance_extend.py
@@ -3,15 +3,17 @@ from pydantic.generics import GenericModel
 
 TypeX = TypeVar('TypeX')
 TypeY = TypeVar('TypeY')
+TypeZ = TypeVar('TypeZ')
 
 
-class BaseClass(GenericModel, Generic[TypeX]):
+class BaseClass(GenericModel, Generic[TypeX, TypeY]):
     x: TypeX
-
-
-class ChildClass(BaseClass[int], Generic[TypeY]):
     y: TypeY
 
 
+class ChildClass(BaseClass[int, TypeY], Generic[TypeY, TypeZ]):
+    z: TypeZ
+
+
 # Replace TypeY by str
-print(ChildClass[str](x=1, y='y'))
+print(ChildClass[str, int](x=1, y='y', z=3))
--- a/docs/examples/models_generics_inheritance_extend.py
+++ b/docs/examples/models_generics_inheritance_extend.py
@@ -1,0 +1,17 @@
+from typing import TypeVar, Generic
+from pydantic.generics import GenericModel
+
+TypeX = TypeVar('TypeX')
+TypeY = TypeVar('TypeY')
+
+
+class BaseClass(GenericModel, Generic[TypeX]):
+    x: TypeX
+
+
+class ChildClass(BaseClass[int], Generic[TypeY]):
+    y: TypeY
+
+
+# Replace TypeY by str
+print(ChildClass[str](x=1, y='y'))
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -319,6 +319,14 @@ To inherit from a GenericModel without replacing the `TypeVar` instance, a class
 ```
 _(This script is complete, it should run "as is")_
 
+You can also create a generic subclass of a `GenericModel` that partially or fully replaces the type 
+parameters in the superclass.
+
+```py
+{!.tmp_examples/models_generics_inheritance_extend.py!}
+```
+_(This script is complete, it should run "as is")_
+
 If the name of the concrete subclasses is important, you can also override the default behavior:
 
 ```py
--- a/tests/test_generics.py
+++ b/tests/test_generics.py
@@ -603,8 +603,8 @@ def test_multiple_specification():
 
 @skip_36
 def test_generic_subclass_of_concrete_generic():
-    T = TypeVar("T")
-    U = TypeVar("U")
+    T = TypeVar('T')
+    U = TypeVar('U')
 
     class GenericBaseModel(GenericModel, Generic[T]):
         data: T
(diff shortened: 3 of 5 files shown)


tests/test_generics.py
```diff
@@ -599,3 +599,22 @@ class Model(GenericModel, Generic[AT, BT]):
...
+
+@skip_36
+def test_generic_subclass_of_concrete_generic():
+    T = TypeVar('T')
+    U = TypeVar('U')
+
+    class GenericBaseModel(GenericModel, Generic[T]):
+        data: T
+
+    class GenericSub(GenericBaseModel[int], Generic[U]):
```
daviskirk, line 612: Not sure if this is even part of the issue or if typing even works that way but can we do stuff like:

```
class GenericSub(GenericBaseModel[U], Generic[U]):
    extra: U

GenericSub[int](data='string')  # raises validation error because data is expected to be int
```

?

tests/test_generics.py
```diff
@@ -599,3 +599,22 @@ class Model(GenericModel, Generic[AT, BT]):
...
+        extra: U
+
+    ConcreteSub = GenericSub[int]
+
+    with pytest.raises(ValidationError):
+        ConcreteSub(data=2, extra='wrong')
+
```
daviskirk, line 619: I'd add a check that data="wrong" also raises an error.  If the type wasn't correctly mapped for some reason it would validate data as "any" which would work

docs/usage/models.md
```diff
@@ -319,6 +319,14 @@ To inherit from a GenericModel without replacing the `TypeVar` instance, a class
 ```
 
+You can also create a generic subclass of a `GenericModel` that partially or fully replaces the type
```
daviskirk, line 322: The example only goes into the "fully replaced" version.  Does a partial replacement also work or is that dependent on #1989  and/or https://github.com/samuelcolvin/pydantic/issues/2002?

If it really does only work inheriting from a fully concrete class, I'd probably leave out this doc section (as that would be the expected behavior anyway I would think).

docs/examples/models_generics_inheritance_extend.py
```diff
@@ -0,0 +1,17 @@
...
+TypeX = TypeVar('TypeX')
+TypeY = TypeVar('TypeY')
+
+
+class BaseClass(GenericModel, Generic[TypeX]):
+    x: TypeX
+
+
+class ChildClass(BaseClass[int], Generic[TypeY]):
+    y: TypeY
```
daviskirk, line 13: See the comment below.  Would this also work?
I would think that this would be the "new feature" that requires documentation.  
While the above should just work anyway (and thanks to your fix it now does ;))

```python
TypeX = TypeVar('TypeX')
TypeY = TypeVar('TypeY')
TypeZ = TypeVar('TypeZ')



class BaseClass(GenericModel, Generic[TypeX, TypeX]):
    x: TypeX
    y: TypeY


class ChildClass(BaseClass[int, TypeY], Generic[TypeY, TypeZ]):
    z: TypeZ

# x is type int, y and z are type str
print(ChildClass[str, str](x=1, y='y', z='z'))

```

daviskirk (changes requested):

daviskirk:
@choogeboom nice!  Just added a few Questions

docs/usage/models.md
choogeboom: I'm guessing the partially replaced version depends on #1989

choogeboom: I can remove it if you think it's superfluous.

tests/test_generics.py
choogeboom: Yes, that is how I would expect it to work.

tests/test_generics.py
choogeboom: Good call. Will do.

docs/examples/models_generics_inheritance_extend.py
choogeboom: Yeah, that should work once #1989 is merged. Unfortunately, the update to the documentation doesn't really fit in either of our PRs, since this example won't work without both.

Should the doc update happen in a separate PR after both are merged, or should we just include it in one of these PRs? I favor updating the documentation in this PR, so it doesn't get forgotten. The documentation should outline how pydantic is _supposed_ to work, right?

Also, I like your example better than mine. If we decide to keep the documentation update here, then I'll update my example to match.

daviskirk (approved):
Looks good to me!

docs/examples/models_generics_inheritance_extend.py
PrettyWood: @choogeboom @daviskirk Great job guys on all the generic part 👍  I agree with you I feel like the doc should be updated in one of your PRs, probably this one as I would first merge #1989 too.

samuelcolvin:
great, thank you.

PR #2018, opened 2020-10-19
Add model contexts and context option to BaseModel.parse_obj

### LINKED ISSUE #1549: Context when validating data
This is discussed at length in #1170 but I'm creating a new issue to make things clearer.

---

Pydantic needs a way of accessing "context" when validating data, serialising data, creating schema.

Might be used via `MyModel.parse_obj(raw_data, context=my_context)`.

Use cases:
* dynamic choices - E.g. when choosing from a select based on a entities you have access to in a db, obviously both the validation and schema for the field should be dynamic - this is very common frustration for me
* dynamic parameters on fields, e.g. max and min
* cases like [this](https://github.com/samuelcolvin/sasstastic/blob/5777b98e328bbdcee99a984f3bc1909e36584e6e/sasstastic/config.py#L77-L88)
* and so many more...

Questions/concerns:
* adding context with `parse_obj`, but I don't see way way to add context with `MyModel(**data)` maybe that's fine
* how do we include context to field definitions? e.g. `foobar: conint(gt=context['...'])` won't work
* how will context work in sub-models?
* will this require breaking changes that need it to go into v2? I hope not.

Another option is to encourage one of the options from #1170 and/or make one of them easier.

Feedback on how this might work and when you would use it very welcome.

daviskirk:
## Change Summary

This pull request adds the possibility for passing context to `BaseModel.parse_obj`

Fixes #1549

While a basic test is implemented, I will do a lot more work in making sure this actually works for all the cases we have.  However, I would love to have some feedback if this is going in the right direction or not!

Disclaimer:

Unfortunately, I had trouble getting my head around all the various lambda combinations in `class_validators.py` and thought that passing the validation arguments by keywords would simplify things (started with code from #1993).  It did, but it also made a smart dent on the benchmarks, so I'll probably go back and rewrite this the hard way.
 
## Related issue number

#1549 

## Checklist

### DIFF
--- a/changes/1993-uriyyo.md
+++ b/changes/1993-uriyyo.md
@@ -1,0 +1,1 @@
+Merge `_generic_validator_basic` and `_generic_validator_cls` into one function
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -1,9 +1,11 @@
+import sys
 import warnings
 from collections import ChainMap
-from functools import wraps
+from functools import lru_cache, update_wrapper, wraps
+from inspect import CO_VARKEYWORDS, Parameter, signature
 from itertools import chain
-from types import FunctionType
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Type, Union, overload
+from types import CodeType, FunctionType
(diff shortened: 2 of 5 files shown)


samuelcolvin:
The solution is #2034 (I just created it since I thought it deserved it's own issue), unfortunately that needs to wait for v2.

samuelcolvin:
This is implemented in pydantic-core, I don't think we should add this in V1.10 so closing this PR.

PR #2028, opened 2020-10-23
Make a v1.6.2 release for python 3.9 support

virtuald:
This branch isn't intended to actually be merged -- just doing it for the CI. It was created like so:

```
git checkout v1.6.1
git cherry-pick 794d0bccf93b006b388a565aabc523ab34920a8c
[make edits to history/version in 63c64659f7935b2a3e5ab425ad205187e2336065]
```

I assume that a new release hasn't been made because there are changes in master that still need consideration before a new release is made. However, it would be great if just the bare minimum for python 3.9 could be pushed as v1.6.2. 

Thanks for your consideration!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.6', '3.7', '3.8']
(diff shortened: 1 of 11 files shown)


samuelcolvin:
Too much changed in #1844 to allow for a patch release.

I'm going to work on merging what I can today, then release 1.7 as soon as possible.

PR #2031, opened 2020-10-25
Add link to issue about GetterDict

nuno-andre:
It's not easy to find some information about `GetterDict` ;)

### DIFF
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -59,8 +59,8 @@ Options:
 : whether to allow usage of [ORM mode](models.md#orm-mode)
 
 **`getter_dict`**
-: a custom class (which should inherit from `GetterDict`) to use when decomposing ORM classes for validation,
-  for use with `orm_mode`
+: a custom class (which should inherit from `GetterDict`; see [#1689](https://github.com/samuelcolvin/pydantic/issues/1689))
+to use when decomposing ORM classes for validation, for use with `orm_mode`
 
 **`alias_generator`**
 : a callable that takes a field name and returns an alias for it


samuelcolvin:
thanks, would you consider adding a proper section to the docs with an example?

nuno-andre:
I finally found that a `Union` of custom types was best suited to my use case, which was a specialization. And way more clear! I'm really impressed with the amount of procedural code that can be replaced with a bunch of pretty and clean typed classes thanks to Pydantic.

But I have some other cases I want to try `GetterDict` with. If I come up with a good example, I will be glad to add it to the docs.

samuelcolvin:
Closing this, @nuno-andre if you're be willing to take the time to add a proper section to the docs about `GetterDict` that would be great.

In terms of an example, I think it's better to keep it understandable than realistic.

How about something like:

> Let's say we hand an ORM which used a class where to properties `_id` and `_name` were important to a document, by default they would be ignored by `GetterDict` and therefore pydantic. We could overcome that by implementing our own `CustomGetterDict` ....

nuno-andre:
Isn't that a case for `fields`? Something like

```
class Model(BaseModel):
    id: int
    name: str

    class Config:
        orm_mode = True
        fields = {'id': '_id', 'name': '_name'}
```


Would an example of managing different sources be representative of `GetterDict`? To take a very straightforward example, we've got the following model and we've got to deal with both ORM objects that come with proper attributes and objects that come with a `full_name` attribute made of `'{last_name}, {first_name}'` instead.

```
class Person(BaseModel):
    first_name: str
    last_name: str
```

PR #2032, opened 2020-10-25
Fix typing coverage

samuelcolvin:
## Change Summary

get coverage back to 100%

## Checklist

### DIFF
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -123,7 +123,7 @@ else:
         """
         args = typing_get_args(tp)
         # the fallback is needed for the same reasons as `get_origin` (see above)
-        return tuple(args or getattr(tp, '__args__', ()) or generic_get_args(tp))
+        return args or getattr(tp, '__args__', ()) or generic_get_args(tp)
 
 
 if TYPE_CHECKING:
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -121,9 +121,8 @@ else:
             get_args(Union[int, Tuple[T, int]][str]) == (int, Tuple[str, int])
             get_args(Callable[[], T][int]) == ([], int)
         """
-        args = typing_get_args(tp)
         # the fallback is needed for the same reasons as `get_origin` (see above)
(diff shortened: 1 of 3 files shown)


pydantic/typing.py
```diff
@@ -121,12 +121,9 @@ def get_args(tp: Type[Any]) -> Tuple[Any, ...]:
             get_args(Union[int, Tuple[T, int]][str]) == (int, Tuple[str, int])
             get_args(Callable[[], T][int]) == ([], int)
         """
-        try:
-            args = typing_get_args(tp)
-        except IndexError:
-            args = ()
+        args = typing_get_args(tp)
```
samuelcolvin, line 124: @PrettyWood I'm not sure about this, I couldn't find a case where `IndexError` is raised, hence removing it.

Do you know what you did to cause the error?

pydantic/typing.py
```diff
@@ -121,12 +121,9 @@ def get_args(tp: Type[Any]) -> Tuple[Any, ...]:
...
         # the fallback is needed for the same reasons as `get_origin` (see above)
-        return args or getattr(tp, '__args__', ()) or generic_get_args(tp)
+        return tuple(args or getattr(tp, '__args__', ()) or generic_get_args(tp))
```
samuelcolvin, line 126: This was need for the `conlist(str)` case in tests which was previously returning a string.

samuelcolvin:
@PrettyWood looks like there are still some differences between how `get_args` works for 3.6 & 3.7. vs. 3.8 and 3.9.

Any suggestions?

pydantic/typing.py
PrettyWood: You're absolutely right ! I did this before splitting `python 3.8` and `python 3.9`.
Now that it's under `sys.version_info >= (3, 9)` it can be removed !

PrettyWood:
@samuelcolvin The main issue is with generics that are not supported on python 3.6. I could have gone further but I didn't when working on supporting python 3.9 as it was a work on generics.
I hope this should be fixed with #1989 where `get_args` has been improved

PR #2040, opened 2020-10-26
build history and uprev version

samuelcolvin:
Prepare for release. I want to get v1.7 released so we can support python 3.9.

@PrettyWood @MrMrRobat are you happy with this as v1.7 or is there anything essential you think we need to add?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,4 +1,16 @@
-## v1.7a1 (2020-10-26)
+## v1.7 (2020-10-26)
+
+Thank you to pydantic's sponsors:
+@timdrijvers, @BCarley, @chdsbd, @tiangolo, @matin, @linusg, @kevinalh, @jorgecarleitao, @koxudaxi, @primer-api 
+for their kind support.
+
+### Highlights
+
+* python 3.9 support, thanks @PrettyWood
+* [Private model attributes](https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes), thanks @MrMrRobat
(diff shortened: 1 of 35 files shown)


PrettyWood:
@samuelcolvin I'm currently writing the doc for https://github.com/samuelcolvin/pydantic/pull/1817. Almost done! But it's your call if you're willing to wait

samuelcolvin:
> @samuelcolvin I'm currently writing the doc for #1817. Almost done! But it's your call if you're willing to wait

Got for it! Happy to wait a few hours for something so useful.

MrMrRobat:
> is there anything essential you think we need to add?

I had thoughts about `BaseModel` support builtin `dict()` signature, allowing usages like:
```py
Model(foo=1, bar=2)
Model({'foo': 1}, bar=2)
Model({'foo': 1, 'bar': 2})

data = {'override': False, 'foo': 2}
Model(data, override=True)
```

But I'm pretty sure it was discussed earlier

In this case `__init__ would look like this (no other changes required, afik)
```py
    def __init__(*self_and_data, **data: Any) -> None:
        self, *dict_arg = self_and_data
        if dict_args:
            data = dict(*dict_arg, **data)
        values, fields_set, validation_error = validate_model(self.__class__, data)
        if validation_error:
            raise validation_error
        object_setattr(self, '__dict__', values)
        object_setattr(self, '__fields_set__', fields_set)
        self._init_private_attributes()
```

samuelcolvin:
I think that's a bigger discussion, please create an issue. Let's skip that for this release.

PR #2046, opened 2020-10-27
Prevent fields from being deleted from BaseModel.__dict__ when a field validator raises an unexpected exception

johnsabath:
## Change Summary

Addressing https://github.com/samuelcolvin/pydantic/issues/2044

## Related issue number

fix https://github.com/samuelcolvin/pydantic/issues/2044

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -384,12 +384,16 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
 
             known_field = self.__fields__.get(name, None)
             if known_field:
-                original_value = self.__dict__[name]
-                value, error_ = known_field.validate(value, self.__dict__, loc=name, cls=self.__class__)
-                if error_:
+                original_value = self.__dict__.pop(name)
+                try:
+                    value, error_ = known_field.validate(value, self.__dict__, loc=name, cls=self.__class__)
+                    if error_:
+                        raise ValidationError([error_], self.__class__)
+                except Exception:
                     self.__dict__[name] = original_value
-                    raise ValidationError([error_], self.__class__)
-                new_values[name] = value
+                    raise
+                else:
+                    new_values[name] = value
 
             errors = []
             for skip_on_failure, validator in self.__post_root_validators__:
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -384,7 +384,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
 
             known_field = self.__fields__.get(name, None)
             if known_field:
-                original_value = self.__dict__.pop(name)
+                original_value = self.__dict__[name]
                 value, error_ = known_field.validate(value, self.__dict__, loc=name, cls=self.__class__)
                 if error_:
                     self.__dict__[name] = original_value
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -1157,20 +1157,25 @@ def test_nested_literal_validator():
     ]
 
 
-def test_field_that_is_being_validated_is_excluded_from_validator_values():
-    validators_values = None
+def test_field_that_is_being_validated_is_excluded_from_validator_values(mocker):
+    check_values = mocker.MagicMock()
 
     class Model(BaseModel):
         foo: str
         bar: str
 
+        class Config:
+            validate_assignment = True
+
         @validator('foo')
         def validate_foo(cls, v, values):
-            nonlocal validators_values
-            validators_values = values
+            check_values({**values})
 
-    Model(foo='foo_value', bar='bar_value')
-    assert validators_values == {'bar': 'bar_value'}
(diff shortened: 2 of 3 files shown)


johnsabath:
I could add some additional unit tests for this edge case as well, just let me know how you'd like to handle it.

pydantic/main.py
```diff
@@ -384,7 +384,7 @@ def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
 
             known_field = self.__fields__.get(name, None)
             if known_field:
-                original_value = self.__dict__.pop(name)
+                original_value = self.__dict__[name]
```
PrettyWood, line 387: We used to check `self.dict(exclude={name})` [before](https://github.com/samuelcolvin/pydantic/pull/2000/files#diff-8708af8d26c26644976866f974e0b810b0940ede43ff487e457057d5a1f07999L370) and I'm the one who suggested to just pop the value and use `self.__dict__` to be faster. I think we may have some undesired behaviour if we keep the whole dictionary since `values` will have the previous value and `v` will have the new one.
I'm trying to think on possible edge cases with validators that would use `values`.
If there are such cases we could use `self.dict(exclude={name})` or use a `try...except...`

samuelcolvin:
what would be wrong with

```py
                try:
                    value, error_ = known_field.validate(value, self.__dict__, loc=name, cls=self.__class__)
                    if error_:
                        raise ValidationError([error_], self.__class__)
                except Exception:
                    self.__dict__[name] = original_value
                    raise
                else:
                    new_values[name] = value
```
?

pydantic/main.py
johnsabath: Ah, I see, that makes sense.

johnsabath:
With @PrettyWood's additional context on why the field should still be removed from __dict__ before calling the validators, that seems like a good compromise to me @samuelcolvin

pydantic/main.py
```diff
@@ -385,11 +385,15 @@ def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
             known_field = self.__fields__.get(name, None)
             if known_field:
                 original_value = self.__dict__.pop(name)
-                value, error_ = known_field.validate(value, self.__dict__, loc=name, cls=self.__class__)
-                if error_:
+                try:
+                    value, error_ = known_field.validate(value, self.__dict__, loc=name, cls=self.__class__)
+                    if error_:
+                        raise ValidationError([error_], self.__class__)
+                except Exception:
```
PrettyWood, line 392: You should add a test for this

johnsabath: on it

tests/test_validators.py
```diff
@@ -1155,3 +1155,38 @@ class Model(BaseModel):
...
+        foo: str
+        bar: str
+
+        @validator('foo')
+        def validate_foo(cls, v, values):
+            nonlocal validators_values
+            validators_values = values
+
+    Model(foo='foo_value', bar='bar_value')
+    assert validators_values == {'bar': 'bar_value'}
```
PrettyWood, line 1173: There are two issues here:
- you don't test with `validate_assignement = True`
- you use the reference of `values` so it won't work as `values` is then updated before being checked with `assert`

Btw you could simply use a mock to check the validation. Something like
```python
def test_field_that_is_being_validated_is_excluded_from_validator_values(mocker):
    check_values = mocker.MagicMock()

    class Model(BaseModel):
        foo: str
        bar: str

        class Config:
            validate_assignment = True

        @validator('bar')
        def validate_bar(cls, v, values):
            check_values({**values})
            return v

    model = Model(foo='foo_value', bar='bar_value')
    check_values.reset_mock()

    model.bar = 'new_bar_value'
    check_values.assert_called_once_with({'foo': 'foo_value'})
```

johnsabath: Apologies, I didn't have my local venv set up, so these changes were untested.

Just pushed up a new set of changes now, and I used a mock to capture the calls like you suggested.

PrettyWood: No worries. Thanks a lot for this!

PrettyWood (approved):
LGTM!

tests/test_validators.py
```diff
@@ -1155,3 +1155,43 @@ class Model(BaseModel):
...
+
+        @validator('foo')
+        def validate_foo(cls, v):
+            if v == 'raise_exception':
+                raise Exception()
+            return v
+
+    model = Model(foo='foo')
+    with pytest.raises(Exception):
+        model.foo = 'raise_exception'
```
samuelcolvin, line 1196:
```suggestion
                raise RuntimeError('test error')
            return v

    model = Model(foo='foo')
    with pytest.raises(RuntimeError, match='test error'):
        model.foo = 'raise_exception'
```

So we're sure we're catching the same error.

samuelcolvin (commented):
otherwise LGTM

PR #2050, opened 2020-10-27
make host optional in AnyUrl

### LINKED ISSUE #1983: AnyUrl doesn't accept file URLs with implicit host
# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /shared/conda/envs/ubide-mastr/lib/python3.6/site-packages/pydantic
               python version: 3.6.10 |Anaconda, Inc.| (default, May  8 2020, 02:54:21)  [GCC 7.3.0]
                     platform: Linux-5.4.0-48-generic-x86_64-with-debian-buster-sid
     optional deps. installed: []
```
```py
import pydantic

    url: pydantic.AnyUrl

MyModel(url='file:///foo/bar')
```
raises:
```
ValidationError: 1 validation error for MyModel
url
  URL host invalid (type=value_error.url.host)
```
However, `MyModel(url='file://hostname/foo/bar')`, i.e. a URL with an explicit hostname, works.

RFC 8089 [says](https://tools.ietf.org/html/rfc8089#section-2) that the hostname is optional. Hence I would expect that `AnyUrl` would accept that.

vgerak:
## Change Summary
- Add `host_required` parameter (`False` in `AnyUrl`, `True` on `HttpUrl`, `PostgresDsn`, `RedisDsn`, `stricturl`, as suggested
- Move (now) failing host validation tests from `AnyUrl` to `HttpUrl`, add relevant tests for Postgres, Redis, stricturl

## Related issue number
Fixes #1983 

## Checklist

* [x] Tests pass on CI & Coverage remains at 100%

### DIFF
--- a/changes/1983-vgerak.md
+++ b/changes/1983-vgerak.md
@@ -1,0 +1,2 @@
+Make host optional in `AnyUrl` to conform to [RFC 8089](https://tools.ietf.org/html/rfc8089#section-2)
+(now allows URLs like `"file:///foo/bar"`).
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -96,6 +96,7 @@ class AnyUrl(str):
     allowed_schemes: Optional[Set[str]] = None
     tld_required: bool = False
     user_required: bool = False
+    host_required: bool = False
 
     __slots__ = ('scheme', 'user', 'password', 'host', 'tld', 'host_type', 'port', 'path', 'query', 'fragment')
 
@@ -110,7 +111,7 @@ class AnyUrl(str):
         scheme: str,
         user: Optional[str] = None,
         password: Optional[str] = None,
-        host: str,
+        host: Optional[str] = None,
         tld: Optional[str] = None,
         host_type: str = 'domain',
         port: Optional[str] = None,
@@ -227,7 +228,8 @@ class AnyUrl(str):
                 break
 
         if host is None:
-            raise errors.UrlHostError()
+            if cls.host_required:
+                raise errors.UrlHostError()
         elif host_type == 'domain':
             is_international = False
             d = ascii_domain_regex().fullmatch(host)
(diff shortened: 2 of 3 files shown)


vgerak:
@samuelcolvin please review whenever convenient

samuelcolvin (commented):
I think this looks great. My only concern is that more URL types get added in future and forget to add `host_required = True`, e.g. #2049 will need to be updated after this or it's behaviour will fundamentally change.

Also people might have their own custom URL classes which inherit from `AnyUrl` which would have their behaviour changed in an important but subtle way.

How about we rename the URL base class from `AnyUrl` to `AbstractUrl`, and set `host_required = True` there.

We could also make all properties undefined on `AbstractUrl` so you had to define all the properties of all its subclasses

Then add `AnyUrl` with the current defaults.

The problem with this solution is that people who are already importing `AnyUrl` and subclassing it would still have the same problem. We could prevent `AnyUrl` from being inherited, but that's quite ugly.

Honestly, I'm not sure of the best way to proceed.

vgerak:
> The problem with this solution is that people who are already importing AnyUrl and subclassing it would still have the same problem.

I'm not sure I understand this. `AnyUrl` both now and after the proposed changes (having `AbstractUrl` as base) would require host to be set and therefore be backwards-compatible, no?

samuelcolvin:
I think we should make `host_required = True` on `AnyUrl`, then add `FileUrl` with `host_required = False`.

We can maybe change `AnyUrl.host_required = False` in v2.

samuelcolvin:
been a long time since the author responded, so I'm closing.

Anyone is welcome to create a new PR for this.

vgerak:
Life happened.. I'll try to get back to it again

samuelcolvin:
> Life happened..

Tell me about it. Not problem at all. Thanks for contributing 🙇

samuelcolvin:
I realise that ["tell me about it"](https://dictionary.cambridge.org/dictionary/english/tell-me-about-it) is a weird english aphorism, sorry.

vgerak:
No worries @samuelcolvin , I'm fluent enough to understand it (and I generally assume good intentions, so there would not be a misunderstanding in any case).

I think I've made a refactor you can check, along with small changes on the docs. Should we reopen this, so the discussion history remains, or you want me to open a new PR?

vgerak:
@samuelcolvin (or perhaps another maintainer - @PrettyWood ?) 
Should we reopen this PR to review the requested changes (while keeping the discussion history) or should I open a new one?

PR #2053, opened 2020-10-27
underscore_attrs_are_private causing TypeError

### LINKED ISSUE #2047: underscore_attrs_are_private causes TypeError
### Checks

# Bug

```
        pydantic version: 1.7
       pydantic compiled: True
            install path: /Users/ahedges/.pyenv/versions/3.7.8/envs/sdf/lib/python3.7/site-packages/pydantic
          python version: 3.7.8 (default, Sep 16 2020, 18:33:23)  [Clang 11.0.3 (clang-1103.0.32.59)]
```

I spent a decent amount of time this weekend trying to make a private field using code posted in #655. I was happy to see Pydantic 1.7 came out today and had support for private fields built in. I upgraded and tried to convert my code, but I encountered some unusual problems. (Even though it doesn't work perfectly, I still appreciate the feature.) Most are type errors from mypy (might report later), but this one is more serious. I simplified the problem below.

The issue is that `underscore_attrs_are_private` causes an exception where `PrivateAttr` does not.

When using `underscore_attrs_are_private` with the following code:

```python
from typing import Any

class TestObject(BaseModel):
    public_field: str
    _private_field: str

    class Config:
        underscore_attrs_are_private = True

    def __init__(self, **data: Any) -> None:
        super().__init__(**data)
        self._private_field = "bar"

print(TestObject(public_field="foo"))
```

I get the following output:

```
test.py:4: DeprecationWarning: __class__ not set defining 'TestObject' as <class '__main__.TestObject'>. Was __classcell__ propagated to type.__new__?
  class TestObject(BaseModel):
  File "test.py", line 15, in <module>
    print(TestObject(public_field="foo"))
  File "test.py", line 12, in __init__
    super().__init__(**data)
  File "pydantic/main.py", line 365, in pydantic.main.BaseModel.__init__
  File "pydantic/main.py", line 424, in pydantic.main.BaseModel._init_private_attributes
  File "pydantic/fields.py", line 821, in pydantic.fields.PrivateAttr.get_default
  File "pydantic/utils.py", line 624, in pydantic.utils.smart_deepcopy
  File "/Users/ahedges/.pyenv/versions/3.7.8/lib/python3.7/copy.py", line 169, in deepcopy
    rv = reductor(4)
TypeError: can't pickle cell objects
```

However, when using `PrivateAttr` with the following code:

```python
from typing import Any

class TestObject(BaseModel):
    public_field: str
    _private_field: str = PrivateAttr()

    def __init__(self, **data: Any) -> None:
        super().__init__(**data)
        self._private_field = "bar"

print(TestObject(public_field="foo"))
```

I get the following, desired output:

```
public_field='foo'
```

I also noticed that removing `__init__()` from the first example also prevents the crash. However, it is needed to set the private field.

Another thing to note is that in my full code, the exception caused by `underscore_attrs_are_private` appears but the `DeprecationWarning` does not. If you think this matters, I can try to reproduce my code without the warning.

samuelcolvin:
## Change Summary

fix `underscore_attrs_are_private` causing `TypeError` when overriding `__init__`

## Related issue number

fix #2047

## Checklist

### DIFF
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -631,4 +631,10 @@ def is_valid_field(name: str) -> bool:
 
 
 def is_valid_private_name(name: str) -> bool:
-    return not is_valid_field(name) and name not in {'__annotations__', '__module__', '__annotations__', '__qualname__'}
+    return not is_valid_field(name) and name not in {
+        '__annotations__',
+        '__module__',
+        '__annotations__',
+        '__qualname__',
+        '__classcell__',
+    }
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -631,10 +631,4 @@ def is_valid_field(name: str) -> bool:
 
 
 def is_valid_private_name(name: str) -> bool:
-    return not is_valid_field(name) and name not in {
-        '__annotations__',
-        '__module__',
-        '__annotations__',
-        '__qualname__',
-        '__classcell__',
-    }
+    return not is_valid_field(name) and name not in {'__annotations__', '__classcell__', '__module__', '__qualname__'}
--- a/Makefile
+++ b/Makefile
@@ -134,5 +134,5 @@ docs-serve:
 .PHONY: publish-docs
 publish-docs:
 	zip -r site.zip site
-	@curl -f -H "Content-Type: application/zip" -H "Authorization: Bearer ${NETLIFY}" \
+	@curl -H "Content-Type: application/zip" -H "Authorization: Bearer ${NETLIFY}" \
 	      --data-binary "@site.zip" https://api.netlify.com/api/v1/sites/pydantic-docs.netlify.com/deploys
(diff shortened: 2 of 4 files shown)


pydantic/utils.py
```diff
@@ -631,4 +631,10 @@ def is_valid_field(name: str) -> bool:
 
 
 def is_valid_private_name(name: str) -> bool:
-    return not is_valid_field(name) and name not in {'__annotations__', '__module__', '__annotations__', '__qualname__'}
+    return not is_valid_field(name) and name not in {
+        '__annotations__',
+        '__module__',
+        '__annotations__',
```
PrettyWood, line 637:
```suggestion
```
while we're at it!

PrettyWood (commented):
👍

pydantic/utils.py
```diff
@@ -631,4 +631,10 @@ def is_valid_field(name: str) -> bool:
...
+        '__qualname__',
+        '__classcell__',
```
aphedges, line 639: It might be good to put this in alphabetical order.

PR #2055, opened 2020-10-27
Add two overload variants to `validate_arguments`

### LINKED ISSUE #2052: mypy "error: <nothing> not callable" when passing config to validate_arguments
# Bug

```
             pydantic version: 1.7
            pydantic compiled: True
                 install path: /home/brian/repos/medigator/venv/lib/python3.6/site-packages/pydantic
               python version: 3.6.9 (default, Oct  8 2020, 12:12:24)  [GCC 8.4.0]
                     platform: Linux-4.15.0-121-generic-x86_64-with-Ubuntu-18.04-bionic
```

The example code from the docs demonstrating the use of passing a config to `validate_arguments` causes an error when running mypy (version 0.790):

```py
from pydantic import ValidationError, validate_arguments


class Foobar:
    def __init__(self, v: str):
        self.v = v

    def __add__(self, other: 'Foobar') -> str:
        return f'{self} + {other}'

    def __str__(self) -> str:
        return f'Foobar({self.v})'


@validate_arguments(config=dict(arbitrary_types_allowed=True))
def add_foobars(a: Foobar, b: Foobar):
    return a + b
```

```
test.py:15: error: <nothing> not callable
```

If I remove the `config` parameter but still call `validate_arguments` it still occurs.
If I remove the call and just decorate with `@validate_arguments` then mypy is happy.

I assume it has something to do with the type annotation of `validate_arguments`.

layday:
## Change Summary

Added two overload variants to `validate_arguments` for the nested decorator signature.

## Related issue number

fix #2052

## Checklist

* [x] ~~Documentation reflects the changes where applicable~~

### DIFF
--- a/changes/2055-layday.md
+++ b/changes/2055-layday.md
@@ -1,0 +1,1 @@
+add two overload variants to `validate_arguments` for the nested decorator signature
--- a/changes/2055-layday.md
+++ b/changes/2055-layday.md
@@ -1,1 +1,1 @@
-add two overload variants to `validate_arguments` for the nested decorator signature
+fix annotation of `validate_arguments` when passing configuration as argument
--- a/tests/mypy/modules/fail4.py
+++ b/tests/mypy/modules/fail4.py
@@ -24,13 +24,3 @@ def bar() -> str:
 
 # return type should be a string
 y: int = bar()
-
-
-# nested decorator should not produce an error
-@validate_arguments(config={'arbitrary_types_allowed': True})
-def baz(a: int, *, c: str = 'x') -> str:
-    return c * a
-
-
-# ok
-z: str = baz(1, c='hello')
--- a/tests/mypy/modules/fail4.py
+++ b/tests/mypy/modules/fail4.py
@@ -24,3 +24,13 @@ def bar() -> str:
 
 # return type should be a string
 y: int = bar()
+
+
+# nested decorator should not produce an error
+@validate_arguments(config={'arbitrary_types_allowed': True})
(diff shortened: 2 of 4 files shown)


tests/mypy/modules/fail4.py
```diff
@@ -24,3 +24,13 @@ def bar() -> str:
 
 # return type should be a string
 y: int = bar()
+
+
+# nested decorator should not produce an error
+@validate_arguments(config={'arbitrary_types_allowed': True})
```
PrettyWood, line 30: Why not put it in `success.py` instead of `fail4.py`?

changes/2055-layday.md
```diff
@@ -0,0 +1 @@
+add two overload variants to `validate_arguments` for the nested decorator signature
```
PrettyWood, line 1: Maybe something more user oriented like "improve/fix typing of `validate_arguments` to be compliant with `mypy`"

PrettyWood (commented):
Other than that looks good 👍

changes/2055-layday.md
layday: Changed to:

> fix annotation of `validate_arguments` when passing configuration as argument

tests/mypy/modules/fail4.py
layday: done

layday:
Unrelatedly, I like how all I've got to do is run `make` locally before pushing - I like this better than the pre-commit workflow adopted by Python projects, but I suppose that's just personal preference.

PrettyWood (approved):
Thanks!

PR #2057, opened 2020-10-27
Refactor PrivateAttr to type-check like Field

### LINKED ISSUE #2048: PrivateAttr causes failing type check
### Checks

# Bug

```
        pydantic version: 1.7
       pydantic compiled: True
            install path: /Users/ahedges/.pyenv/versions/3.7.8/envs/sdf/lib/python3.7/site-packages/pydantic
          python version: 3.7.8 (default, Sep 16 2020, 18:33:23)  [Clang 11.0.3 (clang-1103.0.32.59)]
```

I created a file with the following code:

```py
from pydantic import BaseModel, Field, PrivateAttr

class TestObject(BaseModel):
    public_field: str = Field()
    _private_field: str = PrivateAttr()
```

I then ran mypy (version 0.790) using the pydantic plugin. I received the following type error:

```
test.py:5: error: Incompatible types in assignment (expression has type "PrivateAttr", variable has type "str")
```

`PrivateAttr` should likely not cause a type error here, similarly to how `Field` works.

aphedges:
## Change Summary

- Fix mypy assignment error when using `PrivateAttr`
  - `PrivateAttr` is renamed `ModelPrivateAttr` and modeled after `ModelField`
  - A `PrivateAttr` function now exists, modeled after `Field`
- Documentation and error message are now more consistent between `PrivateAttr` and `Field`
- Made changes to typing where needed

I did not add any unit tests because I'm unsure how to write a unit test for preventing mypy errors.

## Related issue number

Fixes #2048.

## Checklist

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -799,21 +799,10 @@ class ModelField(Representation):
         return args
 
 
-class PrivateAttr(Representation):
-    """
-    Indicates that attribute is only used internally and never mixed with regular fields.
-
-    Types or values of private attrs are not checked by pydantic and it's up to you to keep them relevant.
-
-    Private attrs are stored in model __slots__.
-    """
-
+class ModelPrivateAttr(Representation):
     __slots__ = ('default', 'default_factory')
 
     def __init__(self, default: Any = Undefined, *, default_factory: Optional[NoArgAnyCallable] = None) -> None:
-        if default is not Undefined and default_factory is not None:
-            raise TypeError('default and default_factory args can not be used together')
-
         self.default = default
         self.default_factory = default_factory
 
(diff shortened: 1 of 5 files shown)


pydantic/fields.py
```diff
@@ -825,3 +814,28 @@ def __eq__(self, other: Any) -> bool:
...
+    Types or values of private attrs are not checked by pydantic and it's up to you to keep them relevant.
+
+    Private attrs are stored in model __slots__.
+
+    :param default: the attribute’s default value
+    :param default_factory: callable that will be called when a default value is needed for this attribute
+      If both `default` and `default_factory` are set, an error is raised.
+    """
+    if default is not Undefined and default_factory is not None:
+        raise TypeError('cannot specify both default and default_factory')
```
PrettyWood, line 836:
```suggestion
        raise ValueError('cannot specify both default and default_factory')
```
to be consistent with `ModelField`

PrettyWood (commented):
Nice! 👍 
To make sure it works as expected and fixes the issue you should add an example with `PrivateAttr` in `mypy/modules/success.py`

pydantic/fields.py
aphedges: Good catch! I have committed this change.

aphedges:
Thanks for the feedback! I have added a `PrivateAttr` example to `mypy/modules/success.py`.

PrettyWood (approved):
LGTM! Thanks a lot

pydantic/fields.py
samuelcolvin: Humm, not sure I agree with this. `TypeError` is what gets raised when you call a function with the wrong arguments. I think better to be correct than consistent here (we could change `ModelField` in v2).

But I don't mind that much.

samuelcolvin:
@MrMrRobat you implemented `PrivateAttr` originally, are you happy with this change?

MrMrRobat:
Sure!

pydantic/fields.py
PrettyWood: Ah yes you're right so it was my mistake when I raised the `ValueError` in `ModelField` oops 😬

samuelcolvin: Let's keep this for now and fix in both places later #2059

PR #2071, opened 2020-10-29
Update type annotation for create_model function

uriyyo:
## Update type annotation for create_model function

Updated `create_model` return type annotation to return type which inherits from `__base__` argument.

For instance, before this PR mypy will fail on the code below:
```python

class Point(BaseModel):
    x: int
    y: int

NewPoint = create_model('NewPoint', __base__=Point)

point = NewPoint.parse_obj(...)
point.x = 10  # mypy will fail here
```

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -870,11 +870,11 @@ def create_model(
     __model_name: str,
     *,
     __config__: Type[BaseConfig] = None,
-    __base__: Type[BaseModel] = None,
+    __base__: Type['Model'] = None,
     __module__: str = __name__,
     __validators__: Dict[str, classmethod] = None,
     **field_definitions: Any,
-) -> Type[BaseModel]:
+) -> Type['Model']:
     """
     Dynamically create a model.
     :param __model_name: name of the created model
@@ -887,7 +887,9 @@ def create_model(
         `foobar=(str, ...)` or `foobar=123`, or, for complex use-cases, in the format
         `<name>=<FieldInfo>`, e.g. `foo=Field(default_factory=datetime.utcnow, alias='bar')`
     """
-    if __base__:
+    __base__ = cast(Type['Model'], __base__)  # to make mypy happy
+
+    if __base__ is not None:
         if __config__ is not None:
             raise ConfigError('to avoid confusion __config__ and __base__ cannot be used together')
     else:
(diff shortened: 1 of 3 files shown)


pydantic/main.py
```diff
@@ -887,7 +887,9 @@ def create_model(
         `foobar=(str, ...)` or `foobar=123`, or, for complex use-cases, in the format
         `<name>=<FieldInfo>`, e.g. `foo=Field(default_factory=datetime.utcnow, alias='bar')`
     """
-    if __base__:
+    __base__ = cast(Type['Model'], __base__)  # to make mypy happy
+
+    if __base__ is not None:
```
PrettyWood, line 892: I don't understand why you need this. I thought only setting `'Model'` instead of `BaseModel` in the signature would be enough

uriyyo: Without this cast `mypy` will fail with errors:
```
pydantic/main.py:896: error: Incompatible types in assignment (expression has type "Type[BaseModel]", variable has type "Optional[Type[Model]]")  [assignment]
pydantic/main.py:927: error: Argument 2 to "type" has incompatible type "Tuple[Optional[Type[Model]]]"; expected "Tuple[type, ...]"  [arg-type]
```

As a solution for this problem, I decided to cast `__base__` to `Type['Model']` because it always will be of this type.

PrettyWood: Can you please give a working example ? I just tried to run your test without the cast and it works as expected. I'm probably missing something

uriyyo: When I tried to run `mypy` stage without `cast` it fails. Logs of `mypy`:
```
$ make mypy
mypy pydantic
pydantic/main.py:896: error: Incompatible types in assignment (expression has type "Type[BaseModel]", variable has type "Optional[Type[Model]]")  [assignment]
pydantic/main.py:927: error: Argument 2 to "type" has incompatible type "Tuple[Optional[Type[Model]]]"; expected "Tuple[type, ...]"  [arg-type]
Found 2 errors in 1 file (checked 23 source files)
make: *** [mypy] Error 1
```

PrettyWood: Ah right! Then I would simply write
```python
    if __base__ is not None:
        if __config__ is not None:
            raise ConfigError('to avoid confusion __config__ and __base__ cannot be used together')
    else:
        __base__ = cast(Type['Model'], BaseModel)
```
as the cast is important only when we set `__bases__`. WDYT ?

uriyyo: Thanks, you are totally right.

PrettyWood (approved):
LGTM thanks !

samuelcolvin:
this looks good. As far as I'm aware it shouldn't be included in v1.7.2 so I won't merge it yet.

let me know if I'm missing something.

samuelcolvin:
thanks a lot.

PR #2075, opened 2020-10-30
fix: keep the order of the fields when `validate_assignment` is set

### LINKED ISSUE #2073: `validate_assignment` changes the field order
### Checks

* [*] I added a descriptive title to this issue
* [*] I have searched (google, github) for similar issues and couldn't find anything
* [*] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

```
             pydantic version: 1.7.1
            pydantic compiled: True
                 install path: /home/user/work/bugs/python/pydantic-ordering/.venv/lib/python3.8/site-packages/pydantic
                     platform: Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.29
     optional deps. installed: []
```

Including `validate_assignment = True` in a model's config seems to cause assignment to change the field order when using `dict()`. This change was discovered when upgrading to 1.7.0 and also applies to 1.7.1, but was not present in prior versions.

This behaviour does not seem to agree with <https://pydantic-docs.helpmanual.io/usage/models/#field-ordering>, and is something that is important for my uses (where it generates massive diffs in the output) and surely others.

Running this snippet with and without the `validate_assignment` line produces different output:
```py

class Model(BaseModel):
    title: str = ''
    name: str = ''

    class Config:
        validate_assignment = True
        ...

m = Model(name='name')
m.title = 'title'
print(m.dict())
```

PrettyWood:
## Change Summary
Following #2000 and #2046 we can't `pop` the current value when assigning
a new one (which was probably the most efficient) as we want to keep the
order in the `__dict__`.
So let's do a shallow copy of the `__dict__` without the current value

## Related issue number
fix #2073

## Checklist

### DIFF
--- a/changes/2073-PrettyWood.md
+++ b/changes/2073-PrettyWood.md
@@ -1,0 +1,1 @@
+fix: keep the order of the fields when `validate_assignment` is set
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -384,14 +384,14 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
 
             known_field = self.__fields__.get(name, None)
             if known_field:
-                original_value = self.__dict__.pop(name)
-                try:
-                    value, error_ = known_field.validate(value, self.__dict__, loc=name, cls=self.__class__)
-                    if error_:
-                        raise ValidationError([error_], self.__class__)
-                except Exception:
-                    self.__dict__[name] = original_value
-                    raise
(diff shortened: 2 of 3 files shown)


PrettyWood:
@samuelcolvin it should probably added in v1.7.2 too

samuelcolvin:
🙏

PR #2092, opened 2020-11-04
feat: add `Config.smart_union` option

### LINKED ISSUE #453: When doing a union of types with the same contract, items are converted inappropriately
Example: 
```
from typing import Union
import pydantic

class A(pydantic.BaseModel):
   x: str

class B(pydantic.BaseModel):
   x: str

   attr: Union[B, A]

m = MyModel(attr=A(x="apples"))
print(m)
m = MyModel(attr=B(x="apples"))
print(m)
```

This prints: 
```
MyModel attr=<B x='apples'>
MyModel attr=<B x='apples'>
```

Expected output: 
```
MyModel attr=<A x='apples'>
MyModel attr=<B x='apples'>
```

---

**Workaround**: One workaround is to ensure that your classes do not have equivalent signatures.  Changing to this will avoid the issue: 

```
class A(pydantic.BaseModel):
   a_x: str

class B(pydantic.BaseModel):
   b_x: str
```

---

System Info: 

* OS: Linux
* Python version `3.7.1`
* Pydantic version `0.21`

### LINKED ISSUE #1423: More sophisticated validation for Unions to avoid coercion when not needed

### LINKED ISSUE #3196: Parse error when nested objects include a possible Union
### Checks

# Bug
I'm using pydantic version 1.8.2 with Python 3.9.2 on linux.

I'm trying to wrap an existing API with Pydantic models and I ran into a situation where a Field may take either an object that conforms to one of two Models or a list of objects that conform to either of those models, plus some other metadata. In certain situations, the object does not get parsed properly. Here's a mockup/simplification of the Models associated with this issue:
```py
from typing import Union, Optional


class MetadataModel(BaseModel):
    label: Optional[str] = Field()
    description: Optional[str] = Field()


class ModelA(MetadataModel):
    type: str = Field(...)
    value: str = Field(...)


class ModelB(MetadataModel):
    type: str = Field(...)
    attributes: Optional[dict] = Field()


class MainModel(BaseModel):
    children: Union[ModelA, ModelB, list[Union[ModelA, ModelB]]] = Field()
```

If I run the following six test cases, the last two produce something odd.

```py
# Case 1: One child of type ModelA works
print(
    repr(
        MainModel.parse_obj(
            {
                "children": {
                    "type": "A",
                    "label": "Object of type ModelA",
                    "value": "some string",
                }
            }
        )
    ),
)
# MainModel(children=ModelA(label='Object of type ModelA', description=None, type='A', value='some string'))

# Case 2: A list of one child of type ModelA works
print(
    repr(
        MainModel.parse_obj(
            {
                "children": [
                    {
                        "type": "A",
                        "label": "Object of type ModelA",
                        "value": "some string",
                    }
                ]
            }
        )
    ),
)
# MainModel(children=[ModelA(label='Object of type ModelA', description=None, type='A', value='some string')])

# Case 3: A list of two children of type ModelA works
print(
    repr(
        MainModel.parse_obj(
            {
                "children": [
                    {
                        "type": "A",
                        "label": "Object of type ModelA",
                        "value": "some string",
                    },
                    {
                        "type": "A",
                        "label": "Another object of type ModelA",
                        "value": "some string",
                    },
                ]
            }
        )
    ),
)
# MainModel(children=[ModelA(label='Object of type ModelA', description=None, type='A', value='some string'), ModelA(label='Another object of type ModelA', description=None, type='A', value='some string')])

# Case 4: One child of type ModelB works
print(
    repr(
        MainModel.parse_obj(
            {"children": {"type": "B", "label": "Object of type ModelB"}}
        )
    ),
)
# MainModel(children=ModelB(label='Object of type ModelB', description=None, type='B', attributes=None))

# Case 5: A list of one child of type ModelB DOES NOT work
print(
    repr(
        MainModel.parse_obj(
            {"children": [{"type": "B", "label": "Object of type ModelB"}]}
        )
    ),
)
# The list is misinterpreted and the "label" key became a value
# MainModel(children=ModelB(label=None, description=None, type='label', attributes=None))


# Case 6: A list of two children of type ModelB DOES NOT work
print(
    repr(
        MainModel.parse_obj(
            {
                "children": [
                    {"type": "B", "label": "Object of type ModelB"},
                    {"type": "B", "label": "Another object of type ModelB"},
                ]
            }
        )
    ),
)
# Same error as above
# MainModel(children=ModelB(label=None, description=None, type='label', attributes=None))
```
I found that if I were to reorder the type annotations, the issue (at least this specific issue) goes away. If I redefine `MainModel` like this:
```py
class MainModel(BaseModel):
    children: Union[list[Union[ModelA, ModelB]], ModelA, ModelB] = Field()
```
Then run the test cases above again, the first four work correctly still and the second two work and produce the following output:
```py
# Case 5: Correct
# MainModel(children=[ModelB(label='Object of type ModelB', description=None, type='B', attributes=None)])

# Case 6: Correct
# MainModel(children=[ModelB(label='Object of type ModelB', description=None, type='B', attributes=None), ModelB(label='Another object of type ModelB', description=None, type='B', attributes=None)])
```
I scanned through the documentation and I didn't find anything that seemed to directly impact this scenario. This issue seems problematic and reordering the type annotations seems like a fragile fix. I'm new to pydantic, so perhaps there are best practices that I am not aware of that would prevent this problem from happening or perhaps this is a actually a feature needed for something else.

### LINKED ISSUE #3365: Wrong assignement of values in `Union[Foo, List[Foo]]`
### Checks

# Bug

When validating a model that allows either a single object `Foo` or a list of `Foo` objects, the wrong values are assigned. The key of the second parameter is assigned as value of the first parameter (which is a string).

```
             pydantic version: 1.8.2
            pydantic compiled: True
                 install path: /home/anaconda3/envs/py38/lib/python3.8/site-packages/pydantic
               python version: 3.8.5 (default, Sep  4 2020, 07:30:14)  [GCC 7.3.0]
                     platform: Linux-5.11.0-38-generic-x86_64-with-glibc2.10
```

The following minimal example repruduces the bug: 

```py
from typing import List, Optional, Union


class Foo(BaseModel):
    a: str = Field(
        ...,
    )
    b: Optional[float]


class Bar(BaseModel):
    foo: Union[Foo, List[Foo]]


bar = {
    "foo": [
        {
            "a": "some string",
            "b": 1.0
        },
        {
            "a": "another string",
            "b": 2.0
        }
    ]
}

a = Bar(**bar)

print(a)
```

- Output: `foo=Foo(a='b', b=None)`
- Expected Output: `foo=[Foo(a='some string', b=1.0), Foo(a='another string', b=2.0)]`

I am not sure if this is related somehow to  #2092

PrettyWood:
## Change Summary
Following [this comment](https://github.com/samuelcolvin/pydantic/pull/1951#issuecomment-716186527), here is a proposal for a new option `smart_union` that changes current behaviour of `Union` to avoid coercion if possible

```python
from typing import Dict, List, Union

class Model(BaseModel, smart_union=True):
    v: Union[int, bool, str]

assert Model(v=1).dict() == {'v': 1}
assert Model(v=True).dict() == {'v': True}
assert Model(v='1').dict() == {'v': '1'}
```

@samuelcolvin I'll update the doc if you're fine with it

## Related issue number
A bunch of closed ones 😄 
closes #453
closes #1423
closes #3196 
closes #3365

## Checklist


pydantic/mypy.py
```diff
@@ -95,11 +102,22 @@ def get_class_decorator_hook(self, fullname: str) -> Optional[Callable[[ClassDef
...
+            return _convert_strict_union_type
+        return None
+
     def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
         transformer = PydanticModelTransformer(ctx, self.plugin_config)
         transformer.transform()
 
 
+def _convert_strict_union_type(type_context: AnalyzeTypeContext) -> Type:
+    """Convert StrictUnion[...] type to Union[...]"""
```
samuelcolvin, line 116:
```suggestion
    """
    Convert StrictUnion[...] type to Union[...]
    """
```

pydantic/types.py
```diff
@@ -872,3 +873,31 @@ def to(self, unit: str) -> float:
             raise errors.InvalidByteSizeUnit(unit=unit)
 
         return self / unit_div
+
+
+if TYPE_CHECKING:
+    StrictUnion = Union
+else:
+
+    class StrictUnionMeta(type):
```
samuelcolvin, line 882: I guess this doesn't need to go inside the `if not TYPE_CHECKING`?

tests/mypy/outputs/plugin_success.txt
```diff
@@ -1,3 +1,5 @@
 121: error: Unexpected keyword argument "name" for "AddProject"  [call-arg]
 121: error: Unexpected keyword argument "slug" for "AddProject"  [call-arg]
-121: error: Unexpected keyword argument "description" for "AddProject"  [call-arg]
\ No newline at end of file
+121: error: Unexpected keyword argument "description" for "AddProject"  [call-arg]
+139: error: Variable "pydantic.types.StrictUnion" is not valid as a type  [valid-type]
+139: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
```
samuelcolvin, line 5: we should rename this file to something less confusing (I know i renamed it before to this name 🙈).

Perhaps 'no_plugin_errors.txt` or something?

samuelcolvin (commented):
Humm, this looks good.

I just wonder if we should go the whole hog and implement `SmartUnion` which does this then tries sensible next steps.

Having:
* `Union`
* `StrictUnion`
* `LaxUnion`
* `SmartUnion`
* `DecriminatedUnion`

Isn't very zen of python

> There should be one-- and preferably only one --obvious way to do it.

I know we can't avoid having multiple, but I think we should aim at one type of union which is useful in as many cases as possible -- the behaviour we'll adopt for `Union` if v2.

For me that would be `SmartUnion` which does this and then falls back to the current union behaviour or something smarter.

See the discussion on #1423, e.g. https://github.com/samuelcolvin/pydantic/issues/1423#issuecomment-618962827

marco-neumann-by:
While trying out `StrictUnion`, I stumbled across the issue that `PikaModel.schema()` crashes because pydantic doesn't know what to do with the newly created `StrictUnion` type. Could you please add a test for that? This should probably just create the very same schema as a "normal" `Union`.

tests/mypy/outputs/plugin_success.txt
PrettyWood: https://github.com/samuelcolvin/pydantic/pull/2092/commits/76ae8dfd55510bc2eb4b94354ba07b8eb786679a

pydantic/types.py
PrettyWood: I guess not

pydantic/mypy.py
PrettyWood: https://github.com/samuelcolvin/pydantic/pull/2092/commits/efd803ec6b89832411a7d2b1f2f3a4f4ad8431ee

tests/mypy/modules/plugin_success.py
```diff
@@ -139,3 +139,10 @@ class Model(BaseModel):
 
 dynamic_model = DynamicModel(x=1, y='y')
 dynamic_model.x = 2
+
+
+class PikaModel(BaseModel):
+    x: StrictUnion[int, bool, str]
```
samuelcolvin, line 145: `bool` is a sub-type of `int` so is not required here.

samuelcolvin (commented):
otherwise looks good, but what did you think about my suggestion above of `SmartUnion`?

PrettyWood:
> otherwise looks good, but what did you think about my suggestion above of `SmartUnion`?

I agree but I want to take more time to do it properly. And I also need to support schema like a regular union

samuelcolvin:
Okay, so do you think we should merge this or wait for those improvements?

PrettyWood:
I reckon we should wait. I'll try to work on it soon but I need to put everything in perspective and add some code + tests for schema, mypy...

samuelcolvin:
👍

PrettyWood:
@samuelcolvin Some news regarding this issue!!
I'm not really happy with the current approach with a custom `StrictUnion` class. Since we can't inherit from Union or make a Generic with many types, I add to go with this but it forces me to support the schema generation, the mypy plugin and in the IDE it creates a lot of issues.

I reckon the best approach is to keep the `Union` as is and use a config flag to make it "smart" (flag that can be turned to True by default or even removed in v2). I don't think people will mix Strict and not Strict like I did in my tests. The "`SmartUnion`" should be enough for all cases IMO.

Here is my question: we want `["ab"]` to stay untouched with type `Union[Dict[str, str], List[str]]` where currently _pydantic_ will coerce it into `{"a": "b"}`. Since `isinstance` check works well only with basic types like `int`, `bool` but not with `typing` types, I created a package [typingx](https://github.com/PrettyWood/typingx) a week ago (it was made for testing purpose at the beginning) where I implemented a `isinstance`-like method but for most `typing` types too.

Here is a POC that works well with it: https://github.com/PrettyWood/pydantic/pull/66/files.
I tested it with all the closed issues regarding this matter (~20 issues) and it works perfectly.

But I don't want to add a library (even more when it's mine) randomly. I'll wait for your input on this matter

samuelcolvin:
This is looking awesome, 🎉 I really want some of this functionality.

I'm a bit worried about adding another dependency, particularly because:
* typingx duplicates a lot of the logic in `fields.py`
* `isinstancex` is only used once (albeit in an important/commonly used place)
* using `smart_union` doesn't "feel" like something that should require an extra dependency - it feels like a core part of pydantic
* Although this is optional at the moment, it will become required in the next major release
* This breaks the current logic that `Unions` (like `List`, `Dict` etc.) are implemented as a recursive group of fields which are queried in turn.

I know this is really annoying, but I would prefer to avoid the extra dependency if at all possible?

I would prefer to skip the second pass here or find another route. I'm welcome to input from others though...?

I think overall, the second pass will be less important as pydantic becomes stricter in v2.

PrettyWood:
@samuelcolvin to be honest adding the extra library was a quick win to test it. We can definitely just copy paste part of the logic directly in pydantic now or in v2 :)
I also don't want to add extra dependencies. I love libraries that don't come with a bunch of extra, especially when it's for a "core" functionality

samuelcolvin:
@PrettyWood makes sense.

PrettyWood:
@samuelcolvin Ok I went simple and added a basic step 2. The documentation should be quite explicit.
The logic of `typingx` will be added more or less when a strict mode is added in _pydantic_. At this moment, I think it will make sense to improve further the step 2!
I think it's already a great improvement for most cases. Please review and tell me what you think :)

small-pants:
anyone know the status on this? @PrettyWood are you still making changes?

blazing-gig:
Hey @PrettyWood , any reason why it's still not merged ? This is a pretty serious issue.

small-pants:
@samuelcolvin checking in again on the status of this

PrettyWood:
We'll try to make a v1.9 in September. Samuel is currently under water with both private and professional life, which are more important than OSS
And there are so many PRs to review, not only this one.
So please be patient :) 
In the meantime you're more than welcome to review PRs, try them out in local or in your repo to ensure the implementation suits your needs.

If we can't make a 1.9 next month, I'll try to push a 1.9-alpha with the most awaited features.

docs/usage/model_config.md
```diff
@@ -164,3 +167,28 @@ For example:
 {!.tmp_examples/model_config_alias_precedence.py!}
 ```
+
+## Smart Union
+
+By default, if you set a `Union` type for a field, _pydantic_ tries to validate (and coerce if it can) in the order of the union,
+which can be unexpected in some cases
```
djpugh, line 174: Is it worth adding the warning here from #2839, because python treats unions as singletons so the first definition anywhere in the import chain is what is used and makes that behaviour even more unexpected
so something like

```suggestion

By default, if you set a `Union` type for a field, _pydantic_ tries to validate (and coerce if it can) in the order of the union.

!!! warning
    `typing.Unions` also ignore order when they are [defined](https://docs.python.org/3/library/typing.html#typing.Union),
    so `Union[int, float] == Union[float, int]` which can lead to unexpected behaviour when combined with matching based on the `Union` type order.
    Please note that this can also be affected by third party libraries and their internal type definitions and the import orders.
```

PrettyWood:
@djpugh I updated the doc and moved your section with default `Union` (not smart union) https://smokeshow.helpmanual.io/1q331m1k3x6v06512i6z/usage/types/#unions
Feedback welcome :)
@samuelcolvin I think it's ready for review

docs/usage/types.md
```diff
@@ -246,9 +246,20 @@ the first one that matches. In the above example the `id` of `user_03` was defin
...
+    so `Union[int, float] == Union[float, int]` which can lead to unexpected behaviour
+    when combined with matching based on the `Union` type order.
+    Please note that this can also be [affected by third party libraries](https://github.com/samuelcolvin/pydantic/issues/2835)
+    and their internal type definitions and the import orders.
+
 As such, it is recommended that, when defining `Union` annotations, the most specific type is included first and
-followed by less specific types. In the above example, the `UUID` class should precede the `int` and `str`
-classes to preclude the unexpected representation as such:
+followed by less specific types.
+
```
djpugh, line 258: @PrettyWood - Suggest maybe flipping these two parts for now, as it comes across slightly confusing with the warning then the recommended approach (given the warning applies to the recommended approach), and then maybe refer to Smart Union as the recommended approach to resolve this?

docs/usage/types.md
```diff
@@ -246,9 +246,20 @@ the first one that matches. In the above example the `id` of `user_03` was defin
 is defined under the attribute's `Union` annotation) but as the `uuid.UUID` can be marshalled into an `int` it
 chose to match against the `int` type and disregarded the other types.
 
+!!! warning
+    `typing.Union` also ignores order when [defined](https://docs.python.org/3/library/typing.html#typing.Union),
+    so `Union[int, float] == Union[float, int]` which can lead to unexpected behaviour
+    when combined with matching based on the `Union` type order.
```
djpugh, line 252:
```suggestion
    when combined with matching based on the `Union` type order inside other type definitions, such as `List` and `Dict` types (because python treats these definitions as singletons).
 For example, `Dict[str, Union[int, float]] == Dict[str, Union[float, int]]` with the order based on the first time it was defined.
```
Just went back to #2835, the issue fully manifests as typing doesn't re-initialise Dict and List type definitions because of this so wanted to make it a bit clearer.

docs/usage/types.md
PrettyWood: I went with https://github.com/samuelcolvin/pydantic/pull/2092/commits/b3de1a3546fd9daf8caa0e88aa9b5a9151d36f66

pydantic/fields.py
```diff
@@ -935,6 +935,35 @@ def _validate_singleton(
     ) -> 'ValidateReturn':
         if self.sub_fields:
             errors = []
+
+            if is_union_origin(get_origin(self.type_)) and self.model_config.smart_union:
```
samuelcolvin, line 939:
```suggestion
            if self.model_config.smart_union and is_union_origin(get_origin(self.type_)):
```
Should be quicker since it's just checking a bool, if that is false we skip the `is_union_origin` check.

samuelcolvin (commented):
Otherwise LGTM.

@PrettyWood I know this has been a lonnnnng time, so I can make the change here to get this merged, but I want to leave it for you to check you're happy with it even though it's a pretty trival change.

docs/usage/model_config.md
```diff
@@ -164,3 +167,31 @@ For example:
...
+Know that this is of course slower, especially if your `Union` is quite big.
+
+```py
+{!.tmp_examples/model_config_smart_union_on.py!}
+```
+_(This script is complete, it should run "as is")_
+
+!!! warning
+
+Note that this option **does not support compound types yet** (e.g. differentiate `List[int]` and `List[str]`).
```
samuelcolvin, line 191: I think this is supposed to be the warning text, so should be indented, see [here](https://smokeshow.helpmanual.io/373p4j603c2v6f615z0y/usage/model_config/#smart-union):

image

PrettyWood:
@samuelcolvin updated! Please review 🙏

samuelcolvin:
:tada: thank you so much.

andriilahuta:
@PrettyWood I've been trying this in master and it crashes when `TypedDict` is involved:
```py
class Dict1(TypedDict):
    foo: str


class Dict2(TypedDict):
    bar: str


class M(BaseModel):
    d: Union[Dict2, Dict1]

    class Config:
        smart_union = True


M(d=dict(foo='baz'))  # raises TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

PR #2093, opened 2020-11-05
Add "rediss" scheme.

nuno-andre:
It's the [registered scheme](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) for Redis over TLS.


PrettyWood:
Hi @nuno-andre 
there is already an opened PR #1911

nuno-andre:
Sorry, I didn't search for before. I'm closing this PR.

PR #2097, opened 2020-11-07
Add a Hypothesis plugin

### LINKED ISSUE #2017: Generating Pydantic-specific types with Hypothesis
:wave: I'm a Hypothesis core developer, and I know there are lots of people who like generating Pydantic objects to test their code with Hypothesis.  Automatic generation already works really well for builtin and `typing` generic types - see e.g. https://github.com/samuelcolvin/pydantic/issues/1666, https://github.com/timothycrosley/hypothesis-auto/issues/10, https://github.com/HypothesisWorks/hypothesis/pull/2388, https://github.com/HypothesisWorks/hypothesis/pull/2580, and https://github.com/HypothesisWorks/hypothesis/pull/2640 (cc @MrMrRobat)

I'm proposing that Pydantic adds [a setuptools entry point](https://hypothesis.readthedocs.io/en/latest/strategies.html#registering-strategies-via-setuptools-entry-points) to [register custom strategies](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.register_type_strategy) for [Pydantic-specific or constrained types](https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types).  This is a nice way to offer optional integrations, because *Hypothesis* will load it at import-time without adding any runtime dependency or overhead to Pydantic.  Example code:

```python
# In `setup.py`
setup(
    ...,
    entry_points={"hypothesis": ["_ = pydantic._hypothesis_setup"]},
)

# In `pydantic._hypothesis_setup.py`
import pydantic
import hypothesis.strategies as st

st.register_type_strategy(pydantic.EmailStr, st.emails())
st.register_type_strategy(pydantic.UUID1, st.uuids(version=1))
...
```

Constrained types are a little trickier, since e.g. `confloat(ge=0, lt=1)` dynamically creates a subclass of `ConstrainedFloat`- we'll have to register a function *for* the parent class, which takes the child class and returns the strategy (e.g. `st.floats(min_value=0, max_value=1, exclude_max=True)`).  Fortunately this already works, because it's how we support parametrised generic types!

I'm happy to make a PR if @samuelcolvin or @PrettyWood think this is a good idea, though it might take me a while :slightly_smiling_face:

Zac-HD:
This patch adds a plugin which teaches Hypothesis how to generated examples of Pydantic's custom field types (which closes #2017).  Note that there is no runtime impact or dependency; this module is imported *by Hypothesis* so it only exists at test-time and only when Hypothesis is both installed and actually being used.

--------------

I've ended up dropping the integration for URLs, and for constrained lists and sets... but everything else is ready to go!

<details>
<summary>Why not support constrained lists and sets?</summary>

In short, I couldn't get them to play nicely with Hypothesis' type registry - because the runtime objects are subclasses of a parametrized generic type, and aren't always distinguished by our introspection logic 😭 
</details>

<details>
<summary>Why I decided to leave our URLs (with code)</summary>

I spent a long time tweaking support for URLs, but ultimately gave it up: it's easy to generate valid URLs, *unless* you want to generally really strange ones which will find lots of bugs.  I decided that it was better to make users register their own - and make the tradeoff explicit - than to cover up the complexity and have automatic but weak tests.  This is a standard design choice for Hypothesis, sadly, so I've included my code so far below for the benefit of any future contributor who wants to pick it up.

```python
def idna_encodable(s: str) -> bool:
    # We only need this because the regex patterns aren't fully precise; but
    # rejection sampling is a LOT easier to implement than precise patterns.
    try:
        s.encode('idna')
    except Exception:
        hypothesis.reject()  # type: ignore[no-untyped-call]
    return True


@resolves(pydantic.AnyUrl)
def resolve_anyurl(cls):  # type: ignore[no-untyped-def]
    domains = st.one_of(
        st.from_regex(ascii_domain_regex(), fullmatch=True),
        st.from_regex(int_domain_regex(), fullmatch=True).filter(idna_encodable),
    )
    if cls.tld_required:

        def has_tld(s: str) -> bool:
            assert isinstance(s, str)
            match = ascii_domain_regex().fullmatch(s) or int_domain_regex().fullmatch(s)
            return bool(match and match.group('tld'))

        hosts = domains.filter(has_tld)
    else:
        hosts = domains | st.from_regex(
            r'(?P<ipv4>(?:\d{1,3}\.){3}\d{1,3})' r'|(?P<ipv6>\[[A-F0-9]*:[A-F0-9:]+\])',
            fullmatch=True,
        )

    return st.builds(
        cls.build,
        scheme=(
            st.sampled_from(sorted(cls.allowed_schemes))
            if cls.allowed_schemes
            else st.from_regex(r'(?P<scheme>[a-z][a-z0-9+\-.]+)', fullmatch=True)
        ).filter(idna_encodable),
        user=st.one_of(
            st.nothing() if cls.user_required else st.none(),
            st.from_regex(r'(?P<user>[^\s:/]+)', fullmatch=True).filter(idna_encodable),
        ),
        password=st.none() | st.from_regex(r'(?P<password>[^\s/]*)', fullmatch=True).filter(idna_encodable),
        host=hosts,
        port=st.none() | st.integers(0, 2 ** 16 - 1).map(str),
        path=st.none() | st.from_regex(r'(?P<path>/[^\s?]*)', fullmatch=True).filter(idna_encodable),
        query=st.none() | st.from_regex(r'(?P<query>[^\s#]+)', fullmatch=True).filter(idna_encodable),
        fragment=st.none() | st.from_regex(r'(?P<fragment>\S+)', fullmatch=True).filter(idna_encodable),
    ).filter(lambda url: cls.min_length <= len(url) <= cls.max_length)


st.register_type_strategy(pydantic.AnyUrl, resolve_anyurl)
st.register_type_strategy(pydantic.AnyHttpUrl, resolve_anyurl)
st.register_type_strategy(pydantic.HttpUrl, resolve_anyurl)
st.register_type_strategy(pydantic.PostgresDsn, resolve_anyurl)
st.register_type_strategy(pydantic.RedisDsn, resolve_anyurl)
```
```python
def gen_url_models():
    class AnyUrlModel(pydantic.BaseModel):
        anyurl: pydantic.AnyUrl

    class AnyHttpUrlModel(pydantic.BaseModel):
        anyhttp: pydantic.AnyHttpUrl

    class HttpUrlModel(pydantic.BaseModel):
        http: pydantic.HttpUrl

    class PostgresDsnModel(pydantic.BaseModel):
        postgres: pydantic.PostgresDsn

    class RedisDsnModel(pydantic.BaseModel):
        redis: pydantic.RedisDsn

    yield from (AnyUrlModel, AnyHttpUrlModel, HttpUrlModel, PostgresDsnModel, RedisDsnModel)


@pytest.mark.parametrize('model', gen_url_models())
@settings(suppress_health_check=[HealthCheck.filter_too_much, HealthCheck.too_slow])
@given(data=st.data())
def test_can_construct_urls_model(data, model):
    # This is a separate test because we want a minimal health-check exemption
    instance = data.draw(st.from_type(model))
    assert isinstance(instance, model)
```

</details>

<details>
<summary>`Literal[None]` requires Python 3.7+ or a recent version of Hypothesis</summary>

Because [I literally just fixed that this evening](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-44-0) - since it requires a backport on a security-only Python version, we delayed working out how to support it until we knew it would actually be used before the 3.6 EOL later this year.

</details>

### DIFF
--- a/docs/examples/hypothesis_property_based_test.py
+++ b/docs/examples/hypothesis_property_based_test.py
@@ -1,0 +1,24 @@
+import typing
+from hypothesis import given, strategies as st
+from pydantic import BaseModel, EmailStr, PaymentCardNumber, PositiveFloat
+
+
+class Model(BaseModel):
+    card: PaymentCardNumber
+    price: PositiveFloat
+    users: typing.List[EmailStr]
+
+
+@given(st.builds(Model))
+def test_property(instance):
+    # Hypothesis calls this test function many times with varied Models,
+    # so you can write a test that should pass given *any* instance.
+    assert 0 < instance.price
+    assert all('@' in email for email in instance.users)
+
+
+@given(st.builds(Model, price=st.floats(100, 200)))
+def test_with_discount(instance):
+    # This test shows how you can override specific fields,
+    # and let Hypothesis fill in any you don't care about.
+    assert 100 <= instance.price <= 200
--- a/docs/hypothesis_plugin.md
+++ b/docs/hypothesis_plugin.md
@@ -1,0 +1,28 @@
+[Hypothesis](https://hypothesis.readthedocs.io/) is the Python library for
+[property-based testing](https://increment.com/testing/in-praise-of-property-based-testing/).
+Hypothesis can infer how to construct type-annotated classes, and supports builtin types,
+many standard library types, and generic types from the
+[`typing`](https://docs.python.org/3/library/typing.html) and
+[`typing_extensions`](https://pypi.org/project/typing-extensions/) modules by default.
+
+From Pydantic v1.8 and [Hypothesis v5.29.0](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-29-0),
+Hypothesis will automatically load support for [custom types](usage/types.md) like
+`PaymentCardNumber` and `PositiveFloat`, so that the
+[`st.builds()`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.builds)
+and [`st.from_type()`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.from_type)
+strategies support them without any user configuration.
+
+
+### Example tests
+```py
+{!.tmp_examples/hypothesis_property_based_test.py!}
+```
+_(This script is complete, it should run "as is")_
+
+
+### Use with JSON Schemas
+To test client-side code, you can use [`Model.schema()`](usage/models.md) with the
+[`hypothesis-jsonschema` package](https://pypi.org/project/hypothesis-jsonschema/)
+to generate arbitrary JSON instances matching the schema.
+For web API testing, [Schemathesis](https://schemathesis.readthedocs.io) provides
+a higher-level wrapper and can detect both errors and security vulnerabilities.
--- a/pydantic/_hypothesis_plugin.py
+++ b/pydantic/_hypothesis_plugin.py
@@ -1,0 +1,349 @@
+"""
+Register Hypothesis strategies for Pydantic custom types.
+
+This enables fully-automatic generation of test data for most Pydantic classes.
+
+Note that this module has *no* runtime impact on Pydantic itself; instead it
+is registered as a setuptools entry point and Hypothesis will import it if
+Pydantic is installed.  See also:
+
+https://hypothesis.readthedocs.io/en/latest/strategies.html#registering-strategies-via-setuptools-entry-points
+https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.register_type_strategy
+https://hypothesis.readthedocs.io/en/latest/strategies.html#interaction-with-pytest-cov
+https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types
+
+Note that because our motivation is to *improve user experience*, the strategies
+are always sound (never generate invalid data) but sacrifice completeness for
+maintainability (ie may be unable to generate some tricky but valid data).
+
+Finally, this module makes liberal use of `# type: ignore[<code>]` pragmas.
+This is because Hypothesis annotates `register_type_strategy()` with
+`(T, SearchStrategy[T])`, but in most cases we register e.g. `ConstrainedInt`
+to generate instances of the builtin `int` type which match the constraints.
+"""
+
+import contextlib
+import ipaddress
+import json
+import math
+from fractions import Fraction
+from typing import Callable, Dict, Type, Union, cast, overload
+
+import hypothesis.strategies as st
+
+import pydantic
+import pydantic.color
+import pydantic.types
+
+# FilePath and DirectoryPath are explicitly unsupported, as we'd have to create
+# them on-disk, and that's unsafe in general without being told *where* to do so.
+#
+# URLs are unsupported because it's easy for users to define their own strategy for
+# "normal" URLs, and hard for us to define a general strategy which includes "weird"
+# URLs but doesn't also have unpredictable performance problems.
+#
+# conlist() and conset() are unsupported for now, because the workarounds for
+# Cython and Hypothesis to handle parametrized generic types are incompatible.
+# Once Cython can support 'normal' generics we'll revisit this.
+
+# Emails
+try:
+    import email_validator
+except ImportError:  # pragma: no cover
+    pass
+else:
+
+    def is_valid_email(s: str) -> bool:
+        # Hypothesis' st.emails() occasionally generates emails like 0@A0--0.ac
+        # that are invalid according to email-validator, so we filter those out.
+        try:
+            email_validator.validate_email(s, check_deliverability=False)
+            return True
+        except email_validator.EmailNotValidError:
+            return False
+
+    # Note that these strategies deliberately stay away from any tricky Unicode
+    # or other encoding issues; we're just trying to generate *something* valid.
+    st.register_type_strategy(pydantic.EmailStr, st.emails().filter(is_valid_email))  # type: ignore[arg-type]
+    st.register_type_strategy(
+        pydantic.NameEmail,
+        st.builds(
+            '{} <{}>'.format,  # type: ignore[arg-type]
+            st.from_regex('[A-Za-z0-9_]+( [A-Za-z0-9_]+){0,5}', fullmatch=True),
+            st.emails().filter(is_valid_email),
+        ),
+    )
+
+# PyObject - dotted names, in this case taken from the math module.
+st.register_type_strategy(
+    pydantic.PyObject,
+    st.sampled_from(
+        [cast(pydantic.PyObject, f'math.{name}') for name in sorted(vars(math)) if not name.startswith('_')]
+    ),
+)
+
+# CSS3 Colors; as name, hex, rgb(a) tuples or strings, or hsl strings
+_color_regexes = (
+    '|'.join(
+        (
+            pydantic.color.r_hex_short,
+            pydantic.color.r_hex_long,
+            pydantic.color.r_rgb,
+            pydantic.color.r_rgba,
+            pydantic.color.r_hsl,
+            pydantic.color.r_hsla,
+        )
+    )
+    # Use more precise regex patterns to avoid value-out-of-range errors
+    .replace(pydantic.color._r_sl, r'(?:(\d\d?(?:\.\d+)?|100(?:\.0+)?)%)')
+    .replace(pydantic.color._r_alpha, r'(?:(0(?:\.\d+)?|1(?:\.0+)?|\.\d+|\d{1,2}%))')
+    .replace(pydantic.color._r_255, r'(?:((?:\d|\d\d|[01]\d\d|2[0-4]\d|25[0-4])(?:\.\d+)?|255(?:\.0+)?))')
+)
+st.register_type_strategy(
+    pydantic.color.Color,
+    st.one_of(
+        st.sampled_from(sorted(pydantic.color.COLORS_BY_NAME)),
+        st.tuples(
+            st.integers(0, 255),
+            st.integers(0, 255),
+            st.integers(0, 255),
+            st.none() | st.floats(0, 1) | st.floats(0, 100).map('{}%'.format),
+        ),
+        st.from_regex(_color_regexes, fullmatch=True),
+    ),
+)
+
+
+# Card numbers, valid according to the Luhn algorithm
+
+
+def add_luhn_digit(card_number: str) -> str:
+    # See https://en.wikipedia.org/wiki/Luhn_algorithm
+    for digit in '0123456789':
+        with contextlib.suppress(Exception):
+            pydantic.PaymentCardNumber.validate_luhn_check_digit(card_number + digit)
+            return card_number + digit
+    raise AssertionError('Unreachable')  # pragma: no cover
+
+
+card_patterns = (
+    # Note that these patterns omit the Luhn check digit; that's added by the function above
+    '4[0-9]{14}',  # Visa
+    '5[12345][0-9]{13}',  # Mastercard
+    '3[47][0-9]{12}',  # American Express
+    '[0-26-9][0-9]{10,17}',  # other (incomplete to avoid overlap)
+)
+st.register_type_strategy(
+    pydantic.PaymentCardNumber,
+    st.from_regex('|'.join(card_patterns), fullmatch=True).map(add_luhn_digit),  # type: ignore[arg-type]
+)
+
+# UUIDs
+st.register_type_strategy(pydantic.UUID1, st.uuids(version=1))  # type: ignore[arg-type]
+st.register_type_strategy(pydantic.UUID3, st.uuids(version=3))  # type: ignore[arg-type]
+st.register_type_strategy(pydantic.UUID4, st.uuids(version=4))  # type: ignore[arg-type]
+st.register_type_strategy(pydantic.UUID5, st.uuids(version=5))  # type: ignore[arg-type]
+
+# Secrets
+st.register_type_strategy(pydantic.SecretBytes, st.binary().map(pydantic.SecretBytes))
+st.register_type_strategy(pydantic.SecretStr, st.text().map(pydantic.SecretStr))
+
+# IP addresses, networks, and interfaces
+st.register_type_strategy(pydantic.IPvAnyAddress, st.ip_addresses())
+st.register_type_strategy(
+    pydantic.IPvAnyInterface,
+    st.from_type(ipaddress.IPv4Interface) | st.from_type(ipaddress.IPv6Interface),
+)
+st.register_type_strategy(
+    pydantic.IPvAnyNetwork,
+    st.from_type(ipaddress.IPv4Network) | st.from_type(ipaddress.IPv6Network),
+)
+
+# We hook into the con***() functions and the ConstrainedNumberMeta metaclass,
+# so here we only have to register subclasses for other constrained types which
+# don't go via those mechanisms.  Then there are the registration hooks below.
+st.register_type_strategy(pydantic.StrictBool, st.booleans())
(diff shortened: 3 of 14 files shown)


PrettyWood:
Amazing work @Zac-HD! So glad to see this PR open!
Could you please check the coverage to make sure it remains at 100% please?

Zac-HD:
@PrettyWood - thanks!  I'm looking forward to seeing what people do with it :smiley: 

There was one deliberately unreachable line (assertion that we returned from a loop body), which I've unrolled so we get 100% coverage without needing lots of pragmas.

docs/examples/hypothesis_property_based_test.py
```diff
@@ -0,0 +1,24 @@
+import typing
+import pydantic
```
samuelcolvin, line 2: we normally do `from pydantic import BaseModel, PaymentCardNumber, PositiveFloat, EmailStr` unless there's a reason to required it this way, better to match the rest of the docs.

docs/hypothesis_plugin.md
```diff
@@ -0,0 +1,46 @@
+[Hypothesis](https://hypothesis.readthedocs.io/) is the Python library for
+[property-based testing](https://increment.com/testing/in-praise-of-property-based-testing/).
+
```
samuelcolvin, line 3: You might want to add a not saying this is supported from pydantic v1.8 and hypothesis v5.29.0 onwards only.

docs/hypothesis_plugin.md
```diff
@@ -0,0 +1,46 @@
...
+Hypothesis can infer how to construct type-annotated classes, and supports builtin types,
+many standard library types, and generic types from the
+[`typing`](https://docs.python.org/3/library/typing.html) and
+[`typing_extensions`](https://pypi.org/project/typing-extensions/) modules by default.
+Pydantic's [plugin registers](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-29-0)
```
samuelcolvin, line 8: This sentence confused me.

I think something like

> Hypothesis will detect pydantic via it's `setup.py` entrypoint meaning *pydantic* custom types like `PaymentCardNumber` and `PositiveFloat`, this means that  `st.builds()`...

docs/hypothesis_plugin.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+### Help, constrained types aren't working!
+Because `pydantic.conint(...)`, `pydantic.Json[...]`, and the other constrained types
+create anonymous subtypes at runtime, they can't be automatically registered.
+You can [register them yourself](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.register_type_strategy)
+though:
+
+```py
+# Assign your constrained type to a variable
```
samuelcolvin, line 37: can you put this into a python file in examples. If you need imports you con't want to include in the file (generally this is not advisable as it makes examples harder to use), you can include a `# ignore-above` comment line.

pydantic/_hypothesis_plugin.py
```diff
@@ -0,0 +1,169 @@
+"""Register Hypothesis strategies for Pydantic custom types.
```
samuelcolvin, line 1:
```suggestion
"""
Register Hypothesis strategies for Pydantic custom types.
```

(sorry, my pet dislike of docstrings with no leading blank line.)

pydantic/_hypothesis_plugin.py
```diff
@@ -0,0 +1,169 @@
...
+https://hypothesis.readthedocs.io/en/latest/strategies.html#interaction-with-pytest-cov
+https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types
+
+Note that because our motivation is to *improve user experience*, the strategies
+are always sound (never generate invalid data) but sacrifice completeness for
+maintainability (ie may be unable to generate some tricky but valid data).
+
+Finally, this module makes liberal use of `# type: ignore[<code>]` pragmas.
+This is because Hypothesis annotates register_type_strategy with
+(T, SearchStrategy[T]), but in most cases we register e.g. AnyURL to generate strings,
```
samuelcolvin, line 20:
```suggestion
(T, SearchStrategy[T]), but in most cases we register e.g. AnyUrl to generate strings,
```

pydantic/_hypothesis_plugin.py
```diff
@@ -0,0 +1,169 @@
...
+        ),
+        ensure_ascii=st.booleans(),
+        indent=st.none() | st.integers(0, 16),
+        sort_keys=st.booleans(),
+    ),
+)
+
+# Card numbers, valid according to the Luhn algorithm
+
+
```
samuelcolvin, line 111:
```suggestion

# Card numbers, valid according to the Luhn algorithm
```

pydantic/_hypothesis_plugin.py
```diff
@@ -0,0 +1,169 @@
...
+def fix_luhn_digit(card_number: str) -> str:
+    # See https://en.wikipedia.org/wiki/Luhn_algorithm
+    assert card_number[-1] == 'X', 'expected placeholder check digit'
```
samuelcolvin, line 114: I'm not quite clear what we're doing with the `X` here? If it's just to do with keeping coverage at 100%, better to just add a pragma.

pydantic/color.py
```diff
@@ -45,13 +45,13 @@ def __getitem__(self, item: Any) -> Any:
 # these are not compiled here to avoid import slowdown, they'll be compiled the first time they're used, then cached
 r_hex_short = r'\s*(?:#|0x)?([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])?\s*'
 r_hex_long = r'\s*(?:#|0x)?([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})?\s*'
-_r_255 = r'(\d{1,3}(?:\.\d+)?)'
+_r_255 = r'((?:\d|\d\d|[01]\d\d|2[0-4]\d|25[0-4])(?:\.\d+)?|255(?:\.0+)?)'
```
samuelcolvin, line 48: I think best to add add a separate PR for these changes, sorry.

setup.py
```diff
@@ -134,4 +134,5 @@ def extra(self):
         'dotenv': ['python-dotenv>=0.10.4'],
     },
     ext_modules=ext_modules,
+    entry_points={"hypothesis": ["_ = pydantic._hypothesis_plugin"]},
```
samuelcolvin, line 137:
```suggestion
    entry_points={'hypothesis': ['_ = pydantic._hypothesis_plugin']},
```

:-)

tests/test_hypothesis_plugin.py
```diff
@@ -0,0 +1,89 @@
+import pytest
+from hypothesis import given, settings, strategies as st
+
+import pydantic
+from pydantic.networks import import_email_validator
+
+
+class MiscModel(pydantic.BaseModel):
```
samuelcolvin, line 8: best if possible for models to be declared within tests or fixtures (I know we don't currently stick to this 100%, but it would be good if we could in new code).

You can do something like

```py
def gen_models():
    class MiscModel(pydantic.BaseModel):
        ...
    yield MiscModel

    class StringsModel(pydantic.BaseModel):
        ...
    yield StringsModel
    ...


...

@pytest.mark.parametrize('model', gen_models())
```

That way the models should only be created when the test(s) are executed.

samuelcolvin (commented):
Just a few small things, overall I think this is looking great.

pydantic/color.py
Zac-HD: Done - that's #2155.  

Because this PR requires the more precise regex patterns, it's rebased on that PR but they can be merged separately.

Zac-HD:
OK @samuelcolvin - I think I'm done, and have reverted the #2155 patch in favor of a test-only solution :smile:

tests/test_hypothesis_plugin.py
```diff
@@ -0,0 +1,81 @@
...
+
+@pytest.mark.parametrize('model', gen_models())
+@settings(max_examples=20)
+@given(data=st.data())
+def test_can_construct_models_with_all_fields(data, model):
+    # We take successful creation of an instance to demonstrate that Hypothesis
+    # knows how to provide valid values for each field, so we don't need any
+    # additional assertions.
+    instance = data.draw(st.from_type(model))
+    assert isinstance(instance, model)
```
lsorber, line 81: I was reading these tests and was thrown off by this assert at first: the value of the test is actually in being able to create the instances without raising `ValidationError`s.

The comment does point this out in retrospect, but I took it to refer to the `isinstance` assertion instead of to the absence of a `ValidationError`. I would maybe clarify that in the comment, or remove the assert since it doesn't really add a lot of value on top of being able to create the instance successfully.

Zac-HD: I've split the checks and comments, so that reading top-to-bottom makes it clear what we're checking and why:

```python
def test_can_construct_models_with_all_fields(data, model):
    # The value of this test is to confirm that Hypothesis knows how to provide
    # valid values for each field - otherwise, this would raise ValidationError.
    instance = data.draw(st.from_type(model))

    # We additionally check that the instance really is of type `model`, because
    # an evil implementation could avoid ValidationError by means of e.g.
    # `st.register_type_strategy(model, st.none())`, skipping the constructor.
    assert isinstance(instance, model)
```

lsorber: Very clear, thank you!

lsorber:
@Zac-HD I was wondering, would it be possible to add generic support for constrained types like `ConstrainedStr`. For example:

```python
st.register_type_strategy(
    ConstrainedStr,
    lambda T: st.from_regex(T.regex, fullmatch=True) if T.regex is not None else st.text(min_size=T.min_length, max_size=T.max_length)
)
```

Unfortunately, that type strategy is not picked up on for subclasses of `ConstrainedStr`:

```python
class MyString(ConstrainedStr):
    regex = "[A-Z]{2,8}"

st.from_type(MyString).example()  # Returns empty string
```

This doesn't need to hold back this PR, but it is related so I thought I'd post it here.

Zac-HD:
I was all ready to explain why we couldn't, but in fact there *is* a way... it's just a little more complicated than you might expect.  First, the obvious approach of registering `ConstrainedStr` doesn't do anything for child classes.  That's correct, if unfortunate in this case - otherwise we could just return `builds(object)` for everything!

So the trick is to register a strategy for each child class *when that class is created*, which can be done from the `__init__` method of a metaclass.  The *second* trick is to ensure that this is a noop unless the user is already using Hypothesis, and *that* can be done with a `WeakSet` and a plugin (see https://github.com/Parquery/icontract/pull/181/files plus https://github.com/mristin/icontract-hypothesis/pull/5/files for an example).

I'll update this PR in a week or two, or if you'd like to merge it sooner I can open a follow-up instead 😁

samuelcolvin:
This looks good. I'm happy to merge as soon as tests are passing.

Zac-HD:
Thanks @samuelcolvin :grin: 

I think it'll take another design session; the fundamental concept is working well but there's a few issues still to iron out - `conlist()` and `conset()` have a type parameter and will require some indirection, the `ConstrainedNumberMeta` metaclass introduces a few others, and to finish up I'll need to pay some more attention to test-time performance.  But it's on track and I'm looking forward to shipping it :rocket:

samuelcolvin:
That's great, let me know when it's ready to merge and I'll do so.

samuelcolvin:
great, let me know when you think this is ready to merge.

Zac-HD:
@samuelcolvin - I'm ready to merge 🚀 

- see description above for details about the limited support for URLs, `conset()`, `conlist()`, and `Literal[None]`-on-py36
- the docs declare that this will ship in Pydantic 1.8 - is that correct?

But overall I'm very happy with where we've ended up 😁

PrettyWood:
Great job again @Zac-HD! I'll probably use it as soon as v1.8 is released 🚀 
For `conset` and `conlist`, I think it will be possible to support them in the plugin once rewritten with `Generic[T]`. We are currently waiting for `cython` to [support it](https://github.com/cython/cython/issues/3537)

Makefile
```diff
@@ -60,7 +60,7 @@ mypy:
 
 .PHONY: test
 test:
-	pytest --cov=pydantic
+	pytest -p "no:hypothesispytest" --cov=pydantic
```
PrettyWood, line 63: IMO it would be better to add it in `setup.cfg` with `addopts = -p no:hypothesispytest`
It will allow people to keep using directly `pytest`

pydantic/_hypothesis_plugin.py
```diff
@@ -0,0 +1,334 @@
...
+    st.from_type(ipaddress.IPv4Network) | st.from_type(ipaddress.IPv6Network),
+)
+
+# Constrained types
+# Because a new type is created at runtime for each new set of constraints,
+# we can only register the predefined constrained types.
+st.register_type_strategy(pydantic.StrictBool, st.booleans())
+st.register_type_strategy(pydantic.StrictInt, st.integers())  # type: ignore[arg-type]
+st.register_type_strategy(pydantic.StrictFloat, st.floats())  # type: ignore[arg-type]
+st.register_type_strategy(pydantic.StrictStr, st.text())  # type: ignore[arg-type]
```
PrettyWood, line 189: There are also `NonNegative` and `NonPositive` types. Don't know if you want to add them

pydantic/_hypothesis_plugin.py
```diff
@@ -0,0 +1,334 @@
...
+
+# Emails
+try:
+    import_email_validator()
+except ImportError:  # pragma: no cover
+    pass
+else:
+    # Note that these strategies deliberately stay away from any tricky Unicode
+    # or other encoding issues; we're just trying to generate *something* valid.
+    st.register_type_strategy(pydantic.EmailStr, st.emails())  # type: ignore[arg-type]
```
PrettyWood, line 50: It seems like Hypothesis is less strict for emails. It can generate emails like `0@A0--0.ac` that are invalid for `email-validator`

pydantic/_hypothesis_plugin.py
Zac-HD: They're already supported via the hook into `ConstrainedNumberMeta`... as in fact are `StrictInt` and `StrictFloat`, so I've remove those from here and added some comments.

pydantic/_hypothesis_plugin.py
Zac-HD: Hmm... I'll investigate that upstream, but for now - and to avoid requiring a very new Hypothesis version - I've just added a filter to reject anything that `email_validator` doesn't accept.

Zac-HD:
Thanks for the review @PrettyWood!  I've added your suggestions, along with even more comments to explain what's happening 😄

tests/requirements-linting.txt
```diff
@@ -1,6 +1,7 @@
 black==20.8b1
 flake8==3.8.4
 flake8-quotes==3.2.0
+hypothesis==5.44.0
```
PrettyWood, line 4: Should.be removed from linting deps imo

PrettyWood: Other than this looks great to me!

Zac-HD: Done!

Zac-HD: Alas, `mypy` gives different results if it's installed - so I thought it was better to always have it than to leave it out in CI but have errors if people ran `mypy pydantic` locally with Hypothesis installed.

PrettyWood: Ah! Ok it makes sense I agree with you

samuelcolvin:
this is awesome, thank you so much! 🚀 🙏 🥳 

I'm working through PRs now, v1.8 coming soon.

Zac-HD:
Woohoo!  Pydantic is also awesome, I'm so excited about combining it with Hypothesis 😁 

Can't wait to see what people do with this either :shipit:

PR #2099, opened 2020-11-07
fix(mypy): remove complaints about most custom _pydantic_ types

### LINKED ISSUE #1243: Correctly typed default values for specialized types e.g. PositiveInt, Json, ...
Neither in the documentation nor any code on the repo (e.g. in tests) could I find any example of how to work with more exotic pydantic types such as `PositiveInt` or `Json` that have default values:

Take for instance :

```python
class Foo(BaseModel):
    bar: PositiveInt = 1  # mypy complains about a type mismatch between PositiveInt and int
```

fair enough, let's try:

```python
class Foo(BaseModel):
    bar: PositiveInt = PositiveInt(1)  # typechecks, hooray
```

How about `Json` then ?

```python
class Foo(BaseModel):
    bar: Json = {"hello": "world"}  # same as above, Json is not a type alias of Dict
```

```python
class Foo(BaseModel):
    bar: Json = '{"hello": "world"}'  # again, Json is not a type alias of str
```

however:

```python
class Foo(BaseModel):
    bar: Json = Json({"hello": "world"})  # doesn't work as Json takes no arguments
```

How does one add a default value for a `Json` field (or any other more exotic field for that matter)?

### LINKED ISSUE #2098: How to keep mypy happy using `pydantic.DirectoryPath` with a default value
### Checks

* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and couldn't find an answer

# Question

```
             pydantic version: 1.7.2
            pydantic compiled: True
                 install path: /Users/michaeloliver/echo/app/.venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.0 (default, Oct 27 2020, 14:15:17)  [Clang 12.0.0 (clang-1200.0.32.21)]
```

I would like to use `DirectoryPath` with a default value if no environment override exists (using `BaseSettings`)

```py
from pathlib import Path

from pydantic import BaseSettings, DirectoryPath

BASE_DIR = Path(__file__).parent


class Config(BaseSettings):
    CONTENT_DIR: DirectoryPath = BASE_DIR / "content"

```

`mypy` complains:
```
src/hublib/config.py:8: error: Incompatible types in assignment (expression has type "Path", variable has type "DirectoryPath")
```

What would be the best approach ?

PrettyWood:
## Change Summary
Just add a `if TYPE_CHECKING` check to support directly most custom types

https://user-images.githubusercontent.com/18406791/103153051-e1409900-478d-11eb-9cb8-68f336dc2638.mov

## Related issue number
fix #2098
fix #1243

## Checklist


michael0liver (approved):
Fixes the issue I was having.

samuelcolvin:
change description, otherwise looks good.

pydantic/types.py
```diff
@@ -109,116 +109,223 @@
...
     ModelOrDc = Type[Union['BaseModel', 'Dataclass']]
 
+T = TypeVar('T')
 
-class ConstrainedBytes(bytes):
-    strip_whitespace = False
-    min_length: OptionalInt = None
-    max_length: OptionalInt = None
+
+class ConstrainedNumberMeta(type):
```
PrettyWood, line 115: I went a bit further and added a `If TYPE_CHECKING` for almost all custom types.
The diff is big because I added some comments to make the different sections explicit (see the screenshots in description) and moved some class definitions but no extra logic has been changed!

Now almost all custom types work directly with `mypy`.
The generic ones (`conlist`, `conset`) can be handled thanks to the plugin with the `get_dynamic_class_hook` method.I may work on it in a different PR.

bluetech (commented):
I've prepared a similar change before I thought to look if there is already a PR for it: https://github.com/bluetech/pydantic/commit/17c547b154098bdee73d42c741277496c56e3334?w=1 So this LGTM!

But I wonder about some cases, like `PaymentCardNumber` and `SecretStr`. They have extra functionality on top of their base type, like the `masked` property of `PaymentCardNumber`, so is it correct to completely eliminate them for mypy? It might be more conservative to only do this for the simple wrappers like `Strict*` and `Positive*`.

samuelcolvin:
This looks great, but the conflicts with master since #2097 are not at all trivial to fix.

@PrettyWood could you look at them?

PrettyWood:
I'll have a look at them tonight

samuelcolvin:
thanks so much.

PrettyWood:
@bluetech You're right. I removed custom types that had attributes or extra logic. Thanks!
@samuelcolvin All good 👍

samuelcolvin:
This is great, thank you so much.

Issue #2014, opened 2020-10-18
BaseSettings: Customization of field value priority via Config


espdev:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this feature/change is needed
* [ ] After submitting this, I commit to one of:
    * Look through open issues and helped at least one other person
    * Hit the "watch" button on this repo to receive notifications and I commit to help at least 2 people that ask questions in the future
    * Implement a Pull Request for a confirmed bug

# Feature Request

Currently, there is field value priority (from docs):

```
In the case where a value is specified for the same Settings field in multiple ways, the selected value is determined as follows (in descending order of priority):

1. Arguments passed to the Settings class initialiser.
2. Environment variables, e.g. my_prefix_special_function as described above.
3. Variables loaded from a dotenv (.env) file.
4. The default field values for the Settings model.
```

However, in some cases it would be nice to swap 1 and (2,3) priorities. For example, I would like to create the configuration system based on config files and env vars where envvars/dotenv have highest priority (env vars overrides values from config files that looks reasonable in some cases).

It might look something like this:

```
class Settings(BaseSettings):
    ...

s = Settings(**load_from_config_file(config_path))
```

In this case values from config file will override values from env vars.

Currently, the default behavior cannot be changed without subclassing `BaseSettings` class and overriding `_build_values` method. This is not mentioned in the current docs but previously it was mentioned (see #340, #341, #343).

Would you mind considering the customization of this behavior via `BaseSettings` model `Config`?

layday:
This was briefly discussed deep in a PR: https://github.com/samuelcolvin/pydantic/pull/1011#issuecomment-578158766.

layday:
Perhaps a way this would work would be to introduce a new `Config` option for `BaseSettings`, e.g. `env_has_prio` which would default to `False`.   If `True` the order would become 2, 3, 1, 4.  I don't think there is a use case for lifting `.env` above env vars.

samuelcolvin:
See my suggestion for a more powerful way to achieve this: https://github.com/samuelcolvin/pydantic/pull/2107#issuecomment-735975830

PR #2107, opened 2020-11-09
feat: Add the ability to add extra settings sources

kozlek:
## Change Summary

- refactor BaseSettings internal logic
- expose `filter_relevant_env_vars` and `load_env_vars_from_source ` functions to ease creation of external sources plugins
- add `extra_settings_sources ` ModelConfig key for BaseSettings

More info in the linked issue.

## Related issue number

https://github.com/samuelcolvin/pydantic/issues/2106

## Checklist

### DIFF
--- a/changes/2107-kozlek.md
+++ b/changes/2107-kozlek.md
@@ -1,0 +1,1 @@
+Add the ability to customize settings source (add / disable / change priority order).
--- a/changes/2107-kozlek.md
+++ b/changes/2107-kozlek.md
@@ -1,1 +1,1 @@
-Add the ability to customize settings source (add / disable / change priority order).
+Add the ability to customize settings sources (add / disable / change priority order).
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -208,3 +208,117 @@ the selected value is determined as follows (in descending order of priority):
 3. Variables loaded from a dotenv (`.env`) file.
 4. Variables loaded from the secrets directory.
 5. The default field values for the `Settings` model.
+
+### Customise settings sources priority
+
+If the default order of priority doesn't match your needs, it's possible to change it by overriding a config method:
+
+```py
+class Settings(BaseSettings):
+    ...
+    
+    class Config:
+        @classmethod
+        def customise_sources(
+            cls,
+            init_settings: SettingsSourceCallable,
+            env_settings: SettingsSourceCallable,
+            file_secret_settings: SettingsSourceCallable,
+        ) -> Tuple[SettingsSourceCallable, ...]:
+            return env_settings, init_settings, file_secret_settings
+```
+
+By flipping `env_settings` and `init_settings`, environment variables have now precedence on init kwargs.
+
+## Add and remove settings sources
+
+As explained earlier, *pydantic* ships with multiples built-in settings sources.
+However, changing the order of priority might not be enough in some complex environments, and you may need 
+to add a custom settings source:
+
+```py
+def json_config_settings_source(settings: BaseSettings) -> Dict[str, Any]:
+    """
+    A simple settings source that loads variables from a JSON file at the project's root.
+    """
+    return json.loads(Path('config.json').read_text().strip())
+
+
+class Settings(BaseSettings):
+    ...
+
+    class Config:
+        @classmethod
+        def customise_sources(
+            cls,
+            init_settings: SettingsSourceCallable,
+            env_settings: SettingsSourceCallable,
+            file_secret_settings: SettingsSourceCallable,
+        ) -> Tuple[SettingsSourceCallable, ...]:
+            return init_settings, env_settings, file_secret_settings, json_config_settings_source
+```
+
+If you want to create reusable and configurable settings source, you can also write a class based settings source:
+
+```py
+class RestSecretSettingsSource:
+    """
+    A configurable settings source that request settings variables on a REST server.
+    """
+    
+    def __init__(self, base_url: str, api_token: str):
+        self.base_url = base_url
+        self.api_token = api_token
+        
+    def retrieve_variable(self, var_name: str) -> Any:
+        response = httpx.get(
+            f'{self.base_url}/{var_name}',
+            headers={'Authorization': f'Bearer {self.api_token}'}
+        )
+        response.raise_for_status()
+        return response.json()['value']
+        
+    def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
+        return {
+            field.alias: self.retrieve_variable(var_name=field.name)
+            for field in settings.__fields__.values()
+        }  
+
+
+class Settings(BaseSettings):
+    ...
+
+    class Config:
+        @classmethod
+        def customise_sources(
+            cls,
+            init_settings: SettingsSourceCallable,
+            env_settings: SettingsSourceCallable,
+            file_secret_settings: SettingsSourceCallable,
+        ) -> Tuple[SettingsSourceCallable, ...]:
+            return (
+                init_settings,
+                env_settings,
+                file_secret_settings,
+                RestSecretSettingsSource(api_token='TOKEN', base_url='https://env.lan')
+            )
+```
+
+You might also want to disable a source:
+
+```py
+class Settings(BaseSettings):
+    ...
+
+    class Config:
+        @classmethod
+        def customise_sources(
+            cls,
+            init_settings: SettingsSourceCallable,
+            env_settings: SettingsSourceCallable,
+            file_secret_settings: SettingsSourceCallable,
+        ) -> Tuple[SettingsSourceCallable, ...]:
+            # here we choose to disable entirely environ and .env file sources
+            return init_settings, file_secret_settings
+```
+
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -214,18 +214,7 @@ the selected value is determined as follows (in descending order of priority):
 If the default order of priority doesn't match your needs, it's possible to change it by overriding a config method:
 
 ```py
-class Settings(BaseSettings):
-    ...
-    
-    class Config:
-        @classmethod
-        def customise_sources(
-            cls,
-            init_settings: SettingsSourceCallable,
-            env_settings: SettingsSourceCallable,
-            file_secret_settings: SettingsSourceCallable,
-        ) -> Tuple[SettingsSourceCallable, ...]:
-            return env_settings, init_settings, file_secret_settings
+{!.tmp_examples/settings_env_priority.py!}
 ```
 
 By flipping `env_settings` and `init_settings`, environment variables have now precedence on init kwargs.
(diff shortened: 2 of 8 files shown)


pydantic/env_settings.py
```diff
@@ -44,19 +45,58 @@ def _build_values(
...
+        extra_settings = [source(self) for source in reversed(self.__config__.extra_settings_sources)]
         return deep_update(
-            self._build_secrets_files(_secrets_dir), self._build_environ(_env_file, _env_file_encoding), init_kwargs
+            *extra_settings,
+            self._build_secrets_files(_secrets_dir),
+            self._build_environ(_env_file, _env_file_encoding),
+            init_kwargs,
         )
 
+    def filter_relevant_env_vars(self, env_vars: Mapping[str, Optional[str]]) -> Dict[str, Optional[str]]:
```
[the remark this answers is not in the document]
samuelcolvin: same with most of the methods below.

pydantic/env_settings.py
```diff
@@ -44,19 +45,58 @@ def _build_values(
...
+
+    def load_env_vars_from_source(self, loader: Callable[[str], Any]) -> Dict[str, Optional[str]]:
+        """
+        Helper function to load all relevant fields from an external source using a provided loader.
+        """
+        d: Dict[str, Optional[str]] = {}
+        for field in self.__fields__.values():
+            for env_name in field.field_info.extra['env_names']:
+                value = loader(env_name)
+                if value != undefined:
```
samuelcolvin, line 89: surely we should use `None` or `KeyError` here?

pydantic/env_settings.py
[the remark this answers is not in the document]
samuelcolvin: can you move this down to where the code was before? If so hopefully we can reduce the number of changes and keep history easy to track.

pydantic/env_settings.py
```diff
@@ -44,19 +45,58 @@ def _build_values(
         _env_file_encoding: Optional[str] = None,
         _secrets_dir: Union[Path, str, None] = None,
     ) -> Dict[str, Any]:
+        extra_settings = [source(self) for source in reversed(self.__config__.extra_settings_sources)]
```
samuelcolvin, line 48: why the `reverse`?

Also shouldn't this be
```suggestion
        extra_settings = [self.load_env_vars_from_source(source) for source in reversed(self.__config__.extra_settings_sources)]
```
?

I think this would reduce the logic and complexity in your customer loaders/sources.

samuelcolvin (commented):
I think this could be useful, but I think it needs to be rethought. It will also need tests to pass and lots of docs.

samuelcolvin:
please see #2014 and its PR #2154.

We need a way to make customising the priority of settings sources easier as well as adding new ones. I would propose something like this:

```py
from typing import Tuple, Any, Callable

SettingsSourceCallable = Callable[[str], Any]

class Settings(BaseSettings):
    foo: str
    bar: str

    class Config:
        @classmethod
        def customise_sources(
            cls,
            init_settings: SettingsSourceCallable,
            env_settings: SettingsSourceCallable,
            file_secret_settings: SettingsSourceCallable,
        ) -> Tuple[SettingsSourceCallable, ...]:
            return init_settings, env_settings, file_secret_settings
```

Here `customise_sources` performs the default behaviour, but you could change the behaviour by altering the order of the functions returned, or adding your own. Each function would then be passed to `load_env_vars_from_source` to build dicts which are then merged using `deep_update` like currently in `_build_values`.

Advantages of this approach:
* `customise_sources` (or whatever we end up calling it) is a public function without the risk of clashing with fields
* this allows (almost) complete flexibility with a reasonably simple interface

Disadvantages:
* it's more verbose and complex (though arguably more expressive) than `env_has_prio` suggested in #2014

kozlek:
First of all, thanks for your feedback 🙏

I agree with your solution involving a `customise_sources` classmethod. This lets full operability to the user, while its complexity is not so disturbing as most of people will stick with standard env priority & sources.

## SettingsSourceCallable implementation

About the implementation of `SettingsSourceCallable `, I thought we could split sources in two main categories:
1) Sources that are able to provide a full dict with all the source's variables. This is the case of the .env file or a JSON config file: it is more efficient to load the full file and parse it once, rather than reading it for each settings field.
2) Sources that load variables one by one. This is the case of Docket secrets or any proprietary API that don't provide a list endpoint: we may want to load a variable only if this is absolutely necessary - if the variable cannot be found in higher priority sources.

In the first revision of the PR, I let the `filter_relevant_env_vars` and `load_env_vars_from_source` methods as public utils to help users build custom sources.
`filter_relevant_env_vars` allows to sort useful variables from irrelevant ones thanks to its access to `self.__fields__`. If we don't filter out the extra variables, it forces us to use `extra='ignore'`. Also, we gain access to the field instance and the `is_complex` property, which is useful to load nested settings.
`load_env_vars_from_source` uses the same access to load only the necessary fields.

However, I see now we can simplify the workflow if the custom source handles the "bulk loading" itself (case 1).

```python
import json
from pathlib import Path

class JSONConfigSource:
    def __init__(self, path: Path = 'config.json'):
        self.config_path = path

    @cached_property
    def json_config(self):
        return json.loads(self.config_path.read_text())

    def __call__(self, env_name: str, field: ModelField, settings: BaseSettings):
        return self.json_config.get(env_name, undefined)
```
In this case, all sources can move to the case 2 and we can simplify the workflow as suggest. We still have to edit slightly the `load_env_vars_from_source` method to handle the `is_complex` case but that's not complicated.
Even if most source implementations can work with only the `env_name`, I think we should pass both field and settings instances to the `SettingsSourceCallable` as they provide useful context. I probably won't not use it in my own implementations but others may.

## Settings priority

As you mentioned it in your comments, settings priority seems to be important for some people. 
I think your solution fit perfectly with their needs and go even further as it allows to disable some of the built-in sources !
About the expression of the desired order of the settings, I think we agree to say we declare settings from the highest to lowest priority.
That's why we have to reverse the settings sequence when passing them to the `deep_update` method.


## Custom source config

Built-in sources like .env or docker secrets can be configured directly using `BaseSettings(_env_file=".env", _secrets_dir="/secrets/")`.
For now, custom sources can be configured using class based callable:

```python
from pathlib import Path
from .ext_sources import JSONConfigSource  # defined ealier

class Settings(BaseSettings):
    foo: str
    bar: str

    class Config:
        @classmethod
        def customise_sources(
            cls,
            init_settings: SettingsSourceCallable,
            env_settings: SettingsSourceCallable,
            file_secret_settings: SettingsSourceCallable,
        ) -> Tuple[SettingsSourceCallable, ...]:
            json_config_settings = JSONConfigSource(path=Path("config.json"))
            # here we disable both env_settings and file_secret_settings
            return init_settings, json_config_settings

settings = Settings()
```

That's not ideal but it works so we can start like this.


Based on your comments and the related issues, the whole feature appears clearer to me. Let me know if you want to add / change something else. I'll try to do the necessary changes by the end of the week, including docs & tests 🙂

pydantic/env_settings.py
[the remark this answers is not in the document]
kozlek: I'm going to revert this modification as the new implementation will only use `load_env_vars_from_source`.
Also, `load_env_vars_from_source` will be made private to avoid conflicts as you requested.

pydantic/env_settings.py
kozlek: I would prefer to keep the `undefined` as it allows the source to support any type including `None`. 
About the `KeyError` check, I think it's too specific to dict based sources: in the case of a REST API based source, a call to `loader` can trigger an API call that simply returns a flat value.
What do you think ?

pydantic/env_settings.py
kozlek: As explained in my latest comment, the settings source sequence is expressed from the highest priority to the lowest one.
We need to reverse it to send the settings dicts to `deep_update` from the lowest priority to the highest one.

In the next revision, I'll move to your implementation as all sources can be treated as case 2 sources.

kozlek:
By defining a common type for all settings sources (`SettingsSourceCallable = Callable[[str], Any]`), we consider there are all equals. 
But as of today, this is not true.

## Init settings

`init_settings ` has multiples particularities that make it different from the others sources:
- all init variables must be loaded even if there are not defined in the model fields: depending on the `extra` settings, the model will raise a ValidationError (default, `Extra.forbid`) or ignore the fields.
- init variables are loaded using `field.alias` while others sources use `env_names`.

More than simply messing up with unit tests, this would introduce breaking changes, which we might want to avoid.
Multiple solutions are possible:
1) Accept the following breaking changes:
- Extra init kwargs will be always ignored (the `extra` will be useless)
- Init kwargs will have to be named using the env_name rather the field.name (not really ideal 😕)

2) Treat InitSettingsSource as special case: either force it to be the first settings source (not acceptable for #2014) either always load all variables for this source. Also, this source will use a custom `load_env_from_source` method that will use `field.name` instead of `env_names`.

3) Add field.name to the env_name, so both env_names and field.name will be used: this could lead to an unexpected load of variables after pydantic's upgrade.

4) Move the loading strategy inside the `SettingsSourceCallable`: this will work well and increase the , but it will break the DRY pattern offered by the `load_env_from_source` method.

For now, I'll start my implementation with the solution 4) as this is the most complete and the closest from the actual source code.

kozlek:
I've modified my implementation to include a way of customising the priority of settings sources.
I had to make some tradeoff in my implementation to avoid breaking changes, as explain in my previous post ; let me know if you want me to change something.
I also added some docs with examples to explain how to reorder the settings, add / remove settings sources.

pydantic/env_settings.py
```diff
@@ -44,66 +46,110 @@ def _build_values(
...
+    __slots__ = ('init_kwargs',)
+
+    def __init__(self, init_kwargs: Dict[str, Any]):
+        self.init_kwargs = init_kwargs
 
-    def _build_environ(
-        self, _env_file: Union[Path, str, None] = None, _env_file_encoding: Optional[str] = None
-    ) -> Dict[str, Optional[str]]:
+    def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
+        return self.init_kwargs
```
samuelcolvin, line 121: might be useful to implement repr here to help people implementing their own `customise_sources`

Just something like `f'InitSettingsSource(init_kwargs={self. init_kwargs!r})'`

samuelcolvin: and for `EnvSettingsSource` and `SecretsSettingsSource`

tests/test_settings.py
```diff
@@ -446,7 +453,20 @@ class Settings(BaseSettings):
 
         def _build_values(self, init_kwargs, _env_file, _env_file_encoding, _secrets_dir):
             config_settings = init_kwargs.pop('__config_settings__')
-            return {**config_settings, **init_kwargs, **self._build_environ()}
+            return {
```
samuelcolvin, line 456: are you sure we still need to override `_build_values` here?

I suspect this case could now be accomplished with `customise_sources`.

samuelcolvin (commented):
this is looking great, just a few things to fix.

docs/usage/settings.md
```diff
@@ -208,3 +208,117 @@ the selected value is determined as follows (in descending order of priority):
 3. Variables loaded from a dotenv (`.env`) file.
 4. Variables loaded from the secrets directory.
 5. The default field values for the `Settings` model.
+
+### Customise settings sources priority
+
+If the default order of priority doesn't match your needs, it's possible to change it by overriding a config method:
+
+```py
+class Settings(BaseSettings):
```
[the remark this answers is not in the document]
kozlek: Yep, it's cleaner this way 🙂

docs/usage/settings.md
```diff
@@ -208,3 +208,117 @@ the selected value is determined as follows (in descending order of priority):
...
+            cls,
+            init_settings: SettingsSourceCallable,
+            env_settings: SettingsSourceCallable,
+            file_secret_settings: SettingsSourceCallable,
+        ) -> Tuple[SettingsSourceCallable, ...]:
+            return (
+                init_settings,
+                env_settings,
+                file_secret_settings,
+                RestSecretSettingsSource(api_token='TOKEN', base_url='https://env.lan')
```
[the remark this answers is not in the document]
kozlek: 👌

docs/usage/settings.md
```diff
@@ -208,3 +208,117 @@ the selected value is determined as follows (in descending order of priority):
...
+        def customise_sources(
+            cls,
+            init_settings: SettingsSourceCallable,
+            env_settings: SettingsSourceCallable,
+            file_secret_settings: SettingsSourceCallable,
+        ) -> Tuple[SettingsSourceCallable, ...]:
+            return init_settings, env_settings, file_secret_settings, json_config_settings_source
+```
+
+If you want to create reusable and configurable settings source, you can also write a class based settings source:
```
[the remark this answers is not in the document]
kozlek: I let the example as a second one, but I removed the explanation.

pydantic/env_settings.py
kozlek: Good idea !

tests/test_settings.py
kozlek: Totally, in the first place I kept the test this way to ensure the downstream custom config will still work. 
However, they can easily update their config using this new feature.

michael0liver:
This is a really useful feature. I would love to be able to load configuration via a local `.toml` config file while allowing overriding via environment variables.

PrettyWood (approved):
This customization will be very handy and is very explicit! LGTM! 👍
Samuel may still have some remarks though

PrettyWood (changes requested):
You just forgot to test your custom `__repr__` for `...SettingsSource` and to add a change file

kozlek:
> This customization will be very handy and is very explicit! LGTM! 👍
> Samuel may still have some remarks though

Thanks 🙏


> You just forgot to test your custom `__repr__` for `...SettingsSource` and to add a change file

You're right, I added the missing test and a change file to my latest commit. Tell me if we are missing anything else 🙂

changes/2107-kozlek.md
```diff
@@ -0,0 +1 @@
+Add the ability to customize settings source (add / disable / change priority order).
```
PrettyWood, line 1:
```suggestion
Add the ability to customize settings sources (add / disable / change priority order).
```

PrettyWood (approved):

samuelcolvin:
thanks so much. I've extended the docs slightly as the original docs on this were pretty minimal.

kozlek:
No problem, thanks for taking the time to merge this one 👌
Don't hesitate to ping me about settings-related issues & feature requests, it will be a pleasure to help when I have some time 😉

DomWeldon:
As a thank-you to everyone on this thread, I spent yesterday browsing the master branch and saw this hook before I saw this issue. I was just testing why it didn't work now and saw that it was because this feature isn't released yet! 

It's a really nice implementation and fits my needs exactly - thanks. Looking forward to this hitting a new release. 

In the meantime, my draft implementation - to pull some secrets from SSM in prod - is at the gist below.

https://gist.github.com/DomWeldon/ce7e070283d97368cd9abc5be71b247d

PR #2110, opened 2020-11-10
Updated validators.md with the type of field

oridistor:
As someone started using pydantic, it can be very convenient to know what type we can rely the returned field is, so we know what we can check on it and what fields it has.
After searching the code I found out it has to be of type ModelField and that's why I'm offering it here as something that I think should be clarified in the documentation.

## Checklist


samuelcolvin:
thanks a lot.

PR #2115, opened 2020-11-12
Handle singular item/character values

jillesme:
## Change Summary

Allows for grammatically correct handling of singular values in lists, sets and strings. 

## Related issue number

No issue, change is trivial.

image

## Checklist

Will go through the checklist if this is a desired change.

### DIFF
--- a/pydantic/errors.py
+++ b/pydantic/errors.py
@@ -303,6 +303,8 @@ class ListMinLengthError(PydanticValueError):
     msg_template = 'ensure this value has at least {limit_value} items'
 
     def __init__(self, *, limit_value: int) -> None:
+        if limit_value == 1:
+            self.msg_template = 'ensure this value has at least {limit_value} item'
         super().__init__(limit_value=limit_value)
 
 
@@ -311,6 +313,8 @@ class ListMaxLengthError(PydanticValueError):
     msg_template = 'ensure this value has at most {limit_value} items'
 
     def __init__(self, *, limit_value: int) -> None:
+        if limit_value == 1:
+            self.msg_template = 'ensure this value has at most {limit_value} item'
         super().__init__(limit_value=limit_value)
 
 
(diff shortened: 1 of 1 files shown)


pydantic/errors.py
```diff
@@ -303,6 +303,8 @@ class ListMinLengthError(PydanticValueError):
     msg_template = 'ensure this value has at least {limit_value} items'
 
     def __init__(self, *, limit_value: int) -> None:
+        if limit_value == 1:
+            self.msg_template = 'ensure this value has at least {limit_value} item'
```
jillesme, line 307: I didn't reverse it (e.g. `msg = 'count {limit_value} item' + 's' if limit_value > 1 else ''`) because I thought this was more readable. Would be happy to create a `PluralItemsMixin` or something if preferred 😊 . I'm not doing a ton of Python development these days so any feedback / improvement ideas are welcome.

samuelcolvin:
Surely we can do something more elegant than this? 😄 

What about?

```py
from string import Formatter

class PluralFormatter(Formatter):
    def format_field(self, value, format_spec):
        if format_spec == 'p':
            return '' if value == 1 else 's'
        elif format_spec.startswith('p,'):
            _, single, plural = format_spec.split(',', 2)
            return single if value == 1 else plural
        else:
            # default just calls format
            return format(value, format_spec)

plural_formatter = PluralFormatter()

print(plural_formatter.format('{count} apple{count:p}', count=1))
#> 1 apple
print(plural_formatter.format('{count} apple{count:p}', count=123))
#> 123 apples
print(plural_formatter.format('{count} countr{count:p,y,ies}', count=1))
#> 1 country
print(plural_formatter.format('{count} countr{count:p,y,ies}', count=123))
#> 123 countries
```

jillesme:
@samuelcolvin what you wrote is infinitely times more elegant than what I had. I had no clue `from string import Formatter` even existed. Great feedback. Will update PR!

samuelcolvin:
@jillesme would you be happy to implement what I suggested as a PR, or do you want to close this and I'll create a separate PR.

jillesme:
@samuelcolvin excuse my late response. Unfortunately due to personal time constraints I don't think I can commit to this right now. I'd love to see the PR you create, it will undoubtedly be much better than mine.

alecgerona:
Would it be better to just simplify this to adding `(s)` or `/s` at the end of the word? Simpler than accounting for plural variations.

PR #2118, opened 2020-11-13
Implement MongoDsn field types

### LINKED ISSUE #2117: Feature Request: Mongo(DB)Dsn type
### Checks

# Feature Request

```
             pydantic version: 1.7.2
            pydantic compiled: True
                 install path: /Users/step7212/.pyenv/versions/3.9.0/envs/chas3/lib/python3.9/site-packages/pydantic
               python version: 3.9.0 (default, Oct  5 2020, 14:52:54)  [Clang 11.0.0 (clang-1100.0.33.17)]
     optional deps. installed: ['typing-extensions', 'devtools']
```

```py
import pydantic

   mongodb: pydantic.MongodbDsn

MyModel(mongodb="mongodb://mongodb1.example.com:27317,mongodb2.example.com:27017/?replicaSet=mySet&authSource=authDB")
```

The MongoDB URI scheme is defined here: https://www.iana.org/assignments/uri-schemes/prov/mongodb
as: `mongodb://[username:password@]host1[:port1][,...hostN[:portN]][/[defaultauthdb][?options]]`
some examples of which can be found in the Reference: https://docs.mongodb.com/manual/reference/connection-string/
```
mongodb://mongodb0.example.com:27017
mongodb://myDBReader:D1fficultP%40ssw0rd@mongodb0.example.com:27017/?authSource=admin
mongodb://mongodb1.example.com:27317,mongodb2.example.com:27017/?replicaSet=mySet&authSource=authDB
mongodb://db0.example.com,db1.example.com,db2.example.com/?replicaSet=myRepl&tls=true
```

The sticking point is, I think, the potential for multiple hosts, i.e. domain:port combinations. The current regex only grabs a single host, separate from an optional port. If we were to modify it to group the host and port sections with a repeating operator, Python's `re` module can't capture multiple groups, while the third-party `regex` package can. However, that would add a new dependency to pydantic, which we _probably_ don't want to do.

Another option is to group the whole thing, and split and parse it manually, or just scrap the pattern and write a new one, since the fragment and other items aren't used, and the `query` string is called `options`, and the `path` is used for the `database` to auth to.

A couple other notes lead me to the latter option, namely:

> If the username or password includes the following characters:
> ```
> : / ? # [ ] @
> ```
> those characters must be converted using percent encoding.

and some backwards compatibility:

> **Semi-colon separator for connection string arguments**
> 
> To provide backwards compatibility, drivers currently accept semi-colons (i.e. ;) as option separators.

While I would love to re-use as much functionality from `AnyUrl` as already exists, the quirks of MongoDB connection strings seems to outweigh the similarities it has.

If anyone wants to take a crack at implementing a `MongodbDsn` type, or has a good idea on what to start on for the regex, that'd be great. I just want to put this out there to gauge interest and difficulty of implementation given the above quirks and potentially others that I may not have noticed as of yet.

art049:
## Change Summary

Add  two DSN field types for MongoDB:
- MongoDsn for regular `mongodb://...` connection strings (still todo)
- MongoSrvDsn handling `mongodb+srv://...` aka  [DNS Seed List](https://docs.mongodb.com/manual/reference/connection-string/#dns-seed-list-connection-format) connection strings

## Related issue number

Implements and resolves #2117.

## Checklist

### DIFF
--- a/pydantic/__init__.py
+++ b/pydantic/__init__.py
@@ -50,6 +50,7 @@ __all__ = [
     'IPvAnyNetwork',
     'PostgresDsn',
     'RedisDsn',
+    'MongoSrvDsn',
     'validate_email',
     # parse
     'Protocol',
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -40,6 +40,7 @@ __all__ = [
     'IPvAnyNetwork',
     'PostgresDsn',
     'RedisDsn',
+    'MongoSrvDsn',
     'validate_email',
 ]
 
@@ -280,6 +281,10 @@ class PostgresDsn(AnyUrl):
 
 class RedisDsn(AnyUrl):
     allowed_schemes = {'redis'}
+
+
+class MongoSrvDsn(AnyUrl):
+    allowed_schemes = {'mongodb+srv'}
 
 
 def stricturl(
--- a/tests/test_networks.py
+++ b/tests/test_networks.py
@@ -1,6 +1,17 @@
 import pytest
 
-from pydantic import AnyUrl, BaseModel, EmailStr, HttpUrl, NameEmail, PostgresDsn, RedisDsn, ValidationError, stricturl
+from pydantic import (
+    AnyUrl,
+    BaseModel,
+    EmailStr,
+    HttpUrl,
+    MongoSrvDsn,
+    NameEmail,
+    PostgresDsn,
+    RedisDsn,
+    ValidationError,
+    stricturl,
+)
 from pydantic.networks import validate_email
 
 try:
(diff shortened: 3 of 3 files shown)


samuelcolvin:
Generally I'd be happy to accept Mongo DSNs, but reading #2117 they seem unbelievable complicated. Should that logic really live in pydantic?

`MongoSrvDsn` looks simple enough, but:
* how often is it used?
* is it often used without full blown `MongoDsn`?

E.g. is it worth adding `MongoSrvDsn` to pydantic if it's only used alongside `MongoDsn` when ends up in a separate package?

---

Please excuse my ignorance, I've never really used mongodb.

StephenBrown2:
I think that support for Mongo DSNs could be simplified to just validating a URL with the potential for multiple host:port sections. The added stuff (with authsource, database, etc), can be handled in code, and will likely not be accessed from the validated pydantic field, but rather the mongo module taking a URI would just take the string representation.

I think that discussion in https://github.com/samuelcolvin/pydantic/issues/2117 went a little into the weeds when it shouldn't have. We should keep it simple and only use Pydantic for validation in this case for now.

art049:
@StephenBrown2  it looks better to me that way :sweat_smile: 
Just having the validation without the actual parsing for regular MongoDsn should be way easier to achieve without a lot of duplication (and more suitable for pydantic IMO) than actually handling all the logic behind these (providing host list, ports and all the option brought with the actual dsn).

Does it look reasonable to you @samuelcolvin ?

samuelcolvin:
@art049 sounds good to me.

StephenBrown2:
Just a note, I came across the actual parsing done for a URI in pymongo: https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/uri_parser.py
Which further solidifies the fact that we don't need _all that_, just allowing multiple comma-separated hosts should do the basics.

samuelcolvin:
@art049 can you fix this?

samuelcolvin:
Closing this as no response.

Anyone is free to create another PR for this and resolve the comments above.

PR #2119, opened 2020-11-13
fix: update all modified field values in `root_validator` when `validate_assignment` is on

### LINKED ISSUE #2116: validate_assignment with side effect validators
# Bug

```
             pydantic version: 1.7.2
            pydantic compiled: True
                 install path: /home/arthur/Documents/odmantic/.venv/lib/python3.8/site-packages/pydantic
               python version: 3.8.1 (default, Jan  5 2020, 23:12:00)  [GCC 7.3.0]
                     platform: Linux-4.17.14-041714-generic-x86_64-with-glibc2.27
```

```py
from pydantic import root_validator, BaseModel

class Rectangle(BaseModel):
    width: float
    height: float
    area: float = None

    class Config:
        validate_assignment = True

    @root_validator()
    def set_ts_now(cls, v):
        v["area"] = v["width"] * v["height"]
        return v

r = Rectangle(width=1, height=1)
assert r.area == 1
r.height = 5
assert r.area == 5 # AssertionError, r.area has not been modified
```
I don't know if it should be allowed to use validators this way or if it's a real bug.
I'd be happy to try to implement a fix :smile:

PrettyWood:
## Change Summary
Following #1971 we now run `root_validator` on assignment when `Config.validate_assignment` is set.
But we still only updated the assigned field and not all concerned fields

ℹ️ Not really a regression but more a fix to #1972 in v1.7.0

## Related issue number
closes #2116

## Checklist

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -666,6 +666,28 @@ def test_validating_assignment_post_root_validator_fail():
     ]
 
 
+def test_root_validator_many_values_change():
+    """It should run root_validator on assignment and update ALL concerned fields"""
+
+    class Rectangle(BaseModel):
+        width: float
+        height: float
+        area: float = None
+
+        class Config:
+            validate_assignment = True
+
+        @root_validator
+        def set_ts_now(cls, values):
+            values['area'] = values['width'] * values['height']
+            return values
+
+    r = Rectangle(width=1, height=1)
+    assert r.area == 1
+    r.height = 5
+    assert r.area == 5
+
+
 def test_enum_values():
     FooEnum = Enum('FooEnum', {'foo': 'foo', 'bar': 'bar'})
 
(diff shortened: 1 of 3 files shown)


art049 (approved):

PrettyWood:
@samuelcolvin I took the liberty to add the `bug` label to all the PRs that fix regressions from pydantic 1.7+ to help you see what could be added for a possible v1.7.3
If you don't like this please let me know.
If you do, I guess a better label should be found to differentiate a regular fix and a regression on the latest release

tests/test_main.py
```diff
@@ -666,6 +666,28 @@ def current_lessequal_500(cls, values):
...
+    class Rectangle(BaseModel):
+        width: float
+        height: float
+        area: float = None
+
+        class Config:
+            validate_assignment = True
+
+        @root_validator
+        def set_ts_now(cls, values):
```
art049, line 681: Wrong validator function name from the issue snippet, my bad
```suggestion
        def set_area(cls, values):
```

PR #2149, opened 2020-11-28
feat: add validator for `None`, `NoneType` or `Literal[None]`

### LINKED ISSUE #2095: Pydantic does not seem to have a validator for `NoneType`
I encountered this while working on #2017:

```py
from typing import List
from pydantic import BaseModel, Json

class JsonDemo(BaseModel):
    # I'd expect this to work like typing.List[None], but at definition-time it raises
    # RuntimeError: error checking inheritance of None (type: NoneType)
    none: Json[None]
JsonDemo(none="null")


class JsonWorkaround(BaseModel):
    # The workaround of using type(None) also fails at definition-time, raising
    # RuntimeError: no validator found for <class 'NoneType'>, see `arbitrary_types_allowed` in Config
    # The same error is apparent for List[None], so I think it's a more general bug
    json_null: Json[type(None)]
    list_of_none: List[None]
```

I would expect both `Json[None]` and `List[None]` to work, accepting in respectively `"null"` or a list (or sequence) where all the elements are `None` - casting need not be supported because `None` is a singleton.

PrettyWood:
## Change Summary
_pydantic_ currently doesn't have support for `None` value only. As explained in the issue, it should be supported by _pydantic_ and `None`, `NoneType` (aka `type(None)`) or `Literal[None]` should all be considered as equal as other tools do.
I know there may be some debate on this syntax
```python
class Model(BaseModel):
    my_field: None
```
but IMO we should be consistent everywhere. And it didn't cost much to just replace the default type to `Undefined` (which could be replaced by `_missing` or any other sentinel but not `None`!)

## Related issue number
closes #2095

## Checklist


tests/test_types.py
```diff
@@ -2540,3 +2541,34 @@ class Model(BaseModel):
         v: Deque[int]
 
     assert Model(v=deque((1, 2, 3))).json() == '{"v": [1, 2, 3]}'
+
+
+@pytest.mark.parametrize('value_type', (None, NoneType, NoneLiteral))
```
Zac-HD, line 2546: I'd add a fresh `Literal[None]` here, to check that it works even when the annotation is a different object to the `NoneLiteral` constant.

PrettyWood: Thanks for reviewing! https://github.com/samuelcolvin/pydantic/pull/2149/commits/832068d322ad384def00681f6d26d6225e283e29

samuelcolvin:
`my_field: None` still scares me, but I get that it matches the PEP.

I'm happy to accept this, just need to add docs.

PrettyWood:
@samuelcolvin Added documentation but also schema support!

pydantic/typing.py
```diff
@@ -176,6 +177,9 @@ def get_args(tp: Type[Any]) -> Tuple[Any, ...]:
 
 
 NoneType = None.__class__
+NONE_TYPES: Tuple[Any, ...] = (None, NoneType)
```
MrMrRobat, line 180: Can we use `set()` here for faster lookups?

PrettyWood: Done here https://github.com/samuelcolvin/pydantic/pull/2149/commits/dd3a75c01c7e911dd21dbb1e3a0272588dea29e8

samuelcolvin:
awesome, thank you.

PR #2154, opened 2020-11-29
Add config to prioritize environment variables in BaseSettings

### LINKED ISSUE #2014: BaseSettings: Customization of field value priority via Config

tomgrin10:
## Change Summary
Added `env_has_priority` config option to `BaseSettings` class.
When set to `True`, the class will prioritize environment variables and secrets instead of constructor arguments.

## Related issue number
Fixes #2014

## Checklist

### DIFF
--- a/changes/2154-tomgrin10.md
+++ b/changes/2154-tomgrin10.md
@@ -1,0 +1,1 @@
+Add `env_has_priority` config to `BaseSettings` for prioritizing environment variables and secrets. 
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -208,3 +208,5 @@ the selected value is determined as follows (in descending order of priority):
 3. Variables loaded from a dotenv (`.env`) file.
 4. Variables loaded from the secrets directory.
 5. The default field values for the `Settings` model.
+
+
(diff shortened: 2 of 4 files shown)


samuelcolvin:
see #2107, I think that PR and this one are incompatible.

It seems like people want even more potential sources of settings. With that a simple config flag to switch the order of two of settings sources is not flexible enough.

I'd rather improve #2107 to make it as easy to use as possible.

PR #2155, opened 2020-11-30
Tighten color regex

Zac-HD:
Update the regex patterns for RGB, RGBA, and HSL strings to match only valid colors.

This patch was extracted from #2097, where we use Hypothesis' `from_regex()` strategy to generate test inputs.  It will need to be merged before that PR, but is worth merging and documenting independently - strings like `rgb(300, 300, 300)` now fail to match the regex instead of raising an error later.

There are no new tests for this change, because everything that used to work (or fail) still does work or fail respectively, and we maintain 100% coverage.

### DIFF
--- a/changes/2017-Zac-HD.md
+++ b/changes/2017-Zac-HD.md
@@ -1,0 +1,1 @@
+Update the regex patterns for RGB, RGBA, and HSL strings to match only valid colors.
--- a/pydantic/color.py
+++ b/pydantic/color.py
@@ -45,13 +45,13 @@ class RGBA:
 # these are not compiled here to avoid import slowdown, they'll be compiled the first time they're used, then cached
 r_hex_short = r'\s*(?:#|0x)?([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])?\s*'
 r_hex_long = r'\s*(?:#|0x)?([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})?\s*'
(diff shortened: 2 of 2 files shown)


samuelcolvin:
looks good, can you add some tests for outside the ranges to confirm this is working correctly.

PrettyWood:
I think we can't really add test cases as there are already invalid colors with outside the range values. But it should be possible to simplify the code by remove the "minus 255" logic.

Zac-HD:
Unfortunately (?) this doesn't change the range of valid values at all, so it's fully covered by the existing tests.  The total effect of this change is that 

- Using Hypothesis to generate matching strings only generates valid strings
- Invalid inputs [raise this `ColorError`](https://github.com/Zac-HD/pydantic/blob/9ab34b4aaea064361cf2fe556f79254dc0898ddd/pydantic/color.py#L271) instead of [e.g. this one](https://github.com/Zac-HD/pydantic/blob/9ab34b4aaea064361cf2fe556f79254dc0898ddd/pydantic/color.py#L290).

If the second change is undesirable I can close this PR and just use a more complicated Hypothesis strategy to generate color strings in the plugin.

samuelcolvin:
Then I do think this is a regression.

I was surprised i hadn't implemented the regexes like this to start with, that explains my decision.

Can't you just use those new regexes for hypothesis?

Zac-HD:
Yeah, I'll either use them (involves replicating a bit more logic to reconstruct everything else), or just implement it with format strings (probably more efficient shrinking).

Thanks for the quick review :grin:

PR #2163, opened 2020-12-01
docs: add explicit example of lower camel-case alias generator

huonw:
## Change Summary

Many APIs use lowerCamelCase for properties. This patch adjusts the
documentation to be explicit about an implementation of
`alias_generator` to automatically create such lower camel-case
aliases.  This hopefully simplifies things for users without making
the documentation significantly harder to understand.

(I totally understand if this isn't an overall improvement, I just feel that lowerCamelCase is likely to be a more common alias than UpperCamelCase or most other forms of aliases, and so is worth calling out.)

## Related issue number

N/A, since this is a small doc change.

## Checklist

* [x] (N/A) Unit tests for the changes exist
* [x] (N/A) Tests pass on CI and coverage remains at 100%

### DIFF
--- a/changes/2163-huonw.md
+++ b/changes/2163-huonw.md
@@ -1,0 +1,1 @@
+docs: add explicit example of lower camel-case alias generator
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -109,8 +109,13 @@ you can automatically generate aliases using `alias_generator`:
 _(This script is complete, it should run "as is")_
 
 Here camel case refers to ["upper camel case"](https://en.wikipedia.org/wiki/Camel_case) aka pascal case
-e.g. `CamelCase`. If you'd like instead to use lower camel case e.g. `camelCase`,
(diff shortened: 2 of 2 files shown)


PrettyWood:
Hi @huonw 
Personally I would have just changed the `to_camel` function in `model_config_alias_generator.py` as it's probably more common indeed. But better not overload the documentation with plain python code that is easy to write down or search
WDYT?

huonw:
I'm more than happy with that if that's the preferred approach. I thought that given the code is (as stated) "trivial", it was nicer to include it directly, rather than spend almost the same amount of space telling the reader that they can find/write it themselves.

samuelcolvin:
Python code directly in markdown (rather than in python files in examples) should be avoided if at all possible - is very easy for code like this to become out of date.

Personally I agree with @PrettyWood. That's such a thing as too many examples.

huonw:
Sounds reasonable. Thanks for taking the time to review!

PR #2170, opened 2020-12-02
Fixed KeyError when creating schema from models with same names declared within different classes

### LINKED ISSUE #1912: Generating Json Schema for nested models that have overlapping model names.
# Bug

I'm trying to generate a schema that includes several models that have nested models. The inner-most models have identical name, `Params`, but since they are used only inside the class, there is no name clash. However, when adding outer models to the same schema, both inner-most `Params` models end up in the same ref location: `{'$ref': '#/definitions/Params'}`, which I believe causes the error I see: `KeyError: <class '__main__.A.Params'>`.

I believe some fully-qualified path in '$ref' would resolve this, something like:
```
{'$ref': '#/definitions/A/Params'}
{'$ref': '#/definitions/B/Params'}
```
As a workaround, I wonder if there a way I can adjust prefix of each model independently? Or is there a way to put schema of nested classes "in-place", without creating a "$ref"?

Here is a complete minimal example:

```py
from pydantic import BaseModel, schema


class A(BaseModel):
    class Params(BaseModel):
        a: int

    params: Params  # A.Params

class B(BaseModel):
    class Params(BaseModel):
        b: int

    params: Params  # B.Params


schema.schema([A])  # Works fine.
schema.schema([B])  # Works fine.
try:
    schema.schema([A, B])  # Two schemas together - does not work.
except KeyError as e:
    print(f"{e.__class__.__name__}: {str(e)}")
    """
    KeyError: <class '__main__.A.Params'>
    """
```

```
             pydantic version: 1.6.1
            pydantic compiled: True
               python version: 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 02:25:08)  [GCC 7.5.0]
                     platform: Linux-3.10.0-957.27.2.el7.x86_64-x86_64-with-centos-7.6.1810-Core
```

### LINKED ISSUE #2225: KeyError when having same Inner Class name in different Outer Class
### Checks

# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: C:\Users\Dell\AppData\Local\Programs\Python\Python37\Lib\site-packages\pydantic
               python version: 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)]
```
---
First of all, pydantic is awesome and super useful! Kudos to all the contributors! However, an error occurred when I was accessing the `/openapi.json` endpoint when I was testing out my FastAPI server. 
The response models are located in the models.py file which looks like this.
```py
# models.py
from typing import List

class AModel(BaseModel):
    class _InnerClass(BaseModel):
        error_code: str = "012345"
        
    errors: List[_InnerClass] = _InnerClass()

class BModel(BaseModel):
    class _InnerClass(BaseModel):
        error_code: str = "123456"
        
    errors: List[_InnerClass] = _InnerClass()
```
The FastAPI server file looks like this.
```py
# main.pyimport models
from fastapi import FastAPI

app = FastAPI()

@app.get("/A", response_model=models.AModel)
async def ARoute():
    return models.AModel()


@app.get("/B", response_model=models.BModel)
async def BRoute():
    return models.BModel()
```
When I run the command `uvicorn main:app` and accessed the route `/openapi.json` the following traceback is returned:
```shell
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 384, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\fastapi\applications.py", line 190, in __call__
    await super().__call__(scope, receive, send)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\applications.py", line 111, in __call__
    await self.middleware_stack(scope, receive, send)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\middleware\errors.py", line 181, in __call__
    raise exc from None
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\middleware\errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\exceptions.py", line 82, in __call__
    raise exc from None
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\routing.py", line 566, in __call__
    await route.handle(scope, receive, send)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\routing.py", line 227, in handle
    await self.app(scope, receive, send)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\starlette\routing.py", line 41, in app
    response = await func(request)
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\fastapi\applications.py", line 143, in openapi
    return JSONResponse(self.openapi())
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\fastapi\applications.py", line 128, in openapi
    servers=self.servers,
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\fastapi\openapi\utils.py", line 350, in get_openapi
    flat_models=flat_models, model_name_map=model_name_map  # type: ignore
  File "c:\users\dell\appdata\local\programs\python\python37\lib\site-packages\fastapi\utils.py", line 26, in get_model_definitions
    model, model_name_map=model_name_map, ref_prefix=REF_PREFIX  # type: ignore
  File "pydantic\schema.py", line 455, in pydantic.schema.model_process_schema
  File "pydantic\schema.py", line 491, in pydantic.schema.model_type_schema
  File "pydantic\schema.py", line 185, in pydantic.schema.field_schema
  File "pydantic\schema.py", line 372, in pydantic.schema.field_type_schema
  File "pydantic\schema.py", line 614, in pydantic.schema.field_singleton_schema
  File "pydantic\schema.py", line 539, in pydantic.schema.field_singleton_sub_fields_schema
  File "pydantic\schema.py", line 412, in pydantic.schema.field_type_schema
  File "pydantic\schema.py", line 665, in pydantic.schema.field_singleton_schema
KeyError: <class 'models.AModel._InnerClass'>
```
I temporarily solved the issue by editing the `get_long_model_name` function in [file](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/schema.py) of my local copy to use model.\_\_qualname\_\_ (only for python 3.3+) instead of model.\_\_name\_\_:
```py
...
def get_long_model_name(model: Type['BaseModel']) -> str:
    return f'{model.__module__}__{model.__qualname__}'.replace('.', '__')
...
```

JSextonn:
## Change Summary

Conflicting class names were being resolved but did not account for the possibility of classes to be declared within other classes. The resolving "long name" only consisted of the module and class name.

This PR updates the "long name" to be the full type name which does not lose the context required for distinguishing nested class declarations.

## Related issue number
closes #1912
closes #2225

## Checklist

* [x] `changes/1912-JSextonn.md` file added describing change

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -395,7 +395,7 @@ def get_flat_models_from_models(models: Sequence[Type['BaseModel']]) -> TypeMode
 
 
 def get_long_model_name(model: TypeModelOrEnum) -> str:
-    return f'{model.__module__}__{model.__name__}'.replace('.', '__')
+    return f'{model.__module__}__{model.__qualname__}'.replace('.', '__')
 
 
 def field_type_schema(
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2132,7 +2132,7 @@ class NestedModel(BaseModel):
         'ModelTwo',
         f'{module.__name__}__ModelOne__NestedModel',
         f'{module.__name__}__ModelTwo__NestedModel',
-        f'{module.__name__}__NestedModel'
+        f'{module.__name__}__NestedModel',
     }
     assert model_names == expected_model_names
 
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2097,3 +2097,19 @@ def test_new_type():
         'properties': {'a': {'title': 'A', 'type': 'string'}},
         'required': ['a'],
     }
+
+
+def test_multiple_nested_model_declarations_do_not_raise_error():
+    class ModelOne(BaseModel):
+        class NestedModel(BaseModel):
+            a: float
+
+        nested: NestedModel
+
+    class ModelTwo(BaseModel):
+        class NestedModel(BaseModel):
+            b: float
+
+        nested: NestedModel
+
+    schema([ModelOne, ModelTwo])
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2105,11 +2105,13 @@ def test_multiple_models_with_same_name(create_module):
         """
 from pydantic import BaseModel
 
+
 class ModelOne(BaseModel):
     class NestedModel(BaseModel):
         a: float
 
     nested: NestedModel
+
 
 class ModelTwo(BaseModel):
     class NestedModel(BaseModel):
@@ -2117,9 +2119,10 @@ class ModelTwo(BaseModel):
 
     nested: NestedModel
 
+
 class NestedModel(BaseModel):
     c: float
- """
+        """
     )
 
     models = [module.ModelOne, module.ModelTwo, module.NestedModel]
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2099,20 +2099,39 @@ def test_new_type():
     }
 
 
-def test_multiple_nested_model_declarations_do_not_raise_error():
-    class ModelOne(BaseModel):
-        class NestedModel(BaseModel):
-            a: float
+def test_multiple_models_with_same_name(create_module):
+    module = create_module(
+        # language=Python
+        """
+from pydantic import BaseModel
 
-        nested: NestedModel
+class ModelOne(BaseModel):
+    class NestedModel(BaseModel):
+        a: float
 
-    class ModelTwo(BaseModel):
-        class NestedModel(BaseModel):
-            b: float
+    nested: NestedModel
 
-        nested: NestedModel
+class ModelTwo(BaseModel):
+    class NestedModel(BaseModel):
+        b: float
 
-    schema([ModelOne, ModelTwo])
+    nested: NestedModel
+
+class NestedModel(BaseModel):
+    c: float
+ """
+    )
+
+    models = [module.ModelOne, module.ModelTwo, module.NestedModel]
+    model_names = set(schema(models)['definitions'].keys())
+    expected_model_names = {
+        f'ModelOne',
+        f'ModelTwo',
+        f'{module.__name__}__ModelOne__NestedModel',
+        f'{module.__name__}__ModelTwo__NestedModel',
+        f'{module.__name__}__NestedModel'
+    }
+    assert model_names == expected_model_names
 
 
 def test_multiple_enums_with_same_name(create_module):
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2125,8 +2125,8 @@ class NestedModel(BaseModel):
     models = [module.ModelOne, module.ModelTwo, module.NestedModel]
     model_names = set(schema(models)['definitions'].keys())
     expected_model_names = {
-        f'ModelOne',
-        f'ModelTwo',
+        'ModelOne',
+        'ModelTwo',
         f'{module.__name__}__ModelOne__NestedModel',
         f'{module.__name__}__ModelTwo__NestedModel',
         f'{module.__name__}__NestedModel'
--- a/changes/1912-JSextonn.md
+++ b/changes/1912-JSextonn.md
@@ -1,0 +1,1 @@
+Fixed issue causing KeyError to be raised when building schema from multiple `BaseModel` with the same names declared in separate classes.


pydantic/schema.py
```diff
@@ -395,7 +395,7 @@ def get_flat_models_from_models(models: Sequence[Type['BaseModel']]) -> TypeMode
 
 
 def get_long_model_name(model: TypeModelOrEnum) -> str:
-    return f'{model.__module__}__{model.__name__}'.replace('.', '__')
```
PrettyWood, line 398: I think it's best to keep it this way but just use `__qualname__` instead of `__name__` if possible

JSextonn: Yea I think this is a much better approach and do not see why this would not work. Also, I had no idea `__qualname__` existed, so thanks!

A commit has been pushed with this change.

tests/test_schema.py
```diff
@@ -2097,3 +2097,19 @@ class Model(BaseModel):
...
+
+        nested: NestedModel
+
+    class ModelTwo(BaseModel):
+        class NestedModel(BaseModel):
+            b: float
+
+        nested: NestedModel
+
+    schema([ModelOne, ModelTwo])
```
samuelcolvin, line 2115: We need to check against an expected schema (or at least the names part of the schema), not just confirm no error is raised.

We need a test which locks the exact model name behaviour.

samuelcolvin (commented):
I'm happy with this, my only concern is over whether this might be a breaking change for anyone?

I see no tests have been changed which is a good sign, but I'm still a little unclear about when and where schemas will have changed.

tests/test_schema.py
JSextonn: Something like the below?

```python
schema_definitions = schema([ModelOne, ModelTwo])['definitions']
assert len(schema_definitions) == 4

# All expected model names to be generated
model_names = (
    'tests__test_schema__test_multiple_nested_model_declarations_do_not_raise_error__<locals>__ModelOne__NestedModel',
    'ModelOne',
    'tests__test_schema__test_multiple_nested_model_declarations_do_not_raise_error__<locals>__ModelTwo__NestedModel',
    'ModelTwo'
)
assert all(name in schema_definitions for name in model_names)
```

PrettyWood: Something like this yes. Probably better to use `create_module` fixture

JSextonn: Thanks,  very helpful stuff.

tests/test_schema.py
```diff
@@ -2099,6 +2099,44 @@ class Model(BaseModel):
...
 
+def test_multiple_models_with_same_name(create_module):
+    module = create_module(
+        # language=Python
+        """
+from pydantic import BaseModel
+
+
+class ModelOne(BaseModel):
+    class NestedModel(BaseModel):
```
adriangb, line 2110: The problem does not necessarily arise from the outer model being a Pydantic model; it's purely a namespace issue. Thus I would simplify the test by removing the inheritance from `BaseModel` on the outer models.

tests/test_schema.py
```diff
@@ -2099,6 +2099,44 @@ class Model(BaseModel):
...
+
+
+class NestedModel(BaseModel):
+    c: float
+        """
+    )
+
+    models = [module.ModelOne, module.ModelTwo, module.NestedModel]
+    model_names = set(schema(models)['definitions'].keys())
+    expected_model_names = {
```
adriangb, line 2130: Could this test `get_model_name_map` directly? It'd be the same structure as `test_model_name_maps`.

tests/test_schema.py
PrettyWood: You're right but both linked issues had this kind of issue. I reckon this is why the author went with this

adriangb: Makes sense, that's actually how I found the problem too.

tests/test_schema.py
PrettyWood: I tend to think that what really matters in the end is just to have the right definitions in the schema. We could refactor `get_model_name_map` and still have the same result. It's always the question of "unit test" and if you simply test your function or the expected output. Here IMO testing the output is what has the most value

adriangb: You're right, testing the API is always best. I just thought that maybe `test_model_name_maps` was part of the API, but I too encountered the problem via `schema`, so you're right that testing that may make more sense.

PrettyWood (approved):
LGTM!

PR #2172, opened 2020-12-03
Added expanded message for enum validation errors

fiebiga:
## Change Summary

This is a small change to add the attempted value that was not a valid member value to the exception string representation for an enum validation. This would have saved many hours of frustration in debugging when using pydantic with third party APIs

### DIFF
--- a/pydantic/errors.py
+++ b/pydantic/errors.py
@@ -223,7 +223,7 @@ class EnumMemberError(PydanticTypeError):
 
     def __str__(self) -> str:
         permitted = ', '.join(repr(v.value) for v in self.enum_values)  # type: ignore
-        return f'value <{self.attempted_value}> is not a valid enumeration member; permitted: {permitted}'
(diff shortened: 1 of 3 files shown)


Kludex:
Wouldn't that apply to email as well, for instance? I think the choice of not showing the data is on purpose :thinking:

fiebiga:
@Kludex Why would printing out an enum validation error apply to email?

Kludex:
Well, in a certain way, you're exposing the information that was not validated. I mean, I know nothing about the decisions made for the output message, but looks a similar choice to me.

fiebiga:
Are you basing this on the assumption that raw exceptions are allowed to be bubbled up to consumers of the application?

samuelcolvin:
tests are failing, the pull request template has not be completed, there's no change description and this duplicates a bigger changed scheduled for v2  - #1421, the original value will be included in `ctx` for all errors.

Issue #2129, opened 2020-11-17
Decompose Field components into Annotated


JacobHayes:
# Feature Request

PEP 593's `typing.Annotated` (or `typing_extensions.Annotated`) provides some scaffolding to do a lot of things that are currently handled with `Field`. This was recognized in #837, which suggested adding support for `var: Annotated[type, Field(...)]` in addition to the current `var: type = Field(...)`. I'm proposing we take this one step further and we decompose `Field` into a collection of classes (1 for each `Field` kwarg) that could be instantiated and used in `Annotated` directly. The benefits would be:
- the usage may be shorter (ie: `Annotated[int, Description("...")]` vs `Annotated[int, Field(description="...")]`
- they'd play/look nicer with non-`pydantic` metadata and could replace `**extra`
    - caveat: `**extra` are explicitly meant for `Field`, however `Annotated` values may not. If this is an issue, perhaps we can define a small interface/`FieldAnnotation` base class that could be use to check.
- we (maybe?) wouldn't have to deal with a default set here vs assigned if we didn't add a `Default` to replace the positional arg

One thing we may have to deal with is the ability for there to be _multiple_ instances of any of the classes and define some conflict resolution (likely "take last" or "take outer", however this manifests with `Annotated`). Arguably, we'd _also_ have to handle this for `Annotated[type, Field(...), Field(...)` too, but this decomposition of `Field` would make it much easier to override only a small piece (or perhaps support multiple, in the case of `Alias` for example).

Unless this were a v2 only change (not sure on timeline), we'll of course have to support both this and the existing `var: type = Field(...)` patterns. Separate from this vs `Annotated[type, Field(...)]`, we need to support `Annotated` metadata in general. I have no familiarity with the `pydantic` code base, but would be happy to help out with some guidance.

It seems like with a little extra work, we could replace all of the `con*` with these annotations, which would make mypy much happier. We might be able to do that in a backwards compatible-ish way if we make the `con*` calls wrappers returning the appropriate `Annotated` hint. Probably tricky for v1.

Example of "my custom code outside pydantic" if implemented:
```py
from typing import Annotated

from pydantic import GE, Alias, BaseModel, Description, FieldAnnotation


# Subclass FieldAnnotation so pydantic will identify the type. Perhaps these could define addition
# validation/checks on the field (eg: `count` below), similar to how GE would presumably work (or is that validation
# separate)?
class PII(BaseModel, FieldAnnotation):
    status: bool


[...]

PR #2176, opened 2020-12-04
Support Field(default_factory) in validate_arguments

### LINKED ISSUE #1359: Can Field been used as function default argument value?
# Feature Request

```
             pydantic version: 1.5a1
            pydantic compiled: True
                 install path: /home/xxx/.conda/envs/xxx/lib/python3.7/site-packages/pydantic
               python version: 3.7.4 (default, Aug 13 2019, 20:35:49)  [GCC 7.3.0]
                     platform: Linux-4.18.0-80.el8.x86_64-x86_64-with-centos-8.0.1905-Core
```
Hi, I am trying the function decorator validate_arguments. Is it possible use Field as the default argument value like this?
```py
from datetime import datetime
from pydantic import validate_arguments, Field

@validate_arguments
def foo(dt: datetime = Field(default_factory=datetime.now)):
    print(dt)

```

thomascobb:
## Change Summary

Add preliminary support for Field to validate_arguments

I expect there will be issues with my approach and edge cases, this is a starting point for a discussion

## Related issue number

fix #1359

## Checklist

### DIFF
--- a/changes/2176-thomascobb.md
+++ b/changes/2176-thomascobb.md
@@ -1,0 +1,2 @@
+Allow `Field` with a `default_factory` to be used as an argument to a function
+decorated with `validate_arguments`
--- a/docs/examples/validation_decorator_field.py
+++ b/docs/examples/validation_decorator_field.py
@@ -1,0 +1,22 @@
+from datetime import datetime
+from pydantic import validate_arguments, Field, ValidationError
+from pydantic.typing import Annotated
+
+
+@validate_arguments
+def how_many(num: Annotated[int, Field(gt=10)]):
+    return num
+
+
+try:
+    how_many(1)
+except ValidationError as e:
+    print(e)
+
+
+@validate_arguments
+def when(dt: datetime = Field(default_factory=datetime.now)):
+    return dt
+
+
+print(type(when()))
--- a/docs/usage/validation_decorator.md
+++ b/docs/usage/validation_decorator.md
@@ -55,6 +55,18 @@ To demonstrate all the above parameter types:
 ```
 _(This script is complete, it should run "as is")_
 
+## Using Field to describe function arguments
+
+[Field](schema.md#field-customisation) can also be used with `validate_arguments` to provide extra information about
+the field and validations. In general it should be used in a type hint with
+[Annotated](schema.md#typingannotated-fields), unless `default_factory` is specified, in which case it should be used
+as the default value of the field:
+
+```py
+{!.tmp_examples/validation_decorator_field.py!}
+```
+_(This script is complete, it should run "as is")_
+
 ## Usage with mypy
 
 The `validate_arguments` decorator should work "out of the box" with [mypy](http://mypy-lang.org/) since it's
@@ -93,13 +105,13 @@ _(This script is complete, it should run "as is")_
 
 ## Custom Config
 
-The model behind `validate_arguments` can be customised using a config setting which is equivalent to 
+The model behind `validate_arguments` can be customised using a config setting which is equivalent to
 setting the `Config` sub-class in normal models.
 
 !!! warning
     The `fields` and `alias_generator` properties of `Config` which allow aliases to be configured are not supported
     yet with `@validate_arguments`, using them will raise an error.
-    
+
 Configuration is set using the `config` keyword argument to the decorator, it may be either a config class
 or a dict of properties which are converted to a class later.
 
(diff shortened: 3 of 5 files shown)


samuelcolvin:
This looks like a good start, I'm almost unnerved by how simple the change is.

Are we sure there aren't any edge cases where behaviour has changed?

Has has this had a significant adverse effect on performance?

The other feature which would be cool here would be use of `Annotated[]`, see #2129, e.g. something like:

```py
@validate_arguments
def foo(dt: Annotated[datetime, Field(default_factory=datetime.now)]):
    print(dt)
```

@JacobHayes might have some input here.

Maybe that's a separate PR, but if it reduced the logic change I would personally prefer to only support `Annotated` and not `Field` as a default argument.

samuelcolvin:
Is it possible this would actually be fixed by #2147?

JacobHayes:
#2147 doesn't fix it alone - the model gets instantiated, but the `__fields_set__` filter removes the defaulted values, causing `TypeError: foo() missing 1 required positional argument: 'dt'`: https://github.com/samuelcolvin/pydantic/blob/8bad7bc91105616dfcc23ce1d68fba12f328a2a3/pydantic/decorator.py#L178-L179

Seems like there could be a few paths, but this small tweak on top of #2147 fixes things for the existing + new tests (moving default value -> Annotated in the new test):
```diff
diff --git a/pydantic/decorator.py b/pydantic/decorator.py
index 98bc4e2..85dfdd9 100644
--- a/pydantic/decorator.py
+++ b/pydantic/decorator.py
@@ -81,6 +81,7 @@ class ValidatedFunction:
             )
 
         self.raw_function = function
+        self.raw_function_parameters = parameters
         self.arg_mapping: Dict[int, str] = {}
         self.positional_only_args = set()
         self.v_args_name = 'args'
@@ -176,7 +177,9 @@ class ValidatedFunction:
         return values
 
     def execute(self, m: BaseModel) -> Any:
-        d = {k: v for k, v in m._iter() if k in m.__fields_set__}
+        d = {
+            k: v for k, v in m._iter() if k in m.__fields_set__ or (k in self.raw_function_parameters and v is not None)
+        }
         kwargs = d.pop(self.v_kwargs_name, None)
         if kwargs:
             d.update(kwargs)
```
I'm not sure how solid the `k in self.raw_function_parameters and v is not None` part is though, particularly if a `default_factory` may return `None` or something.

thomascobb:
We've already removed `default` from `Field` with the use of `Annotated`, how about removing `default_factory` too?
```python
@validate_arguments
def foo(dt: datetime = default_factory(datetime.now)):
    print(dt)
```
We would still need the Annotated type so we could use the other validation parts of `Field`:
```python
@validate_arguments
def foo(dt: Annotated[datetime, Field(description="What time is it")] = default_factory(datetime.now)):
    print(dt)
```
We could define default_factory as:
```python
@dataclass
class DefaultFactory:
    factory: Callable

def default_factory(factory: Callable[[], T]) -> T:
    # We will check for default values of type DefaultFactory in validate_arguments and call it at the right time
    df = DefaultFactory(factory)
    # But lie here so mypy thinks we actually passed a default value of the correct type
    return cast(T, df)
```

thomascobb:
@samuelcolvin I'm happy to discard this PR and have a go at making `default_factory` on top of #2147 if you think that's a good approach?

samuelcolvin:
Humm, I'm not sure about `DefaultFactory`, I think `Field(..., default_factory=datetime.now)` is more explicit and clearer for most people.

I'm not clear whether this PR should remain open or closed in favour of #2147? Or needs to wait for #2147 to be merged?

thomascobb:
I've rebased on top of master (and so #2147) and this PR now becomes much simpler. Most cases are handled with #2147, e.g. `foo(a: Annotated[int, Field(description="Something")] = 34):`. We still need a small tweak to support `default_factory`, which is included in this PR.

The issue with mypy still stands as noted in https://github.com/samuelcolvin/pydantic/pull/2176#issuecomment-759913046 

There are two fixes:

1. Pass `Field()` as default value e.g. `foo(a: int = Field(default_factory=lambda: 99)):`
2. Pass something with a matching type or Any in as the default e.g. `foo(a: Annotated[int, Field(default_factory=lambda: 99)] = ANY):`

Both work with this PR (and are in the test).

@samuelcolvin which of the 2 approaches should I put in the docs?

samuelcolvin:
This looks great.

@thomascobb, `foo(a: int = Field(default_factory=lambda: 99))` would be good in the docs.

samuelcolvin:
Great, thank you so much. 🎉

PR #2179, opened 2020-12-06
Propagate`ClassVar`s to sub-models

### LINKED ISSUE #2061: ClassVar inheritance
### Checks

# Feature Request

```
             pydantic version: 1.7
            pydantic compiled: False
                 install path: [...]/pydantic
               python version: 3.9.0 (default, Oct  8 2020, 06:18:59)  [Clang 7.1.0 (tags/RELEASE_710/final)]

```

Currently, it is possible to annotate an attribute as being a class variable and it will be excluded from the model:

```py
from typing import ClassVar

class Foo(BaseModel):
    class_var_to_be_redefined_on_children: ClassVar[str] = "value"


class Bar(Foo):
    class_var_to_be_redefined_on_children = "some other value"

```

However, when inheriitng from a model with a `ClassVar`, overwriting the class variable produces an error:

```
  File "test.py", line 10, in <module>
    class Bar(Foo):
  File "pydantic/main.py", line 284, in __new__
    validate_field_name(bases, var_name)
  File "utils.py", line 144, in validate_field_name
    raise NameError(
NameError: Field name "class_var_to_be_redefined_on_children" shadows a BaseModel attribute; use a different field name with "alias='class_var_to_be_redefined_on_children'".
```

What appears to be happening here is that Pydantic assumes that `class_var_to_be_redefined_on_children` is a Pydantic field but errors out because there is a class attribute with the same name.  I would prefer not to re-annotate subclass class variables with `ClassVar[T]`.  I believe the following should also be interpret as a `ClassVar`, especially in conjunction with generic models (an uninitialised `ClassVar`):

```py
from typing import ClassVar

class Foo(BaseModel):
    class_var_to_be_defined_on_children: ClassVar[str]


class Bar(Foo):
    class_var_to_be_defined_on_children = "some other value"
```

layday:
## Change Summary

Currently, if a `ClassVar` is defined on a model and re-defined
on a sub-model omitting the `ClassVar` annotation, Pydantic produces an
unrelated error:

    NameError: Field name "..." shadows a BaseModel attribute ...

This check was introduced to prevent shadowing Pydantic's own methods
and attributes defined on the `BaseModel` class.  Following this change,
class variables (that is, variables annotated with `ClassVar`)
defined on parent models will be inherited by sub-models and
will be overwritable without having to reapply the annotation.

## Related issue number

fix #2061

## Checklist

### DIFF
--- a/changes/2061-layday.md
+++ b/changes/2061-layday.md
@@ -1,0 +1,1 @@
+allow overwriting `ClassVar`s in sub-models without having to re-annotate them
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -531,9 +531,15 @@ Where `Field` refers to the [field function](schema.md#field-customisation).
     Moreover if you want to validate default values with `validate_all`,
     *pydantic* will need to call the `default_factory`, which could lead to side effects!
 
+## Automatically excluded attributes
+
+Class variables which begin with an underscore and attributes annotated with `typing.ClassVar` will be
+automatically excluded from the model.
+
 ## Private model attributes
 
-If you need to use internal attributes excluded from model fields, you can declare them using `PrivateAttr`:
+If you need to vary or manipulate internal attributes on instances of the model, you can declare them
+using `PrivateAttr`:
 
 ```py
 {!.tmp_examples/private_attributes.py!}
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -16,7 +16,6 @@ from typing import (
     List,
     Mapping,
     Optional,
-    Set,
     Tuple,
     Type,
     TypeVar,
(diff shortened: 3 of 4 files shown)


layday:
I've also updated the documentation to explain what `PrivateAttr` is for, which I thought was unclear.

samuelcolvin:
Humm, I'm not sure about this, I think not using the `ClassVar` annotation on the sub-class is a recipe for enormous confusion. Even for those who understand how pydantic works, let alone those who are new to it.

Personally I would argue that developers should be forced to always use `ClassVar[]`even if they're overwriting an attribute which is a class var on the parent model.

But if many people feel otherwise I'm happy to be overruled, also I think a more descriptive error message would be good.

layday:
If I were to re-annotate subclasses, I would have to parametrise the `ClassVar` for type completeness - type checkers cannot perform 'partial' inference - annotations must be complete.  This results in a whole lot of repetition, even if I were to use a type alias, for what my type checker is able to infer statically via inheritence.  That Pydantic would not honour the class variable is more surprising to me, but YMMV :)

samuelcolvin:
Okay, I'll trust you on this one 😄 .

samuelcolvin:
Thanks a lot.

samuelcolvin:
Question on this in https://github.com/pydantic/pydantic/discussions/4518.

PR #2181, opened 2020-12-07
fix: ensure to always return one of the values in `Literal` field type

### LINKED ISSUE #2166: Literal enum fields are stored as mix-in type instead of enum field when using parse_raw or parse_json
### Checks

* [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this feature/change is needed
* [X] After submitting this, I commit to one of:

# Feature Request

```
             pydantic version: 1.7.1
            pydantic compiled: True
                 install path: /home/user/project/.venv/lib/python3.8/site-packages/pydantic
                     platform: Linux-5.4.0-53-generic-x86_64-with-glibc2.29
```

```py
import enum
from typing import Literal

class Bar(str, enum.Enum):
    FIZ = "fiz"
    FUZ = "fuz"

class Foo(BaseModel):
    bar: Bar
    barfiz: Literal[Bar.FIZ]

Foo.parse_raw('{"bar": "fiz", "barfiz": "fiz"}')
Foo.parse_obj({"bar": "fiz", "barfiz": "fiz"})
# both return: Foo(bar=<Bar.FIZ: 'fiz'>, barfiz='fiz')

# expected: Foo(bar=<Bar.FIZ: 'fiz'>, barfiz=<Bar.FIZ: 'fiz'>)
```

I was expecting that the Pydantic would store `Bar.FIZ` for the `Literal[Bar.FIZ]` field above, as opposed to storing a `str` type (the enum's mix-in type. Mypy flags `Literal[Bar.FIZ]` above as an enum field, so accessing `value` on the `barfiz` attribute, for example, is perfectly fine by Mypy's rules, but breaks during runtime.

I realize that mypy's support for Literal enum fields is relatively new, but they have interesting properties that I am looking to make use of (I would love to see this work with [discriminated union types](https://github.com/samuelcolvin/pydantic/issues/619), for example).

Would you consider adding a Config field to enable this behaviour? Should this be flagged as a bug instead and targetted for a 2.0 release?

Possibly related bugs:
* https://github.com/samuelcolvin/pydantic/issues/1086
* https://github.com/samuelcolvin/pydantic/issues/1747

PrettyWood:
## Change Summary
Ensure we always return a value set in `Literal` even though other values are equal (for example a `StrEnum` where a string **can be equal to a enum member but not identical**)

## Related issue number
closes #2166

## Checklist

### DIFF
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -443,9 +443,13 @@ def make_literal_validator(type_: Any) -> Callable[[Any], Any]:
     allowed_choices_set = set(permitted_choices)
 
     def literal_validator(v: Any) -> Any:
-        if v not in allowed_choices_set:
+        # Ensure to return the value in `Literal`, not the passed value
+        # In some cases they can indeed be different (see `test_literal_validator_str_enum`)
+        for x in allowed_choices_set:
+            if v == x:
+                return x
+        else:
             raise errors.WrongConstantError(given=v, permitted=permitted_choices)
-        return v
 
     return literal_validator
 
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -440,15 +440,17 @@ def int_enum_validator(v: Any) -> IntEnum:
 
 def make_literal_validator(type_: Any) -> Callable[[Any], Any]:
     permitted_choices = all_literal_values(type_)
-    allowed_choices_set = set(permitted_choices)
+
+    # To have a O(1) complexity and still return one of the values set inside the `Literal`,
+    # we create a dict with the set values (a set causes some problems with the way intersection works).
+    # In some cases the set value and checked value can indeed be different (see `test_literal_validator_str_enum`)
+    allowed_choices = {v: v for v in permitted_choices}
 
     def literal_validator(v: Any) -> Any:
-        # Ensure to return the value in `Literal`, not the passed value
-        # In some cases they can indeed be different (see `test_literal_validator_str_enum`)
-        intersection = {v} & allowed_choices_set  # order of operands matters
-        if not intersection:
+        try:
+            return allowed_choices[v]
+        except KeyError:
             raise errors.WrongConstantError(given=v, permitted=permitted_choices)
-        return intersection.pop()
 
     return literal_validator
 
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -445,11 +445,10 @@ def make_literal_validator(type_: Any) -> Callable[[Any], Any]:
     def literal_validator(v: Any) -> Any:
         # Ensure to return the value in `Literal`, not the passed value
         # In some cases they can indeed be different (see `test_literal_validator_str_enum`)
-        for x in allowed_choices_set:
-            if v == x:
-                return x
-        else:
+        intersection = {v} & allowed_choices_set  # order of operands matters
+        if not intersection:
             raise errors.WrongConstantError(given=v, permitted=permitted_choices)
+        return intersection.pop()
 
     return literal_validator
 
--- a/changes/2166-PrettyWood.md
+++ b/changes/2166-PrettyWood.md
@@ -1,0 +1,1 @@
+fix: ensure to always return one of the values in `Literal` field type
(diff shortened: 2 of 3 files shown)


pydantic/validators.py
```diff
@@ -443,9 +443,13 @@ def make_literal_validator(type_: Any) -> Callable[[Any], Any]:
...
     def literal_validator(v: Any) -> Any:
-        if v not in allowed_choices_set:
+        # Ensure to return the value in `Literal`, not the passed value
+        # In some cases they can indeed be different (see `test_literal_validator_str_enum`)
+        for x in allowed_choices_set:
+            if v == x:
+                return x
+        else:
             raise errors.WrongConstantError(given=v, permitted=permitted_choices)
-        return v
```
yobiscus, line 448: I think this bit of code is equivalent, but without the O(n)-ness.

```python
        # Ensure to return the value in `Literal`, not the passed value
        # In some cases they can indeed be different (see `test_literal_validator_str_enum`)
        intersection = {v} & allowed_choices_set  # order of operands matters
        if not intersection:
        return intersection.pop()
```

PrettyWood: You're right! Excellent thanks a bunch 👍

layday: I suggest adding a test for multi-value `Literal`s because I suspect the order wouldn't be respected for all intersections:

```py
from enum import Enum

class Foo(str, Enum):
    bar = "bar"

assert ({"bar"} & {Foo.bar}).pop() is Foo.bar
assert ({"bar"} & {Foo.bar, "baz"}).pop() is Foo.bar
```

PrettyWood: You're absolutely right @layday! There is indeed a weird behaviour with intersection.
```python
from pydantic.typing import all_literal_values, Literal


class Bar(str, Enum):
    FIZ = 'fiz'
    FUZ = 'fuz'

type_ = Literal[Bar.FIZ, Bar.FUZ]

allowed_choices_set = set(permitted_choices)
intersection = allowed_choices_set & {'fiz'}
# intersection == {'fiz'} !!!!
```

@yobiscus I put back my old implementation, which works with many values in `Literal`.

layday: I suppose you could use a dictionary instead, that will still be O(1)-like:

```py
choices = {v: v for v in values_in_literal}

try:
     return choices[value]
except KeyError:
    raise ValueError('invalid')
```

yobiscus: @PrettyWood 

```python
allowed_choices_set = set(permitted_choices)
intersection = allowed_choices_set & {'fiz'}
# intersection == {'fiz'} !!!!
```

This is because the order of the operands matter. If you flip the operands to `intersection = {'fiz'} & allowed_choices_set`, the correct result is returned.

@layday This explains why your example is returning the "wrong" value as well.

layday: That's not what I'm seeing:

```py
In [1]: %paste
from enum import Enum

class Foo(str, Enum):
    bar = "bar"

## -- End pasted text --

In [2]: {"bar"} & {Foo.bar}
Out[2]: {<Foo.bar: 'bar'>}

In [3]: {Foo.bar} & {"bar"}
Out[3]: {'bar'}

In [4]: {Foo.bar, "baz"} & {"bar"}
Out[4]: {'bar'}

In [5]: {"bar"} & {Foo.bar, "baz"}
Out[5]: {'bar'}
```

layday: I'm guessing CPython is optimised to pick the value from the shortest set, preferring the RHS operand.  This is definitely not something we'd want to rely on.

PrettyWood: I have the same as @layday ! Very strange... Does it depend on the python version? 🤔 
Anyway switching to dict always works ! Updated the last commit thanks 👍

yobiscus: @layday Thanks for the clarification. I see that now! Bummer.

```python
allowed_choices = {v: v for v in permitted_choices}
```

This always O(n) now, isn't it? But it's possible that CPython's dict comprehension is faster than a for-loop. Not sure about that though.

PrettyWood: We create the `allowed_choices` *before* in the ModelMetaclass so the validation done by `literal_validator` is still in O(1) ;)

yobiscus: Right! Well this solution is much more intuitive than the obscure set intersection one. Nice work!

PR #2182, opened 2020-12-07
fix: stdlib dataclass converted into pydantic dataclass still equal its stdlib dataclass equivalent

### LINKED ISSUE #2162: Pydantic `dataclasses` break `__eq__` method generated by stdlib `dataclasses` when upgrading from v1.6.1 to v1.7.2
### Checks

# Bug

```
             pydantic version: 1.7.2
            pydantic compiled: True
                 install path: /home/jtrh/.virtualenvs/tmp-efd3764f8d079be/lib/python3.8/site-packages/pydantic
                     platform: Linux-5.4.0-54-generic-x86_64-with-glibc2.29
     optional deps. installed: []
```

```py
import dataclasses
import unittest

import pydantic
import pydantic.dataclasses


class PydanticConfig:


@dataclasses.dataclass(frozen=True)
class StdLibFoo:
    a: str
    b: int


@pydantic.dataclasses.dataclass(frozen=True)
class PydanticFoo:
    a: str
    b: int


@dataclasses.dataclass(frozen=True)
class StdLibBar:
    c: StdLibFoo


@pydantic.dataclasses.dataclass(frozen=True, config=PydanticConfig)
class PydanticBar:
    c: PydanticFoo


@dataclasses.dataclass(frozen=True)
class StdLibBaz:
    c: PydanticFoo


@pydantic.dataclasses.dataclass(frozen=True, config=PydanticConfig)
class PydanticBaz:
    c: StdLibFoo


class DataClassTest(unittest.TestCase):
    def _assert(self, foo, bar):
        self.assertEqual(dataclasses.asdict(foo), dataclasses.asdict(bar.c))
        self.assertEqual(dataclasses.astuple(foo), dataclasses.astuple(bar.c))
        self.assertEqual(foo, bar.c)

    def test_stdlib_only(self):
        foo = StdLibFoo(a='Foo', b=1)
        bar = StdLibBar(c=foo)

        self._assert(foo, bar)

    def test_pydantic_only(self):
        foo = PydanticFoo(a='Foo', b=1)
        bar = PydanticBar(c=foo)

        self._assert(foo, bar)

    def test_pydantic_and_stdlib(self):
        foo = PydanticFoo(a='Foo', b=1)
        bar = StdLibBaz(c=foo)

        self._assert(foo, bar)


    def test_stdlib_and_pydantic(self):
        foo = StdLibFoo(a='Foo', b=1)
        bar = PydanticBaz(c=foo)

        self._assert(foo, bar)


if __name__ == '__main__':
    print(pydantic.version.version_info()); print(); unittest.main(verbosity=2)
```

**Output with `pydantic==1.7.2`**:

```
             pydantic version: 1.7.2
            pydantic compiled: True
                 install path: /home/jtrh/.virtualenvs/tmp-efd3764f8d079be/lib/python3.8/site-packages/pydantic
                     platform: Linux-5.4.0-54-generic-x86_64-with-glibc2.29
     optional deps. installed: []

test_pydantic_and_stdlib (__main__.DataClassTest) ... ok
test_pydantic_only (__main__.DataClassTest) ... ok
test_stdlib_and_pydantic (__main__.DataClassTest) ... FAIL
test_stdlib_only (__main__.DataClassTest) ... ok

======================================================================
FAIL: test_stdlib_and_pydantic (__main__.DataClassTest)
----------------------------------------------------------------------
  File "pydantic_dataclass_equality_error_tests_2020_11_30.py", line 73, in test_stdlib_and_pydantic
    self._assert(foo, bar)
  File "pydantic_dataclass_equality_error_tests_2020_11_30.py", line 48, in _assert
    self.assertEqual(foo, bar.c)
AssertionError: StdLibFoo(a='Foo', b=1) != StdLibFoo(a='Foo', b=1)

----------------------------------------------------------------------
Ran 4 tests in 0.001s

FAILED (failures=1)
```

**Output with `pydantic==1.6.1`**:

```
             pydantic version: 1.6.1
            pydantic compiled: True
                 install path: /home/jtrh/.virtualenvs/tmp-75057fed9773783/lib/python3.8/site-packages/pydantic
                     platform: Linux-5.4.0-54-generic-x86_64-with-glibc2.29
     optional deps. installed: []

test_pydantic_and_stdlib (__main__.DataClassTest) ... ok
test_pydantic_only (__main__.DataClassTest) ... ok
test_stdlib_and_pydantic (__main__.DataClassTest) ... ok
test_stdlib_only (__main__.DataClassTest) ... ok

----------------------------------------------------------------------
Ran 4 tests in 0.001s

OK
```

I think this issue may be related to the automatic conversion of standard library `dataclasses` to Pydantic `dataclasses`. The `__eq__` method generated by the standard library checks that both instances are of the same class (see [`dataclasses._cmp_fn`](https://github.com/python/cpython/blob/v3.8.5/Lib/dataclasses.py#L581) and [`dataclasses._process_class`](https://github.com/python/cpython/blob/v3.8.5/Lib/dataclasses.py#L944-L953)), which is true in 1.6.1 (`assert foo.__class__ is bar.c.__class__` succeeds), but not in 1.7.2 (`assert foo.__class__ is bar.c.__class__` fails, but `assert issubclass(bar.c.__class__, foo.__class__)` succeeds).

PrettyWood:
## Related issue number
closes #2162

## Checklist

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -110,6 +110,15 @@ def _process_class(
         if post_init_post_parse is not None:
             post_init_post_parse(self, *initvars)
 
+    def _converted_pydantic_dataclass_eq(self: 'Dataclass', other: Any) -> bool:
+        """
+        To still support equality between a stdlib dataclass and its converted pydantic dataclass
+        we add a custom `__eq__` method on the converted pydantic dataclass
+        """
+        stdlib_dc = self.__class__.__bases__[0]
+        dc_fields = {k: v for k, v in self.__dict__.items() if k != '__initialised__'}
+        return stdlib_dc(**dc_fields) == other
+
     # If the class is already a dataclass, __post_init__ will not be called automatically
     # so no validation will be added.
     # We hence create dynamically a new dataclass:
@@ -129,6 +138,7 @@ def _process_class(
             (_cls,),
             {
                 '__annotations__': _cls.__annotations__,
+                '__eq__': _converted_pydantic_dataclass_eq,
                 '__post_init__': _pydantic_post_init,
                 # attrs for pickle to find this class
                 '__module__': __name__,
--- a/changes/2162-PrettyWood.md
+++ b/changes/2162-PrettyWood.md
@@ -1,0 +1,1 @@
+fix: stdlib `dataclass` converted into _pydantic_ `dataclass` still equals its stdlib `dataclass` equivalent
--- a/docs/examples/dataclasses_stdlib_to_pydantic_equality.py
+++ b/docs/examples/dataclasses_stdlib_to_pydantic_equality.py
@@ -1,0 +1,14 @@
+import dataclasses
+
+import pydantic
+
+
+@dataclasses.dataclass
+class StdlibUser:
+    name: str
+    age: int
+
+
+PydanticUser = pydantic.dataclasses.dataclass(StdlibUser)
+
+assert StdlibUser(name='pika', age=7) == PydanticUser(name='pika', age='7')
--- a/docs/usage/dataclasses.md
+++ b/docs/usage/dataclasses.md
@@ -59,6 +59,14 @@ them with `pydantic.dataclasses.dataclass`.
 ```
 _(This script is complete, it should run "as is")_
 
+!!! note
+    A _pydantic_ dataclass converted from a stdlib dataclass will still equal its stdlib dataclass equivalent
+
+```py
+{!.tmp_examples/dataclasses_stdlib_to_pydantic_equality.py!}
+```
+_(This script is complete, it should run "as is")_
+
 ### Inherit from stdlib dataclasses
 
 Stdlib dataclasses (nested or not) can also be inherited and _pydantic_ will automatically validate
--- a/tests/test_dataclasses.py
+++ b/tests/test_dataclasses.py
@@ -3,7 +3,7 @@ import pickle
 from collections.abc import Hashable
 from datetime import datetime
 from pathlib import Path
-from typing import ClassVar, Dict, FrozenSet, List, Optional
+from typing import Any, ClassVar, Dict, FrozenSet, Generator, List, Optional, Tuple
 
 import pytest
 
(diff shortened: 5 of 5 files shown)


pydantic/dataclasses.py
```diff
@@ -110,6 +110,15 @@ def _pydantic_post_init(self: 'Dataclass', *initvars: Any) -> None:
         if post_init_post_parse is not None:
             post_init_post_parse(self, *initvars)
 
+    def pydantic_dataclass_eq(self: 'Dataclass', other: Any) -> bool:
```
PrettyWood, line 113: This currently works **only** for automatically converted dataclasses.
This is to remove any regression due to the feature added in v1.7!
Maybe it could make sense to also support it natively like
```python
import dataclasses

import pydantic


class PydanticDataclass:
    a: int
    b: float

@dataclasses.dataclass
class StdlibDataclass:
    a: int
    b: float


py_dc = PydanticDataclass(a=123, b=1.1)
std_dc = StdlibDataclass(a=123, b=1.1)
assert py_dc == std_dc
```

If we want to do this, this could be a new feature

samuelcolvin:
Humm, my only reservation is that standard python dataclasses to clever stuff so that different dataclasses with the same fields (and values) are not equal:

```py

@dataclass(frozen=True)
class Foo:
    a: int
    b: float

@dataclass(frozen=True)
class Bar:
    a: int
    b: float

f = Foo(a=10, b=1.0)
b = Bar(a=10, b=1.0)
debug(f == b)  #> False

f2 = Foo(a=10, b=1.0)
debug(f == f2)  #> True
```

Isn't there a risk that different dataclasses incorrectly being equal is worse than them incorrectly being unequal?

I know "correct" and "incorrect" are subjective here and some might argue (very reasonably) that `f` and `b` are the same above, but I think since dataclasses are a python thing we should do our best to match python behaviour.

If we can't reliably match python's standard behaviour, I'd be happy to just raise an error some people don't assume stuff will work which won't.

samuelcolvin:
I think it shouldn't be that hard to copy python's own logic more or less verabim:

https://github.com/python/cpython/blob/ec3165320e81ac87edcb85c86c452528ddbaec1c/Lib/dataclasses.py#L964-L970

jtrh:
> I think it shouldn't be that hard to copy python's own logic more or less verabim:
> 
> https://github.com/python/cpython/blob/1e5d33e9b9b8631b36f061103a30208b206fd03a/Lib/dataclasses.py#L949-L955

I agree that copying Python's own logic is a good idea, but how would you handle the class equality check? Read below for details.

Near the end of that code, Python [calls `_cmp_fn()`](https://github.com/python/cpython/blob/1e5d33e9b9b8631b36f061103a30208b206fd03a/Lib/dataclasses.py#L953-L955) to create the comparison function, which starts by [checking for class equality](https://github.com/python/cpython/blob/v3.8.5/Lib/dataclasses.py#L581) (`if other.__class__ is self.__class__:`) before even taking a look at the fields themselves.

@samuelcolvin Do you think that replacing the class equality check with something like `if other.__class__ is self.__class__ or other.__class__ is self.__class__.__bases__[0]:` would be reliable enough?

samuelcolvin:
> @samuelcolvin Do you think that replacing the class equality check with something like `if other.__class__ is self.__class__ or other.__class__ is self.__class__.__bases__[0]:` would be reliable enough?

Yes, I think makes sense.

PrettyWood:
FYI I thought again about all the issues related to stdlib dataclass to validated dataclass
I started a draft on implementation in ttps://github.com/samuelcolvin/pydantic/pull/2557.
Feedback more than welcome! 🙏

PrettyWood:
Will be tackled in #2557

PR #2183, opened 2020-12-07
Feature/add anystr lower to config

### LINKED ISSUE #2134: Adding "anystr_lower" to config
# Feature Request

```
             pydantic version: 1.6.1
            pydantic compiled: True
                 install path: ***
               python version: 3.8.0 (v3.8.0:fa919fdf25, Oct 14 2019, 10:23:27)  [Clang 6.0 (clang-600.0.57)]

```
Just like there is "anystr_strip_whitespace" in the config, which runs .strip() on any string value, I would like to add "anystr_lower" which would run .lower() on any string value.

I just find myself use .lower() on the strings in many validators in places in my code, so I thought it would be useful to add it to the config, I assume that was the use-case of "anystr_strip_whitespace" as well.

If this is fine I'll gladly implement it myself, looks like it's a small change.

tayoogunbiyi:
## Change Summary

This change adds the anystr_lower field to the BaseConfig class, allowing users to ensure all string fields are kept as lower case.

## Related issue number

fix #2134

## Checklist

### DIFF
--- a/changes/2134-tayoogunbiyi.md
+++ b/changes/2134-tayoogunbiyi.md
@@ -1,1 +1,1 @@
-* added `Config.anystr_lower` and `lower_str` kwarg to `constr`.
+added `Config.anystr_lower` and `lower_str` kwarg to `constr`.
--- a/changes/2134-tayoogunbiyi.md
+++ b/changes/2134-tayoogunbiyi.md
@@ -1,0 +1,1 @@
+* added `Config.anystr_lower` and `lower_str` kwarg to `constr`.
--- a/changes/2134-tayoogunbiyi.md
+++ b/changes/2134-tayoogunbiyi.md
@@ -1,1 +1,1 @@
-added `Config.anystr_lower` and `lower_str` kwarg to `constr`.
+added `Config.anystr_lower` and `to_lower` kwarg to `constr`.
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -8,6 +8,9 @@ Options:
 **`anystr_strip_whitespace`**
 : whether to strip leading and trailing whitespace for str & byte types (default: `False`)
 
+**`anystr_lower`**
+: whether to make all characters lowercase for str & byte types (default: `False`)
+
 **`min_anystr_length`**
 : the min length for str & byte types (default: `0`)
 
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -727,7 +727,7 @@ The following arguments are available when using the `constr` type function
 - `strict: bool = False`: controls type coercion
 - `min_length: int = None`: minimum length of the string
 - `max_length`: maximum length of the string
-- `curtail_length`: restricts the strings length
+- `curtail_length`: shrinks the string length to the set value when it is longer than the set value
 - `regex: str = None`: regex to validate the string against
 
 ### Arguments to `conbytes`
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -722,21 +722,21 @@ Where `Field` refers to the [field function](schema.md#field-customisation).
 ### Arguments to `constr`
 The following arguments are available when using the `constr` type function
 
--  `strip_whitespace: bool = False`: removes leading and trailing whitespace
-- `to_lower: bool = False`: turns all characters to lowercase
-- `strict: bool = False`: controls type coercion
-- `min_length: int = None`: minimum length of the string
-- `max_length`: maximum length of the string
-- `curtail_length`: shrinks the string length to the set value when it is longer than the set value
-- `regex: str = None`: regex to validate the string against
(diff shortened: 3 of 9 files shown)


pydantic/validators.py
```diff
@@ -472,6 +476,13 @@ def constr_strip_whitespace(v: 'StrBytes', field: 'ModelField', config: 'BaseCon
     return v
 
 
+def constr_lower(v: 'StrBytes', field: 'ModelField', config: 'BaseConfig') -> 'StrBytes':
+    lower = field.type_.lower_str or config.anystr_lower
+    if lower:
+        v = v.strip()
```
PrettyWood, line 482:
```suggestion
        v = v.lower()
```

PrettyWood: You should probably add a test for `constr`

tayoogunbiyi: thanks. I somehow missed that :+1:

changes/2134-tayoogunbiyi.md
```diff
@@ -0,0 +1 @@
+* added `Config.anystr_lower` and `lower_str` kwarg to `constr`.
```
samuelcolvin, line 1:
```suggestion
added `Config.anystr_lower` and `lower_str` kwarg to `constr`.
```

pydantic/types.py
```diff
@@ -112,6 +113,7 @@
 
 class ConstrainedBytes(bytes):
     strip_whitespace = False
+    lower_str = False
```
samuelcolvin, line 116:
```suggestion
    to_lower = False
```

and below.

I think `_str` is tautologous in this context.

`lower` (or `lower_str`) could be confusing, it could be interpreted as meaning the input string must be lowercase, hence I think `to_lower` would be best.

docs/usage/model_config.md
```diff
@@ -8,6 +8,9 @@ Options:
 **`anystr_strip_whitespace`**
 : whether to strip leading and trailing whitespace for str & byte types (default: `False`)
 
+**`anystr_lower`**
+: whether to make all characters lowercase for str & byte types (default: `False`)
```
samuelcolvin, line 12: we should also add a comment about this to [the docs](https://pydantic-docs.helpmanual.io/usage/types/#constrained-types).

Best of all would be a description of all the arguments to `constr` like we do [for URLs](https://pydantic-docs.helpmanual.io/usage/types/#urls)

samuelcolvin (commented):
thanks so much, this looks good.

Just a few small things to fix.

docs/usage/model_config.md
tayoogunbiyi: Done. Let me know what you think

tayoogunbiyi: I'm not sure if this is the best place to ask this but why do we need `curtail_length` when we have `max_length`? 

They seem to do the same thing except I don't fully understand it.

PrettyWood: `curtail_length` does not raise an Error if the string is too long but just shrinks it to the set value

tayoogunbiyi: I see. thanks.

docs/usage/types.md
```diff
@@ -719,6 +719,25 @@ _(This script is complete, it should run "as is")_
 
 Where `Field` refers to the [field function](schema.md#field-customisation).
 
+### Arguments to `constr`
+The following arguments are available when using the `constr` type function
+
+-  `strip_whitespace: bool = False`: removes leading and trailing whitespace
```
PrettyWood, line 725:
```suggestion
- `strip_whitespace: bool = False`: removes leading and trailing whitespace
```
Same under.
Also maybe better to add type and default value for `max_length` and `curtail_length` to be homogeneous

changes/2134-tayoogunbiyi.md
```diff
@@ -0,0 +1 @@
+added `Config.anystr_lower` and `to_lower` kwarg to `constr`.
```
PrettyWood, line 1:
```suggestion
added `Config.anystr_lower` and `to_lower` kwarg to `constr` and `conbytes`.
```

docs/usage/types.md
PrettyWood: I meant to set only 1 space after `-` instead of 2 `strip_whitespace`.
Without space at all it won't be interpreted correctly

PrettyWood: Other than that looks good to me :) I guess we should to do the same for `confloat`, `conint`, `condecimal`, `conset` and `conlist` in another PR

tayoogunbiyi: > Other than that looks good to me :) I guess we should to do the same for `confloat`, `conint`, `condecimal`, `conset` and `conlist` in another PR

I guess I'd do this in another PR once this is settled :)

PrettyWood (approved):
LGTM!

PR #2190, opened 2020-12-09
fix(settings): stop using `Config.env_prefix` for secrets in `BaseSettings`

### LINKED ISSUE #2188: BaseSettings env_prefix also apply for secrets
### Checks

# Bug
The `env_prefix` is used for both env vars and secrets, which makes it impossible to combine env vars (with prefix) and secrets (without prefix) into a single settings model.

```
             pydantic version: 1.7.2
            pydantic compiled: False
                 install path: /usr/local/src/pydantic/pydantic
               python version: 3.7.7 (default, Mar 11 2020, 00:27:03)  [GCC 8.3.0]
                     platform: Linux-5.4.0-54-generic-x86_64-with-debian-10.3
     optional deps. installed: []
```

Environment vars:
```
printenv
RESTAPI_USE_PRODUCTION_DB=1
RESTAPI_DB_READ_ONLY=0
```
Secrets in container:
```
ls /run/secrets/
db_read_only_password  db_read_only_user  db_read_write_password  db_read_write_user
```

```py

class DBSettings(BaseSettings):
    use_production_db: bool = False     # env var
    db_read_only: bool = False          # env var
    db_read_only_user: str = None       # secret
    db_read_only_password: str = None   # secret
    db_read_write_user: str = None      # secret
    db_read_write_password: str = None  # secret

    class Config:
        env_prefix = 'RESTAPI_'
        secrets_dir = '/run/secrets'

db_sett = DBSettings()
```
Tested without the `env_prefix = 'RESTAPI_'` line, which filled the settings from the secrets correctly, env vars were not read as expected.

### Possible solutions
- Make it clear in the documentation that the `env_prefix` is also used for secrets
- Just do not use the `env_prefix` for the secrets
- Make an `secrets_prefix` to be used as secrets prefix

PrettyWood:
## Change Summary
We were using the `env_prefix` for secrets, which doesn't make sense and is very confusing.
Now we just don't

## Related issue number
closes #2188 

## Checklist

### DIFF
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -66,15 +66,14 @@ class BaseSettings(BaseModel):
             raise SettingsError(f'secrets_dir must reference a directory, not a {path_type(secrets_path)}')
 
         for field in self.__fields__.values():
-            for env_name in {field.name}:
-                path = secrets_path / env_name
-                if path.is_file():
-                    secrets[field.alias] = path.read_text().strip()
-                elif path.exists():
-                    warnings.warn(
-                        f'attempted to load secret file "{path}" but found a {path_type(path)} instead.',
-                        stacklevel=4,
-                    )
+            path = secrets_path / field.name
+            if path.is_file():
+                secrets[field.alias] = path.read_text().strip()
+            elif path.exists():
+                warnings.warn(
+                    f'attempted to load secret file "{path}" but found a {path_type(path)} instead.',
+                    stacklevel=4,
+                )
 
         return secrets
 
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -66,7 +66,7 @@ class BaseSettings(BaseModel):
             raise SettingsError(f'secrets_dir must reference a directory, not a {path_type(secrets_path)}')
 
         for field in self.__fields__.values():
-            for env_name in field.field_info.extra['env_names']:
+            for env_name in {field.name}:
                 path = secrets_path / env_name
                 if path.is_file():
                     secrets[field.alias] = path.read_text().strip()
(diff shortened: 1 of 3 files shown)


pydantic/env_settings.py
```diff
@@ -66,7 +66,7 @@ def _build_secrets_files(self, _secrets_dir: Union[Path, str, None] = None) -> D
             raise SettingsError(f'secrets_dir must reference a directory, not a {path_type(secrets_path)}')
 
         for field in self.__fields__.values():
-            for env_name in field.field_info.extra['env_names']:
+            for env_name in {field.name}:
```
PrettyWood: split into 2 commits just to ease review

PrettyWood: We could for extra caution do `{field.name} | field.field_info.extra['env_names']` to be really sure nothing gets broken but imo it's really a bug that isn't expected

samuelcolvin:
I think this is great but we need:
* a prominent `!note` in the docs explaining this
* a `**breaking change:**` prefix on the description

PrettyWood:
Added the "breaking change" mention and a note in the doc
Screen Shot 2021-01-02 at 4 30 58 PM
Hope it's clear enough

samuelcolvin:
Looks good to me, but sadly this has some fairly complicated conflicts.

samuelcolvin:
Wow this is old :see_no_evil:.

@PrettyWood what do you think? V1.10, V2 or close.

PrettyWood:
I would close as much as possible and avoid a breaking change

samuelcolvin:
But should we keep this for V2?

PrettyWood:
> the env_prefix for secrets, which doesn't make sense and is very confusing.

I still agree with myself 😄 If we can change this in v2, that will be great! We can keep this open not to forget

samuelcolvin:
Great, keeping it open.

samuelcolvin:
Closing this, as any new PR would need to be against https://github.com/pydantic/pydantic-settings

PR #2193, opened 2020-12-10
feat: avoid reconstructing models used as fields of another model on validation

### LINKED ISSUE #265: Inherited models as fields are reconstructed as base model
I have a model which has a field of a another base-class model. I want this field to be able to contain any sub-class of this model. But when I assign to this field it gets reconstructed to the base model.
```py
from pydantic import BaseModel, validator, PyObject, ValidationError, BaseConfig

class Foo(BaseModel):
    a: int


class FooSub(Foo):
    b: str


class Container(BaseModel):
    foo: Foo

f = Foo(a=1)
c1 = Container(foo=f)
# print(f'c1: {c1}')


fs = FooSub(a=2, b='hello')
c2 = Container(foo=fs)

# print(f'c2: {c2}')
print(f'fs: {fs} (id: {id(fs)})')
print(f'c2.foo: {c2.foo} ({id(c2.foo)})')
```

this prints

```
fs: FooSub a=2 b='hello' (id: 4566162824)
c2.foo: Foo a=2 (4566162696)
```

What I want is `c2.foo` to be the `FooSub` instance I created and assigned. How can I achieve this? I have tried using custom validators, but without success.

### LINKED ISSUE #1246: Objects are copied duing validation
# Bug

```
$ python3 -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.4a1
            pydantic compiled: False
                 install path: /home/[user]/git/pydantic/pydantic
               python version: 3.7.5 (default, Nov 20 2019, 09:21:52)  [GCC 9.2.1 20191008]
                     platform: Linux-5.3.0-29-generic-x86_64-with-Ubuntu-19.10-eoan
     optional deps. installed: []
```

Recursive models create a copy of the object during validation, this also happens with `Outer(i=Inner())`.  
This stops the `is` operator from working. I would also assume that it impacts performance slightly more than directly using the object. 

```py

class Inner(BaseModel):
    pass

class Outer(BaseModel):
    i: Inner

    # for prints only
    @validator('i', pre=True)
    def print_id_pre(cls, value):
        print("pre_val: ", hex(id(value)))
        return value

    @validator('i')
    def print_id(cls, value):
        print("normal_val: ", hex(id(value)))
        return value

i = Inner()
print("i: ", hex(id(i)))
o = Outer(i=i)
print("o.i: ", hex(id(o.i)))
print("i is o.i: ", i is o.i)
```
Output:
```
i:  0x7fe11da0ab40
pre_val:  0x7fe11da0ab40
normal_val:  0x7fe11da0abe0
o.i:  0x7fe11da0abe0
i is o.i:  False
```

### LINKED ISSUE #1837: Nested basemodel contain different instances of child models in a collection than origially provided.
# Question
I wanted to use an ID based hashing for `BaseModel` instances. It appears to be the case that nested base models create  new objects in a hashable collection like `set`  or a mutable collection like `List`. 

```
           pydantic version: 1.6.1
           pydantic compiled: True
           install path: /home/umesh/personal/pydantic/pydantic
           python version: 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 02:25:08)  [GCC 7.5.0]
           platform: Linux-5.0.0-32-generic-x86_64-with-debian-buster-sid
```
 In this snippet (grabbed from FastAPI examples). Class  `Item` has a list of `Images`, which adhere to equality based on `id `. It appears to be the case that instance variable `images`  of `Item`  contain images are different from the ones originally provided from the constructor. Is there a way to skirt that? 

```py
from typing import List, Set, Optional


class Image(BaseModel):
    url: str
    name: str

    def __hash__(self):
        return id(self)

    def __eq__(self, other):
        return self is other


class Item(BaseModel):
    name: str
    description: Optional[str] = None
    price: float
    tax: Optional[float] = None
    tags: Set[str] = []
    images: Optional[List[Image]] = None


image_1 = Image(url='https://s3.amazon-aws.com/dakl/kals', name='profile_pic_1.png')
image_2 = Image(url='https://s3.amazon-aws.com/dakl/kals', name='profile_pic_1.png')

item = Item(name='my_item1', price=0.050, images={image_1, image_2})
print(image_1 in item.images)

print(id(image_1), id(item.images[0]), id(image_2), id(item.images[1]))
```
Output:
```
139644562343408 139644562344208 139644562343328 139644562344368
```

PrettyWood:
## Change Summary
Right now we always copy a `BaseModel` when validated, which should not be the default behaviour IMHO.

**While working on this I thought about "feature flags" to be able to change the behaviour of all _pydantic_ models but probably other classes / metaclasses** Don't know if we want to put everything in the `BaseConfig` or if we want to separate what's global and local to a `BaseModel`.

ℹ️ I opened a discussion #2194 and added some documentation if we stick to the current way in #2211
Depending on what's decided, I'll update this PR and add some documentation

## Related issue number
closes #265
closes #1246
closes #1837

## Checklist


pydantic/main.py
```diff
@@ -128,6 +128,9 @@ class BaseConfig:
     json_encoders: Dict[Type[Any], AnyCallable] = {}
     underscore_attrs_are_private: bool = False
 
+    # Whether or not inherited models as fields should be reconstructed as base model
+    copy_on_model_validation: bool = False
```
samuelcolvin, line 132:
```suggestion
    copy_on_model_validation: bool = True
```
Surely we have to stick to `True` in v1.

samuelcolvin (commented):
seems fine, but we need docs too.

changes/265-PrettyWood.md
```diff
@@ -0,0 +1,3 @@
+**Possible Breaking Change:** Inherited models as fields are not reconstructed (copied) anymore on validation
+To keep old behaviour, set `copy_on_model_validation = True` in the `Config` of your `BaseModel`.
+To change it globally for all your _pydantic_ models, you can set `BaseConfig.copy_on_model_validation = True`
```
samuelcolvin, line 3: probably best to move some of this to the docs.

PrettyWood: Added the section from the other PR to explain how to change the behaviour globally.
Added the feature flag in the doc and change the default value to avoid any breaking change.
Ready for another review :)

docs/examples/model_config_change_globally_baseconfig.py
```diff
@@ -0,0 +1,11 @@
+from pydantic import BaseConfig, BaseModel
+
+BaseConfig.arbitrary_types_allowed = True
```
samuelcolvin, line 3: we don't document anywhere doing and I'm still not clear on all its implications.

For now, let's stick to setting the `Config` on models. If we think this is a good idea, it should have an issue to discuss it and its own section in the docs.

samuelcolvin: i think for now, best to remove this example.

samuelcolvin (commented):
otherwise LGTM.

docs/examples/model_config_change_globally_baseconfig.py
PrettyWood: I removed it. An issue has already been opened here https://github.com/samuelcolvin/pydantic/issues/2194.

samuelcolvin:
Hi @PrettyWood unfortunately, merging this has caused tests to fail: https://github.com/samuelcolvin/pydantic/runs/1887580866

Any chance you could look into it?

When I get a chance, I'll move these binary builds to be run on PR tests as well as master to avoid problems like this. There's no way you could have predicted this, in fact I have no idea why those tests are any different to earlier tests with pydantic compiled.

PrettyWood:
Damn sorry...on it!

PR #2196, opened 2020-12-12
introduce allow_mutation Field constraint

### LINKED ISSUE #2195: Introduce a allow_mutation Field constraint
### Checks

# Feature Request
I found mentions of a readOnly parameter on the deprecated Schema class but it was not carried over to Field.  Others are using validators to enforce readonly on assignment, but it's a common enough feature that pydantic could provide it natively.  I am not able to use validators to create a readonly Field because I have validators that need to be run on initialization.

```
             pydantic version: 1.7.2
            pydantic compiled: False
                 install path: /Users/scootna/projects/pydantic/pydantic
               python version: 3.8.2 (default, Sep 24 2020, 19:37:08)  [Clang 12.0.0 (clang-1200.0.32.21)]
```

```py

class Entry(BaseModel):
    id: float = Field(allow_mutation=False)
    val: float

    class Config:
        validate_assignment = True

r = Entry(id=1, val=100)
r.id = 2 # raise error when assigning to read_only field

...
```

sblack-usu:
## Change Summary
Introduces an allow_mutation constraint on Field.  When set to False, a TypeError is raised if the field is assigned.  I also added a check for unused constraints when validation_assignment is set to True. 

## Related issue number
fix #2195

## Checklist


StephenBrown2:
Looks nifty! Should this also alter the generated JSON schema to add a `readOnly` property?

sblack-usu:
Great idea @StephenBrown2.  That didn't even occur to me, just made a new commit to include it.

sblack-usu:
It's unclear to me what is going on with the py3.6/3.7 linux tests.  `ModuleNotFoundError: No module named 'typing_extensions'`

Could I get some help resolving this?  I don't see how my most recent commit could be related to that error.

StephenBrown2:
Not sure myself, since it appears to have been installed successfully, as noted in the steps
- https://github.com/samuelcolvin/pydantic/pull/2196/checks?check_run_id=1553385919#step:4:40
- https://github.com/samuelcolvin/pydantic/pull/2196/checks?check_run_id=1553385919#step:4:128

sblack-usu:
Is it possible it's a blip with the test run?  Can we re-run?  I'd like to move this forward, I just don't know what to do to resolve those tests as they seem unrelated.

PrettyWood:
Hi @sblack-usu I can have a look tonight!

PrettyWood:
Ok I restarted your CI just in case and also did a clean PR just for the deps https://github.com/samuelcolvin/pydantic/pull/2201
I think the issue comes from the update on Github side to Ubuntu 20.04. I'll try to dig into it tomorrow

PrettyWood:
Ok I was wrong sorry about that
The issue comes from the new release v3.2.0 (there is a new v3.3.0 but doesn't matter) that introduced a breaking change for us https://github.com/python/importlib_metadata/blob/v3.2.0/importlib_metadata/_compat.py#L7
And pytest didn't pin the version hence the failure in the CI.
For the meantime I did [this](https://github.com/samuelcolvin/pydantic/pull/2201/commits/1711cc964f759209f170cdb1dadde029f78999ee) to see if it works as there may be other issues.

PrettyWood:
@sblack-usu @StephenBrown2 I made a dedicated PR to keep track of the issue and not hide it into a deps bump.
Feedback more than welcome on https://github.com/samuelcolvin/pydantic/pull/2202 :)
If you're ok with this I think we can merge it to allow people to rebase and for the CI to be green again

sblack-usu:
woohoo! Thanks for fixing that dependency @PrettyWood.

pydantic/schema.py
```diff
@@ -277,6 +277,8 @@ def get_field_schema_validations(field: ModelField) -> Dict[str, Any]:
                 f_schema[keyword] = attr
     if field.field_info is not None and field.field_info.const:
         f_schema['const'] = field.default
+    if field.field_info is not None and field.field_info.read_only:
```
samuelcolvin, line 280: This is incorrect I think, as per JSON Schema this suggests the field cannot be set when creating a model (e.g. like `POST`ing to the endpoint who's `OPTIONS` endpoint returns this schema)

pydantic/schema.py
```diff
@@ -946,6 +952,43 @@ def constraint_func(**kwargs: Any) -> Type[Any]:
 
     ans = go(annotation)
 
+    check_unused_constraints(constraints, used_constraints, field_name)
+
+    return ans
+
+
+def check_unused_validation_assignment_constraints(
```
samuelcolvin, line 960: this function name is too long.

docs/usage/schema.md
```diff
@@ -71,6 +71,8 @@ It has the following arguments:
   JSON Schema
 * `max_length`: for string values, this adds a corresponding validation and an annotation of `maxLength` to the
   JSON Schema
+* `read_only`: a boolean which defaults to False. When True, the field raises a TypeError if the field is
```
samuelcolvin, line 74: The name is misleading, (which I think is why the schema problem has come up). 

The function is not "read only" since the field can be set when creating a model. The name should be `allow_mutation` with default `True`. This is both a better description, and matches `Config.allow_mutation`.

samuelcolvin (commented):
I'm happy with the concept, but the name needs to be changed and unless I'm missing something the schema needs to be changed.

docs/usage/schema.md
sblack-usu: I agree allow_mutation is a better fit and less confusing. Will commit the changes shortly.

pydantic/schema.py
sblack-usu: `check_unused_assignment_constraints` is the best alternative I can come up with at the moment.  Please suggest another name if you do not like this one.

pydantic/schema.py
sblack-usu: Admittedly, I didn't put much thought into the JSON schema as it's not something I currently plan to use.  I see that you are correct here and will revert the commit that adds readOnly to the json schema.

docs/usage/schema.md
sblack-usu: @samuelcolvin - how do you feel about the TypeError being raised?  I'm open to other suggestions if you have one.

sblack-usu:
@samuelcolvin - I've addressed all of your feedback and feel it's ready for another pass.

changes/2195-sblack-usu.md
```diff
@@ -0,0 +1 @@
+add allow_mutation constraint to Field
```
samuelcolvin, line 1:
```suggestion
add `allow_mutation` constraint to `Field`
```

pydantic/fields.py
```diff
@@ -136,6 +138,7 @@ def Field(
     max_items: int = None,
     min_length: int = None,
     max_length: int = None,
+    allow_mutation: bool = None,
```
samuelcolvin, line 141:
```suggestion
    allow_mutation: bool = True,
```

It's good to set the correct default here to make the code self-documenting - e.g. many people use the signature hint provided by pycharm or vscode to understand arguments.

samuelcolvin: it'll also simplify the signature below.

pydantic/schema.py
```diff
@@ -946,6 +950,41 @@ def constraint_func(**kwargs: Any) -> Type[Any]:
 
     ans = go(annotation)
 
+    check_unused_constraints(constraints, used_constraints, field_name)
+
+    return ans
+
+
+def check_unused_assignment_constraints(field_info: FieldInfo, field_name: str, validate_assignment: bool) -> None:
```
samuelcolvin, line 958: Maybe I'm missing something, but I think this is no longer required?

tests/test_schema.py
```diff
@@ -1480,6 +1481,23 @@ class Foo(BaseModel):
...
 
 
+@pytest.mark.parametrize(
+    'kwargs,type_,value',
+    [
+        ({'allow_mutation': True}, str, 'foo'),
+        ({'allow_mutation': False}, str, 'foo'),
+    ],
+)
+def test_schema_constraints_on_assignment(kwargs, type_, value):
```
samuelcolvin, line 1491: I think this test can be removed.

samuelcolvin (commented):
looking good, just a few more things.

pydantic/fields.py
sblack-usu: 147 tests fail with that change.  Looks like it's a ValueError with message `the following field constraints are set but not enforced: allow_mutation.`

I'd probably have to add an exception for checking unenforced constraints for allow_mutation because it's default is a True instead of None.  I'll try to dig into a bit and if there is a straightforward solution I'm comfortable with then I'll make a commit.  Suggestions are welcome here.

pydantic/schema.py
sblack-usu: It's still used here https://github.com/samuelcolvin/pydantic/pull/2196/files#diff-bfaaac2ae0a71d9686a272982d29d0bed82bfaacf8b5b30e2b88ed2543e74a3bR321

It's needed because `allow_mutation` requires a check of `validate_assignment`, none of the other constraints need that information.  If there is a different approach to adding this check which you would prefer, I'm open to ideas.  It might make sense to pull the `check_unused_constraints` call out `get_annotation_from_field_info`.  Having that constraint check hidden in there might be part of the confusion.

sblack-usu: I originally had a check within `get_annotation_from_field_info`, but `make` complained it made the function too complex.  Perhaps the function needs to be broken up instead of the approach I took.

pydantic/schema.py
sblack-usu: removed the function.

pydantic/fields.py
sblack-usu: Just made another commit that allows for this by including a check for the default value of the constraint instead of checking only for None.

pydantic/schema.py
sblack-usu: Took another stab at refactoring the `schema.get_annotation_from_field_info` function.   Basically just pulled out the check for unused constraints to be checked outside of the methods.

pydantic/schema.py
```diff
@@ -849,32 +849,65 @@ def encode_default(dft: Any) -> Any:
...
+    :return: the constraints set on field_info
     """
-    Get an annotation with validation implemented for numbers and strings based on the field_info.
+    return {
+        f[0] for f in _field_constraints if getattr(field_info, f[0]) is not None and getattr(field_info, f[0]) != f[1]
+    }
+
 
+def get_annotation_from_field_info(
+    annotation: Any, field_info: FieldInfo, field_name: str, validate_assignment: bool = False
```
sblack-usu, line 879: I had to add a default value for validate_assignment for the fastAPI check to pass as fastApi uses this function.  Should fastAPI be updated to include this parameter?  I haven't looked at the fastAPI codebase yet.

sblack-usu:
@samuelcolvin - I did my best to address your feedback.  I had to break up `schema.get_annotation_from_field_info` a bit to get it to work.  I feel like it's a cleaner approach than my earlier attempt.

pydantic/schema.py
```diff
@@ -849,32 +849,65 @@ def encode_default(dft: Any) -> Any:
...
-    'max_length',
-    'regex',
-    'gt',
-    'lt',
-    'ge',
-    'le',
-    'multiple_of',
-    'min_items',
-    'max_items',
+_field_constraints = {  # field constraints with the default value
```
PrettyWood, line 852: I understand why you added the default value considering Samuel's remark. But I feel like we put some logic at the wrong place. Couldn't we set this somewhere in `fields.py` and use it everywhere?
Maybe it can be done in a separate PR as we already used to have the field names here

pydantic/schema.py
```diff
@@ -849,32 +849,65 @@ def encode_default(dft: Any) -> Any:
...
-def get_annotation_from_field_info(annotation: Any, field_info: FieldInfo, field_name: str) -> Type[Any]:  # noqa: C901
+def get_constraints(field_info: FieldInfo) -> Set[str]:
+    """Gets the constraints set on the field by comparing the constraint value with it's default value
+
+    :param field_info: an instance of FieldInfo
+    :return: the constraints set on field_info
     """
-    Get an annotation with validation implemented for numbers and strings based on the field_info.
+    return {
+        f[0] for f in _field_constraints if getattr(field_info, f[0]) is not None and getattr(field_info, f[0]) != f[1]
```
PrettyWood, line 874: IMO it's easier to read like this
```python
{ attr for attr, default_value in _field_constraints if getattr(field_info, attr, _missing) != default_value }
```

tests/test_schema.py
```diff
@@ -211,8 +211,8 @@ class Model(BaseModel):
 
 def test_schema_repr():
     s = Field(4, title='Foo is Great')
-    assert str(s) == "default=4 title='Foo is Great' extra={}"
-    assert repr(s) == "FieldInfo(default=4, title='Foo is Great', extra={})"
+    assert str(s) == "default=4 title='Foo is Great' allow_mutation=True extra={}"
```
PrettyWood, line 214: I guess we should also overwrite `__repr_args__` as we probably don't want this to be set by default

pydantic/schema.py
sblack-usu: agreed, done.

tests/test_schema.py
sblack-usu: To do so, we'll have to have access to the defaults within the `Representation` class.  This ties in with your previous comment to set the defaults once and make them available everywhere.  I don't think I'm familiar enough with the code to make this decision, but I'm happy to make the changes in this PR if I'm given guidance.  I think updating `__slots__` to include the defaults could possibly work... but several classes extend `Representation`, so not sure if that's the correct approach.

sblack-usu: After thinking about it some more, I found an approach I liked and went ahead and implemented it.  I simply pushed the get_constraints function from `schema.py` to a method on `FieldInfo`.  The rest came naturally after that.

pydantic/schema.py
sblack-usu: https://github.com/samuelcolvin/pydantic/pull/2196#discussion_r550939024

PrettyWood (commented):
Nice feature! Do we want to also add `readOnly` on generated JSON schema?

pydantic/schema.py
```diff
@@ -946,6 +955,18 @@ def constraint_func(**kwargs: Any) -> Type[Any]:
...
+    return ans, used_constraints
+
+
+def check_unused_constraints(constraints: Set[Any], used_constraints: Set[Any], field_name: str) -> None:
+    """
+    Compares constraints to used_constraints and raises a ValueError if there are unused constraints
+
+    :param constraints: a set of constraints on the field
+    :param used_constraints: a set of used_constraints on the field
+    :param field_name: name of the field for use in error messages
```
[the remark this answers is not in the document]
sblack-usu: It was reused in an earlier iteration of this PR.  I agree and will put it back.

sblack-usu:
> Nice feature! Do we want to also add `readOnly` on generated JSON schema?

I added `readOnly` and then reverted it.  see https://github.com/samuelcolvin/pydantic/pull/2196#discussion_r550794426

docs/usage/schema.md
```diff
@@ -71,6 +71,8 @@ It has the following arguments:
   JSON Schema
 * `max_length`: for string values, this adds a corresponding validation and an annotation of `maxLength` to the
   JSON Schema
+* `allow_mutation`: a boolean which defaults to True. When False, the field raises a TypeError if the field is 
+  assigned on an instance.  The BaseModel Config must set validate_assignment to True
```
samuelcolvin, line 75:
```suggestion
* `allow_mutation`: a boolean which defaults to `True`. When False, the field raises a `TypeError` if the field is 
  assigned on an instance.  The model config must set `validate_assignment` to `True` for this check to be performed.
```

pydantic/fields.py
```diff
@@ -115,9 +130,21 @@ def __init__(self, default: Any = Undefined, **kwargs: Any) -> None:
...
+        self.allow_mutation = kwargs.pop('allow_mutation', True)
         self.regex = kwargs.pop('regex', None)
         self.extra = kwargs
 
+    def __repr_args__(self) -> 'ReprArgs':
+        attrs = ((s, getattr(self, s)) for s in self.__slots__)
+        return [(a, v) for a, v in attrs if v != self.__field_constraints__.get(a, None)]
+
+    def get_constraints(self) -> Set[str]:
+        """Gets the constraints set on the field by comparing the constraint value with its default value
```
samuelcolvin, line 142:
```suggestion
        """
        Gets the constraints set on the field by comparing the constraint value with its default value
```

pydantic/fields.py
```diff
@@ -115,9 +130,21 @@ def __init__(self, default: Any = Undefined, **kwargs: Any) -> None:
         self.max_items = kwargs.pop('max_items', None)
         self.min_length = kwargs.pop('min_length', None)
         self.max_length = kwargs.pop('max_length', None)
+        self.allow_mutation = kwargs.pop('allow_mutation', True)
         self.regex = kwargs.pop('regex', None)
         self.extra = kwargs
 
+    def __repr_args__(self) -> 'ReprArgs':
+        attrs = ((s, getattr(self, s)) for s in self.__slots__)
+        return [(a, v) for a, v in attrs if v != self.__field_constraints__.get(a, None)]
```
samuelcolvin, line 139: 👍 smart.

pydantic/main.py
```diff
@@ -388,6 +388,8 @@ def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
                 # - make sure validators are called without the current value for this field inside `values`
                 # - keep other values (e.g. submodels) untouched (using `BaseModel.dict()` will change them into dicts)
                 # - keep the order of the fields
+                if not known_field.field_info.allow_mutation:
+                    raise TypeError(f'"{known_field.name}" has allow_mutation set to False and cannot be assigned"')
```
samuelcolvin, line 392:
```suggestion
                    raise TypeError(f'"{known_field.name}" has allow_mutation set to False and cannot be assigned')
```

pydantic/schema.py
```diff
@@ -849,32 +849,47 @@ def encode_default(dft: Any) -> Any:
...
-    'max_items',
-}
-
-
-def get_annotation_from_field_info(annotation: Any, field_info: FieldInfo, field_name: str) -> Type[Any]:  # noqa: C901
+
+
+def get_annotation_from_field_info(
+    annotation: Any, field_info: FieldInfo, field_name: str, validate_assignment: bool = False
+) -> Type[Any]:  # noqa: C901
```
samuelcolvin, line 856:
```suggestion
) -> Type[Any]:
```
i doubt this is now required.

pydantic/schema.py
```diff
@@ -849,32 +849,47 @@ def encode_default(dft: Any) -> Any:
...
+        raise ValueError(
+            f'On field "{field_name}" the following field constraints are set but not enforced: '
+            f'{", ".join(unused_constraints)}. '
+            f'\nFor more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints'
+        )
+
+    return annotation
+
+
+def get_annotation_with_constraints(annotation: Any, field_info: FieldInfo) -> Tuple[Type[Any], Set[str]]:  # noqa: C901
```
samuelcolvin, line 885: is `# noqa: C901` required here?

samuelcolvin (commented):
otherwise this looks great.

pydantic/schema.py
sblack-usu: When I remove it, flake complains `pydantic/schema.py:885:1: C901 'get_annotation_with_constraints' is too complex (19)`

pydantic/schema.py
sblack-usu: You are correct.

pydantic/schema.py
samuelcolvin: ok, it's required. Just wasn't sure.

sblack-usu:
> otherwise this looks great.

Each suggestion was committed and I resolved the latest merge conflicts.  If everything looks good, I'd love to see this merged.

samuelcolvin:
awesome, thank you.

PR #2209, opened 2020-12-18
Helpful TypeError when user forgets to return `values` in root valida…

masalim2:
## Change Summary

If a user forgot to return the `values` dict in a root validator, they'd see a somewhat opaque exception: 
```
TypeError: __dict__ must be set to a dictionary, not a 'NoneType'`.  
```

Added a quick check in `BaseModel.__init__` to raise a more user-friendly hint in the TypeError:
```
TypeError: BaseModel.__init__ received `None` values. Are you forgetting to return `values` in a root validator?
```

## Checklist

### DIFF
--- a/changes/2209-masalim2.md
+++ b/changes/2209-masalim2.md
@@ -1,1 +1,1 @@
-User-friendly TypeError when a `root_validator` returns `None`.
+Raise a user-friendly TypeError when a root_validator does not return a dict (e.g. None)
--- a/changes/2209-masalim2.md
+++ b/changes/2209-masalim2.md
@@ -1,0 +1,1 @@
+User-friendly TypeError when a `root_validator` returns `None`.
--- a/changes/2209-masalim2.md
+++ b/changes/2209-masalim2.md
@@ -1,1 +1,1 @@
-Raise a user-friendly TypeError when a root_validator does not return a dict (e.g. None)
+Raise a user-friendly `TypeError` when a `root_validator` does not return a `dict` (e.g. `None`)
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -360,11 +360,15 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
         if validation_error:
             raise validation_error
+        if values is None:
+            raise TypeError(
+                'BaseModel.__init__ received `None` values. Are you forgetting to return `values` in a root validator?'
+            )
         object_setattr(__pydantic_self__, '__dict__', values)
         object_setattr(__pydantic_self__, '__fields_set__', fields_set)
         __pydantic_self__._init_private_attributes()
 
-    @no_type_check
+    @no_type_check  # noqa: C901 (ignore complexity)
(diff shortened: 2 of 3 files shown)


pydantic/main.py
```diff
@@ -360,11 +360,15 @@ def __init__(__pydantic_self__, **data: Any) -> None:
         values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
         if validation_error:
             raise validation_error
+        if values is None:
```
PrettyWood, line 363: Good idea! 👍 
But what if you do `return 3` in your `root_validator` you will still have an implicit error `__dict__ must be set to a dictionary, not a 'int'`! Maybe we could handle a more generic case with something like
```python
try:
    object_setattr(__pydantic_self__, '__dict__', values)
except TypeError as e:
    raise TypeError('Did you return a valid dict in your root validators ?') from e
```
?
Just my 2 cents :)

masalim2: Thanks for the feedback!  This looks good -- updated the PR 👍

pydantic/main.py
```diff
@@ -360,11 +360,14 @@ def __init__(__pydantic_self__, **data: Any) -> None:
...
-        object_setattr(__pydantic_self__, '__dict__', values)
+        try:
+            object_setattr(__pydantic_self__, '__dict__', values)
+        except TypeError as e:
+            raise TypeError('Did you return a valid dict in your root validators?') from e
         object_setattr(__pydantic_self__, '__fields_set__', fields_set)
         __pydantic_self__._init_private_attributes()
 
-    @no_type_check
+    @no_type_check  # noqa: C901 (ignore complexity)
```
samuelcolvin, line 370: this looks unrelated.

pydantic/main.py
```diff
@@ -360,11 +360,14 @@ def __init__(__pydantic_self__, **data: Any) -> None:
         values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
         if validation_error:
             raise validation_error
-        object_setattr(__pydantic_self__, '__dict__', values)
+        try:
+            object_setattr(__pydantic_self__, '__dict__', values)
+        except TypeError as e:
+            raise TypeError('Did you return a valid dict in your root validators?') from e
```
samuelcolvin, line 366: the exception message should not be a question, maybe something like

> model values must be a dict, you may not have returned a dictionary from a root validator

samuelcolvin (commented):
otherwise LGTM.

pydantic/main.py
PrettyWood: It's because of https://github.com/samuelcolvin/pydantic/pull/2198#pullrequestreview-550921099

PrettyWood (commented):
@masalim2 Could you change the exception message?

changes/2209-masalim2.md
```diff
@@ -0,0 +1 @@
+User-friendly TypeError when a `root_validator` returns `None`.
```
PrettyWood, line 1: Better add a verb at the beginning like
> Raise a user-friendly `TypeError` when a `root_validator` does not return a dict (e.g. `None`).

Otherwise LGTM 👍

PrettyWood (approved):

PR #2214, opened 2020-12-22
fix: do not ignore annotated fields when type is `Union[Type[...], ...]`

### LINKED ISSUE #2213: Pydantic "Not Recognizing" Field in BaseModel with Default Value of Type
### Checks

* [X ] I added a descriptive title to this issue
* [X ] I have searched (google, github) for similar issues and couldn't find anything
* [X ] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

```
pydantic version: 1.7.3
pydantic compiled: True
platform: Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.29
```

```py
import pydantic
from typing import Type, Sequence, Union

class T:
    pass

    t: Union[Type[T], Sequence[Type[T]]] = T
    
    @pydantic.validator("t", always=True)
    def make_tuple(cls, v):
        if isinstance(v, Sequence):
            return tuple(v)
        return v,

# ConfigError: Validators defined with incorrect fields: make_tuple (use check_fields=False if you're inheriting from the model and intended this)
```

**_Note:_** If the field is marked optional and the default value is set to `None`, the code works without issue.

```py
import pydantic
from typing import Type, Sequence, Union, Optional

class T:
    pass

    t: Optional[Union[Type[T], Sequence[Type[T]]]] = None
    
    @pydantic.validator("t", always=True)
    def make_tuple(cls, v):
        if not v:
            return T,
        elif isinstance(v, Sequence):
            return tuple(v)
        return v,

Model()
# Model(t=(<class '__main__.T'>,))
```

Perhaps I am missing something obvious in my sleep-deprived stupor?

PrettyWood:
## Change Summary
Considering a model
```python
class Model(BaseModel):
    ok_field: Type[FooBar] = FooBar
    ignored_field: Union[Type[FooBar], Sequence[Type[FooBar]]] = FooBar
```
The `ignored_field` was not taken into account because in the metaclass we were only handling `Type[...]`.
When the type is `Union[...]` the origin is `Union` and we need to check the valid types inside the `Union`

## Related issue number
closes #2213

## Checklist

### DIFF
--- a/changes/2213-PrettyWood.md
+++ b/changes/2213-PrettyWood.md
@@ -1,0 +1,1 @@
+Do not ignore annotated fields when type is `Union[Type[...], ...]`
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -34,7 +34,15 @@ from .json import custom_pydantic_encoder, pydantic_encoder
 from .parse import Protocol, load_file, load_str_bytes
 from .schema import default_ref_template, model_schema
 from .types import PyObject, StrBytes
-from .typing import AnyCallable, ForwardRef, get_origin, is_classvar, resolve_annotations, update_field_forward_refs
+from .typing import (
+    AnyCallable,
+    ForwardRef,
+    get_args,
+    get_origin,
+    is_classvar,
+    resolve_annotations,
+    update_field_forward_refs,
+)
 from .utils import (
     ROOT_KEY,
     ClassAttribute,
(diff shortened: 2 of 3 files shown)


mariusvniekerk (approved):
Nice catch

samuelcolvin:
nice.

PR #2216, opened 2020-12-22
feat: add support for `NamedTuple` and `TypedDict` types

### LINKED ISSUE #760: TypedDict support
# Feature Request
It would be great if pydantic supported TypedDict.

* OS: MacOS
* Python version `import sys; print(sys.version)`: **3.7.4**
* Pydantic version `import pydantic; print(pydantic.VERSION)`: **0.32.1**

```py
from mypy_extensions import TypedDict


class Data(TypedDict):
    a: int


class User(BaseModel):
    data: Data

if __name__ == '__main__':
    external_data = {
        'data': {
            'a': 'invalid',
        }
    }
    # should raise exception
    user = User(**external_data)
```

### LINKED ISSUE #1324: Support NamedTuples
# Feature Request

Python's typing library has a nice type called NamedTuple, which doesn't seem to be supported.

```
from typing import List, NamedTuple

class Event(NamedTuple):
    a: int
    b: int
    c: str


class Message(BaseModel):
    events: List[Event]

msg = Message.parse_raw(b'{"events": [[1,2, "c"]]}')

# will raise exception because it's a regular tuple and not a named one.
msg.events[0].a 
```

### LINKED ISSUE #1430: "TypeError: TypedDict does not support instance and class checks" for TypedDict field
# Bug

Models with TypedDict fields currently result in an error when they are instantiated, due to the fact that TypedDict does not support isinstance/issubclass checks. 

```py
from typing import TypedDict

TD = TypedDict('TD', {'a': int})


class ExampleModel(BaseModel):
    td: TD


m = ExampleModel(td={'a': 1})

# output:
#   ...
#   File "pydantic/main.py", line 246, in pydantic.main.ModelMetaclass.__new__
#   File "pydantic/fields.py", line 310, in pydantic.fields.ModelField.infer
#   File "pydantic/fields.py", line 272, in pydantic.fields.ModelField.__init__
#   File "pydantic/fields.py", line 364, in pydantic.fields.ModelField.prepare
#   File "pydantic/fields.py", line 405, in pydantic.fields.ModelField._type_analysis
#   File "/usr/lib/python3.8/typing.py", line 1728, in _check_fails
#     raise TypeError('TypedDict does not support instance and class checks')
# TypeError: TypedDict does not support instance and class checks
```

Previously, models could have TypedDict fields without any problems. This error seems to have been introduced in commit #1266.

```
pydantic version: 1.5.1
            pydantic compiled: False
                 install path: /home/kpberry/Downloads/pydantic/pydantic
                     platform: Linux-5.4.0-26-generic-x86_64-with-glibc2.29
```

PrettyWood:
## Change Summary
Screen Shot 2020-12-23 at 1 58 37 AM

Also expose two utils
- `create_model_from_namedtuple`
- `create_model_from_typeddict`

Screen Shot 2021-01-17 at 11 57 29 PM

## Related issue number

fix #760
fix #1324
fix #1430

## Checklist

### DIFF
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -88,11 +88,12 @@ with custom properties and validation.
 `typing.Tuple`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
 
-`subclass of typing.NamedTuple (or collections.namedtuple)`
-: Same as `tuple` but instantiates with the given namedtuple.  
- _pydantic_ will validate the tuple if you use `typing.NamedTuple` since fields are annotated.  
-  If you use `collections.namedtuple`, no validation will be done.  
+`subclass of typing.NamedTuple`
+: Same as `tuple` but instantiates with the given namedtuple and validates fields since they are annotated.
   See [Annotated Types](#annotated-types) below for more detail on parsing and validation
+
+`subclass of collections.namedtuple`
+: Same as `subclass of typing.NamedTuple` but all fields will have type `Any` since they are not annotated
 
 `typing.Dict`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -420,7 +420,7 @@ _(This script is complete, it should run "as is")_
     This is a new feature of the python standard library as of python 3.8.
     Prior to python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
     But required and optional fields are properly differentiated only since python 3.9.
-    It is hence recommanded using [typing-extensions](https://pypi.org/project/typing-extensions/) with python 3.8 as well.
+    We therefore recommend using [typing-extensions](https://pypi.org/project/typing-extensions/) with python 3.8 as well.
 
 
 ```py
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -85,8 +85,16 @@ with custom properties and validation.
 `typing.Tuple`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
 
+`subclass of typing.NamedTuple (or collections.namedtuple)`
+: Same as `tuple` but instantiates with the given namedtuple.  
+ _pydantic_ will validate the tuple if you use `typing.NamedTuple` since fields are annotated.  
+  If you use `collections.namedtuple`, no validation will be done.
+
 `typing.Dict`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
+
+`subclass of typing.TypedDict`
+: Same as `dict` but _pydantic_ will validate the dictionary since keys are annotated
 
 `typing.Set`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -408,12 +408,21 @@ _(This script is complete, it should run "as is")_
 ## Annotated Types
 
 ### NamedTuple
+
 ```py
 {!.tmp_examples/annotated_types_named_tuple.py!}
 ```
 _(This script is complete, it should run "as is")_
 
 ### TypedDict
+
+!!! note
+    This is a new feature of the python standard library as of python 3.8.
+    Prior to python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
+    But required and optional fields are properly differentiated only since python 3.9.
+    It is hence recommanded using [typing-extensions](https://pypi.org/project/typing-extensions/) with python 3.8 as well.
+
+
 ```py
 {!.tmp_examples/annotated_types_typed_dict.py!}
 ```
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -91,13 +91,15 @@ with custom properties and validation.
 `subclass of typing.NamedTuple (or collections.namedtuple)`
 : Same as `tuple` but instantiates with the given namedtuple.  
  _pydantic_ will validate the tuple if you use `typing.NamedTuple` since fields are annotated.  
-  If you use `collections.namedtuple`, no validation will be done.
+  If you use `collections.namedtuple`, no validation will be done.  
+  See [Annotated Types](#annotated-types) below for more detail on parsing and validation
 
 `typing.Dict`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
 
 `subclass of typing.TypedDict`
-: Same as `dict` but _pydantic_ will validate the dictionary since keys are annotated
+: Same as `dict` but _pydantic_ will validate the dictionary since keys are annotated.  
+  See [Annotated Types](#annotated-types) below for more detail on parsing and validation
 
 `typing.Set`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
@@ -400,6 +402,20 @@ With proper ordering in an annotated `Union`, you can use this to parse types of
 
 ```py
 {!.tmp_examples/types_literal3.py!}
+```
+_(This script is complete, it should run "as is")_
+
+## Annotated Types
+
+### NamedTuple
+```py
+{!.tmp_examples/annotated_types_named_tuple.py!}
+```
+_(This script is complete, it should run "as is")_
+
+### TypedDict
+```py
+{!.tmp_examples/annotated_types_typed_dict.py!}
 ```
 _(This script is complete, it should run "as is")_
 
--- a/pydantic/__init__.py
+++ b/pydantic/__init__.py
@@ -1,6 +1,6 @@
 # flake8: noqa
 from . import dataclasses
-from .annotated_types import namedtuple_to_model, typeddict_to_model
+from .annotated_types import create_model_from_namedtuple, create_model_from_typeddict
 from .class_validators import root_validator, validator
 from .decorator import validate_arguments
 from .env_settings import BaseSettings
@@ -18,8 +18,8 @@ from .version import VERSION
 # please use "from pydantic.errors import ..." instead
 __all__ = [
     # annotated types utils
-    'namedtuple_to_model',
-    'typeddict_to_model',
+    'create_model_from_namedtuple',
+    'create_model_from_typeddict',
     # dataclasses
     'dataclasses',
     # class_validators
--- a/pydantic/__init__.py
+++ b/pydantic/__init__.py
@@ -1,5 +1,6 @@
 # flake8: noqa
 from . import dataclasses
+from .annotated_types import namedtuple_to_model, typeddict_to_model
 from .class_validators import root_validator, validator
 from .decorator import validate_arguments
 from .env_settings import BaseSettings
@@ -16,6 +17,9 @@ from .version import VERSION
 # WARNING __all__ from .errors is not included here, it will be removed as an export here in v2
 # please use "from pydantic.errors import ..." instead
 __all__ = [
+    # annotated types utils
+    'namedtuple_to_model',
+    'typeddict_to_model',
     # dataclasses
     'dataclasses',
     # class_validators
--- a/pydantic/annotated_types.py
+++ b/pydantic/annotated_types.py
@@ -50,10 +50,10 @@ def create_model_from_namedtuple(namedtuple_cls: Type['NamedTuple'], **kwargs: A
     but also with `collections.namedtuple` without any, in which case we consider the type
     of all the fields to be `Any`
     """
-    named_tuple_annotations: Dict[str, Type[Any]] = getattr(
+    namedtuple_annotations: Dict[str, Type[Any]] = getattr(
         namedtuple_cls, '__annotations__', {k: Any for k in namedtuple_cls._fields}
     )
     field_definitions: Dict[str, Any] = {
-        field_name: (field_type, Required) for field_name, field_type in named_tuple_annotations.items()
+        field_name: (field_type, Required) for field_name, field_type in namedtuple_annotations.items()
     }
     return create_model(f'{namedtuple_cls.__name__}Model', **kwargs, **field_definitions)
--- a/pydantic/annotated_types.py
+++ b/pydantic/annotated_types.py
@@ -14,9 +14,9 @@ if TYPE_CHECKING:
 
 def create_model_from_typeddict(typeddict_cls: Type['TypedDict'], **kwargs: Any) -> Type['BaseModel']:
     """
-    Convert a `TypedDict` to a `BaseModel`
+    Create a `BaseModel` based on the fields of a `TypedDict`.
     Since `typing.TypedDict` in Python 3.8 does not store runtime information about optional keys,
-    we warn the user if that's the case (see https://bugs.python.org/issue38834)
+    we warn the user if that's the case (see https://bugs.python.org/issue38834).
     """
     field_definitions: Dict[str, Any]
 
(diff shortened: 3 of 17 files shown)


docs/usage/types.md
```diff
@@ -85,9 +85,17 @@ with custom properties and validation.
...
 
+`subclass of typing.NamedTuple (or collections.namedtuple)`
+: Same as `tuple` but instantiates with the given namedtuple.  
+ _pydantic_ will validate the tuple if you use `typing.NamedTuple` since fields are annotated.  
+  If you use `collections.namedtuple`, no validation will be done.
+
 `typing.Dict`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
 
+`subclass of typing.TypedDict`
```
samuelcolvin, line 96:
```suggestion
`typing.TypedDict`
```
I think the subclass is implicit

docs/usage/types.md
```diff
@@ -85,9 +85,17 @@ with custom properties and validation.
 `typing.Tuple`
 : see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
 
+`subclass of typing.NamedTuple (or collections.namedtuple)`
```
samuelcolvin, line 88:
```suggestion
`typing.NamedTuple` (or `collections.namedtuple`)
```
again I think the "subclass of..." is implicit

pydantic/typing.py
```diff
@@ -258,6 +260,18 @@ def all_literal_values(type_: Type[Any]) -> Tuple[Any, ...]:
...
+def is_named_tuple_type(type_: Type[Any]) -> bool:
+    from .utils import lenient_issubclass
+
+    return lenient_issubclass(type_, tuple) and hasattr(type_, '_fields')
+
+
+def is_typed_dict_type(type_: Type[Any]) -> bool:
+    from .utils import lenient_issubclass
+
+    return lenient_issubclass(type_, dict) and getattr(type_, '__annotations__', None)
```
samuelcolvin, line 272: Humm, I wonder if this is a water tight check. I can't think of a better solution though.

pydantic/validators.py
```diff
@@ -523,6 +530,44 @@ def pattern_validator(v: Any) -> Pattern[str]:
         raise errors.PatternError()
 
 
+NamedTupleT = TypeVar('NamedTupleT', bound=NamedTuple)
+
+
+def make_named_tuple_validator(type_: Type[NamedTupleT]) -> Callable[[Tuple[Any, ...]], NamedTupleT]:
+    from .main import create_model
+
+    # A named tuple can be created with `typing,NamedTuple` with types
```
samuelcolvin, line 539:
```suggestion
    # A named tuple can be created with `typing.NamedTuple` with types
```

pydantic/validators.py
```diff
@@ -523,6 +530,44 @@ def pattern_validator(v: Any) -> Pattern[str]:
...
+        return type_(**validated_dict_values)
+
+    return named_tuple_validator
+
+
+def make_typed_dict_validator(type_: Type['TypedDict']) -> Callable[[Any], Dict[str, Any]]:
+    from .main import create_model
+
+    field_definitions: Dict[str, Any] = {
+        field_name: (field_type, ... if type_.__total__ else None)
```
samuelcolvin, line 560: for performance and clarity
```suggestion
    default_value = ... if type_.__total__ else None
    field_definitions: Dict[str, Any] = {
        field_name: (field_type, default_value)
```

tests/test_main.py
```diff
@@ -1425,3 +1434,90 @@ class M(BaseModel):
         a: int
 
     get_type_hints(M.__config__)
+
+
+def test_named_tuple():
```
samuelcolvin, line 1439: I know test files are a mess, but please can you move this to somewhere else. I try to keep `test_main.py` for the most vanilla cases.

samuelcolvin (commented):
This looks awesome, would be great if we could have a proper example in the docs.

pydantic/typing.py
PrettyWood: Currently there is no easier way afaik. There will be an [public method](https://docs.python.org/3.10/library/typing.html#typing.is_typeddict) in 3.10 though

tests/test_main.py
PrettyWood: Sure thing! TBH I didn't know it was for vanilla cases. Maybe a module docstring on top of the module could help. I'll try to find a better place for these bad boys

docs/usage/types.md
PrettyWood: It used to be the case for me too but I now tend to disagree since we added the support of `enum.Enum` to have a valid enum instance. We could someday add `typing.TypedDict` just to have a valid `TypedDict` instance for example. WDYT?

tests/test_main.py
PrettyWood: I created a new dedicated file `test_annotated_types`. Suggestions welcome if you don't like it

PrettyWood:
Ok I added a dedicated section in the doc for both annotated types and took your remarks into account except the `subclass` thing (cf my remark). I feel like it's ready for another review :)

RobertCraigie:
This looks great, I'll be putting it to good use, thanks!

However there is an issue that this does not respect subclassing, for example, the following will raise a `ValidationError` that the `name` field is missing.
```py
class OptionalUser(TypedDict, total=False):
    name: str

class User(OptionalUser):
    id: int

class Model(BaseModel):
    user: User

Model(user={'id': 1})
```

I had a quick look to try and fix this issue but couldn't find a complete solution.

There are `__required_keys__` and `__optional_keys__` attributes but they are not guaranteed to be present on all versions and we cannot calculate the optional / required keys ourselves by traversing the mro as it is discarded upon class creation.

If there is no universal solution for this issue then `__required_keys__` and `__optional_keys__`  should be respected if they are present.

PrettyWood:
Hi @RobertCraigie
First let me thank you for testing the code! It's so valuable to have feedback before a feature is actually merged.
You're right I didn't get that. My first implementation used `__required_keys__` indeed but I got issues with the CI with older python versions that didn't support it. And I don't know what the best:
- having a better support for python 3.8+ but it means the code will behave differently depending on python version, which I don't like. Maybe we could add a warning in this case?
- having a partial support that is the same (current implementation), which is not perfect either
- the answer C
I'll give answer C another go and see if I can find a way to make it work for all versions.

pydantic/validators.py
```diff
@@ -536,6 +545,62 @@ def pattern_validator(v: Any) -> Pattern[str]:
...
+            field_name: (field_type, ... if field_name in type_.__required_keys__ else None)
+            for field_name, field_type in type_.__annotations__.items()
+        }
+    else:
+        import warnings
+
+        warnings.warn(
+            'You should use `typing_extensions.TypedDict` instead of `typing.TypedDict` for better support! '
+            'Without it, there is no way to differentiate required and optional fields. '
+            'All fields will therefore be considered required.',
```
RobertCraigie, line 588: Should note that all fields can be optional as well

RobertCraigie:
Glad I could help!

I would also like to be able to explicitly forbid passing extra keys, is this possible?

PrettyWood:
@RobertCraigie Sure easy peasy ! :) Just needed to pass down the `config` like I did for `dataclass`.
I didn't pass it down to `NamedTuple` though as I feel like it makes no sense. It can probably be done in a future PR if needed.

pydantic/__init__.py
```diff
@@ -16,6 +17,9 @@
 # WARNING __all__ from .errors is not included here, it will be removed as an export here in v2
 # please use "from pydantic.errors import ..." instead
 __all__ = [
+    # annotated types utils
+    'create_model_from_namedtuple',
```
PrettyWood, line 21: @samuelcolvin I didn't do it directly in this PR but I think it's quite nice to have the helpers exposed properly.
If you're ok with this I reckon we'll be able to add a `create_model_from_dataclass` instead of the not-very-pretty `pydantic.dataclasses.dataclass(StdDataclass).__pydantic_model__`

pydantic/annotated_types.py
```diff
@@ -0,0 +1,59 @@
+from typing import TYPE_CHECKING, Any, Dict, FrozenSet, NamedTuple, Type
+
+from .fields import Required
+from .main import BaseModel, create_model
+
+if TYPE_CHECKING:
+
+    class TypedDict(Dict[str, Any]):
```
PrettyWood, line 8: We could use `typing_extensions.TypedDict` for python < 3.9 or `typing.TypedDict` for 3.9+ but we would need to add `typing_extensions` in linting requirements.

requirements.txt
```diff
@@ -4,5 +4,5 @@ Cython==0.29.21;sys_platform!='win32'
 devtools==0.6.1
 email-validator==1.1.2
 dataclasses==0.6; python_version < '3.7'
-typing-extensions==3.7.4.1; python_version < '3.8'
+typing-extensions==3.7.4.3; python_version < '3.9'
```
PrettyWood, line 7: Had to bump this to have the good version of `TypedDict` for python < 3.9

pydantic/schema.py
```diff
@@ -795,7 +795,7 @@ def field_singleton_schema(  # noqa: C901 (ignore complexity)
         f_schema, schema_overrides = get_field_info_schema(field)
         f_schema.update(get_schema_ref(enum_name, ref_prefix, ref_template, schema_overrides))
         definitions[enum_name] = enum_process_schema(field_type)
-    else:
+    elif not hasattr(field_type, '__pydantic_model__'):
```
PrettyWood, line 798: Felt like the easiest way to have a nice schema support was to add `__pydantic_model__` too like we do for `dataclass`

Thorbenl:
Is there any timeframe for when we can expect this :)

docs/usage/types.md
samuelcolvin: I would still prefer to remove `subclass of` and instead be explicity when we mean the actual type, like `Enum`.

docs/usage/types.md
```diff
@@ -88,9 +88,19 @@ with custom properties and validation.
...
+: Same as `tuple` but instantiates with the given namedtuple.  
+ _pydantic_ will validate the tuple if you use `typing.NamedTuple` since fields are annotated.  
+  If you use `collections.namedtuple`, no validation will be done.
```
samuelcolvin, line 94: I think "no validation will be done" is misleading. I assumed it meant not validation at all will be done, e.g. the field would be equivalent to `Any`.

I think better to make `collections.namedtuple` a different item in this list and instead say "all fields will have type `Any`" or something.

docs/usage/types.md
```diff
@@ -395,6 +405,29 @@ With proper ordering in an annotated `Union`, you can use this to parse types of
...
+```
+_(This script is complete, it should run "as is")_
+
+### TypedDict
+
+!!! note
+    This is a new feature of the python standard library as of python 3.8.
+    Prior to python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
+    But required and optional fields are properly differentiated only since python 3.9.
+    It is hence recommanded using [typing-extensions](https://pypi.org/project/typing-extensions/) with python 3.8 as well.
```
samuelcolvin, line 423:
```suggestion
    We therefore recommend using [typing-extensions](https://pypi.org/project/typing-extensions/) with python 3.8 as well.
```

pydantic/__init__.py
samuelcolvin: yes agree, perhaps we might want `create_model_from` which takes any of the three types, but we definitely want these standalone methods for performance.

Or we could go even further and extend `create_model` to take a single dataclass, named tuple or typed dict instead of fields?

pydantic/annotated_types.py
```diff
@@ -0,0 +1,59 @@
...
+        __annotations__: Dict[str, Type[Any]]
+        __total__: bool
+        __required_keys__: FrozenSet[str]
+        __optional_keys__: FrozenSet[str]
+
+
+def create_model_from_typeddict(typeddict_cls: Type['TypedDict'], **kwargs: Any) -> Type['BaseModel']:
+    """
+    Convert a `TypedDict` to a `BaseModel`
```
samuelcolvin, line 17:
```suggestion
    Create a `BaseModel` based on the fields of a `TypedDict`.
```

The `TypedDict` is not actually "converted".

pydantic/annotated_types.py
```diff
@@ -0,0 +1,59 @@
...
+
+    return create_model(typeddict_cls.__name__, **kwargs, **field_definitions)
+
+
+def create_model_from_namedtuple(namedtuple_cls: Type['NamedTuple'], **kwargs: Any) -> Type['BaseModel']:
+    """
+    Convert a named tuple to a `BaseModel`
+    A named tuple can be created with `typing.NamedTuple` and declared annotations
+    but also with `collections.namedtuple` without any, in which case we consider the type
+    of all the fields to be `Any`
```
samuelcolvin, line 51:
```suggestion
    but also with `collections.namedtuple`, in this case we consider all fields
    to have type `Any`
```

pydantic/validators.py
```diff
@@ -536,6 +540,37 @@ def pattern_validator(v: Any) -> Pattern[str]:
...
+
+
+def make_namedtuple_validator(namedtuple_cls: Type[NamedTupleT]) -> Callable[[Tuple[Any, ...]], NamedTupleT]:
+    from .annotated_types import create_model_from_namedtuple
+
+    NamedTupleModel = create_model_from_namedtuple(namedtuple_cls)
+    namedtuple_cls.__pydantic_model__ = NamedTupleModel  # type: ignore[attr-defined]
+
+    def namedtuple_validator(values: Tuple[Any, ...]) -> NamedTupleT:
+        dict_values: Dict[str, Any] = dict(zip(NamedTupleModel.__annotations__, values))
```
samuelcolvin, line 553:
```py
from typing import NamedTuple

class Point(NamedTuple):
    x: int
    y: int

class Model(BaseModel):
    p: Point

print(Model(p=(1, 2)))
#> p=Point(x=1, y=2)
print(Model(p=(1, 2, 3)))
#> p=Point(x=1, y=2) !!! this should have caused an error
```

```suggestion
        annotations = NamedTupleModel.__annotations__
        if len(annotations) < len(values):
            raise ... some kind of length error
        dict_values: Dict[str, Any] = dict(zip(annotations, values))
```

pydantic/validators.py
```diff
@@ -536,6 +540,37 @@ def pattern_validator(v: Any) -> Pattern[str]:
...
+def make_typeddict_validator(
+    typeddict_cls: Type['TypedDict'], config: Type['BaseConfig']
+) -> Callable[[Any], Dict[str, Any]]:
+    from .annotated_types import create_model_from_typeddict
+
+    TypedDictModel = create_model_from_typeddict(typeddict_cls, __config__=config)
+    typeddict_cls.__pydantic_model__ = TypedDictModel  # type: ignore[attr-defined]
+
+    def typeddict_validator(values: 'TypedDict') -> Dict[str, Any]:
+        return TypedDictModel(**values).dict(exclude_unset=True)
```
samuelcolvin, line 569:
```suggestion
        return TypedDictModel.parse_obj(values).dict(exclude_unset=True)
```
To catch non mapping types.

samuelcolvin (commented):
this is going to be f***ing awesome. Can't wait for it.

Just a few small things.

PrettyWood:
@samuelcolvin Feedback taken except for the rewording on `types.md`.
I'm not really sure how to explain the `Enum` for example.
Would you mind update `types.md` for me please ? I'm sure it will be clearer than what I would come by with

samuelcolvin:
> @samuelcolvin Feedback taken except for the rewording on `types.md`.
> I'm not really sure how to explain the `Enum` for example.
> Would you mind update `types.md` for me please ? I'm sure it will be clearer than what I would come by with

Don't worry about it, the whole file needs rethinking, what's there is fine.

I'll message tomorrow.

samuelcolvin:
🎉 this is great, thank you so much.

PR #2220, opened 2020-12-24
fix: resolve forward refs for inherited dataclasses

### LINKED ISSUE #1668: Inherited dataclasses don't resolve forward refs
# Bug

```
pydantic version: 1.5.1
python version: 3.8.2
```

a.py:
```py
from uuid import UUID

@dataclass
class A:
    uuid: UUID

    # workaround
    # def __post_init__(self):
    #     self.__pydantic_model__.update_forward_refs(**globals())
```

b.py:
```py
from uuid import uuid4
from a import A

@dataclass
class B(A):
    pass

B(uuid=uuid4())
```

`B(uuid=uuid4())` throws `field "uuid" not yet prepared so type is still a ForwardRef, you might need to call B.update_forward_refs()`.

PrettyWood:
## Change Summary
When a stdlib `dataclass` is used inside a `BaseModel`, it is converted into a _pydantic_ `dataclass`. But `__annotations__` passed down when converted were not resolved properly so if we used `'Literal...'` or postponed annotations by default with `from __future__ import annotations`, it failed.
The main problem is that we don't set `__module__ = _cls.__module__` in the _pydantic_ `dataclass` but rather `__module__ = __name__`. This is done to allow pickling. So the `resolve_annotations` done in the metaclass fails.
To support both I suggest we simply call `resolve_annotations` when creating the new _pydantic_ `dataclass`

## Checklist


laevilgenius:
@PrettyWood thanks for the PR. But I believe it doesn't resolve the original issue as I was using pydantic dataclasses, not stdlib ones. 
I was able to workaround it though by calling `update_forward_refs` manually in `__post_init__`.

PrettyWood:
Ah yes thanks for double checking @laevilgenius you're absolutely right!
I caught another issue but yours is still there.
So in your example with `a.py` and `b.py`, the issue lies in  `__annotations__` of `dataclass`
`B.__annotations__` is empty with `BaseModel` but with pydantic `dataclasse`we have
```python
# for class A
namespace == {'__annotations__': {'uuid': 'UUID'}, '__module__': 'a', 'uuid': Ellipsis}

# for class B
namespace == {'__annotations__': {'uuid': 'UUID'}, '__module__': '__main__', 'uuid': Ellipsis}
```
And with postponed evaluation of annotation, we don't have the right module to resolve `UUID` hence the crash...
I don't think we have a way to know if the field 'uuid' is actually the one of the parent or if it has been (re)defined in the class.

samuelcolvin:
@PrettyWood don't we have the same problem on normal models inheriting from `BaseModel`?

Frankly postponed annotations are a big head ache, I'm not all all clear if we can ever have a completely watertight solution, but since they'll become the default in 3.10 I guess we'll have to do our best.

PrettyWood:
@samuelcolvin yes we do have the same problem hence the
> Even though this PR still fixes one problem, I would like to tackle the whole "postponed evaluation" thing properly 
I agree it's a mess but it's something we need to start thinking about. Not directly for v2 but to keep in mind! This PR can probably be closed TBH. But it was a good way for me to start playing with it and have a look at the PEP

samuelcolvin:
makes sense, but it looks like this is still an improvement?

If you think it would be better to merge it, I can easily resolve conflicts and merge?

PrettyWood:
Sorry for the wrong reply I was on my phone yesterday and answered too fast. We don't have the same issue with `BaseModel`. The explanation for `dataclass` can be seen [above](https://github.com/samuelcolvin/pydantic/pull/2220#issuecomment-751391737).
So in the metaclass, resolving annotations works pretty well for `BaseModel` because we don't copy annotations. But for `dataclass` we get a problem because we have all the fields in `__annotations__` including those coming from parent dataclasses.
Honestly we can merge this PR since it still fixes a problem and I'm not sure I'll have time to have a deeper look on the original issue soon

PR #2221, opened 2020-12-24
fix: force need of `update_forward_refs` in recursive models

### LINKED ISSUE #1201: Some recursive models do not require `update_forward_refs` and silently behave incorrectly
# Bug

```
             pydantic version: 1.4
            pydantic compiled: False
                 install path: /Users/dmontague/Programming/oss/pydantic/pydantic
               python version: 3.7.4 (default, Oct 17 2019, 23:32:33)  [Clang 11.0.0 (clang-1100.0.33.8)]
                     platform: Darwin-19.0.0-x86_64-i386-64bit
```

This test passes on master, but it should not:

```py
from typing import Optional, Tuple

def test_parse_nested_tuple():
    class NestedTuple(BaseModel):
        x: Tuple[int, Optional["NestedTuple"]]

    obj = NestedTuple.parse_obj({'x': ('1', {'x': ('2', {'x': ('3', None)})})})
    assert obj.dict() == {'x': (1, {'x': ('2', {'x': ('3', None)})})}
    # Note above: the 2 and 3 did not get parsed into ints !!

    NestedTuple.update_forward_refs()
    obj = NestedTuple.parse_obj({'x': ('1', {'x': ('2', {'x': ('3', None)})})})
    assert obj.dict() == {'x': (1, {'x': (2, {'x': (3, None)})})}
```

Presumably we should either get an error at some point prior to the `NestedRootTuple.update_forward_refs()` call, or the two calls to `obj.dict()` should generate the same output.

I'm guessing this is related to the field being `Optional`, but I'm not sure. I discovered it while working on #1200 .

PrettyWood:
## Change Summary
Currently in nested models, we only check if the type of the field is `ForwardRef` to raise an error.
But we could have `ForwardRef` inside a `Union`, a `Tuple` or anything else...
By not catching this we get a wrong behaviour

## Related issue number
closes #1201

## Checklist

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -25,7 +25,7 @@ from typing import (
 from . import errors as errors_
 from .class_validators import Validator, make_generic_validator, prep_validators
 from .error_wrappers import ErrorWrapper
-from .errors import NoneIsNotAllowedError
+from .errors import ConfigError, NoneIsNotAllowedError
 from .types import Json, JsonWrapper
 from .typing import (
     Callable,
@@ -563,6 +563,12 @@ class ModelField(Representation):
     def validate(
         self, v: Any, values: Dict[str, Any], *, loc: 'LocStr', cls: Optional['ModelOrDc'] = None
     ) -> 'ValidateReturn':
+
+        if self.type_.__class__ is ForwardRef:
+            error_msg = f'field "{self.name}" not yet prepared so type is still a ForwardRef'
+            if cls is not None:
+                error_msg += f', you might need to call {cls.__name__}.update_forward_refs().'
+            raise ConfigError(error_msg)
 
         errors: Optional['ErrorList']
         if self.pre_validators:
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -565,10 +565,11 @@ class ModelField(Representation):
     ) -> 'ValidateReturn':
 
         if self.type_.__class__ is ForwardRef:
-            error_msg = f'field "{self.name}" not yet prepared so type is still a ForwardRef'
-            if cls is not None:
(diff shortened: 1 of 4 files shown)


pydantic/main.py
```diff
@@ -957,12 +957,6 @@ def validate_model(  # noqa: C901 (ignore complexity)
             return {}, set(), ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls_)
 
     for name, field in model.__fields__.items():
-        if field.type_.__class__ == ForwardRef:
```
PrettyWood, line 960: I thought putting this logic in `validate` was easier but we could also change this line into
```python
if ForwardRef in {t.__class__ for t in get_all_args(field.type_)}:
```

where 

```python
def get_all_args(type_: Any) -> Set[type]:
    from .typing import get_args

    args = get_args(type_)
    types: Set[type] = set()

    if args:
        for t in args:
            types.update(get_all_args(t))
    else:
        types.add(type_)

    return types
```

pydantic/fields.py
```diff
@@ -564,6 +564,13 @@ def validate(
         self, v: Any, values: Dict[str, Any], *, loc: 'LocStr', cls: Optional['ModelOrDc'] = None
     ) -> 'ValidateReturn':
 
+        if self.type_.__class__ is ForwardRef:
+            assert cls is not None
```
PrettyWood, line 568: In fact we could probably make `cls` required for `validate()`. AFAIK the only case where `cls` is not set is in two tests. But whatever we do I think it's best to keep with the smallest change as possible.

samuelcolvin:
great, thank you.

PR #2231, opened 2021-01-01
Add exclude as field parameter

### LINKED ISSUE #660: add exclude to Config
From a user perspective I would rather add an exclude in the Config metaclass instead of passing the dict in to .json(). One of my model's fields is a Callable and I would like to call .json() on it, however I need to instead pass a custom encoder to the .json() which still puts the responsibility on the caller.

_Originally posted by @enkoder in https://github.com/samuelcolvin/pydantic/pull/648#issuecomment-511265737_

daviskirk:
## Change Summary

- Add "exclude" as a field parameter so that it can be configured using model config instead of purely at `.dict` / `.json` export time. 
- Main reference used is this comment: https://github.com/samuelcolvin/pydantic/issues/660#issuecomment-515406193

## Related issue number

- Closes #660 

## Checklist


samuelcolvin:
No changes here, just the change description ATM.

tests/test_main.py
```diff
@@ -1209,41 +1270,59 @@ class Bar(BaseModel):
...
-    assert m.dict(exclude=excludes) == {'a': 1, 'bars': [{'y': 2}, {'w': -1, 'z': 3}]}
-    assert excludes == {'bars': {0}}
-    excludes = {'bars': {'__all__'}}
-    assert m.dict(exclude=excludes) == {'a': 1, 'bars': []}
-    assert excludes == {'bars': {'__all__'}}
-    excludes = {'bars': {'__all__': {'w'}}}
-    assert m.dict(exclude=excludes) == {'a': 1, 'bars': [{'x': 1}, {'y': 2}, {'z': 3}]}
-    assert excludes == {'bars': {'__all__': {'w'}}}
+    m = Parent()
+    assert m.dict() == {'a': 1, 'c': {'a': 3, 'c': 'foobar'}, 'd': {'c': 'foobar'}}, 'Unexpected mode export result'
```
daviskirk, line 1325: One of the main design problems with using exclude as a field / model config option is the question of how to deal with nested exclude settings:

Using `.dict(exclude=)` is fairly straightforward because there can also only be one setting.  But using it on fields and model configs is more complex because we can have multiple "exclude" settings on different model levels.

In the test here my intuition as a developer would be that the excludes need to be merged, i.e. "Child.b" is excluded because the Child field exclude is set and "Child.a" is excluded because it is set on the "Parent.d" field (and this test should pass). 

However, since this is probably a bit harder to implement I wanted to check first if this is actually the way we want this designed or if there were already existing thoughts on this.

https://github.com/samuelcolvin/pydantic/pull/1695#issuecomment-753340343 kind of references the confusion that might come about if the excludes were NOT merged), but I'm sure there are arguments against merging as well.

@samuelcolvin sorry for bothering before this is really ready, feel free to say you need a better proof of concept for the merging stuff before we want/can decide anything.

samuelcolvin: Humm, it's very complicated.

On inheritance, exclude should be merged. So if the parent model excludes something, so should the child unless it explicitly decides not to.

On models used as fields in another model it gets more complicated - you would assume that again the field model's exclude should be merged with the outer(?) class. But now what happens when you have both?

There are three problems:
* writing logic that is both robust and performant enough - even the edge cases of `model.dict(exclude=...)` are complex enough that we took a number of shots at it until it worked correctly, and that doesn't require any merging.
* writing documentation that explains such complex behaviour
* having sensible enough behaviour that it does what people who haven't read the documentation (most people) expect

All in all my statement on #660

> we could easily add exclude to a field

Could not have been more wrong.

---

I guess overall the problem is not in the implementation, but in reasoning about the implementation and working out what we expect to happen.

I think the only solution is to have a crack at it and see if the result is usable and intuitive. Does that make sense?

daviskirk: sounds good, I'll give it a try and then see what works and what does not.  Thanks for the feedback!

pydantic/main.py
```diff
@@ -834,6 +834,10 @@ def _iter(
         exclude_none: bool = False,
     ) -> 'TupleGenerator':
 
+        # merge field set excludes with explicit exclude parameter with explicit overriding field set options.
+        field_exclude = {k: v.field_info.exclude for k, v in self.__fields__.items() if v.field_info.exclude}
```
MrMrRobat, line 838: Can we calculate `field_exclude` value once at model creation to save time here?

pydantic/utils.py
```diff
@@ -465,19 +422,30 @@ def __init__(self, value: Any, items: Union['AbstractSetIntStr', 'MappingIntStrA
...
 
         if isinstance(value, (list, tuple)):
             items = self._normalize_indexes(items, len(value))
 
         self._items = items
 
+    @classmethod
+    def _coerce_items(cls, items: Union['AbstractSetIntStr', 'MappingIntStrAny']) -> 'MappingIntStrAny':
+        if isinstance(items, AbstractSet):
+            items = {k: ... for k in items}
```
MrMrRobat, line 436: `dict.fromkeys()` does exact same thing and seems to be faster
```suggestion
            items = dict.fromkeys(items, ...)
```
```py
In [1]: items = {str(i) for i in range(100)}

In [2]: %timeit {k: ... for k in items}
6.39 µs ± 50.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [3]: %timeit dict.fromkeys(items, ...)
3.64 µs ± 31.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

pydantic/utils.py
```diff
@@ -465,19 +422,30 @@ def __init__(self, value: Any, items: Union['AbstractSetIntStr', 'MappingIntStrA
...
 
-        # For further type checks speed-up
-        if isinstance(items, Mapping):
-            self._type = dict
-        elif isinstance(items, AbstractSet):
-            self._type = set
-        else:
-            raise TypeError(f'Unexpected type of exclude value {items.__class__}')
+        items = self._coerce_items(items)
+        self._type = dict
```
MrMrRobat, line 426: Seems like it's redundant now?

MrMrRobat (commented):
Happy to see it's finally getting implemented, thank you!

pydantic/utils.py
daviskirk: Sorry @MrMrRobat I should not have pushed this yet, this is a work in extreme progress ;).  And yes, I'm currently exploring the possibility of removing the necessity for this `self._type` attribute.

pydantic/main.py
daviskirk: Agreed. Currently, I was just using this as a quick way to test the merging functionality.
I still have to combine this with the "model config" option, and when I do that I'll try to pre-compute this is well.

pydantic/utils.py
MrMrRobat: Sure, it's fine 🙂

pydantic/main.py
daviskirk: I haven't done this yet, I'm also not sure how much time this saves us given all the other stuff that is done   in the __iter__ function. If everything else looks good and everyone like the design I'm open to storing this on the class somewhere (but we'd need to think about where)

pydantic/utils.py
daviskirk: did this

pydantic/main.py
MrMrRobat: > I'm also not sure how much time this saves us given all the other stuff that is done in the iter function

In case when there are no excluded or inclided fields, `._iter()` just yields from `.__dict__.items()`, doing something on top of that will be much slower, I think:
```py
if allowed_keys is None and not (to_dict or by_alias or exclude_unset or exclude_defaults or exclude_none):
    # huge boost for plain _iter()
    yield from self.__dict__.items()
```

Also, maybe it's worth precompute some of the other stuff too: for example compute `allowed_keys` from config at model creation might give some performance boost too.

> I'm open to storing this on the class somewhere (but we'd need to think about where)

Computing it in `ModelMetaClass.__new__` after all fields are resolved and storing in one of the `__slots__` should do the trick?  Or you did you mean something more specific?

daviskirk:
Implementation wise, this is just about the way I'd like it.
Since there are probably multiple opinions on how the logic and strategy of merging the various parts should work, it would be great to have some other eyes on this just to make sure I didn't make any stupid assumptions.

A few key points here:

* I consolidated a lot of the "merging" logic from ``ValueItems``. Since I was having to merge in many places the older logic (with 2 mutliple merge functions depending on the keys and types) were to much for my brain to process. If this was done for performance reasons and has to stay, I can try and reimplement it.  The main changes here are:
  * Only one merging method (``ValueItems.merge``)
  * During processing only use dictionaries instead of checking everywhere if the values are dictionaries or sets (or combinations).
* Merge config settings, field level settings and explicit settings.  The order of preference here is: 
  * Explicit exclude/include on dict/json.
  * Field level settings ala ``Field(..., exclude=..., include=...)```
  * Config level settings via "fields":  ``fields = {"a": {"exclude": ..., "include": ...}}``  
* Merging logic:
  * Merging of exclude settings happens as a "union" so overriding a field level exclude of ``{"a"}`` with dict level exclude of ``{"b"}`` will exclude: ``{"a", "b"}``
  * Merging of include settings happens as an "intersection" so overriding a field level include of ``{"a", "b"}`` with a dict level include of ``{"b"}`` will include: ``{"a"}``.   
  * These rules are the most consistent with the way fields are handled on the objects (when I include, ONLY the fields I set in the "include" setting are included, when I exclude everything EXCEPT the fields I set are included). It also should help with not having users to "accidentally" expose a field that was supposed to be kept unexported (for example having set a "mysecretfield" field to "exclude").

MrMrRobat:
> * Merging of include settings happens as an "intersection" so overriding a field level include of `{"a", "b"}` with a dict level include of `{"b"}` will include: `{"a"}`.

I am probably missing something, but why `{"a", "b"} | {"b"}` should result in `{"a"}` rather than `{"b"}`?

daviskirk:
> I am probably missing something, but why `{"a", "b"} | {"b"}` should result in `{"a"}` rather than `{"b"}`?

nope, not missing anything.  Typo on my part, sorry.  I updated the comment.

pydantic/main.py
daviskirk: Did this now, class level exclude/includes are pre-computed and stored on the class.
This does mean that people can't go in and modify the field_info by hand and expect stuff to work... but that's already the case for other attributes as well so I think that's ok.

pydantic/utils.py
```diff
@@ -462,23 +419,15 @@ class ValueItems(Representation):
...
+            self._items: 'MappingIntStrAny'
 
-        # For further type checks speed-up
-        if isinstance(items, Mapping):
-            self._type = dict
-        elif isinstance(items, AbstractSet):
-            self._type = set
-        else:
-            raise TypeError(f'Unexpected type of exclude value {items.__class__}')
+        items = self._coerce_items(items)
```
samuelcolvin, line 424: this cleanup of `ValueItems` makes a lot of sense.

pydantic/utils.py
```diff
@@ -525,38 +465,114 @@ def _normalize_indexes(
...
-                    return normalized_items
-                for i in range(v_length):
-                    normalized_item = normalized_items.setdefault(i, default())
-                    if normalized_item is not ...:
-                        normalized_items[i] = update_normalized_all(normalized_item, all_items)
+
+        if TYPE_CHECKING:
+            normalized_items: 'DictIntStrAny'
+        all_items = None
+        normalized_items = {}
```
samuelcolvin, line 472:
```suggestion
        all_items = None
        normalized_items: 'DictIntStrAny' = {}
```
Should be fine unless I'm missing something?

pydantic/utils.py
```diff
@@ -525,38 +465,114 @@ def _normalize_indexes(
...
+        base = cls._coerce_value(base)
+        if override is None:
+            return base
+        if base is ... or base is None:
+            return override
+        elif override is ...:
+            return base if intersect else override
+
+        if intersect:
+            merge_keys = override.keys() & base.keys()
```
samuelcolvin, line 549: It looks like retaining order can't be guaranteed. I can't work out if that could be a problem? Might be worth adding a note to the docstring?

tests/conftest.py
```diff
@@ -15,7 +15,7 @@
 except ImportError:
     pytest_plugins = []
 else:
-    pytest_plugins = ['hypothesis.extra.pytestplugin']
+    pytest_plugins = []
```
samuelcolvin, line 18: I think this is a mistake?

pydantic/utils.py
```diff
@@ -525,38 +465,114 @@ def _normalize_indexes(
...
+            for i in range(v_length):
+                normalized_items.setdefault(i, ...)
             return normalized_items
+        for i in range(v_length):
+            normalized_item = normalized_items.setdefault(i, {})
+            if normalized_item is not ...:
+                normalized_items[i] = self.merge(all_items, normalized_item)
+        return normalized_items
+
+    @overload
```
samuelcolvin, line 499: Given that this is an internal method and You include the `Optional[Any]` case, maybe all these overloads are overkill?

docs/examples/exporting_models_exclude4.py
```diff
@@ -0,0 +1,26 @@
+from pydantic import BaseModel, Field, SecretStr
+
+
+class User(BaseModel):
+    id: int
+    username: str  # overridden by explicit exclude
+    password: SecretStr = Field(exclude=...)
```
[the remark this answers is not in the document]
daviskirk: I agree. It would probably make sense to have this all the way down the stack and in the other forms (explicit `.dict` exclude for example) as well.  Otherwise, I would find it a bit confusing if `True` were to work sometimes but not other times.

I guess we can't get rid of using `...` though because of backwards compatibility right? Should we just allow both?

```
model.dict(exclude={'test': ...}) == model.dict(exclude={'test': True}) 
model.dict(exclude={'test': {'subfield': True}}) == model.dict(exclude={'test': {'subfield': True}}) 
Field(exclude=True) == Field(exclude=...)
```
?

then discuss if we want to drop the `...` in v2?

pydantic/utils.py
daviskirk: No this looks nicer to me as well

tests/conftest.py
daviskirk: yes, sorry I get a 
```
ValueError: Plugin already registered: hypothesis.extra.pytestplugin=<module 'hypothesis.extra.pytestplugin' from 'envs/pydantic/lib/python3.8/site-packages/hypothesis/extra/pytestplugin.py'>
```
when this is enabled.  But in any case it's a mistake... I have to poke around a bit to try and figure out why that's happening, I'll file an issue if I actually find out what's going on.

Changed it.

pydantic/utils.py
daviskirk: Yes, I got rid of them. I can't really get any consistent typing to work anyway as long as there's no type description for `Ellipsis` and not real support for recursive types.

pydantic/utils.py
daviskirk: Yeah it is if we're using numeric indexes (for example keys -1 and 1).  
Don't know if the slight performance benefit is worth inconsistent behavior in some edge cases. 

I changed the code to preserve ordering for now (if you'd rather keep the direct set operations I can change it back).

Nemesis7:
@samuelcolvin when can we expect this functionality to be released?

docs/examples/exporting_models_exclude4.py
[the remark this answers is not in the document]
samuelcolvin: yes, makes sense.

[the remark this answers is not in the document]
daviskirk: :+1:  thanks! Should be ready from my point of view then. 

A small Question for future PRs: Am I supposed to change the label "awaiting author revision" when the revisions are addressed or just resolve the discussions involved? I could add something to the "contributing" page if you have a preferred flow (and you think it would help cut down on pings if that sort of thing is a problem).

PrettyWood:
@daviskirk Do you think it could be possible to also set a callable as a value?
For example:

```py
class User(BaseModel):
    name: str
    friends: list[str] | None = Field(None, exclude=lambda x: x is None)

assert User(name="Alone", friends=None).json() == '{"name": "Alone"}'
assert User(name="Popular", friends=["Bob", "Alice"]).json() == '{"name": "Popular", "friends": ["Bob", "Alice"]}'
```
WDYT?

daviskirk:
> @daviskirk Do you think it could be possible to also set a callable as a value?

I think that would be possible... I do think that it would open up a pretty large field of possibilities for users to "use"/"misuse" that I'm not sure I can predict right now. 

Given all the larger changes and v2 coming along I'm not sure if we want to commit to such a broad feature without lot's of users actually really wanting it.

For example, I can behavior the behavior of "exclude" (as implemented here already with being able to drill down into nested fields) being popular enough that we might want to expand it to other fields like 
I can imagine that if we allow something like this there will quickly be people that would want the callable to take the entire object as an argument as well (similar to how validators work with the optional "values"), or using this in weird ways using "nested" values as flags and stuff like that.

Not that I'm against this... I think it would be a pretty fancy feature, just trying to look out for you and samuel who might be stuck supporting it ;)

PrettyWood:
You're absolutely right. I just shared a recent need and got away easily with custom `dict`. I didn't have the nesting in mind when I shared the idea but it will surely be expected if a callable support is added. Let's not do that now 😬 
The feature is already super nice and complete 👍 
Thanks for the detailed answer

docs/usage/exporting_models.md
```diff
@@ -162,7 +162,7 @@ sets or dictionaries. This allows nested selection of which fields to export:
 {!.tmp_examples/exporting_models_exclude1.py!}
 ```
 
-The ellipsis (``...``) indicates that we want to exclude or include an entire key, just as if we included it in a set.
+The ``True`` indicates that we want to exclude or include an entire key, just as if we included it in a set.
```
samuelcolvin, line 165:
```suggestion
The `True` indicates that we want to exclude or include an entire key, just as if we included it in a set.
```

samuelcolvin (approved):
otherwise this looks great.

samuelcolvin:
This is awesome. Thank you so much @daviskirk 🎉

rickerp:
Any release predictions on this?

Issue #2234, opened 2021-01-04
Bug: Field ordering is not correct when inheritance.


ghengao:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: /home/ehengao/.local/lib/miniconda3/envs/vscode/lib/python3.8/site-packages/pydantic
               python version: 3.8.3 (default, Jul  2 2020, 16:21:59)  [GCC 7.3.0]
                     platform: Linux-3.10.0-1160.6.1.el7.x86_64-x86_64-with-glibc2.10
     optional deps. installed: ['typing-extensions', 'email-validator']
```

Only related issue i have seen here i https://github.com/samuelcolvin/pydantic/issues/1754, but it is a different concern regardless of the ordering. 

The reason i need the ordering is for the purpose of validation the last field values against other fields with the capability of accessing the `ModelField` information. (Therefore, using `root_validator` does not seem to be convienient.)

```py
from pydantic import BaseModel
import typing as tp

class A(BaseModel):
    a: str = 1
    c: str = 1 
    b: tp.Optional[tp.List[int]]


class B(A):
    d: str = 3
    b: tp.Optional[tp.List[str]]

print(B.__fields__.keys())
# This outputs to dict_keys(['a', 'c', 'b', 'd'])
# It is expected to be dict_keys(['a', 'c', 'd', 'b'])
```

ghengao:
Honestly, i like the idea of dataclassing these models using the `__post_init__` hooks, it would be nice for such hooks to be available within the pydantic model as well.

PrettyWood:
Hello @ghengao! Happy new year and thanks for reporting!
IMO you're absolutely right. Fields that are overwritten in submodels should appear last. The workaround on #1754 was easy but it should have not been necessary. I opened #2236 to show a basic implementation + start discussion. Feedback welcome

samuelcolvin:
I'm really not sure I agree that it's wrong currently. I think the current behaviour could be considered correct.

Either way we need:
* more people to agree that it's currently wrong before we change it
* to wait for v2 for a subtle change like this

leosussan:
[...]

PR #2236, opened 2021-01-04
[wip] change field order to have overwritten fields at the end

PrettyWood:
## Change Summary
When we overwrite parent fields in submodels, we can't trigger validation because the order of the fields is not updated.
⚠️ This PR is just to **showcase** a possible workaround
but it breaks current behaviour and will **NOT** be merged as is of course ⚠️ 

```python
"""
Case that works at the moment on `master` but would break with this PR
"""
from typing import Dict

class M1(BaseModel):
    a: int
    b: int

    @validator('b')
    def assert_b_equal_a(cls, v: int, values: Dict[str, int]) -> int:
        assert 'a' in values and v == values['a']
        return v


class M2(M1):
    a: int = 3

print(repr(M2(b=3)))
# M2(a=3, b=3)
```

```python
"""
Case that would now work and currently fails on `master`
"""
from typing import Dict

class M1(BaseModel):
    a: int
    b: int


class M2(M1):
    a: float = 1.5

    @validator('a', always=True)
    def assert_b_is_twice_a(cls, v: int, values: Dict[str, int]) -> int:
        assert 'b' in values and v == values['b'] / 2
        return v

print(repr(M2(b=3)))
# M2(b=3, a=1.5)
```

## Related issue number
#2234

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -2,6 +2,7 @@ import json
 import sys
 import warnings
 from abc import ABCMeta
+from collections import OrderedDict
 from copy import deepcopy
 from enum import Enum
 from functools import partial
@@ -71,6 +72,7 @@ if TYPE_CHECKING:
 
     ConfigType = Type['BaseConfig']
     Model = TypeVar('Model', bound='BaseModel')
+    ModelFieldsDict = OrderedDict[str, ModelField]
 
     class SchemaExtraCallable(typing_extensions.Protocol):
         @overload
@@ -83,6 +85,14 @@ if TYPE_CHECKING:
 
 
 else:
+
+    class ModelFieldsDict(OrderedDict):
+        """Store fields in the order they were last updated"""
+
+        def __setitem__(self, field_name: str, field: ModelField) -> None:
+            super().__setitem__(field_name, field)
+            self.move_to_end(field_name)
+
     SchemaExtraCallable = Callable[..., None]
 
 try:
@@ -210,7 +220,7 @@ _is_base_model_class_defined = False
 class ModelMetaclass(ABCMeta):
     @no_type_check  # noqa C901
     def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
-        fields: Dict[str, ModelField] = {}
+        fields = ModelFieldsDict()
         config = BaseConfig
         validators: 'ValidatorListDict' = {}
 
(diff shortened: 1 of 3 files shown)


pydantic/main.py
```diff
@@ -71,6 +72,7 @@
 
     ConfigType = Type['BaseConfig']
     Model = TypeVar('Model', bound='BaseModel')
+    ModelFieldsDict = OrderedDict[str, ModelField]
```
PrettyWood, line 75: Had to go with the `if TYPE_CHECKING` because of [this issue](https://mypy.readthedocs.io/en/stable/common_issues.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime)

tests/test_aliases.py
```diff
@@ -345,9 +344,9 @@ def alias_generator(x):
         'e_generator_parent',
     ]
     assert [f.alias for f in Child.__fields__.values()] == [
-        'a_field_child',
         'b_config_child',
         'c_field_parent',
         'd_config_parent',
         'e_generator_child',
+        'a_field_child',
```
ghengao, line 351: Thanks a lot for the quick fix here. I think this behavior is consistent with the dataclasses as well

```python
from dataclasses import dataclass, fields

@dataclass(order=True)
class A:
    a: int
    c: int = 0
    b: int = 1

@dataclass(order=True)
class B(A):
    e: int = 1
    a: int = 1
 
>>> fields(B)
(Field(name='a',type=<class 'int'>,default=1,default_factory=<dataclasses._MISSING_TYPE object at 0x7fedf0188760>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD),
 Field(name='c',type=<class 'int'>,default=0,default_factory=<dataclasses._MISSING_TYPE object at 0x7fedf0188760>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD),
 Field(name='b',type=<class 'int'>,default=1,default_factory=<dataclasses._MISSING_TYPE object at 0x7fedf0188760>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD),
 Field(name='e',type=<class 'int'>,default=1,default_factory=<dataclasses._MISSING_TYPE object at 0x7fedf0188760>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD))
```

Just have two comments here

1. If the child has been override here and move to the end, will that affect inherited validators from parent as well because of this order change? 
2. Maybe it is worth adding a inherited child field in the class body directly as another test scenario as well as adding a inherited field that intentionally changed the order to see if the changed order is taken effects

PrettyWood: The behaviour is actually non consistent with dataclasses but behind the scene it's quite different. Dataclass does not do validation and all annotations are passed to child classes.
This PR was more a showcase to open a discussion than a real proposal (see description of the PR to see why it can't be accepted as is)

samuelcolvin:
see my comment on #2234, not sure about this, but it definitely can't happen now.

Feel free to leave this open or close it.

PrettyWood:
Was definitely just a showcase. Happy to close it ;)

Issue #1673, opened 2020-06-30
'Partial' Equivalent


kalzoo:
# Question

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /.../venv/lib/python3.7/site-packages/pydantic
               python version: 3.7.4 (default, Sep  7 2019, 18:27:02)  [Clang 10.0.1 (clang-1001.0.46.4)]
                     platform: Darwin-19.4.0-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']
```

Hey Samuel & team,

First of all, huge fan of Pydantic, it makes working in Python so much better. I've been through the docs and many of the issues several times, and until this point I've found all the answers I was looking for. If I missed one in this case - sorry in advance!

Looking to be able to transform a model at runtime to another with the same fields but where all are `Optional`, akin to Typescript's [Partial](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialt) (hence the issue title). Something similar is discussed in part in [at least one FastAPI issue](https://github.com/tiangolo/fastapi/issues/125), but since the answer there is basically "duplicate your model" and since this is inherently a `pydantic` thing, I thought I'd ask here.

By way of a little bit of hacking on model internals, I've been able to get the functionality I need for a relatively trivial use case:

```py
from __future__ import annotations

from pydantic import BaseModel, create_model
from typing import Optional

class BaseSchemaModel(BaseModel):
    # ...

    @classmethod
    def to_partial(cls) -> BaseSchemaModel:
        return get_partial_model(cls)


def get_partial_model(model: BaseModel) -> BaseModel:
    """
    Return a model similar to the original, but where all fields are Optional.

    Note: this minimal implementation means that many Pydantic features will be discarded
    (such as alias). This is not by design and could stand to improve.
    """
    new_fields = {
        name: (Optional[model.__annotations__.get(name)], model.__fields__[name].default or None)
        for name in model.__fields__
    }
    return create_model("GeneratedPartialModel", **new_fields, __config__=model.__config__)
```

verified with a simple test:

```
import pytest

from pydantic import ValidationError
from typing import Optional

from app.utilities.pydantic import BaseSchemaModel


def test_partial_model():
    class TestModel(BaseSchemaModel):
        a: int
        b: int = 2
        c: str
        d: Optional[str]

    with pytest.raises(ValidationError):
        TestModel()

[...]

PR #2245, opened 2021-01-07
Allow for fragments of models (partials)

dashavoo:
## Change Summary

This change allows for fragments of models, described in #1673 as "Partial"s, but I personally don't like the overlap with functools.partial.

A `Fragment` of a model validates all received fields against that model, but doesn't require any. An example use case is an HTTP PATCH.

It is intended to be used like this:
```python
from pydantic import BaseModel, conint
from pydantic.fragments import Fragment

class Customer(BaseModel):
    name: str
    age: conint(ge=18)

Fragment[Customer](name="John") .dict() # {'name': 'John'}
Fragment[Customer](age=14)  # validation error
```

I didn't want to submit this until it was completely ready to go, but I'd appreciate feedback and advice on a couple of things.
1) In my current implementation, trying to access the schema of a model which contains a Fragment field fails - I am not certain of where to address this, nor what the schema should actually be!
2) The current solution has some hacky aspects because I am unable to introspect the type of the original model without having it as a field. I'd appreciate some comments from someone who understands generics better than I do. Most of the code the test doesn't currently cover is stuff that I would remove if I could get around this aspect.

## Related issue number

#1673

## Checklist

### DIFF
--- a/pydantic/fragments.py
+++ b/pydantic/fragments.py
@@ -33,22 +33,35 @@ class Fragment(GenericModel, Generic[T]):
     attributes to be set.
     """
 
+    model_: Type[T]
+
     class Config:
         extra = 'allow'  # essentially everything is extra.
         allow_mutation = False
 
-    def __init__(__pydantic_self__, **data: Any):
+    def __init__(__pydantic_self__, __model__: Optional[Type[T]] = None, **data: Any):
         """Initialises the Fragment by validating against __model__'s fields.
         Args:
             __model__: The model this is a fragment of - positional only.
             **data: The fields expected by the __model__.
         """
 
-        if __pydantic_self__.__concrete__:
-            expected_model_type = type(__pydantic_self__)
+        model = __model__  # __model__ is underscored in the method arguments to prevent a name clash.
+
+        if model is None and __pydantic_self__.__concrete__:
+            expected_model_type = __pydantic_self__.__fields__['model_'].type_.__args__[0]
             model = expected_model_type
-        else:
+
+        elif model and not __pydantic_self__.__concrete__:
+            # Still generic
+            expected_model_type = model
+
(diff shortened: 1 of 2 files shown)


dashavoo:
#2260 or #2262 might address my issue with the schemas (or be useful in addressing it).

samuelcolvin:
See my comments on #1673, I'm not yet sure about this for a number of reasons.

A few more problems that are coming to me as I think about this:
* `Fragment` is a confusing name, it sounds like we've got part of the model not an "all optional" version of it
* This will not play nicely with the changes to optional and required fields in v2, see https://github.com/samuelcolvin/pydantic/discussions/2353#discussioncomment-365116


Sorry, I will think about this more, but overall I'm leaning towards closing this PR and either letting people implement their own solutions, or doing something else.

samuelcolvin:
I don't think this is the best wait forward right now. Closing while trying  to cleanup.

dashavoo:
Sorry I missed your previous comments on this - email overload.

If this implementation isn't up to scratch no problem, but to address your points:

* Perhaps I am using the pydantic terminology wrong, but to me Fragment works because like `List[SomeModel]` is a list of `SomeModel` instances, `Fragment[SomeModel]` is a fragment of a `SomeModel` instance - it is incomplete. I don't care too much about the name though.
* This not playing nicely with upcoming changes in pydantic is exactly why I believe this functionality should be part of pydantic and not everyone roll their own. As far as I can tell it isn't possible to do this in a way that works well with mypy without relying on implementation details of pydantic, and if v2 is incompatible with my current implementation I already have a project that won't be upgradable to v2 of pydantic until a better alternative exists.

As far as I can tell your solution using `validate_model` directly doesn't help with a case like in my `test_fragment_container` example.

Issue #2249, opened 2021-01-10
`validate_arguments` allows overriding positional argument with kwarg


cybojenix:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: /home/anthony/.local/share/virtualenvs/project-4BYEWHa2/lib/python3.8/site-packages/pydantic
               python version: 3.8.5 (default, Aug 23 2020, 18:40:40)  [GCC 10.1.0]
                     platform: Linux-5.10.2-2-MANJARO-x86_64-with-glibc2.29
     optional deps. installed: ['typing-extensions']
```

```py

from pydantic import validate_arguments


@validate_arguments
def foo(x: int):
    print(x)


def bar(x: int):
    print(x)

>>> foo(1)
1
>>> foo(1, x=2)
2

>>> bar(1)
1
>>> bar(1, x=2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: bar() got multiple values for argument 'x'
```

Currently, `validate_arguments` does not behave the same as a native function when passing in kwargs that override a positional argument.

expectation is that `foo(1, x=2)` would raise a ValidationError or TypeError

PR #2250, opened 2021-01-10
fix(decorator): detect when a duplicate argument is passed in

cybojenix:
## Change Summary

Currently, it is possible to override a positional argument with a keyword argument using the decorator.

In native python, this raises a TypeError.

Detect this usecase, to prevent keywords overriding positional parameters.

## Related issue number

Closes: #2249

## Checklist


docs/examples/validation_decorator_parameter_types.py
```diff
@@ -65,4 +65,4 @@ def armageddon(
 
 
 print(armageddon(1, 2, e=3))
-print(armageddon(1, 2, 3, 4, 5, 6, c=7, e=8, f=9, g=10, spam=11))
```
cybojenix, line 68: Previously:
```py
print(armageddon(1, 2, e=3))
#> a=1 b=2 c=None d=() e=3 f=None g={}
print(armageddon(1, 2, 3, 4, 5, 6, c=7, e=8, f=9, g=10, spam=11))
#> a=1 b=2 c=7 d=(4, 5, 6) e=8 f=9 g={'spam': 11}
```
Now:
```py
print(armageddon(1, 2, e=3))
#> a=1 b=2 c=None d=() e=3 f=None g={}
print(armageddon(1, 2, 3, 4, 5, 6, e=8, f=9, g=10, spam=11))
#> a=1 b=2 c=3 d=(4, 5, 6) e=8 f=9 g={'spam': 11}
```

I'm unsure if this counts as a breaking change or a bug fix now, as it seems to have been documented that you could override positional arguments with a keyword argument.
While this example did get flagged by mypy in my editor, I'm conscious that people could override parameters using `**{'c': 7}`.

PrettyWood: IMO this counts as a fix : the behaviour should not be different from plain python. It was probably involuntary 🤔

samuelcolvin: agreed

pydantic/decorator.py
```diff
@@ -105,6 +106,7 @@ def __init__(self, function: 'AnyCallableT', config: 'ConfigType'):  # noqa C901
             elif p.kind == Parameter.POSITIONAL_OR_KEYWORD:
                 self.arg_mapping[i] = name
                 fields[name] = annotation, default
+                fields[V_DUPLICATE_KWARGS] = List[str], None
```
samuelcolvin, line 109: This line is very strange, it's either some strange kung fu madness I don't understand, or a typo, do you mean:
```suggestion
                fields[V_DUPLICATE_KWARGS]: List[str] = None
```
?

If not requires a comment explaining it. If that is what you meant, I have no idea how it's passing both mypy and  tests 🙃 .

samuelcolvin (commented):
otherwise LGTM.

pydantic/decorator.py
cybojenix: This is inline with how `V_POSITIONAL_ONLY_NAME` works.

`fields` is a dictionary of annotations + defaults (as tuples), so I think everything here is as expected

samuelcolvin: sorry, yes, my mistake, reading  the whole function makes more sense.

samuelcolvin:
thanks so much.

PR #2252, opened 2021-01-10
fix(decorator): align var args + keywords behaviour with native python

### LINKED ISSUE #2251: `validate_arguments` does not assign `**kwargs` correctly
### Checks

# Bug

```
             pydantic version: 1.7.3
            pydantic compiled: False
                 install path: /home/anthony/experiments/pydantic/pydantic
               python version: 3.8.5 (default, Aug 23 2020, 18:40:40)  [GCC 10.1.0]
                     platform: Linux-5.10.2-2-MANJARO-x86_64-with-glibc2.29
```

```py
from pydantic import validate_arguments

@validate_arguments
def foo(**kwargs):
    print(f"{kwargs=}")


def bar(**kwargs):
    print(f"{kwargs=}")

>>> foo(kwargs=1, b=2)
kwargs={'b': 2}
>>> bar(kwargs=1, b=2)
kwargs={'kwargs': 1, 'b': 2}
```

Kwargs doesn't get assigned as per native python implementation.

It looks like the field gets overwritten, as we're checking all input kwargs against all fields, including the special var_kwarg field.
https://github.com/samuelcolvin/pydantic/blob/8bad7bc91105616dfcc23ce1d68fba12f328a2a3/pydantic/decorator.py#L166

Should be as simple as `set(self.model.__fields__) - {self.v_kwargs_name}` to resolve, will prepare a patch now.

cybojenix:
## Change Summary

Previously, validate_arguments would strip the value if the keyword
was the same as the target args/kwargs argument.

Align with Python's behaviour.

## Related issue number

Closes #2251

## Checklist

### DIFF
--- a/pydantic/decorator.py
+++ b/pydantic/decorator.py
@@ -161,7 +161,7 @@ class ValidatedFunction:
 
         var_kwargs = {}
         wrong_positional_args = []
-        non_var_fields = set(self.model.__fields__) - {self.v_kwargs_name}
+        non_var_fields = set(self.model.__fields__) - {self.v_args_name, self.v_kwargs_name}
         for k, v in kwargs.items():
             if k in non_var_fields:
                 if k in self.positional_only_args:
@@ -178,9 +178,7 @@ class ValidatedFunction:
 
     def execute(self, m: BaseModel) -> Any:
         d = {k: v for k, v in m._iter() if k in m.__fields_set__}
-        kwargs = d.pop(self.v_kwargs_name, None)
-        if kwargs:
-            d.update(kwargs)
+        var_kwargs = d.pop(self.v_kwargs_name, None) or {}
 
         if self.v_args_name in d:
             args_: List[Any] = []
@@ -194,7 +192,7 @@ class ValidatedFunction:
                     in_kwargs = True
                 else:
                     args_.append(value)
-            return self.raw_function(*args_, **kwargs)
+            return self.raw_function(*args_, **kwargs, **var_kwargs)
         elif self.positional_only_args:
             args_ = []
             kwargs = {}
@@ -203,9 +201,9 @@ class ValidatedFunction:
                     args_.append(value)
                 else:
                     kwargs[name] = value
-            return self.raw_function(*args_, **kwargs)
+            return self.raw_function(*args_, **kwargs, **var_kwargs)
         else:
-            return self.raw_function(**d)
+            return self.raw_function(**d, **var_kwargs)
 
     def create_model(self, fields: Dict[str, Any], takes_args: bool, takes_kwargs: bool, config: 'ConfigType') -> None:
         pos_args = len(self.arg_mapping)
--- a/pydantic/decorator.py
+++ b/pydantic/decorator.py
@@ -178,7 +178,7 @@ class ValidatedFunction:
 
     def execute(self, m: BaseModel) -> Any:
         d = {k: v for k, v in m._iter() if k in m.__fields_set__}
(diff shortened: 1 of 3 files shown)


tests/test_decorator.py
```diff
@@ -119,6 +119,7 @@ def foo(a, b, *args, d=3, **kwargs):
     assert foo(1, 2, 3, d=4) == 'a=1, b=2, args=(3,), d=4, kwargs={}'
     assert foo(*[1, 2, 3], d=4) == 'a=1, b=2, args=(3,), d=4, kwargs={}'
     assert foo(1, 2, 3, e=10) == "a=1, b=2, args=(3,), d=3, kwargs={'e': 10}"
+    assert foo(1, 2, kwargs=4, e=5) == "a=1, b=2, args=(), d=3, kwargs={'kwargs': 4, 'e': 5}"
```
PrettyWood, line 122: Hi @cybojenix and thanks for the PR! Nice catch!
Two remarks:
- Could we enforce in the test that we expect the exact same behaviour than plain python.
For example by doing something like
```python
@pytest.mark.parametrize('validated', (True, False))
def test_var_args_kwargs(validated):
    def foo(a, b, *args, d=3, **kwargs):
        return f'a={a!r}, b={b!r}, args={args!r}, d={d!r}, kwargs={kwargs!r}'

    if validated:
        foo = validate_arguments(foo)

    assert ...
```
- Don't we have the same error with `args` i.e. `foo(1, 2, args=4, e=5)`, which doesn't seem to be fixed with this PR. I reckon this could be tackled at the same time ;)

cybojenix: Added. You were right, `args` also gets overwritten.

pydantic/decorator.py
```diff
@@ -177,9 +178,7 @@ def build_values(self, args: Tuple[Any, ...], kwargs: Dict[str, Any]) -> Dict[st
 
     def execute(self, m: BaseModel) -> Any:
         d = {k: v for k, v in m._iter() if k in m.__fields_set__}
-        kwargs = d.pop(self.v_kwargs_name, None)
-        if kwargs:
-            d.update(kwargs)
+        var_kwargs = d.pop(self.v_kwargs_name, None) or {}
```
PrettyWood, line 181:
```suggestion
        var_kwargs = d.pop(self.v_kwargs_name, {})
```

PrettyWood (approved):
Otherwise LGTM!

pydantic/decorator.py
cybojenix: I believe `d[v_kwargs_name]` is actually `None` at this point if no var keywords are provided.

Using `or {}` ensures we have a dictionary, simplifying later code

cybojenix: never mind! all tests pass. realise that this is from a model instance, so it will always be a dictionary.

PrettyWood: When I read `d.pop(self.v_kwargs_name, None) or {}` I thought you wanted to handle the `None` case but I thought it was impossible. For me it was always either unset or a dict hence my suggestion. Sorry if I was mistaken

tests/test_decorator.py
```diff
@@ -110,15 +110,22 @@ def foo(a, b, c='x', *, d='y'):
     assert foo(1, {'x': 2}, c='3', d='4') == "1, {'x': 2}, 3, 4"
 
 
-def test_var_args_kwargs():
-    @validate_arguments
+@pytest.mark.parametrize('validated', (True, False))
```
samuelcolvin, line 113: smart!

samuelcolvin (approved):

PR #2265, opened 2021-01-15
fix: Change non-existent secrets dir into a warning

### LINKED ISSUE #2175: secrets_dir configuration setting raises an error when the directory doesn't exist
### Checks

# Bug

```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: ~/.pyenv/versions/3.8.5/envs/fetcher/lib/python3.8/site-packages/pydantic
               python version: 3.8.5 (default, Sep  4 2020, 14:39:18)  [GCC 9.3.0]
                     platform: Linux-5.8.0-7630-generic-x86_64-with-glibc2.29
     optional deps. installed: ['typing-extensions', 'devtools']
```

I have declared a settings model that can read values from the environment, a `.env` file, or from Docker secrets as follows.

```py
import pydantic

class Settings(BaseSettings):

    password: SecretStr = Field(..., env="APP_PASSWORD")

    class Config:
        """Configure settings values discovery locations."""

        env_file = ".env"
        secrets_dir = "/run/secrets"
```

[As per the documentation](https://pydantic-docs.helpmanual.io/usage/settings/#field-value-priority), checking for a value from a secret is the fourth attempt pydantic makes. However, when I just work in my development environment where the directory `/run/secrets` doesn't exist, instantiating the model raises an error.

```
pydantic.env_settings.SettingsError: directory "/run/secrets" does not exist
```

This is annoying because it happens even though I have the values in a `.env` file. Thus I have to circumvent this by using `Settings(_secrets_dir=".")` which would mean that I need a special case for my dev environment. I think, when values are provided to the instance, or from the environment, or from `.env` first, it shouldn't matter whether the `secrets_dir` exists or not.

davidolrik:
## Change Summary

Previously a non existent secrets directory would throw an error, this has now been turned into a warning.

This makes local development easier, as a secrets directory is usually something that only exists in Kubernetes or Docker.

## Related issue number

Fixes #2175

## Checklist


PrettyWood (commented):
LGTM if this is the desired behaviour.
Note that it will conflict with https://github.com/samuelcolvin/pydantic/pull/2107

pydantic/env_settings.py
```diff
@@ -61,8 +61,10 @@ def _build_secrets_files(self, _secrets_dir: Union[Path, str, None] = None) -> D
         secrets_path = Path(secrets_dir).expanduser()
 
         if not secrets_path.exists():
-            raise SettingsError(f'directory "{secrets_path}" does not exist')
-        if not secrets_path.is_dir():
+            warnings.warn(f'directory "{secrets_path}" does not exist')
+            return secrets
+
+        if secrets_path.exists() and not secrets_path.is_dir():
```
PrettyWood, line 67:
```suggestion
        if not secrets_path.is_dir():
```

changes/2175-davidolrik.md
```diff
@@ -0,0 +1 @@
+fix: Change non-existent secrets dir from exception to warning
```
PrettyWood, line 1:
```suggestion
Allow non-existent secrets directories by issuing only a warning
```

davidolrik:
```suggestion
Allow non-existent secrets directory by only issuing a warning
```

samuelcolvin:
thanks so much. I think this makes a lot of sense.

PR #2282, opened 2021-01-22
fix: keep order of fields with `BaseModel.construct()`

### LINKED ISSUE #2281: BaseModel.construct(): Incorrect field order with default values
### Checks

# Bug

```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: C:\Users\JonasTK\miniconda3\envs\testenv\Lib\site-packages\pydantic
               python version: 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
```

I know this is (sort of) showcased in the docs, but I still think it is a problem that .construct() does not give correct field order when there are fields with default values. Most of the time this is not a problem: comparing instances and .dict() outputs do not depend on the field order. But for JSON export, it does matter.

Shouldn't the JSON export from an instance generated via .construct() with pre-validated input be identical to the JSON export of a properly validated instance?

```py


class Foo(BaseModel):
    a: int
    b: int = 42
    c: float


instance = Foo(a=1, c=3.14)  # Correct field order
instance_construct = Foo.construct(**instance.dict())  # Incorrect field order

print(instance)            # Output: a=1 b=42 c=3.14
print(instance_construct)  # Output: b=42 a=1 c=3.14

print(instance.json())            # Output: {"a": 1, "b": 42, "c": 3.14}
print(instance_construct.json())  # Output: {"b": 42, "a": 1, "c": 3.14}

assert instance == instance_construct  # No Error
assert instance.dict() == instance_construct.dict()  # No Error
assert instance.json() == instance_construct.json()  # AssertionError
```

PrettyWood:
## Change Summary
Order of fields with `BaseModel.construct()` was not maintained with the current implementation.
I changed it a little. It doesn't seem to have a bad impact of performance (which is the whole purpose of `construct`)

```python


class Foo(BaseModel):
    a: int
    b: int = 42
    c: float

values = {k: 1 for k in 'qwertyuiopasdfghjklzxcvbnm'}

def f():
    Foo.construct(**{'a': 1, 'b': 42, 'c': 3.14})
    Foo.construct(**{'a': 1, 'c': 3.14})
    Foo.construct(**{'a': 1, 'c': 3.14, 'qwe': 123, **values})


if __name__ == '__main__':
    import timeit
    print(timeit.timeit("f()", number=10_000_000, setup="from __main__ import f"))

# 133.25740689 BEFORE
# 125.656012296 AFTER
```

## Related issue number
closes #2281

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -586,9 +586,12 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         Default values are respected, but no other validation is performed.
         """
         m = cls.__new__(cls)
-        # default field values
-        fields_values = {name: field.get_default() for name, field in cls.__fields__.items() if not field.required}
-        fields_values.update(values)
+        fields_values: Dict[str, Any] = {}
+        for name, field in cls.__fields__.items():
+            if name in values:
+                fields_values[name] = values[name]
+            elif not field.required:
+                fields_values[name] = field.get_default()
         object_setattr(m, '__dict__', fields_values)
         if _fields_set is None:
             _fields_set = set(values.keys())
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -584,6 +584,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         """
         Creates a new model setting __dict__ and __fields_set__ from trusted or pre-validated data.
         Default values are respected, but no other validation is performed.
+        Behaves as if `Config.extra = 'allow'` was set since it adds all passed values
         """
         m = cls.__new__(cls)
         fields_values: Dict[str, Any] = {}
@@ -592,6 +593,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
                 fields_values[name] = values[name]
             elif not field.required:
                 fields_values[name] = field.get_default()
+        fields_values.update(values)
         object_setattr(m, '__dict__', fields_values)
         if _fields_set is None:
             _fields_set = set(values.keys())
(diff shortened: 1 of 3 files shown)


layday:
This changes the behaviour of `construct` in that it requires the `values` to be in the model's `__fields__`, i.e. it used to behave as if `extra` was set to `allow`.

PrettyWood:
You're absolutely right! Thanks @layday 👍

pydantic/main.py
```diff
@@ -584,10 +584,15 @@ def construct(cls: Type['Model'], _fields_set: Optional['SetStr'] = None, **valu
...
         m = cls.__new__(cls)
-        # default field values
-        fields_values = {name: field.get_default() for name, field in cls.__fields__.items() if not field.required}
+        fields_values: Dict[str, Any] = {}
+        for name, field in cls.__fields__.items():
+            if name in values:
+                fields_values[name] = values[name]
+            elif not field.required:
+                fields_values[name] = field.get_default()
         fields_values.update(values)
```
samuelcolvin, line 596: surely this line can now be removed?

samuelcolvin (commented):
otherwise LGTM

pydantic/main.py
PrettyWood: This is why I did first but if you look at the 2nd commit, I actually had to add it because `construct()` currently behaves as if `Config.extra = 'allow'` is set

samuelcolvin: makes sense, see #2354

samuelcolvin:
thanks so much.

PR #2294, opened 2021-01-27
fix #2293: Properly encode Decimals without any decimal places.

### LINKED ISSUE #2293: Can't roundtrip json encode/prase properly with ConstrainedDecimal
### Checks

# Bug

```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: /Users/hultner/Library/Caches/pypoetry/virtualenvs/hantera-backend-5YPPF0bI-py3.9/lib/python3.9/site-packages/pydantic
               python version: 3.9.0 (default, Oct 10 2020, 11:40:52)  [Clang 11.0.3 (clang-1103.0.32.62)]
```

When using a `ConstrainedDecimal` with zero decimal_places (as provided in example below), pydantic incorrectly encodes the value as a float, thus resulting in failure if one tries to parse the recently encoded json object. 

Using a decimal with `max_digits = x` and `decimal_places = 0` is a great way of representing for instance a `Numeric(22,0)` (where x is 22) from many SQL database schemas. Certain database engines like the popular pyodbc will properly handle and convert such a Decimal value, but won't handle it as an int as this is implicitly interpreted as a 32 bit int by pyodbc. Having a fixed number of max_digits also allows ones query-engine to pre-compile reusable query plans, which other wise would have to be recomputed for every length of of the given number.
In other words, using a ConstrainedDecimal for this type of data is ideal.

I have provided a minimal test/example which can both be executed directly but also ran with pytest to showcase the issue at hand.

```py
"""
Self contained example showcasing problem with decimals using pydantics
default encoder. 
"""
from pydantic import ConstrainedDecimal, BaseModel
from decimal import Decimal


class Id(ConstrainedDecimal):
    max_digits = 22
    decimal_places = 0
    ge = 0


ObjId = Id


class Obj(BaseModel):
    id: ObjId
    name: str
    price: Decimal = Decimal(0.00)


class ObjWithEncoder(BaseModel):
    id: ObjId
    name: str
    price: Decimal = Decimal(0.00)

    class Config:
        json_encoders = {
            Id: int,
        }


def test_con_decimal_encode() -> None:
    test_obj = Obj(id=1, name="Test Obj")
    cycled_obj = Obj.parse_raw(test_obj.json())
    assert test_obj == cycled_obj


def test_con_decimal_encode_custom_encoder() -> None:
    test_obj = ObjWithEncoder(id=1, name="Test Obj")
    cycled_obj = ObjWithEncoder.parse_raw(test_obj.json())
    assert test_obj == cycled_obj


if __name__ == "__main__":
    from pprint import pprint

    print_obj = Obj(id=1, name="Test Obj", price=1.23)
    json_obj = print_obj.json()
    pprint(json_obj)
```

I have a small patch for pydantic.json-module which I can provide as a pull-request. When using said patch all the tests above will pass, notice that I both handle the case where Decimals do have an negative exponent (decimal_places) and the case where it doesn't. The patch handles both these cases as expected and is written in a minimal, and of course readable fashion. 

I am right now in the process of reading through the contributor guidelines to ensure that my patch is up to the standards which the project holds for contributions.

Hultner:
## Change Summary
Adds a custom encoder for Decimals which handles both decimals with decimal places as well as those without.

## Related issue number
Closes: #2293 

## Checklist

  __Tests where at 99% when i ran them before my change, I do have tests covering my changes__


Hultner:
@PrettyWood @samuelcolvin Any thoughts on this?

changes/2293-hultner.md
```diff
@@ -0,0 +1 @@
+fix #2293: Properly encode `Decimal` with, or without any decimal places.
```
samuelcolvin, line 1:
```suggestion
Properly encode `Decimal` with, or without any decimal places.
```

tests/test_decimal.py
```diff
@@ -0,0 +1,28 @@
+"""
```
samuelcolvin, line 1: I know the test files could do with cleaning up, but this doesn't require a standalone test file.

Please move into `test_json`.

samuelcolvin (commented):
otherwise LGTM.

tests/test_decimal.py
Hultner: Have done that now :)

changes/2293-hultner.md
Hultner: Fixed

Hultner:
@samuelcolvin Thanks for the feedback! I know that you are busy. 😃 
I moved the test into test_json, right below the time-delta encoder as that's the functionality I saw the most resemblance with. I also removed the prefix from the changelog entry and resolved the merge conflicts agains master which have arose. 

Does it look good now?

tests/test_json.py
```diff
@@ -170,6 +170,29 @@ class Config:
...
+    Makes sure a decimal with decimal_places = 0, as well as one with places
+    can handle a encode/decode roundtrip.
+    """
+
+    class Id(ConstrainedDecimal):
+        max_digits = 22
+        decimal_places = 0
+        ge = 0
+
+    ObjId = Id
```
samuelcolvin, line 184: This is unnecssary, better to use `condecimal` anyway.

tests/test_json.py
```diff
@@ -170,6 +170,29 @@ class Config:
...
+
+    class Obj(BaseModel):
+        id: ObjId
+        name: str
```
samuelcolvin, line 188: this field is not relevant to the test.

tests/test_json.py
```diff
@@ -170,6 +170,29 @@ class Config:
...
+        price: Decimal = Decimal('0.01')
+
+    test_obj = Obj(id=1, name='Test Obj')
+    cycled_obj = Obj.parse_raw(test_obj.json())
+    assert test_obj == cycled_obj
```
samuelcolvin, line 193: This is smart, but not the cleanest way of writing tests. Tests should demonstrate what they're checking explicitly.

Better to do something like

```py
assert Obj(id=1).json() == '{"id": 1, "price": 0.01}'
```

Hultner: I wanted to include this since the problematic behaviour were strictly related to encoding and then parsing an object. 
But I'll change it if you prefer the other way. However it doesn't capture that parsing the object back also works so maybe add a second parse as well?

Using your method that could look like this.
```python
assert Obj(id=1) == Obj.parse_raw('{"id": 1, "price": " 0.01}')
```

tests/test_json.py
Hultner: Yes, I'll remove it.

tests/test_json.py
Hultner: You mean the ConstrainedDecimal subclass?  I'm cursion to hear why is condecimal better?

I use the subclasses since typers doesn't like the `id: condecimal(args)` approach, and it's neat with reusable types.

Hultner:
@samuelcolvin I updated the test and responded to your comments :)

samuelcolvin:
This is great, thank you.

PR #2312, opened 2021-02-02
Add docs warning about overwriting validate

### LINKED ISSUE #2289: Add some documentation to warn about `BaseModel.validate()` method
### Checks

# Bug
It seems like Literals are not properly evaluated when a root validator is added to a model. In the example below, a Literal would be expected to discriminate which submodel the data is matched to based on the operator value (if an arithmetic operator is provided, it should use the ArithmeticOperators model, if boolean BooleanOperators, etc.). However, the models are not properly determined if the root validator on the BooleanOperators model passes and returns a value. It would be expected that because of the Literal operator, the provided data should not match the BooleanOperators model and should instead match the ArithmeticOperators model first; instead, the model never reaches the ArithmeticOperators validation.

### Example

```py
from typing import Literal, List, Dict, Any, Union

import inflection
import pydantic
from pydantic import Field, root_validator, BaseModel


class Parent(pydantic.BaseModel):
    string_value: str = "string"

    class Config:
        extra = "forbid"

        @classmethod
        def alias_generator(cls, string: str) -> str:
            return inflection.camelize(string, False)


class BooleanOperators(Parent):
    operator: Literal["or", "and", "xor", "nor", "not"]
    args: List[str] = Field(min_items=1)

    @root_validator
    def validate(cls, values: Dict[str, Any]) -> Dict[str, Any]:
        if values["operator"] == "not":
            if len(values["args"]) > 1:
                raise ValueError("'not' may not have more than 1 argument")
        else:
            if len(values["args"]) < 2:
                raise ValueError("other than 'not' require at least 2 arguments")
        return values


class ArithmeticOperators(Parent):
    operator: Literal["+", "-", "*", "/"]
    args: List[str] = Field(min_items=2)


class NonExistentOperators(Parent):
    operator: Literal["foo", "bar"]
    args: List[str] = Field(min_items=1)


class MainModel(BaseModel):
    submodel: Union[BooleanOperators, ArithmeticOperators, NonExistentOperators]

```

Expected:
```py
>>> MainModel(submodel={"operator":"+", "args":["helloWorld","testText"]})
MainModel(submodel=ArithmeticOperators(string_value='string', operator='+', args=['helloWorld', 'testText']))
```
Actual:
```py
>>> MainModel(submodel={"operator":"+", "args":["helloWorld","testText"]})
MainModel(submodel={'operator': '+', 'args': ['helloWorld', 'testText']})
```
### Current Workaround
Reversing the order of BooleanOperators and ArtithmeticOperators in the Union takes care of the issue based on prioritization on models. However, this is not documented to apply in the case of Literals, and it is confusing -- it would make sense that the first model would not match and it would progress to the second, so the order is assumed not to matter in this case.

---
```
$ python -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.7.3
            pydantic compiled: True
               python version: 3.8.6rc1 (tags/v3.8.6rc1:08bd63d, Sep  7 2020, 23:10:23) [MSC v.1927 64 bit (AMD64)]
```

nathanparekh:
## Change Summary
Adds a very short warning against overwriting BaseModel.validate() to the validators documentation, as @PrettyWood suggested would be helpful.

## Related issue number
Closes #2289 

## Checklist

### DIFF
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -5,6 +5,11 @@ Custom validation and complex relationships between objects can be achieved usin
 ```
 _(This script is complete, it should run "as is")_
 
+!!! warning
+  Validators should not overwrite `BaseModel.validate()`. Instead, give the validator function a unique name.
+  
+  This may be fixed in v2 - see [samuelcolvin/pydantic#1001](https://github.com/samuelcolvin/pydantic/issues/1001).
+
 A few things to note on validators:
 
 * validators are "class methods", so the first argument value they receive is the `UserModel` class, not an instance
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -6,11 +6,10 @@ Custom validation and complex relationships between objects can be achieved usin
 _(This script is complete, it should run "as is")_
 
 !!! warning
-  Validators should not overwrite `BaseModel.validate()`. Instead, give the validator function a unique name.
-  
-  This may be fixed in v2 - see [samuelcolvin/pydantic#1001](https://github.com/samuelcolvin/pydantic/issues/1001).
+    Validators should not overwrite `BaseModel.validate()`. Instead, give the validator function a unique name.
+    This may be fixed in v2 - see [samuelcolvin/pydantic#1001](https://github.com/samuelcolvin/pydantic/issues/1001).
 
-A few things to note on validators:
+A few other things to note on validators:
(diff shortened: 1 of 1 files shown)


nathanparekh:
Somehow managed to butcher a docs update. Will fix and resubmit later.

PR #2325, opened 2021-02-07
Properly retain types of Mapping subclasses

### LINKED ISSUE #1536: defaultdict converted to dict after __post_init__
# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: <redacted>/lib/python3.7/site-packages/pydantic
               python version: 3.7.3 | packaged by conda-forge | (default, Dec  6 2019, 08:36:57)  [Clang 9.0.0 (tags/RELEASE_900/final)]
```

If I run the following snippet:
```py
import collections
import dataclasses
from typing import DefaultDict

@dataclass
class X:
    f: DefaultDict[int, int] = dataclasses.field(init=False)

    def __post_init__(self):
        self.f = collections.defaultdict(int)
        self.f[2] += 1
        print(f"__post_init__: {self.f}")


x = X()
print(type(x.f))
print(x.f)
```

It prints the following:
```
__post_init__: defaultdict(<class 'int'>, {2: 1})
<class 'dict'>
{2: 1}
```

In other words, the `defaultdict` field is converted into a `dict` (while preserving the keys), even though the field is annotated as DefaultDict, which is surprising. Is this intended behavior?

### LINKED ISSUE #2323: Mapping types are coerced to dict
# Bug

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: C:\Users\ofek\AppData\Local\Programs\Python\Python38\Lib\site-packages\pydantic
               python version: 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 23:11:46) [MSC v.1916 64 bit (AMD64)]
```

I'm trying to use https://github.com/MagicStack/immutables

```py

from typing import Mapping

from immutables import Map


class Model(BaseModel):
    test: Mapping[str, int]

print(type(Model(test=Map(key=42)).test))
```

produces

```
<class 'dict'>
```

Is there a way to keep the type of encountered Mappings, or create a type representing `immutables.Map`?

I've been at this for hours reading docs/closed issues, trying various approaches, and now I'm simply at a loss here.

ofek:
## Change Summary

This change prevents Mapping subclasses from always being coerced to `dict`

## Related issue number

Resolves #2323 
closes #1536

## Checklist


ofek:
@PrettyWood I'm not sure how to fix the type checking errors

ofek:
@PrettyWood The current job failure is due to https://pypi.org/project/cryptography/3.4.1/ emitting a warning in the test suite of FastAPI

PrettyWood:
Yep can't do anything about that now but it should be handled rapidly on FastAPI side.
In the meantime you can add some tests to keep coverage at 100% :)

ofek:
Done!

PrettyWood (changes requested):
@ofek We should still check the set type and try to coerce when needed.
For example
```py
class M(BaseModel):
    x: Dict[str, int]
```
should coerce to `dict`. Right now there is no difference between `Dict` and `Mapping`...
I updated https://github.com/PrettyWood/pydantic/pull/70/files if you want

-----------------------------------------------------------------

Note that I think we should probably have a proper validator for `Dict` (at least) before `Mapping` to avoid any performance impact on normal dictionaries (like it's done for `Tuple` or `List`, which are treated before `Sequence`)

ofek:
Updated based on your branch

pydantic/fields.py
```diff
@@ -737,7 +738,25 @@ def _validate_mapping(
...
+            return self._get_mapping_value(v, result), None
+
+    def _get_mapping_value(self, original: T, converted: Dict[Any, Any]) -> Union[T, Dict[Any, Any]]:
+        target_type = get_origin(self.outer_type_)
+        original_type = type(original)
+
+        if is_mapping_type(target_type):
+            target_type = original_type
+
+        if target_type is dict:
```
PrettyWood, line 750:
```suggestion
        if target_type in {dict, Dict}:
```

PrettyWood: For python 3.6, `__origin__` is `typing.Dict`, which is just an alias for `dict`

pydantic/fields.py
```diff
@@ -737,7 +738,25 @@ def _validate_mapping(
...
+            return converted
+        elif target_type is defaultdict:
```
PrettyWood, line 752:
```suggestion
        elif target_type in {defaultdict, DefaultDict}:
```
same

pydantic/fields.py
```diff
@@ -737,7 +738,25 @@ def _validate_mapping(
...
+            return defaultdict(getattr(original, 'default_factory', None), **converted)
```
PrettyWood, line 753:
```suggestion
            return defaultdict(getattr(original, 'default_factory', None), converted)
```
and same under.
If we use `kwargs`, it won't work for `int` keys. But it does when passing a dict as `arg`.
Same under

PrettyWood (commented):
I still reckon we should at least have `SHAPE_DICT` (extra could be `SHAPE_DEFAULTDICT`, `SHAPE_COUNTER`...) before `SHAPE_MAPPING` to have no perf impact.
Or at least if we go with it we should have a benchmark to see the impact

ofek:
@PrettyWood Addressed. I'm not familiar with the concept of shapes here. Can you explain?

PrettyWood:
@ofek _pydantic_ is fast because a lot is done in the metaclass. You can have a look at `ModelField._type_analysis()`, which is run in `ModelMetaclass` when `ModelField.infer()` is called.
In `_type_analysis` a lot is done to set the right attributes to each ModelField (type, shape, validators...)
And the `shape` is used to call the right default validators to speed up the parsing

ofek:
I just tried that but many tests are failing now

ofek:
should I revert the last commit for now?

ofek:
I'm still trying but I don't understand what's wrong 🙁

ofek:
@PrettyWood Could you please take a look when you get a chance?

PrettyWood:
I'll have a look tonight or tomorrow 👍

ofek:
Thank you, feel free to push directly to my branch!

PrettyWood:
Hi @ofek
Had a quick look and got all tests working in local https://github.com/PrettyWood/pydantic/pull/70
I have some guests coming around soon but I think you should be able to work it out without a problem
I think you still need to add some tests to showcase a random mapping getting coerced into `dict` when `Dict[...]` is set.
Hope it helps

ofek:
@PrettyWood Done! Added that test too

tests/test_main.py
```diff
@@ -1472,3 +1473,71 @@ class Item(BaseModel):
...
     assert id(image_1) == id(item.images[0])
     assert id(image_2) == id(item.images[1])
+
+
+def test_mapping_retains_type_subclass():
+    class Map(dict):
+        pass
+
+    class Model(BaseModel):
+        field: Mapping[str, Mapping[str, int]]
```
samuelcolvin, line 1483: this is a very confusing name for a field.

tests/test_main.py
```diff
@@ -1472,3 +1473,71 @@ class Item(BaseModel):
 
     assert id(image_1) == id(item.images[0])
     assert id(image_2) == id(item.images[1])
+
+
+def test_mapping_retains_type_subclass():
+    class Map(dict):
```
samuelcolvin, line 1479:
```suggestion
    class CustomMap(dict):
```
might be a clearer name?

samuelcolvin (commented):
otherwise this is looking good.

ofek:
@samuelcolvin Thanks, addressed!

ofek:
Done this look alright now?

ofek:
@samuelcolvin Just let me know if I have to do anything else! FYI, after this is merged and released, every [Datadog Agent integration](https://github.com/DataDog/integrations-core) will exclusively use pydantic for user configuration 🙂

samuelcolvin:
This is great. Thank you.

> @samuelcolvin Just let me know if I have to do anything else! FYI, after this is merged and released, every [Datadog Agent integration](https://github.com/DataDog/integrations-core) will exclusively use pydantic for user configuration 🙂

I don't see pydantic as a dependency, what am I missing?

PrettyWood:
@samuelcolvin https://github.com/DataDog/integrations-core/pull/8675

ofek:
https://github.com/DataDog/integrations-core/pull/8675 is merged and I've started the syncing of every integration: https://github.com/DataDog/integrations-core/pulls 😄

ofek:
https://datadoghq.dev/integrations-core/meta/config-models/

PR #2336, opened 2021-02-09
Support discriminated union

### LINKED ISSUE #619: [Feature Request] Provide a discriminated union type (OpenAPI 3)

### LINKED ISSUE #3113: Validation of fields in tagged unions creates confusing errors
### Checks

# Bug

```
             pydantic version: 1.8.2
            pydantic compiled: False
                 install path: /Users/jonas.obrist/playground/pydantic/pydantic
               python version: 3.8.3 (default, Jul 17 2020, 11:57:37)  [Clang 10.0.0 ]
```

Tested with git checkout 5ccbdcb5904f35834300b01432a665c75dc02296.

See [this test in my branch](https://github.com/ojii/pydantic/blob/54176644ec5494b9dd705d60c88ca2fbd64f5c36/tests/test_errors.py#L470-L491).

```py
from typing import Literal, Union

from pydantic import BaseModel, conint


class A(BaseModel):
    tag: Literal["a"]


class B(BaseModel):
    tag: Literal["b"]
    x: conint(ge=0, le=1)


class Model(BaseModel):
    value: Union[A, B]

Model(value={"tag": "b", "x": 2}) # this will fail with "tag must be 'a'" and "x" must be LE 1
```

I would expect the error to be just "'x' needs to be LE 1", but the error also complains that "'tag' must be 'a'".

This is somewhat related to #619

PrettyWood:
## Change Summary
Add discriminated union support and support open api spec about it

## Related issue number
closes #619
closes #3113

## Checklist


levrik:
OpenAPI 3.x actually doesn't support the `const` option. An `enum` with a single option would be correct here.
JSON schemas support the `const` option.

PrettyWood:
@levrik It's another issue that was already opened regarding proper support of `Literal`.
I made a separate PR #2348 for this

levrik:
@PrettyWood That's great news!

tests/test_schema.py
```diff
@@ -2265,3 +2265,68 @@ class ModelModified(BaseModel):
...
+                ],
+                'discriminator': {
+                    'propertyName': 'pet_type',
+                    'mapping': {
+                        'cat': '#/definitions/Cat',
+                        'dog': '#/definitions/Dog',
+                        'lizard': '#/definitions/Lizard',
+                        'reptile': '#/definitions/Lizard',
+                    },
+                },
```
levrik, line 2303: `discriminator` is supported from OpenAPI 3.x on but that one also seems to use a different way of storing definitions. Seems that they were moved from `definitions` to `components/schema`.
Also in the end Pydantic seems to target JSON schema? Just wondering how this all fits 😅

amagee:
I noticed this does not work if the types I am unioning over are forward references.

Ie. this works:

```
from typing import Literal, Union

class Group(BaseModel):
    type: Literal["Group"]

class Layer(BaseModel):
    type: Literal["Layer"]

class GroupOrLayer(BaseModel):
    __root__: Union["Group", "Layer"] = Field(..., discriminator="type")
```

but if I move the `Group` and `Layer` classes beneath `GroupOrLayer`, I get this:

```
  File "mcve.py", line 4, in <module>
    class GroupOrLayer(BaseModel):
  File "/path/pydantic/main.py", line 291, in __new__
  File "/path/pydantic/fields.py", line 390, in infer
    return cls(
  File "/path/pydantic/fields.py", line 325, in __init__
    self.prepare()
  File "/path/pydantic/fields.py", line 430, in prepare
    self._type_analysis()
  File "/path/pydantic/fields.py", line 539, in _type_analysis
    t_discriminator_type = t.__fields__[discriminator_key].outer_type_
AttributeError: 'ForwardRef' object has no attribute '__fields__'
```

(Just on initialisation of the class; no need to actually do anything with it.)

tests/test_schema.py
PrettyWood: My main concern was to speed up the validation and have a nice and clean error when not valid.
The JSON schema support is messy and I tried to implement the basic one for FastAPI more than myself.
Currently this is the result

| discriminator  | code  | ReDoc  |
|---|---|---|
| without  | without_discriminator_code  | without_discriminator_redoc
  |
| with  | with_discriminator_code  | with_discriminator_redoc  |

I'm not sure I want to put more effort on it without clear expectations.
If the current behaviour is wrong and someone can produce a full example of FastAPI  code with expected ReDoc, I will try to make it work 👍

PrettyWood:
Thanks @amagee. I had to rewrite a bit the logic to support properly `ForwardRef`.
Feedback welcome

vdwees:
I gave this branch a try. Definitely a step in the right direction. In the script below I demonstrate a common situation I encounter where there is more than one discriminating field. I am wondering if you think this is something worth supporting in a discriminated union- would it even be compatible with OpenAPI?


```py
from typing import Literal, Union


class DomainA(BaseModel):
    domain: Literal["A"]


class DomainB(BaseModel):
    domain: Literal["B"]


class FooDomainA(DomainA):
    identifier: Literal["foo"]


class BarDomainA(DomainA):
    identifier: Literal["bar"]


class FooDomainB(DomainB):
    identifier: Literal["foo"]


class BarDomainB(DomainB):
    identifier: Literal["bar"]


class MyClass(BaseModel):
    __root__: Union[FooDomainA, BarDomainA, FooDomainB, BarDomainB] = Field(
        ..., discriminator="domain"
    )


for domain in ("A", "B"):
    for identifier in ("foo", "bar"):
        print(domain, identifier)
        try:
            print(MyClass.parse_obj({"domain": domain, "identifier": identifier}))
        except ValueError as e:
            print(e)
        print()
```
Output:
```
A foo
1 validation error for MyClass
__root__ -> identifier
  unexpected value; permitted: 'bar' (type=value_error.const; given=foo; permitted=('bar',))

A bar
__root__=BarDomainA(domain='A', identifier='bar')

B foo
1 validation error for MyClass
__root__ -> identifier
  unexpected value; permitted: 'bar' (type=value_error.const; given=foo; permitted=('bar',))

B bar
__root__=BarDomainB(domain='B', identifier='bar')
```

PrettyWood:
@vdwees What would be the configuration? an array of discriminators? Whatever we decide I think it can wait for another PR

vdwees:
@PrettyWood I looked into it a bit more, and based on my reading of the openapi schema, I think the discriminator is only a single field. So as much as I would like the discriminator kwarg to support a sequence in addition to a single value, I guess it wouldn't make sense to support that here :cry: 

But I was thinking- even if it is only a single value, it might still be reasonable/possible to support similar functionality if the discriminator was nestable, e.g something like below:

```py
class FooDomainA(BaseModel):
    domain: Literal["A"]
    identifier: Literal["foo"]

class BarDomainA(BaseModel):
    domain: Literal["A"]
    identifier: Literal["bar"]

class FooDomainB(BaseModel):
    domain: Literal["B"]
    identifier: Literal["foo"]

class BarDomainB(BaseModel):
    domain: Literal["B"]
    identifier: Literal["bar"]

class DomainA(BaseModel):
    __root__: Union[FooDomainA, BarDomainA] = Field(..., discriminator="identifier")

class DomainB(BaseModel):
    __root__: Union[FooDomainB, BarDomainB] = Field(..., discriminator="identifier")

class PolymorphClass(BaseModel):
    __root__: Union[DomainA, DomainB] = Field(..., discriminator="domain")
```

vdwees:
Actually, it seems valid from the OpenAPI perspective to do it this way, based on the specs I found here

<details><summary>Example OpenAPI schemas</summary>
<p>

```yaml
components:
  schemas:
    PolymorphClass:
      title: PolymorphClass
      discriminator:
        propertyName: domain
        mapping:
          A: '#/components/schemas/DomainA'
          B: '#/definitions/DomainB'
      anyOf:
        - $ref: '#/components/schemas/DomainA'
        - $ref: '#/components/schemas/DomainB'
    DomainA:
      title: DomainA
      discriminator:
        propertyName: identifier
        mapping:
          foo: '#/components/schemas/FooDomainA'
          bar: '#/components/schemas/BarDomainA'
      anyOf:
        - $ref: '#/components/schemas/FooDomainA'
        - $ref: '#/components/schemas/BarDomainA'
    DomainB:
      title: DomainB
      discriminator:
        propertyName: identifier
        mapping:
          foo: '#/components/schemas/FooDomainB'
          bar: '#/components/schemas/BarDomainB'
      anyOf:
        - $ref: '#/components/schemas/FooDomainB'
        - $ref: '#/components/schemas/BarDomainB'
    FooDomainA:
      title: FooDomainA
      type: object
      properties:
        domain:
          title: Domain
          enum:
            - A
          type: string
        identifier:
          title: Identifier
          enum:
            - foo
          type: string
      required:
        - domain
        - identifier
    BarDomainA:
      title: BarDomainA
      type: object
      properties:
        domain:
          title: Domain
          enum:
            - A
          type: string
        identifier:
          title: Identifier
          enum:
            - bar
          type: string
      required:
        - domain
        - identifier
    FooDomainB:
      title: FooDomainB
      type: object
      properties:
        domain:
          title: Domain
          enum:
            - B
          type: string
        identifier:
          title: Identifier
          enum:
            - foo
          type: string
      required:
        - domain
        - identifier
    BarDomainB:
      title: BarDomainB
      type: object
      properties:
        domain:
          title: Domain
          enum:
            - B
          type: string
        identifier:
          title: Identifier
          enum:
            - bar
          type: string
      required:
        - domain
        - identifier

```

image

</p>
</details>


<details><summary>Example PolymorphClass.schema()</summary>
<p>

```json
{
  "title": "PolymorphClass",
  "discriminator": {
    "propertyName": "domain",
    "mapping": {
      "A": "#/definitions/DomainA",
      "B": "#/definitions/DomainB"
    }
  },
  "anyOf": [
    {
      "$ref": "#/definitions/DomainA"
    },
    {
      "$ref": "#/definitions/DomainB"
    }
  ],
  "definitions": {
    "DomainA": {
      "title": "DomainA",
      "discriminator": {
        "propertyName": "identifier",
        "mapping": {
          "foo": "#/definitions/FooDomainA",
          "bar": "#/definitions/BarDomainA"
        }
      },
      "anyOf": [
        {
          "$ref": "#/definitions/FooDomainA"
        },
        {
          "$ref": "#/definitions/BarDomainA"
        }
      ]
    },
    "DomainB": {
      "title": "DomainB",
      "discriminator": {
        "propertyName": "identifier",
        "mapping": {
          "foo": "#/definitions/FooDomainB",
          "bar": "#/definitions/BarDomainB"
        }
      },
      "anyOf": [
        {
          "$ref": "#/definitions/FooDomainB"
        },
        {
          "$ref": "#/definitions/BarDomainB"
        }
      ]
    },
    "FooDomainA": {
      "title": "FooDomainA",
      "type": "object",
      "properties": {
        "domain": {
          "title": "Domain",
          "const": "A",
          "type": "string"
        },
        "identifier": {
          "title": "Identifier",
          "const": "foo",
          "type": "string"
        }
      },
      "required": ["domain", "identifier"]
    },
    "BarDomainA": {
      "title": "BarDomainA",
      "type": "object",
      "properties": {
        "domain": {
          "title": "Domain",
          "const": "A",
          "type": "string"
        },
        "identifier": {
          "title": "Identifier",
          "const": "bar",
          "type": "string"
        }
      },
      "required": ["domain", "identifier"]
    },
    "FooDomainB": {
      "title": "FooDomainB",
      "type": "object",
      "properties": {
        "domain": {
          "title": "Domain",
          "const": "B",
          "type": "string"
        },
        "identifier": {
          "title": "Identifier",
          "const": "foo",
          "type": "string"
        }
      },
      "required": ["domain", "identifier"]
    },
    "BarDomainB": {
      "title": "BarDomainB",
      "type": "object",
      "properties": {
        "domain": {
          "title": "Domain",
          "const": "B",
          "type": "string"
        },
        "identifier": {
          "title": "Identifier",
          "const": "bar",
          "type": "string"
        }
      },
      "required": ["domain", "identifier"]
    }
  }
}

```
</p>
</details>

Maybe the pydantic API could support a tuple, like this?

```python
class PolymorphClass(BaseModel):
    __root__: Union[DomainA, DomainB] = Field(
        ..., discriminator=("domain", {"A": DomainA, "B": DomainB})
    )
```

samuelcolvin:
Just to say thank you so much for working on this.

Giving this a proper review is weighing heavy on my conscience, but I can't devote the time before v1.8.

PrettyWood:
@samuelcolvin I want to improve it further to support nested discriminated unions. Better wait for v1.9 anyway ;)

cybojenix (commented):
This is looking good.
I've put in some cleanup suggestions to reduce some duplication, and simplify calling `update_forward_refs` (tested locally).

I think an assumption has been made that will cause some headaches though.
It's possible to have duplicated discriminator values.

```py
class Spaniel(BaseModel):
  pet_type: Literal['dog']
  breed: Literal['spaniel']

class JackRussel(BaseModel):
  pet_type: Literal['dog']
  breed: Literal['Jack Russel']

class Cat(BaseModel):
  pet_type: Literal['cat']

class Model(BaseModel):
  pet: Union[Spaniel, JackRussel, Cat] = Field(discriminator='pet_type')
```

I don't think this is an edge case either, we have something to this effect running in production.

In this case, I think the code should be defensive:
- Don't generate the `mapping` in the schema (or exclude mappings that are duplicates)
- have `sub_field_mappings` be defined as `Dict[str, List[ModelField]]`, so it can be used to narrow down validation

cybojenix:
While looking into nested discriminators (as a true fix to the above), You end up with a schema akin to this:

```yml
definitions:
  Pet:
    anyOf:
      - $ref: '#/definitions/Cat'
      - $ref: '#/definitions/Dog'
   discriminator:
      propertyName: pet_type
      mapping:
        cat: '#/definitions/Cat'
        dog: '#/definitions/Dog'
  Cat:
    properties:
      pet_type:
        type: string
        const: cat
  Dog:
    anyOf:
      - $ref: '#/definitions/Spaniel'
      - $ref: '#/definitions/JackRussel'
    discriminator:
      propertyName: breed
      mapping:
        spaniel: '#/definitions/Spaniel'
        jack-russel: '#/definitions/JackRussel'
  Spaniel:
    properties:
      pet_type:
        type: string
        const: dog
      breed:
        type: string
        const: spaniel
      s:
        type: string
  JackRussel:
    properties:
      pet_type:
        type: string
        const: dog
      breed:
        type: string
        const: jack-russel
      j:
        type: string
```

Note how we have `Dog` as an intermediary object, wrapping our breeds.

To express this, we need to define Dog as:
```py
class Dog(BaseModel):
  __root__ = Union[Spaniel, JackRussel] = Field(..., discriminator='breed')

class Model(BaseModel):
        pet: Union[Cat, Dog] = Field(..., discriminator='pet_type')
        number: int
```

However this isn't supported by the current code, raising a KeyError:
`KeyError: "Model 'Dog' needs a discriminator field for key 'pet_type'"`

--- 
An example of using in Redoc: https://redocly.github.io/redoc/?url=https://gist.githubusercontent.com/cybojenix/5258ebe39cd100e99be76e5175f9b41f/raw/6b1708c71cad50ef00fda6bbf1d35ea439d6e2fa/nested-discriminator.yml

This actually [breaks](https://github.com/Redocly/redoc/issues/210) the interface when selecting `dog` as the `pet_type`, though it looks like others are looking for similar functionality.

PrettyWood:
Thanks a lot @cybojenix for the review 🙏  It was still a work in progress but glad to have some constructive feedback on the changes we can do. I'll have a look at it probably next week!

pydantic/fields.py
```diff
@@ -597,6 +623,46 @@ def _type_analysis(self) -> None:  # noqa: C901 (ignore complexity)
...
+                raise TypeError(f'Field {discriminator_key!r} of model {t.__name__!r} needs to be a `Literal`')
+
+            for discriminator_value in all_literal_values(t_discriminator_type):
+                sub_fields_mapping[discriminator_value] = sub_field
+
+            sub_fields.append(sub_field)
+
+        self.sub_fields = sub_fields
+        self.discriminated_union_config.has_forward_refs = False
+        self.discriminated_union_config.sub_fields_mapping = sub_fields_mapping
```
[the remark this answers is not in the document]
PrettyWood: Yes I thought about it but did this to be more explicit

PrettyWood:
Thanks @cybojenix for the example.
I did the changes and started to work on the nested unions.
```py
from typing import Literal, Union

class Cat(BaseModel):
    pet_type: Literal["cat"]


class Spaniel(BaseModel):
    pet_type: Literal["dog"]
    breed: Literal["spaniel"]
    s: str


class JackRussel(BaseModel):
    pet_type: Literal["dog"]
    breed: Literal["jack-russel"]
    j: str


class Dog(BaseModel):
    __root__: Union[Spaniel, JackRussel] = Field(..., discriminator="breed")


class Pet(BaseModel):
    __root__: Union[Cat, Dog] = Field(..., discriminator="pet_type")


assert Pet.schema() == {
  "title": "Pet",
  "discriminator": {
    "propertyName": "pet_type",
    "mapping": {
      "cat": "#/definitions/Cat",
      "dog": "#/definitions/Dog"
    }
  },
  "anyOf": [
    {
      "$ref": "#/definitions/Cat"
    },
    {
      "$ref": "#/definitions/Dog"
    }
  ],
  "definitions": {
    "Cat": {
      "title": "Cat",
      "type": "object",
      "properties": {
        "pet_type": {
          "title": "Pet Type",
          "enum": [
            "cat"
          ],
          "type": "string"
        }
      },
      "required": [
        "pet_type"
      ]
    },
    "Spaniel": {
      "title": "Spaniel",
      "type": "object",
      "properties": {
        "pet_type": {
          "title": "Pet Type",
          "enum": [
            "dog"
          ],
          "type": "string"
        },
        "breed": {
          "title": "Breed",
          "enum": [
            "spaniel"
          ],
          "type": "string"
        },
        "s": {
          "title": "S",
          "type": "string"
        }
      },
      "required": [
        "pet_type",
        "breed",
        "s"
      ]
    },
    "JackRussel": {
      "title": "JackRussel",
      "type": "object",
      "properties": {
        "pet_type": {
          "title": "Pet Type",
          "enum": [
            "dog"
          ],
          "type": "string"
        },
        "breed": {
          "title": "Breed",
          "enum": [
            "jack-russel"
          ],
          "type": "string"
        },
        "j": {
          "title": "J",
          "type": "string"
        }
      },
      "required": [
        "pet_type",
        "breed",
        "j"
      ]
    },
    "Dog": {
      "title": "Dog",
      "discriminator": {
        "propertyName": "breed",
        "mapping": {
          "spaniel": "#/definitions/Spaniel",
          "jack-russel": "#/definitions/JackRussel"
        }
      },
      "anyOf": [
        {
          "$ref": "#/definitions/Spaniel"
        },
        {
          "$ref": "#/definitions/JackRussel"
        }
      ]
    }
  }
}
```

KevOrr:
With this PR, is there any way to express something like the following?
```py
class Dog(BaseModel):
    ...

class Cat(BaseModel):
    ...

class Animal(BaseModel):
    tag: Literal['Dog', 'Cat']
    contents: {'Dog': Dog, 'Cat': Cat}[tag]  # pseudocode
```
No worries if this is out of scope (I can think of a couple workarounds anyway) but curious if this is something that will be supported.

PrettyWood:
Hi @KevOrr 
I'm afraid no. But with https://github.com/samuelcolvin/pydantic/pull/2231 it should be easy to add the tag without adding it

```py
from typing import Annotated, Literal, Union

from pydantic import BaseModel, Field, parse_obj_as


class Cat(BaseModel):
    type: Literal['cat'] = Field('cat', exclude=True)
    cat_name: str


class Dog(BaseModel):
    type: Literal['dog'] = Field('dog', exclude=True)
    dog_name: str


Animal = Annotated[Union[Cat, Dog], Field(discriminator='type')]

assert Cat(cat_name='Titi').dict() == {'cat_name': 'Titi'}
assert parse_obj_as(Animal, {'type': 'cat', 'cat_name': 'Titi'}).dict() == {'cat_name': 'Titi'}
```

foarsitter:
Thanks for this addition! Using this branch already with success. Thanks again!

pydantic/fields.py
```diff
@@ -895,6 +947,36 @@ def _validate_singleton(
...
+                    assert cls is not None
+                    raise ConfigError(
+                        f'field "{self.name}" not yet prepared so type is still a ForwardRef, '
+                        f'you might need to call {cls.__name__}.update_forward_refs().'
+                    )
+
+                discriminator_key = self.discriminated_union_config.discriminator_key
+
+                try:
+                    discriminator_value = v[discriminator_key]
```
kbakk, line 962: Not sure this is intentional - this assumes `v` is a dict, if `v` is a class inherited from `pydantic.BaseModel`, it will fail:

```python
>>> import pydantic, typing
...
... class A(pydantic.BaseModel):
...     l: typing.Literal["a"]
...
... class B(pydantic.BaseModel):
...     l: typing.Literal["b"]
...
... class Top(pydantic.BaseModel):
...     sub: typing.Union[A, B] = pydantic.Field(..., discriminator="l")
...
... t = Top(sub=A(l="a"))
  File "<stdin>", line 12, in <module>
  File "/Users/krisb/Library/Caches/pypoetry/virtualenvs/project-py3.8/lib/python3.8/site-packages/pydantic/main.py", line 400, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for Top
sub
  Discriminator 'l' is missing in value (type=value_error)

1 validation error for Top
sub
  Discriminator 'l' is missing in value (type=value_error)
>>> t = Top(**{"sub": {"l": "a"}})
>>> t
Top(sub=A(l='a'))

>>> class TopWithoutDiscriminator(pydantic.BaseModel):
...     sub: typing.Union[A, B]
... t = TopWithoutDiscriminator(sub=A(l="a"))
>>> t
TopWithoutDiscriminator(sub=A(l='a'))
```

PrettyWood: Thanks a lot @kbakk! It was definitely not intentional ;)

pydantic/fields.py
```diff
@@ -917,6 +969,42 @@ def _validate_singleton(
...
+
+                try:
+                    sub_field = self.discriminated_union_config.sub_fields_mapping[discriminator_value]
+                except KeyError:
+                    allowed_values = ', '.join(map(repr, self.discriminated_union_config.sub_fields_mapping.keys()))
+                    msg_err = (
+                        f'No match for discriminator {discriminator_key!r} and value {discriminator_value!r} '
+                        f'(allowed values: {allowed_values})'
+                    )
+                    return v, ErrorWrapper(ValueError(msg_err), loc)
```
samuelcolvin, line 1002: As above, can we use the existing error infrastructure?

pydantic/fields.py
```diff
@@ -1029,3 +1117,37 @@ class DeferredType:
     """
     Used to postpone field preparation, while creating recursive generic models.
     """
+
+
+def _get_discriminator_values(tp: Any, discriminator_key: str) -> Tuple[str, ...]:
```
samuelcolvin, line 1122: this file is already very long, could we move this logic somewhere else like `utils.py`?

pydantic/tools.py
```diff
@@ -77,3 +80,11 @@ def parse_raw_as(
         json_loads=json_loads,
     )
     return parse_obj_as(type_, obj, type_name=type_name)
+
+
+def schema(type_: Any, *, title: Optional[NameFactory] = None, **schema_kwargs: Any) -> 'DictStrAny':
```
samuelcolvin, line 85: These methods looks useful, but they definitely deserve docstrings, and probably documentation too.

If it's too much for this PR, could we create an issue to remind us to document this in future?

pydantic/typing.py
```diff
@@ -397,3 +405,12 @@ def get_class(type_: Type[Any]) -> Union[None, bool, Type[Any]]:
     except (AttributeError, TypeError):
         pass
     return None
+
+
+def get_sub_types(tp: Any) -> List[Any]:
+    if get_origin(tp) is Annotated:
+        return get_sub_types(get_args(tp)[0])
+    elif get_origin(tp) is Union:
+        return [x for t in get_args(tp) for x in get_sub_types(t)]
```
samuelcolvin, line 414:
```suggestion
    origin = get_origin(tp)
    if origin is Annotated:
        return get_sub_types(get_args(tp)[0])
    elif origin is Union:
        return [x for t in get_args(tp) for x in get_sub_types(t)]
```

Also, please add a docstring.

tests/test_main.py
```diff
@@ -1994,3 +1994,213 @@ class Model(BaseModel, extra='allow'):
 
             class Config:
                 extra = 'forbid'
+
+
+def test_discriminated_union_defined_discriminator():
+    with pytest.raises(KeyError, match="Model 'Cat' needs a discriminator field for key 'pet_type'"):
```
samuelcolvin, line 2000: can you put this around just the code which will raise the error, and below.

samuelcolvin (commented):
This is looking really cool. I'm really looking forward to using it.

These are mostly small style tweaks and moving code around, let me know if anything is unclear.

PrettyWood:
@samuelcolvin Thanks for the review! I added 7 commits, which should be easy to read one by one. Looking forward for your feedback.
Doc added [here](https://smokeshow.helpmanual.io/5t5l1863290s1v70192g/usage/schema/#getting-schema-of-a-specified-type)

tests/test_main.py
```diff
@@ -1994,3 +1994,249 @@ class Model(BaseModel, extra='allow'):
...
+            'loc': ('pet', 'Union[BlackCat, WhiteCat]', 'WhiteCat', 'white_infos'),
+            'msg': 'field required',
+            'type': 'value_error.missing',
+        }
+    ]
+    m = Model.parse_obj({'pet': {'pet_type': 'cat', 'color': 'white', 'white_infos': 'pika'}, 'number': 2})
+    assert isinstance(m.pet, WhiteCat)
+
+
+def test_discrimated_union_basemodel_instance_value():
```
daviskirk, line 2231: First of all, I'm really looking forward to using this!  
Perhaps it would be good to add a few tests marking off the boundaries of the feature more, some of them are just for clarification of the feature, but might help catch breaks in the future, for example I assume non-string literals, will work fine I assume?

```
def test_discriminated_union_int()
    class A(BaseModel):
        l: Literal[1]

    class B(BaseModel):
        l: Literal[2]

    class Top(BaseModel):
        sub: Union[A, B] = Field(..., discriminator='l')

def test_discriminated_union_enum():
    from enum import Enum 
    
    class EnumValue(Enum):
        a = 1
        b = 2

    class A(BaseModel):
        l: Literal[EnumValue.a]

    class B(BaseModel):
        l: Literal[EnumValue.b]

    class Top(BaseModel):
        sub: Union[A, B] = Field(..., discriminator='l')

```

PrettyWood: Thanks for the feedback! I added your tests in 3be7c09

PrettyWood:
Resolved conflicts with master. Please review whenever you can @samuelcolvin. Sorry for the fat PR 😅

pydantic/fields.py
```diff
@@ -563,16 +556,21 @@ def _type_analysis(self) -> None:  # noqa: C901 (ignore complexity)
...
         origin = get_origin(self.type_)
+
+        if origin is Annotated:
+            self.type_ = get_args(self.type_)[0]
+            self._type_analysis()
+            return
+
+        if self.discriminator_key is not None and origin is not Union:
+            raise TypeError('`discriminator` can only be used with `Union` type')
+
```
foarsitter, line 574: I was using this with a List[Union[]] but obviously I'm running here in this error. Is list union discriminator in scope of this pull-request? The version before 13 april was working as following: 

```
devices: List[
        Union[
            RelaySwitchSchema,
            DBRelaySwitchSchema,
            AM2302Schema,
            BME280Schema,
            SinusSchema,
        ]
    ] = Field(..., discriminator="type")
```

PrettyWood: I'm so glad when a feature is used beforehand! Allows to have feedback and ensure it works as expected thanks!
Indeed List of Union should be supported but more like `List[MyCustomUnion]` because `discriminator` needs to be added on the `Union` type, not the whole one.
Can you try with
```py
MySchemas = Annotated[
    Union[
        RelaySwitchSchema,
        DBRelaySwitchSchema,
        AM2302Schema,
        BME280Schema,
        SinusSchema,
    ],
    Field(discriminator="type")
]

class MyModel(BaseModel):
    devices: List[MySchemas]
```
?

foarsitter: Thanks for your quick reply and the effort you are putting in to this! Your purposed usage of Annotated works like a charm.

LyleScott: I'm also in need of exactly this feature, so thanks to the author(s)!

I have a schema setup like the above (list of union), but I have`from_orm` enabled on all models. I receive a traceback when serializing a database model:

```
  File "pydantic/fields.py", line 804, in validate
    v, errors = self._validate_singleton(v, values, loc, cls)
  File "pydantic/fields.py", line 990, in _validate_singleton
    return self._apply_validators(v, values, loc, cls, self.validators)
  File "pydantic/fields.py", line 1034, in _apply_validators
    v = validator(cls, v, values, self, self.model_config)
  File "pydantic/class_validators.py", line 310, in <lambda>
    return lambda cls, v, values, field, config: validator(v)
  File "pydantic/main.py", line 747, in validate
    return cls.from_orm(value)
  File "pydantic/main.py", line 639, in from_orm
    values, fields_set, validation_error = validate_model(cls, obj)
  File "pydantic/main.py", line 1054, in validate_model
    v_, errors_ = field.validate(value, values, loc=field.alias, cls=cls_)
  File "pydantic/fields.py", line 815, in validate
    v, errors = self._validate_sequence_like(v, values, loc, cls)
  File "pydantic/fields.py", line 848, in _validate_sequence_like
    r, ee = self._validate_singleton(v_, values, v_loc, cls)
  File "pydantic/fields.py", line 983, in _validate_singleton
    value, error = field.validate(v, values, loc=loc, cls=cls)
  File "pydantic/fields.py", line 804, in validate
    v, errors = self._validate_singleton(v, values, loc, cls)
  File "pydantic/fields.py", line 990, in _validate_singleton
    return self._apply_validators(v, values, loc, cls, self.validators)
  File "pydantic/fields.py", line 1034, in _apply_validators
    v = validator(cls, v, values, self, self.model_config)
  File "pydantic/class_validators.py", line 310, in <lambda>
    return lambda cls, v, values, field, config: validator(v)
  File "pydantic/main.py", line 747, in validate
    return cls.from_orm(value)
  File "pydantic/main.py", line 639, in from_orm
    values, fields_set, validation_error = validate_model(cls, obj)
  File "pydantic/main.py", line 1054, in validate_model
    v_, errors_ = field.validate(value, values, loc=field.alias, cls=cls_)
  File "pydantic/fields.py", line 815, in validate
    v, errors = self._validate_sequence_like(v, values, loc, cls)
  File "pydantic/fields.py", line 848, in _validate_sequence_like
    r, ee = self._validate_singleton(v_, values, v_loc, cls)
  File "pydantic/fields.py", line 983, in _validate_singleton
    value, error = field.validate(v, values, loc=loc, cls=cls)
  File "pydantic/fields.py", line 804, in validate
    v, errors = self._validate_singleton(v, values, loc, cls)
  File "pydantic/fields.py", line 979, in _validate_singleton
    return self._validate_discriminated_union(v, values, loc, cls)
  File "pydantic/fields.py", line 999, in _validate_discriminated_union
    raise ConfigError(
pydantic.errors.ConfigError: field "_actions" not yet prepared so type is still a ForwardRef, you might need to call EscalationSchema.update_forward_refs().
```

If I actually do the recommendation, I do not receive the traceback and everything is successful.

```python
    @classmethod
    def validate(cls, value):
        cls.update_forward_refs()
        return super().validate(value)
```

Is this expected behavior? I wasn't sure if this was a `from_orm` related bug. I'd be glad to provide more info when I have some time available.

PrettyWood: I can't reproduce. Could you please provide a full reproduction code or even better a test case to show the error? Thanks

foarsitter: Actually, @PrettyWood you have already tested this behavior so it seems as intended behavior to me: https://github.com/samuelcolvin/pydantic/blob/aefcdf1b73a43b43c9b8825aeb3f950ad710300c/tests/test_forward_ref.py#L524

mr-bjerre:
Really great work @PrettyWood 

Do you know if this this planned for next release?

levrik:
@samuelcolvin Can this please get a soon review? I'm currently having to deal once again with a list of 622 validation errors from Pydantic due to this missing. It's not fun at all to track down the actual error like this.

foarsitter:
@levrik no need to wait for a release, just install this repo. 

```
pip install -e git://github.com/PrettyWood/pydantic.git@f/discriminated-union#egg=pydantic
```

levrik:
@foarsitter Sadly that's not really an option in a corporate environment for deploying to production.

pydantic/fields.py
thinkerbot: I ran into this and might have code you can reproduce.  As above the basic error is:

```
# ...
File "/app/src/pydantic/pydantic/fields.py", line 979, in _validate_singleton
  return self._validate_discriminated_union(v, values, loc, cls)
File "/app/src/pydantic/pydantic/fields.py", line 999, in _validate_discriminated_union
  raise ConfigError(
pydantic.errors.ConfigError: field "_items" not yet prepared so type is still a ForwardRef, you might need to call ItemList.update_forward_refs().
```

But the solution is different depending on whether the code is used within a FastAPI app or  in a script. When in FastAPI the solution from above is needed:

```python
@classmethod
def validate(cls, value):
    cls.update_forward_refs()
    return super().validate(value)
```

But when in a script the code seems to need a direct, inline call to `ItemList.update_forward_refs()`. Neither solves the error for the other use case. I have no idea what is going on but hopefully you do!

### Test Case 

With these files and `fastapi==0.67.0`:

```
example.json
example.py
main.py
lib/source.py
```

example.json

```json
{
  "type": "List",
  "items": [
    {
      "type": "A"
    },
    {
      "type": "List",
      "items": [
        {
          "type": "B"
        }
      ]
    }
  ]
}
```

example.py (for the script example)

```json
from lib.source import ItemList
import json

with open('example.json') as f:
    obj = json.loads(f.read())
i = ItemList.parse_obj(obj)
print(i)
```

main.py (for the FastAPI example)

```json
from fastapi import FastAPI
from lib.source import ItemList
app = FastAPI()

@app.post("/data", response_model=ItemList)
def data(input: ItemList):
    return input
```

lib/source.py (used in both examples)

```
from typing import List, Literal, Union
from pydantic import BaseModel,Field

class ItemA(BaseModel):
    type: Literal["A"]

class ItemB(BaseModel):
    type: Literal["B"]

Nodes = Annotated[Union["ItemList", ItemA, ItemB], Field(discriminator='type')]

class ItemList(BaseModel):
    type: Literal["List"]
    items: List[Nodes]

    # comment out to make curl to the fastapi app fail
    @classmethod
    def validate(cls, value):
        cls.update_forward_refs()
        return super().validate(value)

# comment out to make example.py fails
ItemList.update_forward_refs()
```

Then run like:

```shell
# script example
python3 example.py

# FastAPI example
uvicorn main:app --reload --host 0.0.0.0 
curl -H "content-type: application/json" -d @example.json -s http://localhost:8000/data
```

I find both ways of calling `update_forward_refs` are needed. Try commenting out the sections one at a time, repeating the script/curl commands and you should see each section solves it for one but not the other. Something is different about how the code gets used in each case.

Thanks for this feature btw... it's very helpful for my use case and I hope this gets merged soon.

phemmer:
Ran across an edge case while using this. Basically I'm trying to dynamically build the type annotation for the available classes in the discriminated union. To do this the annotation in the referencing object is a forward reference to a function which returns the union annotation (in reality it's a function that collects all subclasses). The problem with this is that it seems various properties of the field are not updated after resolving the forward ref, including the `discriminator_key`.

For example:
```python
class A(BaseModel):
    obj:"obj_annotation()"

class B(BaseModel):
    type: Literal['B']

class C(BaseModel):
    type: Literal['C']

def obj_annotation():
    return Annotated[Union[B,C], Field(discriminator='type')]

A.update_forward_refs()
obj = A.__fields__['obj']
print("%s discriminator=%s"%(obj, obj.discriminator_key))
```
```
name='obj' type=Union[B, C] required=True discriminator=None
```

This is an issue with the forward ref, and everything works fine (using a function to return an annotation) without it. For example by moving `class A` down:
```python
class B(BaseModel):
    type: Literal['B']

class C(BaseModel):
    type: Literal['C']

def obj_annotation():
    return Annotated[Union[B,C], Field(discriminator='type')]

class A(BaseModel):
    obj:obj_annotation()

obj = A.__fields__['obj']
print("%s discriminator=%s"%(obj, obj.discriminator_key))
```
```
name='obj' type=Union[B, C] required=True discriminator=type
```
We see that the discriminator is set properly.

Now as mentioned, there are other field attributes that don't get updated because of the forward ref (for example, `alias`). So this doesn't seem to be something specific to the new discriminator code. However I can't imagine many other use cases for a dynamic annotation other than this one. So I wanted to drop it here and see if this is:
a) Something we even want to address at all
b) Something to address via a discriminator-specific implementation
c) Or address as a more general purpose solution

docs/examples/schema_ad_hoc.py
```diff
@@ -0,0 +1,20 @@
...
+    pet_type: Literal['cat']
+    cat_name: str
+
+
+class Dog(BaseModel):
+    pet_type: Literal['dog']
+    dog_name: str
+
+
+Pet = Annotated[Union[Cat, Dog], Field(discriminator='pet_type')]
```
Dreamsorcerer, line 18: Why does this need an explicit `discriminator`?

Mypy doesn't need this, so can't we do something similar?

I imagine this as just a `Union[Cat, Dog]`. When you pass something to it, it will try to validate it as `Cat`, if `pet_type != 'cat'` it obviously fails and then tries to validate it as `Dog`. I'm not sure why an explicit discriminator is needed...

vdwees: How is the usage you are imagining different from how things currently work without this discriminator feature?

PrettyWood: The whole point of discrimination is to try to validate only on matching types hence gaining time on validation and having more explicit error.
Feel free to use plain Union if you want to fallback on other union classes

Dreamsorcerer: > How is the usage you are imagining different from how things currently work without this discriminator feature?

I think the problem is that I can't use `Pet = Union[Cat, Dog]`, as this is not a `BaseModel`. Maybe this has changed in the new release?

For context, I was sent to this PR in relation to this question: https://github.com/samuelcolvin/pydantic/discussions/2821

wholmen:
This is great work! I have to mirror the sentiment further up in the thread. Is there any plans of releasing this into the official pydantic package? And in that case, when? I would rather not use these kinds of branches in a corporate production environment.

docs/examples/schema_ad_hoc.py
PrettyWood: Why the need of `TypedDict` ?

Dreamsorcerer: Not sure what you mean? My example just demonstrated how I would achieve tagged unions in normal Python code (with Mypy checking), without Pydantic. I don't understand how to use tagged unions with Pydantic models, that was my question.

vdwees: @Dreamsorcerer I've replied to your question on the issue you linked

adriangb:
+1 would love to see this released; this is a pattern I see all the time

spolloni:
+1

AutumnSun1996:
There are some issues when using discriminator with alias.
A minimal example:
```python
from typing import TypeVar, Union
from typing_extensions import Annotated, Literal
from pydantic.fields import Field

T = TypeVar('T')


class ItemA(BaseModel):
    typ: Literal['A'] = Field(alias="T")
    a: str


class ItemB(BaseModel):
    typ: Literal['B'] = Field(alias="T")
    b: str


try:

    class Foo(BaseModel):
        item: Annotated[Union[ItemA, ItemB], Field(discriminator='T')]


except Exception as e:
    print(e)
    # "Model 'ItemA' needs a discriminator field for key 'T'"


class Foo(BaseModel):
    item: Annotated[Union[ItemA, ItemB], Field(discriminator='typ')]


try:
    Foo.validate({'item': {'T': 'A', 'a': 'foo'}})
except Exception as e:
    print(e)
    # 1 validation error for Foo
    # item
    #   Discriminator 'typ' is missing in value (type=value_error.discriminated_union.missing_discriminator; discriminator_key=typ)
```
Is this a bug, or should I change my usage?
@PrettyWood

PrettyWood:
Hi @josteinl
You just need to switch from `Field(..., discriminator='cutlery_type_id')` to `Field(discriminator='cutlery_type_id')`

josteinl:
> Hi @josteinl
> You just need to switch from `Field(..., discriminator='cutlery_type_id')` to `Field(discriminator='cutlery_type_id')`

I tried to do what you suggested, but then I get a new error when I call the endpoint. I have updated the code and the README.md in my example project (https://github.com/josteinl/pydantic-discrimination-union-problem), with more details. 

If I revert to the commit 45db4ad, the example code works both with and without the ```...``` in Field().

tests/test_schema.py
```diff
@@ -2562,3 +2562,203 @@ def resolve(self) -> 'Model':  # noqa
...
+                'discriminator': {
+                    'propertyName': 'pet_type',
+                    'mapping': {
+                        'cat': '#/definitions/Cat',
+                        'dog': '#/definitions/Dog',
+                        'reptile': '#/definitions/Lizard',
+                        'lizard': '#/definitions/Lizard',
+                    },
+                },
+                'anyOf': [
```
bendidi, line 2603: Should this still be `anyOf` instead of `oneOf` ?

StevenLoomanAlliander:
I'd like to have this too!

wholmen:
@samuelcolvin I saw your post on the "bump to 1.9"-thread, and would like to give some feedback on this branch to help you. 

I have used this branch for a large project and huge datamodel for half a year, and I have not met any problems yet. 

When using the package datamodel-code-generator to fetch this from OpenAPI-spec published by FastAPI, the Literal-types is not interpreted correctly, and the loading fails. I don't think this is related to this implementation, though, because we don't get the same problems when loading the models into typescript. That package doesn't support discriminated unions, but it does interpret Literal-typing correctly

hacker-DOM:
From https://github.com/samuelcolvin/pydantic/pull/2336#issuecomment-913498533:

> There are some issues when using discriminator with alias. A minimal example:

I am running into the same exact error messages as @AutumnSun1996 . I don't think the `T = TypeVar('T')` in his example is necessary, though.

pydantic/utils.py
```diff
@@ -665,3 +677,39 @@ def all_identical(left: Iterable[Any], right: Iterable[Any]) -> bool:
...
+
+    if hasattr(tp, '__pydantic_model__'):
+        tp = tp.__pydantic_model__
+
+    if is_root_model or is_union_origin(get_origin(tp)):
+        union_type = tp.__fields__[ROOT_KEY].type_ if is_root_model else tp
+
+        all_values = [get_discriminator_values(t, discriminator_key) for t in get_args(union_type)]
+        if len(set(all_values)) > 1:
+            raise TypeError(f'Field {discriminator_key!r} is not the same for all submodels of {tp.__name__!r}')
```
samuelcolvin, line 700: While playing around, I get an `AttributeError` from `tp.__name__` with the code below.

I'm can imagine this code working, or raising this error, but we shouldn't get the attribute error.

```py
class Cat(BaseModel):
    pet_type: Literal['cat']
    name: str


class Dog(BaseModel):
    pet_type: Literal['dog']
    name: str


CommonPet = Annotated[Union[Cat, Dog], Field(discriminator='pet_type')]


class Lizard(BaseModel):
    pet_type: Literal['reptile', 'lizard']
    name: str


class Model(BaseModel):
    pet: Union[CommonPet, Lizard] = Field(..., discriminator='pet_type')
    n: int
```

samuelcolvin (commented):
This is looking awesome. A few small things to change.

@PrettyWood let me know if you want me to make some of these changes?

samuelcolvin:
please update.

samuelcolvin:
something weird seems to be happening where tests aren't be started. Maybe it's just a matter of time?

ryukinix:
Any estimation date for the next release of pydantic with this feature included?

PrettyWood:
@ryukinix Probably before the end of the year. I'll come back on this PR asap

pydantic/typing.py
```diff
@@ -289,6 +290,19 @@ def is_none_type(type_: Any) -> bool:
         return type_ in NONE_TYPES
 
 
+elif sys.version_info[:2] == (3, 8):
+    # We can use the fast implementation for 3.8 but there is a very weird bug
+    # where it can fail for `Literal[None]`.
+    # We just need to redefine a useless `Literal[None]` inside the function body to fix this
+
+    def is_none_type(type_: Any) -> bool:
+        Literal[None]  # fix edge case
```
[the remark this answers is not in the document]
PrettyWood: @samuelcolvin `master` is broken because of this
You can try on 3.8 with a test file like this (mix of test_generics and test_typing)

```py
import sys
from typing import (
    Any,
    Callable,
    Dict,
    Generic,
    List,
    Optional,
    TypeVar,
    Union,
)

import pytest
from typing_extensions import Literal

from pydantic.generics import GenericModel, iter_contained_typevars

@skip_36
def test_generic():
    data_type = TypeVar('data_type')
    error_type = TypeVar('error_type')

    class Result(GenericModel, Generic[data_type, error_type]):
        data: Optional[List[data_type]]
        error: Optional[error_type]
        positive_number: int

        @validator('error', always=True)
        def validate_error(cls, v: Optional[error_type], values: Dict[str, Any]) -> Optional[error_type]:
            if values.get('data', None) is None and v is None:
                raise ValueError('Must provide data or error')
            if values.get('data', None) is not None and v is not None:
                raise ValueError('Must not provide both data and error')
            return v

        @validator('positive_number')
        def validate_positive_number(cls, v: int) -> int:
            if v < 0:
                raise ValueError
            return v

    class Error(BaseModel):
        message: str

    class Data(BaseModel):
        number: int
        text: str

    success1 = Result[Data, Error](data=[Data(number=1, text='a')], positive_number=1)


@skip_36
def test_complex_nesting():
    T = TypeVar('T')

    class MyModel(GenericModel, Generic[T]):
        item: List[Dict[Union[int, T], str]]

    item = [{1: 'a', 'a': 'a'}]
    model = MyModel[str](item=item)
    assert model.item == item


@skip_36
def test_iter_contained_typevars():
    T = TypeVar('T')
    T2 = TypeVar('T2')

    class Model(GenericModel, Generic[T]):
        a: T

    assert list(iter_contained_typevars(Model[T])) == [T]
    assert list(iter_contained_typevars(Optional[List[Union[str, Model[T]]]])) == [T]
    assert list(iter_contained_typevars(Optional[List[Union[str, Model[int]]]])) == []
    assert list(iter_contained_typevars(Optional[List[Union[str, Model[T], Callable[[T2, T], str]]]])) == [T, T2, T]


def test_is_none_type():
    from pydantic.typing import is_none_type

    assert is_none_type(Literal[None]) is True
    assert is_none_type(None) is True
    assert is_none_type(type(None)) is True
    assert is_none_type(6) is False
    assert is_none_type({}) is False
    # WARNING: It's important to test `typing.Callable` not
    # `collections.abc.Callable` (even with python >= 3.9) as they behave
    # differently
    assert is_none_type(Callable) is False
```

There is a side effect I don't understand (I didn't dig too much tbh)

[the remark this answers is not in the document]
samuelcolvin: This is very weird, how did find the work around.

I've fixed it on master now, might cause conflicts.

pydantic/fields.py
```diff
@@ -945,6 +1000,43 @@ def _validate_singleton(
         else:
             return self._apply_validators(v, values, loc, cls, self.validators)
 
+    def _validate_discriminated_union(
+        self, v: Any, values: Dict[str, Any], loc: 'LocStr', cls: Optional['ModelOrDc']
+    ) -> 'ValidateReturn':
+        assert self.discriminator_key is not None
+
+        if self.sub_fields_mapping is None:
+            assert cls is not None
```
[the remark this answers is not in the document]
PrettyWood: It's for `mypy` (same as https://github.com/samuelcolvin/pydantic/blob/master/pydantic/fields.py#L742). `cls` is always defined! I can use `cast()` or add `# type: ignore` if you prefer

pydantic/utils.py
```diff
@@ -665,3 +677,39 @@ def all_identical(left: Iterable[Any], right: Iterable[Any]) -> bool:
...
+
+        return all_values[0]
+
+    else:
+        try:
+            t_discriminator_type = tp.__fields__[discriminator_key].type_
+        except AttributeError:
+            raise TypeError(f'Type {tp.__name__!r} is not a valid `BaseModel` or `dataclass`')
+        except KeyError:
+            raise KeyError(f'Model {tp.__name__!r} needs a discriminator field for key {discriminator_key!r}')
```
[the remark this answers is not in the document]
PrettyWood: I just changed the first one. Feel free to modify the code directly if you want to change others

PrettyWood:
@samuelcolvin I took most of your remarks into consideration. I also added `alias` support.
TBH I'm not super happy with the implementation, which doesn't feel very robust... But handling all the different cases (root models, annotated unions, same or different discriminator...) is quite annoying with the current codebase.
Feel free to modify the PR directly I'll try to take some time on the two other big PRs (dataclass and computed fields) tonight or tomorrow

PrettyWood:
Please review

pydantic/fields.py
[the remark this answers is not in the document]
samuelcolvin: this is fine.

joaopfg:
@samuelcolvin @PrettyWood 

Thanks for the work on this branch!

I defined some models that are helpful to define my model (`ImageReader`) like below:

```
class T1(BaseModel):
    type: Literal["T1"]
    value: T1

class T5(BaseModel):
    type: Literal["T5"]
    value: bool

class Union4(BaseModel):
	value: Union["T1", "T5"] = Field(..., discriminator="type")

class T6(BaseModel):
    type: Literal["T6"]
    value: str

class T7(BaseModel):
    type: Literal["T7"]
    value: List[Tuple[str, Dict[str, "Union4"]]]

class Union6(BaseModel):
	value: Union["T1", "T6", "T7"] = Field(..., discriminator="type")

class T8(BaseModel):
    type: Literal["T8"]
    value: Dict[str, "Union6"]

class Union7(BaseModel):
	value: Union["T6", "T8"] = Field(..., discriminator="type")

class ImageReader(BaseModel):
	config_dict: Dict[str, Dict[str, "Union7"]]
```

I'm creating an instance of `ImageReader` using the following dict:

```
"config_dict": {
	"camera": {
		"constructor_key": {"value": {"type": "T6", "value": "RealSense"}},
		"kwargs_key": {"value": 
			{"type": "T8", "value": {
				"acquisition_width": {"value": {"type": "T1", "value": 1280}},
				"acquisition_height": {"value": {"type": "T1", "value": 720}},
				"align_to": {"value": {"type": "T6", "value": "color_stream"}},
				"time_acquisition_function": {"value": {"type": "T6", "value": "time_manager_now"}},
				"preset_config": {"value": {"type": "T6", "value": "High Accuracy"}},
				"filters_config": [
					["disparity", {}],
					[
						"spatial",
						{
						"smooth_alpha": {"value": {"type": "T1", "value": 0.6}},
						"smooth_delta": {"value": {"type": "T1", "value": 20.0}},
						},
					],
					["temporal", {}],
					["disparity", {"transform_to_disparity": {"value": {"type": "T5", "value": False}}}],
					[
						"hole_filling",
						{"mode": {"value": {"type": "T1", "value": 2}}},
					], 
				],
				"log_level": {"value": {"type": "T6", "value": "debug"}}
				}
			}	
		},
	}
}
```

I'm getting the following error:

```
  File "/models/image_reader.py", line 100, in <module>
    image_reader = ImageReader(**image_reader_dict)
  File "pydantic/main.py", line 406, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 3 validation errors for ImageReader
config_dict -> camera -> kwargs_key -> value -> type
  unexpected value; permitted: 'T6' (type=value_error.const; given=T8; permitted=('T6',))
config_dict -> camera -> kwargs_key -> value -> value
  str type expected (type=type_error.str)
config_dict -> camera -> kwargs_key -> value -> value -> filters_config -> value
```

Something that is anotated with `Union7` can't recognize `T8`. Is this a bug in nested unions or am I doing something wrong ? If is the last case, what would be the correct way to do it ?

Thanks in advance !

samuelcolvin:
### :tada: 

### :pray:

foarsitter:
Thanks for all your hard work folks!

PR #2337, opened 2021-02-10
chore: unpin pip and fix CI

PrettyWood:
revert https://github.com/samuelcolvin/pydantic/commit/de0657e4a5495de9378db9fe15a17334c2b0fae5

### DIFF
--- a/tests/test_fastapi.sh
+++ b/tests/test_fastapi.sh
@@ -11,4 +11,5 @@ git checkout "${latest_tag}"
 pip install -U flit
 flit install
 
-PYTHONPATH=./docs/src pytest
+# ignore cryptography warning https://github.com/mpdavis/python-jose/issues/208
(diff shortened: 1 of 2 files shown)


tests/test_fastapi.sh
```diff
@@ -11,4 +11,5 @@ git checkout "${latest_tag}"
 pip install -U flit
 flit install
 
-PYTHONPATH=./docs/src pytest
+# ignore cryptography warning https://github.com/mpdavis/python-jose/issues/208
+PYTHONPATH=./docs/src pytest -W ignore::DeprecationWarning:cryptography
```
graingert, line 15:
```suggestion
PYTHONPATH=./docs/src pytest -W 'ignore:int_from_bytes is deprecated, use int\.from_bytes instead:cryptography.utils.CryptographyDeprecationWarning'
```

PrettyWood: Thanks @graingert ! 👍

PR #2358, opened 2021-02-13
try building aarch64 with cibuildwheel

### LINKED ISSUE #2187: Wheel support for linux aarch64[arm64]
**Summary**
Installing pydantic on aarch64 via pip using command "pip3 install pydantic" tries to build wheel from source code

**Problem description**
pydantic doesn't have wheel for aarch64 on PyPI repository. So, while installing pydantic via pip on aarch64, pip builds wheel for same resulting in it takes more time to install pydantic. Making wheel available for aarch64 will benefit aarch64 users by minimizing pydantic installation time.

**Expected Output**
Pip should be able to download pydantic wheel from PyPI repository rather than building it from source code.

@pydantic-team, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make pydantic wheel available for aarch64. It will be a great opportunity for me to work with you.

samuelcolvin:
using https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation

alternative to #2239

fixes #2187

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -221,7 +221,7 @@ jobs:
       env:
         SKIP_CYTHON: 1
 
-    - name: Set up QEMU
+    - name: set up qemu
       if: matrix.os == 'ubuntu'
       uses: docker/setup-qemu-action@v1
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,7 +225,7 @@ jobs:
       if: matrix.os == 'ubuntu'
       uses: docker/setup-qemu-action@v1
       with:
-        platforms: all
+        platforms: arm64
 
     - name: build ${{ matrix.platform || matrix.os }} binaries
       run: cibuildwheel --output-dir dist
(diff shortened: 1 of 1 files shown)


samuelcolvin:
Humm, this seems to be working. The problem is that build time has increased from 9 minutes for that stage to over an hour!

We build wheels on every commit to master to avoid build errors when we deploy, so this change could cause massive backups of builds, which slows development. Additionally the feedback look to fix any error with building wheels would grow enormously.

Is `aarch64` really used enough to warrant this?

Is there another work around? A flag that makes compilation quicker, a way to run these steps nightly only or a way to ask github for our actions to run on a faster machine (I doubt it)?

samuelcolvin:
@odidev you created the original issue, what are your thoughts on this?

odidev:
@samuelcolvin, I can see that it taking around ~30 minutes to build the aarch64 wheels. 

Unfortunately, there is no way to make the wheel build faster for aarch64. One thing you can do is to build aarch64 wheels only on release of a tag in the next step.

Currently, I am working on cross compiling aarch64 wheels through 'cibuildwheel'. The will speedup the wheel build process for aarch64. For more details, please check: https://github.com/joerick/cibuildwheel/issues/598

samuelcolvin:
Thanks so much for looking into this @odidev.

How do you feel about closing this and waiting for a fix in `cibuildwheel`?

odidev:
Yes, please close this PR. I will make another PR when cibuildwheel have cross compilation support.

PR #2364, opened 2021-02-15
Fix generic BaseModel schema generation (rebase)

danni:
## Change Summary

Following on from #2260 this is a rebase onto master with some fixes included.

## Related issue number

* #1578 
* #2260

## Checklist

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -373,7 +373,7 @@ def get_flat_models_from_field(field: ModelField, known_models: TypeModelSet) ->
     field_type = field.type_
     if lenient_issubclass(getattr(field_type, '__pydantic_model__', None), BaseModel):
         field_type = field_type.__pydantic_model__
-    if field.sub_fields:
+    if field.sub_fields and not lenient_issubclass(field_type, BaseModel):
         flat_models |= get_flat_models_from_fields(field.sub_fields, known_models=known_models)
     elif lenient_issubclass(field_type, BaseModel) and field_type not in known_models:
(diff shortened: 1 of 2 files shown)


pydantic/schema.py
```diff
@@ -502,6 +501,8 @@ def field_type_schema(
         )
         definitions.update(f_definitions)
         nested_models.update(f_nested_models)
+    else:
+        raise AssertionError('Unknown shape: %i' % field.shape)
```
samuelcolvin, line 505: f-string please. Also `ValueError` would probably make more sense.

I also suspect this line has no coverage which is why the coverage has gone down.

samuelcolvin: you could also probable provide a more informative error message.

danni: I'm honestly not sure how to trigger this assertion now generics are fixed, as I'm not sure what cases are no longer handled. I simply moved it from a check inside the `else` block to be the `else` block. Which I think aids the readability.

samuelcolvin: pydantic has 100% coverage, I would like to keep it that way.

If we really can't reach this could, it shouldn't exist. But I would be confident we can if we really try.

samuelcolvin: you just need to call the method directly with a custom `ModelField` which has a `shape` which pydantic doesn't recognise.

danni: I fiddled with this for a bit, but the test seemed rather fragile and contrived. I reverted the code back to the way it was which is easier to test.

danni:
Just wondering if anything is still holding this up?

PR #2368, opened 2021-02-15
fix coverage and make typing-extensions a required dependency

### LINKED ISSUE #2367: Put back coverage at 100%
Just saw that. Can have a look tonight

samuelcolvin:
fix #2367

### DIFF
--- a/pydantic/annotated_types.py
+++ b/pydantic/annotated_types.py
@@ -16,29 +16,21 @@ def create_model_from_typeddict(typeddict_cls: Type['TypedDict'], **kwargs: Any)
     """
     Create a `BaseModel` based on the fields of a `TypedDict`.
     Since `typing.TypedDict` in Python 3.8 does not store runtime information about optional keys,
-    we warn the user if that's the case (see https://bugs.python.org/issue38834).
+    we raise an error if this happens (see https://bugs.python.org/issue38834).
     """
     field_definitions: Dict[str, Any]
 
     # Best case scenario: with python 3.9+ or when `TypedDict` is imported from `typing_extensions`
-    if hasattr(typeddict_cls, '__required_keys__'):
-        field_definitions = {
-            field_name: (field_type, Required if field_name in typeddict_cls.__required_keys__ else None)
-            for field_name, field_type in typeddict_cls.__annotations__.items()
-        }
-    else:
-        import warnings
-
-        warnings.warn(
-            'You should use `typing_extensions.TypedDict` instead of `typing.TypedDict` for better support! '
-            'Without it, there is no way to differentiate required and optional fields when subclassed. '
-            'Fields will therefore be considered all required or all optional depending on class totality.',
-            UserWarning,
+    if not hasattr(typeddict_cls, '__required_keys__'):
+        raise TypeError(
+            'You should use `typing_extensions.TypedDict` instead of `typing.TypedDict`. '
+            'Without it, there is no way to differentiate required and optional fields when subclassed.'
         )
-        default_value = Required if typeddict_cls.__total__ else None
-        field_definitions = {
-            field_name: (field_type, default_value) for field_name, field_type in typeddict_cls.__annotations__.items()
-        }
+
+    field_definitions = {
+        field_name: (field_type, Required if field_name in typeddict_cls.__required_keys__ else None)
+        for field_name, field_type in typeddict_cls.__annotations__.items()
+    }
 
     return create_model(typeddict_cls.__name__, **kwargs, **field_definitions)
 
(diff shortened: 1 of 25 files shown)


cybojenix:
Do you think it's worth changing to depending on `typing-extensions>=3.7.4.3` going forward?
That should remove the need for all the custom implementations of `get_args`, and guarantees availability of `Annotated`.

I'm unsure if anyone's raised it, but could also prevent issues with type checkers who don't use `typing-extensions` (Pyright comes to mind here, but I'd have to check if it's an actual issue or not).

tests/test_annotated.py
```diff
@@ -7,8 +7,9 @@
 from pydantic.fields import Undefined
 from pydantic.typing import Annotated
 
+pytestmark = pytest.mark.skipif(not Annotated, reason='typing_extensions not installed')
```
PrettyWood, line 10: 👍

PrettyWood (approved):
LGTM!

samuelcolvin:
This is a change to the `Annotated` solution implemented in #2147, @JacobHayes are you happy with it?

> Do you think it's worth changing to depending on `typing-extensions>=3.7.4.3` going forward?

I think that's a good idea @JacobHayes suggested it on the original PR, I'll see how easy this would be now.

JacobHayes (approved):
PR LGTM! The world is probably a better place without `_FalseMeta` 😁

> Do you think it's worth changing to depending on typing-extensions>=3.7.4.3 going forward?
That should remove the need for all the custom implementations of get_args, and guarantees availability of Annotated.

It's probably a net benefit, but there were a [few caveats](https://github.com/samuelcolvin/pydantic/pull/2147#discussion_r551748778) to making things very clean:

> I was able to refactor the get_args/get_origin stuff a bit with typing_extensions, but it's not a panacea:
>
> - those helper funcs aren't available on 3.6 so I kept the old "polyfills"
> - I had to remove two tests using a bare Callable (the typing_extensions functions seem to expect the Callable generics to be [...]). Might get just as messy again to add the shim if those are truly valid cases (creating a Model with a bare Callable field hint seemed to still work)
> - Looks like this makes the "test compiled without deps" fail (since we now import typing_extensions). Does this need to be tweaked somehow?

~It might be pretty tidy after dropping py 3.6 though, I think(?) the other two are more minor.~ Looks like @samuelcolvin whipped that up pretty quick. 👍

samuelcolvin:
Looks ready. @PrettyWood are you happy with this?

I ❤️ negative PRs 😍.

pydantic/annotated_types.py
```diff
@@ -16,29 +16,21 @@ def create_model_from_typeddict(typeddict_cls: Type['TypedDict'], **kwargs: Any)
...
-    else:
-        import warnings
-
-        warnings.warn(
-            'You should use `typing_extensions.TypedDict` instead of `typing.TypedDict` for better support! '
-            'Without it, there is no way to differentiate required and optional fields when subclassed. '
-            'Fields will therefore be considered all required or all optional depending on class totality.',
-            UserWarning,
+    if not hasattr(typeddict_cls, '__required_keys__'):
+        raise TypeError(
```
cybojenix, line 25: This change should probably be explicitly noted as a breaking change in the changelog, just in case there are people ignoring the warning.

I can also think of scenarios where people are using another libraries TypedDict, but I'm not sure if that's common or not (I've never done this before).

PrettyWood: No need it's in the same new version v1.8 ;)

PrettyWood:
@samuelcolvin We could maybe update `docs/examples/schema_annotated.py` but up to you.
Looks great!

cybojenix (approved):
Nice work. I didn't realise that typing-extensions didn't backport get_args for py36, so future trimmings to come

pydantic/typing.py
```diff
@@ -170,11 +101,55 @@ def get_origin(tp: Type[Any]) -> Type[Any]:
...
-        if Annotated and type(tp).__name__ in AnnotatedTypeNames:
+        if type(tp).__name__ in AnnotatedTypeNames:
             return cast(Type[Any], Annotated)  # mypy complains about _SpecialForm
-        return typing_get_origin(tp) or getattr(tp, '__origin__', None)
+        return _typing_get_origin(tp) or getattr(tp, '__origin__', None)
+
+
+# Annotated[...] is implemented by returning an instance of one of these classes, depending on
+# python/typing_extensions version.
+AnnotatedTypeNames = {'AnnotatedMeta', '_AnnotatedAlias'}
```
JacobHayes, line 111: Should this be moved up since it's referenced in the branches above?

JacobHayes (approved):
Very nice cleanup!

samuelcolvin:
> It might be pretty tidy after dropping py 3.6 though, I think(?) the other two are more minor.

#2374

PR #2414, opened 2021-02-26
preparing for v1.8

samuelcolvin:
:tada: :rocket:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -15,7 +15,7 @@ for their kind support.
 
 ### Changes
 
-* **Breaking Change:**, remove old deprecation aliases from v1, #2415 by @samuelcolvin:
+* **Breaking Change**, remove old deprecation aliases from v1, #2415 by @samuelcolvin:
   * remove notes on migrating to v1 in docs
   * remove `Schema` which was replaced by `Field`
   * remove `Config.case_insensitive` which was replaced by `Config.case_sensitive` (default `False`)
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -8,21 +8,21 @@ for their kind support.
 
 ### Highlights
 
-* [Hypothesis plugin](https://pydantic-docs.helpmanual.io/hypothesis_plugin/) for testing, thanks @Zac-HD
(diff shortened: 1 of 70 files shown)


samuelcolvin:
Humm, I hadn't noticed that the coverage has fallen, I need to fix that before release.

cybojenix:
Just run our tests targeting the master branch, and no regressions.
We don't use all the features, but it's a good sign that even our use of some private APIs works correctly (mainly on `validate_arguments`).

Thanks!

samuelcolvin:
> Just run our tests targeting the master branch, and no regressions.
> We don't use all the features, but it's a good sign that even our use of some private APIs works correctly (mainly on `validate_arguments`).

Thanks for checking! So useful when a few people run it against their tests before rather than after release.

HISTORY.md
```diff
@@ -1,3 +1,102 @@
...
+* fix: update all modified field values in `root_validator` when `validate_assignment` is on, #2116 by @PrettyWood
+* Allow pickling of `pydantic.dataclasses.dataclass` dynamically created from a built-in `dataclasses.dataclass`, #2111 by @aimestereo
+* Fix a regression where Enum fields would not propagate keyword arguments to the schema, #2109 by @bm424
+* Add the ability to customize settings sources (add / disable / change priority order), #2107 by @kozlek
+* Fix mypy complaints about most custom _pydantic_ types, #2098 by @PrettyWood
+* Add a [Hypothesis](https://hypothesis.readthedocs.io/) plugin for easier [property-based testing](https://increment.com/testing/in-praise-of-property-based-testing/) with Pydantic's custom types - [usage details here](https://pydantic-docs.helpmanual.io/hypothesis_plugin/), #2097 by @Zac-HD
+* add validator for `None`, `NoneType` or `Literal[None]`, #2095 by @PrettyWood
+* Handle properly fields of type `Callable` with a default value, #2094 by @PrettyWood
+* Ignore `__doc__` as private attribute when `Config.underscore_attrs_are_private` is set, #2090 by @PrettyWood
```
PrettyWood, line 79:
```suggestion
```
I think some stuff comes from 1.7.3

samuelcolvin: good spot, will remove.

samuelcolvin: done.

PR #2415, opened 2021-02-26
remove DeprecationWarnings from v1 release & fix coverage

samuelcolvin:
To fix coverage before v1.8 #2414.

Most of these warnings should have been removed a long time again.

Removed:
* notes on migrating to v1
* `Schema` which was replaced by `Field`
* `Config.case_insensitive` which was replaced by `Config. case_sensitive` (default `False`)
*  `Config.allow_population_by_alias` which was replaced by `Config.allow_population_by_field_name`
* `model.fields` which was replaced by `model.__fields__`
* `model.to_string()` which was replaced by `str(model)`
* `model.__values__` which was replaced by `__dict__`


pydantic/main.py
```diff
@@ -466,18 +451,18 @@ def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
         self.__fields_set__.add(name)
 
     def __getstate__(self) -> 'DictAny':
+        private_attrs = ((k, getattr(self, k, Undefined)) for k in self.__private_attributes__)
         return {
             '__dict__': self.__dict__,
             '__fields_set__': self.__fields_set__,
-            '__private_attribute_values__': {k: getattr(self, k, Undefined) for k in self.__private_attributes__},
+            '__private_attribute_values__': {k: v for k, v in private_attrs if v is not Undefined},
```
samuelcolvin, line 458: @MrMrRobat I think you wrote this originally, are you happy with the change.

Basically the `if value is not Undefined:` below was always `True`, I because the behaviour of `Undefined` when pickled has changed slightly.

MrMrRobat: LGTM!

samuelcolvin: @MrMrRobat do you have any idea why the `if value is not Undefined:` line in `copy()` now has partial coverage, but used to be covered.

I'm trying to work out, but it makes no sense.

Sorry, I know it's not your problem but it's really confusing me.

samuelcolvin: I think it could perhaps be a change in cython, but I'm not sure. Or it could even be a bug fix in codecov? 😕 🤷

MrMrRobat: Maybe it's because now only present attrs are pickled, while in the past missing attrs used to be pickled as `Undefined`? Just a quick assumption, I didn't dig into new code yet.

samuelcolvin: maybe, it was the same on `copy()`.

Hopefully I fixed the coverage without breaking anything 🤞.

PR #2420, opened 2021-02-27
fix(mypy): fix custom `Path` and `UUID` related types

### LINKED ISSUE #2418: UUID typing hints of Pydantic 1.8 are inaccurate
### Checks

# Bug

It appears commit 7da04d95fa4a7b473a9dae4d1ab9d7eba98a9aa9 introduces a type erasure on `UUID1`/`UUID2`/`UUID3`/`UUID4` that makes them all `Union[UUID, str]`, losing the strict inheritance from `UUID`.

Now:

```python
from uuid import UUID, uuid4
from pydantic import UUID4, BaseModel

def expects_any_uuid(arg: UUID):
    pass

class MyModel(BaseModel):
    my_uuid: UUID4


# While these two are now correct in regards to type checking:
# (they were triggering typing errors in v1.7)
MyModel(my_uuid=uuid4())                                 # 1
MyModel(my_uuid="5259A9AA-D1FD-4D0A-B17D-87803904C233")  # 2

# These two calls are not correct anymore:
expects_any_uuid(MyModel(my_uuid=uuid4()).my_uuid)       # 3
expects_any_uuid(uuid4())                                # 4
# Type checking like MyPy would say something along the following:
#   Argument 1 to "expects_any_uuid" has incompatible type "Union[UUID, str]"; expected "UUID"
```

I would argue that making `# 2` clear the typing error from 1.7 is nice to have, but we could already use something like a `# type: ignore` comment to explicitly tell type-checking, which was acceptable since some `str`s are not `UUID4`s, and as such it makes sense to have to explicitly silent a warning here.

Same with `# 1`, since its type is “erased” to a generic `UUID`, we would need to check again that it is indeed a v4 UUID.

On the other hand, a `UUID4` is always a `UUID`, so `# 3` and `# 4` should never trigger a static typing error. 

I don't have a solution to make all examples `# 1` to `# 4` from above work, so instead I'm opening an issue here in hopes someone would have a solution.

---

```
             pydantic version: 1.8
            pydantic compiled: True
                 install path: <redacted>/venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.2 (default, Feb 19 2021, 17:43:04)  [Clang 12.0.0 (clang-1200.0.32.29)]
                     platform: macOS-11.2.1-x86_64-i386-64bit
```

PrettyWood:
## Change Summary
I went a bit too far with #2099 because I was first focused on "accepted" types and not "validated" types.

## Related issue number
closes #2418

## Checklist

### DIFF
--- a/tests/mypy/modules/success.py
+++ b/tests/mypy/modules/success.py
@@ -220,3 +220,12 @@ class PydanticTypes(BaseModel):
     my_dir_path_str: DirectoryPath = 'root/mydir'
     # Json
     my_json: Json = '{"hello": "world"}'
+
+
+PydanticTypes().my_pyobject_callable(2021, 1, 1)
+PydanticTypes().my_uuid1.hex
+PydanticTypes().my_uuid1_str.hex
+PydanticTypes().my_file_path.absolute()
+PydanticTypes().my_file_path_str.absolute()
+PydanticTypes().my_dir_path.absolute()
+PydanticTypes().my_dir_path_str.absolute()
--- a/tests/mypy/modules/success.py
+++ b/tests/mypy/modules/success.py
@@ -208,24 +208,29 @@ class PydanticTypes(BaseModel):
     # String
     my_strict_str: StrictStr = 'pika'
     # PyObject
-    # TODO: my_pyobject_str: PyObject = 'datetime.date'
+    my_pyobject_str: PyObject = 'datetime.date'  # type: ignore
     my_pyobject_callable: PyObject = date
     # UUID
     my_uuid1: UUID1 = UUID('a8098c1a-f86e-11da-bd1a-00112444be1e')
-    my_uuid1_str: UUID1 = 'a8098c1a-f86e-11da-bd1a-00112444be1e'
+    my_uuid1_str: UUID1 = 'a8098c1a-f86e-11da-bd1a-00112444be1e'  # type: ignore
     # Path
-    my_file_path: FilePath = Path('root') / 'myfile.txt'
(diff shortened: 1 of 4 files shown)


samuelcolvin (approved):
LGTM, @Allezxandre are you happy with this fix?

Allezxandre (approved):
Looks good to me!

samuelcolvin:
great thank you.

tests/mypy/modules/success.py
```diff
@@ -208,15 +208,29 @@ class PydanticTypes(BaseModel):
...
+    my_uuid1_str: UUID1 = 'a8098c1a-f86e-11da-bd1a-00112444be1e'  # type: ignore
     # Path
-    my_file_path: FilePath = Path('root') / 'myfile.txt'
-    my_file_path_str: FilePath = 'root/myfile.txt'
-    my_dir_path: DirectoryPath = Path('root') / 'mydir'
-    my_dir_path_str: DirectoryPath = 'root/mydir'
+    my_file_path: FilePath = Path(__file__)
+    my_file_path_str: FilePath = __file__  # type: ignore
+    my_dir_path: DirectoryPath = Path('.')
+    my_dir_path_str: DirectoryPath = '.'  # type: ignore
```
samuelcolvin, line 220: I guess one day, when we solve https://github.com/koxudaxi/pydantic-pycharm-plugin/issues/36#issuecomment-618478815 we could update  the mypy plugin to avoid the need for all these `# type: ignore`s

PR #2425, opened 2021-02-28
feat: add PastDate and FutureDate types

Kludex:
## Change Summary

Add `PastDate` and `FutureDate` types.

## Related issue number

closes https://github.com/samuelcolvin/pydantic/issues/769

## Checklist


Kludex:
@samuelcolvin Do I need to add it to the `schema_mapping.py`? Just to confirm.

pydantic/types.py
```diff
@@ -1008,3 +1012,42 @@ def to(self, unit: str) -> float:
...
         return self / unit_div
+
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ DATE TYPES ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+class PastDate(date):
+    @classmethod
+    def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
+        field_schema.update(format='past-date')
```
PrettyWood, line 1023: I think this should be removed since `'past-date'` [doesn't seem to be valid](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times).
If you remove both `__modify_schema__`, `format: 'date'` will be added thanks to inheritance

Kludex: Ah! Got it! Thank you.

Kludex: Fixed.

Kludex:
@PrettyWood Do you have more comments?

Thank you for the previous one btw :)

tests/test_types.py
```diff
@@ -2774,3 +2776,147 @@ class Model(BaseModel):
...
+                    'loc': ('foo',),
+                    'msg': f'date "{date.today() + timedelta(1)}" is not in the past',
+                    'type': 'value_error.date.not_in_the_past',
+                    'ctx': {'date': str(date.today() + timedelta(1))},
+                }
+            ],
+        ),
+    ),
+)
+def test_past_date_validation_fails(value, errors):
```
PrettyWood, line 2844: You could probably just write an input value and parse value as params and write the whole error object in the test

PrettyWood (commented):
Otherwise looks good :)

tests/test_types.py
Kludex: True. Going to apply this tonight, thanks :)

Kludex: I've applied some modifications. Is that what you meant? @PrettyWood

PrettyWood: Yes thanks!

pydantic/types.py
```diff
@@ -51,52 +51,52 @@
 )
 
 __all__ = [
-    "NoneStr",
```
StephenBrown2: @ samuelcolvin prefers single quotes, so you'll need to undo this and rerun with `-S`, as specified in https://github.com/samuelcolvin/pydantic/blob/master/Makefile#L3

It would be better in fact to simply run `make format`, which will give all the required parameters needed.

Kludex: I think you've reviewed an old commit :thinking:

StephenBrown2: So I did. Carry on!

pydantic/types.py
```diff
@@ -1008,3 +1012,34 @@ def to(self, unit: str) -> float:
             raise errors.InvalidByteSizeUnit(unit=unit)
 
         return self / unit_div
+
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ DATE TYPES ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+class PastDate(date):
```
PrettyWood, line 1020: I reckon you should add
```py
if TYPE_CHECKING:
    PastDate = date
    FutureDate = date
```
and add two lines in `tests/mypy/modules/success.py` in `PydanticTypes` class

PrettyWood (changes requested):

Kludex:
FastAPI test failing due to:
```bash
E   ImportError: cannot import name 'RowProxy' from 'sqlalchemy.engine.result' (/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/sqlalchemy/engine/result.py)
```
:thinking:

pydantic/types.py
PrettyWood: Just a dumb model with most custom pydantic types to ensure `mypy` is happy 😊

PrettyWood (approved):
LGTM ! Thanks for making the changes.

Note: the CI error with fastapi has nothing to do with your PR don't worry ;)

PrettyWood:
@Kludex you can rebase on master to have the new CI running :)

Kludex:
> @Kludex you can rebase on master to have the new CI running :)

Done :) Thank you!

PrettyWood (approved):
Perfect thanks

pydantic/errors.py
```diff
@@ -66,6 +67,9 @@
     'DecimalWholeDigitsError',
     'DateTimeError',
     'DateError',
+    '_DateValueError',
```
samuelcolvin, line 70:
```suggestion
```

pydantic/errors.py
```diff
@@ -438,6 +442,21 @@ class DateError(PydanticValueError):
...
 
 
+class _DateValueError(PydanticValueError):
+    def __init__(self, *, date: date) -> None:
+        super().__init__(date=str(date))
+
+
+class DateNotInThePastError(_DateValueError):
+    code = 'date.not_in_the_past'
+    msg_template = 'date "{date}" is not in the past'
```
samuelcolvin, line 452: As per #1421, we generally don't include the invalid value in error contexts. In v2 we will always include it.

For now in this case, to keep things simple and consistent with other errors, I think we should remove this and replace it with just `date is not in the past`.

tests/test_types.py
```diff
@@ -2774,3 +2776,81 @@ class Model(BaseModel):
...
+@pytest.mark.parametrize(
+    'value,expected',
+    (
+        (date.today(), str(date.today())),
+        (date.today() + timedelta(1), str(date.today() + timedelta(1))),
+        (datetime.today(), str(date.today())),
+        (datetime.today() + timedelta(1), str(date.today() + timedelta(1))),
+    ),
+)
+def test_past_date_validation_fails(value, expected):
```
samuelcolvin, line 2804: the second argument will not be required after the change requested above, but note for future reference: the second argument is not "expected" rather `date_str` or similar.

tests/test_types.py
```diff
@@ -2774,3 +2776,81 @@ class Model(BaseModel):
...
+    assert Model(foo=value).foo == result
+
+
+@pytest.mark.parametrize(
+    'value,expected',
+    (
+        (date.today(), str(date.today())),
+        (date.today() - timedelta(1), str(date.today() - timedelta(1))),
+        (datetime.today(), str(date.today())),
+        (datetime.today() - timedelta(1), str(date.today() - timedelta(1))),
```
samuelcolvin, line 2840: Please include one explicit case '2064-06-01' or something so we have a hardcoded case.

If pydantic is still alive in 2064, and anyone is still alive to use it, they can fix the test then. 😨

samuelcolvin:
please update.

pydantic/errors.py
Kludex: Do I still pass the value to the error?
```python
class _DateValueError(PydanticValueError):
    def __init__(self, *, date: date) -> None:
        super().__init__(date=str(date))
```

tests/test_types.py
Kludex: Based on this comment, I'll assume the answer for my previous question is "no".

Also, thanks! :)

Kludex:
@samuelcolvin Changes applied.

pydantic/errors.py
```diff
@@ -438,6 +440,21 @@ class DateError(PydanticValueError):
     msg_template = 'invalid date format'
 
 
+class _DateValueError(PydanticValueError):
+    def __init__(self) -> None:
+        super().__init__()
```
PrettyWood, line 445: You probably don't need this anymore and can make `DateNotInThePastError` and `DateNotInTheFutureError` inherit from `PydanticValueError`

PrettyWood (approved):

Kludex:
Thanks @PrettyWood :kissing:

PrettyWood:
@Kludex Can you just rebase/merge with master for the CI to run again properly? I reckon this PR can be merged now :)

tests/test_types.py
```diff
@@ -2813,3 +2815,81 @@ class Model(BaseModel):
...
+
+
+@pytest.mark.parametrize(
+    'value,result',
+    (
+        (date.today() + timedelta(1), date.today() + timedelta(1)),
+        (datetime.today() + timedelta(1), date.today() + timedelta(1)),
+    ),
+)
+def test_future_date_validation_success(value, result):
```
PrettyWood, line 2866: Last thing sorry but I think you could add `'2064-06-01'` also here

PrettyWood:
Thanks!

Issue #2424, opened 2021-02-28
default_factory field fails with inherited stdlib dataclass


ahirner:
### Checks

* [X] I added a descriptive title to this issue
* [X] I have searched (google, github) for similar issues and couldn't find anything
* [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.8
            pydantic compiled: True
                 install path: /Users/dafcok/miniconda3/lib/python3.6/site-packages/pydantic
               python version: 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46)  [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
                     platform: Darwin-19.6.0-x86_64-i386-64bit
     optional deps. installed: ['devtools', 'dotenv', 'email-validator', 'typing-extensions'
```
```bash
$ pip freeze |grep 'dataclasses\|typing'
dataclasses==0.8
typing==3.6.6
typing-extensions==3.7.4.3
```
```py
from dataclasses import dataclass, field

from pydantic.dataclasses import dataclass as validated_dataclass

@dataclass
class Base:
    x: str

# doesn't fail
@dataclass
class Thing(Base):
    y: str = field(default_factory=str)

# fails
@validated_dataclass
class ValidatedThing(Base):
    y: str = field(default_factory=str)
```

`dataclasses` tries to delete `y` attribute which isn't set.
```
tests/test_resources.py:12: in <module>
    class ValidatedThing(Base):
pydantic/dataclasses.py:255: in pydantic.dataclasses.dataclass
    ???
pydantic/dataclasses.py:250: in pydantic.dataclasses.dataclass.wrap
    ???
pydantic/dataclasses.py:158: in pydantic.dataclasses._process_class
    ???
~/lib/python3.6/site-packages/dataclasses.py:958: in dataclass
    return wrap(_cls)
~/lib/python3.6/site-packages/dataclasses.py:950: in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
~/lib/python3.6/site-packages/dataclasses.py:817: in _process_class
    delattr(cls, f.name)
E   AttributeError: y
```

Also fails on `1.7.1` but not on `1.7`. Any ideas? I'd be happy to author a fix.

[...]

PR #2437, opened 2021-03-01
fix: do not ignore `Config` fields information in `dataclass`

### LINKED ISSUE #2426: dataclass config fields not warranted in json schema (1.8)
### Checks

# Bug

```
             pydantic version: 1.8
            pydantic compiled: True
                 install path: /Users/dafcok/miniconda3/lib/python3.6/site-packages/pydantic
               python version: 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46)  [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
     optional deps. installed: ['devtools', 'dotenv', 'email-validator', 'typing-extensions'
```
```py

# works
class A(BaseModel):
    a: str

    class Config:
        fields = {"a": {"description": "descr"}}

assert A.schema()["properties"]["a"]["description"] == "descr"

# works on 1.7.3
# fails on 1.8
@dataclass(config=A.Config)
class Ad:
    a: str

assert Ad.__pydantic_model__.schema()["properties"]["a"]["description"] == "descr"
```
```bash
  File "tests/test_schema_descr.py", line 24, in <module>
    assert Ad.__pydantic_model__.schema()["properties"]["a"]["description"] == "descr"
KeyError: 'description
```
It seems that `config.fields` is not processed in the decorator.

PrettyWood:
## Change Summary
We used to keep only the default value and the type when creating a `dataclass`, which would then follow the default workflow. But now we can use `metadata` in `dataclasses.field` or directly a `pydantic.Field` instance. We hence ignore the information set in the `Config`

## Related issue number
Closes #2426

## Checklist

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -4,12 +4,12 @@ from .class_validators import gather_all_validators
 from .error_wrappers import ValidationError
 from .errors import DataclassTypeError
 from .fields import Field, FieldInfo, Required, Undefined
-from .main import create_model, validate_model
+from .main import BaseConfig, create_model, inherit_config, validate_model
 from .typing import resolve_annotations
 from .utils import ClassAttribute
 
 if TYPE_CHECKING:
-    from .main import BaseConfig, BaseModel  # noqa: F401
+    from .main import BaseModel  # noqa: F401
     from .typing import CallableGenerator, NoArgAnyCallable
 
     DataclassT = TypeVar('DataclassT', bound='Dataclass')
@@ -160,7 +160,9 @@ def _process_class(
     )
     cls.__processed__ = ClassAttribute('__processed__', True)
 
+    config_instance = inherit_config(config, BaseConfig)
     field_definitions: Dict[str, Any] = {}
+
     for field in dataclasses.fields(cls):
         default: Any = Undefined
         default_factory: Optional['NoArgAnyCallable'] = None
@@ -178,7 +180,12 @@ def _process_class(
             field_info = default
             cls.__has_field_info_default__ = True
         else:
-            field_info = Field(default=default, default_factory=default_factory, **field.metadata)
+            field_info = Field(
+                default=default,
+                default_factory=default_factory,
+                **config_instance.get_field_info(field.name),
+                **field.metadata,
+            )
 
         field_definitions[field.name] = (field.type, field_info)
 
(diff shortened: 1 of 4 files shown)


samuelcolvin:
Would this also fix #2424?

PrettyWood:
Unfortunetaly no. #2424 seems trickier.

pydantic/dataclasses.py
```diff
@@ -178,7 +180,12 @@ def _process_class(
             field_info = default
             cls.__has_field_info_default__ = True
         else:
-            field_info = Field(default=default, default_factory=default_factory, **field.metadata)
+            field_info = Field(
+                default=default,
+                default_factory=default_factory,
+                **config_instance.get_field_info(field.name),
```
samuelcolvin, line 186: Humm, I think this could lead to unexpected behaviour - `get_field_info` is not designed to return kwargs for `Field`, I think it's not used for this anywhere else.

Perhaps we could add some tests of using `get_field_info` to create `Fields` and update the docstring?

samuelcolvin: But in theory this should all be done when creating the field as it is normally, maybe we should work out why it's not?

PrettyWood: This is really an issue of merging infos from config to an instance of `Field`. As a matter of fact it's the same with `create_model`

```py
from pydantic import create_model, Field

class Config:
    fields = {'a': {'description': 'descr'}}

Ad = create_model('Ad', __config__=Config, a=(str, ...))
assert Ad.schema()['properties'] == {'a': {'title': 'A', 'description': 'descr', 'type': 'string'}}

Ad = create_model('Ad', __config__=Config, a=(str, Field(...)))
assert Ad.schema()['properties'] == {'a': {'title': 'A', 'type': 'string'}}
```

@samuelcolvin I would suggest we don't include this in `1.8.1` and consider this as a classic "bug".
We can then take time to refactor calmly some stuff if needed as you suggested

samuelcolvin: humm, I understand what you mean, but I think given that it's new since 1.8, we should try to fix it if possible.

I will look into it now.

samuelcolvin: Turns out the same thing even happens with a vanilla pydantic model:

```py

class Foo(BaseModel):
    a: str = Field(...)

    class Config:
        fields = {'a': {'description': 'descr'}}

debug(Foo.schema()['properties'])
```

I'll look into it now.

samuelcolvin:
superseded by #2461

PR #2438, opened 2021-03-01
fix: avoid `RecursionError` when using some types like `Enum` or `Literal` with generic models

### LINKED ISSUE #2436: Combining string enums with generics results in stack overflow in Pydantic 1.8
### Checks

# Bug

```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: C:\Soft_Dev\LIVE_tls_application\back_end\.venv\Lib\site-packages\pydantic
               python version: 3.9.1 (tags/v3.9.1:1e5d33e, Dec  7 2020, 17:08:21) [MSC v.1927 64 bit (AMD64)]
```

Love Pydantic, great stuff, etc etc. 

I came across a bug in the following code (which works fine in 1.7.3) - the iteratively parsing of the model leads to stack overflow. 

```py
from enum import Enum

class SomeStringEnum(str, Enum):
    A = "A"
    B = "B"


T = TypeVar("T")


class SomeGenericModel(GenericModel, Generic[T]):
    some_field: T


class MyModel(BaseModel):
    some_field: SomeGenericModel[SomeStringEnum]


print(MyModel.parse_obj({"some_field": {"some_field": "A"}}))
```

### LINKED ISSUE #2454: RecursionError when using Literal types in GenericModel
## Bug

If we are trying to use Literal types in GenericModel we are getting `RecursionError: maximum recursion depth exceeded in comparison`.

The minimal code example:
```py
from typing import Literal, Dict, TypeVar, Generic

Fields = Literal['foo', 'bar']

FieldType = TypeVar('FieldType')
ValueType = TypeVar('ValueType')

class GModel(GenericModel, Generic[FieldType, ValueType]):
    field: Dict[FieldType, ValueType]

GModelType = GModel[Fields, str]
```

The traceback:
```
  File "scratch_101.py", line 12, in <module>
    GModelType = GModel[Fields, str]
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 110, in __class_getitem__
    {param: None for param in iter_contained_typevars(typevars_map.values())}
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 110, in <dictcomp>
    {param: None for param in iter_contained_typevars(typevars_map.values())}
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 216, in iter_contained_typevars
    yield from iter_contained_typevars(var)
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 220, in iter_contained_typevars
    yield from iter_contained_typevars(arg)
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 216, in iter_contained_typevars
    yield from iter_contained_typevars(var)
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 216, in iter_contained_typevars
    yield from iter_contained_typevars(var)
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 216, in iter_contained_typevars
    yield from iter_contained_typevars(var)
  [Previous line repeated 982 more times]
  File "virtualenvs\foobar-HGIuaRl7-py3.9\lib\site-packages\pydantic\generics.py", line 214, in iter_contained_typevars
    elif isinstance(v, Iterable):
  File "C:\Programs\Python\Python39_x64\lib\typing.py", line 657, in __instancecheck__
    return self.__subclasscheck__(type(obj))
  File "C:\Programs\Python\Python39_x64\lib\typing.py", line 789, in __subclasscheck__
    return issubclass(cls, self.__origin__)
  File "C:\Programs\Python\Python39_x64\lib\abc.py", line 102, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
RecursionError: maximum recursion depth exceeded in comparison
```

----

```
pydantic version: 1.8
        pydantic compiled: True
             install path: /home/esp/.cache/pypoetry/virtualenvs/foobar-n-Q38lsL-py3.9/lib/python3.9/site-packages/pydantic
           python version: 3.9.1 (default, Dec  8 2020, 02:26:20)  [GCC 9.3.0]
                 platform: Linux-4.4.0-19041-Microsoft-x86_64-with-glibc2.31
```

PrettyWood:
## Change Summary
Since an `Enum` is an Iterable, which would cause an infinite loop.

## Related issue number
closes #2436
closes #2454

## Checklist

### DIFF
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -1,13 +1,11 @@
 import sys
 import typing
-from enum import Enum
 from typing import (
     TYPE_CHECKING,
     Any,
     ClassVar,
     Dict,
     Generic,
-    Iterable,
     Iterator,
     List,
     Mapping,
@@ -206,13 +204,16 @@ def check_parameters_count(cls: Type[GenericModel], parameters: Tuple[Any, ...])
         raise TypeError(f'Too {description} parameters for {cls.__name__}; actual {actual}, expected {expected}')
 
 
+DictValues: Type[Any] = {}.values().__class__
+
+
 def iter_contained_typevars(v: Any) -> Iterator[TypeVarType]:
     """Recursively iterate through all subtypes and type args of `v` and yield any typevars that are found."""
     if isinstance(v, TypeVar):
         yield v
     elif hasattr(v, '__parameters__') and not get_origin(v) and lenient_issubclass(v, GenericModel):
         yield from v.__parameters__
-    elif isinstance(v, Iterable) and not lenient_issubclass(v, Enum):
+    elif isinstance(v, (DictValues, list)):
         for var in v:
             yield from iter_contained_typevars(var)
     else:
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -1,5 +1,6 @@
 import sys
 import typing
+from enum import Enum
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -211,7 +212,7 @@ def iter_contained_typevars(v: Any) -> Iterator[TypeVarType]:
         yield v
     elif hasattr(v, '__parameters__') and not get_origin(v) and lenient_issubclass(v, GenericModel):
         yield from v.__parameters__
-    elif isinstance(v, Iterable):
+    elif isinstance(v, Iterable) and not lenient_issubclass(v, Enum):
         for var in v:
             yield from iter_contained_typevars(var)
     else:
--- a/changes/2436-PrettyWood.md
+++ b/changes/2436-PrettyWood.md
@@ -1,0 +1,1 @@
+Support properly `Enum` when combined with generic models
--- a/changes/2436-PrettyWood.md
+++ b/changes/2436-PrettyWood.md
@@ -1,1 +1,1 @@
-Support properly `Enum` when combined with generic models
+Avoid `RecursionError` when using some types like `Enum` or `Literal` with generic models
--- a/tests/test_generics.py
+++ b/tests/test_generics.py
@@ -3,6 +3,7 @@ from enum import Enum
 from typing import Any, Callable, ClassVar, Dict, Generic, List, Optional, Sequence, Tuple, Type, TypeVar, Union
 
 import pytest
+from typing_extensions import Literal
 
 from pydantic import BaseModel, Field, ValidationError, root_validator, validator
 from pydantic.generics import GenericModel, _generic_types_cache, iter_contained_typevars, replace_types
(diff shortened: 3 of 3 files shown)


pydantic/generics.py
```diff
@@ -211,7 +212,7 @@ def iter_contained_typevars(v: Any) -> Iterator[TypeVarType]:
         yield v
     elif hasattr(v, '__parameters__') and not get_origin(v) and lenient_issubclass(v, GenericModel):
         yield from v.__parameters__
-    elif isinstance(v, Iterable):
+    elif isinstance(v, Iterable) and not lenient_issubclass(v, Enum):
```
PrettyWood, line 215: @daviskirk WDYT? Do you think we should improve the check even more to handle other cases?

daviskirk: I was almost sure I would miss something here and this is a good example.

The trick here is that the Type itself has to be an iterable for this to fail.  So an Enum class is a good example because it's a weird hybrid of a normal class and an object/singleton kind of thing.  

There might be other classes like this but I don't know them (and to be honest I didn't even know that the enum class was in iterable, so that's a cool new trick I learned).

So I'm fine with this the way it is unless we find a bunch of other examples at which point we might want to think about how to rework this check.

PrettyWood: Perfect! I had the exact same thought 👍

samuelcolvin: Humm, given the risk of infinite recursion, I'd rather whitelist than blacklist here.

The following seems to pass tests, can either of you think of a case where this might break/change behaviour? If so can we easily add that case to the `isinstance` whitelist too?

```py
DictValues: Type[Any] = {}.values().__class__


def iter_contained_typevars(v: Any) -> Iterator[TypeVarType]:
    """Recursively iterate through all subtypes and type args of `v` and yield any typevars that are found."""
    if isinstance(v, TypeVar):
        yield v
    elif hasattr(v, '__parameters__') and not get_origin(v) and lenient_issubclass(v, GenericModel):
        yield from v.__parameters__
    elif isinstance(v, (DictValues, list)):
        ...
```

daviskirk: I think the only reason for DictValues is the `typevars.values()` in line 110, so I'd be fine with changing that to `list(typevars.values())` there and just have the list check here.

PrettyWood: Lunch break is over. Feel free to modify this PR. Otherwise I'll do the change tonight

samuelcolvin:
LGTM, @PrettyWood & @daviskirk I'll merge this soon if you're happy with it.

PrettyWood:
Thanks @samuelcolvin !

samuelcolvin:
We should add a test for a literal too, just so we don't forget.

daviskirk:
Looks great to me thanks for fixing my oversights, just tried it out with the literal, works fine as well!

```python
@skip_36
def test_generic_literal():
    T = TypeVar('T')

    class SomeGenericModel(GenericModel, Generic[T]):
        some_field: T

    class MyModel(BaseModel):
        my_gen: SomeGenericModel[Literal["foo"]]

    m = MyModel.parse_obj({'my_gen': {'some_field': "foo"}})
    assert m.my_gen.some_field == "foo"

    with pytest.raises(ValidationError, match=".*permitted: 'foo'.*"):
        MyModel.parse_obj({'my_gen': {'some_field': "bar"}})
```

samuelcolvin:
🙏 thanks everyone.

PR #2463, opened 2021-03-03
Add GetterDict example

### LINKED ISSUE #1745: Document GetterDict for ORM mode properly properly
As per #1726 and #1689 we need to document `GetterDict` and how to use it properly.

nuno-andre:
Here is a very simple example of how to subclass `GetterDict` to process a class with no mapping protocol while dealing with the impedance mismatch between models (DOM's attributes and children vs OOP's properties).

I have added three optional fields to show "missing attribute", "missing children", and "missing children attribute".

closes #1745

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -169,7 +169,7 @@ You can also customise class validation using [root_validators](validators.md#ro
 In this case your validator function will be passed a `GetterDict` instance which you may copy and modify.
 
 The `GetterDict` instance will be called for each field with a sentinel as a fallback (if no other default
-value is set). Returning this sentinel (or `object()`) means that the field is missing. Any other value will
+value is set). Returning this sentinel means that the field is missing. Any other value will
 be interpreted as the value of the field.
 
 ```py
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -157,13 +157,26 @@ Here a vanilla class is used to demonstrate the principle, but any ORM class cou
 ```
 _(This script is complete, it should run "as is")_
 
-Arbitrary classes are processed by *pydantic* using the `GetterDict` class
-(see [utils.py](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/utils.py)), which attempts to
+
+### Data binding
+
+Arbitrary classes are processed by *pydantic* using the `GetterDict` class (see
+[utils.py](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/utils.py)), which attempts to
 provide a dictionary-like interface to any class. You can customise how this works by setting your own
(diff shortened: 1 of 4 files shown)


nuno-andre:
I've also documented the `_missing` sentinel behavior. But I'm not quite sure that's the intended way to raise missing field errors, as the `GetterDict` code seems to suggest that `KeyError` should do so (which would make a lot of sense IMO). However, this is not the case.

Could you please confirm this assumption?

PrettyWood:
@nuno-andre Thanks a lot for this doc, it was needed!
And good remark for the `KeyError` thing, which would indeed make sense.
But for now better stick with what you did and maybe open a discussion to suggest a refacto?

nuno-andre:
Of course, just wanted to confirm that I hadn't documented an unintended behavior or a regression.

PrettyWood:
@nuno-andre can you add a change file please? :)

nuno-andre:
Done! I've just realized I'd completely ignored the PR template. Sorry for that.

docs/usage/models.md
```diff
@@ -157,14 +157,27 @@ Here a vanilla class is used to demonstrate the principle, but any ORM class cou
...
+Arbitrary classes are processed by *pydantic* using the `GetterDict` class (see
+[utils.py](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/utils.py)), which attempts to
 provide a dictionary-like interface to any class. You can customise how this works by setting your own
 sub-class of `GetterDict` as the value of `Config.getter_dict` (see [config](model_config.md)).
 
 You can also customise class validation using [root_validators](validators.md#root-validators) with `pre=True`. 
 In this case your validator function will be passed a `GetterDict` instance which you may copy and modify.
 
+The `GetterDict` instance will be called for each field with a sentinel as a fallback (if no other default
+value is set). Returning this sentinel (or `object()`) means that the field is missing. Any other value will
```
samuelcolvin, line 172:
```suggestion
value is set). Returning this sentinel means that the field is missing. Any other value will
```

Unless I'm missing something, you can't just return `object()` as that will give a new unique value which won't match `default`

samuelcolvin (commented):
otherwise looks good.

samuelcolvin:
please update, otherwise I'll accept that one tiny change some time soon and merge.

docs/usage/models.md
nuno-andre: Yes, that's true. So strange, I remember checking it. I probably got mistaken on some test.

nuno-andre:
I guess this CI error isn't related to the PR, right?

samuelcolvin:
no, that was just a problem with that CI script.

PR #2475, opened 2021-03-04
If failed to parse datetime/time then tell what was type received

mas15:
Would be nicer if this message included what type was provided

## Checklist


PrettyWood (commented):
I think it should be done in a more global approach with the context. Decide whether or not the value should be displayed, the type, or any other information.
Adding this info on just one validator feels weird.

samuelcolvin:
See #1421.

This will hopefully be done in a global way at some point in the future, but it will make things more complicated to do it in an ad hoc way first.

samuelcolvin:
Thanks for the contribution though, I know it's annoying when seemingly helpful PRs get closed like this.

PR #2375, opened 2021-02-17
Advanced use of generic fields in schema generation

maximberg:
## Change Summary

Continuing #2262, I started to develop the use of generics as field types in models.
My initial implementation of this functionality was rather naive. I made a revision.
In particular, I changed how `modify_schema` is called for generics and how the nested type schema is generated.

For example, you are using a generic type `MyGeneric[MyCustomType, str]`.
You expect the data to be passed to validation as a Tuple, and a schema must be defined for each nested type and for the entire field.

The sequence of forming the scheme should be, as it seems to me, like this:
1. Collect all nested type schemas into one `allOf` schema.
2. Call `modify_schema` of the generic `MyGeneric`, to where we pass the schema collected at the previous stage. There it can be processed.
3. Update the schema described in the model itself with the data collected in the previous stage.

Moreover, there are two options for nested types - if one type is specified or if several are specified.

For one type, the schema will look like this:
`MyGeneric[MyCustomType]` -> `{"allOf": [{"$ref": "#/definitions/MyCustomType"}]}`

For two nested types like this:
`MyGeneric[MyCustomType, str]` -> `{"allOf": [{"type": "array", "items": [{"$ref": "#/definitions/CustomType"}, {"type": "string "}]}]}`

## Checklist

* [x] Unit tests for the changes exist
* [x] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable
* [x] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)

PR #2497, opened 2021-03-08
feat(schema): set `minItems` and `maxItems` in JSON schema for tuples

### LINKED ISSUE #2496: Empty tuple type does not generate valid JSON schema
### Checks

# Bug

```
             pydantic version: 1.8.1
            pydantic compiled: True
                 install path: /Users/smacpher/.virtualenvs/nucleus/lib/python3.6/site-packages/pydantic
               python version: 3.6.9 (default, Oct 16 2019, 11:14:39)  [GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.10.44.4)]
                     platform: Darwin-17.7.0-x86_64-i386-64bit
```

```py
import jsonschema
import pydantic

from typing import Tuple

    f: Tuple[()]

jsonschema.Draft7Validator.check_schema(Model.schema())
```

This fails with:

```
SchemaError: [] is not valid under any of the given schemas

Failed validating 'anyOf' in metaschema['properties']['properties']['additionalProperties']['properties']['items']:
    {'anyOf': [{'$ref': '#'}, {'$ref': '#/definitions/schemaArray'}],
     'default': True}

On schema['properties']['f']['items']:
    []
```

The generated schema is:

```py
{'title': 'Model',
 'type': 'object',
 'properties': {'f': {'title': 'F', 'type': 'array', 'items': []}},
 'required': ['f']}
```

Is this expected behavior? Would the desired behavior be to set `maxItems: 0` and remove `items: []` like such (that would make the generated schema valid)?

```
{'title': 'Model',
 'type': 'object',
 'properties': {'f': {'title': 'F', 'type': 'array', 'maxItems': 0}},
 'required': ['f']}
```

BTW, glad to see that empty tuples (`Tuple[()]`) are now supported (https://github.com/samuelcolvin/pydantic/issues/2318)! Thank you 🙏 .

PrettyWood:
## Change Summary
We didn't enforce the length of the tuple in the JSON schema. We could hence set smaller or longer tuples.
Also handles case of empty tuple by not setting `items` at all to fix #2496 

## Checklist

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -490,15 +490,17 @@ def field_type_schema(
             definitions.update(sf_definitions)
             nested_models.update(sf_nested_models)
             sub_schema.append(sf_schema)
-        if len(sub_schema) == 1:
-            if field.shape == SHAPE_GENERIC:
-                f_schema = sub_schema[0]
-            else:
-                f_schema = {'type': 'array', 'items': sub_schema[0]}
-        else:
-            f_schema = {'type': 'array', 'items': sub_schema}
+
         if field.shape == SHAPE_GENERIC:
+            f_schema = sub_schema[0] if len(sub_schema) == 1 else {'type': 'array', 'items': sub_schema}
             f_schema = {'allOf': [f_schema]}
+        else:
+            tuple_len = len(sub_fields)
+            f_schema = {'type': 'array', 'minItems': tuple_len, 'maxItems': tuple_len}
+            if tuple_len == 1:
+                f_schema['items'] = sub_schema[0]
+            elif tuple_len > 1:
+                f_schema['items'] = sub_schema
     else:
         assert field.shape in {SHAPE_SINGLETON, SHAPE_GENERIC}, field.shape
         f_schema, f_definitions, f_nested_models = field_singleton_schema(
@@ -835,7 +837,15 @@ def field_singleton_schema(  # noqa: C901 (ignore complexity)
             ref_template=ref_template,
             known_models=known_models,
         )
-        f_schema.update({'type': 'array', 'items': list(sub_schema['properties'].values())})
+        items_schemas = list(sub_schema['properties'].values())
+        f_schema.update(
+            {
+                'type': 'array',
+                'items': items_schemas,
+                'minItems': len(items_schemas),
+                'maxItems': len(items_schemas),
+            }
+        )
     elif not hasattr(field_type, '__pydantic_model__'):
         add_field_type_to_schema(field_type, f_schema)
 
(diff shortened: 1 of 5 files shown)


tests/test_schema.py
```diff
@@ -521,34 +521,36 @@ class Model(BaseModel):
...
+                    {'anyOf': [{'type': 'string'}, {'type': 'integer'}, {'type': 'number'}]},
+                    {'type': 'number'},
+                ],
+                'minItems': 4,
+                'maxItems': 4,
+            },
         ),
-        (Tuple[str], {'type': 'string'}),
+        (Tuple[str], {'items': {'type': 'string'}, 'minItems': 1, 'maxItems': 1}),
+        (Tuple[()], {'maxItems': 0, 'minItems': 0}),
```
smacpher-myst, line 541: Great! 👍

smacpher-myst (approved):

pydantic/schema.py
```diff
@@ -487,9 +489,12 @@ def field_type_schema(
             definitions.update(sf_definitions)
             nested_models.update(sf_nested_models)
             sub_schema.append(sf_schema)
-        if len(sub_schema) == 1:
-            sub_schema = sub_schema[0]  # type: ignore
-        f_schema = {'type': 'array', 'items': sub_schema}
+
+        f_schema = {'type': 'array', 'minItems': tuple_len, 'maxItems': tuple_len}
+        if tuple_len == 1:
+            f_schema['items'] = sub_schema[0]
```
samuelcolvin, line 495: this looks weird, are you sure it's correct as per JSONSchema.

Surely even if there's only one item, `items` should be an array?

samuelcolvin (commented):
This seems good but it will conflict with #2375 which I know @maximberg has been working on for a long time.

If someone can explain to my small brain what #2375 does, I think it would be fair to merge that first.

samuelcolvin:
please update

pydantic/schema.py
PrettyWood: It's valid and more common to directly put the type like [here](https://json-schema.org/understanding-json-schema/reference/array.html#list-validation)
```json
{
  "type": "array",
  "items": {
    "type": "number"
  }
}
```
when the type is the same for all fields (which is the case if there is only one expected item of course)
But both are valid I don't have a strong opinion on this

PrettyWood:
I'll wait for #2375 before updating this PR. If I should update it before please tell me

PrettyWood:
please review 🙏

samuelcolvin:
Thank you.

Issue #1098, opened 2019-12-15
Strict configuration


maxrothman:
# Feature request: strict mode configuration flag

This issue has been a sensitive one in the past, so I'm trying to tread lightly here. Pydantic's position so far has been that because it's a parsing, not a validation library, fields should coerce their values to the specified type if possible (e.g. a field type-hinted with `float` given `"1"` coerces it to `1`), and that `Strict*` fields (e.g. `StrictFloat`, `StrictBool`, …) are available if users prefer to use them. Requests for a way to make default types behave strictly have been closed due to implementation concerns.

However, the codebase has evolved significantly since #578, #360, and #284 were closed, and I think that some of the earlier difficulties in building this feature are no longer present today.

In #284, the reason given for not including a strict mode config was that pydantic would no longer be able to just call `float(x)` and pass the errors along, and that there are edge cases around bools and ints. But in 4f4e22e, the validator for `float` types became `validators.float_validator()` rather than `float()`, `strict_int_validator()` was added in 1b467da and distinguishes between bools and ints, and on master, `float_validator()` actually does _extra work_ compared to `strict_float_validator()` to accept non-float values.

As far as I can tell, the only thing required to implement this flag now would be to build `validators._VALIDATORS` conditionally for each model based on a config value, using the `strict_*` validators rather than the standard ones when the flag was set.

There are several different ways this feature could be implemented: as a global flag, as another property on the model config, etc. I don't have strong opinions on this topic, since in my code I would probably enable this configuration universally.

Thoughts? The lack of this feature is the only thing keeping me from strongly recommending the use of Pydantic in my workplace, and it's clearly a feature that others are interested in having as well, as evidenced by the prior requests. I would be interested in contributing a PR for this change if you're interested in pursuing it.

dmontagu:
Could you describe the use case you have in mind for this feature? I'm trying to understand whether your goal is closer to using it for runtime type-checking when instantiating `BaseModel` subclasses, or if it is specific to the parsing process (e.g., you always want to raise errors when parsing a json payload containing `"1.0"` instead of `1.0`).

-----

I would be in favor of supporting this if we could come up with a simple implementation with minimal impact on existing logic. I can't tell right now from just thinking about it whether I think the approach you have described would meet the bar in terms of simplicity; @samuelcolvin might have different opinions.

-------

[...]

PR #2509, opened 2021-03-11
add strict_int config

amirfru:
## Change Summary

Add a global option `strict_int` to treat all `int` fields as strict

## Checklist

### DIFF
--- a/changes/2509-amirfru.md
+++ b/changes/2509-amirfru.md
@@ -1,0 +1,1 @@
+Added **`strict_int`** option to config
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -116,6 +116,9 @@ _(This script is complete, it should run "as is")_
 **`copy_on_model_validation`**
 : whether or not inherited models used as fields should be reconstructed (copied) on validation instead of being kept untouched (default: `True`)
 
+**`strict_int`**
+: whether to treat `int` fields as strict.
+
 ## Change behaviour globally
 
 If you wish to change the behaviour of _pydantic_ globally, you can create your own custom `BaseModel`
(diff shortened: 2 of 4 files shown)


PrettyWood:
Where is the discussion? Or the linked PR welcome? Why not just use `StrictInt`? What about https://github.com/samuelcolvin/pydantic/issues/1098?
This PR will probably be closed

amirfru:
Hi @PrettyWood, yes it is very-much related to #1098 , I really like this project but I really prefer not to add `StrictInt` (or any typing for that matter) if not needed. i.e. I have many configurations and delta-configuration files and I do not want to go over every single field and decide if it is an `int` or should actually be `float` but someone forgot to put a dot after the number.
Just want to let Pydantic to infer it by the initial input.
and I think that when someone tries to put a `float` in an `int` fields, it should raise an exception, this is the new added option.

amirfru:
its just a micro PR which will make Pydantic much better for me

amirfru:
Meanwhile I'm using [this](https://gist.github.com/samuelcolvin/714e75a74284696204de390a121b0da0) trick in-order to get roughly the same desired behavior

PrettyWood:
> its just a micro PR which will make Pydantic much better for me

@amirfru Believe me a lot of people say this, me included 😄
But the goal of a library is to try to stay consistent and avoid having plenty of different settings that make it harder and harder to use but also maintain.
Tomorrow someone will come with a `strict_float` request and another one with `strict_str`...
This is why I reckon we should avoid "breaking the first window" and tackle the whole problem or not at all.
If a "strict mode" is added, we can probably make it configurable only for some types or some fields.
Your inputs on the "strict mode" issue are more than welcome to ensure the implementation will suit your needs 👍

samuelcolvin:
@PrettyWood is correct here.

I'm afraid this will have to wait for a proper strict mode.

PR #2519, opened 2021-03-13
Allow collections.abc.Callable to be used as type in python 3.9

### LINKED ISSUE #2465: python 3.9 generic alias Callable from collections module does not behave correctly in generics
# Bug

```
             pydantic version: 1.8
            pydantic compiled: False
                 install path: /home/user/Documents/Code/pydantic/pydantic
               python version: 3.9.1 (default, Dec 11 2020, 14:32:07)  [GCC 7.3.0]
                     platform: Linux-5.4.0-66-generic-x86_64-with-glibc2.31
```

```
from typing import Any, Callable, ClassVar, Dict, Generic, List, Literal, Optional, Sequence, Tuple, Type, TypeVar, Union
```

For example, the `test_iter_contained_typevars` test with replacing `typing.Callable` with `collections.abc.Callable`:

```py
from collections.abc import Callable
from typing import Dict, Generic, List, Optional, Union

@skip_36
def test_iter_contained_typevars():
    T = TypeVar('T')
    T2 = TypeVar('T2')

    class Model(GenericModel, Generic[T]):
        a: T

    assert list(iter_contained_typevars(Optional[List[Union[str, Model[T], Callable[[T2, T], str]]]])) == [T, T2, T]
```
results in 

```
----> 1 assert list(iter_contained_typevars(Optional[List[Union[str, Model[T], Callable[[T2, T], str]]]])) == [T, T2, T]

pydantic-39/lib/python3.9/typing.py in inner(*args, **kwds)
    260             except TypeError:
    261                 pass  # All real errors (not unhashable args) are raised below.
--> 262             return func(*args, **kwds)
    263         return inner
    264 

pydantic-39/lib/python3.9/typing.py in __getitem__(self, parameters)
    337     @_tp_cache
    338     def __getitem__(self, parameters):
--> 339         return self._getitem(self, parameters)
    340 
    341 

pydantic-39/lib/python3.9/typing.py in Union(self, parameters)
    449     msg = "Union[arg, ...]: each arg must be a type."
    450     parameters = tuple(_type_check(p, msg) for p in parameters)
--> 451     parameters = _remove_dups_flatten(parameters)
    452     if len(parameters) == 1:
    453         return parameters[0]

pydantic-39/lib/python3.9/typing.py in _remove_dups_flatten(parameters)
    229             params.append(p)
    230 
--> 231     return tuple(_deduplicate(params))
    232 
    233 

pydantic-39/lib/python3.9/typing.py in _deduplicate(params)
    203 def _deduplicate(params):
    204     # Weed out strict duplicates, preserving the first of each occurrence.
--> 205     all_params = set(params)
    206     if len(all_params) < len(params):
    207         new_params = []

TypeError: unhashable type: 'list'
```

This came up while reading the release notes for python 3.10 where this is changed back: https://docs.python.org/3.10/whatsnew/3.10.html#collections-abc 
So in 3.10 this problem won't be present.

However, since I don't think we're testing any of the "collection" module types particularly well, the assumption is that there might be other problems as well.
We should therefore probably add some tests that check these python >= 3.9 features / generic alias cases in case they are missing.

daviskirk:
## Change Summary

Allows using python 3.9 generic alias version of `Callable` (`collections.abc.Callable`)

Fixes #2465

## Related issue number

Address #2465

## Checklist

### DIFF
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -248,6 +248,11 @@ NONE_TYPES: Tuple[Any, Any, Any] = (None, NoneType, Literal[None])
 
 
 if sys.version_info < (3, 8):  # noqa: C901 (ignore complexity)
+    # Even though this implementation is slower, we need it for python 3.6/3.7:
+    # In python 3.6/3.7 "Literal" is not a builtin type and uses a different
+    # mechanism.
+    # for this reason `Literal[None] is Literal[None]` evaluates to `False`,
+    # breaking the faster implementation used for the other python versions.
 
     def is_none_type(type_: Any) -> bool:
         return type_ in NONE_TYPES
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -245,14 +245,14 @@ __all__ = (
 NoneType = None.__class__
 
 
-NONE_TYPES: Tuple[Any, ...] = (None, NoneType, Literal[None])
+NONE_TYPES: Set[Any] = {None, NoneType, Literal[None]}
 
 
 def is_none_type(type_: Any) -> bool:
-    for none_type in NONE_TYPES:
-        if type_ is none_type:
-            return True
-    return False
+    try:
+        return type_ in NONE_TYPES
+    except TypeError:
+        return False
 
 
 def display_as_type(v: Type[Any]) -> str:
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -211,6 +211,7 @@ __all__ = (
     'NoArgAnyCallable',
     'NoneType',
     'NONE_TYPES',
+    'is_none_type',
     'display_as_type',
     'resolve_annotations',
     'is_callable_type',
@@ -242,7 +243,16 @@ __all__ = (
 
 
 NoneType = None.__class__
-NONE_TYPES: Set[Any] = {None, NoneType, Literal[None]}
+
+
+NONE_TYPES: Tuple[Any, ...] = (None, NoneType, Literal[None])
+
+
+def is_none_type(type_: Any) -> bool:
+    for none_type in NONE_TYPES:
+        if type_ is none_type:
+            return True
+    return False
 
 
 def display_as_type(v: Type[Any]) -> str:
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -250,10 +250,7 @@ NONE_TYPES: Tuple[Any, Any, Any] = (None, NoneType, Literal[None])
 if sys.version_info < (3, 8):  # noqa: C901 (ignore complexity)
 
     def is_none_type(type_: Any) -> bool:
-        try:
-            return type_ in NONE_TYPES
-        except TypeError:
-            return False
+        return type_ in NONE_TYPES
 
 
 else:
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -210,7 +210,6 @@ __all__ = (
     'AnyCallable',
     'NoArgAnyCallable',
     'NoneType',
-    'NONE_TYPES',
     'is_none_type',
     'display_as_type',
     'resolve_annotations',
@@ -245,13 +244,24 @@ __all__ = (
 NoneType = None.__class__
 
 
-NONE_TYPES: Set[Any] = {None, NoneType, Literal[None]}
+NONE_TYPES: Tuple[Any, Any, Any] = (None, NoneType, Literal[None])
 
 
-def is_none_type(type_: Any) -> bool:
-    try:
-        return type_ in NONE_TYPES
-    except TypeError:
+if sys.version_info < (3, 8):  # noqa: C901 (ignore complexity)
+
+    def is_none_type(type_: Any) -> bool:
+        try:
+            return type_ in NONE_TYPES
+        except TypeError:
+            return False
+
+
+else:
+
+    def is_none_type(type_: Any) -> bool:
+        for none_type in NONE_TYPES:
+            if type_ is none_type:
+                return True
         return False
 
 
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -656,9 +656,12 @@ def find_validators(  # noqa: C901 (ignore complexity)
     type_type = type_.__class__
     if type_type == ForwardRef or type_type == TypeVar:
         return
-    if type_ in NONE_TYPES:
-        yield none_validator
-        return
+    try:
+        if type_ in NONE_TYPES:
+            yield none_validator
+            return
+    except TypeError:
+        pass  # in case unhashable types are in the type
     if type_ is Pattern:
         yield pattern_validator
         return
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -30,7 +30,6 @@ from typing_extensions import Literal
 from . import errors
 from .datetime_parse import parse_date, parse_datetime, parse_duration, parse_time
 from .typing import (
-    NONE_TYPES,
     AnyCallable,
     ForwardRef,
     all_literal_values,
@@ -39,6 +38,7 @@ from .typing import (
     is_callable_type,
     is_literal_type,
     is_namedtuple,
+    is_none_type,
     is_typeddict,
 )
 from .utils import almost_equal_floats, lenient_issubclass, sequence_like
@@ -656,12 +656,10 @@ def find_validators(  # noqa: C901 (ignore complexity)
     type_type = type_.__class__
     if type_type == ForwardRef or type_type == TypeVar:
         return
-    try:
-        if type_ in NONE_TYPES:
-            yield none_validator
-            return
-    except TypeError:
-        pass  # in case unhashable types are in the type
+
+    if is_none_type(type_):
+        yield none_validator
+        return
     if type_ is Pattern:
         yield pattern_validator
         return
--- a/tests/test_callable.py
+++ b/tests/test_callable.py
@@ -1,11 +1,18 @@
+import sys
 from typing import Callable
 
 import pytest
 
 from pydantic import BaseModel, ValidationError
 
+collection_callable_types = [Callable, Callable[[int], int]]
+if sys.version_info >= (3, 9):
+    from collections.abc import Callable as CollectionsCallable
 
-@pytest.mark.parametrize('annotation', [Callable, Callable[[int], int]])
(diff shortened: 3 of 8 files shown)


tests/test_callable.py
```diff
@@ -1,11 +1,18 @@
...
 import pytest
 
 from pydantic import BaseModel, ValidationError
 
+collection_callable_types = [Callable, Callable[[int], int]]
+if sys.version_info >= (3, 9):
+    from collections.abc import Callable as CollectionsCallable
 
-@pytest.mark.parametrize('annotation', [Callable, Callable[[int], int]])
+    collection_callable_types += [CollectionsCallable, CollectionsCallable[[int], int]]
```
daviskirk, line 12: We could do this with all other generic aliases (Iterable, Sequence and so on).
However, it probably makes some of the tests less readable and I'm unsure on how much sense that would make:  As far as I can tell the "typing" versions are deprecated starting 3.9, so maybe a better approach would be to just to alias them to the "new" generic alias versions for python versions >= 3.9 (except for the cases where they have obviously different behavior like the `Callable` here).  
The old typing versions would still be tested in the old python versions so that should cover almost all cases.

pydantic/validators.py
```diff
@@ -656,9 +656,12 @@ def find_validators(  # noqa: C901 (ignore complexity)
     type_type = type_.__class__
     if type_type == ForwardRef or type_type == TypeVar:
         return
-    if type_ in NONE_TYPES:
-        yield none_validator
-        return
+    try:
+        if type_ in NONE_TYPES:
```
PrettyWood, line 660: I wonder if we should not write a custom set
```py
class IdentitySet(set):
    def __contains__(self, x):
        return any(x is v for v in self)
```
and have
```py
NONE_TYPES = IdentitySet((None, NoneType, Literal[None]))
```
to support all the
```py
whatever in NONE_TYPES
```
instead. @davidolrik WDYT?

daviskirk: Yes, checking this correctly everywhere is a good idea. I tried a few variations and the custom class slowed down the benchmarks even in cython builds for me (I think this part is on the benchmark path).

I ended up just having a simple function that checks by identity, which does not seem to impact the cython benchmarks (At least I switched back and forth between the master and the new version, `make build`, `make benchmark-pydantic` and couldn't see any noticable impact.

In pure python, there might be a slight advantage to still using the try/except since the except is hardly ever called (only for non-hashable types which isn't the case very often, might even only be callables).
But testing using cython that advantage does not exist (cython probably can't optimize the try/except).
Here are a few timings of JUST the function (not pydantic benchmarks).

Using tuple and identity comparison (this is how it's implemented now)

```
In [10]: %%cython
    ...: from typing import Any, Literal
    ...: NoneType = None.__class__
    ...: 
    ...: 
    ...: NONE_TYPES = (None, NoneType, Literal[None])
    ...: 
    ...: 
    ...: def is_none_type(type_: Any) -> bool:
    ...:     for none_type in NONE_TYPES:
    ...:         if type_ is none_type:
    ...:             return True
    ...:     return False
    ...: 

In [11]: %timeit is_none_type(None)
30.2 ns ± 0.574 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [12]: %timeit is_none_type(6)
33.9 ns ± 0.262 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [13]: %timeit is_none_type({})
50.1 ns ± 11.8 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Using a set with try except:

```
In [6]: %%cython
   ...: from typing import Any, Literal
   ...: NoneType = None.__class__
   ...: 
   ...: 
   ...: NONE_TYPES: Set[Any] = {None, NoneType, Literal[None]}
   ...: 
   ...: 
   ...: def is_none_type(type_: Any) -> bool:
   ...:     try:
   ...:         return type_ in NONE_TYPES
   ...:     except TypeError:
   ...:         return False
   ...: 
   ...: 

In [7]: %timeit is_none_type(None)
50 ns ± 12.9 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [8]: %timeit is_none_type(6)
40.9 ns ± 0.546 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [9]: %timeit is_none_type({})
383 ns ± 121 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

For comparison: direct set "contains" (as it was before, but can't handle types like callable and so on), here we probably just save the function call:

```
In [14]: NONE_TYPES: Set[Any] = {None, NoneType, Literal[None]}

In [15]: %timeit None in NONE_TYPES
23.7 ns ± 0.201 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [16]: %timeit 6 in NONE_TYPES
25 ns ± 0.0713 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

daviskirk:
FYI, the fast api tests fail because sqlalchemy 1.4 was released and I'm assumig fastapi didn't pin their test dependencies

PrettyWood:
Thanks! Yep the little `is_none_type` is perfect thanks!
As for the CI I opened https://github.com/encode/databases/pull/299 but still need to work on it

PrettyWood (approved):
:+1:

pydantic/typing.py
```diff
@@ -199,6 +199,7 @@ def get_args(tp: Type[Any]) -> Tuple[Any, ...]:
     'NoArgAnyCallable',
     'NoneType',
     'NONE_TYPES',
```
samuelcolvin, line 201: Can we now remove `NONE_TYPES` here?

pydantic/typing.py
```diff
@@ -229,9 +230,18 @@ def get_args(tp: Type[Any]) -> Tuple[Any, ...]:
...
+
+
 NONE_TYPES: Set[Any] = {None, NoneType, Literal[None]}
 
 
+def is_none_type(type_: Any) -> bool:
+    try:
+        return type_ in NONE_TYPES
+    except TypeError:
+        return False
```
samuelcolvin, line 242: From your benchmarks (and mine), the following seems to be faster:

```py
NONE_TYPES: Tuple[Any, Any, Any] = (None, NoneType, Literal[None])


    for none_type in NONE_TYPES:
        if type_ is none_type:
            return True
    return False
```

samuelcolvin: my benchmarks:

```python
In [1]: %load_ext cython

In [2]: %%cython
   ...: from typing import Any, Literal, Tuple, Set
   ...: NoneType = None.__class__
   ...: 
   ...: 
   ...: NONE_TYPES_TUPLE: Tuple[Any, Any, Any] = (None, NoneType, Literal[None])
   ...: 
   ...: 
   ...: def is_none_type1(type_: Any) -> bool:
   ...:     for none_type in NONE_TYPES_TUPLE:
   ...:         if type_ is none_type:
   ...:             return True
   ...:     return False
   ...: 
   ...: 
   ...: NONE_TYPES_SET: Set[Any] = {None, NoneType, Literal[None]}
   ...: 
   ...: 
   ...: def is_none_type2(type_: Any) -> bool:
   ...:     try:
   ...:         return type_ in NONE_TYPES_SET
   ...:     except TypeError:
   ...:         return False
   ...: 
   ...: 
   ...: def is_none_type3(type_: Any) -> bool:
   ...:     return type_ is None or type_ is NoneType or type_ is Literal[None]
   ...: 

In [3]: print('None:')
   ...: %timeit is_none_type1(None)
   ...: %timeit is_none_type2(None)
   ...: %timeit is_none_type3(None)
   ...: 
   ...: print('type(None):')
   ...: %timeit is_none_type1(type(None))
   ...: %timeit is_none_type2(type(None))
   ...: %timeit is_none_type3(type(None))
   ...: 
   ...: print('Literal[None]:')
   ...: %timeit is_none_type1(Literal[None])
   ...: %timeit is_none_type2(Literal[None])
   ...: %timeit is_none_type3(Literal[None])
   ...: 
   ...: print('6:')
   ...: %timeit is_none_type1(6)
   ...: %timeit is_none_type2(6)
   ...: %timeit is_none_type3(6)
   ...: 
   ...: print('{}:')
   ...: %timeit is_none_type1({})
   ...: %timeit is_none_type2({})
   ...: %timeit is_none_type3({})
   ...: 
None:
49.5 ns ± 0.727 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
65.4 ns ± 1.14 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
43.3 ns ± 0.775 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
type(None):
99.5 ns ± 1.52 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
118 ns ± 1.86 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
92.3 ns ± 1.83 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
Literal[None]:
337 ns ± 5.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
1.08 µs ± 16.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
582 ns ± 11.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
6:
52.6 ns ± 0.762 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
64.5 ns ± 1.48 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
313 ns ± 2.64 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
{}:
65 ns ± 1.05 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
458 ns ± 3.09 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
323 ns ± 1.59 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

I have absolutely no idea why `is_none_type3` is so much slower than `is_none_type1`, but it does seem to be.

samuelcolvin (changes requested):
otherwise LGTM

samuelcolvin:
also I noticed coverage is not complete, please update.

pydantic/typing.py
daviskirk: please review

daviskirk:
please review

PrettyWood (approved):

PR #2549, opened 2021-03-19
Improve generic subclass support

### LINKED ISSUE #2007: GenericModel - Recursive abstract class fails validation
### Checks

# Bug

```
             pydantic version: 1.6.1
            pydantic compiled: False
                 install path: /wayfair/home/ch438l/pydantic/pydantic
               python version: 3.6.8 (default, Aug 10 2019, 06:54:07)  [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
                     platform: Linux-3.10.0-862.3.2.el7.x86_64-x86_64-with-centos-7.5.1804-Core
```

If I try to define an attribute on a generic model to be a generic abstract class, then the model will fail to validate. Here's a failing test case:

```py
@skip_36
def test_abstract_generic_type_recursion():
    T = TypeVar('T')

    class BaseInnerClass(GenericModel, abc.ABC, Generic[T]):
        base_data: T

        @abc.abstractmethod
        def base_abstract(self) -> None:
            pass

    class ConcreteInnerClass(BaseInnerClass[T], Generic[T]):

        def base_abstract(self) -> None:
            return None

    class OuterClass(GenericModel, Generic[T]):
        inner_class: BaseInnerClass[T]

    OuterClass[int](inner_class=ConcreteInnerClass[int](base_data=2))
```

It appears to be failing because the `isinstance(value, cls)` check is failing for the generic type.

diabolo-dan:
## Change Summary

Add parameterised subclasses to `__bases__` when constructing new parameterised classes, so that `A <: B => A[int] <: B[int]`.

This helps to facilitate a use case based on abstract classes, and leads to less unexpected behaviour when comparing between concrete versions of classes.

## Related issue number

fix #2007 

## Checklist


tests/test_generics.py
```diff
@@ -1071,3 +1071,82 @@ class GModel(GenericModel, Generic[FieldType, ValueType]):
...
+
+    class A(GenericModel, Generic[T]):
+        ...
+
+    class B(A[T], Generic[T]):
+        ...
+
+    assert B[int].__name__ == 'B[int]'
+    assert issubclass(B[int], B)
+    assert issubclass(B[int], A[int])
```
choogeboom, line 1088: Probably should add a few negative cases to each of these tests. For example

    assert not issubclass(B[str], A[int])

diabolo-dan: Good point, I've added some suitable negative cases.

diabolo-dan:
test fastAPI is failling with:
```ImportError: cannot import name 'RowProxy' from 'sqlalchemy.engine.result' (/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/sqlalchemy/engine/result.py```

As far as I can tell that's not anything I could've incidentally broken.  I'm not sure what the best way to proceed with it would be?

**Comment by PrettyWood** [comment]
@diabolo-dan Don't worry you can have a rest for the CI! ;)
It's due to `fastapi` tests that uses `databases` without pinning `sqlalchemy` dependency.
I already submitted a fix https://github.com/encode/databases/pull/299 for that

File: pydantic/generics.py [hunk 2]
```diff
@@ -140,6 +144,53 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+        e.g. for
+        ```
+        class A(GenericModel, Generic[T]):
+            ...
+
+        class B(A[V], Generic[V]):
+            ...
+
+        ```
+        then: `A[int] in B.__concrete_bases__({V: int})`
```
diabolo-dan, line 168: This got missed in a rename of the method.  Will correct.

pydantic/generics.py
```diff
@@ -140,6 +144,53 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
         params_component = ', '.join(param_names)
         return f'{cls.__name__}[{params_component}]'
 
+    @classmethod
+    def __parameterized_bases__(cls, typevars_map: Mapping[TypeVarType, Type[Any]]) -> Iterator[Type[Any]]:
```
davidhyman, line 148: `Mapping[TypeVarType, Type[Any]]`

could be aliased and reused

pydantic/generics.py
```diff
@@ -140,6 +144,53 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+                    # type is Generic or
+                    # base model is already concrete, and will be included transitively.
+                    continue
+                elif cls in _assigned_parameters and base_model in _assigned_parameters:
+                    # Subclass will be inferred via cls
+                    continue
+                else:
+                    base_parameters = tuple([mapped_types[param] for param in base_model.__parameters__])
+                    result = base_model.__class_getitem__(base_parameters)
+                    if result != base_model and result != cls:
```
davidhyman, line 191: is equality correct/safe here, or perhaps using `is` would be more correct for explicit comparison of classes?

pydantic/generics.py
```diff
@@ -140,6 +144,53 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+                if not hasattr(base_model, '__parameters__') or not base_model.__parameters__:
+                    # type is Generic or
+                    # base model is already concrete, and will be included transitively.
+                    continue
+                elif cls in _assigned_parameters and base_model in _assigned_parameters:
+                    # Subclass will be inferred via cls
+                    continue
+                else:
+                    base_parameters = tuple([mapped_types[param] for param in base_model.__parameters__])
+                    result = base_model.__class_getitem__(base_parameters)
```
davidhyman, line 190: perhaps a comment to explain where we're at now in this `else` block (consistent with the ones above?)

davidhyman (commented):
Some minor thoughts

pydantic/generics.py
diabolo-dan: replaced with is.

pydantic/generics.py
diabolo-dan: aliased to `Parametrization`

pydantic/generics.py
```diff
@@ -140,6 +144,53 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+        ```
+        class A(GenericModel, Generic[T]):
+            ...
+
+        class B(A[V], Generic[V]):
+            ...
+
+        ```
+        then: `A[int] in B.__concrete_bases__({V: int})`
+        ```
```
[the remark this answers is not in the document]
diabolo-dan: done

tests/test_generics.py
```diff
@@ -1071,3 +1071,89 @@ class GModel(GenericModel, Generic[FieldType, ValueType]):
...
+    T = TypeVar('T')
+    S = TypeVar('S')
+
+    class A(GenericModel, Generic[T]):
+        ...
+
+    class B(A[S], Generic[T, S]):
+        ...
+
+    PartialyAppliedB = B[str, T]
```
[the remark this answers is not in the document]
diabolo-dan: fixed

pydantic/generics.py
```diff
@@ -140,6 +144,53 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+            # class is a partially "bound" form of a generic model
+            # We need to determine the mapping for the base_model parameters
+            mapped_types: Mapping[TypeVarType, Type[Any]] = {
+                key: typevars_map.get(value, value) for key, value in _assigned_parameters[cls].items()
+            }
+        else:
+            mapped_types = typevars_map
+        for base_model in cls.__bases__:
+            if issubclass(base_model, GenericModel):
+                if not hasattr(base_model, '__parameters__') or not base_model.__parameters__:
```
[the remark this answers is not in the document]
diabolo-dan: replaced with `getattr`

pydantic/generics.py
```diff
@@ -140,6 +144,53 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+        if cls in _assigned_parameters:
+            # class is a partially "bound" form of a generic model
+            # We need to determine the mapping for the base_model parameters
+            mapped_types: Mapping[TypeVarType, Type[Any]] = {
+                key: typevars_map.get(value, value) for key, value in _assigned_parameters[cls].items()
+            }
+        else:
+            mapped_types = typevars_map
+        for base_model in cls.__bases__:
+            if issubclass(base_model, GenericModel):
```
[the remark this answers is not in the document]
diabolo-dan: done

diabolo-dan:
I notice that there was significant generic implementation done by @dmontagu , including the line:
```
    assert concrete_model.__name__ == 'Model[int, BT][str]'
```
Which is I believe the only concrete claim about the behaviour of reapplication of partially parameterised generic models. (namely extending from the former).

This behaviour adds complexity to establishing parameterised bases, and makes for less  (because a concrete type could be constructed through arbitrary many type variables).  I also think it's less intuitive than always building from the base, (so that `Model[int, BT][str] is Model[int, str]`).  The alternative isn't perfect, because there's still the issue of checking `issubclass(Model[int, BT][str], Model[int, AT])` which for various assumptions on the TypeVars, should ideally hold, but I don't think it's possible to achieve that (at least through setting `__bases__`).

I'm just wondering if there was a particular reason for implementing it this way, or if it was just the simplest thing that met the requirements at the time.

pydantic/generics.py
diabolo-dan: I've reworked it a bit and added some more explanation.  Could you check if it's any clearer now?

pydantic/generics.py
```diff
@@ -140,6 +145,69 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
         params_component = ', '.join(param_names)
         return f'{cls.__name__}[{params_component}]'
 
+    @classmethod
+    def __parameterized_bases__(cls, typevars_map: Parametrization) -> Iterator[Type[Any]]:
+        """Returns unbound bases of cls parameterised to given type variables
+
```
samuelcolvin, line 151:
```suggestion
        """
        Returns unbound bases of cls parameterised to given type variables
```

pydantic/generics.py
```diff
@@ -140,6 +145,69 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+            if parameterized_base is base_model or parameterized_base is cls:
+                # Avoid duplication in MRO
+                return
+            yield parameterized_base
+
+        for base_model in cls.__bases__:
+            if not issubclass(base_model, GenericModel):
+                # not a class that can be meaningfully parameterized
+                continue
+            elif not getattr(base_model, '__parameters__', ()):
```
samuelcolvin, line 186:
```suggestion
            elif not getattr(base_model, '__parameters__', None):
```

pydantic/generics.py
```diff
@@ -140,6 +145,69 @@ def __concrete_name__(cls: Type[Any], params: Tuple[Type[Any], ...]) -> str:
...
+                # So don't attempt to include A[S][int]
+                continue
+            elif cls in _assigned_parameters and base_model not in _assigned_parameters:
+                # cls is partially parameterized, base_model is original generic
+                # e.g.  cls = B[str, T], base_model = B[S, T]
+                # Need to determine the mapping for the base_model parameters
+                mapped_types: Parametrization = {
+                    key: typevars_map.get(value, value) for key, value in _assigned_parameters[cls].items()
+                }
+                yield from build_base_model(base_model, mapped_types)
```
samuelcolvin, line 205: can't these two cases be covered by

```py
elif cls in _assigned_parameters:
    if base_model in _assigned_parameters:
        ...
    else:
        ...
```
?

pydantic/generics.py
```diff
@@ -28,6 +28,9 @@
 GenericModelT = TypeVar('GenericModelT', bound='GenericModel')
 TypeVarType = Any  # since mypy doesn't allow the use of TypeVar as a type
 
+Parametrization = Mapping[TypeVarType, Type[Any]]
+_assigned_parameters: Dict[Type[Any], Parametrization] = {}
```
samuelcolvin, line 32: Might help to explain what this store is for?

Also, should we use weak refs here?

Will this cause issues with pickle?

samuelcolvin (commented):
Honestly, `generics.py` scares the hell out of me. I don't pretend to remember/understand everything in there and I barely ever (never?) use generics myself.

The only consolation is that this isn't part of core validation, just `isinstance` checks. (I think?)

I've got a few notes and questions, but overall, and from what I can see it looks reasonable.

samuelcolvin:
please update.

pydantic/generics.py
diabolo-dan: > Might help to explain what this store is for?
I've added a (hopefully) clarifying comment.

> Also, should we use weak refs here?
Hmmm, It seems like there are some potential, albeit somewhat obscure, use cases for deleting the classes, however there's at least one other non weak reference  in the form of the `_generic_types_cache`, so there's not much point using weak refs here without changing that as well.  I'm happy to do so, but I wouldn't be surprised if there were some other references to the created class somewhere, so it seems questionable as to whether it's worth it.

> Will this cause issues with pickle?
I don't believe so -- pickle needs the class to be statically defined in a module to unpickle, as captured by the existing tests.  That existence implies the existence of the class as a key in `_assigned_parameters`, so I don't think there's any way for those to get out of sync.

diabolo-dan:
please review.

samuelcolvin:
great, thank you.

samuelcolvin:
Just a note for anyone watching this PR, this changes here _I think_ have caused a memory build up, see #3829.

To mitigate this, I've implemented (and will soon merge) #4083 which limits the size of `_generic_types_cache` and `_assigned_parameters_limit` to 1000. These configured by settings `generics._generic_types_cache_limit` and `generics._assigned_parameters_limit`.

Feedback on the PR welcome, but in order to include this fix in v1.9.1 I won't being waiting for feedback before merging - any further changes would need to be in another PR.

PR #2554, opened 2021-03-20
Support user defined generic field types in generic models.

daviskirk:
Add support for user defined generic type fields in generic models for python < 3.9.

Referencing this comment (and the problem described) here: https://github.com/samuelcolvin/pydantic/issues/2465#issuecomment-801497216

## Change Summary

The actual change (the check for `type_.name_`) is not necessary in python >= 3.9 as described in the code comment but is needed for python < 3.9.

## Related issue number

Work on #2465 (see comments)

## Checklist


pydantic/generics.py
```diff
@@ -167,7 +167,12 @@ def replace_types(type_: Any, type_map: Mapping[Any, Any]) -> Any:
             # If all arguments are the same, there is no need to modify the
             # type or create a new object at all
             return type_
-        if origin_type is not None and isinstance(type_, typing_base) and not isinstance(origin_type, typing_base):
+        if (
+            origin_type is not None
+            and isinstance(type_, typing_base)
+            and not isinstance(origin_type, typing_base)
+            and type_._name is not None
```
daviskirk, line 174: @PrettyWood I can imagine that your work described here: https://github.com/samuelcolvin/pydantic/issues/2465#issuecomment-798766083 would also fix this, but perhaps this is a smaller change and faster fix short term.
In addition, I think adding the tests for user defined generic types is probably a good idea regardless, wdyt?

PrettyWood: My work was pure R&D don't worry! It was more for reference. Thanks 👍

tests/conftest.py
```diff
@@ -15,7 +15,7 @@
 except ImportError:
     pytest_plugins = []
 else:
-    pytest_plugins = ['hypothesis.extra.pytestplugin']
+    pytest_plugins = []
```
PrettyWood, line 18: Is it wanted?

pydantic/generics.py
PrettyWood: Shouldn't we use `type_.get('_name')`? Are you sure `type_._name` is really always defined? I'm a bit paranoid with `typing` types as there may be some weird hidden type without this `_name` attribute 😅

daviskirk: I agree,  changed this

tests/conftest.py
daviskirk: no, I have to find a better way of doing this, for some reason this plugin loads twice and then crashes pytest on my machine, but it should definitely not be commited... second time I did this by accident, sorry.

daviskirk:
Rebased on master to fix tests

samuelcolvin:
thanks so much.

PR #2609, opened 2021-03-30
feat: support generic types and | operator

PrettyWood:
## Change Summary
Now this is possible with python 3.7+
```py

class User(BaseModel, validate_assignment=True):
    first_name: int | str
    last_name: str | None = None
    friends: list[User] = Field(default_factory=list)
    metadata: dict[str, int] | None = None

user = User(first_name='pika', last_name=666)

assert user.last_name == '666'
assert user.friends == []
user.metadata = {b'1': b'1'}
assert m.user.metadata == {'1': 1}
```

## Checklist

### DIFF
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -288,7 +288,7 @@ def resolve_annotations(raw_annotations: Dict[str, Type[Any]], module_name: Opti
         except NameError:
             # this is ok, it can be fixed with update_forward_refs
             pass
-        except TypeError:
+        except TypeError as te:
             if value.__class__ is not ForwardRef:
                 raise
 
@@ -304,7 +304,7 @@ def resolve_annotations(raw_annotations: Dict[str, Type[Any]], module_name: Opti
                     'you need to install `future-typing`',
                     UserWarning,
                 )
-                raise
+                raise te
             else:
                 import typing
 
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -295,7 +295,7 @@ def resolve_annotations(raw_annotations: Dict[str, Type[Any]], module_name: Opti
             # In case of `ForwardRef`, we give it another shot if we want to leverage "new" annotations
             # with older versions of python
             try:
-                from future_typing import TYPING_NAME, transform_annotation
+                from future_typing import transform_annotation
             except ImportError:
                 import warnings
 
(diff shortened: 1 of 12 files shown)


samuelcolvin:
I'm assuming this is still a work in progress? let me know when you want me to review it.

(please update (just for the bot to assign you 😄))

setup.py
```diff
@@ -133,6 +133,7 @@ def extra(self):
     extras_require={
         'email': ['email-validator>=1.0.3'],
         'dotenv': ['python-dotenv>=0.10.4'],
+        'future': ['future-typing>=0.4.0'],
```
samuelcolvin, line 136: If we go with this, I think we'll need a more clear name for this extra.

samuelcolvin (commented):
I've just had a quick read of this, it looks great to me.

pydantic/typing.py
```diff
@@ -288,6 +288,34 @@ def resolve_annotations(raw_annotations: Dict[str, Type[Any]], module_name: Opti
...
+                warnings.warn(
+                    'If you try to use generic builtins or new `|` Union operator, '
+                    'you need to install `pydantic[future-annotations]` or directly `future-typing`.',
+                    UserWarning,
+                )
+                raise te
+            else:
+                # We need a merged version of typing (with typing_extensions) to
+                # ensure we can resolve things like `Literal` or `Annotated` with older versions
+                from . import typing_merge
```
PrettyWood, line 311: At first I did something like
```py
all_typing_mod_name = 'pydantic_all_typing___'
all_typing_mod = types.ModuleType(all_typing_mod_name)
exec("from typing import *\nfrom typing_extensions import *", all_typing_mod.__dict__)
```
But since using `eval()` is problematic with `cython`, I went with this approach.

PrettyWood:
I wrote some documentation. Feedback welcome! Please review

PrettyWood:
I'd rather not add extra dependency if not needed.
Since PEP 563 is no more, I would wait for PEP 649 before trying to support futures than may not be real futures 😄

PrettyWood:
Let's focus on 3.10 and see if people actually want this.
With the switch from PEP 563 to 649, I don't know how it will affect the __future__.annotations.
And supporting only generics and unions but not some types or forwardrefs may create more support.
I tag this PR as deferred in the meantime

elyase:
I was surprised to see that pydantic doesn`t yet support the  `| operator`. Would love to have this merged.

tuchandra:
> Let's focus on 3.10 and see if people actually want this. With the switch from PEP 563 to 649, I don't know how it will affect the **future**.annotations. And supporting only generics and unions but not some types or forwardrefs may create more support. I tag this PR as deferred in the meantime

Hi @PrettyWood - I'd like to give a +1 in support of landing this if still possible. I think the support for unions & generics alone will cover 80+% of cases. It'd be really nice to be able to write "x: str | None" which is what I imagine the most common use case would be. And forward refs in particular have always been a little bit of a thorn for me (even with the stdlib), and I don't think it needs to something like this supporting the basic cases.

PrettyWood:
@tuchandra If you use 3.10, it's already supported. This PR is to have a workaround for python 3.7+ using `from __future__ import annotations` but I don't think it's worth merging it.

tuchandra:
@PrettyWood sorry, I wasn't clear - yeah, I'm on 3.8/3.9 but I understand not merging it. Thanks for the reply!

samuelcolvin:
> @tuchandra If you use 3.10, it's already supported. This PR is to have a workaround for python 3.7+ using `from __future__ import annotations` but I don't think it's worth merging it.

agreed

eseglem:
I understand not merging as well but I would certainly love to see something like this actually get merged. Its not great but unfortunately I have systems stuck with 3.8 for the foreseeable future and it would be nice to use the newer syntax.

It could also be useful for packages that use Pydantic which still support older versions of python as well.

But honestly, it probably better to focus on the future.

samuelcolvin:
I really think this would have been too confusing for users, we were right to drop it.

On a related note, see #4339.

Issue #2011, opened 2020-10-16
Sequence constraint to enforce only unique items


uriyyo:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this feature/change is needed
* [x] After submitting this, I commit to one of:
    * Look through open issues and helped at least one other person
    * Hit the "watch" button on this repo to receive notifications and I commit to help at least 2 people that ask questions in the future
    * Implement a Pull Request for a confirmed bug

# Feature Request

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.6.1
            pydantic compiled: False
                 install path: /Users/yuriikarabas/my-projects/pydantic/pydantic
               python version: 3.6.8 (v3.6.8:3c6b436a57, Dec 24 2018, 02:04:31)  [GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)]
                     platform: Darwin-19.6.0-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions', 'email-validator', 'devtools']

```

The idea of this feature is to add the ability to validate that sequence types (like `list`, `tuple`, or `set`) contains only unique items.

I think the best API for this feature will be something like this:
```python
from typing import List
from pydantic import BaseModel, EmailStr, Field

class User(BaseModel):
    email: EmailStr


class Form(BaseModel):
    users: List[User] = Field(unique=True)
```

In case when data contains duplicates, they will be available using `errors` method.
For instance:
```python
try:
    f = Form(users=[{'email': 'some@email.com'}] * 2)
except ValidationError as e:
    print(e.errors())
```
Will print:
```python
[
    {
        'loc': ('users',),
        'msg': "value contains duplicate items [User(email='some@email.com')]",
        'type': 'type_error.sequencenotunique',
        'ctx': {'duplicates': [User(email='some@email.com')]}
    },
]
```

Also, as an option, we can add a new generic type `Unique`. 
Usage will look like:
```python
from pydantic import BaseModel, EmailStr, Unique

class User(BaseModel):
    email: EmailStr


class Form(BaseModel):
    users: Unique[User]
```

I have already implemented this feature in my fork. [link to commit](https://github.com/uriyyo/pydantic/commit/bb67a2e9f15da0d826ba48da117230284182c1b3)

[...]

PR #2618, opened 2021-04-01
Add unique items validation to constrained lists - POC

nuno-andre:
## Change Summary

Add `unique_items` property and validation to `ConstrainedList` (and `conlist`).

## Related issue number

#2011

## Checklist

## Rationale

This proposal solves two related issues:

- `{"type": "array", "uniqueItems": true}` schemas being interpreted as a Python set.
  JSON Schema specs define `array` as [an ordered sequence of zero or more values](https://tools.ietf.org/html/rfc8259#section-1) and its instance equality as [_both are arrays, and have an equal value item-for-item_](http://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.4.2.2). So both casting JSON arrays to Python sets and defining Python sets as JSON arrays in the schema can lead to validation errors and/or data corruption.
- `conset` used instead of `uniqueItems`
  In addition to the ordering issue, `ConstrainedSet` casts arrays to Python sets with no previous validation of items uniqueness; which is fully consistent with the Python `set` behavior, but opposite to that of the schema with which it is equated.

A `unique_items` validation in `ConstrainedList` would solve both issues while keeping the current `conset` behavior untouched.

This PR is a POC and lacks the proper tests and docs, and also it does not replicate the property into other data models (`FieldInfo`?). If you consider this proposal relevant, I will be glad to complete it for evaluation.

## To be evaluated

- Implementation follows the specs, with one exception: __default value__ is `None` instead of `False` so as `"uniqueItems": false` will not be added to the schema unless explicitly stated.

- In the feature request the __property name__ is `unique`. This PR named it `unique_items` for consistency with `max_items` and `min_items`.

- Current __error message__ is a placeholder. As specs do not state which mathematical object corresponds to this schema, there's no clearly defined concept of multiplicity. E.g. in `[1, 1, 2, 2, 2, 3]` one could understand: two items (1, 2), three items (1, 2, 2), or five items (1, 1, 2, 2, 2). I tend to see it as a set, so the first statement makes more sense to me.


pydantic/types.py
```diff
@@ -479,15 +479,17 @@ class ConstrainedList(list):  # type: ignore
 
     min_items: Optional[int] = None
     max_items: Optional[int] = None
+    unique_items: Optional[bool] = None
     item_type: Type[T]  # type: ignore
 
     @classmethod
     def __get_validators__(cls) -> 'CallableGenerator':
         yield cls.list_length_validator
+        yield cls.unique_items_validator
```
samuelcolvin, line 488:
```suggestion
        if cls.unique_items:
            yield cls.unique_items_validator
```
This should mean the check can occur outside performance sensitive validation stage.

pydantic/types.py
```diff
@@ -505,10 +507,21 @@ def list_length_validator(cls, v: 'Optional[List[T]]') -> 'Optional[List[T]]':
 
         return v
 
+    @classmethod
+    def unique_items_validator(cls, v: 'Optional[List[T]]') -> 'Optional[List[T]]':
+        if cls.unique_items and v and len(set(v)) != len(v):
+            raise errors.ListUniqueItemsError(not_unique=len(v) - len(set(v)))
```
samuelcolvin, line 513: you can avoid duplicating the `len(v)` and `len(set(v))` calls.

samuelcolvin: What are we going to do for lists of unhashable types?

Maybe we just decide that is a limitation of this feature, but at the least we should document it.

samuelcolvin (commented):
Seems like a good start, but needs documentation, fixes and some tweaks.

We need to decide (what do others think?) whether we should support uniqueness checks for things that can't be hashed but can be equalsed - e.g. `list`, `dict`, `set` etc.

samuelcolvin:
please update.

pydantic/types.py
nuno-andre: They are incidentally duplicated because the exception message is just a placeholder.

> Current error message is a placeholder. As specs do not state which mathematical object corresponds to this schema, there's no clearly defined concept of multiplicity. E.g. in `[1, 1, 2, 2, 2, 3]` one could understand: two items (1, 2), three items (1, 2, 2), or five items (1, 1, 2, 2, 2). I tend to see it as a set, so the first statement makes more sense to me.

I still don't know how many elements are supposed to be leftover if validation fails. Do you have an opinion on this? If not I can try to open an issue at the [JSON Schema repo](https://github.comjson-schema-org/json-schema-spec).

But in this case it would also be a micro-optimization, since the second calls are only made on exceptions, and this would save two `STORE_FAST` and two `LOAD_FAST` in a successful validation.

---

We can add a failover for unhashable types. E.g. something such as _(with the appropriate exceptions)_

```python
try:
    if v and len(set(v)) != len(v):
        raise ValueError
except TypeError:
    unique = list()
    if len([unique.append(i) for i in v if i not in unique]) != len(v):
        raise ValueError from None
```

I can benchmark some options. IMHO, as in the previous case, I think it is worth optimizing the check at the expense of a more expensive validation error. What do you think?

nuno-andre: json-schema-org/json-schema-spec#1099

pydantic/types.py
nuno-andre: It's the first checking in the validator but this a better place, indeed.

pydantic/types.py
nuno-andre: The failover for unhashable types seems too tricky to mypy 😅 Would you accept a `type: ignore` if it turns out to be the most performant?

nuno-andre: There's currently no instance-based output structure for validation errors (only schema-based). So, if it's okay with you, I'll continue with a generic error message to move on to the unhashable types validation.

WilliamDEdwards:
Will this be merged without support for unhashable types? I'd love to use this with lists.

nuno-andre:
please review!

@WilliamDEdwards it's already mergeable for hashable and unhashable types 😉

nuno-andre:
@PrettyWood @samuelcolvin I understand that the reduction in coverage is due to the `# type: ignore`. Line 538 is indeed reachable (as shown in tests). Is there any way to adjust the coverage check in a case like this?

WilliamDEdwards:
> please review!
> 
> @WilliamDEdwards it's already mergeable for hashable and unhashable types 😉

Ah, I should've read the full discussion. Can't wait :)

WilliamDEdwards:
Does this PR need any work? I'd contribute, but it looks pretty finished.

pydantic/types.py
```diff
@@ -505,10 +508,28 @@ def list_length_validator(cls, v: 'Optional[List[T]]') -> 'Optional[List[T]]':
 
         return v
 
+    @classmethod
+    def unique_items_validator(cls, v: 'Optional[List[T]]') -> 'Optional[List[T]]':
+        try:
```
samuelcolvin, line 513: I'm not sure about this "try with a set, then fallback approach", I think it might be cleaner to always use the approach below.

samuelcolvin: alternatively, we could give the user a choice of either they want to use a set or not.

pydantic/types.py
```diff
@@ -505,10 +508,28 @@ def list_length_validator(cls, v: 'Optional[List[T]]') -> 'Optional[List[T]]':
...
+            if v and len(set(v)) != len(v):
+                raise errors.ListUniqueItemsError(not_unique=len(v) - len(set(v)))
+        except TypeError:
+            # failover for unhashable types
+            unique = list()
 
-def conlist(item_type: Type[T], *, min_items: int = None, max_items: int = None) -> Type[List[T]]:
+            if v and len([unique.append(i) for i in v if i not in unique]) != len(v):  # type: ignore
```
samuelcolvin, line 520: modifying unique inside a comprehension is somewhat unorthodox, also (if we don't care about how many items are not unique, which I will mostly be the case) we can do this faster by raising the error as soon as we find a duplicate.

Something like
```py
for i, value in enumerate(v, start=1):
    if value in v[i:]:
        raise errors.ListUniqueItemsError()
```

I haven't done any profiling but I feel something like this should be quicker.

samuelcolvin (commented):
a few comments, let me know what you think.

The "hash or not to hash" questions is a difficult one. I'm not sure what I think about it.

samuelcolvin:
please update.

pydantic/types.py
nuno-andre: I've just profiled it and you're totally right about performance. Also, the difference between the `set` and the loop checks is almost negligible (even when the duplicate is at the end of the iterable).

I also agree that the most of the time we don't care about how many items aren't unique, and furthermore, this calculation reduces the performance. So I'm going to refactor this PR to push it again.

pydantic/types.py
nuno-andre: Reviewed below.

nuno-andre:
please review

samuelcolvin:
coverage is failing.

nuno-andre:
Yes, I'm sorry. I mixed up different checks.

But [this issue in `mypy.py`](https://smokeshow.helpmanual.io/2i3a6y6g28205z1i3d01/d_2705a9c05af28909_mypy_py.html#t273) is not related to my PR, isn't it?

samuelcolvin:
thanks so much.

WilliamDEdwards:
@samuelcolvin When are you planning to release a new version which includes this patch?

WilliamDEdwards:
> @samuelcolvin When are you planning to release a new version which includes this patch?

For future readers: this has been included in v1.9.0.

WilliamDEdwards:
I think this may not be working well with optional lists:

`'NoneType' object is not iterable`

Haven't done any further investigation yet, so not 100% sure this is because of `unique_items`.

nuno-andre:
@WilliamDEdwards Please, could you provide a minimal reproducible example? This is currently working in 1.9.0:

```python
from typing import Any, Optional
from pydantic import BaseModel, conlist

class Model(BaseModel):
    prop: Optional[conlist(Any, unique_items=True)]

m = Model()
print(m.prop)
#> None
```

WilliamDEdwards:
@nuno-andre Sorry for not being more specific yesterday. I hadn't been able to make an MRE, but wanted to report the issue (if it is an issue, I don't exclude PEBKAC).

Here's the MRE. It's worth noting that I use `List` with `Field` rather than `conlist`, because mypy doesn't seem to work well with `con*`.

Code:

```
from typing import Optional, List

class TestSchema(BaseModel):
  test: Optional[List[str]] = Field(default=..., unique_items=True)
```

Example:

```
>>> TestSchema(test=None)
  File "pydantic/main.py", line 331, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for TestSchema
test
  'NoneType' object is not iterable (type=type_error)

>>> TestSchema(test=['test'])
TestSchema(test=['test'])
```

WilliamDEdwards:
Hi @nuno-andre,

> @WilliamDEdwards Please, could you provide a minimal reproducible example? This is currently working in 1.9.0:
> 
> ```python
> from typing import Any, Optional
> from pydantic import BaseModel, conlist
> 
> class Model(BaseModel):
>     prop: Optional[conlist(Any, unique_items=True)]
> 
> m = Model()
> print(m.prop)
> #> None
> ```

The problem is in specifying `None` explicitly. `Model()` works, `Model(prop=None)` doesn't, with [this error](https://github.com/pydantic/pydantic/pull/2618#issuecomment-1015632200).

nuno-andre:
@WilliamDEdwards I've just created the issue (#4626). I'm submitting a PR. Thank you for raising this issue!

PR #2625, opened 2021-04-02
feat: add computed fields

### LINKED ISSUE #935: Serializing @property
# Question

* OS: Linux(Ubuntu 18)
* Python version 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)]
* Pydantic version 1.0

I've pored over the documentation and issue and can't seem to find an example for serializing members other than class attributes.  

```py
class Rectangle(BaseModel):
    width: int
    length: int

    @property
    def area(self) -> int:
        return self.width * self.length

r = Rectangle(width=10, length=5)

print(r.json())

# What I'd like to see:
{ "width": 10, "length": 5, "area": 50 }
```
I'd also think that properties without setters would default to readOnly on the schema side as well.

It did seem there was a way to get something close to what I'm after via ORM mode and masking a property on the ORM model, but I'd hate to need to create a class just to contain properties.

Going over [BaseModel._calculate_keys](https://github.com/samuelcolvin/pydantic/blob/25f657fd5ff03f661cb84656747b0a90bad02c46/pydantic/main.py#L653) it seems BaseModel will not yield what I'm after.

Still, I'm hoping that it's just been a long day and I overlooked something :)

### LINKED ISSUE #1241: cached_property issues
# Unable to use cached_property

Hi, 

I am using pydantic for almost any project right now and I find it awesome.

I recently found an handy package, [funcy](https://github.com/Suor/funcy), and I am trying to work with `cached_property` decorator.

What I am doing is something like this:

```py
class A(BaseModel):

    class Config:

    @cached_property
    def test(self):
        return 10
```

The `arbitrary_types_allowed = True` is needed because if not present, a `ValidationError` is raised:

```
File "pydantic/main.py", line 215, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/fields.py", line 277, in pydantic.fields.ModelField.infer
  File "pydantic/fields.py", line 249, in pydantic.fields.ModelField.__init__
  File "pydantic/fields.py", line 335, in pydantic.fields.ModelField.prepare
  File "pydantic/fields.py", line 468, in pydantic.fields.ModelField.populate_validators
  File "pydantic/validators.py", line 574, in find_validators
RuntimeError: no validator found for <class 'funcy.objects.cached_property'>, see `arbitrary_types_allowed` in Config
```

When i do:
```py
a = A()

print(a.test)

# this is what i get
<funcy.objects.cached_property object at 0x7f70466e4748>

# while of course i would like to have 10 as an output.
```

Same happens with python3.8 `cached_property`.
If doing the same with a vanilla python class, the behaviour is the expected one (the property outputs 10).

Is there anything I am missing?
Could anyone help me?

Thank you very much, pydantic is really great and you are doing a terrific job.

### LINKED ISSUE #2313: Pydantic tries to validate method decorated with functools.singledispatchmethod as class attribute
### Checks

See code example:
```py
from functools import singledispatchmethod

class A(BaseModel):
    x: int
        
    @singledispatchmethod
    def add_to_x(self, y):
        raise NotImplementedError()
```

Running this leads to:

```
RuntimeError                              Traceback (most recent call last)
<ipython-input-8-f3e84a8e5cb2> in <module>
      2 from functools import singledispatchmethod
      3 
----> 4 class A(BaseModel):
      5     x: int
      6 

~/.venvs/clarinet/lib/python3.8/site-packages/pydantic/main.cpython-38-darwin.so in pydantic.main.ModelMetaclass.__new__()

~/.venvs/clarinet/lib/python3.8/site-packages/pydantic/fields.cpython-38-darwin.so in pydantic.fields.ModelField.infer()

~/.venvs/clarinet/lib/python3.8/site-packages/pydantic/fields.cpython-38-darwin.so in pydantic.fields.ModelField.__init__()

~/.venvs/clarinet/lib/python3.8/site-packages/pydantic/fields.cpython-38-darwin.so in pydantic.fields.ModelField.prepare()

~/.venvs/clarinet/lib/python3.8/site-packages/pydantic/fields.cpython-38-darwin.so in pydantic.fields.ModelField.populate_validators()

~/.venvs/clarinet/lib/python3.8/site-packages/pydantic/validators.cpython-38-darwin.so in find_validators()

RuntimeError: no validator found for <class 'functools.singledispatchmethod'>, see `arbitrary_types_allowed` in Config
```

Surely this is not intended behaviour? Is it a feature of functools that this property is added as a result of calling `singledispatchmethod`? 

```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: /Users/phinate/.venvs/clarinet/lib/python3.8/site-packages/pydantic
               python version: 3.8.3 (default, May 19 2020, 13:54:14)  [Clang 10.0.0 ]
     optional deps. installed: []
```

PrettyWood:
## Change Summary
I'm waiting for feedback on the API and [doc](https://smokeshow.helpmanual.io/3h0m1f4t2i5e2b26723a/usage/models/#field-with-computed-value-based-on-other-fields) 🙏  

### `property`-like syntax

```py
from random import randint

from pydantic import BaseModel, computed_field


class Square(BaseModel, underscore_attrs_are_private=True):
    side: float
    _private_attr: int = 1

    @computed_field(title='The area')
    @property
    def area(self) -> float:
        return self.side ** 2

    @area.setter
    def area(self, area: float) -> None:
        self.side = area ** .5

    @area.deleter
    def area(self) -> None:
        self.side = 0

    @computed_field(alias='The random number')
    @cached_property
    def random_n(self) -> int:
        return randint(0, 1_000)

    @property
    def public_attr(self) -> int:
        return self._private_attr + 1

    @public_attr.setter
    def public_attr(self, v: int) -> None:
        self._private_attr = v - 1


sq = Square(side=10)
assert sq._private_attr == 1
assert sq.public_attr == 2
sq.public_attr = 5
assert sq._private_attr == 4
assert sq.public_attr == 5
the_random_n = sq.random_n
assert sq.dict(by_alias=True) == {'side': 10, 'area': 100, 'The random number': the_random_n}
sq.area = 49
assert sq.dict() == {'side': 7, 'area': 49, 'random_n': the_random_n}
del sq.area
assert sq.dict() == {'side': 0, 'area': 0, 'random_n': the_random_n}
assert sq.dict(exclude={'random_n'}) == {'side': 0, 'area': 0}
```

### `ComputedField` syntax

```py
from pydantic import BaseModel, ComputedField, ValidationError, validator


class User(BaseModel, validate_assignment=True):
    first_name: str
    surname: str
    full_name: str = ComputedField(lambda self: f'{self.first_name} {self.surname}')

    @validator('full_name')
    def should_not_have_pika(cls, v: str) -> str:
        if 'pika' in v.lower():
            raise ValueError('We love pika but not that much!')
        return v


u = User(first_name='John', surname='Smith')
assert u.dict() == {'first_name': 'John', 'surname': 'Smith', 'full_name': 'John Smith'}
u.first_name = 'Mike'
assert u.dict()['full_name'] == 'Mike Smith'

try:
    User(first_name='Pika', surname='Chu')
    assert False
except ValidationError:
    # pydantic.error_wrappers.ValidationError: 1 validation error for User
    # full_name
    #   We love pika but not that much! (type=value_error)
    assert True
```

See [the tests](https://github.com/PrettyWood/pydantic/blob/field-property/tests/test_computed_fields.py) for more examples

## Related issue number
closes #935
closes #1241
closes #2313

## Checklist

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -574,7 +574,7 @@ _(This script is complete, it should run "as is")_
     If `computed_field` decorator doesn't wrap a decorator like `property`, `cached_property`, ...,
     it will fallback and create `property` itself. This is shorter and may be what you want, but you
     will probably lose IntelliSense in your IDE and have some warnings.
-    This is why it is recommanded to combine `computed_field` with `property`<br>
+    This is why it is recommended to combine `computed_field` with `property`<br>
     **Warning for _mypy_**: Even with this, _mypy_ won't be happy because of [this issue](https://github.com/python/mypy/issues/1362)
 
 ```py
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -554,6 +554,34 @@ Where `Field` refers to the [field function](schema.md#field-customisation).
 !!! warning
     The `default_factory` expects the field type to be set.
 
+## Field with computed value based on other fields
+
+You have two ways of writing a computed field, either with a `property`-like syntax
+
+```py
+{!.tmp_examples/models_computed_field_decorator.py!}
+```
+_(This script is complete, it should run "as is")_
+
+or a `Field`-like one
+
+```py
+{!.tmp_examples/models_computed_field_annotation.py!}
+```
+_(This script is complete, it should run "as is")_
+
+!!! note
+    If `computed_field` decorator doesn't wrap a decorator like `property`, `cached_property`, ...,
+    it will fallback and create `property` itself. This is shorter and may be what you want, but you
+    will probably lose IntelliSense in your IDE and have some warnings.
+    This is why it is recommanded to combine `computed_field` with `property`<br>
+    **Warning for _mypy_**: Even with this, _mypy_ won't be happy because of [this issue](https://github.com/python/mypy/issues/1362)
+
+```py
+{!.tmp_examples/models_computed_field_decorator_details.py!}
+```
+_(This script is complete, it should run "as is")_
+
 ## Automatically excluded attributes
 
 Class variables which begin with an underscore and attributes annotated with `typing.ClassVar` will be
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -339,7 +339,7 @@ def _get_descriptor_infos(descriptor: Any) -> Tuple[Any, 'FGet', Optional['FSet'
 
 
 @overload
-def field(
+def computed_field(
     *,
     alias: Optional[str] = None,
     title: Optional[str] = None,
@@ -351,7 +351,7 @@ def field(
 
 
 @overload
-def field(
+def computed_field(
     func: Any,
     *,
     alias: Optional[str] = None,
@@ -363,7 +363,7 @@ def field(
     ...
 
 
-def field(
+def computed_field(
     func: Optional[Any] = None,
     *,
     alias: Optional[str] = None,
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -282,18 +282,36 @@ SHAPE_NAME_LOOKUP = {
 MAPPING_LIKE_SHAPES: Set[int] = {SHAPE_DEFAULTDICT, SHAPE_DICT, SHAPE_MAPPING}
 
 
-def _descriptor_getter(func: Any) -> Any:
-    with suppress(AttributeError):
-        return func.fget
+class DescriptorInfos:
+    __slots__ = ('type_', 'fget', 'fset')
 
-    with suppress(AttributeError):
-        return func.func
+    def __init__(self, type_: type, fget: Any, fset: Any) -> None:
+        self.type_ = type_
+        self.fget = fget
+        self.fset = fset
 
-    # try to get the first parameter after self
-    from inspect import signature
+    @classmethod
+    def infer(cls, descriptor: Any) -> 'DescriptorInfos':
+        import inspect
 
-    _self_param, getter_param, *_ = signature(func.__class__.__init__).parameters.values()
-    return getattr(func, getter_param.name)
+        fset = None
+
+        if isinstance(descriptor, property):
+            fget = descriptor.fget
+            fset = descriptor.fset
+        else:
+            _self_param, getter_param, *set_del_params = inspect.signature(
+                descriptor.__class__.__init__
+            ).parameters.values()
+            fget = getattr(descriptor, getter_param.name)
+            if len(set_del_params) > 0:
+                fset = getattr(descriptor, set_del_params[0].name)
+
+        type_ = inspect.signature(fget).return_annotation
+        if type_ is inspect._empty:  # type: ignore[attr-defined]
+            type_ = Any
+
+        return cls(type_, fget, fset)
 
 
 def field(
@@ -309,10 +327,7 @@ def field(
         if not (ismethoddescriptor(func_) or isdatadescriptor(func_)):
             func_ = property(func_)
 
-        getter = _descriptor_getter(func_)
-
         return ComputedField(
-            name=getter.__name__,
             descriptor=func_,
             alias=alias,
             title=title,
@@ -327,8 +342,9 @@ def field(
 
 class ComputedField(Representation):
     __slots__ = (
-        'name',
         'descriptor',
+        'descriptor_infos',
+        'name',
         'class_validators',
         'alias',
         'title',
@@ -342,20 +358,20 @@ class ComputedField(Representation):
     def __init__(
         self,
         *,
-        name: str,
         descriptor: Any,
+        name: Optional[str] = None,
         alias: Optional[str] = None,
         title: Optional[str] = None,
         description: Optional[str] = None,
         config: Optional[Type['BaseConfig']] = None,
         model_field: Optional['ModelField'] = None,
     ) -> None:
-        self.name: str = name
         self.descriptor = descriptor
-
+        self.descriptor_infos = DescriptorInfos.infer(descriptor)
+        self.name: str = name or self.descriptor_infos.fget.__name__
         self.class_validators: Optional[Dict[str, 'Validator']] = None
 
-        self.alias: str = alias or name
+        self.alias: str = alias or self.name
         self.title: Optional[str] = title
         self.description: Optional[str] = description
 
@@ -364,21 +380,12 @@ class ComputedField(Representation):
         self.required: bool = False
 
     @property
-    def type_(self) -> Any:
-        import inspect
-
-        getter = _descriptor_getter(self.descriptor)
-        type_ = inspect.signature(getter).return_annotation
-
-        if type_ is inspect._empty:  # type: ignore[attr-defined]
-            return Any
-        else:
-            return type_
-
-    @property
     def field_info(self) -> FieldInfo:
         return FieldInfo(
-            alias=self.alias, title=self.title, description=self.description or self.descriptor.__doc__, read_only=True
+            alias=self.alias,
+            title=self.title,
+            description=self.description or self.descriptor.__doc__,
+            read_only=self.descriptor_infos.fset is None,
         )
 
     def __get__(self, instance: Optional['BaseModel'], owner: Any) -> Any:
@@ -431,7 +438,7 @@ class ComputedField(Representation):
         self.config = config
         self.model_field = ModelField(
             name=self.name,
-            type_=self.type_,
+            type_=self.descriptor_infos.type_,
             class_validators=self.class_validators,
             model_config=self.config,
             default=None,
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -103,6 +103,7 @@ class FieldInfo(Representation):
         'max_length',
         'allow_mutation',
         'regex',
+        'read_only',
         'extra',
     )
 
@@ -140,6 +141,7 @@ class FieldInfo(Representation):
         self.max_length = kwargs.pop('max_length', None)
         self.allow_mutation = kwargs.pop('allow_mutation', True)
         self.regex = kwargs.pop('regex', None)
+        self.read_only = kwargs.pop('read_only', None)
         self.extra = kwargs
 
     def __repr_args__(self) -> 'ReprArgs':
@@ -277,6 +279,128 @@ SHAPE_NAME_LOOKUP = {
 }
 
 MAPPING_LIKE_SHAPES: Set[int] = {SHAPE_DEFAULTDICT, SHAPE_DICT, SHAPE_MAPPING}
+
+
+def field(
+    func: Optional[Any] = None,
+    *,
+    alias: Optional[str] = None,
+    title: Optional[str] = None,
+    description: Optional[str] = None,
+) -> 'Union[Callable[[Any], ComputedField], ComputedField]':
+    import inspect
+
+    def wrap(func_: Any) -> 'ComputedField':
+        if not isinstance(func_, property):
+            func_ = property(func_)
+        type_ = inspect.signature(func_.fget).return_annotation
+        if type_ is inspect._empty:  # type: ignore[attr-defined]
+            type_ = Any
+
+        return ComputedField(
+            name=func_.fget.__name__,
+            type_=type_,
+            fget=func_.fget,
+            alias=alias,
+            title=title,
+            description=description,
+        )
+
+    if func is None:
+        return wrap
+
+    return wrap(func)
+
+
+class ComputedField(Representation):
+    __slots__ = (
+        'name',
+        'type_',
+        'outer_type_',
+        'fget',
+        'fset',
+        'class_validators',
+
+        'alias',
+        'title',
+        'description',
+
+        'config',
+        'model_field',
+        'required',
+    )
+
+    def __init__(
+        self,
+        *,
+        name: str,
+        type_: type,
+        fget: Callable[[Optional['BaseModel']], Any],
+        fset: Any = None,
+        alias: Optional[str] = None,
+        title: Optional[str] = None,
+        description: Optional[str] = None,
+        config: Optional[Type['BaseConfig']] = None,
+        model_field: Optional['ModelField'] = None,
+    ) -> None:
+        self.name: str = name
+        self.type_: type = type_
+        self.outer_type_: type = type_
+        self.fget: Callable[[Optional['BaseModel']], Any] = fget
+        self.fset: Any = fset
+        self.class_validators: Optional[Dict[str, 'Validator']] = None
+
+        self.alias: str = alias or name
+        self.title: Optional[str] = title
+        self.description: Optional[str] = description
+
+        self.config: Optional[Type['BaseConfig']] = config
+        self.model_field: Optional[ModelField] = model_field
+        self.required: bool = False
+
+    @property
+    def field_info(self) -> FieldInfo:
+        return FieldInfo(
+            alias=self.alias,
+            title=self.title,
+            description=self.description or self.fget.__doc__, read_only=True
+        )
+
+    def __get__(self, instance: Optional['BaseModel'], owner: Any) -> Any:
+        if instance is None:
+            return self
+        return self.fget(instance)
+
+    def __set__(self, obj, value):
+        if self.fset is None:
+            raise AttributeError("can't set attribute")
+        self.fset(obj, value)
+
+    def setter(self, fset):
+        return type(self)(
+            name=self.name,
+            type_=self.type_,
+            fget=self.fget,
+            fset=fset,
+            alias=self.alias,
+            title=self.title,
+            description=self.description,
+            config=self.config,
+            model_field=self.model_field,
+        )
+
+    def set_config_and_prepare_field(self, config: Type['BaseConfig']) -> None:
+        self.config = config
+        self.model_field = ModelField(
+            name=self.name,
+            type_=self.type_,
+            class_validators=self.class_validators,
+            model_config=self.config,
+            default=None,
+            required=self.required,
+            alias=self.field_info.alias,
+            field_info=self.field_info,
+        )
 
 
 class ModelField(Representation):
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -82,7 +82,6 @@ if TYPE_CHECKING:
 
     FGet = Callable[[Optional[BaseModel], str], Any]
     FSet = Callable[[Optional[BaseModel], Any], None]
-    FDel = Callable[[Optional[BaseModel], str], None]
 
 
 class FieldInfo(Representation):
@@ -293,8 +292,8 @@ def _get_descriptor_infos(descriptor: Any) -> Tuple[Any, 'FGet', Optional['FSet'
     fset = None
 
     if isinstance(descriptor, property):
-        fget = descriptor.fget
-        fset = descriptor.fset
+        fget = cast('FGet', descriptor.fget)
+        fset = cast('FSet', descriptor.fset)
     else:
         _self_param, getter_param, *set_del_params = inspect.signature(
             descriptor.__class__.__init__
@@ -308,7 +307,7 @@ def _get_descriptor_infos(descriptor: Any) -> Tuple[Any, 'FGet', Optional['FSet'
     if type_ is sig.empty:
         type_ = Any
 
-    return type_, cast('FGet', fget), cast('FSet', fset)
+    return type_, fget, fset
 
 
 @overload
@@ -323,7 +322,7 @@ def field(
 
 @overload
 def field(
-    func: T,
+    func: Any,
     *,
     alias: Optional[str] = None,
     title: Optional[str] = None,
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -349,7 +349,6 @@ class ComputedField(Representation):
         'config',
         'model_field',
         'required',
-        'descriptor',
     )
 
     def __init__(
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -21,6 +21,7 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    overload,
 )
 
 from typing_extensions import Annotated
@@ -308,6 +309,27 @@ def _get_descriptor_infos(descriptor: Any) -> Tuple[Any, 'FGet', Optional['FSet'
         type_ = Any
 
     return type_, cast('FGet', fget), cast('FSet', fset)
+
+
+@overload
+def field(
+    *,
+    alias: Optional[str] = None,
+    title: Optional[str] = None,
+    description: Optional[str] = None,
+) -> 'Callable[[Any], ComputedField]':
+    ...
+
+
+@overload
+def field(
+    func: T,
+    *,
+    alias: Optional[str] = None,
+    title: Optional[str] = None,
+    description: Optional[str] = None,
+) -> Any:
+    ...
 
 
 def field(
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -302,8 +302,9 @@ def _get_descriptor_infos(descriptor: Any) -> Tuple[Any, 'FGet', Optional['FSet'
         if len(set_del_params) > 0:
             fset = getattr(descriptor, set_del_params[0].name)
 
-    type_ = inspect.signature(fget).return_annotation
-    if type_ is inspect._empty:  # type: ignore[attr-defined]
+    sig = inspect.signature(fget)
+    type_ = sig.return_annotation
+    if type_ is sig.empty:
         type_ = Any
 
     return type_, cast('FGet', fget), cast('FSet', fset)
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -1,6 +1,5 @@
 from collections import defaultdict, deque
 from collections.abc import Iterable as CollectionsIterable
-from contextlib import suppress
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -78,6 +77,10 @@ if TYPE_CHECKING:
     ValidateReturn = Tuple[Optional[Any], Optional[ErrorList]]
     LocStr = Union[Tuple[Union[int, str], ...], str]
     BoolUndefined = Union[bool, UndefinedType]
+
+    FGet = Callable[[Optional[BaseModel], str], Any]
+    FSet = Callable[[Optional[BaseModel], Any], None]
+    FDel = Callable[[Optional[BaseModel], str], None]
 
 
 class FieldInfo(Representation):
@@ -285,10 +288,10 @@ MAPPING_LIKE_SHAPES: Set[int] = {SHAPE_DEFAULTDICT, SHAPE_DICT, SHAPE_MAPPING}
 class DescriptorInfos:
     __slots__ = ('type_', 'fget', 'fset')
 
-    def __init__(self, type_: type, fget: Any, fset: Any) -> None:
-        self.type_ = type_
-        self.fget = fget
-        self.fset = fset
+    def __init__(self, type_: type, fget: 'FGet', fset: Optional['FSet']) -> None:
+        self.type_: type = type_
+        self.fget: 'FGet' = fget
+        self.fset: Optional['FSet'] = fset
 
     @classmethod
     def infer(cls, descriptor: Any) -> 'DescriptorInfos':
@@ -311,7 +314,7 @@ class DescriptorInfos:
         if type_ is inspect._empty:  # type: ignore[attr-defined]
             type_ = Any
 
-        return cls(type_, fget, fset)
+        return cls(type_, fget, fset)  # type: ignore
 
 
 def field(
@@ -388,7 +391,7 @@ class ComputedField(Representation):
             read_only=self.descriptor_infos.fset is None,
         )
 
-    def __get__(self, instance: Optional['BaseModel'], owner: Any) -> Any:
+    def __get__(self, instance: Optional['BaseModel'], owner: Any = None) -> Any:
         return self.descriptor.__get__(instance, owner)
 
     def __set__(self, instance: Optional['BaseModel'], value: Any) -> None:
@@ -397,42 +400,27 @@ class ComputedField(Representation):
     def __delete__(self, obj: Any) -> None:
         return self.descriptor.__delete__(obj)
 
-    def getter(self, fget: 'Callable[[Optional[BaseModel], str], Any]') -> 'ComputedField':
-        return type(self)(
-            name=self.name,
-            descriptor=self.descriptor.getter(fget),
-            alias=self.alias,
-            title=self.title,
-            description=self.description,
-            config=self.config,
-            model_field=self.model_field,
-        )
-
-    def setter(self, fset: 'Callable[[Optional[BaseModel], Any], None]') -> 'ComputedField':
-        return type(self)(
-            name=self.name,
-            descriptor=self.descriptor.setter(fset),
-            alias=self.alias,
-            title=self.title,
-            description=self.description,
-            config=self.config,
-            model_field=self.model_field,
-        )
-
-    def deleter(self, fdel: 'Callable[[Optional[BaseModel], str], None]') -> 'ComputedField':
-        return type(self)(
-            name=self.name,
-            descriptor=self.descriptor.deleter(fdel),
-            alias=self.alias,
-            title=self.title,
-            description=self.description,
-            config=self.config,
-            model_field=self.model_field,
-        )
-
     def __set_name__(self, instance: 'BaseModel', name: str) -> None:
-        with suppress(Exception):
+        if hasattr(self.descriptor, '__set_name__'):
             self.descriptor.__set_name__(instance, name)
+
+    def __getattr__(self, name: str) -> Any:
+        # we forward only `setter`, `getter` (public methods)
+        # we don't want to forward things like `__isabstractmethod__`
+        if not name.startswith('_') and hasattr(self.descriptor, name):
+
+            def update_computed_field(*args: Any, **kwargs: Any) -> 'ComputedField':
+                return type(self)(
+                    descriptor=getattr(self.descriptor, name)(*args, **kwargs),
+                    name=self.name,
+                    alias=self.alias,
+                    title=self.title,
+                    description=self.description,
+                    config=self.config,
+                    model_field=self.model_field,
+                )
+
+            return update_computed_field
 
     def set_config_and_prepare_field(self, config: Type['BaseConfig']) -> None:
         self.config = config
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -373,13 +373,8 @@ def field(
     include: Optional['ExcludeInclude'] = None,
 ) -> 'Union[Callable[[Any], ComputedField], ComputedField]':
     def wrap(func_: Any) -> 'ComputedField':
-        from inspect import isdatadescriptor, ismethoddescriptor
-
-        if not (ismethoddescriptor(func_) or isdatadescriptor(func_)):
-            func_ = property(func_)
-
         return ComputedField(
-            descriptor=func_,
+            func_,
             alias=alias,
             title=title,
             description=description,
@@ -393,13 +388,28 @@ def field(
     return wrap(func)
 
 
+class GetAttributeDict:
+    """
+    A tiny wrapper around a dictionary to be able to call a descriptor on it
+    Calling `self.first_name` for example will return `self['first_name']`
+    """
+
+    __slots__ = ('_obj',)
+
+    def __init__(self, obj: Dict[str, Any]) -> None:
+        self._obj = obj
+
+    def __getattr__(self, item: str) -> Any:
+        return self._obj[item]
+
+
 class ComputedField(Representation):
     __slots__ = (
         'descriptor',
         'type_',
         'fget',
         'fset',
-        'name',
+        'defined_name',
         'class_validators',
         'alias',
         'title',
@@ -413,8 +423,8 @@ class ComputedField(Representation):
 
     def __init__(
         self,
+        default_callable: Any,
         *,
-        descriptor: Any,
         name: Optional[str] = None,
         alias: Optional[str] = None,
         title: Optional[str] = None,
@@ -424,12 +434,17 @@ class ComputedField(Representation):
         config: Optional[Type['BaseConfig']] = None,
         model_field: Optional['ModelField'] = None,
     ) -> None:
-        self.descriptor: Any = descriptor
-        type_, fget, fset = _get_descriptor_infos(descriptor)
+        from inspect import isdatadescriptor, ismethoddescriptor
+
+        if not (ismethoddescriptor(default_callable) or isdatadescriptor(default_callable)):
+            default_callable = property(default_callable)
+
+        self.descriptor: Any = default_callable
+        type_, fget, fset = _get_descriptor_infos(default_callable)
         self.type_: Any = type_
         self.fget: 'FGet' = fget
         self.fset: Optional['FSet'] = fset
-        self.name: str = name or self.fget.__name__
+        self.defined_name: Optional[str] = name
         self.class_validators: Optional[Dict[str, 'Validator']] = None
 
         self.alias: str = alias or self.name
@@ -441,6 +456,10 @@ class ComputedField(Representation):
         self.config: Optional[Type['BaseConfig']] = config
         self.model_field: Optional[ModelField] = model_field
         self.required: bool = False
+
+    @property
+    def name(self) -> str:
+        return self.defined_name or self.fget.__name__
 
     @property
     def field_info(self) -> FieldInfo:
@@ -473,7 +492,7 @@ class ComputedField(Representation):
 
             def update_computed_field(*args: Any, **kwargs: Any) -> 'ComputedField':
                 return type(self)(
-                    descriptor=getattr(self.descriptor, name)(*args, **kwargs),
+                    getattr(self.descriptor, name)(*args, **kwargs),
                     name=self.name,
                     alias=self.alias,
                     title=self.title,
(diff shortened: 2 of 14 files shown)


PrettyWood:
Writing `computed_field` instead of `field` seems good but writing
```py
If TYPE_CHECKING:
    computed_field = property
```
won't work with kwargs like `alias` and stuff.

This is why I chose to keep the combo `field + property` (or any other property-like descriptor), which is well supported by mypy and the IDE without any effort. The only issue is that decorating property itself is [not yet handled by mypy](https://github.com/python/mypy/issues/1362) but it's a known issue that may be solved in the near future.

If anyone wants to work on a better mypy support feel free! But I won't do it myself: I have not the will nor the energy to do it since it's a lot of work for IMHO not a big gain.

PrettyWood:
Writing only `@computed_field` instead of `@field` with `@property` is annoying because of `mypy` and IDE. Writing
```py
If TYPE_CHECKING:
    computed_field = property
```
won't work with kwargs like `alias` and stuff.
And it won't work fine with property-like descriptors

This is why I chose to keep the combo `@field + @property`, which is explicit, just a few characters longer and well supported without any effort. The only issue is that decorating property itself is [not yet handled by mypy](https://github.com/python/mypy/issues/1362) but it's a known issue that may be solved in the near future.

If we want to change the API and need to work on a better mypy support fine but I won't do it myself sorry: I have not the will nor the energy to work on the mypy plugin since it's a lot of work for IMHO not a big gain.

layday:
I think this is very well done.  My only concern is that it's synonymous with `Field` (sans the capitalisation) which can be confusing, especially because `field`s are not constructor arguments. Is there a reason not to call this `computed_field` or `dynamic_field` - why do you think it'd need to be tied to `@property` for that?

PrettyWood:
@layday Thanks!
Writing
```py
...
    @computed_field(title='The area')
    def area(self) -> float:
        return self.side ** 2

    @area.setter
    def area(self, area: float) -> None:
        self.side = area ** .5

    @area.deleter
    def area(self) -> None:
        self.side = 0
```
is already supported with
```py
        if not (ismethoddescriptor(func_) or isdatadescriptor(func_)):
            func_ = property(func_)
```
but it doesn't work well with `mypy` and some IDEs
```console
error: Name 'area' already defined on line 10  [no-redef]
error: Name 'area' already defined on line 10  [no-redef]
error: Cannot assign to a method  [assignment]
error: Incompatible types in assignment (expression has type "int", variable has type "Callable[..., Any]")  [assignment]
```
So I felt like keeping `property` (or `cached_property` or whatever) and just add a basic decorator on top of them. It makes everything work properly without extra plugin.
And writing
```py
    @computed_field(title='The area')
    @property
    def area(self) -> float:
```
felt a bit weird to me because a `property` is by definition computed hence the reduced decorator name `'field'`.

I don't have a strong opinion on all of this but as I said in my comment I just don't want to tackle the whole mypy / type checking issue myself.

daviskirk:
I think it's fine to have both @property and @field. For non-pydantic pros this makes it super obvious what this is doing, might make usage of "property" vs. "cached_property" easier as well.
Perhaps it would even be good to have some sort of warning if the user defines a field without specifying a  property decorator?

daviskirk:
I think it's fine to have both @property and @field. For non-pydantic pros this makes it super obvious what this is doing, might make usage of "property" vs. "cached_property" easier as well.
Perhaps it would even be good to have some sort of warning if the user defines a field without specifying a  property decorator?

tests/test_private_attributes.py
```diff
@@ -160,8 +160,8 @@ def __init__(self):
     m = Model()
     for attr in Model.__slots__:
         assert object.__getattribute__(m, attr) == 'spam'
-        with pytest.raises(ValueError, match=f'"Model" object has no field "{attr}"'):
-            setattr(m, attr, 'not spam')
+        setattr(m, attr, 'not spam')
+        assert object.__getattribute__(m, attr) == 'not spam'
```
PrettyWood, line 164: @MrMrRobat I added both computed fields and raw descriptors support for `__set__`.
IMO it's consistent with a raw class but I'd be happy to have your point of view

MrMrRobat: Thanks for mentioning! At first glance, I don't quite understand why the behaviour changed there. I will take a closer look at this later today or in the next few days.

PrettyWood: It's due to my last commit 2ff0541

MrMrRobat: Hm, I see now, thanks. Looks like `if name in self.__private_attributes__:` line in `BaseModel.__setattr__` is redundant now. Not sure if it should be deleted though.

PrettyWood: Yep with the last commit to handle properly raw properties, the special cases for private attributes and computed fields become "almost" redundant. But I feel like we should keep them untouched for debugging and making the whole code more explicit

aiguofer:
Question, what would be the right way to add validation for a property?

Would this be possible?

```
class Thing(BaseModel):
    @property
    def prop(self):
        return self._thing

    @prop.setter
    def set_prop(self, val):
        self._thing = val

    @validator("prop", always=True)
    def validate_prop(cls, v, values, field):
        # do some validation
        return v
```

or would we just have to add the validation in the setter method?

PrettyWood:
@aiguofer It was not decided so I didn't want to make it too complicated. As you can see there are still discussions on the interface and the desired configuration. IMO the PR is already big enough and tackles most of the problem.
If it stays like this you'll need to add validation in the setter yes.
But your feedback is welcome on what would make more sense

aiguofer:
> @aiguofer It was not decided so I didn't want to make it too complicated. As you can see there are still discussions on the interface and the desired configuration. IMO the PR is already big enough and tackles most of the problem.
> If it stays like this you'll need to add validation in the setter yes.
> But your feedback is welcome on what would make more sense

The ability to define validator methods for properties would keep consistency with fields, which would be nice. That being said, adding validation in a setter is easy enough and satisfies all needs. I agree that keeping the PR to only the necessities is the best approach... if a good use case for using validator functions for properties arises later, that could be done in a separate PR.

PrettyWood:
Just resolved conflits. Please review whenever you can for the global approach🙏 There is still some debate on the API and I changed multiple times between `computed_field` and `field` and made my point [here](https://github.com/samuelcolvin/pydantic/pull/2625#issuecomment-813279617).
Once we all agree on the names and expected usage I'll work on the documentation ✍️

docs/usage/models.md
```diff
@@ -554,6 +554,34 @@ Where `Field` refers to the [field function](schema.md#field-customisation).
...
+```py
+{!.tmp_examples/models_computed_field_annotation.py!}
+```
+_(This script is complete, it should run "as is")_
+
+!!! note
+    If `computed_field` decorator doesn't wrap a decorator like `property`, `cached_property`, ...,
+    it will fallback and create `property` itself. This is shorter and may be what you want, but you
+    will probably lose IntelliSense in your IDE and have some warnings.
+    This is why it is recommanded to combine `computed_field` with `property`<br>
```
johtso, line 577: recommanded -> recommended

smarie:
I find this idea very nice. 

 - As already mentioned by others, it would be great if we could control if such a field appears or not in the repr. 
 - Could we imagine that `ComputedField` can receive the class ? In subclassing scenarii that could be worth having. This makes me think that maybe `ComputedField` could be generalized into some kind of `FieldFactory`, that could redefine the field description, type, default value, and other json props, based on the subclass. Or maybe this is something for a separate PR ?

A particular use-case where I see interest is to manually disambiguate `Union` of types that are all subclasses of a common parent basemodel. Using a constant field in the parent class, computed based on the subclass type, does the trick in an elegant way without needing the subclasses to redefine the field themselves. 
(of course PR #2336 is a more integrated alternative, but the computed field could be a good first stab, and I'm sure that it is representative of other similar usage scenarii)

nhuray:
Thanks @PrettyWood for that contribution. 🚀 

Is there's something we can do to have that PR merge any time soon ?

PrettyWood:
> As already mentioned by others, it would be great if we could control if such a field appears or not in the repr.
`repr` has been added. Feedback welcome

> Could we imagine that ComputedField can receive the class ?
Let's wait for some feedback. This PR is already quite huge and a lot may change for v2. But interesting idea thanks!

@nhuray A 1.9 will be done in September. This will most likely be in it

smarie:
Thanks @PrettyWood !

jordi-reinsma (commented):
How would this apply to async methods/properties?

MrMrRobat:
> How would this apply to async methods/properties?

Could you elaborate? AFAIK, there is no async properties in Python.

orfisko:
Guys, can we please get on with this ? :). I miss that @property support really hard!

lucasrabaca:
@MrMrRobat if you need to do an async operation like a database (tortoiseORM) and want to fetch some additional data that has no relation with the model, most orm's don't know how to do it, so async properties could make it work.

e.g:
```
async def get_data_from_a_table(self) -> Optional[OrmModelSchema]:
    return await OrmModel.get(id=1)
```
pydantic doesn't wait for it, so is returned "None" and raises a warning: Coroutine object was never waited.

naught101:
Doc link in first line is broken.

ichorid:
> What is the value of this over just using a property by itself? 

Well, one reason is to define some complex actions that should be taken when _changing_ a value. For instance, [converting a full filesystem path to a relative one](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/config/tribler_config_section.py). Or adding a fixed prefix to a relative path when _reading_ a value.

caniko:
> @PrettyWood What needs to happen for this great PR to be merged?

It is still awaiting review.

caniko:
If you are on poetry, and can't wait:
```
pydantic = { git = "https://github.com/PrettyWood/pydantic.git", branch = "field-property" }
```

caniko:
Crash when abstract classes with compute fields are inherited.
Test code:
```
from abc import ABC

from pydantic import BaseModel, computed_field


class AbstractBase(BaseModel, ABC):
    ren: int

    @computed_field
    @property
    def test_compute_field_field(self) -> int:
        return self.ren * 5


class ConcreteEnd(AbstractBase):
    reni: float
```
Output:
```
  File "/home/can/Software_Projects/BiKiPy/examples/test.py", line 15, in <module>
    class ConcreteEnd(AbstractBase):
  File "pydantic/main.py", line 144, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/utils.py", line 631, in pydantic.utils.smart_deepcopy
  File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 161, in deepcopy
    rv = reductor(4)
TypeError: 'NoneType' object is not callable
```

samuelcolvin (commented):
Over all this looks awesome. I want to use it. Thank you again @PrettyWood for all your effort. :pray: 

My main concern is around performance, particularly performance where compute fields are not being used but the extra checks/logic being used here is still called.

Roughly what performance impact does this have?

The main question I have in terms of implemenation is whether computed field results should be stored in the `__computed_fields__` dict rather than in `__dict__`? Also whether `__computed_fields__` should be separate from `__fields__`.

Would it be possible for computed fields to just be more fields at the end of `__fields__` (end so they could get access to `values`)? If so I think this would significantly improve performance.

In other words, what if computed fields were really just syntactic sugar for

```py
class Rectangle(BaseModel):
    width: int
    length: int
    area: int = None

    @validator('area', always=True, pre=True)
    def calc_area(cls, v, values):
        if (width := values.get('width')) and (length := values.get('length')):
            return width * length
```
?

Other reminders:
* @caniko has a comment about inheritance
* conflicts

samuelcolvin:
please update.

corleyma:
> Over all this looks awesome. I want to use it. Thank you again @PrettyWood for all your effort. 🙏
> 
> My main concern is around performance, particularly performance where compute fields are not being used but the extra checks/logic being used here is still called.
> 
> Roughly what performance impact does this have?
> 
> The main question I have in terms of implemenation is whether computed field results should be stored in the `__computed_fields__` dict rather than in `__dict__`? Also whether `__computed_fields__` should be separate from `__fields__`.
> 
> Would it be possible for computed fields to just be more fields at the end of `__fields__` (end so they could get access to `values`)? If so I think this would significantly improve performance.
> 
> In other words, what if computed fields were really just syntactic sugar for
> 
> ```python
> class Rectangle(BaseModel):
>     width: int
>     length: int
>     area: int = None
> 
>     @validator('area', always=True, pre=True)
>     def calc_area(cls, v, values):
>         if (width := values.get('width')) and (length := values.get('length')):
>             return width * length
> ```
> 
> ?
> 
> Other reminders:
> 
> * @caniko has a comment about inheritance
> * conflicts

One potential drawback with the "computed fields are just fields/sugar" approach would be that it would hint to tooling (e.g. IDEs) that the computed field is a valid/expected attribute when initializing a model instance, when in reality providing a value at initialization time (instead of computing it from the dependencies) should probably be an error.  So however this is implemented, it would be ideal if it can be done in a way that plays nicely with IDEs and the like.

caniko:
Could we make sure that these fields are compatible with `frozen`?

kozzza:
Why hasn't this feature been merged yet? What's the best work-round right until this gets added?

caniko:
> Why hasn't this feature been merged yet? What's the best work-around right now until this gets added?

Use `cached property`, and this in you `BaseModel`:
```
    class Config:
        keep_untouched = (cached_property,)
```

DCMattyG:
I'm testing this out right now, and @computed_field is a dream come true!

I did notice that when using "response_model=<Model>" with FastAPI routes, any Pydantic class using @computed_field causes FastAPI to throw the following error and crash:

TypeError: 'NoneType' object is not callable

Perhaps something I'm missing in the configuration? Simple example:

@computed_field
def running(self) -> bool:
    return self._running

Bobronium:
@DCMattyG, could you perhaps share a complete code snipped that can be ran as is and the full traceback it produces?

DCMattyG:
> @DCMattyG, could you perhaps share a complete code snipped that can be ran as is and the full traceback it produces?

This is the exact same issue that @caniko shared above, which includes a complete code snippet.

TL:DR, if you have a @computed_field on your Pydantic class, and you set a response_model, you get the following error:

TypeError: 'NoneType' object is not callable

Example of attaching the Pydantic class to a route:

@router.get("/{name}", response_model=Profile)
async def get_profile(name):
    for profile in PROFILES:
        if(profile.name == name):
            return profile

caniko:
Another request: `computed_fields` should also be available for validation, `validate("<computed_field>")`. Validating all of computed fields might be computationally expensive, consider adding a config flagg to enable this (default: false; disabled).

The feature might be a little outside the scope of the current PR. Also, considering the newly suggested onion protocol for validation might also get outdated, food for thought!

caniko:
@samuelcolvin regarding the syntactic sugar approach using the `validator` decorator. I couldn't see any caching in the provided example:

```python
class Rectangle(BaseModel):
    width: int
    length: int
    area: int = None

    @validator('area', always=True, pre=True)
    def calc_area(cls, v, values):
        if (width := values.get('width')) and (length := values.get('length')):
            return width * length
```
I don't see this being both syntactic sugar, and have caching without introducing several lines of code. Having no caching as the given example is definitely a waste of compute.

@PrettyWood, what is your opinion?

alemor10:
will this be updated?

Spenhouet:
Since it is unclear when and if this will ever be merged, I wrote my own computed fields function (not for the properties, just for fields which are supposed to be computed (using other model fields).

```python
from typing import Any, Callable, Dict

from pydantic import Field, validator


def ComputedField() -> Any: # NOSONAR # Since this simulates a class, the class naming is used.
    """Wrap the pydantic field to provide None as default value."""
    return Field(None)


def computed_field(name: str):
    """
    Python Decorator for computed fields.

    Args:
        name: The name of the pydantic model field which should be computed.
    """

    def wrapper(func: Callable):
        required = [arg for arg in func.__annotations__.keys() if arg not in ['cls', 'return']]
        func_return_type = func.__annotations__.get('return')

        @validator(name, always=True, allow_reuse=True)
        def inner(cls, v, values: Dict, field):
            # Check that the computed field is unassigned.
            if v is not None:
                raise RuntimeError(f"Field '{name}' is computed and should not be provided.")

            # Check that all fields are defined.
            missing_fields = [r for r in required if r not in values]
            if missing_fields:
                raise RuntimeError(f"The fields {missing_fields} must be defined in model.")

            expected_type = field.type_
            # Check that return type of defined function matches model.
            if func_return_type != expected_type:
                raise TypeError(f"Return type should match type {expected_type} of '{name}' field.")

            computed_value = func(cls, **{r: values[r] for r in required})

            # Check that type of computed value matches model.
            if not isinstance(computed_value, expected_type):
                raise TypeError(f"Computed value should match type {expected_type} of '{name}' field.")

            return computed_value

        return inner

    return wrapper
```

Here some tests to show how it works / can be used:

```python
import unittest

import xmlrunner
from airamed_utils.pydantic_utils import ComputedField, computed_field


class TestComputedField(unittest.TestCase):
    """Test the computed field decorator."""

    def test_sets_computed_value(self):
        """Test that setting a computed field via the compute method works."""
        expected = 'test_value'

        class TestModel(BaseModel):
            computed: str = ComputedField()

            @computed_field('computed')
            def compute_computed(cls) -> str:
                return expected

        test_model = TestModel()

        self.assertEqual(test_model.computed, expected)

    def test_reject_provided_value(self):
        """Test that setting a computed field via the model constructor is not allowed."""

        class TestModel(BaseModel):
            computed: str = ComputedField()

            @computed_field('computed')
            def compute_computed(cls) -> str:
                return 'test'

        with self.assertRaises(RuntimeError):
            TestModel(computed='should not be allowed')

    def test_attribute_injection(self):
        """Test that the injection of model field via parameters works."""

        class TestModel(BaseModel):
            injected: int
            computed: str = ComputedField()

            @computed_field('computed')
            def compute_computed(cls, injected: int) -> str:
                return str(injected)

        test_model = TestModel(injected=42)

        self.assertEqual(test_model.injected, 42)
        self.assertEqual(test_model.computed, '42')

    def test_require_non_existing_field(self):
        """Test that requiring a non existing model field does not work."""

        class TestModel(BaseModel):
            computed: str = ComputedField()

            @computed_field('computed')
            def compute_computed(cls, non_existing_field: str) -> str:
                return 'test'

        with self.assertRaises(RuntimeError):
            TestModel()

    def test_wrong_type_annotation(self):
        """Test that using a non matching type annotation does not work."""

        class TestModel(BaseModel):
            computed: str = ComputedField()

            @computed_field('computed')
            def compute_computed(cls) -> int:
                return 42

        with self.assertRaises(ValidationError):
            TestModel()

    def test_missing_type_annotation(self):
        """Test that using no type annotation does not work."""

        class TestModel(BaseModel):
            computed: str = ComputedField()

            @computed_field('computed')
            def compute_computed(cls):
                return 42

        with self.assertRaises(ValidationError):
            TestModel()

    def test_wrong_return_type(self):
        """Test that returning data with a non matching type does not work."""

        class TestModel(BaseModel):
            computed: str = ComputedField()

            @computed_field('computed')
            def compute_computed(cls) -> str:
                return 42  # type: ignore # NOSONAR # This is on purpose here.

        with self.assertRaises(ValidationError):
            TestModel()

    def test_missing_attribute_assignment(self):
        """Test that the ComputedField assignment is required."""

        class TestModel(BaseModel):
            computed: str  # Missing ComputedField() assignment

            @computed_field('computed')
            def compute_computed(cls) -> str:
                return 'test'

        with self.assertRaises(ValidationError):
            TestModel()


if __name__ == "__main__":
    unittest.main(testRunner=xmlrunner.XMLTestRunner(output='test-reports'))
```

samuelcolvin:
This will be merged.

It'll go in the next minor release.

pydantic/fields.py
```diff
@@ -310,6 +319,239 @@ def Field(
...
+            repr=repr,
+        )
+
+    if func is None:
+        return wrap
+
+    return wrap(func)
+
+
+class GetAttributeDict:
```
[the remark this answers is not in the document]
yicone: > Crash when compute fields are inherited. There should be at least one test for inheritance.
> 
> Test code:
> 
> ```
> from pydantic import BaseModel, computed_field
> 
> 
> class Base(BaseModel):
>     ren: int
> 
>     @computed_field
>     @property
>     def test_compute_field_field(self) -> int:
>         return self.ren * 5
> 
> 
> class ConcreteEnd(Base):
>     reni: float
> ```
> 
> Output:
> 
> ```
> Traceback (most recent call last):
>   File "/home/can/Software_Projects/BiKiPy/examples/test.py", line 13, in <module>
>     class ConcreteEnd(Base):
>   File "pydantic/main.py", line 144, in pydantic.main.ModelMetaclass.__new__
>   File "pydantic/utils.py", line 631, in pydantic.utils.smart_deepcopy
>   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 146, in deepcopy
>     y = copier(x, memo)
>   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 230, in _deepcopy_dict
>     y[deepcopy(key, memo)] = deepcopy(value, memo)
>   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 161, in deepcopy
>     rv = reductor(4)
> TypeError: 'NoneType' object is not callable
> ```
> 
> System info:
> 
> ```
>              pydantic version: 1.8.2
>             pydantic compiled: True
>                  install path: /home/can/.cache/pypoetry/virtualenvs/bikipy-Cb5pIUSo-py3.9/lib/python3.9/site-packages/pydantic
>                python version: 3.9.7 (default, Oct 21 2021, 08:33:25)  [GCC 11.1.0]
>                      platform: Linux-5.15.2-zen1-1-zen-x86_64-with-glibc2.33
>      optional deps. installed: ['typing-extensions']
> ```

My temporary solution.
fix `fields.py`:
```
class ComputedField(Representation):
      ...
   
     def __getattr__(self, name: str) -> Any:
        ...

        raise AttributeError(f'{self.__class__.__name__} object has no attribute {name}')
```

Reference: https://stackoverflow.com/a/49902096/251496

yicone:
> Crash when compute fields are inherited. There should be at least one test for inheritance.
> 
> Test code:
> 
> ```
> from pydantic import BaseModel, computed_field
> 
> 
> class Base(BaseModel):
>     ren: int
> 
>     @computed_field
>     @property
>     def test_compute_field_field(self) -> int:
>         return self.ren * 5
> 
> 
> class ConcreteEnd(Base):
>     reni: float
> ```
> 
> Output:
> 
> ```
> Traceback (most recent call last):
>   File "/home/can/Software_Projects/BiKiPy/examples/test.py", line 13, in <module>
>     class ConcreteEnd(Base):
>   File "pydantic/main.py", line 144, in pydantic.main.ModelMetaclass.__new__
>   File "pydantic/utils.py", line 631, in pydantic.utils.smart_deepcopy
>   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 146, in deepcopy
>     y = copier(x, memo)
>   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 230, in _deepcopy_dict
>     y[deepcopy(key, memo)] = deepcopy(value, memo)
>   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 161, in deepcopy
>     rv = reductor(4)
> TypeError: 'NoneType' object is not callable
> ```
> 
> System info:
> 
> ```
>              pydantic version: 1.8.2
>             pydantic compiled: True
>                  install path: /home/can/.cache/pypoetry/virtualenvs/bikipy-Cb5pIUSo-py3.9/lib/python3.9/site-packages/pydantic
>                python version: 3.9.7 (default, Oct 21 2021, 08:33:25)  [GCC 11.1.0]
>                      platform: Linux-5.15.2-zen1-1-zen-x86_64-with-glibc2.33
>      optional deps. installed: ['typing-extensions']
> ```

My temporary solution.
fix fields.py:

```
class ComputedField(Representation):
      ...
   
     def __getattr__(self, name: str) -> Any:
        ...

        raise AttributeError(f'{self.__class__.__name__} object has no attribute {name}')
```

Reference: https://stackoverflow.com/a/49902096/251496

caniko:
> > Crash when compute fields are inherited. There should be at least one test for inheritance.
> > Test code:
> > ```
> > from pydantic import BaseModel, computed_field
> > 
> > 
> > class Base(BaseModel):
> >     ren: int
> > 
> >     @computed_field
> >     @property
> >     def test_compute_field_field(self) -> int:
> >         return self.ren * 5
> > 
> > 
> > class ConcreteEnd(Base):
> >     reni: float
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Output:
> > ```
> > Traceback (most recent call last):
> >   File "/home/can/Software_Projects/BiKiPy/examples/test.py", line 13, in <module>
> >     class ConcreteEnd(Base):
> >   File "pydantic/main.py", line 144, in pydantic.main.ModelMetaclass.__new__
> >   File "pydantic/utils.py", line 631, in pydantic.utils.smart_deepcopy
> >   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 146, in deepcopy
> >     y = copier(x, memo)
> >   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 230, in _deepcopy_dict
> >     y[deepcopy(key, memo)] = deepcopy(value, memo)
> >   File "/home/can/.pyenv/versions/3.9.7/lib/python3.9/copy.py", line 161, in deepcopy
> >     rv = reductor(4)
> > TypeError: 'NoneType' object is not callable
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > System info:
> > ```
> >              pydantic version: 1.8.2
> >             pydantic compiled: True
> >                  install path: /home/can/.cache/pypoetry/virtualenvs/bikipy-Cb5pIUSo-py3.9/lib/python3.9/site-packages/pydantic
> >                python version: 3.9.7 (default, Oct 21 2021, 08:33:25)  [GCC 11.1.0]
> >                      platform: Linux-5.15.2-zen1-1-zen-x86_64-with-glibc2.33
> >      optional deps. installed: ['typing-extensions']
> > ```
> 
> My temporary solution. fix fields.py:
> 
> ```
> class ComputedField(Representation):
>       ...
>    
>      def __getattr__(self, name: str) -> Any:
>         ...
> 
>         raise AttributeError(f'{self.__class__.__name__} object has no attribute {name}')
> ```
> 
> Reference: https://stackoverflow.com/a/49902096/251496

I am not sure if this solves the problem. Does this mean the problem is in Python, and not the implementation?

PrettyWood:
@caniko Don't worry I'll get back to this PR when I have time. @samuelcolvin didn't mean it would be merged as is

yicone:
> I am not sure if this solves the problem, @yicone. Does this mean the problem is in Python, and not the implementation?

I have changed my previous comment, there are no exception are expected.

pawelswiecki:
Thank you for working on this PR so much 🙇 

Two questions.

**1.** Will I be able to make a field both computed and abstract?

E.g.:

```python
class Shape(BaseModel, ABC):
    @computed_field
    @property  # is this decorator needed when computed_field is already here?
    @abstractmethod  # will this work as expected?
    def area(self) -> float:
        pass

class Square(Shape):
    side: float

    def area(self) -> float:
        return self.side ** 2
```

**2.** Will I be able to alias computed field names? Context: I need to use `type` field name. However, in the case of defining the field as a `@property` I get `NameError: Field name "type" shadows a BaseModel attribute; use a different field name with "alias='type'"`. I'm talking here from the perspective of Pydantic 1.9.0.

Looking forward to using the computed fields feature, it's super-useful!

caniko:
> 1. Will I be able to make a field both computed and abstract?

But why would you do this? I would just use abstract property.

pawelswiecki:
Thank you for answering.

> But why would you do this? I would just use abstract property without including the computed_field.

My use-case:

```python
from enum import Enum

class EventType(Enum):
    JOB_STARTED =  "job_started"
    JOB_FINISHED =  "job_finished"

class BaseEvent(BaseModel):
    id: str
    event_type: EventType
 
    @classmethod
    def create_event(cls) -> BaseEvent:
        id_ = "generated id"
        return BaseEvent(
            id=id_,
            event_type=...,  # without this parameter mypy correctly complains, but what goes here?
        )
    
class JobStartedEvent(BaseEvent):
    event_type = Field(EventType.JOB_STARTED, const=True)

class JobFinishedEvent(BaseEvent):
    event_type = Field(EventType.JOB_FINISHED, const=True)
```

The `const=True` part is the key -- I want to make sure each BaseEvent's subclass won't be instantiated with an incorrect EventType (I only need to remember to include the `const=True` on each subclass).

I tried different combinations of `@abstractmethod`, `@classmethod` and `@property`, and every time something came up that made the whole thing broken on some level (either mypy was complaining, or there was a runtime issue, or it was incompatible with pydantic). When I saw this PR I thought something like this might be the solution:

```python
from abc import ABC, abstractmethod
from enum import Enum

from pydantic import BaseModel, Field, computed_field

class EventType(Enum):
    JOB_STARTED =  "job_started"
    JOB_FINISHED =  "job_finished"

class BaseEvent(BaseModel, ABC):
    id: str
    event_type: EventType
 
    @computed_field
    @property
    @abstractmethod
    def type(cls) -> EventType:
        pass

    @classmethod
    def create_event(cls) -> BaseEvent:
        id_ = "generated id"
        return BaseEvent(id=id_)
    
class JobStartedEvent(BaseEvent):
    @computed_field
    @property
    def type(cls) -> EventType:
        return EventType.JOB_STARTED

class JobFinishedEvent(BaseEvent):
    @computed_field
    @property
    def type(cls) -> EventType:
        return EventType.JOB_FINISHED
```

Another thing is aliasing it -- `type` field won't work in Pydantic, as I've mentioned above.

pawelswiecki:
I'm aware this usage of `@computed_field` is not the intended purpose of it as no computation is really being made. I'm probably just grasping at straws here.

adriangb:
> Also, you don't want to use@abstractmethod. It is deprecated in the newer versions. :)

[`abc.abstractmethod`](https://docs.python.org/3/library/abc.html#abc.abstractmethod) is _not_ deprecated. [`abc.abstractproperty`](https://docs.python.org/3/library/abc.html#abc.abstractproperty) _is_ deprecated.

pawelswiecki:
>[abc.abstractmethod](https://docs.python.org/3/library/abc.html#abc.abstractmethod) is not deprecated. [abc.abstractproperty](https://docs.python.org/3/library/abc.html#abc.abstractproperty) is deprecated.

I know that. I've never mentioned it :)

pawelswiecki:
So, getting back to being able to use `@computed_field` together with `@abstractmethod`. I think it's a reasonable thing -- ABC and abstract method is a well-established part of Python and I'm pretty sure multiple cases are going to emerge.

Mind I have no idea how time-consuming implementing this would be.

yicone:
When reference a property with @comuted_field annotation, the type hint of the reference is gone.

I'm not sure if the bug comes from this branch or the `Python` extension of VS Code currently.

E.g.:
```py
@property
def A(self) -> int:
	print(self.B)  # actual is "(method) B: Any", expect is "(property) B: List[float]"
	return 1

@computed_field
@property
def B(self) -> List[float]:
	return []
```

Editor version info:
> Visual Stuido Code: 1.67.2 
         with `Python` extension : v2022.6.2

Pydantic version info:
>
             pydantic version: 1.9.0
            pydantic compiled: False
                 install path: /Users/Tr/Workspace/yicone/pydantic/pydantic
               python version: 3.9.12 (main, Mar 26 2022, 15:51:15)  [Clang 13.1.6 (clang-1316.0.21.2)]
                     platform: macOS-12.2-x86_64-i386-64bit

david-saeger:
any reason there hasnt been movement on this in a while? are we ever going to get it?

Bobronium:
@yicone, can reproduce. Also can fix by replacing `Any` to `T` in annotation for `func` argument and return annotation of `computed_field`.

pydantic/fields.py
```diff
@@ -310,6 +319,239 @@ def Field(
...
+    exclude: Optional['ExcludeInclude'] = None,
+    include: Optional['ExcludeInclude'] = None,
+    repr: bool = True,
+) -> 'Callable[[Any], ComputedField]':
+    ...
+
+
+@overload
+def computed_field(
+    func: Any,
```
Bobronium, line 361:
```suggestion
    func: T,
```

pydantic/fields.py
```diff
@@ -310,6 +319,239 @@ def Field(
...
+    *,
+    alias: Optional[str] = None,
+    title: Optional[str] = None,
+    description: Optional[str] = None,
+    exclude: Optional['ExcludeInclude'] = None,
+    include: Optional['ExcludeInclude'] = None,
+    repr: bool = True,
+) -> Any:
```
Bobronium, line 369:
```suggestion
) -> T:
```

pydantic/fields.py
```diff
@@ -310,6 +319,239 @@ def Field(
...
+    ...
+
+
+def computed_field(
+    func: Optional[Any] = None,
```
Bobronium, line 374:
```suggestion
    func: Optional[T] = None,
```

pydantic/fields.py
```diff
@@ -310,6 +319,239 @@ def Field(
...
+    *,
+    alias: Optional[str] = None,
+    title: Optional[str] = None,
+    description: Optional[str] = None,
+    exclude: Optional['ExcludeInclude'] = None,
+    include: Optional['ExcludeInclude'] = None,
+    repr: bool = True,
+) -> 'Union[Callable[[Any], ComputedField], ComputedField]':
```
Bobronium, line 382:
```suggestion
) -> 'Union[Callable[[T], ComputedField], ComputedField]':
```

Bobronium (changes requested):
Fixes issue described in https://github.com/samuelcolvin/pydantic/pull/2625#issuecomment-1141740503

Tested on VSCode, please also confirm that there's no regression in PyCharm.

Bobronium:
@pawelswiecki I'd suggest not to dig down into the rabbit hole when it comes to the changes that can be made after the release. I think changes should be gradual and easy to make. Adding `@abstractmethod` is possible after releasing this. Removing such support is not so trivial, since it would introduce breaking changes.

Let's focus on things that can't be easily changed/fixed after release.

<details><summary>As for your example, let me suggest this (click here):</summary>

```py
from enum import Enum
from typing_extensions import Self

class EventType(Enum):
    JOB_STARTED =  "job_started"
    JOB_FINISHED =  "job_finished"

class BaseEvent(BaseModel):
    id: str
    event_type: EventType

    def __init_subclass__(cls, **kwargs) -> None:
        return super().__init_subclass__(**kwargs)
        if (event_type_field := cls.__fields__.get('event_type')) is None or not event_type_field.field_info.const:
            raise NotImplementedError('You must define event_type for concrete event')

    @classmethod
    def create(cls, **data) -> Self:
        id_ = "generated id"
        return cls(id=id_, **data)

class JobStartedEvent(BaseEvent):
    event_type = Field(EventType.JOB_STARTED, const=True)

class JobFinishedEvent(BaseEvent):
    event_type = Field(EventType.JOB_FINISHED, const=True)


print(JobFinishedEvent.create().event_type)  # EventType.JOB_FINISHED


class InvalidEvent(BaseEvent):  ...  # NotImplementedError: You must define event_type for concrete event
```

Bobronium:
To summarize, things to do before merging:
- [ ] Fix typing: https://github.com/samuelcolvin/pydantic/pull/2625#pullrequestreview-649988316
- [ ] Fix inheritance: https://github.com/samuelcolvin/pydantic/pull/2625#issuecomment-1093799303
  - [ ] raise `AttributeError` in `ComputedField.__getattr__` when condition is not met
  - [ ] add tests that include inheritance
- [ ] Decide whether to support `@abstractmethod` (IMO, no, see https://github.com/samuelcolvin/pydantic/pull/2625#issuecomment-1150460832)

Is there something else I'm missing?

PrettyWood:
@Bobronium thanks for the recap. I'll work on it this weekend. Sorry for being inactive, just a lot on my plate with both professional and personal matters

Bobronium:
Can relate. Please take care!

Also, feel free to ask for any help if needed.

caniko:
Suggestion:

I don't disagree with this, and think it makes most sense:
```python
...
    @computed_field
    @cached_property
    def hard_compute():
        <very hard problem>
...
```
However, I suggest that we also introduce shorthand notation for `property` and `cached_property`:
```python
    @cached_computed_field_property
    ...
```
And:
```python
    @computed_field_property
    ...
```
This is nice when `computed_fields` are used frequently in a package, like my case. I don't want to end up in decorator-hell. Hope this makes sense.

Another question:
Would this `computed_field`s work with `numba.jit` decorator?

caniko:
Feature suggestion:
Can we make `computed_field`s work with `numba.jit` decorator? Where it `property` and `cached_property` is optional?

```
@computed_field
@jit
def some_func(self):
    ...
```

bjmc:
`cached_computed_field_property` is rather long.

What about something like?

```
@computed_field(cache=True)
def calculate(self):
    pass
```

caniko:
@bjmc, that is also an option. Where `property` is the default.

samuelcolvin:
This won't be added for v1.10. I don't know if we can use some of this logic in V2 or whether we'll need a new implementation.

pawelswiecki:
@Bobronium Thank you for your suggestion, with the working code, much appreciated!

PrettyWood:
For those waiting for this feature, it will be 100% added in v2!
Even though we may change the implementation, we got a lot of precious feedbacks (thanks everyone ❤️) on the desired APIs: a "property-like" one and a more "type annotation" one.
So we will most likely have [these examples](https://github.com/pydantic/pydantic/pull/2625#issue-849516427) work the same 👍

Bobronium:
Wow. Looks like pylance (pyright) supports custom defined properties through simple descriptor typing.

I've managed to create a `computed_field` type that is interpreted by VSCode as a property (borrowed code for generic property here: https://github.com/python/typing/issues/985#issuecomment-1002770338):

```py
from random import randint
from typing import Any, Callable, Generic, NoReturn, TypeVar, overload



_G1 = TypeVar("_G1")
_G2 = TypeVar("_G2")
_S1 = TypeVar("_S1")
_S2 = TypeVar("_S2")


class computed_field(Generic[_G1, _S1]):
    fget: Callable[[Any], _G1] | None
    fset: Callable[[Any, _S1], None] | None
    fdel: Callable[[Any], None] | None

    @overload
    def __new__(
        cls,
        fget: Callable[[Any], _G2],
        fset: None = ...,
        fdel: Callable[[Any], None] | None = ...,
        doc: str | None = ...,
    ) -> computed_field[_G2, NoReturn]:
        ...

    @overload
    def __new__(
        cls,
        fget: Callable[[Any], _G2],
        fset: Callable[[Any, _S2], None],
        fdel: Callable[[Any], None] | None = ...,
        doc: str | None = ...,
    ) -> computed_field[_G2, _S2]:
        ...

    @overload
    def __new__(
        cls,
        fget: None = ...,
        fset: None = ...,
        fdel: Callable[[Any], None] | None = ...,
        alias: str | None = ...,
        title: str | None = ...,
        description: str | None = ...,
        exclude: set[str] | None = ...,
        include: set[str] | None = ...,
        repr: bool = True,
        doc: str | None = ...,
    ) -> computed_field[NoReturn, NoReturn]:
        ...

    def __call__(self: computed_field[NoReturn, NoReturn], fget: Callable[[Any], _G2]) -> computed_field[_G2, NoReturn]:
        ...

    def getter(self, __fget: Callable[[Any], _G2]) -> computed_field[_G2, _S1]:
        ...

    def setter(self, __fset: Callable[[Any, _S2], None]) -> computed_field[_G1, _S2]:
        ...

    def deleter(self, __fdel: Callable[[Any], None]) -> computed_field[_G1, _S1]:
        ...

    def __get__(self, __obj: Any, __type: type | None = ...) -> _G1:
        ...

    def __set__(self, __obj: Any, __value: _S1) -> None:
        ...

    def __delete__(self, __obj: Any) -> None:
        ...


class Square:
    side: float

    @computed_field
    def area(self) -> float:
        return self.side**2

    @computed_field(alias="The random number")
    def random_n(self) -> int:
        return randint(0, 1_000)


reveal_type(Square().area)
reveal_type(Square().random_n)
```

```py
Type of "Square().area" is "float"
Type of "Square().random_n" is "int"
```

Unfurtunately, both mypy and pycharm still don't know how to handle (generic) descriptors.

caniko:
Another thing that I thought of is to include a keyword argument for computing all computed fields when running `model.dict()` or `model.json()`.

Example:
```python
model.dict(define_computed_fields=True)
```
Will evaluate and provide `dict` with all computed fields _computed_.

nickswiss:
@PrettyWood Any update on this feature, think it is a great addition and gets rid of accomplishing this in validators. Thank you!

samuelcolvin:
Thank you all (especially @PrettyWood) for 2 years (! :see_no_evil:) of patience.

I've migrated this feature to pydantic V2 and pydantic-core, see #5502.

Thank you so much @PrettyWood for your work on this, I've used some of your code, and most of your tests and docs in the new PR :bow: :pray:.

PR #2626, opened 2021-04-03
Update install.md

promsoft:
There must be a typo here - two external dependencies, not three.
It it trivial change.


PrettyWood:
Indeed thanks!

PR #2670, opened 2021-04-13
fix: stop calling parent class `root_validator` if overridden

### LINKED ISSUE #1895: Root validator called in superclass even if overridden in a subclass without calling super
# Bug

```
             pydantic version: 1.6.1
            pydantic compiled: True
                 install path: /Users/jakub/.virtualenvs/server/lib/python3.8/site-packages/pydantic
               python version: 3.8.1 (default, Mar 13 2020, 20:31:03)  [Clang 11.0.0 (clang-1100.0.33.17)]
```

Methods decorated with `root_validator` that are defined in a subclass do not override the method defined in the superclass. For methods decorated with `validator` they do override the method defined in the superclass properly.

Context:
I want to have a model that declares a bunch of fields and some default validators for these fields, which should be applied in every subclass, unless overridden. That works for fields decorated with `validator`, but for fields decorated with `root_validator` the validator in superclass gets called first.

I couldn't find anything in the docs to disable calling the superclass validator when using the `root_validator` decorator. Also, I looked into the generic models (it seems that it would fit composition better than inheritance) and reusing validators (doesn't seem to be a good candidate to validate field contents because in my case the validation depends on the `exercise_type` and I would still have to define something like `_normalized_choices` in every subclass).

```py

from pydantic import BaseModel, root_validator, validator
from typing import Union, List, Optional


class ExerciseBase(BaseModel):

    text: Optional[str] = None
    choices: Optional[List[List[str]]] = None

    @root_validator
    def validate_choices(cls, values):
        print("ExerciseBase validate_choices")
        choices = values.get('choices')
        assert choices is None, 'Choices not allowed for this exercise type'
        return values


class FillBlanksExercise(ExerciseBase):
    @root_validator
    def validate_choices(cls, values):
        print("FillBlanksExercise validate_choices")
        choices = values.get('choices')
        if choices is not None:
            text = values.get('text')
            number_of_blanks = text.count('____')
            assert {len(choice) for choice in choices} == {number_of_blanks}, 'Each choice must match number of blanks'
        return values

EXERCISE_TYPES_MAPPING = {
    'FillBlanksExercise': FillBlanksExercise,
    'ChoiceExercise': ChoiceExercise,
}

exercise_data = {
    'text': 'Python is a type of ____?',
    'exercise_type': 'FillBlanksExercise',
    'choices': [['snake'], ['bird'], ['fish'], ['dinosaur']],
}

exercise = FillBlanksExercise(**exercise_data)


# ExerciseBase validate_choices
# FillBlanksExercise validate_choices
#   File "/Users/jakub/Library/Application Support/JetBrains/PyCharm2020.2/scratches/scratch_2.py", line 85, in <module>
#     exercise = ExerciseBase.build(**exercise_data)
#   File "/Users/jakub/Library/Application Support/JetBrains/PyCharm2020.2/scratches/scratch_2.py", line 22, in build
#     return _class(**kwargs)
#   File "pydantic/main.py", line 346, in pydantic.main.BaseModel.__init__
# pydantic.error_wrappers.ValidationError: 1 validation error for FillBlanksExercise
# __root__
#   Choices not allowed for this exercise type (type=assertion_error)
```

PrettyWood:
## Change Summary
We would keep all the root validators even those that were supposed to be overridden.
Now they actually are

## Related issue number
fix #1895

## Checklist

### DIFF
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -279,7 +279,7 @@ T = TypeVar('T')
 def unique_list(
     input_list: Union[List[T], Tuple[T, ...]],
     *,
-    get_name: Callable[..., str] = lambda x: str(x),
+    name_factory: Callable[[T], str] = str,
 ) -> List[T]:
     """
     Make a list unique while maintaining order.
@@ -289,7 +289,7 @@ def unique_list(
     result: List[T] = []
     result_names: List[str] = []
     for v in input_list:
-        v_name = get_name(v)
+        v_name = name_factory(v)
         if v_name not in result_names:
             result_names.append(v_name)
             result.append(v)
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -277,16 +277,25 @@ def to_camel(string: str) -> str:
 T = TypeVar('T')
 
 
-def unique_list(input_list: Union[List[T], Tuple[T, ...]]) -> List[T]:
+def unique_list(
+    input_list: Union[List[T], Tuple[T, ...]],
+    *,
+    get_name: Callable[..., str] = lambda x: str(x),
+) -> List[T]:
     """
     Make a list unique while maintaining order.
+    We update the list if another one with the same name is set
+    (e.g. root validator overridden in subclass)
     """
-    result = []
-    unique_set = set()
+    result: List[T] = []
+    result_names: List[str] = []
     for v in input_list:
-        if v not in unique_set:
-            unique_set.add(v)
+        v_name = get_name(v)
+        if v_name not in result_names:
+            result_names.append(v_name)
             result.append(v)
+        else:
+            result[result_names.index(v_name)] = v
 
     return result
 
--- a/changes/1895-PrettyWood.md
+++ b/changes/1895-PrettyWood.md
@@ -1,0 +1,1 @@
+stop calling parent class `root_validator` if overridden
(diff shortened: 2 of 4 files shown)


pydantic/utils.py
```diff
@@ -277,16 +277,25 @@ def to_camel(string: str) -> str:
 T = TypeVar('T')
 
 
-def unique_list(input_list: Union[List[T], Tuple[T, ...]]) -> List[T]:
+def unique_list(
+    input_list: Union[List[T], Tuple[T, ...]],
+    *,
+    get_name: Callable[..., str] = lambda x: str(x),
```
samuelcolvin, line 283:
```suggestion
    name_factory: Callable[[T], str] = str,
```
?

samuelcolvin:
please update.

pydantic/utils.py
PrettyWood: No need to be sorry for the change. Way better name, default and typing 😄

PrettyWood:
Updated! Please review 🙏

jkozlowicz:
Thank you @PrettyWood for implementing this! As the person, who opened #1895, I am still waiting for this fix.

@samuelcolvin could you please take a look if this can already be merged? Is there anything related to this fix that I could help with ?

PR #2688, opened 2021-04-21
Ensure type correctness when using `EmailStr`

gherx:
## Change Summary

Hi,

Currently `EmailStr` is always converted to a `str` which leads to false-negatives during type checking, e.g. using mypy, and may cause non-obvious bugs when using `isinstance`, which is useful in combination with `Union`-types, e.g.:

```python
import pydantic as pyd

class Model(pyd.BaseModel):
  v: pyd.EmailStr

m = Model(v=pyd.EmailStr("test@example.com"))
# reveal_type(m.v) == `'pydantic.networks.EmailStr`
assert isinstance(m.v, pyd.EmailStr)  # this raises
```

This change always makes `EmailStr` an `EmailStr` and should be fully backward-compatible, as `EmailStr` is still an instance of `str`.

## Related issue number

None that I know of

## Checklist

### DIFF
--- a/changes/2688-gherx.md
+++ b/changes/2688-gherx.md
@@ -1,0 +1,1 @@
+Using `EmailStr` in a model now always returns an `EmailStr`
--- a/docs/usage/types.md
+++ b/docs/usage/types.md
@@ -446,7 +446,7 @@ _(This script is complete, it should run "as is")_
 
 `EmailStr`
 : requires [email-validator](https://github.com/JoshData/python-email-validator) to be installed;
-  the input string must be a valid email address, and the output is a simple string
+  the input string must be a valid email address, and the output is an `EmailStr` object which is also a simple string
 
 
 
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -359,8 +359,8 @@ class EmailStr(str):
         yield cls.validate
 
     @classmethod
-    def validate(cls, value: Union[str]) -> str:
-        return validate_email(value)[1]
+    def validate(cls, value: Union[str]) -> 'EmailStr':
+        return cls(validate_email(value)[1])
 
 
 class NameEmail(Representation):
--- a/tests/test_networks.py
+++ b/tests/test_networks.py
@@ -484,6 +484,17 @@ def test_email_str():
 
 
 @pytest.mark.skipif(not email_validator, reason='email_validator not installed')
+def test_email_str_type():
+    class Model(BaseModel):
+        v: EmailStr
+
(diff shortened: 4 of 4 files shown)


samuelcolvin:
I disagree with this approach for performance and flexibility the current performance is preferable.

To fix mypy, instead do

```py
if TYPE_CHECKING:
    EmailStr = str
else:
    class EmailStr:
        ...
```

please update.

gherx:
Hi @samuelcolvin,

thank you very much for your reply, I really appreciate it!

To be honest, I don't think your proposed solution fixes the actual problem, nor does mypy cause the trouble.

```python
    login: Union[pydantic.EmailStr, str]
```

A casual programmer using pydantic may be inclined to assume that `login` in the above code might be either an `EmailStr` or some other `str` (e.g. a user might be able to login with email or username). Elsewhere, the programmer may have to handle an `EmailStr` differently and does the following:

```python
def handle(m: MyModel):
    if isinstance(m.login, pydantic.EmailStr):
        # handle email login
    else:
        # handle username login
```

The above seems to be very reasonable and valid, but fails, as `EmailStr` lost all its type information and is just a `str`. So the if-branch is dead code. Mypy cannot (and should not) detect it, as types are not expected to change implicitly. 

Of course the programmer could have read the documentation word by word or the source code and check that detail, but this is not obvious and IMHO should simply be prevented by pydantic.

Regarding performance, I see that this PR requires to create a new `str` (which is short in practice, as it only happens for valid emails), but I don't see that this would have any real impact. Maybe you can tell in which order of magnitude you assume such performance losses to be?

Maybe you can elaborate on what you mean by "flexibility"? An `EmailStr` is still a `str` so I don't get the point here.

Finally, I would like to thank you for the work you have put and are putting into pydantic - Thank you and have a nice day!

PrettyWood:
@samuelcolvin I quite agree with @gherx. I don't think the impact on performance is a problem.
For consistency I find it quite disturbing to have
```py
from pydantic import AnyUrl, BaseModel, EmailStr


class M(BaseModel):
    x: AnyUrl
    y: EmailStr

m = M(x='http://test.com', y='v@lid.com')
assert isinstance(m.x, AnyUrl) is True
assert isinstance(m.y, EmailStr) is False
```
I let you answer

samuelcolvin:
Okay, I agree with this but to avoid any breaking changes, let's defer this until V2.

samuelcolvin:
As per #4516, `EmailStr` has become a true `str` in typing terms in V2.

We're using

```py
if TYPE_CHECKING:
    EmailStr = typing_extensions.Annotated[str, ...]
else:
    ....
```

Hence I'm closing this.

PR #2718, opened 2021-04-30
Fix recursive ORM parsing error

nuno-andre:
The _ORM mode_ doesn't parse dictionaries if it's not explicitly called.

## Change Summary

Prepend the `orm_mode` checking to the `dict`'s.

## Checklist

## Example

```python
from pydantic.utils import GetterDict

class Getter(GetterDict):
    # not relevant, just to show the behavior
    def get(self, key, default):
        key = key + key
        try:
            return self._obj[key]
        except TypeError:
            return getattr(self._obj, key, default)
        except KeyError:
            return default

class Model(BaseModel):
    class Config:
        orm_mode = True
        getter_dict = Getter

class ModelA(Model):
    a: int

class ModelB(Model):
    b: ModelA
```

```python
from types import SimpleNamespace

x = dict(bb=SimpleNamespace(aa=1))
ModelB.from_orm(x)
#> b=ModelA(a=1)

y = dict(bb=dict(aa=1))
ModelB.from_orm(y)
#> ...
#> b -> a
#>   field required (type=value_error.missing)
```

### DIFF
--- a/changes/2718-nuno-andre.md
+++ b/changes/2718-nuno-andre.md
@@ -1,0 +1,1 @@
+Change `orm_mode` checking to allow recursive ORM mode parsing with dicts.
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -725,10 +725,11 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
             return value.copy() if cls.__config__.copy_on_model_validation else value
 
         value = cls._enforce_dict_if_root(value)
-        if isinstance(value, dict):
-            return cls(**value)
-        elif cls.__config__.orm_mode:
+
+        if cls.__config__.orm_mode:
             return cls.from_orm(value)
+        elif isinstance(value, dict):
+            return cls(**value)
         else:
             try:
                 value_as_dict = dict(value)
--- a/tests/test_orm_mode.py
+++ b/tests/test_orm_mode.py
@@ -302,3 +302,38 @@ def test_custom_getter_dict_derived_model_class():
 
     model = ExampleOrm.from_orm(Example())
     assert model.dict() == {'name': 'name', 'col': [0, 1, 2, 3, 4], 'id': 1}
+
+
+def test_recursive_parsing():
+    from types import SimpleNamespace
+
+    class Getter(GetterDict):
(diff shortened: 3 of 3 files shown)


PrettyWood (changes requested):
@nuno-andre Please update with tests and change file thanks

nuno-andre:
@PrettyWood Should I create a new test case or would you prefer I adapt `test_custom_getter_dict_derived_model_class` to check this condition?

https://github.com/samuelcolvin/pydantic/blob/c489445cea40c4962ecfa9e342eb30a0ba819129/tests/test_orm_mode.py#L271-L304

PrettyWood (approved):

PrettyWood:
Thank you!

lsapan:
This change actually prevents models (that are ORM-enabled) from being initialized the normal way from dictionaries. Prior to this change, the following code worked:

```
from types import SimpleNamespace

class User(BaseModel):
    first_name: str
    last_name: str

    class Config:
        orm_mode = True


class State(BaseModel):
    user: User

    class Config:
        orm_mode = True


# Pass an "orm instance"
state = State.from_orm(SimpleNamespace(user=SimpleNamespace(first_name='John', last_name='Appleseed')))

# Pass dictionary data directly
state = State(**{'user': {'first_name': 'John', 'last_name': 'Appleseed'}})
```

After this change, the final line throws a validation error:

```
  File "/Users/luke/Developer/Clients/doctorschoice/flow/orm_issue.py", line 26, in <module>
    state = State(**{'user': {'first_name': 'John', 'last_name': 'Appleseed'}})
  File "/Users/luke/.pyenv/versions/flow/lib/python3.9/site-packages/pydantic/main.py", line 330, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 2 validation errors for State
user -> first_name
user -> last_name
```

This is actually a bit of an issue given the way FastAPI handles response models. During the processing of a request, it serializes the data, and then passes it through a response model (if one was provided) to make sure the response matches the desired output schema. If you're using an ORM-mode model for convenience (like one generated from Tortoise-ORM's `pydantic_model_creator`), it now crashes with a validation error.

PR #2720, opened 2021-05-01
fix: add missing BaseConfig type annotations and handle `max_length = 0`

### LINKED ISSUE #2719: Incomplete type annotations in BaseConfig yield mypy errors when Config is defined from BaseConfig
### Checks

# Bug

1. The `BaseConfig` is missing type annotations for a few fields. Inheriting from `BaseConfig` and setting values yields type errors from `mypy`. 

```python
from pydantic import BaseConfig, BaseModel, Extra


class Record(BaseModel):
    alpha: str
    beta: str

    class Config(BaseConfig):
        title = "Record"
        extra = Extra.ignore
        max_anystr_length = 1234

```

yields

```bash
mypy --version && mypy example.py
mypy 0.812
example.py:9: error: Incompatible types in assignment (expression has type "str", base class "BaseConfig" defined the type as "None")  [assignment]
example.py:11: error: Incompatible types in assignment (expression has type "int", base class "BaseConfig" defined the type as "None")  [assignment]
Found 2 errors in 1 file (checked 1 source file)
```

I believe the type annotations should be:

```python
from typing import Union, Optional

class BaseConfig:
    title: Optional[str] = None
    min_anystr_length: Optional[int] = None # can this just be int = 0 ?
    max_anystr_length: Optional[int] = None
    extra = Union[str, Extra] = Extra.ignore
```

Inheriting from `BaseConfig` is useful for autocomplete and avoids typos like the following.

```python

class Point(BaseModel):
    x: int
    y: int
    
    class Config:
        extras = "forbid"
```

2. The [Config docs](https://pydantic-docs.helpmanual.io/usage/model_config/) look to be inconsistent with the default values defined in `BaseConfig`.

- the default for `max_anystr_length` is listed as `2 ** 16` in the docs, however, `None` is defined in `BaseConfig` (i.e., there is no limit)
- Setting `max_anystr_length = 0` ignores the validation completely and has a different semantic meaning than any other positive integer. In other words, `max_anystr_length = 0` is the same as `max_anystr_length = None`.
- Similarly, the default for `min_anystr_length` is listed as `0` in the docs, but the listed default is `None` in `BaseConfig`. If these are semantically the same, then perhaps set `min_anystr_length:int = 0` as the default?
- In general, it might be useful to clarify the truthyness of these `0` vs `None` values 

```
             pydantic version: 1.8.1
            pydantic compiled: False
               python version: 3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 05:02:20)  [Clang 11.0.1 ]
                     platform: macOS-11.2.3-x86_64-i386-64bit
```

PrettyWood:
## Related issue number
closes #2719

## Checklist

### DIFF
--- a/changes/2719-PrettyWood.md
+++ b/changes/2719-PrettyWood.md
@@ -1,0 +1,1 @@
+add missing type annotations in `BaseConfig` and handle `max_length = 0`
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -32,7 +32,7 @@ _(This script is complete, it should run "as is")_
 : the min length for str & byte types (default: `0`)
 
 **`max_anystr_length`**
-: the max length for str & byte types (default: `2 ** 16`)
+: the max length for str & byte types (default: `None`)
 
 **`validate_all`**
 : whether to validate field defaults (default: `False`)
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -112,21 +112,21 @@ class Extra(str, Enum):
 
 
 class BaseConfig:
-    title = None
-    anystr_lower = False
-    anystr_strip_whitespace = False
-    min_anystr_length = None
-    max_anystr_length = None
-    validate_all = False
-    extra = Extra.ignore
-    allow_mutation = True
-    frozen = False
-    allow_population_by_field_name = False
-    use_enum_values = False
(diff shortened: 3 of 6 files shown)


mpkocher:
Looks good!

PR #2722, opened 2021-05-01
Eliminate redundant variable with direct returns in the if statement

ShahriyarR:
As we have else clause to handle the last case, it is easy to return inside the `if/elif` pairs rather than to assign the value to the `brand` variable.


samuelcolvin:
I don't think this is significant to require a change.

PR #2725, opened 2021-05-01
Remove redundant else clause as the if statement has already return

ShahriyarR:
No need for these else clauses.


samuelcolvin:
I don't agree with this, I think the `else` clause makes the code easier to read.

PR #2739, opened 2021-05-03
add label-and-assign github action

samuelcolvin:
This should making reviewing easier by allowing PR authors to assign reviews (currently me and @PrettyWood) and simplifying the process of us re-assigning the authors after reviews.

@PrettyWood I hope you don't mind me adding you here? It shouldn't make any difference to the number of notifications you get.

## Usage

PR authors can assign reviewers by adding a comment to their PR with the magic string **"please review"**, this will:
* unassign them (if they're assigned) and assign the reviewers
* add the `ready for review` label and remove the `awaiting author revision` label if required

Reviewers can add the magic string **"please update"** which will:
* unassign reviewers and assign the PR author
* add the `awaiting author revision` label and add the `ready for review` label if required

### DIFF
--- a/.github/workflows/label-assign.yml
+++ b/.github/workflows/label-assign.yml
@@ -3,8 +3,8 @@ name: Label & Assign
 on:
   issue_comment:
     types: [created, edited]
-  pull_request_review:
-    types: [submitted, edited]
+#  pull_request_review:
+#    types: [submitted, edited]
 
 permissions:
   pull-requests: write
--- a/.github/workflows/label-assign.yml
+++ b/.github/workflows/label-assign.yml
@@ -1,0 +1,23 @@
+name: Label & Assign
+
+on:
+  issue_comment:
+    types: [created, edited]
+  pull_request_review:
+    types: [submitted, edited]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  label-and-assign:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


samuelcolvin:
Forgot to add, this is will also work with reviews, additionally there's special functionality where the "please update" magic string is not required when reviewers submit reviews with "Request Changes" checked.

Thanks so much to @tiangolo for the prior art, I copied https://github.com/tiangolo/latest-changes a lot with this.

.github/workflows/label-assign.yml
```diff
@@ -0,0 +1,23 @@
+name: Label & Assign
+
+on:
+  issue_comment:
+    types: [created, edited]
+#  pull_request_review:
```
samuelcolvin, line 6: sadly this doesn't work at the moment since `pull_request_review` runs in the context of the PR so doesn't have permissions to update labels or assignees.

I guess we need to wait for `pull_request_review_target` or something, see [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review)

PrettyWood: > @PrettyWood I hope you don't mind me adding you here
Not at all, it won't change anything for me

> `pull_request_review` runs in the context of the PR
Oh that's a bummer...let's hope Github adds this soon

samuelcolvin:
for anyone coming to this in future and wondering how it works, see https://github.com/samuelcolvin/label-and-assign

PR #2745, opened 2021-05-04
Add ability to have custom encoder for builtin types

### LINKED ISSUE #2531: Config json_encoders does not seem to work on builtin types
```
             pydantic version: 1.8.1
            pydantic compiled: True
                 install path: ../MYPROJECTS/MISC_TESTING/.venv/lib/python3.7/site-packages/pydantic
               python version: 3.7.10 (default, Feb 20 2021, 21:21:24)  [GCC 5.4.0 20160609]
                     platform: Linux-4.4.0-18362-Microsoft-x86_64-with-Ubuntu-16.04-xenial
```

I am getting data from a database and passing it to Pydantic models which ensure correct output for API calls.

This database stores True/False values as 1/0, so I wanted to input to a model then ensure they were correct JSON for API payloads on output. 

This does not seem to work for regular data types like `int` and `str`

```py
>>> from pydantic import BaseModel
>>> class BitToBool(BaseModel):
  2     val: int
  3     class Config:
  4         json_encoders = {int: lambda i: bool(i)}

>>> bit_bool = BitToBool(val=1)

>>> bit_to_bool.__json_encoder__
Out[179]: functools.partial(<function custom_pydantic_encoder at 0x7f749c0e79e0>, 
{<class 'int'>: <function BitToBool.Config.<lambda> at 0x7f747c15d320>})

>>> bit_to_bool.dict()
Out[180]: {'val': 1}

>>> bit_to_bool.json()
Out[181]: '{"val": 1}'
...
```
Same thing happens for strings where it is the same story > database field goes into HTTP post with something prepended to satisfy requirements.

```py
>>> class PrependCustomField(BaseModel):
  2     value: str
  3     class Config:
  4         json_encoders = {str: lambda s: f"CF{s}"}

>>> cf = PrependCustomField(value="EventSummary")

>>> cf.json()
Out[189]: '{"value": "EventSummary"}'
```

This does work with a custom class, this is just extra complexity and another class that isn't really intended to use for any other purpose:

```py
>>> class BitBool:
  2     __slots__ = ["_bit"]
  3
  4     def __init__(self, bit: int = 0):
  5         self._bit = bit
  6
  7     def __bool__(self):
  8         if self._bit == 0:
  9             return False
 10         if self._bit == 1:
 11             return True

>>> class BitToBool(BaseModel):
  2     val: BitBool
  3     class Config:
  4         arbitrary_types_allowed = True
  5         json_encoders = {BitBool: lambda i: bool(i)}

>>> bit_to_bool = BitToBool(val=BitBool(1))
>>> bit_to_bool.json()
Out[208]: '{"val": true}'
```

_Note:_ I can get the right behavior with custom validators, but this isn't the best since really I might still want the original data via `dict()`. I really only need it to change during serialization. Is there something obvious I am missing here?

Thanks!

uriyyo:
## Related issue number
Closes #2531 

## Checklist

### DIFF
--- a/pydantic/json.py
+++ b/pydantic/json.py
@@ -2,12 +2,13 @@ import datetime
 import re
 import sys
 from collections import deque
+from dataclasses import dataclass
 from decimal import Decimal
 from enum import Enum
 from ipaddress import IPv4Address, IPv4Interface, IPv4Network, IPv6Address, IPv6Interface, IPv6Network
 from pathlib import Path
 from types import GeneratorType
-from typing import Any, Callable, Dict, Type, Union
+from typing import AbstractSet, Any, Callable, Dict, Type, Union
 from uuid import UUID
 
 if sys.version_info >= (3, 7):
@@ -19,7 +20,32 @@ else:
 from .color import Color
 from .types import SecretBytes, SecretStr
 
-__all__ = 'pydantic_encoder', 'custom_pydantic_encoder', 'timedelta_isoformat'
+__all__ = (
+    'BuiltinTypeWrapper',
+    'pydantic_encoder',
+    'custom_pydantic_encoder',
+    'timedelta_isoformat',
+    'wrap_builtin_types',
+)
+
+
+@dataclass
+class BuiltinTypeWrapper:
+    obj: Any
+
+
+def wrap_builtin_types(obj: Any, types: AbstractSet[Type[Any]]) -> Any:
+    if obj.__class__ is dict:
+        obj = {k: wrap_builtin_types(v, types) for k, v in obj.items()}
+    elif obj.__class__ is list:
+        obj = [wrap_builtin_types(v, types) for v in obj]
+    elif obj.__class__ is tuple:
+        obj = tuple(wrap_builtin_types(v, types) for v in obj)
+
+    if obj.__class__ in types:
+        obj = BuiltinTypeWrapper(obj)
+
+    return obj
 
 
 def isoformat(o: Union[datetime.date, datetime.time]) -> str:
@@ -96,6 +122,9 @@ def pydantic_encoder(obj: Any) -> Any:
 
 
 def custom_pydantic_encoder(type_encoders: Dict[Any, Callable[[Type[Any]], Any]], obj: Any) -> Any:
+    if obj.__class__ is BuiltinTypeWrapper:
+        obj = obj.obj  # type: ignore
+
     # Check the class type and its superclasses for a matching encoder
     for base in obj.__class__.__mro__[:-1]:
         try:
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -346,10 +346,10 @@ class ModelMetaclass(ABCMeta):
         vg.check_for_unused()
         if config.json_encoders:
             json_encoder = partial(custom_pydantic_encoder, config.json_encoders)
-            json_built_in_types = frozenset(t for t in config.json_encoders if t in BUILTIN_TYPES)
+            json_builtin_types = frozenset(t for t in config.json_encoders if t in BUILTIN_TYPES)
         else:
             json_encoder = pydantic_encoder
-            json_built_in_types = frozenset()
+            json_builtin_types = frozenset()
 
         pre_rv_new, post_rv_new = extract_root_validators(namespace)
 
@@ -373,7 +373,7 @@ class ModelMetaclass(ABCMeta):
             '__post_root_validators__': unique_list(post_root_validators + post_rv_new),
             '__schema_cache__': {},
             '__json_encoder__': staticmethod(json_encoder),
-            '__json_builtin_encode_types__': json_built_in_types,
+            '__json_builtin_types_to_encode__': json_builtin_types,
             '__custom_root_type__': _custom_root_type,
             '__private_attributes__': private_attributes,
             '__slots__': slots | private_attributes.keys(),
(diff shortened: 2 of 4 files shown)


uriyyo:
please review

PrettyWood:
I guess it also tackles #3104?

WinnerOK:
Hello, @samuelcolvin, @PrettyWood. Could you please tell what is the state of the review? Do you have any estimates?

I would really like using this feature in my code instead of defining custom `json_dumps` on my models.

pydantic/main.py
```diff
@@ -222,6 +222,18 @@ def hash_function(self_: Any) -> int:
 # the `BaseModel` class, since that's defined immediately after the metaclass.
 _is_base_model_class_defined = False
 
+BUILTIN_TYPES: AbstractSet[Type[Any]] = frozenset(
```
PrettyWood, line 225: What about bytes, set and frozenset ... ?

uriyyo: It's already possible to encode `frozenset`, 'set' and 'bytes' so no need to add them to this set. This set contains types that can't have custom json encoder.

Maybe it's better to rename this variable 🤔

samuelcolvin: I think we have to rename it, not sure what to right now. I'll think about it.

pydantic/main.py
```diff
@@ -277,6 +292,7 @@ def is_untouched(v: Any) -> bool:
             ),
             '__schema_cache__': {},
             '__json_encoder__': staticmethod(json_encoder),
+            '__json_builtin_types_to_encode__': json_builtin_types,
```
samuelcolvin, line 295: I'm not clear why this needs to be stored on the model. Can't `BUILTIN_TYPES` just be a constant that's used in `wrap_builtin_types`?

pydantic/json.py
```diff
@@ -20,7 +20,32 @@
...
+    'pydantic_encoder',
+    'custom_pydantic_encoder',
+    'timedelta_isoformat',
+    'wrap_builtin_types',
+)
+
+
+class BuiltinTypeWrapper:
+    def __init__(self, obj: Any) -> None:
+        self.obj = obj
```
samuelcolvin, line 34: I think we can have a better name than `obj`.

Like `wrapped`, `inner_obj` or similar.

Also, `__slots__`

pydantic/main.py
```diff
@@ -125,6 +125,18 @@ def hash_function(self_: Any) -> int:
...
 
+BUILTIN_TYPES: AbstractSet[Type[Any]] = frozenset(
+    (
+        dict,
+        list,
+        tuple,
+        float,
+        bool,
+        int,
+        str,
```
samuelcolvin, line 136: What about `None`?

pydantic/json.py
```diff
@@ -20,7 +20,32 @@
...
+
+def wrap_builtin_types(obj: Any, types: AbstractSet[Type[Any]]) -> Any:
+    if obj.__class__ is dict:
+        obj = {k: wrap_builtin_types(v, types) for k, v in obj.items()}
+    elif obj.__class__ is list:
+        obj = [wrap_builtin_types(v, types) for v in obj]
+    elif obj.__class__ is tuple:
+        obj = tuple(wrap_builtin_types(v, types) for v in obj)
+
+    if obj.__class__ in types:
```
samuelcolvin, line 45: What happens to subclass of these types? Is there a reason not to use `isinstance`?

samuelcolvin (commented):
please update

samuelcolvin:
please update

pydantic/json.py
uriyyo: Fixed

pydantic/main.py
uriyyo: Good catch, fixed.

pydantic/json.py
uriyyo: Nope, it was a mistake from my side. Fixed.

pydantic/main.py
uriyyo: It's more performance tweak. In case when there are no types that is subclass of a non-encodable then no wrapping will be performed.

RvDijk:
This behaviour is quite nice, looking forward to use it! Is anything blocking this PR?

uriyyo:
please review

pydantic/main.py
```diff
@@ -126,6 +127,17 @@ def hash_function(self_: Any) -> int:
 # the `BaseModel` class, since that's defined immediately after the metaclass.
 _is_base_model_class_defined = False
 
+JSON_UNENCODABLE_TYPES: Tuple[Type[Any], ...] = (
```
samuelcolvin, line 130: This name is extremely confusing, surely it should be called `JSON_TYPES`, also:
* it needs a comment explaining how it's used
* could we move this and its love into `json.py`.

samuelcolvin (commented):
Other than that change, I think this looks good in itself.

If we go ahead with this, we would need to add docs, at least a minimal comment about the change.

But, my question is more about the long term plan - [the plan](https://pydantic-docs.helpmanual.io/blog/pydantic-v2/#improvements-to-dumpingserializationexport) in V2 is to redesign how serialisation works.

In particular, are we going to keep `json_encoders` in V2? If so, I guess it makes sense to go ahead with this PR, but if not - is it slightly weird to make this change for v1.10, then changes again in v2?

@PrettyWood would love your opinion on this.

PrettyWood:
Hmmm I'm not sure. Clearly it's something that I want and that should be added in pydantic but I would have waited for v2.

samuelcolvin:
I think before we merge this we need to decide how we're going to do serialisation in V2 and see if this would be compatible with this.

samuelcolvin:
Sorry, having thought about this, I simply don't have the bandwidth right now to know what the serialisation interface will look like for V2, therefore I'm going to defer this to V2 - depending how much of the logic ends up in python, we might use this or copy it's semantics.

@uriyyo thanks so much for the contribution, the ideas discussed and implemented here really help make V2 as good as possible.

samuelcolvin:
see #4456

pydantic/main.py
CeltiumCDev: k

PR #2748, opened 2021-05-05
mypy: Add create_model overload

uriyyo:
## Add create_model overload

Currently `mypy` will fail on this code:
```python

Foo = create_model("Foo")
```
```
error: Need type annotation for 'Foo'  [var-annotated]
```

With this PR `mypy` will correctly resolve return value type of `create_model`:
```python

Foo = create_model("Foo")
reveal_type(Foo)
```
```
note: Revealed type is 'Type[pydantic.main.BaseModel]'
```

And it will still work for a cases when `__base__` is specified:
```python
from pydantic import create_model, BaseModel


class Bar(BaseModel):
    a: int


Foo = create_model("Foo", __base__=Bar)
reveal_type(Foo)
```
```
note: Revealed type is 'Type[temp.Bar*]'
```

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -21,6 +21,7 @@ from typing import (
     Union,
     cast,
     no_type_check,
+    overload,
 )
 
 from .class_validators import ValidatorGroup, extract_root_validators, extract_validators, inherit_validators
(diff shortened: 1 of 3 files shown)


uriyyo:
please review

pydantic/main.py
```diff
@@ -938,6 +938,32 @@ def __repr_args__(self) -> 'ReprArgs':
...
+    __config__: Type[BaseConfig] = None,
+    __base__: None = None,
+    __module__: str = __name__,
+    __validators__: Dict[str, classmethod] = None,
+    **field_definitions: Any,
+) -> Type['BaseModel']:
+    pass
+
+
+@overload
```
sbv-csis, line 954: Im not an expert by any stretch of the imagination, but isn't this overload the same as the final implementation - is that needed?

uriyyo: It's required to have all overloads defined.

Please see more info here [PEP 484](https://www.python.org/dev/peps/pep-0484/#id45)

PrettyWood (approved):
Great! Thank you so much

Issue #2766, opened 2021-05-09
Can not define `Final` class var in BaseModel


zen-xu:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.8.1
            pydantic compiled: False
                 install path: /Users/zen-xu/.pyenv/versions/3.9.1/lib/python3.9/site-packages/pydantic
               python version: 3.9.1 (default, Mar  3 2021, 21:05:29)  [Clang 12.0.0 (clang-1200.0.32.29)]
                     platform: macOS-11.2.3-arm64-arm-64bit
     optional deps. installed: ['email-validator', 'typing-extensions']
```

```py
from typing import Final, ClassVar
from pydantic import BaseModel

class Template(BaseModel):
    kind: Final[ClassVar[str]] = "xxx"
    content: str

```

Will raise `TypeError: typing.ClassVar[str] is not valid as type argument`

uriyyo:
As [`PEP 591`](https://www.python.org/dev/peps/pep-0591/#id14) say:
> Variables should not be annotated with both ClassVar and Final.

I believe you should use only `Final` in this case.

zen-xu:
Will pydantic support `Final` one day?

PR #2768, opened 2021-05-09
Add ability to use Final in a field type annotation

uriyyo:
## Add ability to use `Final` in a field type annotation
Implementation is based on a `PEP 591`:
> Type checkers should infer a final attribute that is initialized in a class body as being a class variable.

> A final attribute declared in a class body without an initializer must be initialized in the __init__ method

So in this case field `name` will be treated as a `ClassVar`:
```py
from typing import Final

class Model(BaseModel):
    name: Final[str] = 'Yurii'

assert 'name' in Model.__class_vars__
assert 'name' not in Model.__fields__
```

But in a case when a default value is not present than field will be treated as an init-only required field:
```python
from typing import Final

class Model(BaseModel):
    name: Final[str]

assert 'name' not in Model.__class_vars__
assert 'name' in Model.__fields__

_ = Model()  # will fail because name is missed

obj = Model(name='Yurri')
obj.name = 'New Value'  # will fail because final field has been already inited
```

## Related issue number
#2766 

## Checklist


uriyyo:
@samuelcolvin @PrettyWood Could you please review this PR? I want to hear your opinion regarding this such feature.

nuno-andre:
IMHO, this implementation adds an unexpected behavior from the _pydanticish_ perspective (a default value turns what would otherwise be an instance field into a class var) without a clear benefit.

```python

class Model(BaseModel):
    var: Final[str] = Field(max_length=3)
```

I presume that in a case like this no one would expect to have an immutable class var of the type `FieldInfo`.

The aim of a final name is not to be reassigned after initialization, but there's no such initialization in a Model declaration. As with `dataclass`, it's needed to annotate the attribute as a `ClassVar` to be treated as such and excluded from `fields/__fields__`.

```python

class A:
    var = 1

class B(BaseModel):
    var = 1
```

```python
>>> A.var
1
>>> B.var
...
AttributeError: type object 'B' has no attribute 'var'
```

I think `Final` could be more useful as an `allow_mutation = False` synonym that would also prevent the attribute redefinition in subclasses. The problem here would be that `Final` and `ClassVar` should not be together...

The behavior of `Final` for cases such as `dataclass` or `TypedDict` is not defined in the PEP. And, in fact, mypy doesn't seem to know how to deal with it.

```python
from typing import Final, TypedDict

@dataclass
class A:
    var: Final[str]
#> error: Final name must be initialized with a value

class B(TypedDict):
    var: Final[str]
#> error: Final can be only used as an outermost qualifier in a variable annotation

# this also raises a TypeError
#> TypeError: typing.Final[str] is not valid as type argument
```

But mypy's [`Var` AST node](https://github.com/python/mypy/blob/master/mypy/nodes.py#L804) has a `final_set_in_init` attribute that seems a fine approach for Pydantic to mimic in a case like this.

PrettyWood:
Honestly I'm against it. I feel like it's really confusing and the different behaviour with or without a default value goes against pydantic one (required vs optional more or less).
I feel like people should stick with plain `ClassVar` or directly the type, which is explicit

uriyyo:
@nuno-andre @PrettyWood Thanks for your comments.

@nuno-andre Thanks for pointing out case when the field declared as a `Field` call. I have updated the implementation and now it will be resolved correctly.

Regarding the mypy. Currently, pydantic mypy plugin handles a lot of cases and for instance, it generates `__init__` method with the correct type decl for a model. From my perspective, we can update plugin to handle special cases for `Final`.

@PrettyWood Currently `pydantic` has a lot of ways to decl fields:
```python
class Model(BaseModel):
    a: int  # required
    b: int = 10  # optional
    c: int = Field()  # required
    d: int = Field(default=10)  # optional
    e: int = Field(default_factory=lambda: 10)  # optional
```

To simplify things we can always use `Final` type annotation as for instance fields.
```python
class Model(BaseModel):
    a: Final[int] = 10 # instance field with default value 10 same as Field(default=10)
```

Another feature of this PR is that we can have init-only fields:
```python
class Model(BaseModel):
    a: Final[int]


obj = Model(a=10)
obj.a = 100  # will fail because final field can't be changed outside of init method
```

As far as implementation follows `PEP 591` it's pretty easy to predict behavior of this feature.

PrettyWood:
@uriyyo My point was just about class var / field. By setting a default value, we change a field in a class var, which is what the PEP says but feels anti-pydantic to me.
The use of `a: Final[int]` to make a value immutable (since it's only settable once in the init) feels good.
Discard: It's just my opinion ;)

uriyyo:
@PrettyWood Thanks for your opinion, it's really important for me🙂

I totally agree with you that we can use `Final` to make a value immutable and it will be more explicit rather than have a behavior when `Final` makes field as a class var.

PrettyWood:
I think we should get more feedback on this before adding anything or reviewing.
@samuelcolvin what's your opinion on the rationale? others?

uriyyo:
@PrettyWood @samuelcolvin Any chance to review this one? It would be great to hear your opinion. I would like to use this feature in my projects)

pydantic/main.py
```diff
@@ -183,12 +184,17 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
         def is_untouched(v: Any) -> bool:
             return isinstance(v, untouched_types) or v.__class__.__name__ == 'cython_function_or_method'
 
+        def is_finalvar_with_default_val(type_: Type[Any], val: Any) -> bool:
```
samuelcolvin, line 187: This can be moved into the module scope of `fields.py`.

pydantic/main.py
```diff
@@ -347,6 +353,10 @@ def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
             raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')
         elif not self.__config__.allow_mutation or self.__config__.frozen:
             raise TypeError(f'"{self.__class__.__name__}" is immutable and does not support item assignment')
+        elif name in self.__fields__ and self.__fields__[name].final:
+            raise TypeError(
```
samuelcolvin, line 357: https://github.com/samuelcolvin/pydantic-core/issues/219

pydantic/typing.py
```diff
@@ -386,10 +387,21 @@ def _check_classvar(v: Optional[Type[Any]]) -> bool:
     return v.__class__ == ClassVar.__class__ and getattr(v, '_name', None) == 'ClassVar'
 
 
+def _check_finalvar(v: Optional[Type[Any]]) -> bool:
```
samuelcolvin, line 390: can you add a docstring explaining what this is doing.

tests/test_main.py
```diff
@@ -2156,3 +2156,56 @@ class Model(BaseModel):
...
+
+        if value is not None:
+            a = value
+
+    Model.update_forward_refs(ann=ann)
+
+    assert 'a' not in Model.__class_vars__
+    assert 'a' in Model.__fields__
+
+    assert Model.__fields__['a'].final
```
samuelcolvin, line 2183: we we add a check somewhere else that `final` is normally false.

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin:
please update.

samuelcolvin:
you'll also need to rebase/merge master to get tests passing.

uriyyo:
ready for review

uriyyo:
please review

samuelcolvin:
thanks so much.

bzoracler:
A PR has just been opened to change the Python typing specifications that will propagate to any use of dataclass-likes in the Python ecosystem, such as with `@dataclass_transform`.

https://github.com/python/typing/pull/1669

The above typing PR specifies behaviour which is incompatible with the `Final = <value>` (in Pydantic , treated as a `ClassVar` in `BaseModel` subclasses based on a consistent reading of PEP 591) made in this Pydantic PR. The relevant behaviour still exists in Pydantic here:

https://github.com/pydantic/pydantic/blob/8128821b864db533b7c5c0c75589777c10d5cda7/pydantic/_internal/_fields.py#L169-L171

The typing PR will mean that `Final = <value>` become instance fields with a default value of `<value>` instead, in dataclass-like class bodies.

---

As `@dataclass_transform` was created with Pydantic in mind, I think it'd be valuable if anyone would like to weigh in with explicit support or rejection of this change to the typing spec.

Viicos:
@bzoracler see https://github.com/pydantic/pydantic/issues/10474#issuecomment-2478666651. We should be able to change this in V3, as this will be considered a breaking change.

Issue #2798, opened 2021-05-15
Self-references nested inside data type do not appear to update on update_forward_refs()


jleclanche:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Ref: https://pydantic-docs.helpmanual.io/usage/postponed_annotations/

Sample code:

```py
from pydantic import BaseModel

class Order(BaseModel):
    legs: Optional[list["Order"]]

Order.update_forward_refs()
```

Traceback:
```pytb
    class Order(BaseModel):
pydantic/main.py:293: in pydantic.main.ModelMetaclass.__new__
pydantic/fields.py:410: in pydantic.fields.ModelField.infer
pydantic/fields.py:342: in pydantic.fields.ModelField.__init__
pydantic/fields.py:450: in pydantic.fields.ModelField.prepare
pydantic/fields.py:545: in pydantic.fields.ModelField._type_analysis
pydantic/fields.py:629: in pydantic.fields.ModelField._type_analysis
pydantic/fields.py:648: in pydantic.fields.ModelField._create_sub_type
pydantic/fields.py:342: in pydantic.fields.ModelField.__init__
pydantic/fields.py:456: in pydantic.fields.ModelField.prepare
pydantic/fields.py:670: in pydantic.fields.ModelField.populate_validators
pydantic/validators.py:710: in find_validators
E   RuntimeError: error checking inheritance of 'Order' (type: str)
```

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.8.1
            pydantic compiled: True
                 install path: /home/adys/.cache/pypoetry/virtualenvs/alpaca-api-client-JFdRWl93-py3.9/lib/python3.9/site-packages/pydantic
               python version: 3.9.5 (default, May 12 2021, 17:14:51)  [GCC 10.2.0]
                     platform: Linux-5.12.3-arch1-1-x86_64-with-glibc2.33
     optional deps. installed: ['typing-extensions']
```

uriyyo:
The correct way to define self-referencing model will be:
```python
from typing import Optional
from pydantic import BaseModel


class Order(BaseModel):
    legs: "Optional[list[Order]]"


Order.update_forward_refs()
```
Or
```python
from __future__ import annotations
from typing import Optional
[...]

PR #2802, opened 2021-05-16
Fix issue with generic type aliases forward refs resolving

uriyyo:
## Change Summary

Fix issue with generic type aliases forward refs resolving.
Fix issue when `update_forward_refs` fail to resolve `ForwardRef` of field Mapping key.

## Related issue number

#2798
#2803 

## Checklist

### DIFF
--- a/changes/2803-uriyyo.md
+++ b/changes/2803-uriyyo.md
@@ -1,0 +1,2 @@
+Fix issue when `update_forward_refs` fail to resolve `ForwardRef` 
+of field Mapping key.
--- a/changes/2803-uriyyo.md
+++ b/changes/2803-uriyyo.md
@@ -1,2 +1,0 @@
-Fix issue when `update_forward_refs` fail to resolve `ForwardRef` 
-of field Mapping key.
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -20,6 +20,7 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    cast,
 )
 
 from typing_extensions import Annotated
@@ -42,6 +43,7 @@ from .typing import (
     is_new_type,
     is_typeddict,
     new_type_supertype,
+    normalize_type,
 )
 from .utils import PyObjectStr, Representation, ValueItems, lenient_issubclass, sequence_like, smart_deepcopy
 from .validators import constant_validator, dict_validator, find_validators, validate_json
@@ -347,7 +349,7 @@ class ModelField(Representation):
         self.name: str = name
         self.has_alias: bool = bool(alias)
         self.alias: str = alias or name
-        self.type_: Any = type_
+        self.type_: Any = normalize_type(type_)
         self.outer_type_: Any = type_
         self.class_validators = class_validators or {}
         self.default: Any = default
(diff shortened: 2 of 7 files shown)


uriyyo:
please review

FeeeeK:
Will this pull request be accepted?
It also fixes a bug with type aliases like this:
```python
ListOfModels = typing.List["Model"]

class Model(BaseModel):
    pass

class Example(BaseModel):
    a = ListOfModels
```

uriyyo:
Hi @FeeeeK,

Could you please provide more context?

From what I see, there are no problems with code that you provide.
Code below should be executed without errors:
```py
import typing

ListOfModels = typing.List["Model"]


class Model(BaseModel):
    pass


class Example(BaseModel):
    a: ListOfModels


print(Example(a=[Model()]))
```

I will try to add more context on what case this PR will fix.

Basically when you pass str as a typing generic parameter it will be automatically wrapped with `ForwardRef`:
```py
>>> List["Foo"]
typing.List[ForwardRef('Foo')]
```

But in case of a builtin generic aliases it will not be wrapped:
```py
>>> list["Foo"]
list['Foo']
```

FeeeeK:
In my case, I had to call `Example.update_forward_refs()` twice.

Otherwise there would be an error:
`field "response" not yet prepared so type is still a ForwardRef, you might need to call SendUserIdsResponse.update_forward_refs().`
[Model](https://github.com/timoniq/vkbottle-types/blob/d635892e09e2efbd9eeb0c7c7c05cba4f26b5064/vkbottle_types/responses/messages.py#L172)
[Alias](https://github.com/timoniq/vkbottle-types/blob/d635892e09e2efbd9eeb0c7c7c05cba4f26b5064/vkbottle_types/responses/messages.py#L370)

tests/test_generics.py
```diff
@@ -23,6 +23,9 @@
 from pydantic.generics import GenericModel, _generic_types_cache, iter_contained_typevars, replace_types
 
 skip_36 = pytest.mark.skipif(sys.version_info < (3, 7), reason='generics only supported for python 3.7 and above')
+only_39 = pytest.mark.skipif(
```
PrettyWood, line 26: `above_39` maybe since in fact it's also 3.10 now ;)

pydantic/typing.py
```diff
@@ -189,6 +189,22 @@ def get_args(tp: Type[Any]) -> Tuple[Any, ...]:
...
 
+if sys.version_info < (3, 9):
+
+    def normalize_type(tp: Type[Any]) -> Type[Any]:
+        return tp
+
+
+else:
+
+    def normalize_type(tp: Type[Any]) -> Type[Any]:
```
PrettyWood, line 200: Can you add a docstring with a input -> output example pls?

pydantic/fields.py
```diff
@@ -347,7 +348,7 @@ def __init__(
         self.name: str = name
         self.has_alias: bool = bool(alias)
         self.alias: str = alias or name
-        self.type_: Any = type_
+        self.type_: Any = normalize_type(type_)
         self.outer_type_: Any = type_
```
PrettyWood, line 352: What about outer_type?

changes/2803-uriyyo.md
```diff
@@ -0,0 +1,2 @@
+Fix issue when `update_forward_refs` fail to resolve `ForwardRef`
```
PrettyWood, line 1: I think it's easier to have one change file with the PR number with both points inside (key for mappings + normalize)

PrettyWood (changes requested):
Please update

uriyyo:
#3681 has better implementation and will fix the same issues as this PR

PR #2863, opened 2021-05-31
fix: attrs + cattrs benchmark

PrettyWood:
With https://github.com/Tinche/cattrs/commit/41c494599efe3cf49e11f2630dff5415b719fe43, `cattrs` uses `GenConverter` instead of `Converter` as global converter, which raises `KeyError` by default when a required value is missing
Let's just catch this `KeyError` in the benchmark!


tiangolo (approved):
LGTM :rocket: 

I see this fixes the current CI errors for the benchmark tests :sweat_smile: 

Thanks @PrettyWood ! :cake:

PR #2890, opened 2021-06-08
Avoid __dict__ and __weakref__ attributes in AnyUrl and _BaseAddress subclasses

nuno-andre:
## Change Summary

Add an empty `__slots__` declaration to `AnyUrl` and `_BaseAddress` subclasses to avoid the creation of `__dict__` and `__weakreef__` attributes. When inheriting from a class with `__slots__`, [child subclasses will get a `__dict__` and `__weakref__` unless they also define `__slots__`](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots).

```python
from pydantic import AnyUrl, AnyHttpUrl

class AnyOtherHttpUrl(AnyUrl):
    allowed_schemes = {'http', 'https'}
    __slots__ = ()

print(set(dir(AnyHttpUrl)) - set(dir(AnyUrl)))
#> {'__weakref__', '__dict__'}

print(set(dir(AnyOtherHttpUrl)) - set(dir(AnyUrl)))
#> set()
```

## Related issue number

None.

## Checklist


PrettyWood:
Hi @nuno-andre 
What's the matter with having `__dict__`  and `__weakref__`?
What problem does it solve?
AFAIR I haven't seen any issue related to this 🤔

nuno-andre:
Hi @PrettyWood!

One of the `__slots__`'s purposes is [to deny the creation of `__dict__` to save memory](https://docs.python.org/3/reference/datamodel.html#slots).

```python
from pydantic import AnyUrl, AnyHttpUrl
from pympler.asizeof import asizeof


class AnyOtherHttpUrl(AnyUrl):
    allowed_schemes = {'http', 'https'}
    __slots__ = ()


a = AnyUrl.__new__(AnyHttpUrl, 'https://example.com')
print(asizeof(a))
#> 224

b = AnyUrl.__new__(AnyOtherHttpUrl, 'https://example.com')
print(asizeof(b))
#> 120
```

```python


class A(BaseModel):
    url: AnyHttpUrl


class B(BaseModel):
    url: AnyOtherHttpUrl


a = A(url='https://example.com')
print(asizeof(a))
#> 808

b = B(url='https://example.com')
print(asizeof(b))
#> 704
```

I think this is a pretty straightforward and valuable optimization and also makes a more consistent type model.

changes/2890-nuno-andre.md
```diff
@@ -0,0 +1 @@
+Avoid __dict__ and __weakref__ attributes in AnyUrl and _BaseAddress subclasses.
```
samuelcolvin, line 1:
```suggestion
Avoid `__dict__` and `__weakref__` attributes in `AnyUrl` and IP address fields.
```

samuelcolvin (commented):
Seems good to me, I can't see how this will break anyone's code. But I'm sure we'll have someone shouting at us about breaking code in a few weeks from some weird usage.

tl24:
Let the shouting begin... 😃 We had a breaking change as we were reading the slots to be able to clone a RedisDsn and add in a password that was stored in a secret.  We worked around by hard-coding the property names, but is there a better way of doing that?

For reference:
```
def copy_dsn(original, **override_args):
    # object properties don't show up with normal methods,
    # so we need to use the __slots__ property to get field names
    attrs = {n: getattr(original, n) for n in original.__slots__ if getattr(original, n) is not None}
    return type(original)(None, **{**attrs, **override_args})

dsn_with_password = copy_dsn(RedisDsn("redis://localhost:6379/0", password="secret")
```

tl24:
I don't know if this would be a frequent use case, but something like copy on BaseModel would be nice:
```
dsn_with_password = RedisDsn("redis://localhost:6379/0").copy(update={"password":"secret"})
```

PR #2908, opened 2021-06-14
Toml support for `pydantic-mypy` plugin config

jrwalk:
## Change Summary

This PR adds support for configuring the `pydantic-mypy` plugin in `pyproject.toml`, as well as the previous `mypy.ini` config file, enabling full compatibility with `mypy>0.900`. 

Configuration behavior should be consistent with prior behavior from `mypy.ini`, with enumerated slots populated from the `pydantic-mypy` configuration section, and with non-boolean options resulting in a `ValueError`.

Note that this increases the pinned `mypy` version in the testing suite to 0.901, and includes some formatting changes for compatibility to the test cases. However, the plugin behavior is the same in earlier versions of `mypy`.

## Related issue number

Resolves https://github.com/samuelcolvin/pydantic/issues/2895

## Checklist


tests/requirements-linting.txt
```diff
@@ -3,7 +3,7 @@ flake8==3.9.2
 flake8-quotes==3.2.0
 hypothesis==6.13.10
 isort==5.8.0
-mypy==0.812
+mypy==0.901
```
johnthagen, line 6: Why not [`0.902`](https://pypi.org/project/mypy/0.902/)?

johnthagen (changes requested):

tests/requirements-linting.txt
jrwalk: 0.901 was the newest when I started this, 0.902 only released Thursday past. It shouldn't cause problems to bump this.

jrwalk: bumped version in 97efeee, no issues in the tests.

jrwalk:
This should be ready for review!

jrwalk:
This should be ready, please review!

pydantic/mypy.py
```diff
@@ -1,6 +1,7 @@
 from configparser import ConfigParser
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Type as TypingType, Union
 
+import toml
```
SnoopJeDi, line 4: This PR does not add `toml` to `pydantic`'s dependencies, so this import and its use later in the file should be more cautious as written, i.e.

```suggestion
try:
    import toml
except ImportError:
    import warnings
    warnings.warn("'toml'  module not installed, will not be able to read configuration from pyproject.toml without it")
    toml = None  # and then check `if toml and options.config_file.endswith('.toml'):` during usage
```

However, note that [`tomli` is part of `mypy`'s dependencies](https://github.com/python/mypy/blob/5e73cd7b0dc9b871716acf0086b2ca0376fcff39/setup.py#L196) and should therefore already be present on the user's system. I'd suggest rewriting to use it.

SnoopJeDi:
@samuelcolvin @PrettyWood is anything other than resolving the implicit dependency needed to get this merged? I'd love to stop organizing project configuration around the plugin's idiosyncratic parsing, and the PR looks great otherwise.

pydantic/mypy.py
jrwalk: That seems reasonable to me. For practical purposes the plugin file would only be read in the case of mypy's use, which as you noted is a hard dependency for `mypy` itself and thus would be present, but it doesn't hurt anything to check here.

jrwalk: resolved in 3aaa75a5ecc60691a99ca0e19e9978980ccb5967.

SnoopJeDi: Point of clarification: `mypy` depends on [`tomli`](https://pypi.org/project/tomli/), which is a different library than [`toml`](https://pypi.org/project/toml/). The PR as written will require users to install `toml` to use `pyproject.toml`, but the changes look sufficient to keep it from exploding on a user who doesn't have it.

SnoopJeDi (approved):

pydantic/mypy.py
jrwalk: Good looking out - looks like `mypy` switch to requiring `tomli` rather than `toml` since this PR was submitted (https://github.com/python/mypy/commit/aede3aefd75332f3911a43bb4a6a98baacce785a).  We should definitely change that dependency.

jrwalk: added in 70d85fc

jrwalk: actually, looks like that's not released yet. 0.910 still requires `toml`.  I'll add some checks around that.

jrwalk: OK, it'll default to `toml` now, fall back to `tomli` for upcoming releases, and error out otherwise now, so the bases should be covered.

changes/2908-jrwalk.md
```diff
@@ -0,0 +1,4 @@
+Adds configuration logic to the `PydanticPluginConfig` constructor, enabling parsing of the configuration slots from `pyproject.toml` for `mypy>0.900`.
```
PrettyWood, line 1: No need to give the implementation details, just the enduser explanation
You can probably write something like
```
Make `pydantic-mypy` plugin compatible with `pyproject.toml`. see the [doc](...) for more information```

File: pydantic/mypy.py [hunk 4]
```diff
@@ -1,6 +1,16 @@
 from configparser import ConfigParser
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Type as TypingType, Union
 
+try:
+    import toml
+except ImportError:
+    # future-proofing for upcoming `mypy` releases which will switch dependencies
+    try:
```
PrettyWood, line 8: Great!

tests/mypy/outputs/plugin-fail-strict.txt
```diff
@@ -15,7 +15,7 @@
 87: error: Missing named argument "c" for "DefaultTestingModel"  [call-arg]
 87: error: Missing named argument "d" for "DefaultTestingModel"  [call-arg]
 87: error: Missing named argument "e" for "DefaultTestingModel"  [call-arg]
-91: error: Name 'Undefined' is not defined  [name-defined]
+91: error: Name "Undefined" is not defined  [name-defined]
```
PrettyWood, line 18: Ah sorry I bumped all the deps together yesterday including mypy. Can you please rebase/merge?

PrettyWood (changes requested):
Looks great! Thanks a lot 😊
Just need to rebase/merge and update the changelog to be more digest

PrettyWood:
Please update.

changes/2908-jrwalk.md
jrwalk: changed in https://github.com/samuelcolvin/pydantic/pull/2908/commits/af411103cdc416f0e884ff2b3bee64c74c696506

pydantic/mypy.py
jrwalk: :D

tests/mypy/outputs/plugin-fail-strict.txt
jrwalk: should be squared away now in https://github.com/samuelcolvin/pydantic/pull/2908/commits/c7491208644dfa8f3389882ca236ef9ccca1f7e6

jrwalk:
@PrettyWood should be good to go!

Makefile
```diff
@@ -5,6 +5,7 @@ black = black -S -l 120 --target-version py38 pydantic tests
 .PHONY: install-linting
 install-linting:
 	pip install -r tests/requirements-linting.txt
+	mypy --install-types
```
PrettyWood, line 8: What stubs are missing? If I remember correctly this installs missing stubs caught in the last run. Here we didn't even run mypy. Are you sure it's necessary?

changes/2908-jrwalk.md
```diff
@@ -0,0 +1 @@
+Make `pydantic-mypy` plugin compatible with `pyproject.toml` configuration, consistent with `mypy` changes. see the [doc](https://pydantic-docs.helpmanual.io/mypy_plugin/#configuring-the-plugin) for more information```
```
PrettyWood, line 1: Leading back quotes ;)

pydantic/mypy.py
```diff
@@ -1,6 +1,16 @@
 from configparser import ConfigParser
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Type as TypingType, Union
 
+try:
+    import toml
+except ImportError:
```
PrettyWood, line 6: Probably a pragma no cover herr

changes/2908-jrwalk.md
jrwalk: oops, fixed

Makefile
jrwalk: you know, I think that was for VSCode's `mypy` runs to stop complaining (on `toml` as I recall). The CLI runs seem perfectly happy without it in a clean environment, so I'll remove it.

jrwalk: actually, yep `make mypy` complains about the `toml` stubs.  Lighter touch to just go ahead and add that specifically to the dev requirements.

jrwalk: done in https://github.com/samuelcolvin/pydantic/pull/2908/commits/067d7880402d1aa0f5ba673a17f135a00d327162

PrettyWood (approved):
Thank you so much 🙏

jrwalk:
Thanks for the review @PrettyWood, glad to help!

kkirsche:
Is there a plan for when this will be released? The last tagged release of pydantic appears to have been in May

Issue #2895, opened 2021-06-09
`pydantic-mypy` plugin is incompatible with `pyproject.toml` config in `mypy>0.900`


jrwalk:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:

```
pydantic version: 1.8.2
pydantic compiled: True
   install path: /opt/venv/lib/python3.8/site-packages/pydantic
   python version: 3.8.5 (default, Sep 10 2020, 16:58:22)  [GCC 8.3.0]
       platform: Linux-5.10.25-linuxkit-x86_64-with-glibc2.2.5
optional deps. installed: ['typing-extensions']
```

The newest release of `mypy` (0.900-0.901) has introduced `pyproject.toml` configuration in addition to the older `mypy.ini` in accordance with PEP-518. However, the `pydantic-mypy` plugin still expects an INI-formatted config file, such that TOML-configured mypy with the plugin enabled results in a parse error:

```py
Error constructing plugin instance of PydanticPlugin

Traceback (most recent call last):
  File "/opt/venv/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/opt/venv/lib/python3.8/site-packages/mypy/__main__.py", line 11, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "mypy/main.py", line 98, in main
  File "mypy/build.py", line 179, in build
  File "mypy/build.py", line 229, in _build
  File "mypy/build.py", line 475, in load_plugins
  File "mypy/build.py", line 453, in load_plugins_from_config
  File "pydantic/mypy.py", line 77, in pydantic.mypy.PydanticPlugin.__init__
  File "pydantic/mypy.py", line 114, in pydantic.mypy.PydanticPluginConfig.__init__
  File "/usr/local/lib/python3.8/configparser.py", line 697, in read
    self._read(fp, filename)
  File "/usr/local/lib/python3.8/configparser.py", line 1113, in _read
    raise e
configparser.ParsingError: Source contains parsing errors: 'pyproject.toml'
        [line 12]: ']\n'
        [line 17]: ']\n'
        [line 65]: ']\n'
...
```

Note that the plugin still works in `mypy>0.900` provided an INI configuration file is still provided. To bring the plugin up to speed for the newest mypy releases I believe we'll need a TOML parser option in addition to the default `ConfigParser` in the plugin constructor.

johnthagen:
[...]

PR #2984, opened 2021-07-14
Fix mypy plugin without a config or with pyproject.toml

JacobHayes:
## Change Summary

The mypy plugin's `PydanticPluginConfig` relied on instance level booleans, but those were never set without a config
file. Separately, for users using mypy with a pyproject.toml, the `PydanticPluginConfig` would try to parse the config
with ConfigParser.

To solve both of those issues, I just default both cases to setting all of the flags to the strict mode.

The modified code is under `# pragma: no cover`, but I confirmed w/ a local run of things. Fair to mark "Unit tests for the changes exist" ✔️?

## Checklist

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -109,12 +109,14 @@ class PydanticPluginConfig:
 
     def __init__(self, options: Options) -> None:
         if options.config_file is None:  # pragma: no cover
-            return
-        plugin_config = ConfigParser()
-        plugin_config.read(options.config_file)
-        for key in self.__slots__:
-            setting = plugin_config.getboolean(CONFIGFILE_KEY, key, fallback=False)
-            setattr(self, key, setting)
+            for key in self.__slots__:
+                setattr(self, key, True)  # default to strict without a config
+        else:
+            plugin_config = ConfigParser()
+            plugin_config.read(options.config_file)
+            for key in self.__slots__:
+                setting = plugin_config.getboolean(CONFIGFILE_KEY, key, fallback=False)
+                setattr(self, key, setting)
 
 
 def from_orm_callback(ctx: MethodContext) -> Type:
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -108,7 +108,8 @@ class PydanticPluginConfig:
     warn_untyped_fields: bool
 
     def __init__(self, options: Options) -> None:
-        if options.config_file is None:  # pragma: no cover
+        # NOTE: pyproject.toml reading not supported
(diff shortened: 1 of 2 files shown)


JacobHayes:
I think the only failing test is an unrelated flake, so please review when you have a chance. I know there are [concerns around the `toml` package](https://github.com/samuelcolvin/pydantic/pull/2151#issuecomment-735413630), so I avoided including that.

Just noticed https://github.com/samuelcolvin/pydantic/pull/2908, which is admittedly more thorough, but ends up importing `toml` (I don't think it requires a new dependency, given any `mypy` users will also have `toml` installed). So... throwing this out there as a simpler "at least don't break" alternative. 😁

johnthagen:
Cross referencing this to #2895

pydantic/mypy.py
```diff
@@ -108,13 +108,16 @@ class PydanticPluginConfig:
...
-            return
-        plugin_config = ConfigParser()
-        plugin_config.read(options.config_file)
-        for key in self.__slots__:
-            setting = plugin_config.getboolean(CONFIGFILE_KEY, key, fallback=False)
-            setattr(self, key, setting)
+        # NOTE: pyproject.toml reading not supported
+        if options.config_file is None or options.config_file.endswith('.toml'):  # pragma: no cover
+            for key in self.__slots__:
+                setattr(self, key, True)  # default to strict without a config
```
PrettyWood, line 114: Thank you for the PR!
I just reviewed the other one that adds support. I know toml is messed up but it's still stdlib library so no extra dependency on pydantic side. Once they switch to tomli, it will be in user en.
So for this PR there is still the question of "when there is no config file found" and I would have said "everything to False".
I don't even know if that's really important but you probably took more time than me to think about it. I would love to have your opinion

johnthagen: >  I know toml is messed up but it's still stdlib library so no extra dependency on pydantic side

What do you mean by this? That the `toml` PyPI package is part of the stdlib? It's third party as well (https://pypi.org/project/toml/).

Just wanted to make sure there was no confusion.

PrettyWood: There was oops 🙈 2am...50 reviews in 1 day.
I'm going to bed!
Thanks!

PrettyWood:
Btw if we go we
> Still this PR can be useful when there is no config file found and I would have said "everything is set to False" and not True.

which makes more sense to me, then we can simply close this PR right in favor of #2908?

JacobHayes:
Gonna close this in favor of #2908. As @PrettyWood notes, defaulting most settings to `False` (less strict) is probably closer to most user's desires.

PR #3000, opened 2021-07-19
chore(ci): update python 3.10 version

PrettyWood:
Following https://github.com/samuelcolvin/pydantic/pull/2885#issuecomment-882559655, let's update the python 3.10 version and fix the enum repr


samuelcolvin:
Lgtm

PR #3005, opened 2021-07-21
Add logic which handles case where either alias or field are acceptable

jpribyl:
See issue here https://github.com/samuelcolvin/pydantic/issues/3004
And related issue here https://github.com/tiangolo/fastapi/issues/3559


PrettyWood:
Please add the expected vs current behaviour before adding tests and change to your PR
I think the problem of schemas and aliases is deeper than that and should be tackled as a whole

samuelcolvin:
I think best that we wait for V2 for this.

PR #3018, opened 2021-07-24
Fix support for using a subclass of an annotation as a default

JacobHayes:
## Change Summary

A field can be set with a default that is a subclass of the type hint when set together:

```python
class MyStr(str):
    pass

class Simple(BaseModel):
    x: str = MyStr("test")
```

Additionally, a model can be instantiated with a subclass of the type hint: `Simple(x=MyStr("works"))`.

However, when using model subclasses where the base sets a type hint, but the default (eg: `MyStr`) is set in a
subclass, the `The type of {name}.{var_name} differs from the new default value` exception is raised:

```python
class MyStr(str):
    pass

class Base(BaseModel):
    x: str

class Sub(Base):
    x = MyStr("test")
```

I think rather than checking for an exact match on the type, we should be able to use `lenient_issubclass`. This seems
pretty sane and didn't break any of the existing tests. My use case is actually using an `Annotation(BaseModel)` as the
type hint, but setting subclass defaults like `Vendor(Annotation)`, but that should follow from this simpler test case.

## Checklist

### DIFF
--- a/changes/3018-JacobHayes.md
+++ b/changes/3018-JacobHayes.md
@@ -1,0 +1,1 @@
+Fix support for using a subclass of an annotation as a default
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -216,11 +216,14 @@ class ModelMetaclass(ABCMeta):
                         class_validators=vg.get_validators(var_name),
                         config=config,
                     )
-                    if var_name in fields and inferred.type_ != fields[var_name].type_:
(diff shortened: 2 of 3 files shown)


JacobHayes:
Please review when you have a moment. The change is a bit simpler ignoring whitespace changes.

PrettyWood (approved):
Makes sense! Thank you 🙏

Issue #1494, opened 2020-05-08
Improve BaseModel.__eq__ performance


zoopp:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /home/mihai/.local/share/virtualenvs/vmcs-smsc-nYWUkkuU/lib/python3.8/site-packages/pydantic
               python version: 3.8.2 (default, Apr  8 2020, 14:31:25)  [GCC 9.3.0]
                     platform: Linux-5.5.19-1-ck-x86_64-with-glibc2.2.5
     optional deps. installed: ['devtools']

```
I've been trying out pydantic in one of my projects and came across a serious performance issue that unfortunately I am not able to reproduce in isolation. At this point I'm unsure if I am doing something wrong or if there is an issue in pydantic.

Essentially, I have the following class declared in my project that I was initially using with standard dataclasses:

```python
import pydantic
from pydantic import dataclasses

#@dataclasses.dataclass
#class ProtocolValue:
class ProtocolValue(pydantic.BaseModel):
    type: str
    value: Any

#    @classmethod
#    def construct(cls, **kwargs):
#        return cls(**kwargs)
```

This class is used extensively. Instances are created, passed around and destroyed heavily and because of this, I made sure to use `construct` everywhere.

I have a couple of unit tests that simulate heavy workload which when I run them with the class deriving from `pydantic.BaseModel` they take around 3 minutes to run. If I use the `dataclass` form by removing the comment markers and fixing up the code above the the total time drops down to about 44 seconds.

By looking over some profiling data that I captured using `pytest-profile` the following two entries seem suspicious (they are in 2nd and 3rd place sorted by total time; 1st and 2nd sorted by total number of calls):
```
 ncalls          tottime    percall cumtime  percall filename:lineno(function)
 54371496   21.687    0.000   42.711    0.000   /usr/lib/python3.8/abc.py:96(__instancecheck__)
 54371496   21.023    0.000   21.024    0.000   {built-in method _abc._abc_instancecheck}
```

In comparison to pydantic.BaseModel, the dataclass variant has an insignificant number of calls:
```
      126    0.000    0.000    0.001    0.000 /usr/lib/python3.8/abc.py:96(__instancecheck__)
      126    0.000    0.000    0.000    0.000 {built-in method _abc._abc_instancecheck} 
 ```

[...]

Issue #578, opened 2019-06-05
remove use of "validation" replace with "parse"


samuelcolvin:
ref #284 and #360 

Pydantic is no a validation library, it's a parsing library.

It makes no guarantee whatsoever about checking the form of data you give to it; instead it makes a guarantee about the form of data you get out of it.

This sounds like an esoteric difference, but it has real practical consequences, eg.
* if you pass `"3"` (which is not an int) to an `int` field, pydantic will convert it to an int
* if you pass `3.14` (which again is not an int) to an `int` field, pydantic will convert it to an int (thereby "loosing information")

I think this is correct and I'm not interested in changing it, but we should be clear about what pydantic is/does - while I was annoyed by the manor of the question in #284 (sorry, wrong issue, I meant #360) I do understand the motivation for the question.

To fix this we should:
* [ ] rename `validate_model` to `parse_model` (this will require the old version to continue to work with a deprecation warning for 2 versions)
* [ ] rename `__get_validators__` to `__get_parser_functions__` or similar (again will require the old function to continue to work with a deprecation warning)
* [ ] add the above statement prominently to the docs
* [ ] remove use of "validate"/"validation" from the docs

This is a big, backwards incompatible change for no material benefit but I think it's worth it for clarity.

gangefors:
I have no objection to this renaming. I also think it's a better fit for pydantic use-cases.

We use it as an advanced dataclass so that we can ensure that the data we want to save in our DB is consistent with the data model we have specified. We also use it to mangle messages sent on a queue into a data model that we can easily use in our code and store it as json documents that adhere to the specifications of the models we have written.

We were using cerberus but it was way harder to read the model, and we really didn't need to validate the incoming data as long as it was convertible into the data type we wanted in the model. We can even mangle the data in really advanced ways using for example pre-parsing.

tldr;
I support this.

tiangolo:
Agreed. Makes sense.

dmontagu:
This was probably implied, but since it wasn't specifically called out in the checkboxes, I wanted to add just in case:

* the config values referencing `validate` should probably also be changed to say `parse` (namely `validate_all` and `validate_always`).

samuelcolvin:
I've decided to chill out on this a lot, I'm going to add roughly the above warning to the docs. but avoid all the renaming.

"Validate" seems to be much more commonly used than "parse", also one step of parsing is validating, so like it or not; pydantic does do validation. Just that it's primary purpose isn't validation.

[...]

PR #3066, opened 2021-08-07
Include types during model equality check

JacobHayes:
Currently, `BaseModel.__eq__` only considers the _dict_ representation of the two models. This leads to incorrect (imo) equality for models of different types but the same fields. For example (not contrived, ~real code I wrote):

```python3

class Type(BaseModel):
    description: str = ""
    class Config:
        frozen = True  # Only relevant for hash comparison below

class Int8(Type):
    pass

class Int16(Type):
    pass

# then...

>>> i8, i16 = Int8(), Int16()
>>> i8 == i16
True
>>> hash(i8) == hash(i16)
False
```

## Change Summary

This PR adds the model classes into the equality check to distinguish on type+content. Note that this new behavior more closely matches `__hash__` (for frozen models), which mixes the class + the values (not sure why the keys aren't included :shrug:)

@samuelcolvin suggests a very similar change in https://github.com/samuelcolvin/pydantic/issues/1494#issuecomment-630144720. I avoided most of the "breaking change" described there - we're still converting the models to dicts. However, the class check may still be a breaking change for some esoteric uses, including 1 (fixed) test: `test_generic_recursive_models`, which was arguably incorrect in the first place (comparing an instance with the TypeVar against one without).

## Related issue number

Tangentially: #1494

## Checklist

### DIFF
--- a/changes/3066-JacobHayes.md
+++ b/changes/3066-JacobHayes.md
@@ -1,0 +1,1 @@
+Include types during model equality check
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -822,7 +822,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
 
     def __eq__(self, other: Any) -> bool:
         if isinstance(other, BaseModel):
-            return self.dict() == other.dict()
+            return (self.__class__, self.dict()) == (other.__class__, other.dict())
         else:
             return self.dict() == other
 
--- a/tests/test_generics.py
+++ b/tests/test_generics.py
@@ -1077,7 +1077,7 @@ def test_generic_recursive_models(create_module):
     Model1 = module.Model1
     Model2 = module.Model2
     result = Model1[str].parse_obj(dict(ref=dict(ref=dict(ref=dict(ref=123)))))
-    assert result == Model1(ref=Model2(ref=Model1(ref=Model2(ref='123'))))
+    assert result == Model1[str](ref=Model2(ref=Model1(ref=Model2(ref='123'))))
 
 
 @skip_36
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -2047,11 +2047,19 @@ def test_new_union_origin():
 
 
 def test_equality():
-    class A(BaseModel):
-        description: str = ''
+    class Animal(BaseModel):
+        name: str = ''
 
-    class B(BaseModel):
-        description: str = ''
+    class Dog(Animal):
+        pass
 
-    assert A() == A()
-    assert A() != B()
+    class Cat(Animal):
+        pass
+
+    class Owner(BaseModel):
+        name: str = ''
+        pet: Animal
+
+    assert Dog() == Dog()
+    assert Dog() != Cat()
+    assert Owner(pet=Dog()) != Owner(pet=Cat())
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -2048,10 +2048,10 @@ def test_new_union_origin():
 
 def test_equality():
     class A(BaseModel):
-        description: str = ""
+        description: str = ''
 
     class B(BaseModel):
-        description: str = ""
+        description: str = ''
 
     assert A() == A()
     assert A() != B()
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -2044,3 +2044,14 @@ def test_new_union_origin():
         'properties': {'x': {'title': 'X', 'anyOf': [{'type': 'integer'}, {'type': 'string'}]}},
         'required': ['x'],
     }
+
+
+def test_equality():
+    class A(BaseModel):
+        description: str = ""
+
+    class B(BaseModel):
+        description: str = ""
+
+    assert A() == A()
+    assert A() != B()


JacobHayes:
If this extra `__eq__` check makes sense, please review when you have a minute @PrettyWood! 🙇

JacobHayes:
Hmm, this works fine for simple models, but doesn't work on nested ones as they are just turned into and compared as dicts.

Any suggestions on how that should be handled?

PrettyWood:
@JacobHayes I agree the current behaviour of `__eq__` is not always desired.
Personally I overwrote it on one of my project where I rely on it with
```py
        return self.__class__ == other.__class__ and tuple(self._iter()) == tuple(other._iter())
```
This is more consistent with `dataclass` (and `pydantic.dataclass`!) behaviour.
But this change needs some work because it breaks some stuff with generics and custom types.
And most importantly it will probably break some code for endusers that "don't care" about types and just want to compare the content.
Quite a lot of people want pydantic to be stricter but at the same time the "looseness" of it is one of its strength 🤷

JacobHayes:
Thanks for the snippet, I'll use your simpler one! I was about to suggest a page like the old [`flask` snippets](https://web.archive.org/web/20170608150532/http://flask.pocoo.org/snippets), but looks like they decided otherwise and deleted it. 😁

As you note on the breaks, do you think it best to just close this PR (fine with me)? I've seen some other talk about having a strict flag (and doing so here actually seems easy enough), but maybe that also falls into too much complexity for a relatively simple override.

PrettyWood:
I honestly think it would make sense for v2.
The current behaviour with `dict` comparison is even simpler to overwrite and this "loose" way of checking equality could be done by people who actually rely on it.

sveinugu:
Ran into this issue in my use case, and wanted to ping it. As mentioned above, I believe the Python standard is to include class equality in an equality check, and I think the opposite behaviour (only considering data content) is confusing and non-intuitive. The comparison to `dataclass` is a good example of this.

I also wanted to add another argument for implementing this in v2:

It has been a long-contented issue whether what pydantic does should be described using the term "parse" or "validate" (see https://github.com/pydantic/pydantic/issues/578). I see in the [v2 blog post](https://docs.pydantic.dev/blog/pydantic-v2/#im-dropping-the-word-parse-and-just-using-validate) that the issue will finally be settled, and "validate" is the winner. I am not very happy with this conclusion (and it also seems @samuelcolvin was not fully satisfied with this conclusion either, according to the smiley), but that is another issue and the ship has sailed for that one, I believe. (Granted, I did find pydantic myself by googling "python validation library" or similar, so there are at least practical arguments for this.)

I believe the main arguments for using the term "parse" can be most easily grasped by reading the influential blog post ["Parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) by Alexis King (which I btw discovered from a comment in the above-mentioned issue: https://github.com/pydantic/pydantic/issues/578#issuecomment-626575276). So the main point of this concept (also termed "type-driven development") is to parse data once according to the required model and then let this added information (that the data now adheres to the model) be stored/remembered by the data type. In that way, you would not need to validate the data ever again in your code base.  [Btw: adding a section on this in the pydantic docs would be awesome and a way to not fully cave in to the "validate" crowd! :) ]

Consequently, if one wants to use pydantic as the vessel for "type-driven design", the model itself is important for downstream logic, not only the parsed data. Hence, I think it would be natural for __eq__ to consider also include the data types in the checks.

The main counter-argument is, I assume, that lots of existing code (and especially tests) will start failing if this is added, so if this is to be changed, v2 at least seems the proper place to do it.

I believe the best solution would be to add a config flag to enable/disable this behaviour (as mentioned above: https://github.com/pydantic/pydantic/pull/3066#issuecomment-894636366). However, I don't think this should be the same as the `strict` flag, as one would probably not set `strict` to `True` if following the "type-driven design" concepts. One could argue that checking types in equality tests should be the default if `strict=False`, however this seems counter-intuitive on a semantic level, as the equality check itself is actually more strict in this case. So a `include_types_in_equality` (or better-worded version) is probably the best solution, IMO. It also follows the [zen of Python](https://peps.python.org/pep-0020/) ("explicit is better than implicit").

I would also argue for `include_types_in_equality` being set to `True` as default for v2, as `strict` is by default set to `False` (I believe).

dmontagu:
Revisiting this in v2; I think there are a few pieces of state beyond `__dict__` that we may want to check for equality (beyond the type, which I agree should match). In particular, you can see essentially the full object state information here:

https://github.com/pydantic/pydantic/blob/83b3c49e99ceb4599d9286a3d793cea44ac36d4b/pydantic/main.py#L245-L251

This includes `__dict__`, but also `__fields_set__` and `__private_attrs__`.

Currently, I am thinking we should have `__eq__` look for matches on `__dict__` and `__private_attrs__`, but _not_ `__fields_set__`. Would anyone object to that? @samuelcolvin let me know if I should create a separate issue for this.

An additional concern: If we do matching on type, there are some nuisances with equality checks for generics with different parametrizations. In particular, note the following behavior:

```python
T = TypeVar("T")
S = TypeVar("S")
class MyModel(BaseModel, Generic[T]):
    x: T
assert MyModel[T] is MyModel
assert MyModel[S] is not MyModel  # a subclass is created because S != T
```
Therefore, you could end up with:
```python
MyModel[S](x=1) != MyModel[T](x=1)  # due to type inequality
```
I don't think this inequality is meaningful, but I also am not super keen to go down the rabbit hole of ensuring we get exact type equality if you reparametrize a type with a single typevar (I expect this would wreak some havoc on the current approach to resolving recursive generic types, and I really don't want to revisit that without a better reason than this edge case.)

I am open to opinions about how to resolve the generic types equality issue (or to just leave it unresolved).

(Note that currently, `MyModel[int, str] != MyModel[int, T][str]`, but I think the caching implementation can be improved to fix this, so I'm less concerned about this as a source of equality misbehavior. I think it will be harder to resolve with generics that are re-parametrized with different typevars though.)

PR #3088, opened 2021-08-11
Reproduce and Simplify logical expression for Some Functions

yezz123:
## Change Summary

Function `Color.as_named` Reproduce with the following changes:

- Swap if/else branches
- Remove unnecessary else after guard condition

Function `Color.as_rgb_tuple` Reproduce with the following changes:

- Merge duplicate blocks `in` conditional

Function `parse_str` Reproduce with the following changes:

- Replace `if statement` with `if expression`

Function `parse_hsl` Reproduce with the following changes:

- Replace assignment with augmented assignment

Function FieldInfo.__repr_args__ Reproduce with the following changes:

- Replace `dict.get(x, None)` with `dict.get(x)`

Function `ModelField._get_mapping_value` Reproduce with the following changes:

- Replace multiple comparisons of same variable with `in` operator

Function `import_string.all_identical` Reproduce with the following changes:

- Use `any()` instead of for loop

### DIFF
--- a/pydantic/color.py
+++ b/pydantic/color.py
@@ -90,17 +90,17 @@ class Color(Representation):
         return self._original
 
     def as_named(self, *, fallback: bool = False) -> str:
-        if self._rgba.alpha is None:
-            rgb = cast(Tuple[int, int, int], self.as_rgb_tuple())
-            try:
-                return COLORS_BY_VALUE[rgb]
-            except KeyError as e:
-                if fallback:
-                    return self.as_hex()
-                else:
-                    raise ValueError('no named color found, use fallback=True, as_hex() or as_rgb()') from e
-        else:
+        if self._rgba.alpha is not None:
             return self.as_hex()
+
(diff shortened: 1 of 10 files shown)


PrettyWood:
I've always wondered the purpose of this kind of PRs (you're not the first one)
No disrespect or anything but what are you trying to achieve here? Is it for performance reasons?
Because after having a quick look at it, I feel like it's mostly a question of style.
Though I agree with most changes (more "pythonic"), I feel like a lot are less explicit and IMHO if there is no notable performance improvement explicitness is far more important

yezz123:
This kind of PRs can help others understand the code more easily, also support devs to be more simple with code Environment,  not hard for helping others understand it fast.
About my PR, the concept is to Simplify logical expression for Most of the Functions, also for the Code Quality can be an addons 😀

samuelcolvin:
some of these I agree with, some I disagree with.

But overall I don't think they're worth the changes, conflicts and potential for subtle changes to the logic.

Thanks for the submission though.

PR #3101, opened 2021-08-15
Check for `collections.abc.Mapping` instead of dict when converting a…

qhefb:
… pydantic field to a dict.

## Change Summary

We would like to be able use a custom immutable`frozendict` in fields as pydantic and have `dict()` continue to work as expected. So we should use a more generic type like `collections.abc.Mapping` instead of dict to make sure that we still convert it to a dict.

## Related issue number

N/A

## Checklist

please review

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -1,3 +1,4 @@
+import collections
 import sys
 import warnings
 from abc import ABCMeta
@@ -678,7 +679,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         value_exclude = ValueItems(v, exclude) if exclude else None
         value_include = ValueItems(v, include) if include else None
 
-        if isinstance(v, dict):
+        if isinstance(v, collections.abc.Mapping):
             return {
                 k_: cls._get_value(
                     v_,


qhefb:
please review

pydantic/main.py
```diff
@@ -678,7 +679,7 @@ def _get_value(
         value_exclude = ValueItems(v, exclude) if exclude else None
         value_include = ValueItems(v, include) if include else None
 
-        if isinstance(v, dict):
+        if isinstance(v, collections.abc.Mapping):
```
PrettyWood, line 682: It should be `typing.Mapping`.
But I'm not sure it's expected.
If I have
```py
    class Model(BaseModel):
        a: MyCustomMap[str, int]
```
I expect my value in the dict representation to be a instance of `MyCustomMap`, not a dict

PrettyWood:
If I have
```py
    class Model(BaseModel):
        a: MyCustomMap[str, int]
```
I expect my value in the dict representation to be a instance of `MyCustomMap`, not a dict.
Please open an issue before to talk about this

Issue #2345, opened 2021-02-12
OverflowError if np.inf for integer values


andreas-vester:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Pydantic Version: 1.7.3

The following code generates an ``OverflowError: cannot convert float infinity to integer``

```
import numpy as np
from pydantic import BaseModel


class Model(BaseModel):
    num: int


Model(num=np.inf)
```

On the other hand, using ``np.nan`` generates a ``type_error.integer``, which is more intuitive, in my view.

Shouldn't we get the same error for ``np.inf``, too?

PrettyWood:
Hi @KraxelHuber 
We could maybe add `OverflowError` in the caught errors here
https://github.com/samuelcolvin/pydantic/blob/61b45340e78a66af132d975ba063fae691db25a0/pydantic/validators.py#L121-L124
The only problem is that depending on the used libraries it won't have the same meaning and I'm afraid we'll need to add more and more exceptions that are not directly related to python or pydantic
It may be better to add your own validator

KraxelHuber:
I think ``np.inf`` should raise the same error as ``np.nan``. It simply is not a valid ``int``.

Kludex:
On V2 we have the following validation error:

```python
num
  Input should be a finite number [type=finite_number, input_value=inf, input_type=float]
```

PR #3112, opened 2021-08-18
fix int_validator not catching overflows

ojii:
## Change Summary

`int(value)` can raise an `OverflowError`, for example when called with very large floats.

## Related issue number

#2345 

## Checklist

### DIFF
--- a/changes/3112-ojii.md
+++ b/changes/3112-ojii.md
@@ -1,0 +1,1 @@
+fix overflow errors in int_validator
--- a/changes/3112-ojii.md
+++ b/changes/3112-ojii.md
@@ -1,1 +1,1 @@
-fix overflow errors in int_validator
+Catch overflow errors in `int_validator`
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -125,7 +125,7 @@ def int_validator(v: Any) -> int:
 
     try:
         return int(v)
-    except (TypeError, ValueError):
+    except (TypeError, ValueError, OverflowError):
         raise errors.IntegerError()
 
 
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -43,6 +43,18 @@ def test_int_validation():
     assert Model(a=4.5).a == 4
 
 
+@pytest.mark.parametrize('value', [2.2250738585072011e308, float('nan'), float('inf')])
+def test_int_overflow_validation(value):
+    class Model(BaseModel):
+        a: int
+
+    with pytest.raises(ValidationError) as exc_info:
+        Model(a=value)
+    assert exc_info.value.errors() == [
+        {'loc': ('a',), 'msg': 'value is not a valid integer', 'type': 'type_error.integer'}
+    ]
+
+
 def test_frozenset_validation():
     class Model(BaseModel):
         a: frozenset


ojii:
please review

PrettyWood:
I reckon it would be better to catch the `OverflowError` in a separate `except` clause with a more explicit error. We could probably do the exact same for `float`

ojii:
> I reckon it would be better to catch the `OverflowError` in a separate `except` clause with a more explicit error. 

Sorry, first time contributing to pydantic. Does that mean just changing the `msg` or the `type` of the error? If the latter, is there a way type codes are chosen?

pydantic/validators.py
```diff
@@ -126,7 +126,7 @@ def int_validator(v: Any) -> int:
 
     try:
         return int(v)
-    except (TypeError, ValueError):
+    except (TypeError, ValueError, OverflowError):
         raise errors.IntegerError()
```
dimaqq, line 130:
```suggestion
    except (TypeError, ValueError):
        raise errors.IntegerError()
    except OverflowError:
        raise errors.IntegerOverflowError()
```

I think this is meant.

dimaqq: Then, like 144

```diff
-        return float(v)
+        rv = float(v)
+        if rv in (float("inf"), float("-inf")):
+            raise errors.FloatOverflowError()
+        if rv != rv:
+            raise errors.FloatNanError()
+        return rv
```

P.S. the weird line of code from https://stackoverflow.com/a/44154660/705086

samuelcolvin: floats are a separate case. I would say that `float('inf')` is a valid float, arguably so is `float('nan')`.

tests/test_validators.py
```diff
@@ -41,6 +41,11 @@ class Model(BaseModel):
     assert Model(a=True).a == 1
     assert Model(a=False).a == 0
     assert Model(a=4.5).a == 4
+    with pytest.raises(ValidationError) as exc_info:
+        Model(a=2.2250738585072011e308)
+    assert exc_info.value.errors() == [
+        {'loc': ('a',), 'msg': 'value is not a valid integer', 'type': 'type_error.integer'}
+    ]
```
samuelcolvin, line 48: can you add the case of passing `float('inf')` etc.

samuelcolvin (commented):
This is looking good, just a few things to fix and more test cases to add.

samuelcolvin:
please update.

ojii:
> This is looking good, just a few things to fix and more test cases to add.

@samuelcolvin i've added more test cases, do you think the error code/msg should be handled different? If so, how are new messages/codes decided?

samuelcolvin:
At this stage, breast if you can roughly follow the code from pydantic-core to minimize the incompatibility with v2.

samuelcolvin:
https://github.com/samuelcolvin/pydantic-core/blob/fdf42b8b0d0684bc2cac04d34c40fe4b0d33498d/tests/validators/test_int.py#L207

ojii:
Since the structure of the error dict is different between v1 and v2 anyway, does it make sense trying to make it v2-like? I feel making it fit in with the rest of v1 makes more sense (eg in v1, error codes seem to always be prefixed with `type_error.` or `value_error.` but in v2 not?)

samuelcolvin:
Yes, that's what I meant by roughly follow, e.g. use `value_error.int_nan`?

ojii:
> Yes, that's what I meant by roughly follow, e.g. use `value_error.int_nan`?

When I change it to that, interestingly enough `float('nan')` gets caught in the already existing exceptions (`ValueError`). This proposed change (catching `OverflowError`) only catches `float('inf')` or other huge floats, so `int_nan` is probably not the right thing here.

changes/3112-ojii.md
```diff
@@ -0,0 +1 @@
+fix overflow errors in int_validator
```
samuelcolvin, line 1:
```suggestion
Catch overflow errors in `int_validator`
```

samuelcolvin (commented):
Okay, let's just stick with this then.

samuelcolvin:
Good catch, thanks so much for the PR.

PR #3121, opened 2021-08-19
✨ Add a config field to make all model fields optional

christophelec:
## Change Summary

This PR adds a Config to turn all fields of a model Optional with a None default.

The underlying use is to create a copy of an existing model, but where fields are Optional, to allow for an update of only the fields provided in FastAPI.

An example is provided in the issue below.

## Related issue number

Closes https://github.com/samuelcolvin/pydantic/issues/3120

It seems it would also help with https://github.com/samuelcolvin/pydantic/issues/2272#issuecomment-771205557

## Checklist

### DIFF
--- a/changes/3121-christophelec.md
+++ b/changes/3121-christophelec.md
@@ -1,0 +1,1 @@
+Add a `total` configuration field to create models with all fields Optional
--- a/changes/3121-christophelec.md
+++ b/changes/3121-christophelec.md
@@ -1,1 +1,1 @@
-Add a `total` configuration field to create models with all fields Optional
+Add a `total` configuration field to create models with all fields `Optional`
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -62,6 +62,7 @@ class BaseConfig:
     json_dumps: Callable[..., str] = json.dumps
     json_encoders: Dict[Type[Any], AnyCallable] = {}
     underscore_attrs_are_private: bool = False
+    all_optionals: bool = False
 
     # Whether or not inherited models as fields should be reconstructed as base model
     copy_on_model_validation: bool = True
@@ -96,7 +97,9 @@ class BaseConfig:
         """
         Optional hook to check or modify fields during model creation.
         """
-        pass
+        if cls.all_optionals:
+            field.type_ = Optional[field.type_]
+            field.required = False
 
 
 def inherit_config(self_config: 'ConfigType', parent_config: 'ConfigType', **namespace: Any) -> 'ConfigType':
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -62,7 +62,7 @@ class BaseConfig:
     json_dumps: Callable[..., str] = json.dumps
     json_encoders: Dict[Type[Any], AnyCallable] = {}
     underscore_attrs_are_private: bool = False
-    all_optionals: bool = False
+    total: bool = True
 
     # Whether or not inherited models as fields should be reconstructed as base model
     copy_on_model_validation: bool = True
@@ -97,7 +97,7 @@ class BaseConfig:
         """
         Optional hook to check or modify fields during model creation.
         """
-        if cls.all_optionals:
+        if not cls.total:
             field.type_ = Optional[field.type_]
             field.required = False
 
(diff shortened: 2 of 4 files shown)


christophelec:
The PR is ready, please review

changes/3121-christophelec.md
```diff
@@ -0,0 +1 @@
+Add a `total` configuration field to create models with all fields Optional
```
PrettyWood, line 1:
```suggestion
Add a `total` configuration field to create models with all fields `Optional`
```

PrettyWood (commented):
LGTM otherwise! Not the first time people asked for this.
And it can properly be used afterwards by FastAPI for update payloads where only a subset of your model is often sent
I'll still wait for @samuelcolvin opinion for new config params

PrettyWood:
Please update

christophelec:
Ok, the doc is updated, please review

lsapan:
@christophelec thanks for this! I actually just submitted a PR for the exact same thing accidentally 😅.

I'm wondering if you'd be open to checking mine out and possibly adding the helper class method? You can find the PR here:
https://github.com/samuelcolvin/pydantic/pull/3179

Basically, it just allows you to declare your full model, and then call `partial_model()` on it to generate the version that makes all the fields optional. Let me know what you think.

lsapan:
Actually I take that back - this is similar to my PR but not quite the same. This PR would still allow `null` values to make it into fields that should not be `null`, as long as they're still explicitly passed in the data. I'd encourage you to check out my PR as it doesn't allow non-nullable fields to be passed `null`.

pydantic/config.py
```diff
@@ -96,7 +97,9 @@ def prepare_field(cls, field: 'ModelField') -> None:
         """
         Optional hook to check or modify fields during model creation.
         """
-        pass
+        if not cls.total:
+            field.type_ = Optional[field.type_]
+            field.required = False
```
PrettyWood, line 102:
```suggestion
        if not cls.total:
            field.required = False
```
@lsapan is right. We shouldn't change the type and accept `None`.
Making it non required is enough

PrettyWood (commented):
I think it would be great to
- add a schema example to show there are no required keys
- add a test with "not nullable" keys like @lsapan said
- maybe a helper to generate it. If so doc as well

christophelec:
Sorry about the delay @lsapan, I checked your PR and found some key difference, that might explain why I found it necessary to change the type to Optional and allow null as you mention : 
In your PR, if the field is not set, it never appears when exporting to a dict, whether exclude_unset is set or not.
To talk in terms of code, in [your PR](https://github.com/samuelcolvin/pydantic/pull/3179/files#diff-a13941bfb0678e0826e2f725908c867214b98caca917e5a7df5d5a3d8f368344R1910): 

```python
    class Foo(BaseModel):
        a: str = Field(...)
        b: str = Field(...)
        c: Optional[str] = Field()

    Bar = create_partial_model(Foo)

    b = Bar(a='x', c=None)
    assert b.dict() == b.dict(exclude_unset=True) == {'a': 'x', 'c': None}
```

While in mine : 
```python
    b = Bar(a='x', c=None)
    assert b.dict() == {'a': 'x', 'b': None, 'c': None}
    assert b.dict(exclude_unset=True) == {'a': 'x', 'c': None}
```

I understand why you made this decision : it allows not to muddy the type by making it Optional, and overall fits better the use case of a patch for FastAPI (as I expect forgetting to use `exclude_unset` in this context would be major trap).

However, I think the fact that the field does not appear at all, whether `exclude_unset` is set or not is a bit surprising in general.

What do you think @PrettyWood @lsapan ? To me if we're ok with having the field not appear whether `exclude_unset` is set or not, I think we should merge your PR @lsapan .

Otherwise, I found an issue with my PR about the handling of `Field(...)`, I'll try to dig it down this weekend and update the PR with your requirements

lsapan:
@christophelec no problem at all! So for what it's worth, that's actually a slightly different discussion than the `Optional` one. My PR could always be changed to show all keys when you don't pass `exclude_unset=True` without making all the fields optional.

The problem with making all the fields optional is that (from your example above), you could do:
```
b = Bar(a=None)
```

According to your model, `a` should never be allowed to be _explicitly_ set as `None`, but it'll allow it. This means that if you have a non-nullable field in your model, and you go to allow PATCHing to it, it can suddenly be set as `null`, which is most likely not what you want.

samuelcolvin:
Thanks so much for this, I'm closing it in favour of #3179.

Although this PR was created first, #3179 is more complete and I prefer "partial" over "total"

PR #3133, opened 2021-08-25
Allow ellipsis on Fields inside Annotated

ezegomez:
## Change Summary

Allow ellipsis on Fields inside Annotated. 

The usecase is TypedDicts, which can't have default values otherwise.
Meaning that adding Optionals to Annotated drops them from the list of
required fields, which the patch fixes.

I believe the change is minor enough to not deserve its own issue. 

## Checklist


tests/test_annotated.py
```diff
@@ -26,6 +26,11 @@
             5,
             id='annotated-field-value-default',
         ),
+        pytest.param(
+            lambda: Annotated[int, Field(..., description='Test')],
+            5,
```
PrettyWood, line 31: I makes no sense to me to have both `...` (i.e. make it required) in `Field` and `5` as default value 🤔 
Can you explain the usage?

ezegomez: This was just me trying to make sure Annotated + Field with ellipsis worked on the test where you explicitly cover those. My usecase is the one from the other added test: being able to annotate Optional ("nullable") fields in a NamedDict so they get i.e. better descriptions, while keeping them required on the json schema. 

On looking at it again you're right it doesn't make sense to have ellipsis and a default, and accounting for not having a value set probably overcomplicates this test, so I'll just remove this.

tests/test_annotated_types.py
```diff
@@ -281,3 +281,34 @@ class Model(BaseModel):
...
+
+def test_typeddict_annotated_nonoptional():
+    class DataTD(TypedDict):
+        a: Optional[int]
+        b: Annotated[Optional[int], Field(...)]
+        c: Annotated[Optional[int], Field(..., description='Test')]
+        d: Annotated[Optional[int], Field()]
+
+    class Model(BaseModel):
+        data_td: DataTD
```
samuelcolvin, line 294: Can we add some tests for actually using the model, not just generating a schema?

pydantic/fields.py
```diff
@@ -401,7 +401,7 @@ def _get_field_info(
             field_info = next(iter(field_infos), None)
             if field_info is not None:
                 field_info.update_from_config(field_info_from_config)
-                if field_info.default is not Undefined:
+                if field_info.default not in (Undefined, Required):
```
samuelcolvin, line 404:
```suggestion
                if field_info.default not in {Undefined, Required}:
```
Weirdly I've seen in the past that checking containment in sets can be faster even when they're tiny.

samuelcolvin: But if you're not sure, might be worth checking.

samuelcolvin (commented):
Otherwise LGTM.

Almost scary how small this change was :confused:

samuelcolvin:
please update.

pydantic/fields.py
ezegomez: Some napkin testing on an i7 2020 macbook, homebrew 3.9 python:
```
from timeit import timeit
from enum import Enum
class TestEnum(Enum):
    Undefined = 1
    Required = 2
    Whatever = 3
def test():
    [a not in (TestEnum.Undefined, TestEnum.Required) for a in TestEnum]
def test_set():
    [a not in {TestEnum.Undefined, TestEnum.Required} for a in TestEnum]
timeit(test)
2.068943857000022
timeit(test_set)
3.550850099999934
```

Now granted this is not super accurate (not the same enum size, testing all options in order instead of sorting the most common tests, etc) but for the purposes of what we're checking, testing over what's effectively an array of size 2 is quite faster, vs finding hashes of all members, storing them, etc which is what you'd expect to happen with the set option.

tests/test_annotated_types.py
ezegomez: Can try, although all I had interest originally was the schema itself.

ezegomez:
> please update.

should be done (dunno how to remove the "needs author revision" tag)

samuelcolvin:
> > please update.
> 
> should be done (dunno how to remove the "needs author revision" tag)

As per the note in the body of this PR, you can just add "please review".

I'm looking now.

samuelcolvin:
Good catch, thanks so much.

PR #3159, opened 2021-09-01
Nested env

Air-Mark:
## Change Summary
- new ```env_nested_delimiter option``` in Config class
- optional ```_env_nested_delimiter``` kwarg in init ```BaseSettings```
- patched ```EnvSettingsSource```
Depending on ```env_nested_delimiter``` trying to find env variables with that delimiter and split those variable into nested dicts. Merge those dicts if they have same nesting structure.
So lets say we have 
```
class A(BaseSettings):
    foo: str


class B(BaseSettings):
    bar: A


class C(BaseSettings):
    baz: B
```
now we are able to fill the ```A.foo``` with env variable like
```BAZ__BAR__FOO='hello' ```

## Related issue number

#2304 #2232

## Checklist

### CODE REVIEW

**Comment by MadJlzz** [comment]
Exactly what I was looking for! I believe that `pydantic` is already doing an excellent job ; removing boiler plate code thanks to this PR will help us centralize settings configuration.

I can help if needed (doc, testing, ...)

cc @daviskirk, because https://github.com/daviskirk/climatecontrol#pydantic seems to have this already implemented.

**Comment by Air-Mark** [comment]
@MadJlzz it is nice to hear that it was useful for you. Sorry for disappearing for almost a moth... I cleaned up code and logic a bit. Added some docs and tests. I think this PR is now ready for submission. And of course It would be great to have some help if needed.

**Comment by Air-Mark** [comment]
please review

File: pydantic/env_settings.py [hunk 1]
```diff
@@ -132,18 +142,16 @@ def __repr__(self) -> str:
 
 
 class EnvSettingsSource:
```
samuelcolvin, line 144: I think on reflection, having reviewed the changes to this class, that this feature would better implemented as a separate `Source`. It's actually quite an elegant use of the sources approach.

pydantic/env_settings.py
```diff
@@ -158,6 +166,33 @@ def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
                     ),
                     **env_vars,
                 }
+        return env_vars
+
+    def explode_env_vars(self, settings: BaseSettings, env_vars: Mapping[str, Optional[str]]) -> Dict[str, Any]:
```
samuelcolvin, line 171: This equivalent method in the new source will need a docstring and comments to explain how it's working.

samuelcolvin (commented):
Thanks for this, I've wanted this since I first build pydantic, I didn't include it in the first release and never got around to implementing it.

I'm really excited to use this once it's done.

samuelcolvin:
please update

pydantic/env_settings.py
samuelcolvin: Humm, on reflection, I'm not so sure about this, but we can definitely make the logic clearer.

Air-Mark: It looks like is better to keep this new logic in one class. If we'll separate it for two `Source`s than we'll need to choose which `Source` to use in the `BaseSettings._build_values` method. Because we cannot use them both at the same time in a lack of efficiency, because they both will be reading the env_file.

Air-Mark:
please review

pydantic/env_settings.py
samuelcolvin: agreed.

pydantic/env_settings.py
```diff
@@ -170,25 +172,70 @@ def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
...
+                - Assign the value to the deepest key
+        """
+        result: Dict[str, Any] = {}
+        for env_name, env_val in env_vars.items():
+            keys = env_name.split(self.env_nested_delimiter)
+            env_var = result
+            for idx, key in enumerate(keys):
+                if idx == len(keys) - 1:
+                    try:
+                        env_val = settings.__config__.json_loads(env_val)  # type: ignore
```
samuelcolvin, line 231: I don't understand why you need this here? Also not sure why we're catching and ignoring `TypeError`

pydantic/env_settings.py
```diff
@@ -170,25 +172,70 @@ def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
...
+            - Before going through each key define env_var as a root (same as the final result)
+            - Go trough each key, add a new dict item to the env_var and redefine env_var within this item
+            - At the last iteration:
+                - Silently try to parse value with the json, leave it as it is if it fails
+                - Assign the value to the deepest key
+        """
+        result: Dict[str, Any] = {}
+        for env_name, env_val in env_vars.items():
+            keys = env_name.split(self.env_nested_delimiter)
+            env_var = result
```
samuelcolvin, line 227: I don't think this approach works.

If you change the order of env variables in your test below, the result changes. E.g. put `env.set('top__sub__v5', '5')` first.

samuelcolvin (commented):
A few things to change, you also need to rethink `explode_env_vars` significantly.

samuelcolvin:
please update.

pydantic/env_settings.py
```diff
@@ -132,32 +142,24 @@ def __repr__(self) -> str:
...
-
-        if self.env_file is not None:
-            env_path = Path(self.env_file).expanduser()
-            if env_path.is_file():
-                env_vars = {
-                    **read_env_file(
-                        env_path, encoding=self.env_file_encoding, case_sensitive=settings.__config__.case_sensitive
-                    ),
-                    **env_vars,
-                }
```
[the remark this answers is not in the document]
Air-Mark: Sure, returned this pice to its place. But was forced to add `noqa: C901` because now `__call__` is to complex.

pydantic/env_settings.py
```diff
@@ -170,25 +172,70 @@ def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
...
+    def explode_env_vars(self, settings: BaseSettings, env_vars: Mapping[str, Optional[str]]) -> Dict[str, Any]:
+        """
+        Go trough environment variables.
+        Within the each key-value pair:
+            - Split key with the env_nested_delimiter
+            - Before going through each key define env_var as a root (same as the final result)
+            - Go trough each key, add a new dict item to the env_var and redefine env_var within this item
+            - At the last iteration:
+                - Silently try to parse value with the json, leave it as it is if it fails
+                - Assign the value to the deepest key
```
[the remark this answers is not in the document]
Air-Mark: Agreed, just misunderstood you about the "explanation" for this method.

docs/examples/settings_nested_env.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+os.environ.update({
+    'top': '{"v1": "1", "v2": "2"}',
+    'v0': '0',
+    'top__v3': '3',
+    'top__sub': '{"sub_sub": {"v6": "6"}}',
+    'top__sub__v4': '4',
+    'top__sub__v5': '5',
+})
```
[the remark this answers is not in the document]
Air-Mark: Fixed. Misunderstood you last time, where exactly I should place this.

pydantic/env_settings.py
Air-Mark: Thank you for pointing on this. Yes there was a bug. Redesigned this method and test a bit according to this bug and logic from my answer on your comment above.

pydantic/env_settings.py
Air-Mark: Agreed about the `TypeError` we process only env values here, so there are only sting values.
We need this piece because env_val can be a json string. And there could be a situation when you want to mix both approaches of defining nested models:
```bash
export TOP__SUB__V5='5'
export TOP='{"sub": {"v6": "6"}}'
```
So the correct behaviour in my mind was that we should merge those variables into one, instead of overwriting.

Air-Mark:
Thank you for review. Did changes that you been asking for. Improved and redesigned `explode_env_vars` a bit. 
Please review

samuelcolvin:
Thanks so much for this, I'm really looking forward to using it.

I've made some changes to the logic to avoid a few errors I found. Please let me know if you have any and feedback or want to make any more changes, otherwie I'll merge over the next few days.

samuelcolvin:
Thanks so much for this, I've actually merged now as I want to include this in v1.9 and I want to get a pre-release version of v1.9 out today.

If you want to make any changes to this, feel free to create new PR.

Air-Mark:
That is awesome. Thank you for review and help. I can use a pre-release version in some projects, and will be glad to help with bugs if needed.

Issue #1223, opened 2020-02-12
How to have an “optional” field but if present required to conform to non None value?


mgcdanny:
# How to have an “optional” field but if present required to conform to non None value?

How can I have an optional field where None is not allowed?  Meaning a field may be missing but if it is present it should not be None. 

```py
from pydantic import BaseModel

class Foo(BaseModel):
    count: int
    size: float = None  # how to make this an optional float? 

 >>> Foo(count=5)
 Foo(count=5, size=None)  # GOOD - "size" is not present, value of None is OK

 >>> Foo(count=5, size=None)
 Foo(count=5, size=None) # BAD - if field "size" is present, it should be a float

 # BONUS
 >>> Foo(count=5)
 Foo(count=5)  # BEST - "size" is not present, it is not required to be present, so we don't care about about validating it all.  We are using Foo.json(exclude_unset=True) handles this for us which is fine.
```

I cross posted  SO:
https://stackoverflow.com/questions/60191270/python-pydantic-how-to-have-an-optional-field-but-if-present-required-to-con

samuelcolvin:
use a [validator](https://pydantic-docs.helpmanual.io/usage/validators/)

```py
from pydantic import BaseModel, validator

class Foo(BaseModel):
    count: int
    size: Optional[float] = None

    @validator('size')
    def prevent_none(cls, v):
        assert v is not None, 'size may not be None'
        return v
```

dmontagu:
You should be able to add a validator to the field that raises an error if called on `None`.

This works because, by default, validators aren't called on arguments that are not provided (though there is a keyword argument on `@validator` that can make this happen).

Another route would be to use a sentinel value (e.g., `object()`) as the default (this might require some `Config` changes to make it work, I'm not 100% sure), and add an `always=True` validator that converts the exact default value to `None`, and raises an error if `None` was provided.

-----

Note that the partial lack of idempotency may cause trouble with certain frameworks (like FastAPI) which may convert the model to dict form (which may have the None), and then call the initializer again on the dumped data.

There may be a way to achieve a similar pattern that can track whether the root source of the value was default initialization (defeating the idempotency issues), but you'd have to avoid the use of `None`. (Something that you ensure JSON-encodes to `None` might work though, if that's the context you are working in.)

acnebs:
[...]

PR #3179, opened 2021-09-06
Add support for partial model validation

lsapan:
## Change Summary

This is a simple change that will allow Pydantic models to be used for PATCH operations. It adds a `partial` config option which allows required fields to be omitted during validation. There has been various discussion on the topic of using pydantic for partial updates (particularly in the realm of FastAPI), but it seems like the suggested solutions were too complex for the perceived benefit. This change doesn't require adding any new types, it simply doesn't require fields to be present if the model has `partial = True` configured.

To illustrate why this change is necessary, imagine a scenario where you have a model with required fields, but you want to have an endpoint that can update just a few fields at a time:

```python
from typing import Optional

from pydantic import BaseModel, constr


class User(BaseModel):
    first_name: constr(min_length=1)
    last_name: constr(min_length=1)
    phone: Optional[constr(min_length=1)]

# We normally want the model to fail validation if last_name is left off, or null. i.e.:
User(first_name='John')
User(first_name='John', last_name=None)

# However, in the case of partial updates, we should be able to leave off last_name. We can do that now:
class PartialUser(BaseModel):
    first_name: constr(min_length=1)
    last_name: constr(min_length=1)
    phone: Optional[constr(min_length=1)]

    class Config:
        partial = True

# This no longer fails validation
PartialUser(first_name='John')

# This still fails validation, because last_name is still a required field
PartialUser(first_name='John', last_name=None)
``` 

There is an important distinction here. It has been recommended in the past to simply make every field optional, and use `dict(exclude_unset=True)`. However, that would allow setting `last_name=None`, bypassing the requirement that it always have a value. This change allows us to cleanly leave fields off when we aren't interested in updating them, while still ensuring they have a valid value when they're provided.

Speaking of clean, it's a bit impractical to define the same model twice! This change also introduces a class method to quickly generate a partial version of the model:

```python
PartialUser = User.partial_model()
```

## Related issue number

There are various issues in both this repo and FastAPI that have been seeking this functionality. My apologies for not opening a separate issue first, I just wanted to present the solution at the same time because it is an incredibly small change that adds a lot of flexibility. 👍 

## Checklist

### DIFF
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -116,7 +116,7 @@ not be included in the model schemas. **Note**: this means that attributes on th
 : whether or not inherited models used as fields should be reconstructed (copied) on validation instead of being kept untouched (default: `True`)
 
 **`partial`**
-: whether or not the model should allow required fields to be omitted. Useful when working with CRUD operations, and you need to use an existing model for a PATCH operation. (default: `False`)
+: whether or not the model should allow required fields to be omitted. Useful when working with CRUD operations, and you need to use an existing model for a PATCH operation. Subclassing a partial model will result in any additional fields also not being required. (default: `False`)
 
 ## Change behaviour globally
 
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -115,6 +115,9 @@ not be included in the model schemas. **Note**: this means that attributes on th
 **`copy_on_model_validation`**
 : whether or not inherited models used as fields should be reconstructed (copied) on validation instead of being kept untouched (default: `True`)
 
+**`partial`**
+: whether or not the model should allow required fields to be omitted. Useful when working with CRUD operations, and you need to use an existing model for a PATCH operation. (default: `False`)
+
 ## Change behaviour globally
 
 If you wish to change the behaviour of _pydantic_ globally, you can create your own custom `BaseModel`
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -62,6 +62,7 @@ class BaseConfig:
     json_dumps: Callable[..., str] = json.dumps
     json_encoders: Dict[Type[Any], AnyCallable] = {}
     underscore_attrs_are_private: bool = False
+    partial: bool = False
 
     # Whether or not inherited models as fields should be reconstructed as base model
     copy_on_model_validation: bool = True
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -97,7 +97,8 @@ class BaseConfig:
         """
         Optional hook to check or modify fields during model creation.
         """
-        pass
+        if cls.partial:
+            field.required = False
 
 
 def inherit_config(self_config: 'ConfigType', parent_config: 'ConfigType', **namespace: Any) -> 'ConfigType':
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -981,6 +981,8 @@ def validate_model(  # noqa: C901 (ignore complexity)
             if field.required:
                 errors.append(ErrorWrapper(MissingError(), loc=field.alias))
                 continue
+            elif config.partial:
+                continue
 
             value = field.get_default()
 
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -756,6 +756,20 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         for f in cls.__fields__.values():
             update_field_forward_refs(f, globalns=globalns, localns=localns)
 
+    @classmethod
+    def partial_model(cls: Type['Model']) -> Type['Model']:
+        """
+        Create a version of the model that is suitable for partial validation.
+        Useful when working with CRUD operations, and you need to use an
+        existing model for a PATCH operation.
+        """
+
+        class PartialModel(cls):  # type: ignore
+            class Config(cls.Config):  # type: ignore
+                partial = True
+
+        return PartialModel
+
     def __iter__(self) -> 'TupleGenerator':
         """
         so `dict(model)` works
@@ -978,6 +992,9 @@ def validate_model(  # noqa: C901 (ignore complexity)
             using_name = True
 
         if value is _missing:
+            if config.partial:
+                continue
+
             if field.required:
                 errors.append(ErrorWrapper(MissingError(), loc=field.alias))
                 continue
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -978,9 +978,6 @@ def validate_model(  # noqa: C901 (ignore complexity)
             using_name = True
 
         if value is _missing:
-            if config.partial:
-                continue
-
             if field.required:
                 errors.append(ErrorWrapper(MissingError(), loc=field.alias))
                 continue
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -756,20 +756,6 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         for f in cls.__fields__.values():
             update_field_forward_refs(f, globalns=globalns, localns=localns)
 
-    @classmethod
-    def partial_model(cls: Type['Model']) -> Type['Model']:
-        """
-        Create a version of the model that is suitable for partial validation.
-        Useful when working with CRUD operations, and you need to use an
-        existing model for a PATCH operation.
-        """
-
-        class PartialModel(cls):  # type: ignore
-            class Config(cls.Config):  # type: ignore
-                partial = True
-
-        return PartialModel
-
     def __iter__(self) -> 'TupleGenerator':
         """
         so `dict(model)` works
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -91,4 +91,5 @@ def create_partial_model(model: Type[BaseModel]) -> Type[BaseModel]:
         class Config(model.Config):  # type: ignore
             partial = True
 
+    PartialModel.__name__ = model.__name__
     return PartialModel
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -3,11 +3,12 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Any, Callable, Optional, Type, TypeVar, Union
 
+from .main import BaseModel
 from .parse import Protocol, load_file, load_str_bytes
 from .types import StrBytes
 from .typing import display_as_type
 
-__all__ = ('parse_file_as', 'parse_obj_as', 'parse_raw_as')
+__all__ = ('parse_file_as', 'parse_obj_as', 'parse_raw_as', 'create_partial_model')
 
 NameFactory = Union[str, Callable[[Type[Any]], str]]
 
@@ -77,3 +78,17 @@ def parse_raw_as(
         json_loads=json_loads,
     )
     return parse_obj_as(type_, obj, type_name=type_name)
+
+
+def create_partial_model(model: Type[BaseModel]) -> Type[BaseModel]:
+    """
+    Create a version of the model that is suitable for partial validation.
+    Useful when working with CRUD operations, and you need to use an
+    existing model for a PATCH operation.
+    """
+
+    class PartialModel(model):  # type: ignore
+        class Config(model.Config):  # type: ignore
+            partial = True
+
+    return PartialModel
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -1899,17 +1899,54 @@ def test_config_partial():
     class Foo(BaseModel):
         a: str = Field(...)
         b: str = Field(...)
+        c: Optional[str] = Field()
 
     Bar = create_partial_model(Foo)
 
+    # Ensure partial is only set on the new class
     assert Foo.Config.partial is False
     assert Bar.Config.partial is True
 
-    b = Bar(a='x')
-    assert b.dict(exclude_unset=True) == {'a': 'x'}
+    # Verify b (a required field) can be omitted on the partial model
+    b = Bar(a='x', c=None)
+    assert b.dict() == {'a': 'x', 'c': None}
+    assert b.dict(exclude_unset=True) == {'a': 'x', 'c': None}
+    assert dict(b) == {'a': 'x', 'c': None}
 
+    # Verify b cannot be passed explicitly as None
     with pytest.raises(ValidationError) as exc_info:
         b = Bar(a='x', b=None)
     assert exc_info.value.errors() == [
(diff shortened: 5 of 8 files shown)


lsapan:
Thanks for such an awesome library! Please review when you have time 😄

pydantic/main.py
```diff
@@ -756,6 +756,20 @@ def update_forward_refs(cls, **localns: Any) -> None:
         for f in cls.__fields__.values():
             update_field_forward_refs(f, globalns=globalns, localns=localns)
 
+    @classmethod
+    def partial_model(cls: Type['Model']) -> Type['Model']:
```
PrettyWood, line 760: Adding again a new classmethod scares me a bit. Could we maybe have a more functional approach like `create_model`?
Something added in `pydantic.tools` perhaps?

docs/usage/model_config.md
```diff
@@ -115,6 +115,9 @@ not be included in the model schemas. **Note**: this means that attributes on th
 **`copy_on_model_validation`**
 : whether or not inherited models used as fields should be reconstructed (copied) on validation instead of being kept untouched (default: `True`)
 
+**`partial`**
```
PrettyWood, line 118: I let you see with the author of the other PR and others for the best wording (`total` was my suggestion to mimic `TypedDict` but that's just an idea)

tests/test_edge_cases.py
```diff
@@ -1892,3 +1892,17 @@ class Config:
...
+    class Foo(BaseModel):
+        a: str = Field(...)
+        b: str = Field(...)
+
+    Bar = Foo.partial_model()
+
+    assert Foo.Config.partial is False
+    assert Bar.Config.partial is True
+
+    b = Bar(a='x')
```
PrettyWood, line 1907: You can add something like in the PR description with the `None` usecase

pydantic/main.py
```diff
@@ -978,6 +992,9 @@ def validate_model(  # noqa: C901 (ignore complexity)
             using_name = True
 
         if value is _missing:
+            if config.partial:
```
PrettyWood, line 995: I feel like making the field non required is better and more expected.
If you introspect your models or generate a JSON schema, you expect nothing in required keys

pydantic/main.py
lsapan: Haha, absolutely. I was just trying to avoid another import in any file that is doing a PATCH operation (in the context of FastAPI).

docs/usage/model_config.md
lsapan: Yeah I'm quite flexible on phrasing. I was basing it off of Django REST Framework as that's where I'm coming to FastAPI / pydantic from.

pydantic/main.py
lsapan: That's a good point! I was being overly protective in the event someone didn't pass `exclude_unset=True` to `dict()`. By short-circuiting here, it protects against ever accidentally getting `null` values for non-nullable fields.

lsapan:
@PrettyWood thanks for the review! I changed it from a classmethod to a function in `pydantic.tools`. I also added additional test coverage to demonstrate that passing `null` raises a validation error.

Didn't rename it yet, though that can be easily done! It looked like you were waiting to hear from @samuelcolvin in the other PR so I figured I'd just leave it for now until he chimes in.

As for the last bit about setting `required = False` versus leaving it out of the data altogether, I can see the benefits of both, so I'll definitely defer to whatever you guys land on.

Thanks again! 🎉

PrettyWood:
IMO having the right JSON schema is important especially when you recall why people want this (mostly for FastAPI).
Setting only `required=False` won't make the field allow `None` as value.

lsapan:
@PrettyWood sure thing! I just changed it to set `required = False` on all the fields instead. Would you be open to the `partial` name? A few reasons:

1. It's easier to talk about the model as a "partial model", rather than a "model with total = False", or "model without totality", etc.
2. The name of the helper function would also be a bit confusing (for similar reasons): `create_model_without_totality` ?
3. Because of the behavior I just changed, it actually deviates from `TypedDict`'s handling of totality:

```python
Python 3.9.4 (default, Apr 28 2021, 16:15:19) 
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
>>> from typing import TypedDict
>>> class Movie(TypedDict, total=False):
...     name: str
...     year: int
... 
>>> m: Movie = {}
>>> m2: Movie = {'year': 2015}
>>> m
{}
>>> m2
{'year': 2015}
```

`TypedDict` instances with `total = False` don't show the missing values as `None`, so it may actually make sense to use a separate name as to avoid confusion.

Just my two cents, happy to change it if desired!

PrettyWood:
You can still add
```diff
             if field.required:
                 errors.append(ErrorWrapper(MissingError(), loc=field.alias))
                 continue
+            elif config.partial:
+                continue
```
to support `dict/json` serialization ;) Best of both worlds.

PrettyWood:
There are still some questions that need to be answered in the doc:
What happens if you inherit a class from another? Do you expect all the fields to be not required or only the one in your class?

----------
I can't wait to work on v2 and https://github.com/samuelcolvin/pydantic/discussions/2353.
Making a clear difference between not required and nullable fields ;)

lsapan:
@PrettyWood

> You can still add
> 
> ```diff
>              if field.required:
>                  errors.append(ErrorWrapper(MissingError(), loc=field.alias))
>                  continue
> +            elif config.partial:
> +                continue
> ```
> 
> to support `dict/json` serialization ;) Best of both worlds.

Ah, I had misread what you were asking for earlier! Added this back.

> Can you add tests with `dict` serialization and `schema`? 🙏
> Also there are still some questions that need to be answered in the doc:
> What happens if you inherit a class from another? Do you expect all the fields to be not required or only the one in your class?
> 
> I can't wait to work on v2 and #2353.
> Making a clear difference between not required and nullable fields ;)

Sure thing! I've added a bunch more coverage. Let me know if there's anything else you want to see. I'm also quite excited for v2!

I noticed one interesting thing when testing subclasses of partial models. The subclass inherits partial (as would be expected), so no fields are required, **however**:

```python

class ModelA(BaseModel):
    a: str = Field(...)

    class Config:
        partial = True


class ModelB(ModelA):
    b: str = Field(...)


class ModelC(ModelA):
    c: str


# No error
ModelB(b=None)

# Validation error
ModelC(c=None)
```

I'd expect both to produce a validation error. I'm not familiar enough with the magic happening in `Field(...)` to understand what could possibly be happening here. Any ideas?

PrettyWood:
Yup you should probably put an extra condition when `infer`ring the field in the metaclass
```diff
-        if value is Required:
+        if config.partial:
+            required = False
+        elif value is Required:
```

lsapan:
@PrettyWood that did the trick! It's passing with the `= Field(...)` syntax now. Thanks! 🎉

lsapan:
@PrettyWood do you need anything else from me for approval? 🚀  Not trying to rush you at all - just want to make sure you're not waiting on me. Thanks!

pydantic/tools.py
```diff
@@ -77,3 +78,18 @@ def parse_raw_as(
...
+    Create a version of the model that is suitable for partial validation.
+    Useful when working with CRUD operations, and you need to use an
+    existing model for a PATCH operation.
+    """
+
+    class PartialModel(model):  # type: ignore
+        class Config(model.Config):  # type: ignore
+            partial = True
+
+    PartialModel.__name__ = model.__name__
```
christophelec, line 94: I'm curious, why do you make the model name the same as the original one ?

I personally think it's a bit confusing but I don't understand why you want to do it in the first place, so I might be missing a piece here

tests/test_edge_cases.py
```diff
@@ -1892,3 +1893,60 @@ class Config:
...
+
+    Bar = create_partial_model(Foo)
+
+    # Ensure partial is only set on the new class
+    assert Foo.Config.partial is False
+    assert Bar.Config.partial is True
+
+    # Verify b (a required field) can be omitted on the partial model
+    b = Bar(a='x', c=None)
+    assert b.dict() == {'a': 'x', 'c': None}
```
christophelec, line 1912: For reference as I mentioned it in [my PR](https://github.com/samuelcolvin/pydantic/pull/3121#issuecomment-925955894) : having `b` not appear even though we did not set `exclude_unset` is a surprising behavior to me, but it might be because I implemented it in a different way which shaped my expectations differently

pydantic/tools.py
lsapan: @christophelec it would show up as `PartialModel` otherwise 🙂

tests/test_edge_cases.py
lsapan: @christophelec yeah as I mentioned [here](https://github.com/samuelcolvin/pydantic/pull/3121#issuecomment-925960693), it can be easily changed. It just made more sense to me in the context of a "partial" model coming from a Django REST Framework background.

lsapan: @christophelec also just to flesh out the logic behind not showing the fields a bit more:

Given the model:
```python
class Foo(BaseModel):
    a: str
    b: Optional[str]
```

Looking at this, it's clear that `a` will always have at least a string value, and `b` may or may not.

As such, it's a bit misleading to ever assign `None` to a, if we were to include it:
```
f = Foo(b='hi')

# It feels strange / risky to assign None to a value that normally could not be None, based on the model's definition
f.dict()  # {'a': None, 'b': 'hi'}
```

Granted yes, the user should be using `exclude_unset=True` _anyway_, but returning None on a non-nullable field still feels like an impractical decision. Just my two cents though! Happy to change it if others disagree!

camflan:
Bump, this would be really useful to have built-in!

docs/usage/model_config.md
samuelcolvin: I agree on "partial", because:
1. It matches (roughly) TypeScript which is kind of the gold standard for type hints in dynamic languages
2. It means you can talk about "partial models" which is cleaner than "non-total models"

samuelcolvin: But can we add a note to the docs explaining that the word "total" could also be used with the opposite meaning.

pydantic/config.py
```diff
@@ -96,7 +97,8 @@ def prepare_field(cls, field: 'ModelField') -> None:
         """
         Optional hook to check or modify fields during model creation.
         """
-        pass
+        if cls.partial:
```
samuelcolvin, line 100: I think this could be a breaking/confusing change, since the default `prepare_field` does nothing, people will override it without calling `super()` and wonder why `partial` is not being respected.

pydantic/tools.py
samuelcolvin: How about something like

```suggestion
    PartialModel.__name__ = f'Partial{model.__name__}'
```
?

pydantic/main.py
```diff
@@ -981,6 +981,8 @@ def validate_model(  # noqa: C901 (ignore complexity)
             if field.required:
                 errors.append(ErrorWrapper(MissingError(), loc=field.alias))
                 continue
+            elif config.partial:
+                continue
```
samuelcolvin, line 985: This is kind of the cruz of the question about partial and how it should behave.

The problem is that with this (or any implementation of partial/non-total) the runtime instance no-longer matches the definition of the class in code.

I can imagine people legitimating wanting this to work both ways.

What if we added another config property `config.partial_omit`, then based on that, either omitted the attribute or made it `None`?

I think it would also be smart if we could change the `AttributeError` when accessing omitted properties to something like

    AttributeError: 'MyModel' object has no attribute 'bar'  - this is a partial model and the field was omitted

But only if we can do this easily and performantly.

tests/test_edge_cases.py
```diff
@@ -1892,3 +1893,60 @@ class Config:
...
+    assert b.dict(exclude_unset=True) == {'a': 'x', 'c': None}
+    assert dict(b) == {'a': 'x', 'c': None}
+
+    # Verify b cannot be passed explicitly as None
+    with pytest.raises(ValidationError) as exc_info:
+        b = Bar(a='x', b=None)
+    assert exc_info.value.errors() == [
+        {'loc': ('b',), 'msg': 'none is not an allowed value', 'type': 'type_error.none.not_allowed'}
+    ]
+
```
samuelcolvin, line 1922: Can we split this into multiple tests, even if it involves duplicating model definitions.

samuelcolvin (commented):
Looking good. This will be really useful.

A few things to change, then we will need a lot more docs.

bbatliner:
One of the original use cases discussed in #1223 was nested partial models, e.g.:
```python3
class Foo(BaseModel):
  a: str
  b: str

class Bar(BaseModel):
  foo: Foo
  c: str
```

See @sborovic's [solution](https://github.com/samuelcolvin/pydantic/issues/1223#issuecomment-985030887) or my [solution](https://github.com/samuelcolvin/pydantic/issues/1223#issuecomment-998160737) on that issue for a couple of reference implementations.

Correct me if I'm wrong, but this PR only "optionalizes" the top layer of a model, not any nested model fields? Will this work?

```python3
PartialFoo = Foo.partial_model()
PartialBar = Bar.partial_model()
PartialBar(foo=PartialFoo(b="abc"))
```

macintacos:
Friendly ping - @lsapan are there any updates expected for this PR, or should someone else take over?

lsapan:
Hey @macintacos! Sorry for the radio silence on my end - I'm working on a different project that isn't FastAPI based and I haven't had time to sit down and work on this. I'm hoping to in the next few months, but definitely happy to have someone else take over if they have time!

samuelcolvin:
Honestly, I think we should probably pause this.

I've never been comfortable with partial models since the data no longer matches the types.

This will be much much more elegant in pydantic v2 when typeddicts will be first class citizens.

I know it's annoying, bit I think the best answer is to wait for v2.

BrutalSimplicity:
This is an all or none approach, but is there a simple way to say exclude this single field?

bhushanatdukaan:
`    product_description: Optional[constr(max_length=500)] = ""
    product_category: Optional[constr(max_length=100)] = ""
    quantity: conint(ge=1)
    gst_rate: Optional[condecimal(ge=0, le=100, decimal_places=2, max_digits=5)] = 0`

You all can use Optional and give pass an default field check constr or any field validation to add partial model validation. You can also pass default as well.

samuelcolvin:
TypedDicts in pydantic-core already support `total=False` which achieves the same features as this in a type-safe way.

Closing this feature and deferring the feature until V2.

mrlubos:
@samuelcolvin (im)patiently waiting for v2! 🚀

samuelcolvin:
Me too.

munkim:
So when is V2 coming 😭

PR #3201, opened 2021-09-11
fix: models copied via `Config.copy_on_model_validation` always have all fields

### LINKED ISSUE #3195: Field(..., exclude=True) are missing from nested BaseModel objects
### Checks

# Bug

```
             pydantic version: 1.8.2
            pydantic compiled: False
                 install path: /home/venv/lib/python3.7/site-packages/pydantic
               python version: 3.7.10 (default, May  4 2021, 13:51:36)  [GCC 11.1.1 20210428 (Red Hat 11.1.1-1)]
                     platform: Linux-4.4.0-19041-Microsoft-x86_64-with-fedora-34-Thirty_Four
```

```py
from pydantic import BaseModel, Field, SecretStr


class User(BaseModel):
    id: int
    username: str
    password: SecretStr = Field(..., exclude=True)


class Transaction(BaseModel):
    id: str
    user: User = Field(..., exclude={'username'})
    value: int

    class Config:
        fields = {'value': {'exclude': True}}


t = Transaction(
    id='1234567890',
    user=User(
        id=42,
        username='JohnDoe',
        password='hashedpassword'
    ),
    value=9876543210,
)

print(t.user.password)
print(t.dict())
```

Hi, 
I'm currently interested by the exclude parameter (tested from commit 45db4ad3aa558879824a91dd3b011d0449eb2977).
After BaseModel initialization, `excluded` fields are missing from nested objects.
the BaseModel __init__ uses the `validate()` function that uses `copy()` ( and -> _iter) under the hood. 
https://github.com/samuelcolvin/pydantic/blob/45db4ad3aa558879824a91dd3b011d0449eb2977/pydantic/main.py#L652-L656

as a result, all excluded fields are missing after validation. 
I do not know if this behavior is intended. 
But as far as I can see the exclude should only affect exports function, right ? 

my issue looks in contradiction with this comment bellow. let me know if I misinterpreted the exclude purpose.
https://github.com/samuelcolvin/pydantic/issues/830#issuecomment-842313022

thank you very much

PrettyWood:
Small regression in #2231.
The shallow copy done with `Config.copy_on_model_validation = True` (default behaviour)
was using excluded / included fields when it should just copy everything

closes #3195

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -580,6 +580,24 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         m._init_private_attributes()
         return m
 
+    def _copy_and_set_values(self: 'Model', values: 'DictStrAny', fields_set: 'SetStr', *, deep: bool) -> 'Model':
+        if deep:
+            # chances of having empty dict here are quite low for using smart_deepcopy
+            values = deepcopy(values)
+
+        cls = self.__class__
+        m = cls.__new__(cls)
+        object_setattr(m, '__dict__', values)
+        object_setattr(m, '__fields_set__', fields_set)
+        for name in self.__private_attributes__:
+            value = getattr(self, name, Undefined)
+            if value is not Undefined:
+                if deep:
+                    value = deepcopy(value)
+                object_setattr(m, name, value)
+
+        return m
+
     def copy(
         self: 'Model',
         *,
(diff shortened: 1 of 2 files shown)


PrettyWood:
Please review

NunchakusLei:
Is this PR raised the issue https://github.com/samuelcolvin/pydantic/issues/3641?

PR #3264, opened 2021-09-26
Added Runtype to benchmarks

erezsh:
## Change Summary

Added a new benchmark, for the library Runtype.

Please review.

### DIFF
--- a/benchmarks/requirements.txt
+++ b/benchmarks/requirements.txt
@@ -12,3 +12,4 @@ cattrs
 cerberus
 voluptuous
 schematics
+runtype
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -64,6 +64,12 @@ except Exception as e:
     print('WARNING: unable to import TestSchematics')
     TestSchematics = None
 
+try:
+    from test_runtype import TestRuntype
+except Exception as e:
+    print('WARNING: unable to import TestRuntype')
+    TestSchematics = None
+
 PUNCTUATION = ' \t\n!"#$%&\'()*+,-./'
 LETTERS = string.ascii_letters
 UNICODE = '\xa0\xad¡¢£¤¥¦§¨©ª«¬ ®¯°±²³´µ¶·¸¹º»¼½¾¿ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ'
@@ -80,6 +86,7 @@ other_tests = [
     TestSchematics,
     TestDRF,
     TestCerberus,
+    TestRuntype,
 ]
 active_other_tests = [t for t in other_tests if t is not None]
 
--- a/benchmarks/test_runtype.py
+++ b/benchmarks/test_runtype.py
@@ -1,0 +1,55 @@
+from datetime import datetime
+from typing import List
+
+from runtype import __version__, dataclass, String, Int
+
+
+class TestRuntype:
+    package = 'runtype'
+    version = __version__
+
+    def __init__(self, allow_extra):
+        @dataclass(check_types='cast')
+        class Model:
+            id: int
+            client_name: String(max_length=255)
+            sort_index: float
+            # client_email: EmailStr = None
+            client_phone: String(max_length=255) = None
+
+            @dataclass
+            class Location:
+                latitude: float = None
+                longitude: float = None
+
+            location: Location = None
+
+            contractor: Int(min=0) = None
+            upstream_http_referrer: String(max_length=1023) = None
+            grecaptcha_response: String(min_length=20, max_length=1000)
+            last_updated: datetime = None
+
+            @dataclass
+            class Skill:
+                subject: str
+                subject_id: int
+                category: str
+                qual_level: str
+                qual_level_id: int
+                qual_level_ranking: float = 0
+
+            skills: List[Skill] = []
+
+            class Config:
+                extra: dict = dict
+
+        self.model = Model
+
+    def validate(self, data):
+        try:
+            return True, self.model(**data)
+        except TypeError as e:
+            return False, str(e)
+
+    def to_json(self, model):
+        return model.json()


ehiggs:
Could you add `runtype` to the `benchmarks/requirements.txt`?

ehiggs:
Great, using `make install-benchmarks` and then `python3 benchmarks/run.py` on my box runs the benchmark suite fine:

```
                pydantic best=104.764μs/iter avg=165.047μs/iter stdev=53.357μs/iter version=1.8.2
           attrs + cattrs best=125.587μs/iter avg=129.364μs/iter stdev=6.426μs/iter version=21.2.0
                 valideer best=170.376μs/iter avg=174.882μs/iter stdev=4.087μs/iter version=0.4.2
              marshmallow best=391.559μs/iter avg=404.856μs/iter stdev=10.565μs/iter version=3.14.0
               voluptuous best=367.023μs/iter avg=393.055μs/iter stdev=17.915μs/iter version=0.12.2
                 trafaret best=423.219μs/iter avg=431.278μs/iter stdev=7.389μs/iter version=2.1.0
               schematics best=790.590μs/iter avg=1097.016μs/iter stdev=355.515μs/iter version=2.1.1
    django-rest-framework best=915.720μs/iter avg=1172.045μs/iter stdev=229.796μs/iter version=3.12.4
                 cerberus best=1248.596μs/iter avg=1344.066μs/iter stdev=124.771μs/iter version=1.3.4
                  runtype best=38.760μs/iter avg=39.820μs/iter stdev=1.032μs/iter version=0.2.4
```

PrettyWood:
Sorry but for now we can't accept this since we already refused a lot of other libraries for benchmark
See https://github.com/samuelcolvin/pydantic/pull/1810#issuecomment-706050989

erezsh:
@PrettyWood Are all the other libraries you refused also much faster than Pydantic?

If that's the case, imho you should remove the misleading benchmarks from your documentation, because reading it gives the wrong impression that Pydantic is the fastest of the bunch.

That false impression is further induced with the phrase "Feel free to suggest more packages to benchmark or improve an existing one." which is likely to lead one to believe that no faster libraries have been suggested.

wyfo:
> @PrettyWood Are all the other libraries you refused also much faster than Pydantic?

A lot of existing libraries are a lot faster than Pydantic, which is especially slow in serialization. [apischema](https://github.com/wyfo/apischema) is more than [5x faster (up to 40x in serialization)](https://wyfo.github.io/apischema/0.17/optimizations_and_benchmark/), and it's not present in Pydantic benchmark, as well as [mashumaro](https://github.com/Fatal1ty/mashumaro) or [typical](https://github.com/seandstewart/typical). cattrs, which is a lot faster too is present, but its benchmark is completely biased by the `datetime` parsing, see https://github.com/samuelcolvin/pydantic/pull/1568.

Also, most of the libraries present in the benchmark doesn't even use type annotations, whereas apischema, typical, mashumaro and runtype do. 

Last commit of valideer is 3 years ago, trafaret 1 year ago; mashumaro have more download/month than valideer + trafaret + schematics. Omegaconf, which was rejected [last year](https://github.com/samuelcolvin/pydantic/issues/1939) has now 1k stars and 1M download/month too.

In my opinion, Pydantic benchmark is no more relevant at all, and I agree with @erezsh, it gives the wrong impression than Pydantic is ahead of other libraries in term of performance, while it clearly doesn't. Even marshmallow can perform better for serialization.

> Sorry but for now we can't accept this since we already refused a lot of other libraries for benchmark

@PrettyWood Then valideer should be removed now. 
Regarding download/month, omegaconf and mashumaro are relevant to be added. 
Apischema has also more download/month than valideer, and regarding features, it's one of the few that provides for example JSON schema generation, and can be considered as an alternative to pydantic in this domain.

*Disclaimer: I'm the author of apischema, my opinion can be partial.*

samuelcolvin:
Our benchmarks only include other libraries that are more popular than pydantic, or similarly popular. I don't really want to change that.

The other problem is that there's a lot of different things that could be included in the benchmarks.

Honestly I'm tempted to just remove the benchmarks.

wyfo:
> Our benchmarks only include other libraries that are more popular than pydantic, or similarly popular. I don't really want to change that.

I've said [before](https://github.com/samuelcolvin/pydantic/pull/1525#issuecomment-630435646) that I kind of agree with that. 
The issue is that today, Pydantic **is** IMO the most popular library, especially since the advent of FastAPI. Pydantic also was a lot [focused on performance](https://github.com/samuelcolvin/pydantic/issues/1157#issuecomment-572694302), claiming its lead with its benchmark. It has thus become a comparison point for all libraries in this domain (it was mine too), especially regarding performance.
And the fact is that today, Pydantic is surpassed (in term of performance at least, but not only) by a bunch of younger libraries.

However, the benchmark page is still here, and seems to be quite out of date. For example I pointed last year that valideer was no more maintained, they answer me that it had still 13k download/month. apischema has reach this number now (yes, it's still quite a few), and his more similar to Pydantic than any other libraries of the benchmark, as well as being the fastest available.

> Honestly I'm tempted to just remove the benchmarks.

And I think you should. Performance can barely be an argument for Pydantic anymore (especially regarding serialization).  When you know it, this benchmark seems more marketing than anything, and it provokes all the PRs you have about it.
Notably because of the formulation "Feel free to suggest more packages to benchmark or improve an existing one." I know you already had reword *sumbit* to *suggest*, but as @erezsh noticed, it's kind of misleading.

I can also add the sentence "Below are the results of crude benchmarks comparing pydantic to other validation libraries." where the word _popular_ is missing, but that's nitpicking, because I think updating the page will not solve the issue of performance as an argument being out of date.

> The other problem is that there's a lot of different things that could be included in the benchmarks.

Your library has become a reference, that's quite rewarding! But yes, success has a price, and these PRs are part of it.

erezsh:
@samuelcolvin 

> Our benchmarks only include other libraries that are more popular than pydantic, or similarly popular. I don't really want to change that.

Just out of curiosity, why is that? I understand you can't list every little package, but certainly there are candidates who are worth knowing as alternatives to pydantic. (I'm not even talking about mine).

Why not be supportive of the open-source community, and give those challengers another way to grow?

> Honestly I'm tempted to just remove the benchmarks.

I agree with @wyfo, that's probably for the best. Or at least put a big and clear disclaimer, explaining that the page is outdated, and that several faster alternatives to pydantic exist.

PR #3304, opened 2021-10-08
fix(secret): uppercase in filename on linux

### LINKED ISSUE #3273: Pydantic fails to read content of files in secrets_dir on Linux
### Checks

# Bug

To reproduce, on Linux OS: 
- build a config manager that subclasses `BaseSettings`
- add `secrets_dir` config pointing to dir with text file(s), as per docs
- add class attributes to use the files in `secrets_dir`, as per docs
- run the code and inspect/print the attributes
- observe that vars coming from .env file work fine, but those from `secrets_dir` are None, or error if not Optional, and there is no user warning for the `secrets_dir` itself

Repeat the same on MacOS - it should work OK, the content is read correctly.

I have added a repo with Dockerfile replicating the setup: https://github.com/dimitri-b/pydantic-secrets-dir-issue

```
             pydantic version: 1.8.2
            pydantic compiled: True
               python version: 3.8.12 (default, Sep  3 2021, 02:35:42)  [GCC 8.3.0]
                     platform: Linux-5.10.47-linuxkit-x86_64-with-glibc2.2.5
```

```py
import pydantic

class SettingsLocal(pydantic.BaseSettings):
    NORMAL_VAR: Optional[str]
    SECRET_VAR: Optional[str]

    class Config:
        env_file = "./.env"
        secrets_dir = "./secrets-in-project"
...
```

JeanArhancet:
## Change Summary

Hi,

It's my first Pull request :)
I worked on the problem of path and secret on Linux

## Related issue number

fix #3273

## Checklist


JeanArhancet:
please review

pydantic/env_settings.py
```diff
@@ -244,3 +240,16 @@ def read_env_file(
         return {k.lower(): v for k, v in file_vars.items()}
     else:
         return file_vars
+
+
+def read_secret_path(secrets_path: Path, env_name: str, case_sensitive: bool = False) -> Optional[str]:
+    paths = [secrets_path / env_name] if case_sensitive else [secrets_path / env_name, secrets_path / env_name.upper()]
```
samuelcolvin, line 246: This doesn't work if the filename is partially uppercase, e.g. `SECRET_var`.

To fix this, I think you'll need a utility function like 

```py
def find_case_insensitive_path(path: Path) -> Path:
    """
    Find a file within path's directory with a case-insensitive matching filename.
    
    Note: this won't if the rest of the path has a case mismatch.
    """
    expected_name = path.name.lower()
    for f in path.parent.iterdir():
        if f.name.lower() == expected_name:
            return f
    return path
```

Then you can reverse your changes in `__call__` above, and just add

```py
path = secrets_path / env_name
if settings.__config__.case_sensitive not path.exists():
    path = find_case_insensitive_path(path)
```

tests/test_settings.py
```diff
@@ -816,6 +817,30 @@ class Config:
     assert Settings().dict() == {'foo': 'foo_secret_value_str'}
 
 
+@pytest.mark.skipif(sys.platform != 'linux', reason='Linux tests')
```
samuelcolvin, line 820: do we actually need this? Does it fail on mac or windows?

tests/test_settings.py
```diff
@@ -877,7 +902,9 @@ class Settings(BaseSettings):
         class Config:
             secrets_dir = tmp_path
 
-    with pytest.warns(UserWarning, match=f'attempted to load secret file "{tmp_path}/foo" but found a directory inste'):
+    with pytest.warns(
```
samuelcolvin, line 905: not sure this should have changed?

tests/test_settings.py
```diff
@@ -816,6 +817,30 @@ class Config:
     assert Settings().dict() == {'foo': 'foo_secret_value_str'}
 
 
+@pytest.mark.skipif(sys.platform != 'linux', reason='Linux tests')
+def test_secrets_uppercase_filename(tmp_path):
+    (tmp_path / 'SECRET_VAR').write_text('foo_env_value_str')
+
+    class Settings(BaseSettings):
+        SECRET_VAR: str
```
samuelcolvin, line 825: This test doesn't have `case_sensitive = True` - therefore `secret_var` shouldn't be populated.

You'll need two tests one with `case_sensitive = True` and one without showing the functionality is not correct.

changes/3273-JeanArhancet.md
```diff
@@ -0,0 +1 @@
+Fix Pydantic error to read content of files in secrets_dir on Linux
```
samuelcolvin, line 1:
```suggestion
Update `SecretsSettingsSource` to respect `config.case_sensitive`
```
We know it's an error with pydantic, this is pydantic  :smile:

samuelcolvin (commented):
> It's my first Pull request :)

Congratulations :tada:, and well done.

This is a good start, but some things to fix.

By the way, I also altered the body of your pull request to add "fix " before the issue ID, so the issue is automatically closed when the PR is merged.

Thanks for your contribution.

samuelcolvin:
please update.

Also, you have conflicts to resolve.

tests/test_settings.py
JeanArhancet: Yes, we need this because the issue exist only on Linux.

File names with capital letters are returned on macos and windows

pydantic/env_settings.py
JeanArhancet: Thanks for the feedback :) 

But, we have two scenarios:

- Case sensitive and the filename which has the same syntax
- Case insensitive and read files with linux (issues #3304)

tests/test_settings.py
JeanArhancet: I have removed the skip

JeanArhancet:
please review

pydantic/env_settings.py
```diff
@@ -249,3 +251,19 @@ def read_env_file(
         return {k.lower(): v for k, v in file_vars.items()}
     else:
         return file_vars
+
+
+def find_case_path(path: Path, case_sensitive: bool) -> Optional[Path]:
+    """
+    Find a file within path's directory matching filename.
+
+    If filename is case-sensitive returns the file which has the same syntax.
```
samuelcolvin, line 260:
```suggestion
    If filename is case-sensitive returns the file which has the same exact name.
```

pydantic/env_settings.py
samuelcolvin: not sure what you mean by "same syntax"?

samuelcolvin:
Tests are failing, but I'm also a bit confused by what you mean by "... path with the same syntax" - what do you mean by "syntax" in this context?

Please update.

pydantic/env_settings.py
JeanArhancet: Same syntax for the name of files (for example when you have a file in upper case if you are case sensitive you should not read it if the variable name is in lower case)

JeanArhancet:
please review

samuelcolvin:
thanks a lot.

PR #3344, opened 2021-10-22
Add `NanoTime` data type

### LINKED ISSUE #432: Richer timedelta / timestamp parsers?
I was wondering whether it would make sense to steal a parser bit from pandas, allowing for much richer timedelta and timestamp parsing? (e.g., see [here](https://github.com/pandas-dev/pandas/blob/58a4151c0a7909ea20191ce11b8eebe08b6f2af0/pandas/_libs/tslibs/timedeltas.pyx#L323) for timedeltas)

This way, you could have `'3 micros', '1h2m'` etc as valid timedeltas which is arguably much nicer.

kirill-shtrykov:
## Change Summary

Add `NanoTime` data type

## Related issue number

fix #432

## Checklist


kkirsche:
Would it make sense to use a generic Time[T] type instead of the NanoTime type, where T is the time type (nano, milli, sec, min, etc.)? It looks like the class implementation is aware of the various time units, but this specifically implements Nano.

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
                 raise errors.DateNotInTheFutureError()
 
             return value
+
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ NANO TIME TYPE ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TIME_UNITS = {
+    'ns': 1,
```
samuelcolvin, line 1064:
```suggestion
    'ns': 1,
    'μs': 10 ** 3,
```
?

tests/test_types.py
```diff
@@ -2915,3 +2916,61 @@ class Model(BaseModel):
...
+    assert m.time.human_readable() == human
+
+
+def test_nanotime_to():
+    class Model(BaseModel):
+        time: NanoTime
+
+    m = Model(time='2d')
+
+    assert m.time.to('d') == pytest.approx(2)
```
samuelcolvin, line 2954: `pytest.mark.parametrize` again.

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+
+    @classmethod
+    def validate(cls, v: StrIntFloat) -> 'NanoTime':
+
+        result: int = 0
+        pos: int = 0
+        try:
+            return cls(int(v))
+        except ValueError:
+            pass
```
samuelcolvin, line 1087:
```suggestion
        try:
            v_int = int(v)
        except ValueError:
            pass
        else:
            return cls(v_int)
```

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+class NanoTime(int):
+    @classmethod
+    def __get_validators__(cls) -> 'CallableGenerator':
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v: StrIntFloat) -> 'NanoTime':
+
+        result: int = 0
+        pos: int = 0
```
samuelcolvin, line 1083: we don't these until after the first check, move them down.

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+        try:
+            return cls(int(v))
+        except ValueError:
+            pass
+
+        while pos < len(str(v)):
+            match = time_string_re.match(str(v), pos=pos)
```
samuelcolvin, line 1090:
```suggestion
        v_str = str_validator(v)
        while pos < len(v_str):
            match = time_string_re.match(v_str, pos=pos)
```

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+            if match is None:
+                raise errors.InvalidNanoTime()
+
+            scalar, unit = match.groups()
+            if unit is None:
+                if match.end() == len(str(v)):
```
samuelcolvin, line 1096:
```suggestion
                if match.end() == len(v_str):
```

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+
+TIME_UNITS = {
+    'ns': 1,
+    'ms': 10 ** 6,
+    's': 10 ** 9,
+    'm': 60 * 10 ** 9,
+    'h': 60 * 60 * 10 ** 9,
+    'd': 24 * 60 * 60 * 10 ** 9
+}
+time_string_re = re.compile(r'\s*(\d+\.?\d*)\s*([a-zA-Z]+)?', re.IGNORECASE)
```
samuelcolvin, line 1071:
```suggestion
time_string_re = re.compile(r'\s*(\d+(?:\.\d+)?)\s*([a-z]{1,2})?', re.IGNORECASE)
```

Unless you think `1. ns` should be allowed?

samuelcolvin: Also can you use `re.match` in the function instead of `re.compile` here, see #1132

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+                    unit = 'ns'
+                else:
+                    raise errors.OmittedNanoTimeUnit()
+
+            try:
+                unit_mult = TIME_UNITS[unit.lower()]
+            except KeyError:
+                raise errors.InvalidNanoTimeUnit(unit=unit)
+
+            result += int(float(scalar) * unit_mult)
```
samuelcolvin, line 1106: why do we need `float()` here?

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+            pos = match.end()
+
+        return cls(result)
+
+    def human_readable(self) -> str:
+
+        result: list = []
```
samuelcolvin, line 1113:
```suggestion
        result: List[str] = []
```

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+time_string_re = re.compile(r'\s*(\d+\.?\d*)\s*([a-zA-Z]+)?', re.IGNORECASE)
+
+
+class NanoTime(int):
+    @classmethod
+    def __get_validators__(cls) -> 'CallableGenerator':
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v: StrIntFloat) -> 'NanoTime':
```
samuelcolvin, line 1080: I can imagine people might well want to use the string parsing part of this function for `timedelta` parsing, for now could we just extract it into a standalone method so it can be called from validators?

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ NANO TIME TYPE ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TIME_UNITS = {
+    'ns': 1,
+    'ms': 10 ** 6,
+    's': 10 ** 9,
+    'm': 60 * 10 ** 9,
+    'h': 60 * 60 * 10 ** 9,
+    'd': 24 * 60 * 60 * 10 ** 9
```
samuelcolvin, line 1069: what about "seconds", "secs", "mins", "minutes", "hours", "hrs", "days", even "microseconds"?

Could we add something like
```py
UNIT_LOOKUP = {
    'min': 'm',
    'minute': 'm',
    ...
}

...
            unit_l = unit.lower()
            try:
                unit_mult = TIME_UNITS[unit_l]
            except KeyError:
                try:
                    unit_mult = TIME_UNITS[UNIT_LOOKUP[unit_l.rstrip('s')]
                except KeyError:
                    raise errors.InvalidNanoTimeUnit(unit=unit)
```

or something?

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
...
+        result: list = []
+        for unit in reversed(TIME_UNITS):
+            quotient = int(self / TIME_UNITS[unit])
+            if quotient > 0:
+                result.append(f'{quotient}{unit}')
+                self -= quotient * TIME_UNITS[unit]
+
+        return ' '.join(result)
+
+    def to(self, unit: str) -> float:
```
samuelcolvin, line 1122:
```suggestion
    def to(self, unit: Literal[...]) -> float:
```

samuelcolvin: in fact if you have this type, you can use it above in `TIME_UNITS` and `UNIT_LOOKUP`

docs/usage/types.md
```diff
@@ -183,6 +183,9 @@ with custom properties and validation.
 `ByteSize`
 : converts a bytes string with units to bytes
 
+`NanoTime`
+: converts a time string with units to nanoseconds
```
samuelcolvin, line 187:
```suggestion
: converts a time string with units to nanoseconds, `NanoTime` is a subclass of `int`
```

pydantic/types.py
```diff
@@ -1055,3 +1056,74 @@ def validate(cls, value: date) -> date:
                 raise errors.DateNotInTheFutureError()
 
             return value
+
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ NANO TIME TYPE ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
samuelcolvin, line 1061: can this all move into `datetime_parse.py`?

samuelcolvin (commented):
This looks like it could be really useful, quite a few things to change to make it even better :-)

samuelcolvin:
please update.

samuelcolvin:
One more question, instead of the loop, can't we do the whole parsing thing with a regex like `standard_duration_re` and `iso8601_duration_re`?

samuelcolvin:
@kirill-shtrykov  are you still working on this?

samuelcolvin:
On reflection, I'm closing this I think it's confusing two things - more permission parsing and supporting nanoseconds.

I think any further parsing leniency should be either:
* additions to [speedate](https://github.com/samuelcolvin/speedate)
* or a new type which allows more "human timestamps and diffs"

The matter of supporting nanoseconds is more complicated, if someone really wants it, we should discuss it in a new issue.

PR #3346, opened 2021-10-22
feat(timedelta): add scientific notation

### LINKED ISSUE #3315: Serialization -> de-serialisation fails for small timedelta (< 100 microseconds)
### Checks

# Bug

The serialization -> de-serialization of a model with small `timedelta` raises a `ValidationError`. The de-serialization fails only when the `timedelta` is below 100 microseconds, see the following example:

```py
from datetime import timedelta

class Model(BaseModel):
    duration: timedelta


# This works
model = Model(duration=timedelta(microseconds=100))
Model.parse_raw(model.json())
# This Fails
model = Model(duration=timedelta(microseconds=99))
Model.parse_raw(model.json())
```

Last line throws the following error:
```py
duration
  invalid duration format (type=value_error.duration)
```

I believe the error comes from the `parse_duration` function, and in particular the line where the input value is converted to `str`.
https://github.com/samuelcolvin/pydantic/blob/c256dccbb383a7fd462f62fcb5d55558eb3cb108/pydantic/datetime_parse.py#L226-L231
Indeed `str(0.0001)` gives `"0.0001"` but `str(0.000099)` gives `"9.9e-5"`, thus the `re.match` fails.

Changing `value = str(value)` to `value = "f{value:.6f}"` should fix this. I would be happy to create a PR to solve the issue.

# System information

```
             pydantic version: 1.8.2
            pydantic compiled: True
                 install path: <my-home>/.pyenv/versions/3.7.11/envs/pydantic/lib/python3.7/site-packages/pydantic
               python version: 3.7.11 (default, Aug 31 2021, 20:43:02)  [Clang 10.0.1 (clang-1001.0.46.4)]
```

JeanArhancet:
## Change Summary

There is an issue with scientific notations and timedelta

## Related issue number

fix #3315

## Checklist

### DIFF
--- a/changes/3315-JeanArhancet.md
+++ b/changes/3315-JeanArhancet.md
@@ -1,1 +1,1 @@
-Add scientific notation in timedelta
+Allow scientific notation for numbers in timedelta fields
--- a/changes/3315-JeanArhancet.md
+++ b/changes/3315-JeanArhancet.md
@@ -1,0 +1,1 @@
+Add scientific notation in timedelta
--- a/pydantic/datetime_parse.py
+++ b/pydantic/datetime_parse.py
@@ -40,6 +40,8 @@ standard_duration_re = re.compile(
     r'(?:\.(?P<microseconds>\d{1,6})\d{0,6})?'
     r'$'
 )
+# Support scientific notation for duration
+scientific_notation_duration_re = re.compile(r'^([+\-]?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+\-]?\d+))?$')
 
 # Support the sections of ISO 8601 date representation that are accepted by timedelta
 iso8601_duration_re = re.compile(
@@ -228,6 +230,8 @@ def parse_duration(value: StrBytesIntFloat) -> timedelta:
         value = value.decode()
 
     try:
+        if scientific_notation_duration_re.match(value):
+            value = f"{float(value):.6f}"
         match = standard_duration_re.match(value) or iso8601_duration_re.match(value)
     except TypeError:
         raise TypeError('invalid type; expected timedelta, string, bytes, int or float')
--- a/pydantic/datetime_parse.py
+++ b/pydantic/datetime_parse.py
@@ -55,7 +55,7 @@ iso8601_duration_re = re.compile(
 )
 
 # Support scientific notation
-scientific_notation = re.compile(r'^(?P<sign>[-+]?)' r'(?P<scientific_notation>\d+(.\d+)?[eE][+\-]?\d+)?' r'$')
+scientific_notation = re.compile(r'^(?P<sign>[-+]?)(?P<scientific_notation>\d+(.\d+)?[eE][+\-]?\d+)?$')
 
 EPOCH = datetime(1970, 1, 1)
 # if greater than this, the number is in ms, if less than or equal it's in seconds
@@ -249,7 +249,8 @@ def parse_duration(value: StrBytesIntFloat) -> timedelta:
         kw['microseconds'] = '-' + kw['microseconds']
 
     if kw.get('scientific_notation'):
-        kw['seconds'] = ("%.17s" % kw.pop('scientific_notation')).rstrip('0').rstrip('.')
+        # convert scientific notation to decimals
+        kw['seconds'] = '{:.8s}'.format(kw.pop('scientific_notation'))
 
     kw_ = {k: float(v) for k, v in kw.items() if v is not None}
 
--- a/pydantic/datetime_parse.py
+++ b/pydantic/datetime_parse.py
@@ -40,8 +40,6 @@ standard_duration_re = re.compile(
     r'(?:\.(?P<microseconds>\d{1,6})\d{0,6})?'
     r'$'
 )
-# Support scientific notation for duration
-scientific_notation_duration_re = re.compile(r'^([+\-]?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+\-]?\d+))?$')
 
 # Support the sections of ISO 8601 date representation that are accepted by timedelta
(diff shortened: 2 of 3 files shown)


pydantic/datetime_parse.py
```diff
@@ -228,6 +230,8 @@ def parse_duration(value: StrBytesIntFloat) -> timedelta:
         value = value.decode()
 
     try:
+        if scientific_notation_duration_re.match(value):
```
ehiggs, line 233: This looks like it's being run on every attempt to parse a timedelta. Seems like a chunk regex to run on everything just for the corner case. And we pass the value to the float constructor.

Wouldn't it be easier to do `try .. float(...) Except ValueError: pass` to skip the regex algoether?

JeanArhancet: The regex allows to make a first control and to avoid exceptions every time

ehiggs: With the regex:
```
In [1]: from datetime import timedelta
In [2]: from pydantic import BaseModel
In [3]: class Foo(BaseModel):    x: timedelta
In [4]: %timeit Foo.parse_raw('{"x": "3e4"}')
15.4 µs ± 165 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Using just `float(value)` with a try-except block:

```
In [1]: from datetime import timedelta
In [2]: from pydantic import BaseModel
In [3]: class Foo(BaseModel):    x: timedelta
In [4]: %timeit Foo.parse_raw('{"x": "3e4"}')
14.7 µs ± 199 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

There's almost nothing between them so should be fine imo. :)

JeanArhancet: Thank you for your feedback, I have added a try/catch in `isinstance(value, (str))` in the case of `value` is a `Dict`

ehiggs (approved):
lgtm!

pydantic/datetime_parse.py
```diff
@@ -226,7 +226,14 @@ def parse_duration(value: StrBytesIntFloat) -> timedelta:
         value = str(value)
     elif isinstance(value, bytes):
         value = value.decode()
-
+    
+    if isinstance(value, (str)):
```
samuelcolvin, line 230: Instead of adding this into the logic for all cases, I think you can match all the cases described in your tests with another regex, which can be appended to the current regexes checked and therefore have no effect on performance if it's not called.

changes/3315-JeanArhancet.md
```diff
@@ -0,0 +1 @@
+Add scientific notation in timedelta
```
samuelcolvin, line 1:
```suggestion
Allow scientific notation for numbers in timedelta fields
```

samuelcolvin (commented):
good catch, but I think we can implement in a cleaner way.

samuelcolvin:
please update

pydantic/datetime_parse.py
JeanArhancet: Thanks for you feedback. 

The regex in the commit  https://github.com/samuelcolvin/pydantic/pull/3346/commits/88ffae7f8cf14bd33f3a22f57cc302c2a3021aaf  is good for you?

ehiggs: please review

JeanArhancet:
please review

pydantic/datetime_parse.py
```diff
@@ -54,6 +54,9 @@
     r'$'
 )
 
+# Support scientific notation
+scientific_notation = re.compile(r'^(?P<sign>[-+]?)' r'(?P<scientific_notation>\d+(.\d+)?[eE][+\-]?\d+)?' r'$')
```
samuelcolvin, line 58:
```suggestion
scientific_notation = re.compile(r'^(?P<sign>[-+]?)(?P<scientific_notation>\d+(.\d+)?[eE][+\-]?\d+)?$')
```

pydantic/datetime_parse.py
```diff
@@ -243,6 +248,9 @@ def parse_duration(value: StrBytesIntFloat) -> timedelta:
     if kw.get('seconds') and kw.get('microseconds') and kw['seconds'].startswith('-'):
         kw['microseconds'] = '-' + kw['microseconds']
 
+    if kw.get('scientific_notation'):
+        kw['seconds'] = ("%.17s" % kw.pop('scientific_notation')).rstrip('0').rstrip('.')
```
samuelcolvin, line 252:
```suggestion
        kw['seconds'] = ("%.17s" % kw.pop('scientific_notation')).rstrip('0').rstrip('.')
```
this needs a comment to explain what's going on, also no need to use percent subsitution, you can use an f-string or format.

Also, I'm not quite clear what you're trying to do, but I would guess you can do it with standard string formatting, no need for `rstrip`.

samuelcolvin (commented):
otherwlise I think looking good.

samuelcolvin:
please update.

pydantic/datetime_parse.py
JeanArhancet: I've changed, it's ok for you ?

JeanArhancet:
please review

JeanArhancet:
@samuelcolvin this pull request still valid?

samuelcolvin:
Yes, hopefully this can be included in v1.10.

No idea what's going to happen to this on V2, but I guess we cross that bridge when we get there.

samuelcolvin:
I'm really not sure about this, we don't support this in V2 and adding it to speedate wouldn't be trivial.

We can fix the original issue in #3315 by serialising timedeltas better.

samuelcolvin:
replaced by #4329 which i think is simpler.

Thanks for the contribution, it was useful in working towards a fix.

Issue #3337, opened 2021-10-20
PostgresDsn fails for multiple hosts


StupidHobbit:
Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
pydantic version: 1.8.2
pydantic compiled: False
install path: /home/stupidhobbit/arc/arcadia/contrib/python/pydantic/pydantic
python version: 3.9.7 (default, Oct 13 2021, 18:49:29)  [Clang 12.0.1 ]
platform: Linux-5.4.0-66-generic-x86_64-with-glibc2.27
optional deps. installed: ['typing-extensions']
```

```py
from pydantic import BaseSettings, PostgresDsn


class PGSettings(BaseSettings):
    dsn: PostgresDsn = 'postgres://user:pass@host1.db.net,host2.db.net:6432/db'
```
This code gives you this error:

```
pydantic.error_wrappers.ValidationError: 1 validation error for PGSettings
dsn
  URL host invalid (type=value_error.url.host)
```
But this is a valid dsn for postgres and it works just fine with psycopg2

PR #3349, opened 2021-10-24
fix(PostgresDsn): Add support for multiple hosts in PostgresDsn #3337

rglsk:
## Change Summary

Add support for multiple hosts validation in `PostgresDsn` class.

Postgres supports multiple hosts in the DSN like: `postgres://user:pass@host1.db.net:4321,host2.db.net:6432/app`.
The change keeps old PostgreDsn working for single host as it was before and introduces new attribute `hosts` to keep track of all added hosts with validation.
I have also separated some regex to be able to reuse them in the new matching regexes. 

## Related issue number

Fixes: #3337

## Checklist


pydantic/networks.py
```diff
@@ -75,28 +82,53 @@ class Parts(TypedDict, total=False):
...
         _url_regex_cache = re.compile(
-            r'(?:(?P<scheme>[a-z][a-z0-9+\-.]+)://)?'  # scheme https://tools.ietf.org/html/rfc3986#appendix-A
-            r'(?:(?P<user>[^\s:/]*)(?::(?P<password>[^\s/]*))?@)?'  # user info
+            rf'{_scheme_regex}{_user_info_regex}{_host_regex}{_path_regex}{_query_regex}{_fragment_regex}',
+            re.IGNORECASE,
+        )
+    return _url_regex_cache
+
+
+def postgres_url_regex() -> Pattern[str]:
```
samuelcolvin, line 115: please add a docstring to describe why this is different.

pydantic/networks.py
```diff
@@ -75,28 +82,53 @@ class Parts(TypedDict, total=False):
...
+    global _postgres_url_regex_cache
+    if _postgres_url_regex_cache is None:
+        _postgres_url_regex_cache = re.compile(
+            rf'{_scheme_regex}{_user_info_regex}'
             r'(?:'
-            r'(?P<ipv4>(?:\d{1,3}\.){3}\d{1,3})(?=$|[/:#?])|'  # ipv4
-            r'(?P<ipv6>\[[A-F0-9]*:[A-F0-9:]+\])(?=$|[/:#?])|'  # ipv6
-            r'(?P<domain>[^\s/:?#]+)'  # domain, validation occurs later
+            r'(?P<hosts>('  # hosts, validation occurs later
+            r'([(?:\d{1,3}\.){3}\d{1,3})(?=$|[/:#?])]|'  # ipv4
```
samuelcolvin, line 122: can we reuse these bits from `_host_regex`

pydantic/networks.py
```diff
@@ -351,6 +393,114 @@ class PostgresDsn(AnyUrl):
...
+        if cls.user_required and user is None:
+            raise errors.UrlUserInfoError()
+
+        return parts
+
+    @classmethod
+    def validate_host_parts(cls, parts: 'HostParts') -> 'HostParts':
+        """
+        A method used to validate parts of an URL.
+        Could be overridden to set default values for parts if missing
```
samuelcolvin, line 421: this docstring is wrong I think?

pydantic/networks.py
```diff
@@ -351,6 +393,114 @@ class PostgresDsn(AnyUrl):
     }
     user_required = True
 
+    __slots__ = ('hosts',)
```
samuelcolvin, line 396: are you sure this is right? Should we not do something like `AnyUrl.__slots__ + ('hosts',)`?

pydantic/networks.py
```diff
@@ -351,6 +393,114 @@ class PostgresDsn(AnyUrl):
...
+
+    def __init__(self, *args, hosts: Optional[List['HostParts']] = None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.hosts = hosts
+
+    @classmethod
```
samuelcolvin, line 402: surely we can make some fairly small changes to `AnyUrl` to reduce the amount we need to duplicate here?

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin:
please update, also conflicts.

pydantic/networks.py
rglsk: Great catch! Thanks!

pydantic/networks.py
rglsk: Makes sense, DRY rule 💪

rglsk:
Thank you for code review, please review

pydantic/networks.py
```diff
@@ -350,6 +397,98 @@ class PostgresDsn(AnyUrl):
...
     user_required = True
 
+    __slots__ = AnyUrl.__slots__ + ('hosts',)
+
+    def __init__(self, *args, hosts: Optional[List['HostParts']] = None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.hosts = hosts
+
+    @classmethod
+    def validate_host_parts(cls, parts: 'HostParts') -> 'HostParts':
```
samuelcolvin, line 407: move this to the parent, and call it from there and here to avoid duplication.

pydantic/networks.py
```diff
@@ -350,6 +397,98 @@ class PostgresDsn(AnyUrl):
...
+        if port is not None and int(port) > 65_535:
+            raise errors.UrlPortError()
+
+        return parts
+
+    @classmethod
+    def validate_multi_host(cls, hosts: List[str]) -> List['HostParts']:
+        hosts_parts: List['HostParts'] = []
+        for host in hosts:
+            hm = host_regex().match(host)
```
samuelcolvin, line 421: with this we're re-running the host regex. Instead of running this bit of the regex twice, surely have just `r'(?P<hosts>.+?)` in `_postgres_url_regex_cache` above, then do the validation here?

pydantic/networks.py
```diff
@@ -350,6 +397,98 @@ class PostgresDsn(AnyUrl):
...
+                    'rebuild': rebuild,
+                    'port': original_parts.get('port'),
+                },
+            )
+            host_parts = cls.validate_host_parts(host_parts)
+            hosts_parts.append(host_parts)
+        return hosts_parts
+
+    @classmethod
+    def validate(cls, value: Any, field: 'ModelField', config: 'BaseConfig') -> 'PostgresDsn':
```
samuelcolvin, line 439: There's still to much duplication here.

There are a number of ways to work around this while maintaining performance and not altering the external interface to `AnyUrl`.

This class should only override `validate_host`, or perhaps a new method `validate_netloc`.

pydantic/networks.py
```diff
@@ -350,6 +397,98 @@ class PostgresDsn(AnyUrl):
...
+            hm = host_regex().match(host)
+            original_parts = cast('Parts', hm.groupdict())
+            host, tld, host_type, rebuild = cls.validate_host(original_parts)
+            host_parts = cast(
+                'HostParts',
+                {
+                    'host': host,
+                    'host_type': host_type,
+                    'tld': tld,
+                    'rebuild': rebuild,
```
samuelcolvin, line 430: This is not correct. There's no point in using types if you're just going to ignore them.

samuelcolvin (commented):
There's still quite a lot to do here.

Also as per https://github.com/samuelcolvin/pydantic/issues/2117#issuecomment-726420310 mongo DSN's can also have multiple hosts, so we should have a subclass of `AnyUrl` called `MultiHostDsn` which is then inherited by `PostgresDsn`.

Let me know if this is getting too involved, and I'll take over.

samuelcolvin:
please update.

also, you need to run mypy before committing.

pydantic/networks.py
rglsk: Makes sense 💪

pydantic/networks.py
rglsk: This part is the most tricky.

It is hard to override it since we need to build a new objects with new field `hosts`. We could map `host` and other fields and join them by `,` but the problem would be that we will lost which port relies to which host ect.

I will think more what would be the best way to merge it.

@samuelcolvin Please let me know if you have already something in mind.

rglsk:
please review

samuelcolvin:
Thanks so much for your contribution @rglsk, I made a few tweaks to get things passing, this should now be merged.

PR #3375, opened 2021-10-31
Allow for passing keyword arguments to from_orm

dosisod:
## Change Summary

You can already use `Field(alias='whatever')` on the model to allow for
converting from ORM models that might be named differently, but this
means that the model knows about the existence of a database (or at
least it's columns), which seems like a leaky abstraction.

With this modification though, the database will be the one to handle
the naming conflict, which keeps the model more succinct.

## Related issue number

None that I can find

## Checklist


dosisod:
please review

jordi-reinsma:
This feature is VERY much appreciated. If my ORM model has A, B, C..., W and X fields but I want the resulting PydanticModel to have an extra Y and Z fields, I would've needed to:
- Do some Config.getter_dict trickeries if available
- `PydanticModel(A=orm.A, B=orm.B, ..., X=orm.X, y=y, z=z)`
- root_validator trickery if Y and Z fields are based on any of the other fields

Just this one feature and all of these are solved. In a possible v2 those other options like Field alias could be dropped as this solution is way cleaner and performant.

Have to point out though that the kwargs overrides the original ORM model fields, e.g.:

```
class UserORM:
  name: str
class User(BaseModel):
  name: str
  class Config:
    orm_mode=True
print(User.from_orm(UserORM("Daniel"), name="Cooler Daniel"))
# prints User(name=Cooler Daniel)
```

samuelcolvin:
LGTM, please can you update the docs including an example.

Please update.

dosisod:
@samuelcolvin Docs have been updated

jordi-reinsma:
@dosisod please resolve the now conflicted branch (:

dosisod:
ping @samuelcolvin

dosisod:
@samuelcolvin is there anything else needed for this PR to be merged?

axgb:
Also pinging for relevancy

samuelcolvin:
I'll try to merge for v1.10 soon.

hauh:
There is an issue with this approach when an ORM object has descriptors (`@property` decorated methods).
When you unpack an object `{**obj, **kwargs}`, each attribute gets accessed, even if it is not defined in pydantic model, therefore all property methods are called, which is not always appropriate and probably can lead to slower performance.
Example:
`
class Model(BaseModel):

    class Config:
        orm_mode = True

    a: int

class Test:

    a = 1

    @property
    def test(self):
        print('called')


Model.from_orm(Test(), extra=42)
`

samuelcolvin:
@hauh is correct, if we're to support this, we would need to modify `_decompose_class` to try kwargs on get before `getattr`.

Please update.

docs/examples/models_orm_mode_kwargs.py
```diff
@@ -0,0 +1,31 @@
+import typing
+
+from pydantic import BaseModel
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+
+
+class MyModel(BaseModel):
```
samuelcolvin, line 8: This is a pretty confusing example.

Better to do something like

```py
class MyModel(BaseModel):
    foo: str
    bar: int
    spam: bytes

    class Config:
        orm_mode = True


BaseModel = declarative_base()


class SQLModel(BaseModel):
    __tablename__ = 'my_table'
    id = sa.Column('id', sa.Integer, primary_key=True)
    foo = sa.Column('metadata', sa.String(32))
    bar = sa.Column('metadata', sa.Integer)

...

pydantic_model = MyModel.from_orm(sql_model, bar=123, spam=b'apple')
```
Or similar to show the behaviour.

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin:
please update.

pydantic/utils.py
```diff
@@ -413,19 +413,20 @@ class GetterDict(Representation):
     We can't inherit from Mapping[str, Any] because it upsets cython so we have to implement all methods ourselves.
     """
 
-    __slots__ = ('_obj',)
+    __slots__ = ('_obj', '_kwargs')
 
-    def __init__(self, obj: Any):
+    def __init__(self, obj: Any, **kwargs: Any):
```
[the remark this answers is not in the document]
dosisod: I adjusted this slightly and typed `kwargs` as an `Optional[Dict[str, Any]]` instead, since anyone who might've subclassed `GetterDict` would now be missing the new `kwargs` param (if they overrode the `__init__` method). I can change it back if need be.

tests/test_orm_mode.py
```diff
@@ -357,3 +357,47 @@ class Config:
...
+        def __init__(self, username: str, full_name: str) -> None:
+            self.username = username
+            self.full_name = full_name
+
+    user = UserCls('admin', 'john smith')
+
+    converted = User.from_orm(user, name=user.full_name)
+
+    assert user.username == converted.username
+    assert user.full_name == converted.name
```
samuelcolvin, line 383: This test is pretty confusing, better to take the logic from the docs example and strip out the sqlalchemy dependency.

samuelcolvin (commented):
Looking good, just a few small things to fix. Please update.

Please remember to use the manage phrase described in the PR template to assign us once this is ready to review, otherwise it'll be missed.

pydantic/utils.py
```diff
@@ -413,18 +413,29 @@ class GetterDict(Representation):
...
         self._obj = obj
+        self._kwargs = kwargs or None
 
     def __getitem__(self, key: str) -> Any:
         try:
+            if self._kwargs:
+                try:
+                    return self._kwargs[key]
+                except KeyError:
+                    pass
```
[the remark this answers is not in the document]
dosisod: If there are `kwargs`, but the key isn't contained, we have to fallback on the `getattr()` call. I updated it to the following:

```python
            if self._kwargs and key in self._kwargs:
                return self._kwargs[key]
            return getattr(self._obj, key)
```

The `try` block was one way, this is another way. Is there a better way you think?

dosisod:
please review.

I don't know why the doc building is failing now, worked last commit, and works on my local machine.

samuelcolvin:
The docs build is failing because #4339 was merged.

It'll be failing locally and the error message explains how to fix it, so should be easy to fix. 

Please update.

docs/usage/models.md
```diff
@@ -152,6 +152,14 @@ _(This script is complete, it should run "as is")_
     The example above works because aliases have priority over field names for
     field population. Accessing `SQLModel`'s `metadata` attribute would lead to a `ValidationError`.
 
+You can also achieve the same thing by passing keyword arguments to `from_orm`, instead of
+using Field aliases:
+
+```py
+{!.tmp_examples/models_orm_mode_kwargs.py!}
+```
+_(This script is complete, it should run "as is")_
```
Bobronium, line 161: @dosisod, this should fix docs build:

````suggestion
{!.tmp_examples/models_orm_mode_kwargs.md!}
````

You can test this locally as well, but you will need to pull changes from remote.

File: pydantic/main.py [hunk 6]
```diff
@@ -568,10 +568,10 @@ def parse_file(
         return cls.parse_obj(obj)
 
     @classmethod
-    def from_orm(cls: Type['Model'], obj: Any) -> 'Model':
+    def from_orm(cls: Type['Model'], obj: Any, **kwargs: Any) -> 'Model':
         if not cls.__config__.orm_mode:
             raise ConfigError('You must have the config attribute orm_mode=True to use from_orm')
-        obj = {ROOT_KEY: obj} if cls.__custom_root_type__ else cls._decompose_class(obj)
+        obj = {ROOT_KEY: obj} if cls.__custom_root_type__ else cls._decompose_class(obj, kwargs)
```
Bobronium, line 574: Seems like, `kwargs` will be silently ignored if `cls.__custom_root_type__` is true. I don't think it should be the case.

pydantic/main.py
```diff
@@ -698,10 +698,10 @@ def validate(cls: Type['Model'], value: Any) -> 'Model':
             return cls(**value_as_dict)
 
     @classmethod
-    def _decompose_class(cls: Type['Model'], obj: Any) -> GetterDict:
+    def _decompose_class(cls: Type['Model'], obj: Any, kwargs: Dict[str, Any]) -> GetterDict:
         if isinstance(obj, GetterDict):
             return obj
-        return cls.__config__.getter_dict(obj)
+        return cls.__config__.getter_dict(obj, kwargs)
```
Bobronium, line 704: This is pretty much a breaking change and should be reflected in docs and history.

pydantic/utils.py
```diff
@@ -413,18 +413,26 @@ class GetterDict(Representation):
...
 
-    def __init__(self, obj: Any):
+    def __init__(self, obj: Any, kwargs: Optional[Dict[str, Any]] = None):
         self._obj = obj
+        self._kwargs = kwargs or None
 
     def __getitem__(self, key: str) -> Any:
         try:
+            if self._kwargs and key in self._kwargs:
+                return self._kwargs[key]
```
Bobronium, line 425: IMO, this should be out of `try/except AttrubuteError` block.

And perhaps we should ask for forgiveness, not permission (like you've done it in `get` method)?

pydantic/utils.py
```diff
@@ -413,18 +413,26 @@ class GetterDict(Representation):
...
             return getattr(self._obj, key)
         except AttributeError as e:
             raise KeyError(key) from e
 
     def get(self, key: Any, default: Any = None) -> Any:
+        if self._kwargs:
```
Bobronium, line 431: IMO, The way of getting key out of kwargs should match one in `__getitem__`

Bobronium (changes requested):
@samuelcolvin, please beware that this PR introduces breaking changes

This was previously a valid usecase, but with these changes it will raise an error:
```py
def test_custom_getter_dict():
    class TestCls:
        x = 1
        y = 2

    def custom_getter_dict(obj):
        assert isinstance(obj, TestCls)
        return {'x': 42, 'y': 24}
```

samuelcolvin:
Thanks so much @Bobronium for looking into this.

I wonder if we should delay this feature to V2?

@Bobronium @dosisod @PrettyWood @hramezani what do you think?

Bobronium:
> I wonder if we should delay this feature to V2?

It's one option.

Another is checking if `getter_dict` has `kwargs` argument when validating a config:

```py
if "kwargs" not in inspect.signature(config.getter_dict).parameters:
    orig_getter_dict = config.getter_dict
    @functools.wraps(orig_getter_dict)
    def updated_getter_dict(obj, kwargs):
        return orig_getter_dict(obj)
    
    config.getter_dict = updated_getter_dict
```

Or perhaps, not changing the `GetterDict` in the first place and always just wrapp it to account for kwargs? 
So we'll have an obj behind two proxy layers: `kwargs_wrapper` -> `getter_dict` -> `obj`.

All this seems quite not right, but I don't quite have a perfect solution in mind.

samuelcolvin:
The best option long term is https://github.com/pydantic/pydantic-core/issues/223, I think that there's quite an elegant solution in rust and we don't have to worry about breaking changes.

My worry is that any solution for this in V1.10 is either going to be hacky or a break changing in some edge case, or both.

I'm leaning towards closing this and waiting for V2, although I feel pretty guilty for @dosisod who has worked on this a lot and responded to my feedback, they would understandably find it pretty annoying if this PR gets closed now. 🙏

PrettyWood:
> they would understandably find it pretty annoying if this PR gets closed now.

computed fields PR will probably be closed so trust me I know the feeling 😆 but this is part of the deal when working on OSS.

----------------------------

If there is a way to avoid any breaking change like @Bobronium I would go for it. But if there is a risk of breaking change I would defer this PR. I would love 1.10 to be a smooth and stable release before the big bang

dosisod:
@samuelcolvin I don't mind closing this PR, this feature seems to be introducing more friction then a feature of this size should introduce. And, if there are plans to handle this in v2, we might be better off waiting.

Thank you all for your feedback and support!

samuelcolvin:
Thanks so much for understanding, we'll do our best to make this happen in V2.

PR #3393, opened 2021-11-05
Dataclass extra

### LINKED ISSUE #986: Pydantic dataclasses do not respect the extra-fields in Config
# Question
Pydantic dataclasses do not seem to respect the configuration passed in the `@dataclass`-decorator. I am wondering if this is a bug or expected functionality? The [documentation](https://pydantic-docs.helpmanual.io/usage/dataclasses/) indicates that the `Config` is respected.

> pydantic.dataclasses.dataclass's arguments are the same as the standard decorator, except one extra keyword argument config which has the same meaning as Config.

In particular I would like to ignore extra fields when initialising the object. The documentation suggests that the default behaviour is `Extra.ignore`, but it does not seem to work. Making this explicit does not seem to have an effect either..

```python
>>> class C:
...     extra = Extra.ignore
...
>>> @dataclass(config=C)
... class Foo:
...     a: str
...     b: str
...
>>> Foo(**{"a": "bar", "b": "foo", "c": 1})
TypeError: __init__() got an unexpected keyword argument 'c'
```

Please complete:
* OS: **os x**
* Python version: 
```
>>> import sys; print(sys.version)
3.7.2 (default, Mar 22 2019, 11:08:57)
[Clang 10.0.0 (clang-1000.10.44.4)]
```
* Pydantic version: 
```
>>> import pydantic; print(pydantic.VERSION)
1.0
```

DetachHead:
## Change Summary

forked from @PrettyWood's change which allows using the `extra` config option for dataclasses.

see https://github.com/samuelcolvin/pydantic/issues/986#issuecomment-744093319

## Related issue number

fix #986

## Checklist

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Type, TypeVar, Union, overload
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Type, TypeVar, Union, get_type_hints, overload
 
 from .class_validators import gather_all_validators
 from .error_wrappers import ValidationError
@@ -190,7 +190,9 @@ def _process_class(
 
     if getattr(config, 'extra', Extra.forbid) is not Extra.forbid:
         def allow_extra_init(self, *args, **kwargs):
-            self.__original_init__(*args, **{k: v for k, v in kwargs.items() if k in self.__annotations__})
+            self.__original_init__(*args, **{
+                k: v for k, v in kwargs.items() if k in get_type_hints(type(self))
+            })
 
         cls.__original_init__ = cls.__init__
         cls.__init__ = allow_extra_init
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -4,7 +4,7 @@ from .class_validators import gather_all_validators
 from .error_wrappers import ValidationError
 from .errors import DataclassTypeError
 from .fields import Field, FieldInfo, Required, Undefined
-from .main import create_model, validate_model
+from .main import Extra, create_model, validate_model
 from .typing import resolve_annotations
 from .utils import ClassAttribute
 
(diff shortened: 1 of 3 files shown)


DetachHead:
please review

tests/test_dataclasses.py
```diff
@@ -945,3 +945,30 @@ def __new__(cls, *args, **kwargs):
...
+def test_allow_extra():
+    class C:
+        extra = Extra.ignore
+
+    @pydantic.dataclasses.dataclass(config=C)
+    class Foo:
+        a: str
+        b: str
+
+    Foo(**{"a": "bar", "b": "foo", "c": 1})
```
samuelcolvin, line 959: * single quotes
* `Foo(a='bar'...)` would be cleaner
* you need to add `assert not hasattr(foo, 'c')`
* you need to add another test for `Extra.allow`

pydantic/dataclasses.py
```diff
@@ -188,6 +188,15 @@ def _process_class(
         cls.__name__, __config__=config, __module__=_cls.__module__, __validators__=validators, **field_definitions
     )
 
+    if getattr(config, 'extra', Extra.forbid) is not Extra.forbid:
+        def allow_extra_init(self, *args, **kwargs):
+            self.__original_init__(*args, **{
+                k: v for k, v in kwargs.items() if k in get_type_hints(type(self))
+            })
+
+        cls.__original_init__ = cls.__init__
```
samuelcolvin, line 197: Looks like this is never called. We need a test for this.

pydantic/dataclasses.py
```diff
@@ -188,6 +188,15 @@ def _process_class(
...
+        cls.__init__ = allow_extra_init
```
samuelcolvin, line 198: @PrettyWood I remember we had lots of problems with overriding `__init__` at some points.

Do you think this will break anything?

What about conflicts with #2557, will it just be code conflicts, or does this approach need to be rethought?

samuelcolvin (commented):
I've reopened the original issue, no idea why I closed it, must have been a mistake.

I'm also a bit confused about why tests weren't run, hopefully if you commit again they'll be triggered.

samuelcolvin:
please update.

pydantic/dataclasses.py
PrettyWood: I'm sorry @DetachHead but I felt like it was way easier to add it in https://github.com/samuelcolvin/pydantic/pull/2557/commits/593c9bb4abba752af31b16ecd229ee645331fe66 rather than handling conflicts
@samuelcolvin IMO if the commit above is good we can close this PR

DetachHead: @PrettyWood looks good. though i can't find the branch that commit is on? is there a PR for it?

DetachHead:
closing since #2557 was merged

PR #3413, opened 2021-11-14
bug: Add annotation to ModelField for use in BaseModel.__signature__

adriangb:
Currently, `ModelField.outer_type_` is used for `Model.__signature__`.

This leads to signatures that differ from dataclasses or classes with a manual `__init__` of the same types.

```python
from inspect import signature
from typing import Optional

@dataclass
class Dataclass:
    foo: Optional[int] = None

class CustomClass:
    def __init__(self, foo: Optional[int] = None) -> None:
        ...

class PydanticModel(BaseModel):
    foo: Optional[int] = None


print(signature(Dataclass).parameters["foo"].annotation)  # Optional[int]
print(signature(CustomClass).parameters["foo"].annotation)  # Optional[int]
print(signature(PydanticModel).parameters["foo"].annotation)  # <class 'int'>
```

This PR fixes this by storing the annotation in `ModelField.annotation` and then using that for the signature instead of `outer_type_`.
I realize this is adding a public attribute to `ModelField`, and that it is not as simple as storing the `type_` passed into the constructor (I tried to handle the case where the type is inferred from the default value, I think successfully), so this may not be the best solution, but it is the best I could come up with today.


samuelcolvin (changes requested):
Seems good to me, but please add a change description.

I don't think this can go in v1.9.1 as it's not a bug fix for v1.9

samuelcolvin:
please update.

adriangb:
Hmm. While I am adding a public attribute, and that could be considered a feature, I am only doing it to fix what I would consider a bug (incorrect results from `inspect.signsture`). One of my main questions here was: can you think of a way of doing this that does not add a public attribute?

samuelcolvin:
I don't mind the public attribute, I don't see another (simply) way.

Which of my points were you questioning? The current description, out which release this goes in?

adriangb:
I was just questioning if this is a bug fix or a feature. I think the underlying problem is a bug, but if we're adding a public attribute it's more like a feature.

I'll add the change description.

changes/3413-fix-inspect-signature.md
```diff
@@ -0,0 +1 @@
+fix mangling of inspect.signature for BaseModel
```
samuelcolvin, line 1:
```suggestion
fix mangling of `inspect.signature` for `BaseModel`
```

samuelcolvin (commented):
Otherwise lgtm

adriangb:
@samuelcolvin I guess we didn't merge this 😅

samuelcolvin:
> @samuelcolvin I guess we didn't merge this sweat_smile

Got there in the end :sweat:.

Thank so much, as ever, _sorry for the delay_.

JacobHayes:
@samuelcolvin looks like this one won't auto-merge because of the [old "request changes"](https://github.com/samuelcolvin/pydantic/pull/3413#pullrequestreview-929568495) - I think it'll need an approval to override.

samuelcolvin (approved):

samuelcolvin:
Thanks @JacobHayes, missed that.

JacobHayes:
Yup, I missed the auto-merge announcement so excited to find it in the wild!

samuelcolvin:
Ye, it's been a god sent for this work.

PR #3430, opened 2021-11-19
Add mypy static checking for default and default_factory

### LINKED ISSUE #3429: Mypy plugin does not check the type of default and default_factory
### Checks

# Bug

```
pydantic version: 1.8.2
            pydantic compiled: False
                 install path: /Users/lucr/pydantic/pydantic
               python version: 3.9.5 (default, Nov  6 2021, 11:31:10)  [Clang 12.0.5 (clang-1205.0.22.11)]
                     platform: macOS-11.4-x86_64-i386-64bit
```

```py
import pydantic

class ExampleMode(BaseModel):
   a: int = Field("test")
   b: list = Field(default_factory=set)
   c: int = Field("test", default_factory=set)
...
```

The model above should raise three `mypy` errors, since:
1. `a` is defined as an `int`, but its default is a string
2. `b` is defined as a `list`, but the default factory is a `set`
3. `c` will create a runtime error, since both default and default factory are present

I propose (will create PR shortly to show the desired behavior) to enhance the `get_function_hook` of the `mypy` plugin to make `mypy` solve the three errors above, by changing the `return_type` of the function depending on the argument type.

klaa97:
## Change Summary

This PR enhances the `mypy` Pydantic plugin to dynamically change the return type of the `Field` function, based on the `default` and `default_factory` arguments.

Moreover, it adds an error in case the default and default_factory arguments are both specified.

The main goal is to enhance static checks, since they allow for a way more robust code and avoid runtime errors.

Suggestions on the `mypy` plugin APIs usage are welcome (documentation on that is not very big 😄) 

## Related issue number

Fixes #3429 

## Checklist


klaa97:
please review

tests/mypy/outputs/plugin-fail.txt
```diff
@@ -25,4 +26,7 @@
 197: note: Possible overload variant:
 197: note:     def dataclass(*, init: bool = ..., repr: bool = ..., eq: bool = ..., order: bool = ..., unsafe_hash: bool = ..., frozen: bool = ..., config: Optional[Type[Any]] = ...) -> Callable[[Type[Any]], Type[Dataclass]]
 197: note:     <1 more non-matching overload not shown>
-219: error: Property "y" defined in "FrozenModel" is read-only  [misc]
\ No newline at end of file
+219: error: Property "y" defined in "FrozenModel" is read-only  [misc]
+239: error: Incompatible types in assignment (expression has type "str", variable has type "int")  [assignment]
```
Bobronium, line 30: Looking at implementation, this message does make sense. However it may not be obvious, especially when you're using some custom functions (you don't see a return type just by glancing at the line). Can we make it more obvious by changing the message? Would it be possible to fail fast at [pydantic.mypy:160](https://github.com/samuelcolvin/pydantic/pull/3430/files#diff-469037bbe55bbf9aa359480a16040d368c676adad736e133fb07e5e20d6ac523R160) with some custom message like 
```py
f"Incompatible type in `default_factory` ({default_factory_type.fullname} has return type {default_factory_type.ret_type!r}, variable has type {...})"
```

Bobronium (changes requested):
Thank you, that's a great feature! I have some small concerns, but otherwise LGTM 🚀!

pydantic/mypy.py
```diff
@@ -110,6 +123,44 @@ def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
...
+        * Check whether the default and default_factory argument is specified.
+        * Output an error if both are specified.
+        * Retrieve the type of the argument which is specified, and use it as return type for the function.
+        """
+        default_any_type = ctx.default_return_type
+
+        default_args = ctx.args[0]
+        default_factory_args = ctx.args[1]
+
+        default_specified = bool(default_args)
```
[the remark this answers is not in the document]
klaa97: Thank you! You are completely right, fixed!

tests/mypy/outputs/plugin-fail.txt
```diff
@@ -8,6 +8,7 @@
 58: error: Invalid value for "Config.orm_mode"  [pydantic-config]
 63: error: Invalid value for "Config.orm_mode"  [pydantic-config]
 74: error: Incompatible types in assignment (expression has type "ellipsis", variable has type "int")  [assignment]
+83: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
```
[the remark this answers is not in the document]
klaa97: Thank you - I am not sure though that I correctly understood why this should not fail.

In both strict and non-strict mode, shouldn't a code like this fail?
```
x: int = None
```

That is basically what it is being done in that instance; we are assigning `None` to a variable that has `int` as a type. I admit I am very noob when it comes to strict/non-strict mypy differences, but I thought this kind of check should fail in both modes.

Following your comment https://github.com/samuelcolvin/pydantic/pull/3430#discussion_r762456160 I think it would be even better though to have a message like:
```Incompatible type in `default`(...) ```

It would make way more clear what is failing to the reader; to do so, my best bet is to use the `is_overlapping_type` API from `mypy`... I'll give it a try, thank you for the feedback!

File: tests/mypy/outputs/plugin-fail.txt (hunk 1 above)
Comment (klaa97) [reply tests/mypy/outputs/plugin-fail.txt:30]: Thank you and totally agree, will follow your suggestion using the `overlapping_type` API from mypy

File: tests/mypy/outputs/plugin-fail.txt (hunk 3 above)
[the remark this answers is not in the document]
Comment (Bobronium) [reply tests/mypy/outputs/plugin-fail.txt:11]: Perhaps, you can look at this example in docs: https://pydantic-docs.helpmanual.io/usage/models/#recursive-models

Previously, you could find same pattern in the first example: https://github.com/samuelcolvin/pydantic/issues/1072

You can think of declaring pydantic fields as declaring function arguments, where `a: int = None` is allowed.

However, it also worth looking on the reason why the original example was changed. Maybe pydantic is moving towards more conventional typing model, in which case such change in the plugin should be welcomed.

Nevertheless IMO it should be introduced in separate PR.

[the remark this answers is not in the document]
Comment (Bobronium) [reply tests/mypy/outputs/plugin-fail.txt:11]: Thinking about this more, I came to conclusion that `a: int = Field(None)` should produce the same result as `a: int = None` would. 

I didn't find such assignment in the test file and assumed it doesn't produce error. If it does, then field default should indeed raise error as well.

TL;DR: let's add `a: int = None` and match it's behavior with Field \o/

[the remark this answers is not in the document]
Comment (klaa97) [reply tests/mypy/outputs/plugin-fail.txt:11]: Thank you again for the feedback, super helpful!

I will add the example you specified then 🚀

File: tests/mypy/outputs/plugin-fail.txt (hunk 1 above)
Comment (klaa97) [reply tests/mypy/outputs/plugin-fail.txt:30]: I tried to implement your suggestion, but I find a hard roadblock due to the separation of parsing expression into types and the function/class plugin hook.

What I currently do is "fake" the return value of the `Field` function in the mypy function hook, so that the `mypy` plugin will autonomously raise errors for type mismatch.

What you suggest cannot be done in the function hook: I do not have any information there on the type of the variable I am assigning the function to, therefore I cannot compare the types and raise the error.

Similarly, I tried to move this check to the class hook (where I have information about the variable types); the main problem there is that I do not have info on the function `Field` and its argument types, since the type checking comes later during the function hook :( .

I agree that it would be better to be more specific, but nonetheless I think that it is an improvement to add this check for the `default_factory`, and the message is not unclear 🤔 

What do you think? In case, I can also revert the change on the `default_factory` and keep only the check for the `default` argument.

Comment (Bobronium) [reply tests/mypy/outputs/plugin-fail.txt:30]: Your current implementation is definitely a big improvement. If it's not trivial to customize the message, as you say, I'd stick with current method for now.

File: tests/mypy/outputs/plugin-fail.txt (hunk 3 above)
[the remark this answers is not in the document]
Comment (klaa97) [reply tests/mypy/outputs/plugin-fail.txt:11]: Added the test you suggest here: https://github.com/samuelcolvin/pydantic/pull/3430/commits/0380b43c7d2c87d6fde3379ade36cc68ba9563e3 Thank you!

File: pydantic/mypy.py [hunk 4]
```diff
@@ -608,6 +657,10 @@ def error_untyped_fields(api: SemanticAnalyzerPluginInterface, context: Context)
     api.fail('Untyped fields disallowed', context, code=ERROR_UNTYPED)
 
 
+def error_default_and_default_factory_specified(api: CheckerPluginInterface, context: Context) -> None:
```
samuelcolvin, line 660: is there a good reason to make this a standalone function?

tests/mypy/modules/plugin_fail.py
```diff
@@ -226,3 +226,22 @@ class Config:
 
 inheriting2 = InheritingModel2(x=1, y='c')
 inheriting2.y = 'd'
+
+
+class FieldDefaultTestingModel(BaseModel):
```
samuelcolvin, line 231: All the valid cases should go in `success.py`, only the errors should be in this file.

pydantic/mypy.py
```diff
@@ -110,6 +123,41 @@ def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
...
+
+        In particular:
+        * Check whether the default and default_factory argument is specified.
+        * Output an error if both are specified.
+        * Retrieve the type of the argument which is specified, and use it as return type for the function.
+        """
+        default_any_type = ctx.default_return_type
+
+        default_args = ctx.args[0]
+        default_factory_args = ctx.args[1]
```
samuelcolvin, line 138: Looks like this could be quite brittle, e.g. if we changed the order of arguments in `Field`.

Could you add something like
```py
assert ctx.callee_arg_names[0] == 'default', '"default" is no longer first argument in Field()'
assert ctx.callee_arg_names[1] == 'default_factory', '"default_factory" is no longer second argument in Field()'
```
or something

pydantic/mypy.py
```diff
@@ -110,6 +123,41 @@ def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
...
+        if default_args:
+            default_type = ctx.arg_types[0][0]
+            default_arg = default_args[0]
+
+            if isinstance(default_arg, EllipsisExpr):  # Fallback to default Any type if the field is required
+                default_type = default_any_type
+
+            return default_type
+
+        if default_factory_args:
```
samuelcolvin, line 153:
```suggestion
        elif default_factory_args:
```

pydantic/mypy.py
```diff
@@ -110,6 +123,41 @@ def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
...
+            return default_any_type
+
+        if default_args:
+            default_type = ctx.arg_types[0][0]
+            default_arg = default_args[0]
+
+            if isinstance(default_arg, EllipsisExpr):  # Fallback to default Any type if the field is required
+                default_type = default_any_type
+
+            return default_type
```
samuelcolvin, line 151:
```suggestion
            # Fallback to default Any type if the field is required
            if not isinstance(default_arg, EllipsisExpr):
                return default_type
```

tests/mypy/outputs/plugin-fail.txt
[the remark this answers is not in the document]
samuelcolvin: I agree with you both.

Pydantic will be changing the way it deals with optional in v2, see #2353.

But for now I agree with this.

samuelcolvin (commented):
LGTM, just a few changes.

samuelcolvin:
please update.

pydantic/mypy.py
klaa97: The main reason is that all the `api.fail` are wrapped into functions for clarity, so I thought of being consistent with the rest of the codebase. What do you think?

tests/mypy/modules/plugin_fail.py
klaa97: Fixed, moving the valid cases to `success.py`, thanks!

pydantic/mypy.py
klaa97: Added the asserts!

klaa97:
please review

pydantic/mypy.py
samuelcolvin: sounds good.

samuelcolvin (approved):
LGTM

samuelcolvin:
Thanks so much for this.

klaa97:
Thank you for pydantic! 

This PR is quite old and the `mypy` tests apparently changed in master, I will look into this later tonight and see if they are fixable 🙏

samuelcolvin:
Thanks @klaa97, if we need to remove some older mypy that's fine, but we'll need to be explicit about which versions we support and which we drop.

samuelcolvin:
please update (just adding this so I keep track of which PRs I need to review).

klaa97:
please review

I should have fixed the tests, now it works also for mypy 0.910 (which had a few different way of handling things 😞 ). It also works with mypy 0.971, which is not currently tested in the CI.

To check what is going to change validation-side in the plug-in, I think the best way is to check the [`plugin-fail.txt` changes](https://github.com/klaa97/pydantic/blob/fix/mypy-warn-default-type/tests/mypy/outputs/plugin-fail.txt). Thank you!

klaa97:
Fixing coverage...

samuelcolvin:
> It also works with mypy 0.971, which is not currently tested in the CI.

Is tested, it's run in the main test job of CI.

samuelcolvin:
thanks so much.

PR #3434, opened 2021-11-21
Add optional `field` argument to `__modify_schema__`

jasujm:
## Change Summary

Add optional `field` parameter to `model.__modify_schema__()` to opt in to receiving the `ModelField` object of the field. This makes it possible e.g. for the implementation of `__modify_schema__()` of generic type to know the runtime type of its subfields.

The concept is inspired by the optional parameters in validators, implemented in the `pydantic.class_validators` module. There are some differences:

* Only `field` is supported. `values` does not make sense in this context, since validation deals with concrete *object*, and this with the class itself. Model `config` might be supported, but is not included in this PR.
* `field` is `Optional[ModelField]` as opposed to `ModelField`, unlike validators. This is because the `schema` methods can be used on things that are not model fields, in which case `None` will be passed to as the `field` argument.
* Whereas `class_validators` does plenty of checks on the validator method signature before wrapping it, my implementation simply checks for the presence of `field`, and calls `__modify_schema__()` with or without the `field` parameter. This is because with validators there is separate prepare and execute steps, whereas `__modify_schema__()` is called directly when generating schema. If the signature doesn't match the expected, the error message and stack trace will be clear enough.

## Related issue number

There are multiple requests for this:
* https://github.com/samuelcolvin/pydantic/discussions/2902
* https://github.com/samuelcolvin/pydantic/discussions/3445
* My own [hrefs](https://github.com/jasujm/hrefs) project also uses generic types as model fields and needs to access the field spec to properly generate schema

## Checklist

* [X] Tests pass on CI and coverage remains at 100% (expected to pass since actually running on CI is gate keeped)

### DIFF
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -137,6 +137,21 @@ For versions of Python prior to 3.9, `typing_extensions.Annotated` can be used.
 Custom field types can customise the schema generated for them using the `__modify_schema__` class method;
 see [Custom Data Types](types.md#custom-data-types) for more details.
 
+`__modify_schema__` can also take a `field` argument which will have type `Optional[ModelField]`.
+*pydantic* will inspect the signature of `__modify_schema__` to determine whether the `field` argument should be
+included.
+
+```py
+{!.tmp_examples/schema_with_field.py!}
+```
+_(This script is complete, it should run "as is")_
+
+Outputs:
+
+```json
+{!.tmp_examples/schema_with_field.json!}
+```
+
 ## JSON Schema Types
 
 Types, custom field types, and constraints (like `max_length`) are mapped to the corresponding spec formats in the
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -89,6 +89,19 @@ TypeModelOrEnum = Union[Type['BaseModel'], Type[Enum]]
 TypeModelSet = Set[TypeModelOrEnum]
 
 
+def _apply_modify_schema(
+    modify_schema: Callable[..., None], field: Optional[ModelField], field_schema: Dict[str, Any]
+) -> None:
+    from inspect import signature
+
+    sig = signature(modify_schema)
+    args = set(sig.parameters.keys())
+    if 'field' in args or 'kwargs' in args:
+        modify_schema(field_schema, field=field)
+    else:
+        modify_schema(field_schema)
+
+
 def schema(
     models: Sequence[Union[Type['BaseModel'], Type['Dataclass']]],
     *,
@@ -303,7 +316,7 @@ def get_field_schema_validations(field: ModelField) -> Dict[str, Any]:
         f_schema.update(field.field_info.extra)
     modify_schema = getattr(field.outer_type_, '__modify_schema__', None)
     if modify_schema:
-        modify_schema(f_schema)
+        _apply_modify_schema(modify_schema, field, f_schema)
     return f_schema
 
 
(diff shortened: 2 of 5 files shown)


pydantic/schema.py
```diff
@@ -87,6 +87,16 @@
 TypeModelSet = Set[TypeModelOrEnum]
 
 
+def _apply_modify_schema(modify_schema: Callable, field: ModelField, schema: Dict[str, Any]):
```
jasujm, line 90: The `field` should be `Optional[ModelField]` and not `ModelField`, since `__modify_schema__()` is not _only_ called for fields. If it's called for entire model, there is no `ModelField` in this context.

jasujm: Fixed in newer revision :point_up:

jasujm:
please review

docs/usage/schema.md
```diff
@@ -131,6 +131,11 @@ For versions of Python prior to 3.9, `typing_extensions.Annotated` can be used.
 Custom field types can customise the schema generated for them using the `__modify_schema__` class method;
 see [Custom Data Types](types.md#custom-data-types) for more details.
 
+You can also add any of the following arguments to the signature to use them in the implementation:
+
+* `field`: the field whose schema is customized. Type of object is `pydantic.fields.ModelField`.
+* `**kwargs`: if provided, the above argument will be provided in they `kwargs` dictionary
```
samuelcolvin, line 137:
```suggestion
`__modify_schema__` can also take a `field` argument which will have type `Optional[ModelField]`, pydantic will inspect the signature of `__modify_schema__` to determine
whether the `field` argument should be included.
```
Or something. Explaining about `**kwargs` is overkill here I think.

Might also be worth adding an example.

tests/test_schema.py
```diff
@@ -2562,3 +2562,43 @@ def resolve(self) -> 'Model':  # noqa
         },
         '$ref': '#/definitions/Model',
     }
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 7), reason='schema generation for generic fields is not available in python < 3.7'
```
samuelcolvin, line 2568: The test doesn't need to use a generic, so can remove the skipif

tests/test_schema.py
```diff
@@ -2562,3 +2562,43 @@ def resolve(self) -> 'Model':  # noqa
...
+)
+def test_schema_for_generic_field_with_field_parameter():
+    T = TypeVar('T')
+
+    class GenModel(Generic[T]):
+        @classmethod
+        def __modify_schema__(cls, field_schema, field):
```
samuelcolvin, line 2575:
```suggestion
        def __modify_schema__(cls, field_schema, field):
            assert isinstance(field, ModelField)
```

tests/test_schema.py
samuelcolvin: If there's a good reason to add a test for a generic here too, you can do that as a separate test.

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin:
please update

docs/usage/schema.md
jasujm: Simplified the documentation as suggested, and added an example

tests/test_schema.py
jasujm: No there's no reason for it to be generic, so changed the test to work without them

jasujm:
Thank you for the suggestions. I updated and rebased the PR accordingly. please review

samuelcolvin:
thanks so much.

PR #3513, opened 2021-12-11
chore: add pre-commit

PrettyWood:
Just a suggestion because quite a lot of PRs have lint issues once pushed. So better run `make lint` automatically

Screen Shot 2021-12-11 at 8 15 03 PM

WDYT @samuelcolvin?


samuelcolvin:
LGTM. do you think we need any docs?

Otherwise happy for you to merge.

PrettyWood:
I don't think so. At worst I'll add some doc later

PR #2650, opened 2021-04-10
feat: make JSON serialization of referenced pydantic models possible

PrettyWood:
## Change Summary
Allow to customize serialization of fields with `BaseModel` type

```py
from typing import List, Optional

from pydantic import BaseModel


class Address(BaseModel):
    city: str
    country: str


class User(BaseModel):
    name: str
    address: Address
    friends: Optional[List['User']] = None

    class Config:
        json_encoders = {
            Address: lambda a: f'{a.city} ({a.country})',
            'User': lambda u: f'{u.name} in {u.address.city} '
                              f'({u.address.country[:2].upper()})',
        }


User.update_forward_refs()

wolfgang = User(
    name='Wolfgang',
    address=Address(city='Berlin', country='Deutschland'),
    friends=[
        User(name='Pierre', address=Address(city='Paris', country='France')),
        User(name='John', address=Address(city='London', country='UK')),
    ],
)
print(wolfgang.json())
# {"name": "Wolfgang", "address": "Berlin (Deutschland)", "friends": ["Pierre in Paris (FR)", "John in London (UK)"]}
```

## Related issue number
https://github.com/samuelcolvin/pydantic/discussions/2622

## Checklist

* [x] Unit tests for the changes exist
* [x] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable
* [x] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)

PR #3542, opened 2021-12-19
fix: keep old behaviour of `json()` by default

PrettyWood:
## Change Summary
fix regression on 1.9 introduced in #2650 (see https://github.com/samuelcolvin/pydantic/discussions/3538#discussioncomment-1843005)

## Checklist

* [ ] ~Documentation reflects the changes where applicable~
* [ ] ~`changes/<pull request or issue id>-<github username>.md` file added describing change~ (I didn't add a change file since it's due to a new feature of 1.9)
~(see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)~

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -454,7 +454,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         exclude_defaults: bool = False,
         exclude_none: bool = False,
         encoder: Optional[Callable[[Any], Any]] = None,
-        to_dict: bool = True,
+        models_as_dict: bool = True,
         **dumps_kwargs: Any,
     ) -> str:
         """
@@ -475,7 +475,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         # This allows users to write custom JSON encoders for given `BaseModel` classes.
         data = dict(
             self._iter(
-                to_dict=to_dict,
+                to_dict=models_as_dict,
                 by_alias=by_alias,
                 include=include,
                 exclude=exclude,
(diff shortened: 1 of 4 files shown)


PrettyWood:
please review

layday:
I'm not sure if this is a good workaround. IIUC the (broader) issue with #2650 is that the transformation is no longer recursive. This will manifest in other ways (and it already has: #3544 :) It'd probably make more sense to expose `to_dict` in `json`, but adding more options always comes at a cost.

layday:
If you don't want to delay v1.9 then reverting #2650 might be the best course of action for the time being?

PrettyWood:
Yeah I saw new issues pop up. I will take some time tomorrow to check properly. But if no easy workaround is possible, I'll revert #2650

pydantic/main.py
```diff
@@ -469,11 +470,12 @@ def json(
             exclude_unset = skip_defaults
         encoder = cast(Callable[[Any], Any], encoder or self.__json_encoder__)
 
-        # We don't directly call `self.dict()`, which does exactly the same thing but
-        # with `to_dict = True` because we want to keep raw `BaseModel` instances and not as `dict`.
+        # We don't directly call `self.dict()`, which is does exactly this with `to_dict=True`
```
christianbundy, line 473:
```suggestion
        # We don't directly call `self.dict()`, which does exactly this with `to_dict=True`
```

samuelcolvin:
@PrettyWood did you have time to think about this?

I've haven't dug into it in detail, but I will try to find time soon.

christianbundy:
FWIW after the force-push this seems to solve the problem (and adds a test from #3544). At least from my perspective, this looks Fine™️.

PrettyWood:
This should solve everything yes

samuelcolvin:
awesome, thank you.

PR #3572, opened 2021-12-24
Support and test mypy 0.910, 0.920, and 0.921

christianbundy:
## Change Summary

Problem: We aren't testing mypy version compatibility, which makes it hard to know when we're making/breaking compatibility. In my local testing, it seems that we've recently broken mypy 0.910 compatibility.

Solution: Add mypy version matrix to CI and fix compatibility regressions.

## Related issue number

Link: https://github.com/samuelcolvin/pydantic/discussions/3538#discussioncomment-1867912

See-also: #3175 

## Checklist

* [x] Unit tests for the changes exist
* [x] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable
* [x] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)
* [x] My PR is ready to review, **please add a comment including the phrase "please review" to assign reviewers**

PR #3594, opened 2021-12-29
Support mypy 0.910 to 0.930 including CI tests

samuelcolvin:
Some cleanup from #3573, also stealing logic from #3572.

I normally don't think changing the versions of mypy which the plugin is compatible with would be a problem in a minor release, but since we've already made two beta release of v1.9, breaking mypy compatibility now would be ugly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,8 +164,52 @@ jobs:
         name: coverage
         path: coverage
 
+  test-old-mypy:
+    name: test mypy v${{ matrix.mypy-version }}
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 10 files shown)


samuelcolvin:
Right now, pypi is partially down, hence all the failures, see [here](https://twitter.com/samuel_colvin/status/1476257837890387979).

samuelcolvin:
@christianbundy, @PrettyWood I'd love you feedback on this.

I still need to get tests passing once pypi is back up and add a change description.

christianbundy (approved):
Looks great to me.

samuelcolvin:
LGTM. I'll merge tomorrow.

PR #3616, opened 2022-01-03
Guard default object dict() from list of dictionaries

### LINKED ISSUE #3614: Object construction yields incorrect results when passed a list of dictionaries as a field initializer
### Checks

# Bug

```
pydantic version: 1.9.0\n            pydantic compiled: False\n                 install path: /Users/mkromer/.pyenv/versions/3.6.8/envs/tcex3-3.6.8/lib/python3.6/site-packages/pydantic\n               python version: 3.6.8 (default, Aug 25 2021, 09:55:39)  [GCC Apple LLVM 12.0.0 (clang-1200.0.32.29)]\n                     platform: Darwin-19.6.0-x86_64-i386-64bit\n     optional deps. installed: ['typing-extensions']
```

When initializing a union input with a list of dictionaries, the handler in validate() and parse_obj() will attempt to see if the input value can be converted to a dictionary with `dict()`.  However, if the input is a *list of dictionaries*, (at least, a list with an even number of keys in the first value's dictionary) dict() will succeed but will return a dictionary based on the pairwise key *names*, i.e. `dict([{'key_a': 'value_a', 'key_b': 'value_b'}])` will return `{ 'key_a': 'key_b' }` which is not the desired result.

Because of this, model inputs which accept `Union[Thing, List[Thing]]` will fail, but `Union[List[Thing], Thing]` will succeed.  It is highly desirable to have both formats work regardless of order.

```py
from typing import Any, Optional, Union, List

class KV(BaseModel):
    key: str
    type: Optional[str]
    value: Any


class MyModel(BaseModel):
    input_1: Union[KV, List[KV]]

data = [
    {"key": "domain", "value": "supercalifragilisticexpialidocious.com"},
    {"key": "specialsauce", "value": "another.com"},
]

mm = MyModel(input_1=data)
print(mm)

```

mkromer-tc:
## Change Summary

Guard use of `dict()` such that if the input to dict is a *list of dictionaries* a ValueError is raised, rather than allow
dict to construct a new dictionary based on the keys of element 0 which yields incorrect results.

## Related issue number

fix #3614 

## Checklist

### DIFF
--- a/changes/3614-mkromer-tc.md
+++ b/changes/3614-mkromer-tc.md
@@ -1,0 +1,11 @@
+# Fix 3614
+
+Insert `_guard_list_of_dicts()` as a guard point against trying
+to instantiate a dict() of a list of dictionaries, which will inevitably give
+incorrect results.
+
+Python will convert `dict([{'key_a': 'value_a', 'key_b': 'value_b'}])` to `{'key_a': 'key_b'}`
+but because the conversion succeeds (albeit with incorrect results) it is impossible to allow
+model fields of the type `Union[Thing, List[Thing]]` whereas `Union[List[Thing], Thing]` works as
+expected.  Raising an exception when trying to perform a `dict()` operation on a list of
+dictionaries allows both forms to work.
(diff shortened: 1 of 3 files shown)


mkromer-tc:
Please review

PR #3640, opened 2022-01-07
fix ability to mix/merge json env variables with env_nested_delimiter…

Air-Mark:
## Change Summary
fix ability to mix/merge json env variables with variables defined according to env_nested_delimiter

## Related issue number
Found a corner-case when we are mixing 2 types of nested environment variables definition.
#3159

## Checklist

### DIFF
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -1,5 +1,6 @@
 import os
 import warnings
+from json import JSONDecodeError
 from pathlib import Path
 from typing import AbstractSet, Any, Callable, ClassVar, Dict, List, Mapping, Optional, Tuple, Type, Union
 
@@ -183,7 +184,7 @@ class EnvSettingsSource:
             if is_complex:
                 if env_val is None:
                     # field is complex but no value found so far, try explode_env_vars
-                    env_val_built = self.explode_env_vars(field, env_vars)
+                    env_val_built = self.explode_env_vars(field, env_vars, settings)
                     if env_val_built:
                         d[field.alias] = env_val_built
                 else:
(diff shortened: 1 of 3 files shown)


Air-Mark:
please review

pydantic/env_settings.py
```diff
@@ -232,7 +235,16 @@ def explode_env_vars(self, field: ModelField, env_vars: Mapping[str, Optional[st
             env_var = result
             for key in keys:
                 env_var = env_var.setdefault(key, {})
-            env_var[last_key] = env_val
+
+            if env_val:
+                try:
+                    env_val = settings.__config__.json_loads(env_val)
+                except JSONDecodeError:
+                    ...
```
samuelcolvin, line 243:
```suggestion
                    pass
```

pydantic/env_settings.py
```diff
@@ -232,7 +235,16 @@ def explode_env_vars(self, field: ModelField, env_vars: Mapping[str, Optional[st
             env_var = result
             for key in keys:
                 env_var = env_var.setdefault(key, {})
-            env_var[last_key] = env_val
+
+            if env_val:
+                try:
+                    env_val = settings.__config__.json_loads(env_val)
+                except JSONDecodeError:
```
samuelcolvin, line 242:
```suggestion
                except ValueError:
```
More consistent to just use `ValueError`

pydantic/env_settings.py
```diff
@@ -232,7 +235,16 @@ def explode_env_vars(self, field: ModelField, env_vars: Mapping[str, Optional[st
             env_var = result
             for key in keys:
                 env_var = env_var.setdefault(key, {})
-            env_var[last_key] = env_val
+
+            if env_val:
+                try:
```
samuelcolvin, line 240: The problem here is we're calling `json.loads` on every field which:
1. could have some weird side effects - e.g. you want `"foobar"` but `json_loads` converts it to `foobar`
2. It could effect performance

We could:
* not do this and document it?
* check if `env_val` starts with `{` or `[` and only parse json in that case

samuelcolvin (commented):
I get the request, but I'm not sure about the best way to proceed.

samuelcolvin:
please update.

samuelcolvin:
I think this overlaps a lot with #4216, but that solution is more complete, even that we're deferring to V2.

PR #3652, opened 2022-01-10
Fix missing port in HttpUrl.build() result

dolfinus:
## Change Summary

Previously  passing the `port` argument into `HttpUrl.build()` method has lead to including this port number to the output:
```python
assert 'http://example.com:123' == HttpUrl.build(schema='http', host='example.com', port='123')
```

But a regression was introduced in #2447.

It looks like the desired behavior was to hide default port number for the protocol, but instead any port number was just being ignored.

So I've fixed a check, that only default port numbers become hidden, like `80` for `http` or `443` to `https`. Non-default port numbers are left unchanged.
Also I've added a test to cover this case.

## Related issue number

No

## Checklist


dolfinus:
please review

pydantic/networks.py
```diff
@@ -25,13 +25,14 @@
     no_type_check,
 )
 
+from typing_extensions import TypedDict
+
```
samuelcolvin, line 29: I don't think this is required.

tests/test_networks.py
```diff
@@ -564,6 +565,34 @@ def test_build_url(kwargs, expected):
     assert AnyUrl(None, **kwargs) == expected
 
 
+@pytest.mark.parametrize(
+    'kwargs,expected',
+    [
+        (dict(scheme='http', host='example.net'), 'http://example.net'),
+        (dict(scheme='https', host='example.net'), 'https://example.net'),
+        (dict(scheme='http', user='foo', host='example.net'), 'http://foo@example.net'),
+        (dict(scheme='https', user='foo', host='example.net'), 'https://foo@example.net'),
```
samuelcolvin, line 574: What about the cases of
```suggestion
        (dict(scheme='http', user='foo', host='example.net', port=80), 'http://foo@example.net'),
        (dict(scheme='https', user='foo', host='example.net', port=443), 'https://foo@example.net'),
```
?

tests/test_networks.py
```diff
@@ -564,6 +565,34 @@ def test_build_url(kwargs, expected):
...
+        (dict(scheme='https', host='example.net', port='1234'), 'https://example.net:1234'),
+    ],
+)
+@pytest.mark.parametrize(
+    'klass',
+    [
+        AnyHttpUrl,
+        HttpUrl,
+    ],
+)
```
samuelcolvin, line 591:
```suggestion
@pytest.mark.parametrize('klass', [AnyHttpUrl, HttpUrl])
```

pydantic/networks.py
```diff
@@ -176,6 +181,18 @@ def build(
         fragment: Optional[str] = None,
         **_kwargs: str,
     ) -> str:
+        parts = Parts(
```
samuelcolvin, line 184: not sure why this is required and/or part of this change?

samuelcolvin (commented):
otherwise I think makes sense.

samuelcolvin:
please update.

pydantic/networks.py
dolfinus: This is necessary because `get_default_parts` (there the default port value is checked) and `build` are both class methods, and they cannot share any information using `self` object. Also `get_default_args` does not support kwargs, only the `Parts` class object, so I need to create it manually.

tests/test_networks.py
dolfinus: AnyHttpUrl does not hide port number, only HttpUrl does. Moved suggested cases to new test

dolfinus:
Fixed, please check

changes/3652-dolfinus.md
```diff
@@ -0,0 +1 @@
+Port number is no longer being ignored by `HttpUrl.build()`
```
samuelcolvin, line 1: sorry for the delay, this description is a bit unclear - when reading it I'm not sure whether this is the "no longer being ignored" is the bug or the fix state.

Maybe you could tweak it to be clear about the new behaviour.

samuelcolvin:
thanks so much.

PR #3674, opened 2022-01-14
support `kw_only` on dataclasses

DetachHead:
## Related issue number

fix #3670 (it's a discussion tho because the issue template didn't let me raise a feature request)
## Checklist

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -31,6 +31,7 @@ This means we **don't want to create a new dataclass that inherits from it**
 The trick is to create a wrapper around `M` that will act as a proxy to trigger
 validation without altering default `M` behaviour.
 """
+import sys
 from contextlib import contextmanager
 from functools import wraps
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Dict, Generator, Optional, Type, TypeVar, Union, overload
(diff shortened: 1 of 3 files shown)


DetachHead:
please review

DetachHead:
what's up with the CI on this project? the MR says it's awaiting approval but it's already run? https://github.com/DetachHead/pydantic/actions/runs/1695980747

DetachHead:
how do i make `plugin-fail.txt` expect a different output depending on the python version?

pydantic/dataclasses.py
```diff
@@ -156,8 +158,14 @@ def _process_class(
...
-    cls: Type['Dataclass'] = dataclasses.dataclass(  # type: ignore
-        _cls, init=init, repr=repr, eq=eq, order=order, unsafe_hash=unsafe_hash, frozen=frozen
+    cls: Type['Dataclass'] = (
+        dataclasses.dataclass(  # type: ignore
+            _cls, init=init, repr=repr, eq=eq, order=order, unsafe_hash=unsafe_hash, frozen=frozen, kw_only=kw_only
+        )
+        if sys.version_info >= (3, 10)
+        else dataclasses.dataclass(  # type: ignore
+            _cls, init=init, repr=repr, eq=eq, order=order, unsafe_hash=unsafe_hash, frozen=frozen
+        )
```
samuelcolvin, line 168: This should be a full `if ...: else:...`, not a "one liner".

pydantic/dataclasses.py
```diff
@@ -199,60 +207,124 @@ def _process_class(
...
+            unsafe_hash: bool = False,
+            frozen: bool = False,
+            config: Type[Any] = None,
+            kw_only: bool = False,
+        ) -> Union[Callable[[Type[Any]], Type['Dataclass']], Type['Dataclass']]:
+            ...
+
+else:
+
+    def dataclass(
```
samuelcolvin, line 302: not sure this should be in `else:`, I think it should be run either with or without `TYPE_CHECKING`?

tests/test_dataclasses.py
```diff
@@ -989,3 +990,17 @@ def __new__(cls, *args, **kwargs):
         instance = cls(a=test_string)
         assert instance._special_property == 1
         assert instance.a == test_string
+
+
+if sys.version_info >= (3, 10):
```
samuelcolvin, line 995: use `skipIf`, rather than `if` like this.

samuelcolvin (commented):
> how do i make `plugin-fail.txt` expect a different output depending on the python version?

One solution (and it would make the overloads much cleaner below), is to always include `kw_only` in the type definition, then just raise a clear error for older versions?

Otherwise LGTM.

samuelcolvin:
please update.

pydantic/dataclasses.py
DetachHead: unfortunately i can't due to https://github.com/python/mypy/pull/1071. the workaround requires both sides of the `sys.version_info` check to have a "fake" implementation, so if the real implementation wasn't in the `else` mypy would complain that it's a duplicate definition.

i'll add some more comments to make this more clear

samuelcolvin: See my message above. Just use the same signature for all versions, then raise an error internally.

Should solve most of these problems.

DetachHead: but that way you only get a runtime error. imo it's better to allow type checkers to catch errors like these wherever possible.

if you are happy with this approach, i've made a change to the mypy tests to allow version specific output files which i believe solves that problem

duducp:
Is there any prediction for this feature?

KotlinIsland:
# 🔮
I predict that it will be... BASED!!!

samuelcolvin:
Should go into v1.10 which I'll start working on after we get v1.9.1 out.

pydantic/dataclasses.py
DetachHead: https://github.com/python/mypy/pull/10712 has now been merged which solves the issue

samuelcolvin (approved):

samuelcolvin:
Thanks so much.

Issue #2678, opened 2021-04-15
IMPORTANT: PEP 563, PEP 649 and the future of pydantic


samuelcolvin:
[PEP 563](https://www.python.org/dev/peps/pep-0563/) "Postponed Evaluation of Annotations" was introduced in python 3.7 behind the `from __future__ import annotations` switch, it basically meant all annotations are strings, not python objects.

Pydantic has tried to supported Postponed Annotations since `v0.18.0`, see #348.

The problem however is that trying to evaluate those strings to get the real annotation objects is really hard, perhaps impossible to always do correctly, see #248 #234 #397 #415 #486 #531 #545 #635 #655 #704 #1298 #1332 #1345 #1370 #1668 #1736 #1873 #1895 #2112 #2314 #2323 #2411 

The reasons are complicated but basically `typing.get_type_hints()` doesn't work all the time and neither do the numerous hacks we've introduced to try and get fix it. Even if `typing.get_type_hints()` was faultless, it would still be massively slower than the current semantics or PEP 649 (see below).

In short - pydantic doesn't work very well with postponed annotations, perhaps it never will.

# The Problem

The problem is that postponed annotations are set to be come default in 3.10, features of which will be frozen in about three week.

Even worse, there's no way to switch back to the current behaviour.

# The Solution

The solution to this is [PEP 649](https://www.python.org/dev/peps/pep-0649/) developed by Larry Hastings. This basically means annotations evaluation is lazy - the work of building `__annotations__` isn't done until you access `__annotations__`.

As far as I can tell this is the best of both worlds.

# The Second Problem

The sad reality however is that it seems very possible that PEP 649 will get rejected and when python 3.10 is released it will break much of pydantic, and thereby FastAPI and all the other libraries that rely on pydantic (as well as other libraries that use annotations at runtime like [enforce](https://github.com/RussBaz/enforce) and [typer](https://github.com/tiangolo/typer) presumably).

See [this very long discussion](https://mail.python.org/archives/list/python-dev@python.org/thread/QSASX6PZ3LIIFIANHQQFS752BJYFUFPY/#UITB2A657TAINAGWGRD6GCKWFC5PEBIZ) of the issue on **python-dev** about whether PEP 649 is a good idea.

---

# The Point

So why am I saying all this apart from whinging about something I don't agree with?

## 1. This is fair warning

That pydantic might break in a big way if python's core developers continue value principle over pragmatism.

## 2.  You can help

[...]

PR #3701, opened 2022-01-20
fix: support local refs natively in future annotation mode

henryiii:
## Change Summary

This is trying out the suggestion in #2678. When the class is instantiated, type annotations are resolved using the enclosing scopes, enabling locally defined types to be used.

Performance should be checked. This should only affect class creation, though - not usage. This could be optimized, perhaps, by trying to resolve first, maybe with the inmost locals, and then trying again with the full locals construction only if some annotations remain unresolved.

## Related issue number

Improves the situation in #2678 by supporting local type annotation, or annotations in an enclosing scope, but not global.

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -1,4 +1,5 @@
 import warnings
+import sys
 from abc import ABCMeta
 from copy import deepcopy
 from enum import Enum
@@ -289,7 +290,19 @@ class ModelMetaclass(ABCMeta):
         cls = super().__new__(mcs, name, bases, new_namespace, **kwargs)
         # set __signature__ attr only for model class, but not for its instances
         cls.__signature__ = ClassAttribute('__signature__', generate_model_signature(cls.__init__, fields, config))
-        cls.__try_update_forward_refs__()
+
+        localns = {}
+
+        # Pull in the currently defined local namespace and any namespaces above 
+        if hasattr(sys, "_getframe"):
+            frame = sys._getframe(1)
+            while frame is not None:
+                for k, v in frame.f_locals.items():
+                    if k not in localns:
+                        localns[k] = v
+                frame = frame.f_back
+
+        cls.__try_update_forward_refs__(**localns)
 
         return cls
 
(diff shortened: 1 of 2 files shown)


pydantic/main.py
```diff
@@ -289,7 +290,19 @@ def is_untouched(v: Any) -> bool:
...
+
+        # Pull in the currently defined local namespace and any namespaces above 
+        if hasattr(sys, "_getframe"):
+            frame = sys._getframe(1)
+            while frame is not None:
+                for k, v in frame.f_locals.items():
+                    if k not in localns:
+                        localns[k] = v
+                frame = frame.f_back
+
```
henryiii, line 304: This is the "all enclosing scopes version" - if you are only worried about adding local definitions, at least as first, this simpler version would suffice:

```suggestion

localns = sys._getframe(1).f_locals if hasattr(sys, "_getframe") else {}
```

Also, this doesn't handle classes defined directly within classes, which might have been important too? That could be added as well by including the class namespace (`new_namespace` above).

alexmojaki: `frame.f_back` goes to the calling function. This isn't 'enclosing' scopes, it's all scopes being called on the stack, which is very likely to cause problems as it may use completely unrelated names.

I gave a suggestion for one way to choose which frame to get locals from here: https://github.com/samuelcolvin/pydantic/issues/2678#issuecomment-822685329

> Also, this doesn't handle classes defined directly within classes

I'm not 100% sure what you have in mind, but inner classes don't have direct access to the enclosing class scope:

```python
class A:
    x = 1
    class B:
        print(x)  # NameError
```

henryiii: Yes, you are right, going "up" the stack would be too much. Getting the immediate locals, though, would help quite a bit with most common issues.

I was assuming the issue with typing in classes looked more like this:

```python
class A:
    class B: pass

    x: B
```

And you want to resolve `B` from `A`.

samuelcolvin:
thanks so much, Interesting idea.

I'll refer back to this when we have a decision on what's happening with python.

PR #3715, opened 2022-01-21
Add Robusta.dev to list of Pydantic users

aantn:
If possible, we would be honored to be listed on the homepage. We're really excited about how we're utilizing Pydantic to automate Kubernetes troubleshooting. Pydantic makes formerly tedious code simple and elegant.


samuelcolvin:
Thanks so much, I'll be rearranging all this at some point, but I'll keep Robusta in any new list of users.

aantn:
Awesome, thanks

PR #3724, opened 2022-01-23
New type StrLimited

### LINKED ISSUE #2104: Simplified length limited string
I often want `constr(max_length=XXX, strip_whitespace=True)`. Could we add a shortcut for this that worked better with mypy?

E.g. something like

```py
class Foobar(BaseModel):
    name: StrLimit[63]
```

This would be equivalent to `constr(max_length=63, strip_whitespace=True)`, e.g a field with a max length of `63`, no min limit, `strip_whitespace` enabled.

The name `StrLimit` is debatable. I think using `ConStr` might be confusing, but perhaps would make more sense?

We could also allow `StrLimit[10, 63]` which would be equivalent to `constr(min_length=10, max_length=63, strip_whitespace=True)`. However this would not be very pythonic in that the first argument meaning would change. Also I find myself wanting this very rarely.

thoughts?

lucastosetto:
## Change Summary

This represents a simplified length limited string and is an alias for the following syntax:

`StrLimited[max_length]` = `constr(max_length=max_length, strip_whitespace=True)`

Usage example:

```
class Foo:
    bar: StrLimited[10] # bar is a str with maximum length of 10 characters
```

## Related issue number

fix #2104

## Checklist

### DIFF
--- a/changes/2104-lucastosetto.md
+++ b/changes/2104-lucastosetto.md
@@ -1,1 +1,1 @@
-Added simplified length limited string type, declared as StrLimited[max_length].
+Added simplified length limited string type, declared as `StrLimited[max_length]`.
--- a/changes/2104-lucastosetto.md
+++ b/changes/2104-lucastosetto.md
@@ -1,0 +1,1 @@
+Added simplified length limited string type, declared as StrLimited[max_length].
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -97,6 +97,7 @@ __all__ = [
     'StrictBytes',
     'StrictInt',
     'StrictFloat',
+    'StrLimited',
     'PaymentCardNumber',
     'ByteSize',
     'PastDate',
@@ -434,6 +435,12 @@ def constr(
         regex=regex and re.compile(regex),
     )
     return _registered(type('ConstrainedStrValue', (ConstrainedStr,), namespace))
+
+
+class StrLimited:
+    @classmethod
+    def __class_getitem__(cls, max_length: int) -> Type[str]:
+        return constr(max_length=max_length, strip_whitespace=True)
 
 
 if TYPE_CHECKING:
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -439,7 +439,7 @@ def constr(
 
 class StrLimited:
     @classmethod
-    def __class_getitem__(cls, max_length: int) -> Type[str]:
+    def __class_getitem__(cls, max_length: int) -> Type[ConstrainedStr]:
         return constr(max_length=max_length, strip_whitespace=True)
 
 
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -738,7 +738,7 @@ def test_str_limited_good():
     max_length = 5
 
     class Model(BaseModel):
-        v: StrLimited[max_length]
+        v: StrLimited[5]
 
     m = Model(v='abcde')
     assert m.v == 'abcde'
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -61,6 +61,7 @@ from pydantic import (
     StrictFloat,
     StrictInt,
     StrictStr,
+    StrLimited,
     ValidationError,
     conbytes,
     condecimal,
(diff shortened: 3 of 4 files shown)


lucastosetto:
please review

changes/2104-lucastosetto.md
```diff
@@ -0,0 +1 @@
+Added simplified length limited string type, declared as StrLimited[max_length].
```
samuelcolvin, line 1:
```suggestion
Added simplified length limited string type, declared as `StrLimited[max_length]`.
```

pydantic/types.py
```diff
@@ -436,6 +437,12 @@ def constr(
     return _registered(type('ConstrainedStrValue', (ConstrainedStr,), namespace))
 
 
+class StrLimited:
+    @classmethod
+    def __class_getitem__(cls, max_length: int) -> Type[str]:
```
samuelcolvin, line 442:
```suggestion
    def __class_getitem__(cls, max_length: int) -> Type[ConstrainedStr]:
```

pydantic/types.py
```diff
@@ -436,6 +437,12 @@ def constr(
     return _registered(type('ConstrainedStrValue', (ConstrainedStr,), namespace))
 
 
+class StrLimited:
```
samuelcolvin, line 440: you might want to implement `__get_validators__` and raise an error people try to use `StrLimited` with parameterising it.

tests/test_types.py
```diff
@@ -733,6 +734,34 @@ class Model(BaseModel):
     ]
 
 
+def test_str_limited_good():
+    max_length = 5
+
+    class Model(BaseModel):
+        v: StrLimited[max_length]
```
samuelcolvin, line 741: cleanest to hard code the length
```suggestion
        v: StrLimited[5]
```

tests/test_types.py
```diff
@@ -733,6 +734,34 @@ class Model(BaseModel):
...
+
+    m = Model(v='abcde')
```
samuelcolvin, line 743:
```suggestion
    m = Model(v='abcde ')
```
show that it also strips white space.

tests/test_types.py
```diff
@@ -733,6 +734,34 @@ class Model(BaseModel):
...
+    assert m.v == 'abcde'
+
+
+def test_str_limited_too_long():
+    max_length = 5
+
+    class Model(BaseModel):
+        v: StrLimited[max_length]
```
samuelcolvin, line 751:
```suggestion
        v: StrLimited[5]
```
same, always best to hard code tests unless you need to vary the value, which you don't here.

pydantic/types.py
samuelcolvin, line 442: We could also be clever and allow slice usage:

```py
class MyModel(BaseModel):
    has_max_length: StrLimited[10]
    has_min_and_max_length: StrLimited[5:10]
```

WDYT?

samuelcolvin (commented):
Looking good, will need to be added to docs.

samuelcolvin:
please update.

tests/test_types.py
```diff
@@ -733,6 +734,34 @@ class Model(BaseModel):
     ]
 
 
+def test_str_limited_good():
+    max_length = 5
```
samuelcolvin, line 738: you can remove this, it's not used.

tests/test_types.py
```diff
@@ -733,6 +734,34 @@ class Model(BaseModel):
...
+
+    class Model(BaseModel):
+        v: StrLimited[5]
+
+    m = Model(v='abcde ')
+    assert m.v == 'abcde'
+
+
+def test_str_limited_too_long():
+    max_length = 5
```
samuelcolvin, line 748: you can remove this and just use a hard-coded value below, it's not adding anything.

samuelcolvin (changes requested):
Looking good, but a few other things still to do:
* `__get_validators__` as suggested above
* add docs, as suggested above

samuelcolvin:
@lucastosetto any chance you can fix this?

samuelcolvin:

Or if you need this in V1.10 but don't have time to complete it (or aren't the author), please comment here and on #4324

cloudwalk-tests:
> @lucastosetto any chance you can fix this?

@samuelcolvin Sure, I can work on it this week, but I won't have time to do this for v1.10

lucastosetto:
> @lucastosetto any chance you can fix this?

Sure, I can work on it this week, but I won't have time to do it for v1.10 release.

samuelcolvin:
Ok, I might try to finish it for you.

If it doesn't make it into v1.10 much of the implementation will have to change for V2.

samuelcolvin:
> Ok, I might try to finish it for you.

Ignore that, I'm not going to have time to work on this, I think it'll have to wait for V2.

pydantic/types.py
```diff
@@ -436,6 +437,12 @@ def constr(
...
+    @classmethod
+    def __class_getitem__(cls, max_length: int) -> Type[ConstrainedStr]:
+        return constr(max_length=max_length, strip_whitespace=True)
```
antonagestam, line 443: This usage of `__class_getitem__` isn't compatible with typing, at least mypy will give an error for `StrLimited[10]`, and I believe the other major type checkers would too. Mypy gives: `error: The type "Type[StrLimited]" is not generic and not indexable [misc]`. I'd suggest redesigning this to make `StrLimited` an object that implements `__get_item__` instead.

PR #3740, opened 2022-01-27
Initial implementation of ConstrainedDate, condate

hottwaj:
## Change Summary

Initial implementation of ConstrainedDate, condate, with tests and docs

- ConstrainedDate piggybacks off `ConstrainedNumberMeta` and `number_size_validator` to implement the gt/ge/lt/le logic

- Perhaps those should be renamed to hint that they are not just applicable to numbers?

- if accepted, more doc changes will be needed to highlight that date validation can be constrained

- also if accepted suggest similar classes for datetime, time may be needed?

Many thanks for great library!

(p.s. I replaced date with datetime.date throughout types.py to avoid inevitable confusion that otherwise arises between  datetime module vs datetime class)

## Related issue number

fix #3739

## Checklist

* [?] Tests pass on CI and coverage remains at 100%


hottwaj:
seems like using `ConstrainedNumberMeta` as metaclass for `ConstrainedDate` has broken the tests relating to Hypothesis strategies (_hypothesis_plugin.py)

Would appreciate some guidance on whether this approach makes sense and how to add necessary changes to _hypothesis_plugin.py

samuelcolvin:
> seems like using `ConstrainedNumberMeta` as metaclass for `ConstrainedDate` has broken the tests relating to Hypothesis strategies (_hypothesis_plugin.py)
> 
> Would appreciate some guidance on whether this approach makes sense and how to add necessary changes to _hypothesis_plugin.py

@Zac-HD any ideas?

pydantic/_hypothesis_plugin.py
```diff
@@ -325,6 +325,24 @@ def resolve_conint(cls):  # type: ignore[no-untyped-def]
...
+        min_value = cls.gt
+    if cls.lt is not None:
+        assert max_value is None, 'Set `lt` or `le`, but not both'
+        max_value = cls.lt
+    s = st.dates(min_value, max_value)
+    if cls.lt is not None:
+        s = s.filter(lambda d: d < cls.lt)
+    if cls.gt is not None:
+        s = s.filter(lambda d: cls.gt < d)
+    return s
```
Zac-HD, line 343: `st.dates()` doesn't accept `None` as a minimum value, because you could instead pass `date.min`.  You can't don't need to make these assertions, since you can add one day to your minumum (/substract from maximum).  Therefore:

```suggestion
    if cls.ge is not None:
        assert cls.gt is None, 'Set `gt` or `ge`, but not both'
        min_value = cls.ge
    elif cls.gt is not None:
        min_value = cls.gt + datetime.timedelta(days=1)
    else:
        min_value = datetime.date.min
    if cls.le is not None:
        assert cls.lt is None, 'Set `lt` or `le`, but not both'
        max_value = cls.le
    elif cls.lt is not None:
        max_value = cls.lt - datetime.timedelta(days=1)
    else:
        max_value = datetime.date.max
    return st.dates(min_value, max_value)
```

Zac-HD (commented):
Suggested fix below, though I haven't tested it.  I'd also recommend renaming the class `ConstrainedScalarMeta`, because a date (or datetime) is definitely not a number.

Zac-HD:
Unsubscribing because this looks like standard "read CI logs and fix errors" territory; ping me again if you have a Hypothesis question.

hottwaj:
Thanks for the help and changes, I've now got initial version working and all tests pass.

Some follow up questions for @samuelcolvin 
- I agree with Zac-HD's suggestion to rename ConstrainedNumberMeta to ConstrainedScalarMeta - but I have not implemented this yet as it'll make more widespread changes.  at the moment it is fairly straightforward to review the changes.  let me know if you'd like me to go ahead with that rename?
- more docs changes are needed to highlight that date validation can be constrained.  I can add these as long as your are happy with basic implementation
- should I add similar classes for datetime.datetime, and maybe even datetime.time?  might be overkill?

Thanks!

hottwaj:
ready for review

hottwaj:
please review

docs/usage/types.md
```diff
@@ -904,6 +908,15 @@ The following arguments are available when using the `conbytes` type function
 - `max_length: int = None`: maximum length of the byte string
 - `strict: bool = False`: controls type coercion
 
+### Arguments to `condate`
+The following arguments are available when using the `condate` type function
+
+- `strict_formats: List[str] = None`: list of datetime.datetime.strpformat style date formats that will be accepted
```
samuelcolvin, line 914: Let's remove this, it's not supported in V2 so it'll be confusing if we add it now in V1.

(the bounds constraints are already implemented in V2).

pydantic/types.py
```diff
@@ -1087,32 +1090,78 @@ def to(self, unit: str) -> float:
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ DATE TYPES ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 if TYPE_CHECKING:
-    PastDate = date
-    FutureDate = date
+    PastDate = datetime.date
```
samuelcolvin, line 1093: Best to revert these changes.

tests/test_condate.py
```diff
@@ -0,0 +1,75 @@
+from datetime import date, datetime
```
samuelcolvin, line 1: can you move this into `test_datetime_parse.py`, We could do with rearranging tests, putting some tests in this file will cause more confusion.

pydantic/types.py
```diff
@@ -1087,32 +1090,78 @@ def to(self, unit: str) -> float:
...
+    ge: OptionalDate = None
+    lt: OptionalDate = None
+    le: OptionalDate = None
+
+    @classmethod
+    def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
+        update_not_none(field_schema, exclusiveMinimum=cls.gt, exclusiveMaximum=cls.lt, minimum=cls.ge, maximum=cls.le)
+
+    @classmethod
+    def strict_date_validator(cls, value: Any) -> datetime.date:
```
samuelcolvin, line 1136: as per my above comment, remove this.

samuelcolvin (commented):
Looking good, some things to change.

samuelcolvin:
you'll also need to rebase or merge master to get all a green tick.

hottwaj:
Hopefully that covers everything you mentioned :)

Sorry if some of this overlaps with changes already envisioned for V2

And thanks for the great work!

changes/3740-hottwaj.md
```diff
@@ -0,0 +1 @@
+Added ConstrainedDate, condate
```
samuelcolvin, line 1:
```suggestion
Added `ConstrainedDate` and `condate`
```

pydantic/types.py
```diff
@@ -1113,32 +1116,62 @@ def to(self, unit: str) -> float:
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ DATE TYPES ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 if TYPE_CHECKING:
+    from datetime import date
```
samuelcolvin, line 1119: This is not correct, we should just do

```py
from datetime import date
```

above in the module scope, then use `date` etc. everywhere. This will reduce the size of the change significantly.

pydantic/types.py
```diff
@@ -1113,32 +1116,62 @@ def to(self, unit: str) -> float:
...
+
+class ConstrainedDate(datetime.date, metaclass=ConstrainedNumberMeta):
+    gt: OptionalDate = None
+    ge: OptionalDate = None
+    lt: OptionalDate = None
+    le: OptionalDate = None
+
+    @classmethod
+    def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
+        update_not_none(field_schema, exclusiveMinimum=cls.gt, exclusiveMaximum=cls.lt, minimum=cls.ge, maximum=cls.le)
```
samuelcolvin, line 1160: can you add tests to check that schema is correct.

Specifically are dates encoded correctly in `schema_json()`?

docs/usage/types.md
```diff
@@ -928,6 +932,14 @@ The following arguments are available when using the `conbytes` type function
 - `max_length: int = None`: maximum length of the byte string
 - `strict: bool = False`: controls type coercion
 
+### Arguments to `condate`
+The following arguments are available when using the `condate` type function
+
+- `gt: float = None`: enforces date to be greater than the set value
+- `ge: float = None`: enforces date to be greater than or equal to the set value
+- `lt: float = None`: enforces date to be less than the set value
+- `le: float = None`: enforces date to be less than or equal to the set value
```
samuelcolvin, line 941:
```suggestion
- `gt: date = None`: enforces date to be greater than the set value
- `ge: date = None`: enforces date to be greater than or equal to the set value
- `lt: date = None`: enforces date to be less than the set value
- `le: date = None`: enforces date to be less than or equal to the set value
```

surely?

pydantic/_hypothesis_plugin.py
```diff
@@ -327,6 +328,25 @@ def resolve_conint(cls):  # type: ignore[no-untyped-def]
     return st.integers(min_value, max_value).map(lambda x: x * cls.multiple_of)
 
 
+@resolves(pydantic.ConstrainedDate)
+def resolve_condate(cls):  # type: ignore[no-untyped-def]
```
samuelcolvin, line 332: could you extend the hypothesis tests to confirm this is working correctly.

samuelcolvin (commented):
Looking good but a few more things to fix.

pydantic/types.py
hottwaj: Sure - just trying to find an example to follow or test to modify...

Would this be the right place to do that?
https://github.com/pydantic/pydantic/blob/master/tests/test_schema.py#L459

samuelcolvin: well probably best to do something like

```py
assert json.loads(model.schema_json()) == {
    ...
}
```
to avoid the test being brittle while testing the json data.

hottwaj: only just saw this comment - thanks.  Will submit another change shortly to adjust

tests/test_schema.py
```diff
@@ -459,26 +460,17 @@ def test_json_schema():
     class Model(BaseModel):
         a = b'foobar'
         b = Decimal('12.34')
+        c = date(2022, 8, 11)
```
samuelcolvin, line 463: This isn't testing `condate`! hence the coverage is not 100%

hottwaj: sorry I got the wrong idea about what to test.  think I've got it now, will push shortly

hottwaj:
Hopefully nearly there now :)

PR #3783, opened 2022-02-05
Fix mypy version check

KotlinIsland:
## Change Summary

Basically the mypy plugin is broken with any dev version of mypy, I changed how the version number is parsed.

```
>mypy --version 
mypy 0.940+dev.04cac4b5d911c4f9529e6ce86a27b44f28846f5d.dirty

>mypy test.py       
pyproject.toml:1: error: Error importing plugin "pydantic.mypy": could not convert string to float: 
Found 1 error in 1 file (errors prevented further checking)
```

Also mypy will switch to a version number like `x.y.z` which the current check will fail on. It will be better to future proof this check to prevent a sudden wave of reports of "could not convert string to float".

## Related issue number

N/A
 

## Checklist

* [N/A] Unit tests for the changes exist

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -76,7 +76,8 @@ BASEMODEL_FULLNAME = 'pydantic.main.BaseModel'
 BASESETTINGS_FULLNAME = 'pydantic.env_settings.BaseSettings'
 FIELD_FULLNAME = 'pydantic.fields.Field'
 DATACLASS_FULLNAME = 'pydantic.dataclasses.dataclass'
-BUILTINS_NAME = 'builtins' if float(mypy_version) >= 0.930 else '__builtins__'
+MYPY_VERSION = tuple(int(part) for part in mypy_version.split('+')[0].split('.'))
+BUILTINS_NAME = 'builtins' if MYPY_VERSION >= (0, 930) else '__builtins__'
 
 
 def plugin(version: str) -> 'TypingType[Plugin]':
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -61,7 +61,6 @@ from mypy.types import (
 from mypy.typevars import fill_typevars
 from mypy.util import get_unique_redefinition_name
 from mypy.version import __version__ as mypy_version
-from typing_extensions import Literal
 
 from pydantic.utils import is_valid_field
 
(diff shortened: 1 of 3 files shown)


KotlinIsland:
please review

pydantic/mypy.py
```diff
@@ -68,7 +68,8 @@
 BASESETTINGS_FULLNAME = 'pydantic.env_settings.BaseSettings'
 FIELD_FULLNAME = 'pydantic.fields.Field'
 DATACLASS_FULLNAME = 'pydantic.dataclasses.dataclass'
-BUILTINS_NAME = 'builtins' if float(mypy_version) >= 0.930 else '__builtins__'
+MYPY_VERSION = tuple(int(part) for part in mypy_version.split('+')[0].split('.'))
+BUILTINS_NAME = 'builtins' if MYPY_VERSION >= (0, 930) else '__builtins__'
 
 
 def plugin(version: str) -> 'TypingType[Plugin]':
```
sobolevn, line 75: You can use `version` string here as well.

KotlinIsland: What benefit would that provide over importing it directly?

KotlinIsland: Something like this?
```py
BUILTINS_NAME: str


def plugin(version: str) -> 'TypingType[Plugin]':
    global BUILTINS_NAME
    BUILTINS_NAME = tuple(int(part) for part in version.split('+')[0].split('.'))

    return PydanticPlugin
```

samuelcolvin:
LGTM, please add a change description.

KotlinIsland:
This is not optimal, it should ideally use the version string passed from mypy, not import from `mypy.version`.

pydantic/mypy.py
```diff
@@ -68,7 +68,8 @@
 BASESETTINGS_FULLNAME = 'pydantic.env_settings.BaseSettings'
 FIELD_FULLNAME = 'pydantic.fields.Field'
 DATACLASS_FULLNAME = 'pydantic.dataclasses.dataclass'
-BUILTINS_NAME = 'builtins' if float(mypy_version) >= 0.930 else '__builtins__'
+MYPY_VERSION = tuple(int(part) for part in mypy_version.split('+')[0].split('.'))
+BUILTINS_NAME = 'builtins' if MYPY_VERSION >= (0, 930) else '__builtins__'
```
samuelcolvin, line 72: Please move this into a function so we can test some real life values.

samuelcolvin (commented):
Looking good, you'll need to rebase or merge master to get tests passing.

samuelcolvin:
please update.

pydantic/mypy.py
```diff
@@ -68,7 +68,8 @@
 BASESETTINGS_FULLNAME = 'pydantic.env_settings.BaseSettings'
 FIELD_FULLNAME = 'pydantic.fields.Field'
 DATACLASS_FULLNAME = 'pydantic.dataclasses.dataclass'
-BUILTINS_NAME = 'builtins' if float(mypy_version) >= 0.930 else '__builtins__'
+MYPY_VERSION = tuple(int(part) for part in mypy_version.split('+')[0].split('.'))
```
samuelcolvin, line 71:
```suggestion
MYPY_VERSION = tuple(int(part) for part in mypy_version.split('+', 1)[0].split('.'))
```

pydantic/mypy.py
KotlinIsland: Like that?

samuelcolvin:
thanks so much.

PR #3824, opened 2022-02-16
Don't override `__init__` method in MyPy plugin if it already exists

patrick91:
## Change Summary

This PR updates the MyPy plugin to not override the `__init__` method if there's already one in the model.

Here's an example of the current issue:

https://replit.com/@patrick91/AngryThoseMath#main.py

running `mypy main.py` returns the following:

```
main.py:13: error: Too many positional arguments for "Person"
main.py:13: error: Missing named argument "name" for "Person"
main.py:13: error: Missing named argument "birth_year" for "Person"
main.py:15: note: Revealed type is "def (*, id: Any, name: Any, birth_year: Any, **kwargs: Any) -> main.Person"
main.py:16: note: Revealed type is "def (__pydantic_self__: main.Person, *, id: Any, name: Any, birth_year: Any, **kwargs: Any)"
```

while the code works since we have a custom constructor:

```python


class Person(BaseModel):
    id: int
    name: str
    birth_year: int

    def __init__(self, id: int) -> None:
        super().__init__(id=id, name="Patrick", birth_year=1991)


print(Person(1))

reveal_type(Person)
reveal_type(Person.__init__)

```

## Checklist

### DIFF
--- a/changes/3824-patrick91.md
+++ b/changes/3824-patrick91.md
@@ -1,0 +1,1 @@
+Fix MyPy plugin to not override pre-existing `__init__` method in models.
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -392,7 +392,9 @@ class PydanticModelTransformer:
         if not self.should_init_forbid_extra(fields, config):
             var = Var('kwargs')
             init_arguments.append(Argument(var, AnyType(TypeOfAny.explicit), None, ARG_STAR2))
-        add_method(ctx, '__init__', init_arguments, NoneType())
+
+        if '__init__' not in ctx.cls.info.names:
+            add_method(ctx, '__init__', init_arguments, NoneType())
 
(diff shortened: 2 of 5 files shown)


patrick91:
coverage fails because we don't hit these lines anymore: https://smokeshow.helpmanual.io/5k4z1l3e0n1q425g2r40/d_2705a9c05af28909_mypy_py.html#t644

I'm not sure if they are needed anymore

patrick91:
please review

samuelcolvin:
this is great, thank you so much.

samuelcolvin:
Looks like the coverage has dropped, hence this not merging, see [here](https://smokeshow.helpmanual.io/0c5f6q6v6b105c3u0q21/d_2705a9c05af28909_mypy_py.html).

Looks like the code which is no-longer covered can be removed, but if you're not sure, maybe safer to add `no cover`.

Please update.

patrick91:
done! 

please review

PR #3826, opened 2022-02-17
speed up __eq__

zrothberg:
Please don't merge right away I think I can speed up a lot more. 

__eq__ is written really simple and can be greatly speed up. This one line resulted in about 5000x times faster comparison for cases of is the same object. Breaking up the .dict() into generators and checking keys one at a time would results in significantly faster code for one of the most common expressions used. I have to verify order integrity and a few other things but there is likely a lot that can be done for this method and result in MUCH faster code. Including but not limited to, tracking len, smart search optimization (simple types first), and some other hash based fuckery.

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -870,6 +870,9 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
         return keys
 
     def __eq__(self, other: Any) -> bool:
+        if self is other:
+            return True
+
         if isinstance(other, BaseModel):
             return self.dict() == other.dict()
         else:
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -691,6 +691,36 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
 
     @classmethod
     @no_type_check
+    def _get_value_lazy(
+        cls,
+        v: Any,
+        by_alias: bool,
+        include: Optional[Union['AbstractSetIntStr', 'MappingIntStrAny']],
+        exclude: Optional[Union['AbstractSetIntStr', 'MappingIntStrAny']],
+        exclude_unset: bool,
+        exclude_defaults: bool,
+        exclude_none: bool,
+    ) -> Any:
+
+        if isinstance(v, BaseModel):
+            v_dict = partial(v._dict_prep,
+                             to_dict=True,
+                             by_alias=by_alias,
+                             exclude_unset=exclude_unset,
+                             exclude_defaults=exclude_defaults,
+                             include=include,
+                             exclude=exclude,
+                             exclude_none=exclude_none,
+                             )
+            # no idea how to handle rootkey
+            # if ROOT_KEY in v_dict:
+            #     return v_dict[ROOT_KEY]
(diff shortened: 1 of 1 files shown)


zrothberg:
@samuelcolvin @PrettyWood This still needs some work but early benchmarks are pretty crazy. They are a docstring in the eq method right now. 

Optimized is 5-1000 times faster at returning not equal on good cases (same object and nothing similar at all). It can get faster for more cases depending on the behavior of lists. Right now Im treating them as simple types. I am not 100% if we have defined anywhere how lists should compare is it element wise or set like? If its element wise I can speed this up a great deal further. 

Instead of returning the object in full like it does right now I am returning the partials and only parsing them as we compare this makes it so the sooner it fails the more we saved.

samuelcolvin:
These improvements look great in principle, but I think we should wait for V2 so we can break existing behaviour.

@zrothberg what do you think about implementing `__eq__` in rust as part of pydantic-core? If you're interested in the challenge, please create an issue on pydantic-core to discuss.

zrothberg:
@samuelcolvin I have actually messed around with some other stuff for this a fair deal. I think a fairly significant amount of progress can be made on the pure python side by switching to using a custom abstractmapping in multiple places. It would allow the ability to push to push arguments like the current dict method else without the need to parse the object outright. It would save on significant amount of intermediary conversions and memory creation. Especially for places where objects are being set or being destroyed outright. Using that custom mapping as the target for the rust part would also leave as much customizability into core pydanitc as possible. 

I can open a PR if you want to see an example of what I am talking about against the current setup so you can evaluate it relative to the pydantic-core setup. I am not entirely sure how flexible that side is and haven't written enough rust to make a proper eval.

samuelcolvin:
That sound really interesting, PR welcome. I don't mind if you create it here or on pydantic-core.

When you say "custom mapping", do you mean in place of `model.__dict__`? Creating the dict at the core of validation is one of the slowest parts of pydantic-core, it would be amazing if we could replace that with something where the underlying object is a rust type, wrapped for access from python. But I'm not sure we could make it faster than a plain python dict.

Currently pydantic-core doesn't import anything from pure python, I think we could do it if it's genuinely faster.

zrothberg:
Yea so instead of needing to convert the object to a series of dicts inorder to turn it into a json or to send it to other objects as arguments. You can instead just wrap the object in a [Mapping. ](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) This lets you avoid actually creating the dict in memory and just do so lazily. For child models you just do the same thing and keep returning wrapped objects. I use the trick to avoid turning normal dataclasses into dictionaries. Let me show you that one as it will make more sense from there. It's pretty straightforward to use. 

As far as doing it in pure rust, I don't see any issue other then you need to make sure a few odd fields are set to the correct value so that the c code from python can detect its a mapping. https://docs.python.org/3/c-api/mapping.html#c.PyMapping_Check That covers some of them and I believe you have to specify some other marker called TP_FlAGS but I am not super familiar with that part of the C-api.

PR #3881, opened 2022-03-04
Add jsons benchmark

mbroedel:
Not sure if I should have created an issue first but this is pretty straightforward. Please review

### DIFF
--- a/benchmarks/requirements.txt
+++ b/benchmarks/requirements.txt
@@ -12,3 +12,4 @@ cattrs
 cerberus
 voluptuous
 schematics
+jsons
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -64,6 +64,12 @@ except Exception as e:
     print('WARNING: unable to import TestSchematics')
     TestSchematics = None
 
+try:
+    from test_jsons import TestJsons
+except Exception as e:
+    print('WARNING: unable to import TestJsons')
+    TestJsons = None
+
 PUNCTUATION = ' \t\n!"#$%&\'()*+,-./'
 LETTERS = string.ascii_letters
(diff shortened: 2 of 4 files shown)


mbroedel:
My benchmark results are very different from those in the docs, I'm guessing it's because I'm on ARM:  

Package | Version | Relative Performance | Mean validation time
--- | --- | --- | ---
attrs + cattrs | `21.4.0` |  | 23.0μs
valideer | `0.4.2` | 1.4x slower | 31.7μs
pydantic | `1.9.0` | 1.8x slower | 42.2μs
marshmallow | `3.14.1` | 3.0x slower | 68.3μs
voluptuous | `0.12.2` | 3.2x slower | 72.9μs
trafaret | `2.1.0` | 3.4x slower | 79.1μs
jsons | `1.6.1` | 10.9x slower | 249.1μs
schematics | `2.1.1` | 11.9x slower | 272.5μs
django-rest-framework | `3.13.1` | 14.2x slower | 325.9μs
cerberus | `1.3.4` | 27.4x slower | 629.6μs

I've refrained from pushing them for this reason.

samuelcolvin:
Sorry but I can't accept this. See https://github.com/samuelcolvin/pydantic/pull/1810#issuecomment-706050989

I'll remove the benchmarks soon, they're wasting peoples time (like this, I'm so sorry). They need a lot of work and that's outside the bounds of pydantic.

mbroedel:
I see, that's unfortunate. Looking further into it I think wyfo makes a lot of good points here https://github.com/samuelcolvin/pydantic/pull/3264 and the benchmarks really ought to be removed, glad you agree. I originally became interested in the benchmarks because pydantic often didn't behave the way I wanted it to and its supposed performance edge and in particular the benchmarks were pointed to as the reason why, see https://github.com/samuelcolvin/pydantic/issues/1157#issuecomment-572694302. I feel a little misled but the comment is admittedly pretty old and a lot of things seem to have changed since then (though the benchmarks page doesn't reflect that unfortunately)

Issue #2329, opened 2021-02-08
StrictBool, StrictStr etc types are not drop in replacements for normal types


spookylukey:
Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.7.3
            pydantic compiled: True
                 install path: /home/luke/.virtualenvs/rpr/lib/python3.9/site-packages/pydantic
               python version: 3.9.1 (default, Jan 26 2021, 09:48:27)  [GCC 9.3.0]
                     platform: Linux-5.4.0-58-generic-x86_64-with-glibc2.31
     optional deps. installed: []
```

In places where you are trying to tighten validation by going from `int` to `StrictInt` etc, these types fail as drop in replacements in several ways. Specifically, if you call the types directly, they produce instances of `StrictInt` etc, instead of instances of `int` etc. This produces some incompatibilities.

For `StrictBool`, `repr()` output:

```py
>>> from pydantic import StrictBool
>>> repr(StrictBool(False))
'0'
# Compared with:
>>> repr(bool(False))
'False'
```

For `StrictBool`, `isinstance()` checks:
```py
>>> from pydantic import StrictBool
>>> isinstance(StrictBool(False), bool)
False
```

For `StrictBool`, any code that does `is` checks, which is a legitimate technique, especially when you want to check for `False` as distinct from `None`:

```py
>>> StrictBool(False) is False
False
# Compared with
>>> bool(False) is False
True
```

For all of `StrictInt`, `StrictStr`, `StrictBool`, any code that does `type()` checking will behave differently. This is arguably a bug of that code (it should be doing `isinstance()`), but it is worth asking whether the current behaviour is helpful. When you change the type to `StrictInt` instead of `int`, you do so only for different validation behaviour.

## Use case - type driven deserialization

Here is a fuller use case - doing type driven deserialization, similar to `BaseModel.parse_obj`, but with my own logic. 
Starting with these classes:

```py
from enum import Enum
from pydantic import BaseModel


class Color(Enum):
    RED = 'r'
    BLUE = 'b'


class Flag(BaseModel):
    color: Color
    square: StrictBool

```
Suppose we have a serialized instance of `Flag` as JSON string `{"color": "r", "square": true}` (just like `.json()`), which becomes the Python dict `{"color": "r", "square": True}` after JSON decoding. So we make a `deserializer()` like this:

```py
def deserialize(data, cls):
[...]

PR #3885, opened 2022-03-07
Fix StrictBool etc. constructors should return base types

spookylukey:
This PR changes the constructors of `StrictBool`, `StrictInt`, `StrictFloat`, `StrictBytes` and `StrictStr` so that they return the base types instead of a subclass. The motivation for this is on #2329 

In terms of implementation. For `StrictInt` etc, there were tests that ensured that subclassing worked: https://github.com/samuelcolvin/pydantic/blob/d7a8272d7e0c151b0bd43df596be02e0d436ebdf/tests/test_types.py#L1682

I'm not sure why this is needed, but to preserve that behaviour the implementation was constrained such that the change only affects `StrictInt`, and not further subclasses. The logic here is that if you go from `int` to `StrictInt`, the only behaviour change you expect is stricter validation, as mentioned in the ticket. If you have your own further subclass, you presumably have other behaviour you want, and you are responsible for that yourself.

`StrictBool` is a slight exception to this:
- it doesn't inherit from `bool`, but `int` (it seems it's not possible to inherit from `bool`)
- there were no tests for subclassing `StrictBool`
- it's difficult to imagine a use case for subclassing bool

So for `StrictBool.__new__` only, we unconditionally defer to `bool.__new__`.

## Checks

* [x] Documentation reflects the changes where applicable - not needed

### DIFF
--- a/changes/2329-spookylukey.md
+++ b/changes/2329-spookylukey.md
@@ -1,0 +1,1 @@
+Fix StrictBool, StrictInt, StrictFloat, StrictStr, StrictBytes constructors to return instances of the base class
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -167,6 +167,9 @@ else:
         StrictBool to allow for bools which are not type-coerced.
         """
 
+        def __new__(self, value=False):
+            return bool(value)
+
         @classmethod
         def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
             field_schema.update(type='boolean')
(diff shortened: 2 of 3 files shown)


spookylukey:
please review

samuelcolvin:
I'm a bit confused on where this fits in with #2688?

My instinct is to defer this until V2 to avoid any breaking behaviour. WDYT?

spookylukey:
Personally I've worked around this issue, so I'm happy to wait for V2.

I think the decision here is independent of that in #2688 - in that case there is an argument for having the output be a separate type from the base type, but here I don't think there is.

samuelcolvin:
great, please you agree.

adriangb:
So I think we already implemented the above in #4516. I think this resolves most of the issues in the original request:

```python
>>> from pydantic import StrictBool
>>> repr(StrictBool(False))
'False'
>>> isinstance(StrictBool(False), bool)
True
>>> assert StrictBool(False) is False
>>>
```

I think we can close this and #2329.

samuelcolvin:
@spookylukey, thanks so much for your contribution, looks like this is all being solved on the road to V2.

PR #3921, opened 2022-03-20
BaseModel: Document and test structural pattern matching

### LINKED ISSUE #3920: Support python3.10 structural pattern matching (PEP 634-636)
### Checks

# Feature

Python 3.10 introduces the new `match` syntax. https://peps.python.org/pep-0636/

Example:
```python
num = 5

match num:
    case 5:
        print('is five')
    case default:
        print('is not five')
```

They support matching on custom classes too, with the `__match_args__` attribute.

```python
class MyClass:
    __match_args__ = ('num1', 'num2')

    def __init__(self, num1, num2):
        self.num1 = num1
        self.num2 = num2


myinstance = MyClass(5, 10)


match myinstance:
    case MyClass(5, num2):
        print(f'is 5 and {num2}')
    case MyClass(_, num2):
        print(f'is not 5 and is {num2}')
```

I've found myself adding this attribute to pydantic classes by hand, eventhough it seems pretty intuitive to me that the default behavior should match fields in the order that they were specified. So my code looks kind of like this:

```python
class UpdateThing(BaseModel):
    __match_args__ = ("thing", "value")
    thing: Thing
    value: str


class AddThing(BaseModel):
    __match_args__ = ("index", "thing")
    index: int
    thing: Thing
```

Might as well add it natively?

I'm opening this issue to couple it with a PR :)

irgolic:
## Change Summary

Adds `__match_args__` attribute to BaseModel namespace, to support python3.10 structural pattern matching (`match` syntax).

## Related issue number

Fix #3920

## Checklist


irgolic:
please review

samuelcolvin:
This looks great, thank so much.

Please add tests and some to documentation.

There are lots of example of tests with new syntax in `test_forward_ref.py`

samuelcolvin:
@irgolic will you be able to work on this and get it ready to merge? I'd really love to add this feature.

irgolic:
> @irgolic will you be able to work on this and get it ready to merge? I'd really love to add this feature.

Hey, sorry for the late reply. I'd love to, I'll get on it.

One concern I've thought of – is there ever a case when `__match_args__` should be updated? Or can we assume that the order and structure of fields is always preserved since class initialization?

samuelcolvin:
After class init, yes it should be pretty well constant. I'm theory it could change after instantiation because of extra, but I think we can ignore that for now.

irgolic:
`make lint` and `make docs` are passing locally for me after amending `--target-version` to `py310` in Makefile. Any idea why it's failing as if it hasn't been updated in the runner?

samuelcolvin:
Are you using the same version of libraries? You might want to merge with master to get the latest versions.

Also please use the "fix #xxx" syntax in the pr body, then use the issue I'd in the change file.

But more importantly, I'm worried about the performance impact of this, it looks like we have to validate a whole new model for each match. Is that right? I don't know anything about pattern matching yet as I almost never get to write python 3.10 only code.

irgolic:
> Are you using the same version of libraries? You might want to merge with master to get the latest versions.

I rebased and reinstalled before starting. I realized I had to write the tests into a string rather than code, otherwise test runners below 3.10 throw syntax errors.

> Also please use the "fix #xxx" syntax in the pr body, then use the issue I'd in the change file.

Nice, thanks!

> But more importantly, I'm worried about the performance impact of this, it looks like we have to validate a whole new model for each match. Is that right? I don't know anything about pattern matching yet as I almost never get to write python 3.10 only code.

I'm not very experienced with it yet either, but from my understanding, it doesn't actually instantiate a new model to compare it to. It's just sugar for _"get these attributes and compare them"_. I just confirmed this by setting a breakpoint in BaseModel's `__init__`, and seeing that it's not called in the `case` statements.

irgolic:
Hmm, I seem to still be doing something wrong with the changefile.

irgolic:
please review

PrettyWood (approved):
> it doesn't actually instantiate a new model to compare it to. It's just sugar for "get these attributes and compare them". 

Yep exactly

------------------------------------------------

LGTM ! Thanks a lot

samuelcolvin:
> Hmm, I seem to still be doing something wrong with the changefile.
> 
> EDIT: Seems like it doesn't support link syntax, only numbers.

Makes sense, it's a [simple regex](https://github.com/pydantic/hooky/blob/21b47182c699b283212480aa81b792c8ebca5a1c/src/logic.py#L206)

Thanks for this, I'll review next week, though I'm probably on holiday until until Friday.

irgolic:
> Makes sense, it's a [simple regex](https://github.com/pydantic/hooky/blob/21b47182c699b283212480aa81b792c8ebca5a1c/src/logic.py#L206)

I was just looking at that. Here's a PR that should include the functionality. https://github.com/pydantic/hooky/pull/1

pydantic/main.py
```diff
@@ -277,6 +277,7 @@ def is_untouched(v: Any) -> bool:
             '__slots__': slots | private_attributes.keys(),
             '__hash__': hash_func,
             '__class_vars__': class_vars,
+            '__match_args__': tuple(fields),
```
samuelcolvin, line 280: I'm actually not sure we should add this, without it pattern matching still works, just not with positional arguments.

Since pydantic models don't support positional arguments, I think adding pseudo support for them in pattern matching could be very confusing.

Either way, the rest of this PR would still be a great addition to pydantic.

docs/usage/models.md
```diff
@@ -643,3 +644,11 @@ In addition, the `**data` argument will always be present in the signature if `C
...
     not necessarily all the types that can actually be provided to that field.
     This may be fixed one day once [#1055](https://github.com/pydantic/pydantic/issues/1055) is solved.
+
+## Structural pattern matching
+
+*pydantic* supports structural pattern matching for models, as introduced by [PEP 634](https://peps.python.org/pep-0622/) in Python 3.10.
+
+```py
+{!.tmp_examples/models_structural_pattern_matching.py!}
+```
```
samuelcolvin, line 654: can we add a [note admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#admonition-icons) explaining that a new model is not created, and this is just syntactic super form matching?

Dumb or ill-informed people (like me) who haven't used pattern matching before could be confused and worry about a performance overhead for creating models which are not actually created.

samuelcolvin (commented):
Overall, this is looking great.

The only real question is whether `__match_args__` should be added as I questioned below.

pydantic/main.py
PrettyWood: Ah I thought this was needed! If kwargs matching already works I reckon we shouldn't add this. Allowing positional arguments only here but not elsewhere is confusing

samuelcolvin: Agreed.

That was my assumption, but I tried it and it seems to work fine without this.

tests/test_match.py
```diff
@@ -0,0 +1,74 @@
+import sys
```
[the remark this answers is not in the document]
irgolic: renamed to `test_structural_pattern_matching.py`

pydantic/main.py
irgolic: I agree. Even though it peforms a completely different function, the match-case syntax looks the same as instantiation, so it looks offputting given that BaseModel doesn't use args.

docs/usage/models.md
```diff
@@ -643,3 +644,11 @@ In addition, the `**data` argument will always be present in the signature if `C
     Types in the model signature are the same as declared in model annotations, 
     not necessarily all the types that can actually be provided to that field.
     This may be fixed one day once [#1055](https://github.com/pydantic/pydantic/issues/1055) is solved.
+
+## Structural pattern matching
+
+*pydantic* supports structural pattern matching for models, as introduced by [PEP 634](https://peps.python.org/pep-0622/) in Python 3.10.
```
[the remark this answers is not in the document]
irgolic: Nice catch on these, PEP 636 is the tutorial and the latest one, probably best one to link.

irgolic:
I've removed the codebase change and made this PR a docs/tests-only change.
(as well as a minor "" => '' change in the ci yaml, seems like there was a big docs change merged today that already updated the docs build python version to 3.10)

please review

samuelcolvin:

irgolic:
> This is great, thanks so much.

Glad to, thanks for pydantic.

One concern... It says "Python 3.7 and above", is there any way to change that to 3.10? Are the 3.9/3.10 versions of code (with the typing changes) autogenerated?
Screenshot 2022-08-16 at 21 59 42

samuelcolvin:
Luckily @Bobronium already thought about that, should be fixed in #4384, let me know if that's okay.

Bobronium:
This is awesome, btw!

irgolic:
> This is awesome, btw!

It's a cool new syntax we'll all need some getting used to, I think :D Reminds me of haskell. Hopefully docs like this in commonly used libraries will help people get acquainted with it.

PR #3941, opened 2022-03-26
Add nested json encoding

### LINKED ISSUE #2277: Json Encoders are ignored in nested structures
### Checks

# Bug

When nesting a model with customer encoders, they are ignored by the parent.

```python
from datetime import datetime, timedelta 
from pydantic.json import timedelta_isoformat 
 
 
class WithCustomEncoders(BaseModel): 
    dt: datetime 
    diff: timedelta 
 
    class Config: 
        json_encoders = { 
            datetime: lambda v: v.timestamp(), 
            timedelta: timedelta_isoformat, 
        } 
 
class ParentWithoutEncoders(BaseModel): 
    child: WithCustomEncoders
 
m = WithCustomEncoders(dt=datetime(2032, 6, 1), diff=timedelta(hours=100)) 
print(m.json())

p = ParentWithoutEncoders(child=m)
print(p.json())


#> {"dt": 1969671600.0, "diff": "P4DT4H0M0.000000S"}
#> {"child": {"dt": "2032-06-01T00:00:00", "diff": 360000.0}}
```

Expected output:
```python
#> {"dt": 1969671600.0, "diff": "P4DT4H0M0.000000S"}
#> {"child": {"dt": 1969671600.0, "diff": "P4DT4H0M0.000000S"}}
```

lilyminium:
#3935 contains a pin to jinja; unsure if I should wait on that or pin myself.

## Change Summary

* adds `encode_as_json` keyword to a bunch of serialisation functions to try getting nested JSON encoders
* adds `use_nested_encoders` to `BaseModel.json`
* `use_nested_encoders` does *not* pair with `models_as_dict=False`; in that case, only the global encoder is used

## Related issue number

fix #2277

## Checklist


changes/3941-lilyminium.md
```diff
@@ -0,0 +1 @@
+add use_nested_encoders keyword argument to BaseModel.json to allow encoding nested subclasses with encoders specified in the inner classes
```
samuelcolvin, line 1:
```suggestion
add `use_nested_encoders` keyword argument to `BaseModel.json` to allow encoding nested subclasses with encoders specified in the inner classes
```

docs/examples/exporting_models_json_nested_encoders.py
```diff
@@ -0,0 +1,42 @@
...
+
+class ParentWithCustomEncoders(BaseModel):
+    diff: timedelta
+    child: WithCustomEncoders
+
+    class Config:
+        json_encoders = {
+            timedelta: lambda v: v.total_seconds(),
+            WithCustomEncoders: lambda _: 'using parent encoder',
+        }
```
samuelcolvin, line 25: I think this makes the example more complicated than it needs to be - I think it's sufficient just to show that `ParentModel` uses `child.Config.json_encoders`.

If there's a more subtle point to be made about how `ParentModel` interacts, either mention it in the code or add another example.

tests/test_json.py
```diff
@@ -325,6 +325,76 @@ class Config:
     )
 
 
+def test_nested_models_with_inner_encoders():
```
samuelcolvin, line 328: please can you split this into multiple tests, you can move some of the model definitions out of the test function(s) to avoid duplication.

pydantic/main.py
```diff
@@ -420,6 +420,7 @@ def dict(
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
         exclude_none: bool = False,
+        encode_as_json: bool = False,
```
samuelcolvin, line 423: do you really need this on `.dict()`?

samuelcolvin (commented):
I think this looks great, a few things  to change and will need to wait for v1.10, but doesn't need to wait until v2.

pydantic/main.py
lilyminium: The function calls go `parent.json()` -> `parent._iter()` -> `parent._get_value()` -> (if child is a BaseModel), `child.dict()` -> `child._iter()` ....

So no, I could alter `_get_value` and call `_iter()` directly. I thought about it but assumed there was a reason to use `dict()` in `_get_value` -- also that maybe having `encode_as_json` in `dict()` could be a feature. Any thoughts?

docs/examples/exporting_models_json_nested_encoders.py
lilyminium: I think this was highlighting that `ParentModel` doesn't use `child.Config.json_encoders` when `models_as_dict=False`, whether `use_nested_encoders=True|False`. That's mentioned in the code as a comment, should it be more explicit?

tests/test_json.py
lilyminium: I split this into a class of tests, because it seemed neater and made sense to me to group the related tests. I did notice that the tests are normally written in quite a flat structure, however. Let me know if I should just convert them to functions.

samuelcolvin: Best if tests are just functions. Class based tests too easily end in a mess.

pydantic/main.py
samuelcolvin: Ok, I understand, best to keep it like this.

samuelcolvin:
Once this is ready to review, please use the special phrase in a comment so I know it's my responsibility.

https://github.com/samuelcolvin/pydantic/blob/8997cc5961139dd2695761a33c06a66adbf1430a/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L21

lilyminium:
please review

07pepa:
@lilyminium  i think this will satify this feature request  https://github.com/samuelcolvin/pydantic/discussions/3910 👍 (i need to serialize date times time deltas and other python stuff) 

also if you will add tests for negative time deltas there is small bug in it time https://github.com/samuelcolvin/pydantic/pull/3909 so do not do it yet or do it after this fix is merge (but i think curent state will suffice. This is just heads up since i saw it in test used.)

barringtonhaynes:
Thanks for working on this, it's a feature I need. When I've tried running these tests using the above code though, if you define any primitive types, string for example it raises a `TypeError: Object of type 'str' is not JSON serialisable`. I've amended the test slightly to show this.

```python
from datetime import datetime, timedelta
from pydantic.json import timedelta_isoformat


class CustomChildModel(BaseModel):
    dt: datetime
    diff: timedelta
    text: str

    class Config:
        json_encoders = {
            datetime: lambda v: v.timestamp(),
            timedelta: timedelta_isoformat,
        }

class ParentModel(BaseModel):
    diff: timedelta
    child: CustomChildModel

    class Config:
        json_encoders = {
            timedelta: lambda v: v.total_seconds(),
            CustomChildModel: lambda _: 'using parent encoder',
        }

child = CustomChildModel(dt=datetime(2032, 6, 1), diff=timedelta(hours=100), text='hello')
parent = ParentModel(diff=timedelta(hours=3), child=child)

# default encoder uses total_seconds() for diff
print(parent.json())

# nested encoder uses isoformat
print(parent.json(use_nested_encoders=True))

# turning off models_as_dict only uses the top-level formatter, however

print(parent.json(models_as_dict=False, use_nested_encoders=True))

print(parent.json(models_as_dict=False, use_nested_encoders=False))
```

Also, in the following example I couldn't get it to json encode the `GrandChildModel` using my custom encoder, but I could if it was a standard python data class.

```python
from datetime import datetime, timedelta
from pydantic.json import timedelta_isoformat


class GrandChildModel(BaseModel):
    dt: datetime
    diff: timedelta

class CustomChildModel(BaseModel):
    dt: datetime
    diff: timedelta
    grand_child: GrandChildModel

    class Config:
        json_encoders = {
            datetime: lambda v: v.timestamp(),
            timedelta: timedelta_isoformat,
            GrandChildModel: lambda v: 'foobar',
        }


class ParentModel(BaseModel):
    diff: timedelta
    child: CustomChildModel

    class Config:
        json_encoders = {
            timedelta: lambda v: v.total_seconds(),
            CustomChildModel: lambda _: 'using parent encoder',
        }


grand_child = GrandChildModel(dt=datetime(2040, 7, 13), diff=timedelta(hours=10))
child = CustomChildModel(dt=datetime(2032, 6, 1), diff=timedelta(hours=100), grand_child=grand_child)
parent = ParentModel(diff=timedelta(hours=3), child=child)

# default encoder uses total_seconds() for diff
print(parent.json())

# nested encoder uses isoformat
print(parent.json(use_nested_encoders=True))

# turning off models_as_dict only uses the top-level formatter, however

print(parent.json(models_as_dict=False, use_nested_encoders=True))

print(parent.json(models_as_dict=False, use_nested_encoders=False))
```

maestro-1 (approved):

samuelcolvin:
thanks so much for this.

pydantic/main.py
mymedia2: It may be worth documenting this new parameter. It still looks useful when actual JSON-encoding occurs later.

samuelcolvin:
So sorry @barringtonhaynes, I've been crashing through PRs and missed your comment.

I guess we should either revert this or work on a separate fix. Do you think you could submit a fix?

barringtonhaynes:
> So sorry @barringtonhaynes, I've been crashing through PRs and missed your comment.
> 
> 
> 
> I guess we should either revert this or work on a separate fix. Do you think you could submit a fix?

I would love to but don't have any time at the moment due to other commitments, it's also possible that I missed something and those use cases do work. It's a super useful feature, but think it needs a little TLC. If it's still around in a month or so I'm happy to have a look.

samuelcolvin:
Okay, I'll look into it. Thanks for the heads up.

And thanks again to @lilyminium for the PR.

samuelcolvin:
Okay, having looked into it, this introduces a lot of problems, for example even in this very simple case the output is not valid JSON:

```py

class Child(BaseModel):
    b: str
    c: int

    class Config:
        json_encoders = {
            str: lambda v: 'custom value!',
        }

class Parent(BaseModel):
    a: str
    child: Child

x = Parent(a='a', child=dict(b='b', c=1))
print(x.json(use_nested_encoders=True))
```

Outputs:

```json
{"a": "\"a\"", "child": {"b": "\"b\"", "c": "1"}}
```
**Note:**:
* `a` is not valid JSON
* `child.b` is not valid JSON
* `child.c` is a string when it should be an int

I'll revert the PR, and if someone else wants to take a stab at this before V1.10 is locked next week, I'll try and review. But realistically, we'll probably need to wait for V2 when serialisation will be completely re-thought.

lilyminium:
Huge apologies for falling off the map -- Google was helpfully filtering out a substantial chunk of my GitHub notifications. Thanks for noticing the problems, @barringtonhaynes and sorry for any complications this introduced.

Issue #3617, opened 2022-01-03
Autocomplete for vscode does not register `Field` defaults when using positional argument


jellis18:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```bash
             pydantic version: 1.9.0
            pydantic compiled: True
                 install path: /Users/.../.../venv/lib/python3.8/site-packages/pydantic
               python version: 3.8.6 (default, Feb  5 2021, 13:58:38)  [Clang 12.0.0 (clang-1200.0.32.21)]
                     platform: macOS-10.16-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']
```

```py
from pydantic import BaseModel, Field


class Thing(BaseModel):
    x: bool = Field(False, description="hello")
    y: float

# this will give a pylance error saying "Argument missing for parameter x"
thing = Thing(y=4.5)

class Thing2(BaseModel):
    x: bool = Field(default=False, description="hello")
    y: float

# this is fine
thing = Thing2(y=4.5)
```
With screenshots we can see the error:
![image](https://user-images.githubusercontent.com/3458354/147979029-cd60681c-bfb1-434d-b4ca-60dd95f96d0a.png)

... and without the error
![image](https://user-images.githubusercontent.com/3458354/147979120-2298448b-e041-45ac-8909-adbd7b29759c.png)

So it looks like there is a bug in the autocomplete when using a positional argument for `default` as opposed to a keyword argument.

I am not hugely familiar with the new [dataclass transform](https://github.com/microsoft/pyright/blob/main/specs/dataclass_transforms.md#field-descriptors) but it looks like it requires that `default` be a keyword argument and not a positional argument.

I see that this is actually on [purpose](https://github.com/samuelcolvin/pydantic/pull/2721/files#diff-8708af8d26c26644976866f974e0b810b0940ede43ff487e457057d5a1f07999R130) but I'm not sure why. If this is needed (and I'm sure it it) then adding a warning or note to the documentation would be helpful.

iwoloschin:
[...]

PR #3972, opened 2022-04-04
test pyright with pydantic

### LINKED ISSUE #3753: BaseSettings - pylance complains for not setting default values to fields
Hi! I am a huge fan. I am just encountering an error at the moment. The code below used to have no issue with `pylance` using previous `pydantic` version. But, right now I can no longer create an instance without having to pass values to fields that are supposed to be set only in `.env` or loaded from environment. It is really nice that it now offers autocomplete (same goes to `BaseModel`) and it actually still works. It's just that I really do not like the view of squiggly lines :smile: Would there be a quick fix for this?

Screenshot from 2022-01-30 10-12-43

These are my `vscode` settings, btw.

```json
"python.analysis.typeCheckingMode": "basic",
"python.languageServer": "Pylance",
```

Thank you in advance for your response.

samuelcolvin:
## Change Summary

Run pyright checks on pydantic code in CI.

## Related issue number

fix #3753

ref #3617 - I'll use this PR to see if I can fix this or find a workaround

## Checklist


florian-sattler:
As mentioned in #3753 I don't think the proposed solution
```python
settings = Settings.parse_obj({})
```
makes it very obvious what is happening. IMHO a type checking bug should be handled on a type checking level.

samuelcolvin:
@florian-sattler I kind of agree but it is a workaround.

Hopefully in v2 we can remove dataclass transforms from `BaseSettings`, unless then, this is better than what went before.

PR #1810, opened 2020-08-07
benchmark of validx library

mathbeal:
Comparing similar libraries is encouraged on pydantic benchmark page [https://pydantic-docs.helpmanual.io/benchmarks/](url)

## Change Summary

- test_validx.py added to compare performance of validx library against pydantic
- datetime.now() replaced by time.perf_counter (see [https://docs.python.org/3/library/time.html#time.perf_counter](url))

## Related issue number

closes #1816

## Checklist

* [x] Unit tests for the changes exist
* [x] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable
* [x] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/samuelcolvin/pydantic/blob/master/changes/README.md) for details)

Issue #1425, opened 2020-04-24
A method to recheck the model


gk-patel:
# Feature Request

Is it possible to add a recheck() method to the model ? 

For example, lets say I have a pydantic BaseModel class called Foo and I have done foo = Foo(**{}) and now I am passing foo to scripts written by others and when I get it back I want to recheck it. 

I could implement something like this,
```
class Foo(BaseModel): 
      v1: int 
      v2: List[float] ..... etc

     def recheck(self): 
           return self.__class__.parse_obj(self.dict())
```
but I dont know if this is the most optimal way. 

Thank you for your time and the wonderful library.

samuelcolvin:
Best to use [`validate_assignment`](https://pydantic-docs.helpmanual.io/usage/model_config/) or add your own method as you have here.

PR #3973, opened 2022-04-04
Remove benchmarks completely

samuelcolvin:
Benchmarks have just started failing, see #3972, so it seems like a good time to remove them.

See #3881, #1810, https://github.com/samuelcolvin/pydantic/issues/1939#issuecomment-695799667 #1425 and many other issues for an explanation of why I'm removing benchmarks.

TL;DR - they're wrong, they're a burden to maintain, they're controversial, they don't really belong in pydantic - they've outlived their usefulness.

In future it would be great if:
* we had a way of benchmarking pydantic between versions to improve performance and avoid performance slipping - I don't have time to do this now, I might in future
* it might also be cool if someone build a comparison of different data validation libraries, a bit like the [framework benchmarks](https://www.techempower.com/benchmarks/). But this would be a big piece of work and is way outside the bounds of pydantic


Kludex:
> we had a way of benchmarking pydantic between versions to improve performance and avoid performance slipping - I don't have time to do this now, I might in future

What do you have in mind here? I know that `h11` has a script that runs a microbenchmark, and people are able to add in the PR, see: https://github.com/python-hyper/h11/pull/119. Are you looking for something like that?

PR #3974, opened 2022-04-04
try speeding up ci using caching and rearranging jobs

samuelcolvin:
## Change Summary

Rearrange steps and using caching  to try to speedup CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
 
     - name: install
       run: make install-testing
-      if: steps.cache.outputs.cache-hit != 'true'
+#      if: steps.cache.outputs.cache-hit != 'true'
 
     - name: compile
       run: |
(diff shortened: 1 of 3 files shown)


Zac-HD:
> Any idea what's going on here? Would caching the `.hypothesis` directory help?

I *don't* have any real idea what's going on here; so while I wouldn't *expect* caching the `.hypothesis` directory to help that might still be worth trying.

My best guess is that the new configuration isn't [registering `NameEmail`, probably due to an import error](https://github.com/samuelcolvin/pydantic/blob/f96a6131a3ad569b1f1ccbb073a4042209d472f4/pydantic/_hypothesis_plugin.py#L50-L75).  You can check this by inspecting the contents of the `hypothesis.strategies._internal.types._global_type_lookup` dict before and afterwards, and/or directly checking that you can `import email_validator`.

PR #3975, opened 2022-04-04
Fix #3807: matching characters in nested env delimeter and env prefix

### LINKED ISSUE #3807: Unexpected behaviour if the prefix contains the env_nested_delimiter
### Checks

# Bug

```
$ python -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.9.0
            pydantic compiled: True
               python version: 3.8.12 (default, Oct 12 2021, 03:05:25)  [GCC 10.2.1 20210110]
                     platform: Linux-5.13.0-28-generic-x86_64-with-glibc2.2.5
```

```py
import os
import pydantic
import pytest


@pytest.fixture
def env():
    original_env = os.environ.copy()
    yield os.environ
    os.environ.clear()
    os.environ.update(original_env)


def test_env_nested_delimiter_in_prefix(env):
    """Test pydantic fails to set value where the prefix has the delimiter.

    Pydantic will break down `settings_sub_value` to
    `{sub: {value: ...}}` and it will pass it to `settings.sub`, so it will fail to
    populate `settings.sub.value` and it will try to set `settings.sub.sub`.
    """

    class SubSettings(pydantic.BaseSettings):
        value: str

    class Settings(pydantic.BaseSettings):
        sub: SubSettings

        class Config:
            env_prefix = "settings_"
            env_nested_delimiter = "_"

    env["settings_sub_value"] = "won't be set"

    expected_error = (
        # settings.sub.value doesn't get set
        r".*sub -> value.*\n?.*value_error\.missing.*\n?"
        # Tries to assign settings.sub.sub
        r".*sub -> sub.*\n?.*value_error\.extra"
    )

    with pytest.raises(pydantic.ValidationError, match=expected_error):
        Settings()


def test_env_nested_delimiter_in_field(env):
    """Test pydantic fails to set value where the field has the delimiter.

    Pydantic will break down `with_delimiter_value` to
    `{delimiter: {value: ...}}` and it will pass it to `settings.with_delimiter`,
    so it will fail to populate `settings.with_delimiter.value` and it will try 
    to set `settings.with_delimiter.delimiter`.
    """

    class SubSettings(pydantic.BaseSettings):
        value: str

    class Settings(pydantic.BaseSettings):
        with_delimiter: SubSettings

        class Config:
            env_nested_delimiter = "_"

    env["with_delimiter_value"] = "won't be set"

    expected_error = (
        # settings.with_delimiter.value doesn't get set
        r".*with_delimiter -> value.*\n?.*value_error\.missing.*\n?"
        # Tries to set settings.with_delimiter.delimiter
        r".*with_delimiter -> delimiter.*\n?.*value_error\.extra"
    )

    with pytest.raises(pydantic.ValidationError, match=expected_error):
        Settings()
```
In my opinion this behaviour is wrong. Pydantic should either support this or raise an error that the class is wrongly configured. But the way it behaves right now is confusing since there's no error of wrong class setup, but the fields don't get set as expected from the environment variables.

Here's a possible solution on adding support for allowing to have the delimiter in both in the prefix and in the field name:

```py
class PrefixFixedEnvSettingsSource(EnvSettingsSource):
    def explode_env_vars(
        self, field: ModelField, env_vars: Mapping[str, Optional[str]]
    ) -> Dict[str, Any]:
        """
        Fix original explode_env_vars which splits the prefix with
        env_nested_delimiter. This implemetation doesn't split the prefix.
        """
        prefixes = [
            f"{env_name}{self.env_nested_delimiter}"
            for env_name in field.field_info.extra["env_names"]
        ]
        result: Dict[str, Any] = {}
        for env_name, env_val in env_vars.items():
            matching_prefixes = [
                prefix for prefix in prefixes if env_name.startswith(prefix)
            ]
            if not matching_prefixes:
                continue
            longest_prefix = max(len(prefix) for prefix in matching_prefixes)
            *keys, last_key = env_name[longest_prefix:].split(self.env_nested_delimiter)
            env_var = result
            for key in keys:
                env_var = env_var.setdefault(key, {})
            env_var[last_key] = env_val

        return result
```

arsenron:
## Change Summary
Hi! It is my first pull request to the project which I really, really appreciate :) The problem is when characters in ```env_prefix``` and ```env_nested_delimiter``` match, it occurs an unexpected behaviour as the initial splitting was wrong in this case.

I added a new ```env_prefix``` field to ```EnvSettingsSource``` to fix the issue without parsing ```ModelField``` which I think can create a little "code mess".

## Related issue number

fix #3807

## Checklist

### DIFF
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -63,6 +63,7 @@ class BaseSettings(BaseModel):
             env_nested_delimiter=(
                 _env_nested_delimiter if _env_nested_delimiter is not None else self.__config__.env_nested_delimiter
             ),
+            env_prefix=self.__config__.env_prefix,
         )
         file_secret_settings = SecretsSettingsSource(secrets_dir=_secrets_dir or self.__config__.secrets_dir)
         # Provide a hook to set built-in sources priority and add / remove sources
@@ -142,14 +143,19 @@ class InitSettingsSource:
 
 
 class EnvSettingsSource:
-    __slots__ = ('env_file', 'env_file_encoding', 'env_nested_delimiter')
+    __slots__ = ('env_file', 'env_file_encoding', 'env_nested_delimiter', 'env_prefix')
 
     def __init__(
-        self, env_file: Optional[StrPath], env_file_encoding: Optional[str], env_nested_delimiter: Optional[str] = None
+        self,
+        env_file: Optional[StrPath],
+        env_file_encoding: Optional[str],
+        env_nested_delimiter: Optional[str] = None,
+        env_prefix: str = '',
     ):
         self.env_file: Optional[StrPath] = env_file
         self.env_file_encoding: Optional[str] = env_file_encoding
         self.env_nested_delimiter: Optional[str] = env_nested_delimiter
+        self.env_prefix: str = env_prefix
 
     def __call__(self, settings: BaseSettings) -> Dict[str, Any]:  # noqa C901
         """
@@ -228,7 +234,8 @@ class EnvSettingsSource:
         for env_name, env_val in env_vars.items():
             if not any(env_name.startswith(prefix) for prefix in prefixes):
                 continue
-            _, *keys, last_key = env_name.split(self.env_nested_delimiter)
+            env_name_without_prefix = env_name[len(self.env_prefix) :]
+            _, *keys, last_key = env_name_without_prefix.split(self.env_nested_delimiter)
             env_var = result
             for key in keys:
                 env_var = env_var.setdefault(key, {})
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -234,6 +234,7 @@ class EnvSettingsSource:
         for env_name, env_val in env_vars.items():
             if not any(env_name.startswith(prefix) for prefix in prefixes):
                 continue
+            # we remove the prefix before splitting in case the prefix has characters in common with the delimiter
             env_name_without_prefix = env_name[self.env_prefix_len :]
             _, *keys, last_key = env_name_without_prefix.split(self.env_nested_delimiter)
             env_var = result
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -63,7 +63,7 @@ class BaseSettings(BaseModel):
             env_nested_delimiter=(
                 _env_nested_delimiter if _env_nested_delimiter is not None else self.__config__.env_nested_delimiter
             ),
-            env_prefix=self.__config__.env_prefix,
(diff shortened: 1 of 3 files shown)


pydantic/env_settings.py
```diff
@@ -228,7 +234,8 @@ def explode_env_vars(self, field: ModelField, env_vars: Mapping[str, Optional[st
         for env_name, env_val in env_vars.items():
             if not any(env_name.startswith(prefix) for prefix in prefixes):
                 continue
-            _, *keys, last_key = env_name.split(self.env_nested_delimiter)
+            env_name_without_prefix = env_name[len(self.env_prefix) :]
```
samuelcolvin, line 237: can you add a comment here explaining why we're doing this?

pydantic/env_settings.py
```diff
@@ -142,14 +143,19 @@ def __repr__(self) -> str:
...
-    __slots__ = ('env_file', 'env_file_encoding', 'env_nested_delimiter')
+    __slots__ = ('env_file', 'env_file_encoding', 'env_nested_delimiter', 'env_prefix')
 
     def __init__(
-        self, env_file: Optional[StrPath], env_file_encoding: Optional[str], env_nested_delimiter: Optional[str] = None
+        self,
+        env_file: Optional[StrPath],
+        env_file_encoding: Optional[str],
+        env_nested_delimiter: Optional[str] = None,
+        env_prefix: str = '',
```
samuelcolvin, line 153:
```suggestion
        env_prefix_len: int = 0,
```
Might be clearer if we passed the length, not the string here.

samuelcolvin:
please update.

pydantic/env_settings.py
arsenron: Sure. We are splitting an environment variable key into `keys` and `last_key` by `self.env_nested_delimiter` and in case of matching characters in both `self.env_nested_delimiter` and `self.env_prefix`, we were adding to `keys` an environment variable prefix. This was before, but now we just strip a prefix and can safely split by a delimeter because there will be no any more conflicts.

pydantic/env_settings.py
arsenron: Agree

pydantic/env_settings.py
samuelcolvin: No, when I say "please add a comment", I mean add a comment in the code. I know why we're doing this, but we comment the code to make it easier for someone reading this in future (including me when I've forgotten this conversation).

arsenron: Got it, sorry for misunderstanding

PR #3977, opened 2022-04-05
Fix #1458 - Allow for custom parsing of environment variables via env_parse

### LINKED ISSUE #1458: Settings: Custom parsing of environment variables (non-JSON)
# Settings: Custom parsing of environment variables (non-JSON)

Often we want complex environment variables that are not represented as JSON. One example is a list of items. It's not uncommon to comma-delimit lists like this in bash:

```python
import os
from typing import List

os.environ['options'] = "a,b,c"

class Settings(BaseSettings):
    options: List

s = Settings()
```

This results in a `JSONDecodeError`. Writing a list of items as valid json is error prone and not human friendly to read in the context of environment variables.

### Workaround

One workaround is to store the variable as valid json, which is tricky to type correctly in lots of systems where you have to enter environment variables:

```
OPTIONS='["a","b","c"]'
```

Another (simplified) workaround is to set `json_loads`. but its not very elegant since `json_loads` doesn't know what field its is parsing, which could be error prone:

```python
import json
import os
from typing import List

from pydantic import BaseSettings, BaseModel, root_validator

os.environ['options'] = "a,b,c"
    
def list_parse_fallback(v):
    try:
        return json.loads(v)
    except Exception as e:
        return v.split(",")

class Settings(BaseSettings):
    options: List
        
    class Config:
        json_loads = list_parse_fallback
       
s = Settings()
```

I can see a couple options for implementing the fix:

### 1. Store the parsing method in the field info extra:

```python
parse_func = lambda x: x.split(",")

class Settings(BaseSettings):
    options: List = Field(..., env_parse=parse_func)
```

If we take this approach, I think that we can update this logic branch:
https://github.com/samuelcolvin/pydantic/blob/42395056e18dfaa3ef299373374ab3b12bb196ac/pydantic/env_settings.py#L170-L174

Adding something like the following:

```python
if field.is_complex():
    if field.extra.get("env_parse", None) is not None:
        try:
            env_val = field.extra["env_parse"](env_val)  # type: ignore
        except ValueError as e:
            raise SettingsError(f'error with custom parsing function for "{env_name}"') from e
    else:
        try:
            env_val = self.__config__.json_loads(env_val)  # type: ignore
        except ValueError as e:
            raise SettingsError(f'error parsing JSON for "{env_name}"') from e
d[field.alias] = env_val
```

### 2. Add a new config option just for Settings for overriding how env vars are parsed
Another implementation option is to add a new property like `Settings.Config.parse_env_var` which takes the `field` and the value so that it can be overridden to handle dispatching to different parsing methods for different names/properties of `field` (currently, just overriding `json_loads` means you are passed a value without knowing where it will be stored so you have to test for and handle all of the possible settings values.

```python
class BaseSettings:
    ...
    class Config:
         ...
        @classmethod
        def parse_env_var(cls, field, raw_val):
            return cls.json_loads(raw_val)
```

Then the following line is the only change that is needed:
https://github.com/samuelcolvin/pydantic/blob/master/pydantic/env_settings.py#L62

Changes to `self.__config__.parse_env_var(field, env_val)`

### 3. Call field validators on the raw string instead of trying to load from json first

Change the same line to:
```
env_val, ee = field.validate(env_val)

# collect ee and raise errors
```

Pros:
 - Adding validators for fields is well documented / understood

Cons:
 - Breaks existing JSON functionality if those fields already have validators
 - Mixes up the abstractions in that the functions would now do parsing and validation

### 4. Custom (de)serialization

Let fields implement custom serialization/deserialization methods. Currently there is `json_encoders` but not an equivalent `json_decoders` for use per-field.

There's some discussion of this here: https://github.com/samuelcolvin/pydantic/issues/951

### 5. Something else

Other ideas? Happy to implement a different suggestion.

--------------

```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: /Users/bull/miniconda3/envs/sandbox/lib/python3.7/site-packages/pydantic
               python version: 3.7.6 (default, Jan  8 2020, 13:42:34)  [Clang 4.0.1 (tags/RELEASE_401/final)]
     optional deps. installed: []

```

acmiyaguchi:
## Change Summary

This PR adds an `env_parse` extra into Fields to allow for custom parsing of the environment within Settings. This is useful when specifying a custom format in environment strings e.g. lists (`1,2,3`) or key-value pairs (`a=1,b=2,c=3`). String-encoded JSON is not ideal in certain situations, such as when the environment is specified inside of a YAML file (leading to string-escaped string-encoded JSON). 

## Related issue number

fixes #1458

## Checklist

### DIFF
--- a/changes/3977-acmiyaguchi.md
+++ b/changes/3977-acmiyaguchi.md
@@ -1,0 +1,1 @@
+Allow for custom parsing of environment variables via `env_parse` in `Field`.
--- a/docs/examples/settings_with_custom_parsing.py
+++ b/docs/examples/settings_with_custom_parsing.py
@@ -1,0 +1,17 @@
+# output-json
+import os
+from typing import List
+
+from pydantic import BaseSettings, Field
+
+
+def parse_list(s: str) -> List[int]:
+    return [int(x.strip()) for x in s.split(',')]
+
+
+class Settings(BaseSettings):
+    numbers: List[int] = Field(env_parse=parse_list)
+
+
+os.environ['numbers'] = '1,2,3'
+print(Settings().dict())
--- a/docs/examples/settings_with_custom_parsing_validator.py
+++ b/docs/examples/settings_with_custom_parsing_validator.py
@@ -1,0 +1,17 @@
+# output-json
+import os
+from typing import List
+
+from pydantic import BaseSettings, Field, validator
+
+
+class Settings(BaseSettings):
+    numbers: List[int] = Field(env_parse=str)
+
+    @validator('numbers', pre=True)
+    def validate_numbers(cls, s):
+        return [int(x.strip()) for x in s.split(',')]
+
+
+os.environ['numbers'] = '1,2,3'
+print(Settings().dict())
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -99,16 +99,12 @@ Nested environment variables take precedence over the top-level environment vari
 You may also populate a complex type by providing your own parsing function to
 `env_parse` in the field extras.
 
-```py
 {!.tmp_examples/settings_with_custom_parsing.py!}
-```
 
 You might choose to pass the environment string value through to pydantic and
 transform in a validator instead.
 
-```py
 {!.tmp_examples/settings_with_custom_parsing_validator.py!}
-```
 
 
 ## Dotenv (.env) support
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -99,12 +99,12 @@ Nested environment variables take precedence over the top-level environment vari
 You may also populate a complex type by providing your own parsing function to
 `env_parse` in the field extras.
 
-{!.tmp_examples/settings_with_custom_parsing.py!}
+{!.tmp_examples/settings_with_custom_parsing.md!}
 
 You might choose to pass the environment string value through to pydantic and
 transform in a validator instead.
 
-{!.tmp_examples/settings_with_custom_parsing_validator.py!}
+{!.tmp_examples/settings_with_custom_parsing_validator.md!}
 
 
 ## Dotenv (.env) support
--- a/docs/usage/settings.md
+++ b/docs/usage/settings.md
@@ -87,7 +87,7 @@ export SUB_MODEL__DEEP__V4=v4
 You could load a settings module thus:
 {!.tmp_examples/settings_nested_env.md!}
 
-`env_nested_delimiter` can be configured via the `Config` class as shown above, or via the 
+`env_nested_delimiter` can be configured via the `Config` class as shown above, or via the
 `_env_nested_delimiter` keyword argument on instantiation.
 
 JSON is only parsed in top-level fields, if you need to parse JSON in sub-models, you will need to implement
@@ -95,6 +95,21 @@ validators on those models.
 
 Nested environment variables take precedence over the top-level environment variable JSON
 (e.g. in the example above, `SUB_MODEL__V2` trumps `SUB_MODEL`).
+
+You may also populate a complex type by providing your own parsing function to
+`env_parse` in the field extras.
+
+```py
+{!.tmp_examples/settings_with_custom_parsing.py!}
+```
+
+You might choose to pass the environment string value through to pydantic and
+transform in a validator instead.
+
+```py
+{!.tmp_examples/settings_with_custom_parsing_validator.py!}
+```
+
 
 ## Dotenv (.env) support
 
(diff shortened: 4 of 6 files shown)


acmiyaguchi:
please review

acmiyaguchi:
I appreciate the bump on the review.

> What about the case where I want to pre-process/parse a simple value? Let's say strip and sanitize a value during parsing. One could do that w/ this same mechanism if we extended the parse function to simple types as well [...].

I want to limit the scope of this PR to the related issue. If I understand correctly, you can use a validator to transform a simple value instead of using a custom parsing function (which is only necessary when trying to parse from an external source like the environment). While weird, minor modifications like stripping and sanitizing are idiomatic in Pydantic. I might be misunderstanding, though -- discussing the parent issue or a new issue might be in order.

samuelcolvin:
Looks good, but having reviewed the PR, I think this would be better deal with using suggestion 2 from the original PR: looking for a `parse_env_var` method on `Config`.

@acmiyaguchi do you think you could please update the PR?

acmiyaguchi:
Ack, I'll look into implementing the second method this week. Thanks for the review!

samuelcolvin:

Or if you need this in V1.10 but don't have time to complete it (or aren't the author), please comment here and on #4324.

acmiyaguchi:
Apologies for being last minute with implementation and discussion here, but I'd like to make a case against option 2:

> Looks good, but having reviewed the PR, I think this would be better deal with using suggestion 2 from the original PR: looking for a `parse_env_var` method on `Config`.

Suppose I had two fields in a setting class: one of them parses using `_parse_custom_dict: Callable[[str], Dict[int, str]]` and the other `_parse_custom_list: Callable[[str], List[str]]`. The current implementation in this PR has the following shape:

```python
class Settings(BaseSettings):
    my_list: List[str] = Field(env_parse=_parse_custom_list)
    my_dict: Dict[int, str] = Field(env_parse=_parse_custom_dict)
```

Whereas suggestion two take on the following look:

```python
class BaseSettings:
    my_list: List[str]
    my_dict: Dict[int, str]

    class Config:
        @classmethod
        def parse_env_var(cls, field, raw_val):
            if field == "my_list":
                return _parse_custom_list(raw_val)
            elif field == "my_dict":
                return _parse_custom_dict(raw_val)
            return cls.json_loads(raw_val)
```

I personally prefer the former solution in the context of the codebase where I'm currently using pydantic. I want to use a custom parsing function for lists and key-value pairs across several setting classes. Being forced to dispatch the parsing function via `parse_env_var` in the `Config` file is less flexible, because it requires knowing the names of the fields up front or doing some actual parsing to determine at runtime whether the environment variable is "listy" or "dicty". 

I believe the tests in the PR show a bit of flexibility in how parsing could be done with the option on the Field. An alternative to my first implementation above could be done by passing `str` as the parsing function and using a validator to transform the str into the appropriate value:

```python
class Settings(BaseSettings):
    my_list: List[str] = Field(env_parse=str)
    my_dict: Dict[int, str] = Field(env_parse=str)
    
    @validator('my_list', pre=True)
    def val_list(cls, v):
        assert isinstance(v, str)
        return _parse_custom_list(v)

    @validator('my_list', pre=True)
    def val_dict(cls, v):
        assert isinstance(v, str)
        return _parse_custom_dict(v)
```

It's a little verbose, but it seems to fit the style of pydantic.

acmiyaguchi:
I've implemented the alternative method via `Config.parse_env_var` in #4406. I've been using a workaround using a union type and custom validator, so there's no urgency on my end to make it into the next version of the package.

acmiyaguchi:
please review

samuelcolvin:
I think your implementation in #4406 is much better, I'm going to close this and accept that PR once a few small things are tweaked.

> Apologies for being last minute with implementation and discussion here, but I'd like to make a case against option 2:

Having thought about this I'm afraid I disagree. I get that `Field(env_parse=str)` could be ergonomic and useful, and I think it's something we should think about in V2 where settings might become a separate package, `pydantic-settings`.

The reasons I prefer `parse_env_var` are:
1. It's closer to the way the rest of pydantic works, in a minor release I want to make as few conceptual changes as possible
2. I think it's simpler to implement and therefore less likely to introduce bugs and breaking changes
3. If we do want per-field parsing, I think we should use something like the `Json` type, or #1848, not as a `Field` argument - unless we move to using `Field()` more for stuff like this in pydantic V2
4. With lots of fields on settings, I personally would prefer to have one function that takes care of parsing for multiple fields, rather than have to remember to set/change `Field.env_parse` on every field

---

Thanks so much for working on this, and for #4406. I'd love your help more when we get to redesign settings management in V2.

PR #4028, opened 2022-04-29
Add mypy type checked shallow and deep copy methods to BaseModel

thejaminator:
## Change Summary
This is a proposal and implementation for the mypy plugin to type check new `shallow_copy` and `deep_copy` methods.
If there is support for this PR , i will runtime validation into the methods as well. And update the docs.

### The Problem
Currently we don't have a nice way to typecheck a model field copying.

E.g.
```python
class Model(BaseModel):
    name: str

initial_model = Model(name="something")
new = initial_model.copy(update={"name_misspelled":123})
```
Mypy can't check if the field name and the type is correct in the update dict.

You can do
```python
new = initial_model.copy()
new.name = "somethingelse"
```
The assignment will be typechecked by the current mypy plugin. But what if your model has the config `frozen`?    
Then you can't assign in first place.

## Solution
This PR adds shallow_copy and deep_copy methods which the mypy plugin can type check. The user passes the field to update in as a keyword param.

These now fail with mypy errors
```python
# error: Unexpected keyword argument "weq" for "shallow_copy" of "MyModel"  [call-arg]
Model(name='foo').shallow_copy(weq=1)
# try_copy.py:24: error: Argument "name" to "shallow_copy" of "MyModel" has incompatible type "int"; expected "str"  [arg-type]
Model(name='foo').shallow_copy(name=1)
```

## Issues
Currently you will get a warning about overriding the parent's copy() which disobeys Liskov Substitution. Figuring out how to disable this

## Checklist


samuelcolvin:
@PrettyWood what do you think? I believe you're plan is to remove `copy` in V2? If so I guess we shouldn't add this now.

PrettyWood:
Yep. IMO better to close this and rely on stdlib and magic (aka dunder) methods

samuelcolvin:
Thanks for your contribution @thejaminator, however since we're going to remove `copy` in V2, I don't think it makes sense to add these checks in the last minor release of V1.

PR #4066, opened 2022-05-11
fix corner case for NoneType in __root__

mike0sv:
## Change Summary

I faced this edge case where I need to use a Model with `NoneType` for `__root__` field. It failed validation since `None` value is not allowed for this field. I fixed this check to allow it.

## Checklist

### DIFF
--- a/changes/4066-mike0sv.md
+++ b/changes/4066-mike0sv.md
@@ -1,0 +1,1 @@
+Fix edge case with `NoneType` as `__root__` field type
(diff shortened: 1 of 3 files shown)


mike0sv:
please review

samuelcolvin:
Please can you explain why this is necessary. I'm inclined to refuse the change unless many people are facing this problem.

mike0sv:
We need this because our tool has its own type system, and those types are translated to pydantic models to set up openapi schema for fastapi. We create models dynamically and recursively with `create_model` and it works for containers (`create_model(..., __root__=(List[...], ...))`) and for all the primitives (`create_model(..., __root__=(int, ...))`) except for the NoneType. I doubt lot of people will have the same issue, but it feels wrong that `__root__: int` works and `__root__: NoneType` does not

Issue #3829, opened 2022-02-18
Memory leak after updating to Pydantic v1.9.0


MarkusSintonen:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
$ python -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.9.0
            pydantic compiled: True
                 install path: /opt/pydist/.venv/lib/python3.10/site-packages/pydantic
               python version: 3.10.1 (main, Dec 21 2021, 09:01:08) [GCC 10.2.1 20210110]
                     platform: Linux-5.10.93-x86_64-with-glibc2.31
     optional deps. installed: ['dotenv', 'typing-extensions']
```

**As a heads-up we are seeing a memory leak in production after upgrading to Pydantic version 1.9.0.** Our application is a large FastAPI based API service (`fastapi 0.73.0, uvicorn 0.17.4`). The application uses Pydantic `BaseModel`s, `GenericModel`s and `pydantic.dataclasses.dataclass`es for FastAPI response/request data models. It also uses Pydantic models for some internal data validators. We deployed a single change to production, which changed the version of Pydantic from 1.8.2 to 1.9.0. After that we started to observe abnormal memory usage. The memory usage increased until the version upgrade got reverted back to the older 1.8.2 version (no other change again). After deploying the revert, the memory usage patterns went back to normal.

Unfortunately I haven't been able to reproduce the issue locally (it is an effort to simulate the prod-like load to get the leak to show up). So I can not yet add further information from isolating it or information from eg. memory analyzers. I tried to also look into the changelog and diff but its so big release that I can not spot anything obvious that may cause this. I have attached below a picture from our production metrics about the abnormal behaviour after using v1.9.0. We have never observed such a memory issue until now when upgrading to v1.9.0.

Thank you all for the awesome work and the great library! I wish I would have more information about the leak for the ticket.

<img width="1235" alt="Näyttökuva 2022-2-18 kello 8 48 24" src="https://user-images.githubusercontent.com/12939780/154632303-d7fa908a-1c66-4004-9ce4-66c996220354.png">
[...]

PR #4083, opened 2022-05-16
limit the length of `generics._limit_assigned_parameters`

samuelcolvin:
## Change Summary

As per discussion in #3829 this should avoid memory "leaks" when generating lots of generics.

## Related issue number

#3829 - I won't mark this as fixed until confirmed by users.

## Checklist

### DIFF
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -24,23 +24,20 @@ from .fields import DeferredType
 from .main import BaseModel, create_model
 from .types import JsonWrapper
 from .typing import display_as_type, get_all_type_hints, get_args, get_origin, typing_base
-from .utils import all_identical, lenient_issubclass
+from .utils import LimitedDict, all_identical, lenient_issubclass
 
 GenericModelT = TypeVar('GenericModelT', bound='GenericModel')
 TypeVarType = Any  # since mypy doesn't allow the use of TypeVar as a type
 
 Parametrization = Mapping[TypeVarType, Type[Any]]
 
-_generic_types_cache: Dict[Tuple[Type[Any], Union[Any, Tuple[Any, ...]]], Type[BaseModel]] = {}
-_generic_types_cache_limit = 1000
+_generic_types_cache: LimitedDict[Tuple[Type[Any], Union[Any, Tuple[Any, ...]]], Type[BaseModel]] = LimitedDict()
 # _assigned_parameters is a Mapping from parametrized version of generic models to assigned types of parametrizations
 # as captured during construction of the class (not instances).
 # E.g., for generic model `Model[A, B]`, when parametrized model `Model[int, str]` is created,
 # `Model[int, str]`: {A: int, B: str}` will be stored in `_assigned_parameters`.
 # (This information is only otherwise available after creation from the class name string).
-_assigned_parameters: Dict[Type[Any], Parametrization] = {}
-# _assigned_parameters size is limited by _limit_assigned_parameters() below
-_assigned_parameters_limit = 1000
+_assigned_parameters: LimitedDict[Type[Any], Parametrization] = LimitedDict()
 
 
 class GenericModel(BaseModel):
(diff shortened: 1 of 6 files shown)


pydantic/generics.py
```diff
@@ -38,6 +38,8 @@
 # `Model[int, str]`: {A: int, B: str}` will be stored in `_assigned_parameters`.
 # (This information is only otherwise available after creation from the class name string).
 _assigned_parameters: Dict[Type[Any], Parametrization] = {}
+# _assigned_parameters size is limited by _limit_assigned_parameters() below
+_assigned_parameters_max_length = 1000
```
MarkusSintonen, line 42: If `_assigned_parameters` is a kind of cache we can limit in size, could we instead also use `lru_cache` for `__parameterized_bases__`?:
```
@classmethod
@lru_cache(maxsize=1024)
def __parameterized_bases__(cls, typevars_map: Parametrization) -> List[Type[Any]]:
```
And then remove the `_assigned_parameters` "cache"?

MarkusSintonen, line 42: If `_assigned_parameters` is a kind of cache we can limit in size, could we instead also use `lru_cache` for `__parameterized_bases__`?:
```
@classmethod
@lru_cache(maxsize=1024)
def __parameterized_bases__(cls, typevars_map: Parametrization) -> List[Type[Any]]:
```
And then remove the `_assigned_parameters` "cache"?

samuelcolvin: Yes we could, in fact that's how I had it on the first commit. But I wanted to keep the interface identical (in case people are introspecting `_assigned_parameters`) and also allow people to change the limit. 

Hence this approach.

samuelcolvin: Sorry, I realise I never pushed the second commit, should make more sense now.

pydantic/generics.py
```diff
@@ -358,3 +363,16 @@ def _prepare_model_fields(
         field.outer_type_ = concrete_type
         field.prepare()
         created_model.__annotations__[key] = concrete_type
+
+
+def _limit_cache_size(cache_dict: Dict[Any, Any], size_limit: int) -> None:
+    """
+    Limit the size/length of a dict used for caching to avoid unlimited increase in memory usage.
+
+    Since the dict is ordered, and we always remove elements from the beginning, this is effectively an LRU cache.
```
samuelcolvin, line 372:
```suggestion
    Since the dict is ordered, and we always remove elements from the beginning, this is effectively a
    FIFO cache.
```

pydantic/generics.py
```diff
@@ -131,6 +135,7 @@ def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[T
         _generic_types_cache[(cls, params)] = created_model
         if len(params) == 1:
             _generic_types_cache[(cls, params[0])] = created_model
+            _limit_cache_size(_generic_types_cache, _generic_types_cache_limit)
```
davidhyman, line 138: possibly an alternative solution that doesn't require maintainers to remember to call this cache limiter every time the dict is modified, whilst recalling the reason for avoiding lru_cache - 

could the cache limiting be implemented directly on the dict? presumably as a subclass of dict (or userdict) so that anyone relying on introspection of the dict would see the same api

i.o.w override `__setitem__` ?

samuelcolvin: good idea, though I wonder how many people are really using `_generic_types_cache`, through probably still worth doing for the same of correctness.

samuelcolvin: Done, see c04cd8e.

samuelcolvin: Well that was much slower and uglier than I had hoped, cython complained, but only on CI (I couldn't replicate the exception locally on Ubuntu) so I had to stop inheriting from `dict` or `UserDict`.

samuelcolvin: :disappointed: this is very annoying, last try.

PR #4084, opened 2022-05-17
Support cross-process pickling of Pydantic-wrapped built-in dataclasses

### LINKED ISSUE #3453: dataclass conversion breaks multiprocessing pickling
### Checks

# Bug

```
             pydantic version: 1.8.2
            pydantic compiled: True
                 install path: /home/kazesberger/.cache/pypoetry/virtualenvs/pydantic-pickling-uaEvjRfL-py3.8/lib/python3.8/site-packages/pydantic
               python version: 3.8.10 (default, Sep 28 2021, 16:10:42)  [GCC 9.3.0]
                     platform: Linux-5.11.0-38-generic-x86_64-with-glibc2.29

```

```py
import dataclasses
import multiprocessing

import pydantic


@dataclasses.dataclass()   # using pydantic.dataclasses.dataclass() works
class Foo:
    bar: int


class SomeModel(pydantic.BaseModel):
    baaz: int
    foo: Foo


def do_it(q: multiprocessing.Queue):
    model = q.get()
    print(model)

def main():
    multiprocessing.set_start_method(method="spawn")
    q = multiprocessing.Queue()
    p = multiprocessing.Process(target=do_it, args=(q,))
    p.start()
    model = SomeModel(baaz=0, foo=Foo(bar=42))
    q.put(model)
    q.put(None)
    p.join()

if __name__ == '__main__':
    main()

```

```console
  File "/usr/lib/python3.8/multiprocessing/queues.py", line 116, in get
    return _ForkingPickler.loads(res)
AttributeError: Can't get attribute '_Pydantic_Foo_41343024' on <module 'pydantic.dataclasses' from '/home/kazesberger/.cache/pypoetry/virtualenvs/pydantic-pickling-uaEvjRfL-py3.8/lib/python3.8/site-packages/pydantic/dataclasses.cpython-38-x86_64-linux-gnu.so'>
```

iammaart:
## Change Summary

Pull request https://github.com/samuelcolvin/pydantic/pull/2114 has added support for picking built-in dataclasses that are used as fields in Pydantic models (or ones that are directly wrapped by `pydantic.dataclasses.dataclass`). However, the currently implemented solution only works within a single Python process.

The problem is that Pydantic dynamically creates wrapper classes for the built-in dataclasses and assigns names to them based on `id(origninal_class)`. These generated names will be different for each invocation of the Python interpreter. In a multiprocessing setting, or even for different runs of the same code, pickle won't be able to find the right class to unpickle to as a result.

This change adds support for cross-process pickling by making the generated class names deterministic.

As explained in [my analysis of issue #3453](https://github.com/samuelcolvin/pydantic/issues/3453#issuecomment-1126065386), pickling will still not work properly across different processes in some scenarios. This could either be considered a separate issue, or this PR is only a partial fix for https://github.com/samuelcolvin/pydantic/issues/3453.

## Related issue number

Fix #3453

## Checklist

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -106,6 +106,27 @@ def _generate_pydantic_post_init(
     return _pydantic_post_init
 
 
+def _generate_builtin_dataclass_wrapper_name(cls: Type[Any]) -> str:
+    """
+    Generate a name for the class that wraps a built-in dataclass that is:
+    - a valid Python identifier (usable as attribute name),
+    - unique (no collisions with different classes),
+    - deterministic (so it works across different processes), and
+    - recognisable as wrapper for the dataclass (to ease debugging).
+    """
+
+    # Assumption: the combination of `__module__` and `__qualname__` attributes
+    # uniquely identifies the class. This should be a safe assumption because pickle
+    # uses these two attributes to find the class.
+    fully_qualified_cls_name = f'{cls.__module__}.{cls.__qualname__}'
+
+    # We hex-encode the fully qualified name of the source class because it contains
+    # one or more period (.) characters, making it unusable as identifier.
+    unique_id = fully_qualified_cls_name.encode('utf-8').hex()
+
+    return f'_Pydantic_{cls.__name__}_{unique_id}'
+
+
 def _process_class(
     _cls: Type[Any],
     init: bool,
@@ -140,7 +161,7 @@ def _process_class(
     # and register it on module level to address pickle problem:
     # https://github.com/samuelcolvin/pydantic/issues/2111
     if is_builtin_dataclass(_cls):
-        uniq_class_name = f'_Pydantic_{_cls.__name__}_{id(_cls)}'
+        uniq_class_name = _generate_builtin_dataclass_wrapper_name(_cls)
         _cls = type(
             # for pretty output new class will have the name as original
             _cls.__name__,
--- a/changes/4084-iammaart.md
+++ b/changes/4084-iammaart.md
@@ -1,0 +1,1 @@
+Support cross-process pickling of built-in dataclasses wrapped by Pydantic
--- a/tests/test_dataclasses.py
+++ b/tests/test_dataclasses.py
@@ -1,6 +1,8 @@
 import dataclasses
+import multiprocessing
 import pickle
 from collections.abc import Hashable
+from concurrent.futures import ProcessPoolExecutor
 from datetime import datetime
 from pathlib import Path
 from typing import Callable, ClassVar, Dict, FrozenSet, List, Optional, Union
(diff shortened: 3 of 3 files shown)


iammaart:
Please review

iammaart:
In case you're wondering why a few commits from `master` got rewritten in an earlier version of this pull request: I redacted my email address from the commit history after I creating the PR. In my initial attempt, I didn't scope the `git filter-branch` correctly. That's been fixed now.

maestro-1 (commented):
This looks pretty good. It might be worth taking out a few of the comments. Those could have easily been added in the commit message as opposed to just sitting on the code base, especially since your function docstring also contains explanations of what you are attempting to achieve.
The comments in the test cases are super useful, though, so definitely leave those in

maestro-1 (approved):

elfjes:
This is great, I would love to see this merged. We're working on providing windows support for our software and this is currently hindering us. On posix we can use forked Processes, but that's unfortunately not an option on Windows.

samuelcolvin:
I'll get to this soon.

FWIW, on @maestro-1's review: I think commenting like this is absolutely great, it's true that (in theory) you can have "too many comments", but that's not the case here. Clarity wins - we don't have to print the code base.

pydantic/dataclasses.py
```diff
@@ -106,6 +106,27 @@ def _pydantic_post_init(self: 'Dataclass', *initvars: Any) -> None:
...
+    """
+
+    # Assumption: the combination of `__module__` and `__qualname__` attributes
+    # uniquely identifies the class. This should be a safe assumption because pickle
+    # uses these two attributes to find the class.
+    fully_qualified_cls_name = f'{cls.__module__}.{cls.__qualname__}'
+
+    # We hex-encode the fully qualified name of the source class because it contains
+    # one or more period (.) characters, making it unusable as identifier.
+    unique_id = fully_qualified_cls_name.encode('utf-8').hex()
```
samuelcolvin, line 125: do we really need to use hex and make the name (arguably) harder to read?

Could we use just `f'{cls.__module__}ᚑ{cls.__qualname__}'`? I know that means in theory that we could get clashes, but it seems vanishingly unlike unless someone is trying to clash.

(the `ᚑ` is a arbitarily selected high order unicode character which is [valid in identifiers](https://gist.github.com/samuelcolvin/68f56c96bf18b03bb8a3ffa988380966), we could use something else if you prefer)

samuelcolvin (commented):
mostly me being pedantic, and suggesting a crazy unicode value, as I'm [somewhat fascinated](https://twitter.com/samuel_colvin/status/1472283581087158273) by their use in identifiers.

Otherwise LGTM.

samuelcolvin:
please update.

samuelcolvin:
also conflicts.

Issue #3641, opened 2022-01-08
copy_on_model_validation config doesn't copy List field anymore in v1.9.0


NunchakusLei:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

The output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"` when I use v1.9.0a1,
```
             pydantic version: 1.9.0a1
            pydantic compiled: True
                 install path: <*my-project-dir*>/venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.9 (main, Nov 16 2021, 03:05:18)  [GCC 7.5.0]
                     platform: Linux-5.4.0-91-generic-x86_64-with-glibc2.27
     optional deps. installed: ['typing-extensions']
```
The output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"` when I use v1.9.0,
```
             pydantic version: 1.9.0
            pydantic compiled: True
                 install path: <*my-project-dir*>/venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.9 (main, Nov 16 2021, 03:05:18)  [GCC 7.5.0]
                     platform: Linux-5.4.0-91-generic-x86_64-with-glibc2.27
     optional deps. installed: ['typing-extensions']
```

Looks like there is accidental behaviour change from v1.9.0a1 to v1.9.0 regarding the `copy_on_model_validation` config.

```py
from pydantic import BaseModel
from typing import List


class A(BaseModel):
    name: str
    
    class Config:
        copy_on_model_validation = True  # Default behaviour
        

class B(BaseModel):
    list_a: List[A]


a = A(name="a")
b = B(list_a=[a])

print(a)
print(b)
a.name = "b"  # make a change to variable a's field
print(a)
print(b)

```
The output using v1.9.0a1,
```
A(name='a')
B(list_a=[A(name='a')])
A(name='b')
B(list_a=[A(name='a')])
```

The output using v1.9.0 changed to,
```
A(name='a')
B(list_a=[A(name='a')])
A(name='b')
B(list_a=[A(name='b')])
```

NunchakusLei:
This issue happen again in the latest version 1.9.2.

PR #4093, opened 2022-05-19
allow for shallow copies

### LINKED ISSUE #4092: 1.9.1 Breaking change with model deep copying
### Checks

# Bug
https://github.com/samuelcolvin/pydantic/pull/3642/files
The above PR changes the way models are copied and has created a breaking change in our code base. When running our test suite we now get the following error:

```log
INTERNALERROR>   File "/Users/worx/ape/src/ape_test/providers.py", line 19, in __init__
INTERNALERROR>     super().__init__(**data)
INTERNALERROR>   File "pydantic/main.py", line 339, in pydantic.main.BaseModel.__init__
INTERNALERROR>   File "pydantic/main.py", line 1038, in pydantic.main.validate_model
INTERNALERROR>   File "pydantic/fields.py", line 857, in pydantic.fields.ModelField.validate
INTERNALERROR>   File "pydantic/fields.py", line 1074, in pydantic.fields.ModelField._validate_singleton
INTERNALERROR>   File "pydantic/fields.py", line 1121, in pydantic.fields.ModelField._apply_validators
INTERNALERROR>   File "pydantic/class_validators.py", line 313, in pydantic.class_validators._generic_validator_basic.lambda12
INTERNALERROR>   File "pydantic/main.py", line 679, in pydantic.main.BaseModel.validate
INTERNALERROR>   File "pydantic/main.py", line 605, in pydantic.main.BaseModel._copy_and_set_values
INTERNALERROR>   File "/Users/worx/.pyenv/versions/3.9.2/lib/python3.9/copy.py", line 146, in deepcopy
INTERNALERROR>     y = copier(x, memo)
INTERNALERROR>   File "/Users/worx/.pyenv/versions/3.9.2/lib/python3.9/copy.py", line 230, in _deepcopy_dict
INTERNALERROR>     y[deepcopy(key, memo)] = deepcopy(value, memo)
INTERNALERROR>   File "/Users/worx/.pyenv/versions/3.9.2/lib/python3.9/copy.py", line 151, in deepcopy
INTERNALERROR>     copier = getattr(x, "__deepcopy__", None)
INTERNALERROR>   File "/Users/worx/ape/src/ape/api/networks.py", line 183, in __getattr__
INTERNALERROR>     return self.get_network(network_name)
INTERNALERROR>   File "/Users/worx/ape/src/ape/api/networks.py", line 320, in get_network
INTERNALERROR>     raise NetworkNotFoundError(network_name)
INTERNALERROR> ape.exceptions.NetworkNotFoundError: No network named '--deepcopy--'.
```
Pinning to 1.9.0 as current fix.

```
pydantic version: 1.9.1
            pydantic compiled: True
                 install path: /Users/worx/.pyenv/versions/3.9.2/envs/ape/lib/python3.9/site-packages/pydantic
               python version: 3.9.2 (default, Dec  2 2021, 09:57:33)  [Clang 13.0.0 (clang-1300.0.29.3)]
                     platform: macOS-12.3.1-arm64-arm-64bit
```

```py
import pydantic

...
```

timkpaine:
## Related issue number

Fix #4092.

## Checklist


timkpaine:
please review

samuelcolvin:
Thanks so much. I'm just moving house at the moment, but hopefully will find time to review this in the next few days.

timkpaine:
@samuelcolvin no worries! pinning 1.9.0 is a workaround for now (or setting `Config.copy_on_model_validation = False`, though the behavior is somewhat different).

pydantic/config.py
```diff
@@ -62,8 +62,11 @@ class BaseConfig:
     json_encoders: Dict[Union[Type[Any], str, ForwardRef], AnyCallable] = {}
     underscore_attrs_are_private: bool = False
 
-    # whether inherited models as fields should be reconstructed as base model
+    # whether inherited models as fields should be reconstructed as base model,
+    # and whether such a copy should be shallow or deep
     copy_on_model_validation: bool = True
+    copy_on_model_validation_shallow: bool = False
```
PrettyWood, line 68: If we want this, I would rather switch to an enum for
`copy_on_model_validation` with options `'disable', 'shallow', 'deep'`

samuelcolvin: Or just a literal?

samuelcolvin:
You need to change the base for this PR to `1.9.X-fixes`, not master.

pydantic/config.py
timkpaine: string literal better than enum I would think

pydantic/config.py
```diff
@@ -36,6 +36,12 @@ class Extra(str, Enum):
     forbid = 'forbid'
 
 
+class Copy(str, Enum):
```
samuelcolvin, line 39: Breast to use a `Literal` not an enum. Also empty string should not be an option.

pydantic/main.py
```diff
@@ -675,7 +675,9 @@ def __get_validators__(cls) -> 'CallableGenerator':
     @classmethod
     def validate(cls: Type['Model'], value: Any) -> 'Model':
         if isinstance(value, cls):
-            if cls.__config__.copy_on_model_validation:
+            if cls.__config__.copy_on_model_validation == Copy.shallow:
```
samuelcolvin, line 678: We should raise a warning for booleans.

samuelcolvin (commented):
Also need to add something to the documentation.

pydantic/main.py
```diff
@@ -675,9 +675,18 @@ def __get_validators__(cls) -> 'CallableGenerator':
     @classmethod
     def validate(cls: Type['Model'], value: Any) -> 'Model':
         if isinstance(value, cls):
-            if cls.__config__.copy_on_model_validation:
+            if isinstance(cls.__config__.copy_on_model_validation, bool):
+                # Warn about deprecated behavior
+                warnings.warn('`copy_on_model_validation` should be a string of value ("deep", "shallow", "none")', RuntimeWarning)
+
+            if cls.__config__.copy_on_model_validation == "shallow":
```
AngellusMortis, line 682: This is introducing another breaking change. `copy_on_model_validation` being a bool should restore the pre-1.9.1 behavior. 

```suggestion
            if isinstance(cls.__config__.copy_on_model_validation, bool):
                # Warn about deprecated behavior
                warnings.warn('`copy_on_model_validation` should be a string of value ("deep", "shallow", "none")', RuntimeWarning)

            if cls.__config__.copy_on_model_validation == "shallow" or cls.__config__.copy_on_model_validation is True:
```

samuelcolvin:
At this stage, I think most people have worked around this, I think best we leave behaviour as it is.

AngellusMortis:
We have not worked around it in Home Assistant. Instead, we blacklisted this release since it is a breaking change. Introducing breaking changes in a bugfix revision and then not fixing them is not a good way to handle versioning.

samuelcolvin:
I understand this isn't ideal @AngellusMortis, the breaking change wasn't intentional.

As far as I can tell, this change wasn't a simple fix for the original problem, it introduced new settings, also as you mentioned it introduced a new breaking change.

If you (or anyone else) can create a fix for the issue, I'll happily review it and try to get this fixed. Otherwise I'll try to get to it soon.

AngellusMortis:
Even if this PR is canceled, as long as it is fixed is the important part. Introducing a breaking change in a bugfix versions and then _not fixing it_ erodes the trust in the package. It would be good enough reason for me to consider an alternative solution as I do not want to use a package whose response to semver is "oh there was a breaking change, oh well, I am sure people will work around it".

timkpaine:
> If you (or anyone else) can create a fix for the issue
This PR was a compromise solution that kept the breaking change and added in functionality to revert to the old. Simply changing the functionality back will itself be a breaking change (no doubt released in a patch version), so I recommend not doing this and going with this PR instead.

samuelcolvin:
> I understand this isn't ideal 

Again, I'm really sorry about this. Finding time to work on pydantic when I'm not paid by a big company to do so has been hard, hence I've had to prioritise. 

Since I've been working on it full time, I've been mostly concentrating on preparing for V2. One of the big reason to rewrite pydantic from the ground up in V2 is to make the codebase easier to work with so we can reduce the chance of breaking changes in minor or patch releases. Again, **we try to avoid breaking changes as much as possible**.

AngellusMortis:
I personally though this PR was a great compromise so the feature could be kept without needing the breaking change. Just need the above suggestion to remove the new breaking change introduce by it.

samuelcolvin:
Okay, I've re-opened this PR.

@timkpaine any chance you could fix the above? And we'll try to get this fixed and released asap.

PrettyWood:
@AngellusMortis I come back from holidays on Thursday. I'll try to work on it before the end of the week.
You are right to say we should fix it and I'm sorry I introduced a breaking change. This was definitely not intended. But please remember that all of this is done for free on nights on some individuals ❤️

samuelcolvin:
@AngellusMortis is there a PR or issue you could link to on Home Assistant where this broke stuff?

In preparation for both V1.10 and V2 it would be good to have a list of open source projects which use pydantic where we can test functionality and thereby try to avoid more breaking changes?

AngellusMortis:
> @AngellusMortis is there a PR or issue you could link to on Home Assistant where this broke stuff?
> 
> In preparation for both V1.10 and V2 it would be good to have a list of open source projects which use pydantic where we can test functionality and thereby try to avoid more breaking changes?

I unfortunately do not. HA directly does not use pydantic. HA is large a combination of a lot of integrations with various other platforms. Two integrations, one I maintain for UniFi Protect and the Xbox integration, both make use of Pydantic. The UniFi Protect integration was adversely affected. Since I am a core maintainer on the library for the UniFi Protect integration, [I did not make a PR](https://github.com/briis/pyunifiprotect/commit/9efbedb6906b6b3bc230802780dee224ea36ca69) like I should have.

> But please remember that all of this is done for free on nights on some individuals

We are all doing this in our free time. That is open source. But having stable libraries is very important in choosing projects for open source since breaking changes like this also requires _me_ to spend more of my free time finding a workaround.

samuelcolvin:
> I am also now saying

"not saying"?

I accept it was a mistake to close this PR :bow:.

AngellusMortis:
I do that all the time. Sorry :joy:. now -> not always for me. Lol

PrettyWood:
FWIW @samuelcolvin I agree with @timkpaine that this PR has the right approach now (it is what I would have done https://github.com/pydantic/pydantic/pull/4093#discussion_r877953270)
In v2 we'll go with explicit calls of copy or deepcopy but in the meantime having this literal is probably the best to be 1.9.0 compliant

samuelcolvin:
Ok, please comment if there's anything you think needs changing and I'll
try to merge tomorrow and get a new release out.

Is there anything else which should go in v1.9.2?

peterschutt:
> In preparation for both V1.10 and V2 it would be good to have a list of open source projects which use pydantic where we can test functionality and thereby try to avoid more breaking changes?

One for your list - [starlite](https://github.com/starlite-api/starlite) is an API framework that heavily leverages pydantic.

pydantic/main.py
samuelcolvin: @AngellusMortis I think I've resolved this.

samuelcolvin:
Okay I think this is fixed and ready to be deployed.

@AngellusMortis @PrettyWood please can you review and confirm you're happy?

samuelcolvin:
> In preparation for both V1.10 and V2 it would be good to have a list of open source projects which use pydantic where we can test functionality and thereby try to avoid more breaking changes?

see #4359, moving discussion to there.

changes/4093-timkpaine.md
```diff
@@ -0,0 +1,3 @@
+Allow for shallow copies of attributes, adjusting the behavior of #3642
+`Config.copy_on_model_validation` is now a str enum of `["none", "deep", "shallow"]` corresponding to 
+not copying, deep copy, shallow copy.
```
PrettyWood, line 3: My main concern is on the default value

With the test example
```py
class User(BaseModel):
    hobbies: list[str]

my_user = User(hobbies=['scuba diving'])

class Transaction(BaseModel):
    user: User

t = Transaction(user=my_user)

print(t.user is my_user)
print(t.user.hobbies is my_user.hobbies)
```

we used to have in v1.9.0
```
copy_on_model_validation = True (default): False, True
copy_on_model_validation = False: True, True
```

which got changed (the famous breaking change) in 1.9.1

```
copy_on_model_validation = True (default): False, False <-💥
copy_on_model_validation = False: True, True
```

We now have with this patch
```
copy_on_model_validation = 'deep' : False, False
copy_on_model_validation = 'shallow': False, True
copy_on_model_validation = 'none': True, True
```

I wonder if we should write a note saying a breaking change has been introduced by mistake in a release version and that the old 1.9.0 behaviour has been fixed. In which case, the default value should be `'shallow'`.

PrettyWood (commented):
Except the default value and the note looks good! Thank you!

changes/4093-timkpaine.md
samuelcolvin: I agree, we should make the default `shallow`.

I'll try to fix this today.

samuelcolvin: default changed.

PrettyWood (approved):
LGTM! Thanks a lot
@samuelcolvin we'll need to add a note at the start of the release note for this issue and patch

samuelcolvin:
Thanks for the review, will do.

I'll do the release tomorrow.

NunchakusLei:
Looks like this PR reproduce the issue https://github.com/pydantic/pydantic/issues/3641

```python
from typing import List


class A(BaseModel):
    name: str
        

class B(BaseModel):
    list_a: List[A]


a = A(name="a")
b = B(list_a=[a])

print(a)
print(b)
a.name = "b"  # make a change to variable a's field
print(a)
print(b)
```

Output in 1.6.2, 1.7.4, 1.8.1, 1.8.2, and 1.9.1,
```
A(name='a')
B(list_a=[A(name='a')])
A(name='b')
B(list_a=[A(name='a')])
```

Output in 1.9.0 and 1.9.2,
```
A(name='a')
B(list_a=[A(name='a')])
A(name='b')
B(list_a=[A(name='b')])
```

Is this the expected change of behaviour from previous version to v1.9.x?

timkpaine:
@NunchakusLei yes, this PR restores state to some combination of 1.9.0 and 1.9.1 behavior, but lets you pick between them.

timkpaine:
@NunchakusLei from the docs:
```
Allow for shallow copies of model fields, Config.copy_on_model_validation is now a str which must be 'none', 'deep', or 'shallow' corresponding to not copying, deep copy & shallow copy; default 'shallow',
```

samuelcolvin:
See #4369, I need to update the docs.

NunchakusLei:
> @NunchakusLei yes, this PR restores state to some combination of 1.9.0 and 1.9.1 behavior, but lets you pick between them.
> 
> from the docs:
> 
> ```
> Allow for shallow copies of model fields, Config.copy_on_model_validation is now a str which must be 'none', 'deep', or 'shallow' corresponding to not copying, deep copy & shallow copy; default 'shallow',
> ```
> 
> So you'll want `deep` to restore the (erroneously changed) 1.9.1 behavior

Do you know if the 1.8.2 and previous are equivalent to `deep` behaviour? That's the version behaviour I try to restored to.

timkpaine:
pre 1.9.1 was default shallow or "none", 1.9.1 was default deep or none, 1.9.2 is none/deep/shallow

NunchakusLei:
@timkpaine Base on my test https://github.com/pydantic/pydantic/pull/4093#issuecomment-1220778338, pre 1.9.0 was default deep or none (tested in 1.6.2, 1.7.4, 1.8.1, and 1.8.2). 

The test result conflicting the statement of,

> pre 1.9.1 was default shallow or "none" ...

I got very confuse now so want to clarify.

samuelcolvin:
On my phone but I think:
* Pre 1.9.0: default SHALLOW, config setting for NONE
* 1.9.0, 1.9.1: default DEEP, config setting for NONE
* 1.9.2: default SHALLOW, config settings for NONE or DEEP

NunchakusLei:
@samuelcolvin 

> * Pre 1.9.0: default SHALLOW, config setting for NONE

This doesn't matched my test result in 1.6.2, 1.7.4, 1.8.1, and 1.8.2. https://github.com/pydantic/pydantic/pull/4093#issuecomment-1220778338

timkpaine:
@NunchakusLei your tests results might be wrong, but also...it doesnt really matter at this point, all behaviors are supported now.

samuelcolvin:
> @samuelcolvin
> 
> > * Pre 1.9.0: default SHALLOW, config setting for NONE
> 
> This doesn't matched my test result in 1.6.2, 1.7.4, 1.8.1, and 1.8.2. [#4093 (comment)](https://github.com/pydantic/pydantic/pull/4093#issuecomment-1220778338)

See: https://github.com/pydantic/pydantic/blob/v1.8.2/pydantic/main.py#L731 pretty sure I'm correct.

efisoft-elt:
```python
class M(BaseModel):
    class Config:
        copy_on_model_validation = "!! Some non sens str !!"
```

Does not raise any error.  This is probably due to the deprecation warning .

```python
            if copy_on_model_validation not in {'deep', 'shallow', 'none'}:
                # Warn about deprecated behavior
                warnings.warn(
                    "`copy_on_model_validation` should be a string: 'deep', 'shallow' or 'none'", DeprecationWarning
                )
                if copy_on_model_validation:
                    deep_copy = False

```

PR #4152, opened 2022-06-11
Add default to JSON schema when const is True

aminalaee:
## Change Summary

Right now for the following Model:

```python

class MyModel(BaseModel):
    x: int = Field(default=0, const=True)
```

The JSON schema is:

```json
{
  "title": "MyModel",
  "type": "object",
  "properties": {
    "x": {
      "title": "X",
      "const": 0,
      "type": "integer"
    }
  }
}
```

Which will include the `default`:

```json
{
  "title": "MyModel",
  "type": "object",
  "properties": {
    "x": {
      "title": "X",
      "const": 0,
      "default": 0,
      "type": "integer"
    }
  }
}
```

Fixes 

## Related issue number

Fixes https://github.com/samuelcolvin/pydantic/issues/4031

## Checklist

### DIFF
--- a/changes/4031-aminalaee.md
+++ b/changes/4031-aminalaee.md
@@ -1,0 +1,1 @@
+Add `default` value in JSON Schema when `const=True`
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -210,12 +210,7 @@ def get_field_info_schema(field: ModelField, schema_overrides: bool = False) -> 
         schema_['description'] = field.field_info.description
         schema_overrides = True
 
-    if (
-        not field.required
-        and not field.field_info.const
-        and field.default is not None
-        and not is_callable_type(field.outer_type_)
-    ):
+    if not field.required and field.default is not None and not is_callable_type(field.outer_type_):
         schema_['default'] = encode_default(field.default)
         schema_overrides = True
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -585,12 +585,14 @@ def test_const_list():
         'properties': {
             'a': {
                 'const': [SubModel(b=1), SubModel(b=2), SubModel(b=3)],
+                'default': [{'b': 1}, {'b': 2}, {'b': 3}],
                 'items': {'$ref': '#/definitions/SubModel'},
                 'title': 'A',
                 'type': 'array',
             },
             'b': {
                 'const': [{'b': 4}, {'b': 5}, {'b': 6}],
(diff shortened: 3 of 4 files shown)


aminalaee:
please review

hramezani:
> I couldn't find the reason why it was introduced in https://github.com/samuelcolvin/pydantic/pull/2094 but seems like the main purpose was something else.

The `and not field.field_info.const` was introduced originally in [Implement const keyword in Schema](https://github.com/samuelcolvin/pydantic/commit/8149e0933238773b78e3046ea16b23d7e03efe0b) and just reformatted in https://github.com/samuelcolvin/pydantic/pull/2094. IMO, seems the original implementation was wrong.

hramezani (approved):

PrettyWood:
IMO the `const` keyword should be deprecated in favor of `Literal`. With `Literal`, is the behaviour the expected one?

aminalaee:
@PrettyWood Not entirely sure, but they will produce different schemas which seem equivalent 🤷 

with `const` it will include: `"const": 10` and with `Literal` it will include `"enum": [10]`

aminalaee:
@PrettyWood 

Not sure if I understand correctly but this will not include a `default` either:

```python
class Test(BaseModel):
    x: Literal[10]
```

```json
{"title": "Test", "type": "object", "properties": {"x": {"title": "X", "enum": [10], "type": "integer"}}, "required": ["x"]}
```

But this will which I'm not sure if is correct:

```python
class Test(BaseModel):
    x: Literal[10] = Field(default=10)
```

```json
{"title": "Test", "type": "object", "properties": {"x": {"title": "X", "default": 10, "enum": [10], "type": "integer"}}}
```

aminalaee:
And about the `const` deprecation, `const=True` will produce `"const": 10` in schema but `Literal[10]` will produce `"enum": [10]` which should serve the same purpose but they are both possible according the the json schema, I could not find any references to the difference though.

samuelcolvin:
> IMO the `const` keyword should be deprecated in favor of `Literal`. With `Literal`, is the behaviour the expected one? edit: I don't say we should deprecate it now or that this PR should not be merged! :P

I think what @PrettyWood means is that `Literal[10]` (just one option in the literal), should result in `"const": 10` in the schema? I think he just means we should get rid of the kwarg from `Field` (at least that's what I think he means :raised_eyebrow:)

samuelcolvin:
thanks so much for this.

PR #4155, opened 2022-06-12
Fix schema for set and frozenset with default value

aminalaee:
## Change Summary

Fixes JSON schema for `set` and `frozenset` when they have default values, the output json is not valid and cannot be serialized.

For example:

```python
import json

class Test(BaseModel):
    x = {1}

print(Test.schema())
```

It will produce:

```json
{
  "title": "Test",
  "type": "object",
  "properties": {
    "x": {
      "title": "X",
      "type": "array",
      "default": {1},
      "items": {
        "type": "integer"
      },
      "uniqueItems": true
    }
  },
  "required": [
    "x"
  ]
}
```

But instead should be:

```json
{
  "title": "Test",
  "type": "object",
  "properties": {
    "x": {
      "title": "X",
      "default": [
        1,
      ],
      "type": "array",
      "items": {},
      "uniqueItems": true
    }
  }
}
```

## Related issue number

Fixes https://github.com/samuelcolvin/pydantic/issues/4136

## Checklist

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -77,7 +77,7 @@ from .typing import (
     is_none_type,
     is_union,
 )
-from .utils import ROOT_KEY, get_model, lenient_issubclass, sequence_like
+from .utils import ROOT_KEY, get_model, lenient_issubclass
 
 if TYPE_CHECKING:
     from .dataclasses import Dataclass
@@ -984,7 +984,7 @@ def encode_default(dft: Any) -> Any:
         return dft.value
     elif isinstance(dft, (int, float, str)):
         return dft
-    elif sequence_like(dft):
+    elif isinstance(dft, (list, tuple)):
         t = dft.__class__
         seq_args = (encode_default(v) for v in dft)
         return t(*seq_args) if is_namedtuple(t) else t(seq_args)
--- a/changes/4155-aminalaee.md
+++ b/changes/4155-aminalaee.md
@@ -1,0 +1,1 @@
+Fix JSON schema for `set` and `frozenset` when they include default values.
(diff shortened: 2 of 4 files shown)


aminalaee:
please review

pydantic/schema.py
```diff
@@ -984,7 +984,7 @@ def encode_default(dft: Any) -> Any:
         return dft.value
     elif isinstance(dft, (int, float, str)):
         return dft
-    elif sequence_like(dft):
+    elif isinstance(dft, (list, tuple)):
```
LordFckHelmchen, line 987: I'm really not into the code base - so this is just a non-constructive critisism, but this change would narrow the scope compared to the original sequence_like check. What happens when these instances (set, frozenset, ...) are passed in?

aminalaee: I'm also very new to the codebase, but I think the `sequence_like` usage is not very accurate here.

If you look further, after these conditions it will try to use the pydantic_encoder for all unhandled types which in this case is set and frozenset.
And also there are tests for all cases and new ones for set and frozenset, so I think this is not narrowing down any type, just setting up different encoders for these types.

LordFckHelmchen: sounds reasonable! now we only need @samuelcolvin or @PrettyWood to approve :-)

hramezani:
Thanks @aminalaee for this patch :+1: 

I think we need to add `frozenset` to the documentation about [The field schema mapping from Python / pydantic](https://pydantic-docs.helpmanual.io/usage/schema/#json-schema-types)

hramezani:
please update

aminalaee:
please review

hramezani (approved):

samuelcolvin:
Thanks so much @aminalaee.

PR #4179, opened 2022-06-22
Copy edit model_config docs

bennuttall:
I noticed a missing closing bracket, then realised that line was different to the others, so I made them all the same.


bennuttall:
Maybe links with the text "dedicated section" should be updated to use titles

lukehsiao (approved):
Came here to do this same thing.

hramezani (approved):

PR #4187, opened 2022-06-27
Catch Exceptions in smart_deepcopy

### LINKED ISSUE #4184: smart_deepcopy causes sqlalchemy boolean value error on import
### Checks

# Bug

```
root@20116b25f897:/[REDACTED] # python -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.8.2
            pydantic compiled: False
                 install path: /usr/local/lib/python3.10/site-packages/pydantic
               python version: 3.10.5 (main, Jun  7 2022, 18:49:47) [GCC 10.2.1 20210110]
                     platform: Linux-5.10.104-linuxkit-x86_64-with-glibc2.31
```

```py
"""Base class for all database models."""

from typing import Any, Dict, Optional
from uuid import UUID, uuid4

import sqlalchemy as sa
import sqlalchemy.dialects.postgresql as pg
from inflection import underscore
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import declared_attr
from sqlalchemy_utils import Timestamp

from pydantic import BaseModel as _BaseModel

class CommonBase(_BaseModel, Timestamp):
    """SQLAlchemy Base common base class for all database models."""

    id: UUID = sa.Column(pg.UUID(as_uuid=True), primary_key=True, default=uuid4)

    @declared_attr
    def __tablename__(cls) -> Optional[str]:  # pylint: disable=no-self-argument
        """Define table name for all models as the snake case of the model's name."""
        return underscore(cls.__name__)

    def as_dict(self) -> Dict[str, Any]:
        """Return python dictionary of SQLAlchemy model."""
        return self.__dict__

BaseModel = declarative_base(cls=CommonBase)
```

On import of this BaseModel, we are getting a hard to track down error that seems to be related to sqlalchemy boolean expressions, but stemming from the smart_deepcopy line in pydantic and I am not quite sure what the real issue is, and whether it is caused by pydantic, sqlalchemy, or my setup (most likely).

The stack trace received on import is:

```
app/models/base_model.py:15: in <module>
    class CommonBase(_BaseModel, Timestamp):
/usr/local/lib/python3.10/site-packages/pydantic/main.py:299: in __new__
/usr/local/lib/python3.10/site-packages/pydantic/fields.py:411: in infer
    return cls(
/usr/local/lib/python3.10/site-packages/pydantic/fields.py:342: in __init__
    self.prepare()
/usr/local/lib/python3.10/site-packages/pydantic/fields.py:445: in prepare
    self._set_default_and_type()
/usr/local/lib/python3.10/site-packages/pydantic/fields.py:473: in _set_default_and_type
    default_value = self.get_default()
/usr/local/lib/python3.10/site-packages/pydantic/fields.py:345: in get_default
    return smart_deepcopy(self.default) if self.default_factory is None else self.default_factory()
/usr/local/lib/python3.10/site-packages/pydantic/utils.py:627: in smart_deepcopy
    elif not obj and obj_type in BUILTIN_COLLECTIONS:
/usr/local/lib/python3.10/site-packages/sqlalchemy/sql/elements.py:582: in __bool__
    raise TypeError("Boolean value of this clause is not defined")
E   TypeError: Boolean value of this clause is not defined
```

I placed a debugger just before the raise and found that:

```
(Pdb) self
Column(None, UUID(as_uuid=True), table=None, primary_key=True, nullable=False, default=ColumnDefault(<function uuid4 at 0x7f99d085e200>))
(Pdb) self.__dict__
{'key': None, 'name': None, 'table': None, 'type': UUID(as_uuid=True), 'is_literal': False, 'primary_key': True, '_user_defined_nullable': symbol('NULL_UNSPECIFIED'), 'nullable': False, 'default': ColumnDefault(<function uuid4 at 0x7f99d085e200>), 'server_default': None, 'server_onupdate': None, 'index': None, 'unique': None, 'system': False, 'doc': None, 'onupdate': None, 'autoincrement': 'auto', 'constraints': set(), 'foreign_keys': set(), 'comment': None, 'computed': None, 'identity': None, '_creation_order': 88, 'comparator': <sqlalchemy.sql.type_api.TypeEngine.Comparator object at 0x7f99d07f0a00>, '_memoized_keys': frozenset({'comparator'})}
```

This is not new code in pydantic, and it also seems this check in sqlalchemy was [added almost a decade ago](https://docs.sqlalchemy.org/en/14/changelog/migration_06.html) so I am sure that is must be something wrong with either my expectatins or setup here. While investigating, it seems that this indeed is the correct way to setup a postgres UUID pk column within a pydantic model backed by sqlalchemy, but I am clearly missing something.

Thanks for the incredible work you do on this library!

coneybeare:
## Change Summary

This PR comes from an issue where a `sqlalchemy` column definition was raising when `not obj` was called in `smart_deepcopy`. Whether or not the raise is warranted is a question for `sqlalchemy`, but catching the error here and falling back to `deepcopy` on seems to be a decent approach as a raise of this nature prevents application import and load.

## Related issue number

Closes #4184 

## Checklist

* [X] Tests pass on CI and coverage remains at 100% (coverage is 99.08, but only because of unrelated files missed coverage)

### DIFF
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -113,13 +113,6 @@ BUILTIN_COLLECTIONS: Set[Type[Any]] = {
     deque,
 }
 
-# There are error classes caught in smart_deepcopy
-CAUGHT_SMART_DEEPCOPY_ERRORS: Set[Type[Exception]] = {
-    TypeError,
-    ValueError,
-    RuntimeError,
-}
-
 
 def import_string(dotted_path: str) -> Any:
     """
@@ -662,7 +655,7 @@ def smart_deepcopy(obj: Obj) -> Obj:
         if not obj and obj_type in BUILTIN_COLLECTIONS:
             # faster way for empty collections, no need to copy its members
             return obj if obj_type is tuple else obj.copy()  # type: ignore  # tuple doesn't have copy method
-    except tuple(CAUGHT_SMART_DEEPCOPY_ERRORS):
+    except (TypeError, ValueError, RuntimeError):
         # do we really dare to catch ALL errors? Seems a bit risky
         pass
 
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -113,6 +113,13 @@ BUILTIN_COLLECTIONS: Set[Type[Any]] = {
     deque,
 }
 
+# There are error classes caught in smart_deepcopy
+CAUGHT_SMART_DEEPCOPY_ERRORS: Set[Type[Exception]] = {
+    TypeError,
+    ValueError,
+    RuntimeError,
+}
+
 
 def import_string(dotted_path: str) -> Any:
     """
@@ -651,9 +658,14 @@ def smart_deepcopy(obj: Obj) -> Obj:
     obj_type = obj.__class__
     if obj_type in IMMUTABLE_NON_COLLECTIONS_TYPES:
         return obj  # fastest case: obj is immutable and not collection therefore will not be copied anyway
-    elif not obj and obj_type in BUILTIN_COLLECTIONS:
-        # faster way for empty collections, no need to copy its members
-        return obj if obj_type is tuple else obj.copy()  # type: ignore  # tuple doesn't have copy method
+    try:
+        if not obj and obj_type in BUILTIN_COLLECTIONS:
+            # faster way for empty collections, no need to copy its members
+            return obj if obj_type is tuple else obj.copy()  # type: ignore  # tuple doesn't have copy method
+    except tuple(CAUGHT_SMART_DEEPCOPY_ERRORS):
+        # do we really dare to catch ALL errors? Seems a bit risky
+        pass
+
     return deepcopy(obj)  # slowest way when we actually might need a deepcopy
 
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -27,7 +27,6 @@ from pydantic.typing import (
 )
 from pydantic.utils import (
     BUILTIN_COLLECTIONS,
-    CAUGHT_SMART_DEEPCOPY_ERRORS,
     ClassAttribute,
     LimitedDict,
     ValueItems,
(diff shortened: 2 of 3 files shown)


coneybeare:
please review @samuelcolvin

pydantic/utils.py
```diff
@@ -113,6 +113,13 @@
     deque,
 }
 
+# There are error classes caught in smart_deepcopy
+CAUGHT_SMART_DEEPCOPY_ERRORS: Set[Type[Exception]] = {
```
samuelcolvin, line 117: I don't think you need to define this separately, just hard code it in the except block.

tests/test_utils.py
```diff
@@ -461,6 +462,18 @@ def test_smart_deepcopy_collection(collection, mocker):
...
 
+@pytest.mark.parametrize('error', CAUGHT_SMART_DEEPCOPY_ERRORS)
+def test_smart_deepcopy_error(error, mocker):
+    class RaiseOnBooleanOperation(str):
+        def __bool__(self):
+            raise error('raised error')
+
+    obj = RaiseOnBooleanOperation()
+    expected_value = object()
+    mocker.patch('pydantic.utils.deepcopy', return_value=expected_value)
```
samuelcolvin, line 473: why do we need to patch this? I don't think it's necessary.

pydantic/utils.py
```diff
@@ -651,9 +658,14 @@ def smart_deepcopy(obj: Obj) -> Obj:
...
         return obj  # fastest case: obj is immutable and not collection therefore will not be copied anyway
-    elif not obj and obj_type in BUILTIN_COLLECTIONS:
-        # faster way for empty collections, no need to copy its members
-        return obj if obj_type is tuple else obj.copy()  # type: ignore  # tuple doesn't have copy method
+    try:
+        if not obj and obj_type in BUILTIN_COLLECTIONS:
+            # faster way for empty collections, no need to copy its members
+            return obj if obj_type is tuple else obj.copy()  # type: ignore  # tuple doesn't have copy method
+    except tuple(CAUGHT_SMART_DEEPCOPY_ERRORS):
+        # do we really dare to catch ALL errors? Seems a bit risky
```
samuelcolvin, line 666: I think what you have here is a good start.

pydantic/utils.py
samuelcolvin: then manually parameterise the test to the same exception types.

samuelcolvin (commented):
Looking good, just one thing to change.

samuelcolvin:
please update.

pydantic/utils.py
coneybeare: https://github.com/pydantic/pydantic/pull/4187/commits/624a0aa0be4a15923907501d90e516e6e5b778d4

tests/test_utils.py
coneybeare: https://github.com/pydantic/pydantic/pull/4187/commits/624a0aa0be4a15923907501d90e516e6e5b778d4

coneybeare: We were mocking out the underlying method (not under test) to assert `is`. I removed, called the underlying method, and now am using `==` as the test because the real response is a different in-memory object.

samuelcolvin:
great, thanks so much.

In future, it's safest to use the magic comment "please review" to assign the PR to me and make sure I know the PR needs my attention.

PR #4194, opened 2022-06-29
Fix for bug in root_validator value propogation with pre=True

jacquelinegarrahan:
## Change Summary

During validation on assignment at present, the value returned from a root_validator with pre=True is not propagated and assigned. The reason for this is that the known_field validators operate on the original value.

```python

class M(BaseModel):

    x: int = 1

    class Config:
        validate_assignment = True

    @root_validator(pre=True)
    def validate_all(cls, values):
        print('root validator', values)
        values["x"] = 99  # Always hijack x
        print('done', values)
        return values    


m = M(x=3)
m.x = None  # ValidationError: 1 validation error for M
```

This PR adds a single line updating the value with the output of the pre root_validator.

## Checklist

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -363,6 +363,7 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
             for validator in self.__pre_root_validators__:
                 try:
                     new_values = validator(self.__class__, new_values)
+                    value = new_values[name]
                 except (ValueError, TypeError, AssertionError) as exc:
                     raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], self.__class__)
 
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -363,7 +363,6 @@ class BaseModel(Representation, metaclass=ModelMetaclass):
             for validator in self.__pre_root_validators__:
                 try:
                     new_values = validator(self.__class__, new_values)
-                    value = new_values[name]
                 except (ValueError, TypeError, AssertionError) as exc:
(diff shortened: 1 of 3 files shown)


jacquelinegarrahan:
Please review

hramezani:
Thanks @jacquelinegarrahan for this patch :+1: 

I am not sure if validating the changed value is the right behaviour or not because we explicitly set the value in the assignment.
On the other hand, we have this behavior (changing the assigned value) in the normal validator. for example:

```python

class M(BaseModel):

    x: int = 1

    class Config:
        validate_assignment = True

    @validator('x')
    def validate_x(cls, v):
        return 4

m = M(x=3)
print('after init x = ', m.x)  # after init x =  4
m.x = 5
print('after assignment x = ', m.x)  # after assignment x =  4
```

What do you think @samuelcolvin and @PrettyWood?

pydantic/main.py
```diff
@@ -363,6 +363,7 @@ def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
             for validator in self.__pre_root_validators__:
                 try:
                     new_values = validator(self.__class__, new_values)
+                    value = new_values[name]
```
hramezani, line 366: You can change line 378 and pass directly new value there:
```python
 value, error_ = known_field.validate(new_values[name], dict_without_original_value, loc=name, cls=self.__class__)
```

samuelcolvin:
Thanks for the contribution.

I think this behaviour is right, but the change in behaviour is too big for a minor release, let's wait for V2.

PR #4224, opened 2022-07-10
Fix `AnyUrl.build` doesn't do percent encoding (#3061)

### LINKED ISSUE #3061: *Url.build functions doesn't percent-encode arguments
### Checks

# Bug
Using (AnyUrl, HttpUrl, AnyHttpUrl) build functions results an unusable URL if characters like: `:`, `/`, `?`, `#`, `[`, ` `, `]`, `@` exists in the following arguments: `user`, `password`, `path`, `query`, `fragment`.

In other words these arguments should be [Percent-Encoded](https://tools.ietf.org/html/rfc3986#section-2.1) to work properly and not
produce a Url that have characters that conflicts with the [url-spec](https://www.w3.org/Addressing/URL/url-spec.txt).

```
             pydantic version: 1.8.2
            pydantic compiled: True
                 install path: /.../.venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.6 (default, Jul  3 2021, 17:50:42)  [GCC 7.5.0]
                     platform: Linux-5.4.0-80-generic-x86_64-with-glibc2.27
```

# Example

```py
>>> from pydantic import AnyUrl, HttpUrl, AnyHttpUrl
>>>
>>> AnyUrl.build(
...    scheme="ws",
...    user="@AwesomeUser",
...    password="Pa##w0rd?:/",
...    host="localhost",
...    port="5000"
...)
'ws://@AwesomeUser:Pa##w0rd?:/@localhost:5000'
>>> # ...
>>> # The same result with `HttpUrl`, `AnyHttpUrl`
```

The easiest way to percent-encode in Python is to use the `urllib.parse.quote` function

FaresAhmedb:
## Change Summary

- Add `percent_encode` function to `pydantic.utils`
- Add tests for `percent_encode`
- Fix `AnyUrl.build`
- Add new tests for `AnyUrl.build`

## Related issue number

fix #3061

## Checklist

* [x]  Documentation reflects the changes where applicable, *not applicable*


FaresAhmedb:
please review

pydantic/utils.py
```diff
@@ -787,3 +787,29 @@ def __setitem__(self, __key: Any, __value: Any) -> None:
         def __class_getitem__(cls, *args: Any) -> Any:
             # to avoid errors with 3.7
             pass
+
+
+URI_RESERVED_CHARACTERS = frozenset(":/?#[]@!$&'()*+,;=% ")
+
+
+def percent_encode(string: str, *, plus: bool = False) -> str:
```
samuelcolvin, line 795: what's wrong with `urllib.parse.quote`?

pydantic/utils.py
```diff
@@ -787,3 +787,29 @@ def __setitem__(self, __key: Any, __value: Any) -> None:
...
+        return string
+
+    string = string.replace('%', f'%{ord("%"):02X}')
+
+    if plus:
+        string = string.replace('+', f'%{ord("+"):02X}')
+        string = string.replace(' ', '+')
+        characters = URI_RESERVED_CHARACTERS.symmetric_difference('+% ')
+    else:
+        characters = URI_RESERVED_CHARACTERS.symmetric_difference('%')
```
samuelcolvin, line 809: surely we can just remove `%` from `URI_RESERVED_CHARACTERS` and remove this line.

samuelcolvin (commented):
Looks good in principle, can we get some benchmarks to see how much this slows down validation

samuelcolvin:
please update.

pydantic/utils.py
FaresAhmedb: I just didn't feel like introducing a new import to pydantic since `urllib.prase` is a big module which can slow down importing pydantic but now after testing with `-X importtime`, it seems that there's no difference since `pathlib` already imports `urllib.parse`.

Here's some benchmarks for both functions tested on Python 3.10.5, let me know what you think

```
$ ./percent_encode_urllib_parse_benchmark 
short(16ch):
pydantic.utils.percent_encode: 50000 loops, best of 5: 4.4 usec per loop
urllib.parse.quote:            50000 loops, best of 5: 2.53 usec per loop

long(256ch):
pydantic.utils.percent_encode: 50000 loops, best of 5: 12.1 usec per loop
urllib.parse.quote:            50000 loops, best of 5: 18 usec per loop

large(4096ch):
pydantic.utils.percent_encode: 50000 loops, best of 5: 77.9 usec per loop
urllib.parse.quote:            50000 loops, best of 5: 255 usec per loop
```

```
$ ./percent_encode_urllib_parse_benchmark --plus
short(16ch):
pydantic.utils.percent_encode: 50000 loops, best of 5: 5.39 usec per loop
urllib.parse.quote:            50000 loops, best of 5: 2.59 usec per loop

long(256ch):
pydantic.utils.percent_encode: 50000 loops, best of 5: 15 usec per loop
urllib.parse.quote:            50000 loops, best of 5: 17.9 usec per loop

large(4096ch):
pydantic.utils.percent_encode: 50000 loops, best of 5: 79.5 usec per loop
urllib.parse.quote:            50000 loops, best of 5: 260 usec per loop
```

```python
#!/usr/bin/env python

import random
import string
import sys
import timeit

TIMEIT_ARGV = ["-n", "50000", "-u", "usec"]

if "--plus" not in sys.argv:
    quote_timeit_argv = TIMEIT_ARGV + ["--setup", "from urllib.parse import quote"]
    percent_encode_timeit_argv = TIMEIT_ARGV + ["--setup", "from pydantic.utils import percent_encode"]
else:
    quote_timeit_argv = TIMEIT_ARGV + ["--setup", "from urllib.parse import quote_plus as quote"]
    percent_encode_timeit_argv = TIMEIT_ARGV + [
        "--setup",
        ";".join(
            [
                "from functools import partial",
                "from pydantic.utils import percent_encode",
                "percent_encode = partial(percent_encode, plus=True)",
            ]
        ),
    ]


def benchmark(string: str) -> None:
    print("pydantic.utils.percent_encode: ", end="")
    timeit.main(percent_encode_timeit_argv + [f"percent_encode({string!r}, plus=True)"])
    print("urllib.parse.quote:            ", end="")
    timeit.main(quote_timeit_argv + [f"quote({string!r})"])


SHORT_STRING = "".join(random.choices(string.printable, k=16))
LONG_STRING = "".join(random.choices(string.printable, k=256))
LARGE_STRING = "".join(random.choices(string.printable, k=4096))

print(f"short({len(SHORT_STRING)}ch):")
benchmark(SHORT_STRING)

print(f"\nlong({len(LONG_STRING)}ch):")
benchmark(LONG_STRING)

print(f"\nlarge({len(LARGE_STRING)}ch):")
benchmark(LARGE_STRING)
```

Here's also a plot for strings with the length 1-24 to the power of 2, lower is bettter:

percent_encode_urllib_parse_benchmark_plot

FaresAhmedb:
> Looks good in principle, can we get some benchmarks to see how much this slows down validation

I'm not sure how this affects validation or how to benchmark pydantic's validation, can you assist me on this?

samuelcolvin:
Thanks so much for the benchmarks, I think we should go with `from urllib.parse import quote_plus` for the following reasons:
* keep it simple, minimise our own logic
* I would guess most URLs will be short enough that's `quote_plus` will be marginally quicker

> I'm not sure how this affects validation or how to benchmark pydantic's validation, can you assist me on this?

I think we can skip this, we have to get this right, so there's no option of not quote encoding.

The real question long terms is whether we should move this logic to rust, but provided the API doesn't change we could do that in a minor release after V2.

FaresAhmedb:
Okay, I removed `pydantic.utils.percent_encode` in the latest commit.

> The real question long terms is whether we should move this logic to rust, but provided the API doesn't change we could do that in a minor release after V2.

I think this should be moved to rust since `urllib.parse` is just pure python.

I will drop some benchmarks later for `urllib.prase.quote` against [percent-encoding](https://crates.io/crates/percent-encoding) and [urlencoding](https://crates.io/crates/urlencoding)

pydantic/networks.py
```diff
@@ -153,6 +154,7 @@ def __init__(
         path: Optional[str] = None,
         query: Optional[str] = None,
         fragment: Optional[str] = None,
+        plus: bool = False,
```
samuelcolvin, line 157:
```suggestion
        quote_plus: bool = False,
```
might be a clearer name.

changes/3061-FaresAhmedb.md
```diff
@@ -0,0 +1 @@
+Fix `AnyUrl.build` doesn't do percent encoding
```
samuelcolvin, line 1:
```suggestion
Add percent encoding in `AnyUrl` and descendent types
```

pydantic/networks.py
samuelcolvin: Also, is there a good reason not to have `quote_plus = True` as the default?

pydantic/networks.py
```diff
@@ -192,20 +195,23 @@ def build(
             **_kwargs,  # type: ignore[misc]
         )
 
+        do_quote = quote_plus if plus else quote
```
samuelcolvin, line 198: please can you add this as a method on the class so it can be used externally.

pydantic/networks.py
samuelcolvin: Please add `quote_plus` to `stricturl`

samuelcolvin (commented):
Looking good, my main concerns are:
* this could in theory be considered a breaking change, I think it's fine because it's pretty clear than without this the built URLs are "wrong", @PrettyWood do you agree?
* there's no way to set `quote_plus` on a URL field, other than `stricturl` - maybe that's fine?

please can you add to the docs:
* An example of this in action
* an example with `stricturl` of using `quote_plus=False` (or whatever the non-default is)
* a "note" admonition saying "percent encoding was added in V1.10 ..."

Please update.

samuelcolvin:
Also, @FaresAhmedb please remember to use the magic comment "please  review" (with one space) to request a review, otherwise we might miss updates on the PR.

pydantic/networks.py
FaresAhmedb: > Also, is there a good reason not to have `quote_plus = True` as the default?

Just conforming to the standard ([RFC 3986](https://www.ietf.org/rfc/rfc3986.txt))

FaresAhmedb:
please review

docs/usage/types.md
```diff
@@ -723,6 +724,20 @@ _(This script is complete, it should run "as is")_
...
     (or at least big) company.
 
+#### Building URLs
+
+You can build URLs from seperate [URL Properties](#url-properties) using the `build` method in [Pydantic URL types](#urls) or any type that inherits from them; see [Custom Data Types](#custom-Data-Types).
+
+By default *pydantic* percent encodes the following URL properties: `user`, `password`, `path`, `query` as per [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) without replacing spaces with `+` but this can be changed using the `stricturl` method:
+
+!!! note
+    Percent encoding was added in V2.0
```
samuelcolvin, line 734:
```suggestion
    Percent encoding was added in V1.10
```
We don't want to wait for V2 for this!

docs/usage/types.md
```diff
@@ -723,6 +724,20 @@ _(This script is complete, it should run "as is")_
...
+
+```py
+{!.tmp_examples/types_url_building.py!}
+```
+_(This script is complete, it should run "as is")_
```
samuelcolvin, line 739: can you rebase and change this to just 
```suggestion
{!.tmp_examples/types_url_building.py!}
```
To match the recent docs changes.

samuelcolvin (commented):
Looking good, just a few small things here, but you need to fix linting in the docs so build succeeds and we can check the docs look right.

Please update.

FaresAhmedb:
please review

samuelcolvin:
Thanks so much.

PR #4249, opened 2022-07-18
Resolve ForwardRef on Field.outer_type_

JacobHayes:
## Change Summary

When resolving `ForwardRef`s, the `Field.outer_type_` is not resolved, only `Field.type_`. This PR updates `update_field_forward_refs` to resolve both attributes, while still only calling `Field.prepare` once.

`Field.outer_type_` can be a nested object (eg: `List[str]`, etc), but I only check if the _entire_ annotation is a `ForwardRef` for simplicity. This should cover most cases, namely:
- all `from __future__ import annoations` annotations
- manually stringized annotations as long as the _whole_ annotation is stringized (eg: `"List[RecursiveType]"` will be resolved, `List["RecursiveType"]` will not)

I think I've seen [other instances](https://github.com/samuelcolvin/pydantic/discussions/4185) where `ForwardRefs` inside nested types aren't full resolved, so 🤞 this is fine for now.

## Checklist

### DIFF
--- a/changes/4249-JacobHayes.md
+++ b/changes/4249-JacobHayes.md
@@ -1,0 +1,1 @@
+Update `ForwardRef`s in `Field.outer_type_`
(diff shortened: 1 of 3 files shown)


samuelcolvin:
thanks so much.

PR #4252, opened 2022-07-19
Empty string is a valid JSON-key

### LINKED ISSUE #4253: An empty string is a valid JSON attribute name
### Checks

* [ x] I added a descriptive title to this issue
* [ x] I have searched (google, github) for similar issues and couldn't find anything
* [ x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

```
             pydantic version: 1.9.0
            pydantic compiled: False
                 install path: /Users/s_tsaplin/projects/github.com/sergeytsaplin/pydantic/pydantic
               python version: 3.9.13 (main, May 19 2022, 14:10:54)  [Clang 13.1.6 (clang-1316.0.21.2)]
                     platform: macOS-12.4-x86_64-i386-64bit
```

```py

class Model(BaseModel):
        empty_string_key: int = Field(alias='')

data = {'': 123}
m = Model(**data)
m.dict(by_alias=True)
```

## Current behavior:

The code causes the exception:

```
  File "<masked>/.pyenv/versions/3.10.1/lib/python3.10/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
empty_string_key
 ```

## Expected result:

The code should print:

```
{'': 123}
```

Links:
* https://stackoverflow.com/questions/58916957/is-an-empty-string-a-valid-json-key
* https://datatracker.ietf.org/doc/html/rfc6901#section-5

### PR DESCRIPTION
**Pull request by SergeyTsaplin**
## Change Summary

An empty string is a valid JSON attribute name, so aliases checking cannot be just like `bool(alias)` it should be strictly compared with `None`.

## Related issue number

fix #4253

## Checklist

### DIFF
--- a/changes/4253-sergeytsaplin.md
+++ b/changes/4253-sergeytsaplin.md
@@ -1,1 +1,1 @@
-Field aliases are compared with None instead of simple `bool(alias)` because an empty string is a valid JSON attribute name.
+Allow empty string aliases by using a `alias is not None` check, rather than `bool(alias)`
--- a/changes/4253-sergeytsaplin.md
+++ b/changes/4253-sergeytsaplin.md
@@ -1,0 +1,1 @@
+Field aliases are compared with None instead of simple `bool(alias)` because an empty string is a valid JSON attribute name.
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -148,7 +148,7 @@ class FieldInfo(Representation):
         self.default = default
         self.default_factory = kwargs.pop('default_factory', None)
         self.alias = kwargs.pop('alias', None)
-        self.alias_priority = kwargs.pop('alias_priority', 2 if self.alias else None)
+        self.alias_priority = kwargs.pop('alias_priority', 2 if self.alias is not None else None)
         self.title = kwargs.pop('title', None)
         self.description = kwargs.pop('description', None)
         self.exclude = kwargs.pop('exclude', None)
@@ -395,7 +395,7 @@ class ModelField(Representation):
 
         self.name: str = name
         self.has_alias: bool = bool(alias)
-        self.alias: str = alias or name
+        self.alias: str = alias if alias is not None else name
         self.type_: Any = convert_generics(type_)
         self.outer_type_: Any = type_
         self.class_validators = class_validators or {}
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -394,7 +394,7 @@ class ModelField(Representation):
     ) -> None:
 
         self.name: str = name
-        self.has_alias: bool = bool(alias)
+        self.has_alias: bool = alias is not None
         self.alias: str = alias if alias is not None else name
         self.type_: Any = convert_generics(type_)
         self.outer_type_: Any = type_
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by samuelcolvin** [comment]
Thanks for the contribution.

While I think you're technically correct. I'm not sure I would support changing this at this late stage for pydantic V1.

I believe this is fixed in pydantic-core (the basis for pydantic V2) - if you're willing, it would be very useful to add unit tests for this to [pydantic-core](https://github.com/samuelcolvin/pydantic-core)?

[The plan](https://pydantic-docs.helpmanual.io/blog/pydantic-v2/) is to get v1.10 released with mostly cosmetic changes and some new features that having been hanging around for a long time, then move ahead with V2.

What do others think? @PrettyWood @hramezani

**Comment by hramezani** [comment]
Thanks @SergeyTsaplin for the patch :+1: and Thanks @samuelcolvin for explaining and asking my opinion.

> I'm not sure I would support changing this at this late stage for pydantic V1.

Why would be the problem? seems we don't need a big change to support it? 

> it would be very useful to add unit tests for this to [pydantic-core](https://github.com/samuelcolvin/pydantic-core)

:+1:

**Comment by samuelcolvin** [comment]
ye, I agree it's a small change. Just there's a lot in `fields.py` that scares me, its pretty complicated. My worry is subtle side-effects not covered tests, but maybe I'm being a wimp. :hand: 

Let's wait and see what @PrettyWood thinks.

**Comment by SergeyTsaplin** [comment]
@samuelcolvin I don't think that the change will break some existed code, it can break only when a user uses `Field(alias='')`  for not "aliased" fields (I have no idea why to do it:)), but of course, the final decision is up to you. For me, it's not a problem to wait for 1.11 if you accidentally decide to merge it to the V1.

About the unit-test to pydantic-core - I can do it, but not soon. It may take me a few days to find a time for it.

File: changes/4253-sergeytsaplin.md [hunk 1]
```diff
@@ -0,0 +1 @@
+Field aliases are compared with None instead of simple `bool(alias)` because an empty string is a valid JSON attribute name.
```
samuelcolvin, line 1: I think we can make this slightly clearer, something like

```suggestion
Allow empty string aliases by using a `alias is not None` check, rather than `bool(alias)`
```

samuelcolvin (commented):
otherwise LGTM

Let's go ahead with this.

samuelcolvin:
> About the unit-test to pydantic-core - I can do it, but not soon. It may take me a few days to find a time for it.

No problem with a wait, this is a marathon, not a sprint.

samuelcolvin:
no, just keep committing, we squash on merge.

PrettyWood (approved):
Looks good! I don't see any risk for this in v1.10

hramezani (approved):

hramezani:
Let's wait for @samuelcolvin for final review and probably merge.

samuelcolvin:
thanks so much.

samuelcolvin:
Well (while I still think this is okay to be included in v1.10), my paranoia was well placed, tis was breaking starlite, see  https://github.com/starlite-api/starlite/pull/402.

Basically with pydantic v1's massive public (or semi-public) API, basically every change can break something. Bring on V2.

PR #4267, opened 2022-07-22
ci: add python 3.11.0b4

PrettyWood:


samuelcolvin:
Looks like this failure should be easy to fix, either by removing or skipping the test or catching the warning.

@PrettyWood will you fix this?

samuelcolvin:
:eyes: have we found something real?

PrettyWood:
@samuelcolvin 
```py
In [1]: from typing import *

In [2]: get_args(Tuple[()])
Out[2]: ((),)

In [3]: get_args(tuple[()])
Out[3]: ()

In [4]: get_args(tuple)
Out[4]: ()
```
🤦

PrettyWood:
I suspect `typing.Tuple` to actually be `tuple` in 3.11. But now we don't differentiate `tuple[()]` and `tuple`.
I guess we could change `get_args` to return `None` when nothing is defined or add an explicit check to `__args__`, which is defined only for `tuple[()]`...
This whole typing situation starts to be annoying. Every single version adds something new

samuelcolvin:
> This whole typing situation starts to be annoying. Every single version adds something new

I agree completely. But then ... that's one of the reason people use pydantic, we wrap that mess in something nice.

On this occasion, it looks like 3.11 is more correct than 3.10:

```py 
>>> from typing import get_args, Tuple
>>> get_args(Tuple[()])
()
>>> get_args(tuple[()])
()
>>> get_args(tuple)
()
>>> sys.version
'3.11.0b4 (main, Jul 22 2022, 17:54:47) [GCC 11.2.0]'
```

PrettyWood:
Yeah in 3.11 `Tuple` is `tuple` so this mess is over. But it highlights a bug in 3.10: we don't support `tuple[()]`

samuelcolvin:
but what does `tuple[()]` mean as a type annotation? Surely it's invalid?

PrettyWood:
Empty tuple

samuelcolvin:
Ye true (sorry, long day).

I think that's not the most serious bug in pydantic v1. Any solution that's consistent between python versions and clear to the user is fine for me.

PrettyWood:
Ok I think I fixed everything. I didn't do anything complicated and tried to be as explicit as possible in the commits. Please review 🙏

hramezani:
Thanks @PrettyWood for the patch :+1: 

it seems this is [part of 3.11 changes](https://docs.python.org/3.11/whatsnew/3.11.html#typing)
> The representation of empty tuple types (Tuple[()]) is simplified. This affects introspection, e.g. get_args(Tuple[()]) now evaluates to () instead of ((),). (Contributed by Serhiy Storchaka in [gh-91137](https://github.com/python/cpython/issues/91137).)

So, the question is which representation is the right one (`((),)` or `()`)?

BTW, The changes LGTM

samuelcolvin:
I think `()` is correct, but pydantic until now returns `((),)`, hence keeping it the same.

LGTM, I'll check again and merge tomorrow.

PrettyWood:
@samuelcolvin @hramezani the fix for `Tuple[()]` didn't handle `tuple[()]`. I think the new fix is better. I added some tests for this and a comment. We'll probably clarify things for pydantic v2

PrettyWood:
Ok I fixed the test when compiled with `cython`.
I think it's an edge case that is almost never used so it's ok to "sit on it".
Still I reckon it's a good idea to have the test to track this. Since pydantic v2 will be a full python package, it should make things work

hramezani (approved):
LGTM :+1: Thanks!

PR #4287, opened 2022-07-27
Add ClusterfuzzLite

AdamKorcz:
Signed-off-by: AdamKorcz <Adam@adalogics.com>

## Change Summary

This PR adds [ClusterfuzzLite](https://google.github.io/clusterfuzzlite/) and a fuzzer for the parser. 

ClusterfuzzLite will run fuzzers in the CI when PRs are made. It can be extended beyond this PR with code coverage and batch fuzzing: https://google.github.io/clusterfuzzlite/running-clusterfuzzlite/github-actions/

## Checklist


.github/workflows/cflite.yaml
```diff
@@ -0,0 +1,29 @@
...
+      id: build
+      uses: google/clusterfuzzlite/actions/build_fuzzers@v1
+      with:
+        sanitizer: ${{ matrix.sanitizer }}
+        language: python
+    - name: Run Fuzzers (${{ matrix.sanitizer }})
+      id: run
+      uses: google/clusterfuzzlite/actions/run_fuzzers@v1
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
```
samuelcolvin, line 26: why do we need the github token?

pydantic/fuzz_parse.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/python3
```
samuelcolvin, line 1: why do we need this file? Surely the source shouldn't have to change.

samuelcolvin (commented):
Thanks for the PR, this is interesting.

Perhaps it would be good to start by adding this to [pydantic-core](https://github.com/samuelcolvin/pydantic-core) which will be the basis for future versions of pydantic?

pydantic/fuzz_parse.py
AdamKorcz: This file is the fuzz harness/test itself. The fuzz engine - [Atheris](https://github.com/google/atheris) - will run this test repeatedly.

This PR does not alter the pydantic source code but adds a fuzz test. The fuzzer can be placed anywhere in this repository though.

.github/workflows/cflite.yaml
AdamKorcz: It is a requirement for ClusterfuzzLite to work.

pydantic/fuzz_parse.py
samuelcolvin: Okay, well it definitely shouldn't be in `pydantic`, best to put it in `tests`

.github/workflows/cflite.yaml
samuelcolvin: Well I assume that, but why?

samuelcolvin:
@Zac-HD I would love your opinion on this?

What do you think of [ClusterfuzzLite](https://google.github.io/clusterfuzzlite/), will it provide any utility here?

Zac-HD:
I think fuzzing can be really helpful for native code, and the Clusterfuzz / OSS-Fuzz team are great.

For Python code, you can repurpose any or all of your existing property-based tests as fuzz targets ([hypothesis docs](https://hypothesis.readthedocs.io/en/latest/details.html#use-with-external-fuzzers), [OSS-fuzz docs](https://google.github.io/oss-fuzz/getting-started/new-project-guide/python-lang/#hypothesis)) and get the benefit of Hypothesis-style replay, minimal counterexamples, explain-mode, etc.  (I'm pretty fond of [HypoFuzz](https://hypofuzz.com/) for Python projects, but don't yet have a great story for sharing the database between Actions-for-fuzzing and Actions-for-CI.  In the meantime, Clusterfuzz does handle that and offer free compute!)

So I'd try it out, ideally with a focus on compiled code, and you can always drop it later if it's not worth the hassle!

samuelcolvin:
Thanks, fuzzing found all the errors in speedate, so I'm a complete convert for native code.

samuelcolvin:
It looks like tests are failing while "building the fuzzer" because the environment is somehow set up wrong and it's trying to use `pydantic/typing.py` on `from typing import ...` commands:

```
Compiling libFuzzer to /usr/lib/libFuzzingEngine.a...  done.
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.8/site-packages/atheris/__init__.py", line 17, in <module>
    from typing import List
  File "/src/pydantic/typing.py", line 4, in <module>
    from typing import (  # type: ignore
ImportError: cannot import name 'TYPE_CHECKING'
```

@AdamKorcz any idea why this is happening? I guess the next question is to ask on https://github.com/google/atheris or https://github.com/google/clusterfuzzlite

AdamKorcz:
> It looks like tests are failing while "building the fuzzer" because the environment is somehow set up wrong and it's trying to use `pydantic/typing.py` on `from typing import ...` commands:
> 
> ```
> Compiling libFuzzer to /usr/lib/libFuzzingEngine.a...  done.
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "/usr/local/lib/python3.8/site-packages/atheris/__init__.py", line 17, in <module>
>     from typing import List
>   File "/src/pydantic/typing.py", line 4, in <module>
>     from typing import (  # type: ignore
> ImportError: cannot import name 'TYPE_CHECKING'
> ```
> 
> @AdamKorcz any idea why this is happening? I guess the next question is to ask on https://github.com/google/atheris or https://github.com/google/clusterfuzzlite

Let me have a look to see if I can see the issue.

samuelcolvin:
No updates and I fail to see how this approach could realistically find anything interesting in pydantic.

@AdamKorcz if you'd like to submit a PR to the [speedate}(https://github.com/pydantic/speedate) project which is also on this organisation, I think that could definitely be useful. In particular speedate does quite a lot of manual parsing and already has fuzz tests that ClusterfuzzLite could perhaps hook into.

PR #4294, opened 2022-07-28
Change Union with None to Optional

hramezani:
## Checklist

### DIFF
--- a/pydantic/color.py
+++ b/pydantic/color.py
@@ -290,7 +290,7 @@ def parse_color_value(value: Union[int, str], max_val: int = 255) -> float:
         raise ColorError(reason=f'color values must be in the range 0 to {max_val}')
 
 
-def parse_float_alpha(value: Union[None, str, float, int]) -> Optional[float]:
+def parse_float_alpha(value: Optional[Union[str, float, int]]) -> Optional[float]:
     """
     Parse a value checking it's a valid float in the range 0 to 1
     """
--- a/pydantic/datetime_parse.py
+++ b/pydantic/datetime_parse.py
@@ -63,7 +63,7 @@ MAX_NUMBER = int(3e20)
 StrBytesIntFloat = Union[str, bytes, int, float]
 
 
-def get_numeric(value: StrBytesIntFloat, native_expected_type: str) -> Union[None, int, float]:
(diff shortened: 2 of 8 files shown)


hramezani:
please review

samuelcolvin:
Seems weird it was like that before.

Could this be a weird artifact of cython? 

Are there other cases where we didn't do it this way?

Is there anything in git blame to suggest these lines were edited at the same time?

If there's no specific reason and cython is happy, LGTM.

hramezani:
> Could this be a weird artifact of cython?

 I have no idea

> Are there other cases where we didn't do it this way?

In `pydantic/color.py`, we have 

* `def parse_float_alpha(value: Union[None, str, float, int]) -> Optional[float]:`
* `def parse_hsl(h: str, h_units: str, s: str, l: str, alpha: Optional[float] = None) -> RGBA:`

in `parse_float_alpha` we have `value: Union[None, str, float, int]` and `Optional[float]` as return value.(why we didn't use `Optional` for return value :thinking: ?)

On the other hand, we have `Optional[float]` in `parse_hsl`.

Note: Both functions were added in the same commit(https://github.com/samuelcolvin/pydantic/commit/22f134475d8df84479239f37f68711fc569a05a9)
 
> Is there anything in git blame to suggest these lines were edited at the same time?
No, I couldn't find it.

samuelcolvin:
Wait I had completely mis-read this, it seems you're suggestion switching:
1. from `Union[None, Foo, Bar]`
2. to `Optional[Union[Foo, Bar]]`

I think this is wrong, I think 1) is more correct.

Do you have any source that recommends this?

@PrettyWood do you have an opinion on this?

hramezani:
I think `Optional[X] = Union[X, None]`. so `Optional[Union[Foo, Bar]] = Union[None, Union[Foo, Bar]]`. 
in theory `Union[None, Union[Foo, Bar]] = Union[None, Foo, Bar]`. but I am not sure how python considers that.

BTW, we can ignore the change if you think it may cause a problem.

hramezani:
I find this from stackoverflow

> There is technically no difference between using Optional[] on a Union[], or just adding None to the Union[]. So Optional[Union[str, int]] and Union[str, int, None] are exactly the same thing.

https://stackoverflow.com/a/51710151/3465428

samuelcolvin:
They're definitely the same, it's just semantics.

I prefer how it is now.

PR #4332, opened 2022-08-05
Typecheck Json type

Bobronium:
## Change Summary
Make Json a proper generic via `Annotated`. Typecheckers will now infer the type of field as `inner_type` of `Json`.
These changes are *soft* breaking,  one should use `Json[Any]` instead of plain `Json`, as the later will result in error from mypy.
In runtime, all should remain the same, except `Json[Any]` is now allowed.
```py
from typing import Any

from pydantic import BaseModel, Json, ValidationError


class AnyJsonModel(BaseModel):
    json_obj: Json[Any]


class ConstrainedJsonModel(BaseModel):
    json_obj: Json[list[int]]

reveal_type(ConstrainedJsonModel().json_obj)  # Revealed type is "list[int]"
```
## Checklist

### DIFF
--- a/tests/mypy/modules/success.py
+++ b/tests/mypy/modules/success.py
@@ -230,7 +230,8 @@ class PydanticTypes(BaseModel):
     my_dir_path: DirectoryPath = Path('.')
     my_dir_path_str: DirectoryPath = '.'  # type: ignore
     # Json
-    my_json: Json = '{"hello": "world"}'
+    my_json: Json[dict[str, str]] = '{"hello": "world"}'  # type: ignore
+    my_json_list: Json[list[str]] = '["hello", "world"]'  # type: ignore
     # Date
     my_past_date: PastDate = date.today() - timedelta(1)
     my_future_date: FutureDate = date.today() + timedelta(1)
@@ -248,6 +249,8 @@ validated.my_file_path.absolute()
 validated.my_file_path_str.absolute()
 validated.my_dir_path.absolute()
 validated.my_dir_path_str.absolute()
+validated.my_json['hello'].capitalize()
+validated.my_json_list[0].capitalize()
 
 stricturl(allowed_schemes={'http'})
 stricturl(allowed_schemes=frozenset({'http'}))
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .idea/
 env/
 venv/
+.venv/
 env3*/
 Pipfile
 *.lock
--- a/docs/examples/types_json_type.py
+++ b/docs/examples/types_json_type.py
@@ -1,29 +1,29 @@
-from typing import List
+from typing import Any, List
 
 from pydantic import BaseModel, Json, ValidationError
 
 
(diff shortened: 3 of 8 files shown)


Bobronium:
please review :)

tests/mypy/modules/success.py
```diff
@@ -230,7 +230,8 @@ class PydanticTypes(BaseModel):
     my_dir_path: DirectoryPath = Path('.')
     my_dir_path_str: DirectoryPath = '.'  # type: ignore
     # Json
-    my_json: Json = '{"hello": "world"}'
+    my_json: Json[Dict[str, str]] = '{"hello": "world"}'  # type: ignore
+    my_json_list: Json[List[str]] = '["hello", "world"]'  # type: ignore
```
samuelcolvin, line 234: Humm, problem is this makes the output type compatibility better, but makes the input type compatibility worse.

 Hence the "type ignore" here.

samuelcolvin (commented):
Thanks for this

Two questions:
* is this better than the status quo - I think "yes"
* is this suitable for a patch release since some legitimate usage will "break" (at least for type checkers) - I think probably not

I would suggest we defer this to V2.

We need to think a lot about serialisation and in particular round-trip validation which affects `Json`, but perhaps not this.

tests/mypy/modules/success.py
Bobronium: Yes, indeed. I thought that it's alright since it's kind of consistent throughout pydantic: you can even see same `type: ignore`'s above for other types.

samuelcolvin: Yes, agreed, also presumably it allows `bytes`.

Hence why I said this is better than the status quo on balance.

Bobronium:
I think it certainly not suitable for a patch release.
As for minor release (still not `V2`), it's totally your call. I can imagine it breaks CI here and there, maybe prevent something from getting deployed. On the other hand, runtime behaviour should be intact, so one might vouch for slipping it under a minor.

I also don't have enough context about V2 release date. 

As for changing inner logic, I'd like to see something entirely different in terms of inner types support in V2. Currently I have a case where I need to decode and validate JWS, would be nice to be able to do something like:
```py
signed_payload: Annotated[PayloadModel, Field(decode=decode_jws_payload)
```
This would allow for usages like
```py
T = TypeVar("T")
JWS = Annotated[T, Field(decode=decode_jws_payload)

signed_payload: JWS[PayloadModel]
```

samuelcolvin:
Ye, I see your point on runtime behaviour not changing.

---

Regarding `Annotated[PayloadModel, Field(decode=decode_jws_payload)`, I think it should be possible. `decode` would probably be `validator`, which should then be a pre-function validator. Since all pydantic is doing in V2 is building a pydantic-core schema, allowing this via an argument to `Field` seems pretty simple (at least from here, when we haven't started implementing it :wink:). Could you create an issue so we remember about this?

samuelcolvin:
thanks so much.

PR #4339, opened 2022-08-06
Generate docs exampels for Python 3.10 and above

### LINKED ISSUE #4334: Docs: have two examples versions: Python 3.6+ and 3.10+
### Initial Checks

- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this feature is missing

### Description

I just noticed that FastAPI has two versions of examples: one for 3.6+ and another for latest python version:
https://fastapi.tiangolo.com/tutorial/query-params-str-validations/

I haven't looked at how it's achieved on a docs syntax/build level, but automating conversion from 3.6 syntax to 3.10+ can be done with https://github.com/asottile/pyupgrade, so the whole process should not require any additional effort after initial implementation.

What do you think?

### Affected Components

Bobronium:
## Change Summary
image

Embedding examples workflow is simplified:

<table>
<tr>
<th>Current</th>
<th>New</th>
</tr>
<tr>
<td>
  
````markdown
```py
{!.tmp_examples/schema_main.py!}
```

Outputs:

```json
{!.tmp_examples/schema_main.json!}
```
````
  
</td>
<td>

````markdown
{!.tmp_examples/schema_main.md!}
````

</td>
</tr>
</table>

`docs/build/exec_examples.py` will now:
1. check if example uses all features of minimal supported version via pyupgrade
2. create and add versions of the file with upgraded syntax that correspond to major python releases, if applicable
3. add `This script is complete, it should run "as is"` if applicable
4. add json output if applicable

## Screenshots 
<details>
<summary>Tabs</summary>

<table>
<tr>
</tr>
<tr>
<td>
  
image
  
</td>
<td>
image

</td>
</tr>
</table>
</details>

<details>
<summary>Json output, single tab, incomplete script</summary>

image
</details>

## Related issue number
Closes #4334

## Checklist

### CODE REVIEW

**Comment by Bobronium** [comment]
Please review. I mostly need feedback on the approach and end result, rather than detailed review of code/style/etc.
Once I get confirmation that this path is applicable, I'll iron out the implementation and will rerequest full review.

**Comment by samuelcolvin** [comment]
Without reviewing the code at all, I think the idea is great and the approach is great.

One note, I think I'll change how examples work a lot in V2, see dirty-equals for an example of what I want to do.

That shouldn't affect this, I'll just copy your code or reimplement the same approach when I'm doing the docs for V2.

**Comment by Bobronium** [comment]
Ok, this should be ready. Plesase review.

I've cleaned up the implementation and allowed multiple future versions to be generated, e.g. Python 3.7, 3.9, 3.10, where it's applicable.

File: docs/examples/postponed_annotations_works.py [hunk 1]
```diff
@@ -1,10 +1,10 @@
 from __future__ import annotations
-from typing import List  # <-- List is defined in the module's global scope
+from typing import Any, List
```
samuelcolvin, line 2: this comment is import for context, please revert.

Also better to use `int` to keep the meaning of the example as clear as possible.

docs/usage/models.md
```diff
@@ -225,9 +210,7 @@ Each error object contains:
 
 As a demonstration:
 
-```py
-{!.tmp_examples/models_errors1.py!}
-```
+{!.tmp_examples/models_errors1.md!}
 _(This script is complete, it should run "as is". `json()` has `indent=2` set by default, but I've tweaked the
```
samuelcolvin, line 214: this comment can probably be removed.

docs/examples/postponed_annotations_broken.py
```diff
@@ -3,11 +3,12 @@
 
 
 def this_is_broken():
-    # List is defined inside the function so is not in the module's
+    # Annotations are defined inside the function so is not in the module's
```
samuelcolvin, line 6: Probably better to revert this, it was slightly clear to the layman.

docs/build/exec_examples.py
```diff
@@ -46,34 +67,28 @@ def to_string(value: Any) -> str:
 
 
 class MockPrint:
-    def __init__(self, file: Path):
+    def __init__(self, file: Path) -> None:
```
samuelcolvin, line 70: perhaps we should add the script to `lint`, `format` and `mypy` in `Makefile`?

samuelcolvin (commented):
Overall I think this looks great :tada:, thanks so much.

A few small things, but my only significant suggestion is:

Could we change the default tab is the newest version? It seems this isn't possible directly in mkdocs, so options are:
* reverse options so the newest version comes first - I think this is kind of fine???
* use JS to change tab on page load, a bit :vomiting_face: but might work
* ask mkdocs to add this feature - could be slow

samuelcolvin:
please update.

docs/requirements.txt
```diff
@@ -10,3 +10,5 @@ mkdocs-material==8.3.9
 sqlalchemy
 orjson
 ujson
+autoflake
+pyupgrade
```
samuelcolvin, line 14: Could you move these up and pin them.

docs/examples/postponed_annotations_broken.py
Bobronium: The reason it was changed is that `List` is removed by pyupgrade in 3.9+, so I added `Any` that shouldn't be removed in any case. Other option is to revert changes and add # dont-upgrade to the top of the file, but it might be unclear why there's no future versions in this example.

And the third way — rethink the whole example so it would fit both versions without confusion on any of them.

docs/build/exec_examples.py
Bobronium: Done!

docs/examples/postponed_annotations_works.py
Bobronium: agreed, but waiting for response in https://github.com/pydantic/pydantic/pull/4339#discussion_r941665874

Bobronium:
> Could we change the default tab is the newest version? It seems this isn't possible directly in mkdocs, so options are:

Don't know, but the good thing is that mkdocs will remember preffered tab throught the docs, so maybe it's ok for now?


Please review. CI failed with mypy complaining since it can't find libraries used in docs. 

My guess is that since mypy previously was used only on pydantic, which didn't have any hard dependencies, it didn't complain much nor required install anything. 

I think mypy should really run in matrix with tests, since results may vary from installation to installation, but that's for another issue.

So options for now:
1. install docs dependencies for linting
2. don't run mypy on docs
3. ~~`# type ignore` the thing~~ — this will cause `unused type ignore comment` errors locally

Suggestions?

samuelcolvin:
You just need to add new sections to `setup.cfg` like:

https://github.com/pydantic/pydantic/blob/4fb872ef3aada7ca9e64b464af9bd5c4f3d232f2/setup.cfg#L83-L84

Please update.

samuelcolvin:
I've made those changes, should be ready now.

Please review.

Bobronium:
@samuelcolvin, what about this discussion https://github.com/pydantic/pydantic/pull/4339/files#r941665874?

I was going to address it on a weekend with an approach that going to be equivalent on all supported versions.

docs/examples/postponed_annotations_broken.py
samuelcolvin: Yes, I think better to import one of the pydantic types like `HttpUrl` inside the function?

samuelcolvin:
good catch, I've replied.

I'll assign back to you: please update.

docs/examples/postponed_annotations_works.py
Bobronium: Replaced `List` and `Any` with `HttpUrl` and added a comment

docs/examples/postponed_annotations_broken.py
Bobronium: Done!

Bobronium:
Please review.

samuelcolvin:
this is really exciting :tada:, thanks so much for the contribution.

Issue #4324, opened 2022-08-04
V1.10 release


samuelcolvin:
To do/decide:
* [x] #2557 - **merged**
* [x] #2745 - needs some tweaks, but we need to decide if it's a good idea before V2
* [x] #2190 - **deferred**
* [x] cherry pick stuff from v1.9 branch, maybe just history #4350
* [x] #3346
* [x] #3593 -  **deferred**
* [x] #3946
* [x] #4028 - **API will change in v2**
* [x] #4354
* [x] #4216
* [x] #4191
* [x] #3941 - revert or fix
* [x] #4339
* [x] #4356

samuelcolvin:
Ideally I'd like to get a pre-release of V1.10 out early next week.

If you have a change you'd like to be included in V1.10, please make sure your PRs are ready to merge and use the magic comment "please review" to assign us.

DetachHead:
#4356 is ready to go i think

satheler:
#4165 is ready to go i think

invokermain:
#3674 would be amazing 🙏

samuelcolvin:
> #3674 would be amazing pray

this is already merged, so will be included.

samuelcolvin:
so are #4165 and #4356.

PR #4356, opened 2022-08-09
fix dataclass containing `Any`

### LINKED ISSUE #4355: `dataclass` type contains `Any`
### Initial Checks

### Description

i use the `disallow-any-expr` rule in mypy, however the `dataclass` decorator contains `Any` types which means i need to `type:ignore` the error on each usage of the pydantic `dataclass` decorator

### Example Code

```Python

@dataclass # error: expression contains Any
class Foo:
    ...
```

```Text
1.9.1
```

### Affected Components

- [X] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)

DetachHead:
## Change Summary
remove `Any` types that contributed to usages of the `dataclass` decorator causing mypy errors when using the `disallow-any-expr` rule

## Related issue number

fixes #4355
## Checklist

  (see [changes/README.md](https://github.com/pydantic/pydantic/blob/master/changes/README.md) for details)

### DIFF
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -4,7 +4,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, ForwardRef, Optional, Tup
 
 from typing_extensions import Literal, Protocol
 
-from .typing import AnyCallable
+from .typing import AnyArgTCallable, AnyCallable
 from .utils import GetterDict
 from .version import compiled
 
@@ -62,10 +62,10 @@ if not compiled:
         getter_dict: Type[GetterDict]
         alias_generator: Optional[Callable[[str], str]]
         keep_untouched: Tuple[type, ...]
-        schema_extra: Union[Dict[str, Any], 'SchemaExtraCallable']
-        json_loads: Callable[[str], Any]
-        json_dumps: Callable[..., str]
-        json_encoders: Dict[Type[Any], AnyCallable]
+        schema_extra: Union[Dict[str, object], 'SchemaExtraCallable']
+        json_loads: Callable[[str], object]
+        json_dumps: AnyArgTCallable[str]
+        json_encoders: Dict[Type[object], AnyCallable]
         underscore_attrs_are_private: bool
 
         # whether or not inherited models as fields should be reconstructed as base model
@@ -146,7 +146,7 @@ class BaseConfig:
         pass
 
 
-def get_config(config: Union[ConfigDict, Type[BaseConfig], None]) -> Type[BaseConfig]:
+def get_config(config: Union[ConfigDict, Type[object], None]) -> Type[BaseConfig]:
     if config is None:
         return BaseConfig
 
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -68,7 +68,7 @@ if TYPE_CHECKING:
         __pydantic_validate_values__: ClassVar[Callable[['Dataclass'], None]]
         __pydantic_has_field_info_default__: ClassVar[bool]  # whether a `pydantic.Field` is used as default value
 
-        def __init__(self, *args: Any, **kwargs: Any) -> None:
+        def __init__(self, *args: object, **kwargs: object) -> None:
             pass
 
         @classmethod
@@ -88,6 +88,8 @@ __all__ = [
     'make_dataclass_validator',
 ]
 
+_T = TypeVar('_T')
+
 if sys.version_info >= (3, 10):
 
     @dataclass_transform(kw_only_default=True, field_descriptors=(Field, FieldInfo))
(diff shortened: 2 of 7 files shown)


DetachHead:
please review

pydantic/typing.py
```diff
@@ -76,9 +77,14 @@ def get_all_type_hints(obj: Any, globalns: Any = None, localns: Any = None) -> A
         return get_type_hints(obj, globalns, localns, include_extras=True)
 
 
+_T = TypeVar('_T')
+
 AnyCallable = TypingCallable[..., Any]
 NoArgAnyCallable = TypingCallable[[], Any]
 
+# workaround for https://github.com/python/mypy/issues/9496
+AnyArgAnyCallable = TypingCallable[..., _T]
```
samuelcolvin, line 86:
```suggestion
AnyArgTCallable = TypingCallable[..., _T]
```
Or something, since its return type is paramaterised.

pydantic/config.py
```diff
@@ -62,10 +62,10 @@ class ConfigDict(TypedDict, total=False):
         getter_dict: Type[GetterDict]
         alias_generator: Optional[Callable[[str], str]]
         keep_untouched: Tuple[type, ...]
-        schema_extra: Union[Dict[str, Any], 'SchemaExtraCallable']
-        json_loads: Callable[[str], Any]
-        json_dumps: Callable[..., str]
-        json_encoders: Dict[Type[Any], AnyCallable]
+        schema_extra: Union[Dict[str, object], 'SchemaExtraCallable']
```
samuelcolvin, line 65: are these really all required?

I would rather keep the change as small as possible while fixing the original issue.

pydantic/dataclasses.py
```diff
@@ -86,6 +86,8 @@ def __validate__(cls: Type['DataclassT'], v: Any) -> 'DataclassT':
     'make_dataclass_validator',
 ]
 
+_T = TypeVar('_T')
```
samuelcolvin, line 89:
```suggestion
T = TypeVar('T')
```
Unless we need the leading underscore?

tests/mypy/configs/mypy-plugin-strict-no-any.ini
```diff
@@ -0,0 +1,23 @@
+[mypy]
```
samuelcolvin, line 1: can't we just use `strict = true`?

tests/mypy/modules/no_any.py
```diff
@@ -0,0 +1,11 @@
+from pydantic.dataclasses import dataclass
+
+
+@dataclass
+class Foo:
+    ...
```
samuelcolvin, line 6: can you just add a field, just in case that makes a difference.

samuelcolvin (commented):
looking good, but I wonder if we can reduce the size of the change somewhat?

pydantic/dataclasses.py
DetachHead: i usually do that with `TypeVar`s and other variables/functions to indicate that they are "private" ie. shouldn't be imported elsewhere.

not sure what the standard is so i'll just remove it if you want

tests/mypy/configs/mypy-plugin-strict-no-any.ini
DetachHead: no unfortunately because `strict` doesn't enable any of the `Any` rules

pydantic/config.py
DetachHead: yeah they are, changing any one of them back to `Any` causes the mypy error to come back. i made sure to only change types that were absolutely necessary to make the error go away.

pydantic/dataclasses.py
DetachHead: looks like there are already `TypeVar`s in this repo following the same pattern i do

https://github.com/pydantic/pydantic/blob/fa6b9788a1001591cbac44923476b5d12a14c1fc/pydantic/main.py#L95

samuelcolvin: okay, that's fine.

samuelcolvin:
This is a fix since V1.9, so really needs to be included before we release v1.10.

---

#### Notice 

Or if you need this in V1.10 but don't have time to complete it (or aren't the author), please comment here and on #4324.

DetachHead:
@samuelcolvin all outstanding conversations are resolved, unless there's something i'm missing? i think it's ready to merge

samuelcolvin:
Okay, I'll look. 

In future, best if you can say "please review" so the label changes and reviewers get assigned. I basically only look at PRs with the "ready for review" label.

DetachHead:
my bad, i didn't realize i had to comment it again after making changes, i assumed clicking "re-request review" would've done it

samuelcolvin:
> my bad, i didn't realize i had to comment it again after making changes, i assumed clicking "re-request review" would've done it

we might be able to somehow extend hooky to detect "re-request review" if there's a webhook for it. But it doesn't work at present.

samuelcolvin:

PR #4374, opened 2022-08-14
Build binaries for Python 3.11

samuelcolvin:
See https://github.com/pypa/cibuildwheel/pull/1226, cibuildwheel now support building python 3.11 binaries using `rc1`.

Although it's pretty early to release a package with python 3.11 binaries, the chance of a major change to 3.11 between now and it's release is very small. Also I want pydantic V1.10 to support python 3.11 but I don't want to wait for [October 3rd](https://www.python.org/downloads/release/python-3110rc1/) to release it.

See #4267 for the actual changes that were required to pydantic to support Python 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,7 +308,7 @@ jobs:
   build:
     name: build py3.${{ matrix.python-version }} on ${{ matrix.platform || matrix.os }}
     needs: [lint, test-linux-compiled, test-not-compiled, test-old-mypy, test-fastapi]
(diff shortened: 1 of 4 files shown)


samuelcolvin:
(waiting for 0999c9f which should run full CI including building wheels, then this can be merged AFAIK)

PrettyWood:
maybe add 'Programming Language :: Python :: 3.11' in setup.py ?

samuelcolvin:
Merged to avoid asking obvious questions, but please comment if you have any further suggestions/changes.

PR #4384, opened 2022-08-16
fix pattern matching version

samuelcolvin:
ref https://github.com/pydantic/pydantic/pull/3921#issuecomment-1217160250

skip change file check

### DIFF
--- a/docs/examples/models_structural_pattern_matching.py
+++ b/docs/examples/models_structural_pattern_matching.py
@@ -1,3 +1,4 @@
+# requires python3.10
 from pydantic import BaseModel
 
 


samuelcolvin:
please review.

Bobronium (approved):
LGTM! 😸

irgolic:
Can't find a docs preview in CI, but conceptually LGTM

samuelcolvin:
> Can't find a docs preview in CI, but conceptually LGTM
> 
> EDIT: Found it now that the rest of CI finished :) looks good

https://smokeshow.helpmanual.io/40174q1w4h6d071c084t/usage/models/#structural-pattern-matching

Bobronium:
We also might want to automate some basic checks for minimal required version in examples.

Is there a known tool for that guesses minimal required version by checking for new syntax/other features?

Or, we can just run docs-build in matrix. There will be warnings on lower versions, but shouldn't be any errors.

samuelcolvin:
> Or, we can just run docs-build in matrix. There will be warnings on lower versions, but shouldn't be any errors.

makes sense, but let's wait for V2 when I want to change the docs to build like watchfiles.

PR #4406, opened 2022-08-19
Fix #1458 - Allow for custom parsing of environment variables via parse_env_var in Config object

### LINKED ISSUE #1458: Settings: Custom parsing of environment variables (non-JSON)

acmiyaguchi:
## Change Summary

This PR adds an `parse_env_var` classmethod into the Config object to allow for custom parsing of the environment within Settings. This is useful when specifying a custom format in environment strings e.g. lists (1,2,3) or key-value pairs (a=1,b=2,c=3). String-encoded JSON is not ideal in certain situations, such as when the environment is specified inside of a YAML file (leading to string-escaped string-encoded JSON).

This is an alternative to https://github.com/pydantic/pydantic/pull/3977

## Related issue number

fixes https://github.com/pydantic/pydantic/issues/1458

## Checklist

  (see [changes/README.md](https://github.com/pydantic/pydantic/blob/main/changes/README.md) for details)


acmiyaguchi:
please review

pydantic/env_settings.py
```diff
@@ -190,10 +194,10 @@ def __call__(self, settings: BaseSettings) -> Dict[str, Any]:  # noqa C901
                 else:
                     # field is complex and there's a value, decode that as JSON, then add explode_env_vars
                     try:
-                        env_val = settings.__config__.json_loads(env_val)
+                        env_val = settings.__config__.parse_env_var(field.name, env_val)
                     except ValueError as e:
                         if not allow_json_failure:
-                            raise SettingsError(f'error parsing JSON for "{env_name}"') from e
+                            raise SettingsError(f'error parsing envvar "{env_name}"') from e
```
samuelcolvin, line 200:
```suggestion
                            raise SettingsError(f'error parsing env var "{env_name}"') from e
```

pydantic/env_settings.py
```diff
@@ -299,9 +303,9 @@ def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
                     secret_value = path.read_text().strip()
                     if field.is_complex():
                         try:
-                            secret_value = settings.__config__.json_loads(secret_value)
+                            secret_value = settings.__config__.parse_env_var(field.name, secret_value)
                         except ValueError as e:
-                            raise SettingsError(f'error parsing JSON for "{env_name}"') from e
+                            raise SettingsError(f'error parsing envvar "{env_name}"') from e
```
samuelcolvin, line 308:
```suggestion
                            raise SettingsError(f'error parsing env var "{env_name}"') from e
```

pydantic/env_settings.py
```diff
@@ -190,10 +194,10 @@ def __call__(self, settings: BaseSettings) -> Dict[str, Any]:  # noqa C901
                 else:
                     # field is complex and there's a value, decode that as JSON, then add explode_env_vars
                     try:
-                        env_val = settings.__config__.json_loads(env_val)
+                        env_val = settings.__config__.parse_env_var(field.name, env_val)
```
samuelcolvin, line 197: please can you also rename `allow_json_failure` both here and in `field_is_complex`

samuelcolvin (commented):
Otherwise, I think this looks great.

Please update.

docs/examples/settings_with_custom_parsing.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Settings(BaseSettings):
+    numbers: List[int]
+
+    class Config:
+        @classmethod
+        def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
+            if field_name == 'numbers':
+                return parse_list(raw_val)
```
samuelcolvin, line 19:
```suggestion
                return [int(x) for x in s.split(',')]
```

then remove `parse_list` above, I think this is just as easy to read and more compact.

I don't think you need the strip since `int('  123  ') == 123`

docs/examples/settings_with_custom_parsing.py
```diff
@@ -0,0 +1,24 @@
+# output-json
```
samuelcolvin, line 1:
```suggestion
```
Clearer just to use the standard print handling.

samuelcolvin:
For the sake of speed, I've made the alterations I suggested - AFAIK this is the last change before V1.10 beta is released. :tada:

LGTM, @PrettyWood @hramezani please review.

tests/test_settings.py
```diff
@@ -1215,3 +1215,66 @@ def test_builtins_settings_source_repr():
...
+    with pytest.raises(SettingsError, match='error parsing env var "top"'):
+        Settings()
+
+
+def test_secret_settings_source_custom_env_parse(tmp_path):
+    p = tmp_path / 'top'
+    p.write_text('1=apple,2=banana')
+
+    class Settings(BaseSettings):
+        top: Dict[int, str] = Field(env_parse=_parse_custom_dict)
```
hramezani, line 1268: Do we need to pass `env_parse=_parse_custom_dict`?

hramezani:
This is great :+1: 

I just left a small comment other than that LGTM :rocket:

samuelcolvin:
Okay, I'm going to go for it. I want to get the beta out today.

@PrettyWood if you see anything important, we can create another pR.

@acmiyaguchi thanks so much for this.

acmiyaguchi:
Thank you for the review and accepting the patch into v1.10 beta! I appreciate Pydantic and look forward to seeing it mature.

samuelcolvin:
Thanks.

Somehow the change for this feature didn't get included in the changelog for v1.10a1, but the feature is there. The change will be included in the changelog for the next (pre) release.

PR #4407, opened 2022-08-20
Fix PEP487 `__set_name__` protocol in BaseModel

tlambert03:
## Change Summary

This is a small change that brings BaseModel back into compliance with the the `__set_name__` protocol defined in [[PEP 487](https://peps.python.org/pep-0487/](https://peps.python.org/pep-0487/#proposal)), and described in the [creating the class object](https://docs.python.org/3/reference/datamodel.html#creating-the-class-object) portion of the python documentation, which states:

> the following additional customization steps are invoked after creating the class object:
>
> - The `type.__new__` method collects all of the attributes in the class namespace that define a [`__set_name__()`](https://docs.python.org/3/reference/datamodel.html#object.__set_name__) method;
> - Those `__set_name__` methods are called with the class being defined and the assigned name of that particular attribute

In the case of `BaseModel`, some names in the original namespace (such as instances of `ModelPrivateAttr`) are stripped before calling `type.__new__` with the `new_namespace`.  As such, objects implementing this `__set_name__` protocol are skipped.

Rather than calling `type.__new__` with the original namespace, this PR implements the protocol directly by calling `__set_name__` on any objects in the namespace that provide the method.

## Checklist

  (see [changes/README.md](https://github.com/pydantic/pydantic/blob/main/changes/README.md) for details)

## Context

For context, I've been using [sqlmodel](https://github.com/tiangolo/sqlmodel) recently and would like to to implement this pattern

```python
from sqlmodel import SQLModel

class SomeObject(SQLModel, table=True):
    name: str
    slug: str | None

    @on_before_save
    def _do_stuff(self):
        self.slug = slugify(self.name)
```

to bind object-specific listeners to various SqlAlchemy [ORM events](https://docs.sqlalchemy.org/orm/events.html)... however, this requires that the decorator know the class in which it is being called (which, among other things, seems to have been a primary motivator of `__set_name__`)

### DIFF
--- a/tests/test_create_model.py
+++ b/tests/test_create_model.py
@@ -2,30 +2,38 @@ from typing import Generic, TypeVar
 
 import pytest
 
-from pydantic import BaseModel, Extra, Field, ValidationError, create_model, errors, validator
+from pydantic import (
+    BaseModel,
+    Extra,
+    Field,
+    ValidationError,
+    create_model,
+    errors,
+    validator,
+)
 from pydantic.generics import GenericModel
 
 
 def test_create_model():
-    model = create_model('FooModel', foo=(str, ...), bar=123)
+    model = create_model("FooModel", foo=(str, ...), bar=123)
     assert issubclass(model, BaseModel)
     assert issubclass(model.__config__, BaseModel.Config)
-    assert model.__name__ == 'FooModel'
-    assert model.__fields__.keys() == {'foo', 'bar'}
+    assert model.__name__ == "FooModel"
+    assert model.__fields__.keys() == {"foo", "bar"}
     assert model.__validators__ == {}
-    assert model.__config__.__name__ == 'Config'
-    assert model.__module__ == 'pydantic.main'
+    assert model.__config__.__name__ == "Config"
+    assert model.__module__ == "pydantic.main"
 
 
 def test_create_model_usage():
-    model = create_model('FooModel', foo=(str, ...), bar=123)
-    m = model(foo='hello')
-    assert m.foo == 'hello'
+    model = create_model("FooModel", foo=(str, ...), bar=123)
+    m = model(foo="hello")
+    assert m.foo == "hello"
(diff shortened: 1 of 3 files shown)


tlambert03:
please review

tests/test_create_model.py
```diff
@@ -222,3 +222,27 @@ class TestGenericModel(GenericModel):
     result = AAModel[int](aa=1)
     assert result.aa == 1
     assert result.__config__.orm_mode is True
+
+
+def test_set_name():
+
+    from unittest.mock import Mock
```
samuelcolvin, line 229: you don't really need this, you could just do

```py
calls = []
...
def __set_name__(self, owner, name):
    calls.append((owner, name))
```
But if you do want to use mock, please make it a global import to match the style of other tests.

tests/test_create_model.py
```diff
@@ -222,3 +222,27 @@ class TestGenericModel(GenericModel):
...
+
+    from pydantic.fields import ModelPrivateAttr
```
samuelcolvin, line 231: again, please make this a global import.

tests/test_create_model.py
```diff
@@ -222,3 +222,27 @@ class TestGenericModel(GenericModel):
...
+        def __init__(self, fn):
+            super().__init__()
+            self.fn = fn
+
+        def __set_name__(self, owner, name):
+            mock(owner, name)
+
+    class A(BaseModel):
+        @class_deco
+        def _some_func(self):
```
samuelcolvin, line 245: can you also call `_some_func` to check it's a proper bound method.

samuelcolvin (commented):
otherwise, LGTM.

Please update.

samuelcolvin:
Looks good, thanks so much.

tests/test_create_model.py
tlambert03: working on this one still, don't merge yet :)

PR #4410, opened 2022-08-21
Fix duplicate calls to `__set_name__` for non-private attributes in `BaseModel.__new__`

tlambert03:
## Change Summary

My apologies @samuelcolvin, as I was working on adding the test you requested in https://github.com/pydantic/pydantic/pull/4407#discussion_r950847546, I found a bug in my implementation, which is that `__set_name__` was now getting called twice for public attributes (i.e. anything in `new_namespace`), and only once for private attributes.  I didn't notice the auto-merge in time :)

This PR fixes that, adds a test that `__set_name__` is called only once both for regular `object` subclasses as well as `ModelPrivateAttrs`, and adds the test requested above to make sure the bound method works (for regular objects... privateAttrs likely won't work for this pattern, as commented in the test)

skip change file check

## Checklist

* [x] Unit tests for the changes exist
* [x] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable
* [x] `changes/<pull request or issue id>-<github username>.md` file added describing change
  (see [changes/README.md](https://github.com/pydantic/pydantic/blob/main/changes/README.md) for details)
* [x] My PR is ready to review, **please add a comment including the phrase "please review" to assign reviewers**

PR #4411, opened 2022-08-21
Revert "Fix PEP487 `__set_name__` protocol in BaseModel"

samuelcolvin:
Reverts #4407

See #4410 for explanation.

skip change file check


samuelcolvin:
This was a mistake, I manually force pushed to remove it.

PR #4432, opened 2022-08-24
Warn and ignore `__slots__` argument to `create_model`

samuelcolvin:
Found while investigating error in https://github.com/starlite-api/starlite/pull/401

## Change Summary

In v1.9.2 `__slots__` could be passed to `create_model`, I'm not sure if it was honoured or ignored, but it didn't cause an error.

In v1.10.0a2 however it caused a warning, and if that warning was ignored, an error creating the model.

This was happening [here](https://github.com/starlite-api/starlite/blob/1bff6d0477e29ea7b4aa57b25786de9735abb6b7/starlite/typing.py#L54)

Even more unusually, this only seemed to be an error if pydantic is compiled.

With this change a `RuntimeWarning` is raised, and `__slots__` is then ignored.

## Related issue number

Caused by #3946 I think.

### DIFF
--- a/pydantic/typing.py
+++ b/pydantic/typing.py
@@ -333,15 +333,16 @@ if sys.version_info < (3, 8):
         return type_ in NONE_TYPES
 
 elif sys.version_info[:2] == (3, 8):
-    # We can use the fast implementation for 3.8 but there is a very weird bug
-    # where it can fail for `Literal[None]`.
-    # We just need to redefine a useless `Literal[None]` inside the function body to fix this
 
     def is_none_type(type_: Any) -> bool:
-        Literal[None]  # fix edge case
         for none_type in NONE_TYPES:
             if type_ is none_type:
                 return True
+        # With python 3.8, specifically 3.8.10, Literal "is" check sare very flakey
+        # can change on very subtle changes like use of types in other modules,
+        # hopefully this check avoids that issue.
+        if is_literal_type(type_):
+            return all_literal_values(type_) == (None,)
         return False
 
 else:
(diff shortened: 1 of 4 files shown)


samuelcolvin:
please review.

hramezani:
Except the failing CI, LGTM 👍

samuelcolvin:
I'll wait for @hot123s (author of #3946) and @PrettyWood to review, then I think we're ready for a final pre-release - `v1.10.0b1`.

pydantic/typing.py
```diff
@@ -333,15 +333,16 @@ def is_none_type(type_: Any) -> bool:
...
-    # We just need to redefine a useless `Literal[None]` inside the function body to fix this
 
     def is_none_type(type_: Any) -> bool:
-        Literal[None]  # fix edge case
         for none_type in NONE_TYPES:
             if type_ is none_type:
                 return True
+        # With python 3.8, specifically 3.8.10, Literal "is" check sare very flakey
+        # can change on very subtle changes like use of types in other modules,
+        # hopefully this check avoids that issue.
```
PrettyWood, line 343: I would love to know what is compiled behind the scene to understand how this edge case arrives. Such a weird bug

PrettyWood (approved):

pydantic/typing.py
samuelcolvin: I agree, so weird.

The line that caused the issue again was `__slots__: Optional[Tuple[str, ...]] = None,`, if I changed it to `__slots__: `Optional[Tuple[str, ...]]` = None,` it went again. But I decided this was a safer fix.

samuelcolvin:
okay, I want to get the (hopefully) last pre-release out, so I'm going to merge this.

Issue #312, opened 2018-11-24
Check on duplicate validators breaks validators in jupyter notebooks


amogorkon:
Check on duplicate validators breaks validators in jupyter notebooks if a cell is executed more than once.

```
from pydantic import BaseModel, validator
import yaml
from ipaddress import ip_address

class Config(BaseModel):
    ip: str
    name: str
    port: int
        
    @validator("ip")
    def check_ip(cls, value):
        return ip_address(value)            
            

with open("config") as f:
    d = yaml.safe_load(f)
    c = Config(**d)
    print(c)
```
Returns

```
Config ip=IPv4Address('123.0.0.200') name='asdf' port=2392
when the cell is called first and raises

ConfigError Traceback (most recent call last)
in ()
3 from ipaddress import ip_address
4
----> 5 class Config(BaseModel):
6 ip: str
7 name: str

in Config()
8 port: int
9
---> 10 @validator("ip")
11 def check_ip(cls, value):
12 return ip_address(value)

~/anaconda3/lib/python3.7/site-packages/pydantic/main.py in dec(f)
449 ref = f.module + '.' + f.qualname
450 if ref in _FUNCS:
--> 451 raise ConfigError(f'duplicate validator function "{ref}"')
452 _FUNCS.add(ref)
453 f_cls = classmethod(f)

ConfigError: duplicate validator function "main.Config.check_ip"
```

The second time the same cell is called.

I was somewhat surprised by this behaviour and when I checked the code, I didn't really get why this exception is raised since we're dealing with a set here, so behaviour shouldn't change.
I would suggest to replace this exception with a logger message, warning level, which won't mess with people's code and is well supported by jupyter.

samuelcolvin:
Would be great if you could format code or output like this to make it more readable.

I see what you mean. What about an environment variable and/or some check if we're in the jupyter before doing the duplicate check?

samuelcolvin:
I understand what you mean about warnings, but this catches a lot of potential problems for me alone, and probably for others too.

I'd rather keep it as an error but allow a workaround.

Abraxos:
I am having this problem as well, and I do not really understand what kind of behavior this is supposed to prevent. Could you please explain and perhaps I can re-write my code better?

samuelcolvin:
The test for this behavior demonstrates what it's there to avoid:

[...]

PR #4436, opened 2022-08-25
Replace duplicate validator detection mechanism

nzig:
I was confused about when exactly `allow_reuse=True` was required, so I looked into the implementation. As far as I can tell, the motivation behind the feature is outlined in https://github.com/pydantic/pydantic/issues/312#issuecomment-449875744. 

I think the same goal can be accomplished more cleanly with the metaclass's [`__prepare__`](https://docs.python.org/3/reference/datamodel.html#preparing-the-class-namespace) method, without requiring the library user to set `allow_reuse=True`.
## Change Summary
This PR replaces the "duplicate validator" detection mechanism. 
Before, there was a `set` of qualified validator names, and if a validator was added with a qualified name that already existed in the set, this was detected as a duplicate.

The new mechanism is to directly detect when a validator is re-assigned within the class body, by using a custom `dict` subclass as the class namespace. This is more convenient because it works without the user needing to specify `allow_reuse=True`, while still detecting validators with duplicate names.

I think it would be best to merge/implement this for pydantic V2, and retire the `allow_reuse` parameter for V2. However, if you want this for V1, this PR essentially deprecates the `allow_reuse` parameter, which no longer has any effect. It can be merged into V1 with the appropriate documentation changes.

## Related issue number
None

## Checklist

* [x] Unit tests for the changes exist - covered by the existing `test_duplicates`.
* [x] Documentation reflects the changes where applicable - not relevant unless you plan to merge this to V1. For V2 this would just be removing the documentation for `allow_reuse`.
  If we're not merging to V1 then we can skip change file check.

### DIFF
--- a/pydantic/class_validators.py
+++ b/pydantic/class_validators.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional,
 
 from .errors import ConfigError
 from .typing import AnyCallable
-from .utils import ROOT_KEY, in_ipython
+from .utils import ROOT_KEY
 
 if TYPE_CHECKING:
     from .typing import AnyClassMethod
@@ -44,7 +44,6 @@ if TYPE_CHECKING:
     ValidatorsList = List[ValidatorCallable]
     ValidatorListDict = Dict[str, List[Validator]]
 
-_FUNCS: Set[str] = set()
 VALIDATOR_CONFIG_KEY = '__validator_config__'
 ROOT_VALIDATOR_CONFIG_KEY = '__root_validator_config__'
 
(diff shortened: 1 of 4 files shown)


nzig:
please review

samuelcolvin:
This will have to wait for V2 by which time I suspect most of this code will be completely replaced.

If therefore delay any further work on this.

I also think we should make this a warning not an error in V2.

nzig:
No problem. Thanks @samuelcolvin !

samuelcolvin:
I think the principle here could still be very useful, but this will need to be rebased on `main` for V2. It might be easier to restart it given how much has changed.

samuelcolvin:
The plan is to take this logic and move it onto the new code in `main`.

samuelcolvin:
assigned both @hramezani and @adriangb, whoever gets here first, please work on it.

adriangb:
@nzig we're taking your (excellent) idea and using it in https://github.com/pydantic/pydantic/pull/5398. Please feel free to take a look there and give any feedback you have. Thanks for the contribution and sorry it took so long to get to it!

nzig:
@adriangb  Looks good. Thanks for getting this in 😄 🚀 !

Issue #4223, opened 2022-07-09
"Config" overrides symbol of same name in class "BaseSettings"


adithep:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
 pydantic version: 1.9.1
            pydantic compiled: True
                 install path: /Users/adithepsrinarula/repo/prometheus-hoarding/dist/export/python/virtualenv/3.10.5/lib/python3.10/site-packages/pydantic
               python version: 3.10.5 (main, Jul  4 2022, 16:56:19) [Clang 13.1.6 (clang-1316.0.21.2.5)]
                     platform: macOS-12.4-arm64-arm-64bit
     optional deps. installed: ['dotenv', 'typing-extensions']

```

```py
import pydantic

class Settings(pydantic.BaseSettings):

    class Config:
        env_file = ".env"
        env_file_encoding = "utf-8"
```

I get error:

![Screen Shot 2565-07-09 at 18 33 49](https://user-images.githubusercontent.com/6279977/178103823-35166b3a-23eb-4a58-aa9d-3b6532a5cf9c.png)

RobertCraigie:
@adithep Pylance is correct here. You have to explicitly subclass the `Config`, e.g.

```py
import pydantic

class Settings(pydantic.BaseSettings):
    class Config(pydantic.BaseSettings.Config):
        env_file = '.env'
        env_file_encoding = 'utf-8'
```

jannisko:
Should this maybe be changed in the docs as well? The example given [here](https://pydantic-docs.helpmanual.io/usage/settings/) still causes an error in type checkers.

DbCrWk:
Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:

```
pydantic version: 1.9.2
            pydantic compiled: True
                 install path: /Users/dev.dabke/Documents/repo/ninja/nibbler/.venv/lib/python3.10/site-packages/pydantic
               python version: 3.10.4 (main, Jul 18 2022, 16:17:04) [Clang 13.1.6 (clang-1316.0.21.2.5)]
                     platform: macOS-12.5.1-arm64-arm-64bit
     optional deps. installed: ['devtools', 'dotenv', 'typing-extensions']
```


For what it's worth, running

> ```python
> import pydantic
> 
> class Settings(pydantic.BaseSettings):
>     class Config(pydantic.BaseSettings.Config):
>         env_file = '.env'
[...]

PR #4450, opened 2022-08-30
fix type hints in `BaseSettings.Config`

samuelcolvin:
## Change Summary

Add type hints to `BaseSettings.Config` to avoid mypy errors, also correct mypy version compatibility comment in the docs.

## Related issue number

See https://github.com/pydantic/pydantic/issues/4223#issuecomment-1230894434

### DIFF
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -79,15 +79,15 @@ class BaseSettings(BaseModel):
             return {}
 
     class Config(BaseConfig):
-        env_prefix = ''
-        env_file = None
-        env_file_encoding = None
-        env_nested_delimiter = None
-        secrets_dir = None
-        validate_all = True
-        extra = Extra.forbid
-        arbitrary_types_allowed = True
-        case_sensitive = False
+        env_prefix: str = ''
+        env_file: Optional[DotenvType] = None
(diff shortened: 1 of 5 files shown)


samuelcolvin:
I think this is ready, please review.

pydantic/env_settings.py
```diff
@@ -79,15 +79,15 @@ def _build_values(
...
-        env_prefix = ''
-        env_file = None
-        env_file_encoding = None
-        env_nested_delimiter = None
-        secrets_dir = None
-        validate_all = True
-        extra = Extra.forbid
-        arbitrary_types_allowed = True
-        case_sensitive = False
+        env_prefix: str = ''
```
hramezani, line 82: can't we use `ClassVar` here?

samuelcolvin: maybe, but is there any advantage?

hramezani: probably nothing :)

samuelcolvin: I think given this will be either the final beta, or included directly in the release, I want to keep it as simple as possible.

hramezani (approved):
LGTM :+1:

PR #4453, opened 2022-08-30
Documentation update - Update settings_main.py

sanderbrandenburg:
## Change Summary

mypy complains about "incompatible types in assignment" when I follow the documentation.

According to the existing mypy test, the str must be boxed in a Field:
https://github.com/pydantic/pydantic/pull/1849/commits/af40e98bb86488345f8dd9ff1e3a0e101c18a8ec#diff-c639ffa4761f60e358be073ec6cc2eebcb6f547c7fdee9d492d1ed4f866e468b

## Related issue number
#1849 - defines a test for PyObjects, which is not followed in the example

## Checklist
skip change file check # <-- I think this is ok?

### DIFF
--- a/docs/examples/settings_main.py
+++ b/docs/examples/settings_main.py
@@ -24,7 +24,7 @@ class Settings(BaseSettings):
     pg_dsn: PostgresDsn = 'postgres://user:pass@localhost:5432/foobar'
     amqp_dsn: AmqpDsn = 'amqp://user:pass@localhost:5672/'
 
-    special_function: PyObject = 'math.cos'
+    special_function: PyObject = Field('math.cos')
 
     # to override domains:
     # export my_prefix_domains='["foo.com", "bar.com"]'


samuelcolvin:
I'm not sure I agree about this. It's a bit confusing since the use of `Field` isn't required at runtime, so adding it could be confusing.

At the very least we need a comment like 

> Field(..) used here to mollify mypy et al.

samuelcolvin:
No reply, so closing for sake of tidiness.

I think we should document this properly in V2, but I don't think this change is quite right.

sanderbrandenburg:
I was trying to figure this out, but I'm new to mypy. I think BaseModel is somehow simulating type coercion at runtime as it requires a type at runtime to actually work. However, this form of type coercion is not supported by the python language, so can never be supported by the static analyzer.

Given that the source code, tests and examples are strewn by types it's hard to understand 'pydantic only selectively supports mypy', but maybe that's just me.

samuelcolvin:
The reason this works is that `Field` is a function which returns `Any`:

https://github.com/pydantic/pydantic/blob/317bef33b06e05f65f57b1ba009dfd949b462689/pydantic/fields.py#L250

I agree it can be confusing, we'll try to document it better in V2. But in the end, typing is python is an imperfect art, there's no one authoritative type checker, there's pydantic, mypy, pyright, pyre and others. This is not like typescript and rust where there's one source of truth.

PR #4461, opened 2022-09-01
Bugfix/anyurl escapes percent sign in url

### LINKED ISSUE #4458: AnyUrl escapes percent sign in URLs without TLD that are already escaped.
### Initial Checks

### Description

### Example Code

```Python


class Foo(BaseModel):
    url: AnyUrl


class Bar(BaseModel):
    url: str


```

```Text
             pydantic version: 1.10.1
            pydantic compiled: True
```

### Affected Components

vetedde:
## Change Summary

Fix re-quote when build url. I added relevant cases to test.
Also fix mypy error, because without it I couldn't create commit

## Related issue number
Fix #4458 

## Checklist

* [] My PR is ready to review, **please add a comment including the phrase "please review" to assign reviewers**

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -121,7 +121,7 @@ class PydanticPlugin(Plugin):
 
     def get_class_decorator_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
         if fullname == DATACLASS_FULLNAME:
-            return dataclasses.dataclass_class_maker_callback  # type: ignore[return-value]
+            return dataclasses.dataclass_class_maker_callback
         return None
 
     def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -121,7 +121,7 @@ class PydanticPlugin(Plugin):
 
     def get_class_decorator_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
         if fullname == DATACLASS_FULLNAME:
-            return dataclasses.dataclass_class_maker_callback
+            return dataclasses.dataclass_class_maker_callback  # type: ignore[return-value]
         return None
 
     def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -396,6 +396,10 @@ class AnyUrl(str):
 
     @classmethod
     def quote(cls, string: str, safe: str = '') -> str:
+        pattern = r'^([\w]+|(%\d{2}))+$'
+        if re.match(pattern, string) is not None:
+            return string
+
         return quote_plus(string, safe) if cls.quote_plus else quote(string, safe)
 
     def __repr__(self) -> str:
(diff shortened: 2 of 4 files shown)


pydantic/mypy.py
```diff
@@ -121,7 +121,7 @@ def get_method_hook(self, fullname: str) -> Optional[Callable[[MethodContext], T
 
     def get_class_decorator_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
         if fullname == DATACLASS_FULLNAME:
-            return dataclasses.dataclass_class_maker_callback  # type: ignore[return-value]
+            return dataclasses.dataclass_class_maker_callback
```
hramezani, line 124: I think we don't need this change. you may have an old version of mypy in your local setup

hramezani:
Please add tests for your change into [test_build_url_quote_plus](https://github.com/pydantic/pydantic/blob/e34ff9255f281eaccac10d0436c082626d9146c3/tests/test_networks.py#L703)

hramezani:
Thanks @vetedde for this patch :+1: 
I left some comments. please update

pydantic/networks.py
```diff
@@ -396,6 +396,10 @@ def apply_default_parts(cls, parts: 'Parts') -> 'Parts':
 
     @classmethod
     def quote(cls, string: str, safe: str = '') -> str:
+        pattern = r'^([\w]+|(%\d{2}))+$'
```
hramezani, line 399: please compile the pattern and then use the `match` on the compiled pattern.
Take a look at the same implementation in `datetime_parse.py`
https://github.com/pydantic/pydantic/blob/32ea8853450b50e5b963ce677ab1fff95bc2e56a/pydantic/datetime_parse.py#L30

vetedde:
@hramezani found a problem with case "foo%20bar", when quote_plus is True. My solution returns foo%20bar, but need foo+bar. I'm thinking of adding an extra check, but it's nested check and still have to unquote at some point

```python
string_with_precent_code_re = re.compile(r'^([\w]+|(%\d{2}))+$')

if string_with_precent_code_re.match(string) is not None:
    if cls.quote_plus:
        string = unquote_plus(string)
    else:
        return string

return quote_plus(string, safe) if cls.quote_plus else quote(string, safe)
```

Maybe it's better to do it at once:

```python
string = unquote_plus(string, safe) if cls.quote_plus else unquote(string, safe)
return quote_plus(string, safe) if cls.quote_plus else quote(string, safe)
```
 or always do unquote if match %xx

What do you think?

pydantic/networks.py
samuelcolvin: Can you add a comment explaining what this regex is doing?

I also think it could be simplified.

hramezani:
@vetedde I think there are 3 possible solutions:

1. unquote string in any case and then quote it
2. unquote string if it is already quoted(checking by regex for example) and then quote it
3. revert this feature because seems it is complex and is not good performance-wise (we need to unquote and then quote)

What do you think @samuelcolvin

samuelcolvin:
please have a look at #4469, I think it's a cleaner approach to solve this. I've also added a large number of test cases to at least lock the behaviour. I'll reply on the original issue with some more context.

Thanks so much @vetedde for looking into this, I'll leave this PR open in case it is decided that it's a better approach than #4469.

hramezani:
I think we still will have the problem reported in https://github.com/pydantic/pydantic/issues/4468 even with this patch. which is a breaking change.

samuelcolvin:
After some consideration and discussions, I've decided to replace this by #4470. @vetedde thanks again for your contribution.

See #4470 for a full explanation.

Issue #4468, opened 2022-09-02
AnyUrl percent encoding (#3061) introduced v1.10.0 breaks database connection string


doc0815:
## Checks

- [x] I added a descriptive title to this issue
- [x] I have searched (google, github) for similar issues and couldn't find anything
- [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

## Bug
This bug-fix https://github.com/pydantic/pydantic/issues/3061 introduced percent encoding for the `query` attribute of `AnyUrl`. It lists a set of characters that get percent encoded and refers to the [url-spec](https://www.w3.org/Addressing/URL/url-spec.txt).

According to the url-spec, the `+` character is in the group of safe characters and **should not be encoded** (page 16 in the doc).

As a consequence of the introduced characters encoding, my database connection string to a MS SQL Server DB got unusable:

```
import pydantic


url = pydantic.AnyUrl.build(
    scheme="mssql+pyodbc",
    user="user",
    password="password",
    host="127.0.0.1",
    port="1433",
    path="/database",
    query="driver=ODBC+Driver+17+for+SQL+Server",
)
print(pydantic.VERSION)
print(url)


# 1.9.1
# mssql+pyodbc://user:password@127.0.0.1:1433/database?driver=ODBC+Driver+17+for+SQL+Server

# 1.10.1
# mssql+pyodbc://user:password@127.0.0.1:1433/database?driver=ODBC%2BDriver%2B17%2Bfor%2BSQL%2BServer
```

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:

```
pydantic version: 1.9.1
pydantic compiled: True
install path: C:\Users\xxx\PycharmProjects\xxx\venv\Lib\site-packages\pydantic
python version: 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64 bit (AMD64)]
platform: Windows-10-10.0.19044-SP0
optional deps. installed: ['typing-extensions']
```

It would be cool to have a MSSQLServerDsn similar to Postgres, Redis, or MongoDB instead of the general AnyUrl (https://pydantic-docs.helpmanual.io/usage/types/#urls).

hramezani:
maybe related to https://github.com/pydantic/pydantic/issues/4458

samuelcolvin:
duplicate of #4458

samuelcolvin:
[...]

PR #4469, opened 2022-09-02
Fix existing percent encoding

### LINKED ISSUE #4458: AnyUrl escapes percent sign in URLs without TLD that are already escaped.

samuelcolvin:
## Change Summary

As described in the issue, #4458 was double encoding URLs, this splits the URL on encoding groups with the regex `(.*?)(%[\dA-F]{2})` to (hopefully) fix the issue.

Changes:
* existing "percentage encodings" are not re-encoded - they're just cast to upper case to match RFC 3986 and httpe - e.g. `foo%2Bbar` is left unchanged while `foo%2bbar` is converted to `foo%2Bbar`
* forward slash `/` is no longer encoded in query paramters, e.g. `http://example.com?a=b/c` is left unchanged

## Related issue number

fix #4458

## Checklist

### DIFF
--- a/changes/4458-samuelcolvin.md
+++ b/changes/4458-samuelcolvin.md
@@ -1,0 +1,1 @@
+Fix percent encoding of URLs with which are already percent encoded.
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -1,5 +1,4 @@
 import re
-from functools import partial
 from ipaddress import (
     IPv4Address,
     IPv4Interface,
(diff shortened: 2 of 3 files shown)


hramezani:
I think we still will have the problem reported in https://github.com/pydantic/pydantic/issues/4468 even with this patch. which is a breaking change.

samuelcolvin:
Okay, this problem is bigger than I thought due to the strange logic to decide when to rebuild urls, see

https://github.com/pydantic/pydantic/blob/317bef33b06e05f65f57b1ba009dfd949b462689/pydantic/networks.py#L297-L300

and 

https://github.com/pydantic/pydantic/blob/317bef33b06e05f65f57b1ba009dfd949b462689/pydantic/networks.py#L377-L379

which is completely out of date from the URL quoting - we previously only rebuild domains when required to cope with "international" TLDs, but now we should be rebuilding if there's a path or a query or a user or anything which might need quoting.

However if I just force rebuilding of all URLs, lots of tests are failing.

samuelcolvin:
> I think we still will have the problem reported in #4468 even with this patch. which is a breaking change.

Good point, should be fixed by the last commit.

samuelcolvin:
replaced by #4470.

PR #4473, opened 2022-09-02
Remove Cython & Move to `pyproject.toml`

samuelcolvin:
skip change file check

Starting for on V2, first speed up build by removing cython.

PR #4476, opened 2022-09-04
Use tomllib on Python 3.11

hauntsaninja:
## Change Summary

Python 3.11 includes tomllib in the standard library. When installed on Python 3.11, mypy will use tomllib and does not have tomli or toml as a dependency.

### DIFF
--- a/changes/4476-hauntsaninja.md
+++ b/changes/4476-hauntsaninja.md
@@ -1,0 +1,1 @@
+Use `tomllib` on Python 3.11 when parsing `mypy` configuration.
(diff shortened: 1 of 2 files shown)


samuelcolvin:
great, thanks so much. Will be included in v1.10 next patch.

PrettyWood:
I'm surprised `tomli` was not in the requirements 🤔  Btw @samuelcolvin I guess we can remove `dataclasses==0.6; python_version < '3.7'` now that python 3.6 support has been dropped

samuelcolvin:
:+1: done in #4473

PR #4480, opened 2022-09-05
prevent long strings as int inputs

### LINKED ISSUE #1477: potential DOS risk with pydantic - fix pending
I have been made aware of a potential DOS attack risk in pydantic.

The fix I believe is relatively trivial, I will release:
* `v1.5.2` based of the current `v1.5.1` tag
* `v1.4.1` based of the current `v1.4` tag

These releases will be made just after **1pm UTC** on **2020/5/11**, that's next Monday.

If you require a fix to any other version, please let me know on this issue.

samuelcolvin:
see #1477 and in turn, python/cpython#95778 this check should be unnecessary once patch releases are out for 3.7, 3.8, 3.9 and 3.10 but better to check here until then.

See [this twitter thread](https://twitter.com/samuel_colvin/status/1565999997396680704) the python patch releases should be out in the next few days, but I know how slow people are to upgrade these things, so I think better to fix here.

### NOTICE:

This does not full protect user from the DOS risk since the standard library JSON implementation
(and other std lib modules like xml) use `int()` and are called before this, the best workaround is to

1. update to the latest patch release of python when it's released
2. (until then) use a different JSON library like ujson

## Related issue number

fix #1477

## Checklist

### DIFF
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -131,9 +131,9 @@ def int_validator(v: Any) -> int:
     # see https://github.com/pydantic/pydantic/issues/1477 and in turn, https://github.com/python/cpython/issues/95778
     # this check should be unnecessary once patch releases are out for 3.7, 3.8, 3.9 and 3.10
     # but better to check here until then.
-    # NOTICE: this does not full protect user from the DOS risk since the standard library JSON implementation
-    # (and other std lib modules like xml) use `int()` and are called before this, the best workaround is to
-    # 1. update to the latest patch release of python, 2. use a different JSON library like ujson
+    # NOTICE: this does not fully protect user from the DOS risk since the standard library JSON implementation
+    # (and other std lib modules like xml) use `int()` and are likely called before this, the best workaround is to
+    # 1. update to the latest patch release of python once released, 2. use a different JSON library like ujson
     if isinstance(v, (str, bytes, bytearray)) and len(v) > max_str_int:
         raise errors.IntegerError()
 
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -120,9 +120,22 @@ def bool_validator(v: Any) -> bool:
     raise errors.BoolError()
 
 
+# matches the default limit cpython, see https://github.com/python/cpython/pull/96500
+max_str_int = 4_300
+
+
 def int_validator(v: Any) -> int:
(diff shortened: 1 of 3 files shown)


samuelcolvin:
should now pass, please review.

samuelcolvin:
@pydantic/pydantic-maintainers do you think we should backport this to v1.9 and v1.8?

hramezani (approved):
LGTM :+1: 

I think we don't need backporting this because this is not Pydantic issue and the proper fix will be out in Python patch releases. Also, 1.10.x is mostly backward compatible release and users can have the fix by upgrading to 1.10.x

hramezani (approved):
LGTM :+1: 

I think we don't need backporting this because this is not Pydantic issue and the proper fix will be out in Python patch releases. Also, 1.10.x is mostly backward compatible release and users can have the fix by upgrading to 1.10.x

hramezani (approved):
LGTM :+1: 

I think we don't need backporting this because this is not Pydantic issue and the proper fix will be out in Python patch releases. Also, 1.10.x is mostly backward compatible release and users can have the fix by upgrading to 1.10.x

hramezani:
sorry. I think it was my internet problem. I approved the PR 3 times :man_facepalming:

samuelcolvin:
No problem. Good to see you're sure :smile:.

pydantic/validators.py
```diff
@@ -120,10 +120,23 @@ def bool_validator(v: Any) -> bool:
...
     if isinstance(v, int) and not (v is True or v is False):
         return v
 
+    # see https://github.com/pydantic/pydantic/issues/1477 and in turn, https://github.com/python/cpython/issues/95778
+    # this check should be unnecessary once patch releases are out for 3.7, 3.8, 3.9 and 3.10
+    # but better to check here until then.
+    # NOTICE: this does not fully protect user from the DOS risk since the standard library JSON implementation
+    # (and other std lib modules like xml) use `int()` and are likely called before this, the best workaround is to
+    # 1. update to the latest patch release of python once released, 2. use a different JSON library like ujson
+    if isinstance(v, (str, bytes, bytearray)) and len(v) > max_str_int:
```
cmyui, line 137: maybe overkill, but should this include `memoryview` as well?

```py


class A(BaseModel):
    x: int

A(x=memoryview(b"1" * 5000))
```

PR #4482, opened 2022-09-05
Basic fix of GenericModel cache to detect order of arguments in Union models

### LINKED ISSUE #4474: First instantiation of generic Union model defines type coercion of models instantiated later
### Initial Checks

### Description

When pydantic generic models are instantiated with Unions, the order of a particular set of types are fixed by the first instantiation, e.g.:

```python
from typing import Generic, TypeVar, Union

NumberT = TypeVar('NumberT')


class NumberModel(GenericModel, Generic[NumberT]):
    data: NumberT


FloatOrIntModel = NumberModel[Union[float, int]]
print(FloatOrIntModel(data='1').data)

IntOrFloatModel = NumberModel[Union[int, float]]
print(IntOrFloatModel(data='1').data)
```
prints
```
1.0
1.0
```
While changing the order of the instantiations changes type coercion of both models:
```
1
1
```

In pydantic, type coercion of Union models depends on the order of the types. However, as explained in the documentation on [Unions](https://pydantic-docs.helpmanual.io/usage/types/#unions):

> typing.Union also ignores order when [defined](https://docs.python.org/3/library/typing.html#typing.Union), so Union[int, float] == Union[float, int] which can lead to unexpected behaviour when combined with matching based on the Union type order inside other type definitions

I think this should be considered a bug in pydantic, even though the user is warned against 'unexpected behaviour' combined with other code that depend on Union type equality. The unexpected behaviour in this case is caused by pydantic itself, there is no "matching based on the Union type order inside other type definitions" in the example code, nor are any third-party code imported. It is not difficult to envision situations where this dependency could cause very hard-to-find bugs.

It seems the cause of the error is the `_generic_types_cache` in `generics.py`, which caches the first Union model and uses this as the blueprint of later matching models (i.e. with the same set of types).

Setting `smart_union` to `True` has no impact.

Related to https://github.com/pydantic/pydantic/issues/2835.

### Example Code

_No response_

```Text
pydantic version: 1.10.1
            pydantic compiled: True
                 install path: /Users/sveinugu/Library/Caches/pypoetry/virtualenvs/unifair-myuXak-6-py3.10/lib/python3.10/site-packages/pydantic
               python version: 3.10.4 | packaged by conda-forge | (main, Mar 24 2022, 17:42:03) [Clang 12.0.1 ]
                     platform: macOS-11.2.3-arm64-arm-64bit
```

### Affected Components

- [X] [Generic Models](https://pydantic-docs.helpmanual.io/usage/models/#generic-models)

sveinugu:
## Change Summary

Basic fix of GenericModel cache to include the order of arguments in Union models into the cache key. 
This fixes the basic problem described in issue #4474, so that e.g. Model[Union[float, int]] is not used as a
blueprint for Model[Union[float, int]].  Added tests for basic and nested versions of the problem. Fixing the
nested problem depends on the fixing of the almost identical issue described in
 https://github.com/python/cpython/issues/86483 in CPython. The test was thus skipped, but kept for 
documentation.

## Related issue number

fix #4474

## Checklist

* [ x ] Unit tests for the changes exist
* [ x ] Tests pass on CI and coverage remains at 100%
* [ x ] Documentation reflects the changes where applicable
* [ x ] `changes/<pull request or issue id>-<github username>.md` file added describing change
* [ x ] My PR is ready to review, **please add a comment including the phrase "please review" to assign reviewers**

### DIFF
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -62,7 +62,11 @@ class GenericModel(BaseModel):
             returned as is.
 
         """
-        cached = _generic_types_cache.get((cls, params))
+
+        def _cache_key(_params: Any) -> Tuple[Type[GenericModelT], Any, Tuple[Any, ...]]:
+            return cls, _params, get_args(_params)
+
+        cached = _generic_types_cache.get(_cache_key(params))
         if cached is not None:
             return cached
         if cls.__concrete__ and Generic not in cls.__bases__:
@@ -128,9 +132,9 @@ class GenericModel(BaseModel):
 
         # Save created model in cache so we don't end up creating duplicate
         # models that should be identical.
-        _generic_types_cache[(cls, params)] = created_model
+        _generic_types_cache[_cache_key(params)] = created_model
         if len(params) == 1:
-            _generic_types_cache[(cls, params[0])] = created_model
+            _generic_types_cache[_cache_key(params[0])] = created_model
 
         # Recursively walk class type hints and replace generic typevars
         # with concrete types that were passed.
--- a/changes/4474-sveinugu.md
+++ b/changes/4474-sveinugu.md
@@ -1,0 +1,1 @@
+Basic fix of GenericModel cache to detect order of arguments in Union models
--- a/tests/test_generics.py
+++ b/tests/test_generics.py
@@ -688,6 +688,52 @@ def test_generic_model_redefined_without_cache_fail(create_module, monkeypatch):
         assert globals()['MyGeneric[Model]__'] is third_concrete
 
 
+def test_generic_model_caching_detect_order_of_union_args_basic(create_module):
+    # Basic variant of https://github.com/pydantic/pydantic/issues/4474
+    @create_module
+    def module():
+        from typing import Generic, TypeVar, Union
+
+        from pydantic.generics import GenericModel
+
+        t = TypeVar('t')
+
+        class Model(GenericModel, Generic[t]):
+            data: t
+
+        int_or_float_model = Model[Union[int, float]]
+        float_or_int_model = Model[Union[float, int]]
+
+        assert type(int_or_float_model(data='1').data) is int
+        assert type(float_or_int_model(data='1').data) is float
+
+
+@pytest.mark.skip(
+    reason="""
+Depends on similar issue in CPython itself: https://github.com/python/cpython/issues/86483
+Documented and skipped for possible fix later.
+"""
+)
+def test_generic_model_caching_detect_order_of_union_args_nested(create_module):
+    # Nested variant of https://github.com/pydantic/pydantic/issues/4474
+    @create_module
+    def module():
+        from typing import Generic, List, TypeVar, Union
+
+        from pydantic.generics import GenericModel
+
+        t = TypeVar('t')
+
+        class Model(GenericModel, Generic[t]):
+            data: t
+
+        int_or_float_model = Model[List[Union[int, float]]]
+        float_or_int_model = Model[List[Union[float, int]]]
+
+        assert type(int_or_float_model(data=['1']).data[0]) is int
+        assert type(float_or_int_model(data=['1']).data[0]) is float
+
+
 def test_get_caller_frame_info(create_module):
     @create_module
     def module():


pydantic/generics.py
```diff
@@ -62,7 +62,13 @@ def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[T
             returned as is.
 
         """
-        cached = _generic_types_cache.get((cls, params))
+
+        def _cache_key(
+            _params: Union[Type[Any], Tuple[Type[Any], ...]]
+        ) -> Tuple[Type[GenericModelT], Union[Type[Any], Tuple[Type[Any], ...]], Tuple[Any, ...]]:
```
samuelcolvin, line 68: Just use, this is an internal method, so that's sufficient.

```suggestion
        def _cache_key(_params: Any) -> Tuple[Type[GenericModelT], Any, Tuple[Any, ...]]:
```

samuelcolvin (commented):
otherwise LGTM, you just need to add "fix " to the issue ID in the PR body to satisfy the checks and also allow github to auto close the issue.

pydantic/generics.py
sveinugu: Ah! Tried to replace with `Type[Any]` instead of `Any` but that did not work.

Also, I was wondering about closing the issue, as the nested variant of the issue is not fixed. Do you want to keep the issue open for this or is the documented and skipped test good enough?

samuelcolvin: I don't mind, if you want to keep the issue open, use the PR ID in the change file.

Either way this is urgent since people are (understandably) waiting on the patch release.

I just want this to pass so I can merge it and get the release done.

sveinugu: Added 'fix'. You can reopen the issue after, if needed.

samuelcolvin: great, thanks.

samuelcolvin:
LGTM.

@PrettyWood please review,  you happy with this?

PrettyWood (approved):
LGTM! Thanks

samuelcolvin:
Thanks so much @sveinugu.

sveinugu:
> Thanks so much @sveinugu.

You're welcome! Happy to get the fix in. And thank you for creating such a useful library! Looking forward to v2.

pepastach:
Hi @sveinugu @samuelcolvin 

We recently bumped Pydantic from 1.10.1 to 1.10.2 and our tests started failing. After some lengthy investigation I found this change to be the root cause. I'd like to ask you for your opinions.

We have a Pydantic model with a field such as
```python
f: MyGenericModel[Callable[[Task, Dict[str, Any]], Iterable[Result]]] = Field(...)
```

Just importing Python module with this code gives me:

```
cls = <class 'foo.MyGenericModel'>
params = typing.Callable[[foo.Task, typing.Dict[str, typing.Any]], typing.Iterable[foo.Result]]

    def __class_getitem__(cls: Type[GenericModelT], params: Union[Type[Any], Tuple[Type[Any], ...]]) -> Type[Any]:
        """Instantiates a new class from a generic class `cls` and type variables `params`.
    
        :param params: Tuple of types the class . Given a generic class
            `Model` with 2 type variables and a concrete model `Model[str, int]`,
            the value `(str, int)` would be passed to `params`.
        :return: New model class inheriting from `cls` with instantiated
            types described by `params`. If no parameters are given, `cls` is
            returned as is.
    
        """
    
        def _cache_key(_params: Any) -> Tuple[Type[GenericModelT], Any, Tuple[Any, ...]]:
            return cls, _params, get_args(_params)
    
>       cached = _generic_types_cache.get(_cache_key(params))
E       TypeError: unhashable type: 'list'
```

The problem seems to be that the function `GenericModel._cache_key()` now calls `get_args()` which in turns calls Python's `typing.get_args()` -> and this function returns a tuple with a list in it. And that makes it unhashable.

```python
def get_args(tp):
    """Get type arguments with all substitutions performed.

    For unions, basic simplifications used by Union constructor are performed.
    Examples::
        get_args(Dict[str, int]) == (str, int)
        get_args(int) == ()
        get_args(Union[int, Union[T, int], str][int]) == (int, str)
        get_args(Union[int, Tuple[T, int]][str]) == (int, Tuple[str, int])
        get_args(Callable[[], T][int]) == ([], int)
    """
    if isinstance(tp, _GenericAlias) and not tp._special:
        res = tp.__args__
        if get_origin(tp) is collections.abc.Callable and res[0] is not Ellipsis:
            res = (list(res[:-1]), res[-1])             # <======== this list is a problem
        return res
    return ()
```

If I skip the callable typing and only define the field as 
```python
f: MyGenericModel[Callable] = Field(...)
```
then it runs okay.


Thanks in advance!

sveinugu:
@pepastach Hmm... Seems we did not think of that. Surprised that typing uses lists and not tuples for Callable arguments, but there is probably a good reason for this (or it might be just that it would be too much a nuisance to switch between brackets and parentheses...). In any case, there might be other examples of unhashable return values from `get_params`, so a simple solution could be to just catch `TypeError` and in that case default to the previous _cache_key generation.

I am new as contributor to pydantic, but I assume it would be cleanest if you could create another issue for this.

samuelcolvin:
Yes, new issue which references this PR.

But honestly, generics are one of the things that needs to be refactored properly in V2.

This patches-on-patches-on-patches are getting more and more complex - no one's fault, but I really hope we can get away from some of these anti-patterns in V2.

Anyway, apart from me whinging, this shouldn't be too hard to fix/work around as @sveinugu suggests.

samuelcolvin:
> Yes, new issue which references this PR.

I've creatd #4551

PR #4484, opened 2022-09-05
fix: dataclass wrapper was not always called

### LINKED ISSUE #4477: __post_init__ and validators are not triggered in class inherited from stdlib dataclass in 1.10.1
### Initial Checks

### Description

__post_init__ and validator are not triggered in the following code. However, uncommenting any of the lines: `py_dataclass(D_C_Base)` or `b: str = "string"` solves the issue.
Versions 1.9.x could trigger both __post_init__ and validators without these lines.

### Example Code

```Python
from pydantic.dataclasses import dataclass as py_dataclass
from pydantic import validator


@dataclass
class D_C_Base:
    a: float = 1.0

# py_dataclass(D_C_Base)

@py_dataclass
class D_C_Child(D_C_Base):
    # b: str = "string"
    
    def __post_init__(self):
        print('POST_INIT')
        
    @validator('a')
    def validate_a(cls, value):
        print(f'Validating "a" is {value}')

d_c_class = D_C_Child()
```

```Text
pydantic version: 1.10.1
pydantic compiled: True
install path: C:\Users\TestUser\.conda\envs\py38\Lib\site-packages\pydantic
               python version: 3.8.13 (default, Mar 28 2022, 06:59:08) [MSC v.1916 64 bit (AMD64)]
platform: Windows-10-10.0.22000-SP0

Also,

pydantic version: 1.10.1
pydantic compiled: True
install path: /local/envs/py39_test/lib/python3.9/site-packages/pydantic
               python version: 3.9.1 (default, Dec 11 2020, 14:32:07)  [GCC 7.3.0]
platform: Linux-3.10.0-1160.66.1.el7.x86_64-x86_64-with-glibc2.17
```

### Affected Components

- [x] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)

PrettyWood:
## Change Summary
In a stdlib dataclass without `__post_init__`, `__post_init__` call is not added in the generated `__init__`.
So when we inherit from a stdlib dataclass and we add  a `__post_init__`, it won't be called. So all its related code + validation will be ignored.
Now it should be okish (not ok because it's a mess).

In v2, we should rely on stdlib dataclasses and go with a more functional approach

## Related issue number
fix #4477

## Checklist

### DIFF
--- a/changes/4477-PrettyWood.md
+++ b/changes/4477-PrettyWood.md
@@ -1,0 +1,1 @@
+fix: dataclass wrapper was not always called
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -34,7 +34,20 @@ validation without altering default `M` behaviour.
 import sys
 from contextlib import contextmanager
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Dict, Generator, Optional, Type, TypeVar, Union, overload
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Generator,
+    Optional,
+    Set,
+    Type,
+    TypeVar,
+    Union,
+    overload,
+)
 
 from typing_extensions import dataclass_transform
 
@@ -184,7 +197,7 @@ def dataclass(
     def wrap(cls: Type[Any]) -> 'DataclassClassOrWrapper':
         import dataclasses
 
-        if is_builtin_dataclass(cls):
+        if is_builtin_dataclass(cls) and _extra_dc_args(_cls) == _extra_dc_args(_cls.__bases__[0]):  # type: ignore
             dc_cls_doc = ''
             dc_cls = DataclassProxy(cls)
             default_validate_on_init = False
@@ -416,6 +429,14 @@ def _dataclass_validate_assignment_setattr(self: 'Dataclass', name: str, value: 
                 raise ValidationError([error_], self.__class__)
 
     object.__setattr__(self, name, value)
+
+
+def _extra_dc_args(cls: Type[Any]) -> Set[str]:
+    return {
+        x
+        for x in dir(cls)
+        if x not in getattr(cls, '__dataclass_fields__', {}) and not (x.startswith('__') and x.endswith('__'))
+    }
 
 
 def is_builtin_dataclass(_cls: Type[Any]) -> bool:
--- a/tests/test_dataclasses.py
+++ b/tests/test_dataclasses.py
@@ -1395,3 +1395,80 @@ def test_extra_forbid_list_error():
         @pydantic.dataclasses.dataclass
         class Foo:
             a: List[Bar(a=1)]
+
+
+def test_parent_post_init():
+    @dataclasses.dataclass
+    class A:
+        a: float = 1
+
+        def __post_init__(self):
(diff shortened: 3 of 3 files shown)


samuelcolvin (commented):
LGTM, I'll fix the markdown in the change file when building history (no need for extra builds).

changes/4477-PrettyWood.md
```diff
@@ -0,0 +1 @@
+fix: always call __post_init__ when set in dataclasses
```
samuelcolvin, line 1:
```suggestion
fix: always call `__post_init__` when set in dataclasses
```

pydantic/dataclasses.py
```diff
@@ -285,8 +284,23 @@ def new_post_init(self: 'Dataclass', *args: Any, **kwargs: Any) -> None:
...
+                handle_extra_init(self, *args, **kwargs)
+                new_post_init(self, *args, **kwargs)
+
+            setattr(dc_cls, '__init__', new_init)
+            setattr(dc_cls, '__post_init__', new_post_init)
+        else:
+
+            @wraps(init)
+            def new_init(self: 'Dataclass', *args: Any, **kwargs: Any) -> None:
+                handle_extra_init(self, *args, **kwargs)
```
samuelcolvin, line 300: Looks like this line isn't covered.

samuelcolvin (commented):
otherwise LGTM, please update.

changes/4477-PrettyWood.md
PrettyWood: it's what I wrote but somehow it has been added in bold

pydantic/dataclasses.py
PrettyWood: I'll check

changes/4477-PrettyWood.md
samuelcolvin: it's markdown... this is the change file, not the PR title.

PrettyWood: I know. Just a VS code extension that messed up

PrettyWood:
The fix is wrong. I'm pushing something new that handles properly inheritance and stuff.
But it's not acceptable yet. The check on `is_builtin_dataclass` needs to be more resilient (it's a mess honestly...)

samuelcolvin:
:disappointed: I agree, let me know when it's ready to review.

PrettyWood:
@samuelcolvin it's good! Sorry I had client calls all day. Hard to do that at the same time 😓

samuelcolvin:
> @samuelcolvin it's good! Sorry I had client calls all day. Hard to do that at the same time sweat

I'm sure! Sorry about this.

samuelcolvin:
LGTM, thanks so much.

PrettyWood:
I feel like `__pydantic_model__` and all should be attached to the dataclass wrapper and not the dataclass itself. But I'm too tired to think about it now. Hopefully we don't have new bugs with dataclasses and we'll rely directly on stdlib in v2

samuelcolvin:
agreed, the whole idea of the `Argument` type (https://github.com/pydantic/pydantic-core/pull/203) is that we can just call

```py
args, kwargs = schema_validator.validate_python(input)
return dc(*args, **kwargs)
```
and never touch the dataclass.

PrettyWood:
Yup! That's great 👍

PR #4491, opened 2022-09-06
fix `copy_on_model_validation` type hint

samuelcolvin:
## Change Summary

Fix to type hint spotted by @PrettyWood on #4473.

skip change file check.


PrettyWood (approved):
thx!

PR #4493, opened 2022-09-06
fix: avoid multiple calls of `__post_init__` when dataclasses are inherited

### LINKED ISSUE #4487: Dataclass post_init_post_parse called multiple times.
### Initial Checks

### Description

When using inheritance with `pydantic.dataclasses.dataclass` based objects with `pydantic>=1.10` I seem to have encountered a bug that calls `__post_init_post_parse__` methods multiple times if a `__post_init__` method is called further down an inheritance chain.

Running the below code snippet outputs the print statement three times, whereas if `__post_init__` is placed in `AbstractClass` the `"post init post parse called"` statement is called twice.

The exact same code snippet on `pydantic<1.10` works as expected with the print statement only printed the once.

Including an overridden `__post_init__` method in `ConcreteClass` only prints the statement once - but of course if there is more useful logic contained in `BaseClass.__post_init__` (as would be the case for this pattern) then this is ignored.

Removing the `@dataclass` decorators from the base classes works fine however, except **if** we didn't have access to the code for those underlying base classes.

EDIT: It appears its due to these lines:
https://github.com/pydantic/pydantic/blob/v1.10.2/pydantic/dataclasses.py#L290-L304

### Example Code

```Python

@dataclass
class BaseClass:
    def __post_init__(self):
        pass

@dataclass
class AbstractClass(BaseClass):
    pass

@dataclass
class ConcreteClass(AbstractClass):
    def __post_init_post_parse__(self):
        print("post init post parse called")

ConcreteClass()

# prints:
# post init post parse called
# post init post parse called
# post init post parse called


# Example of nested dataclasses from docs: https://pydantic-docs.helpmanual.io/usage/dataclasses/#convert-stdlib-dataclasses-into-pydantic-dataclasses

from datetime import datetime
from typing import Optional

@dataclass
class Meta:
    modified_date: Optional[datetime]
    seen_count: int
    def __post_init__(self):
        print("POST INIT")

@dataclass
class File(Meta):
    filename: str
    def __post_init_post_parse__(self):
        print("POST INIT POST PARSE")

File(filename="file.txt", modified_date=datetime.now(), seen_count=1)

# prints:
# POST INIT
# POST INIT POST PARSE
# POST INIT POST PARSE
```

```Text
pydantic version: 1.10.2
            pydantic compiled: True
                 install path: *****
               python version: 3.8.13 (default, Mar 28 2022, 06:16:26)  [Clang 12.0.0 ]
```

### Affected Components

- [X] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)

PrettyWood:
When `__post_init__` was defined on a parent class, `new_post_init` would be called many times because it "auto-refers" itself (it refers the parent one).
Now we call the right `__post_init__`

fix #4487


PrettyWood:
Please review

hramezani (approved):
LGTM :+1:

samuelcolvin (approved):
LGTM, @PrettyWood feel free to merge or add other fixes.

samuelcolvin:
@PrettyWood what do you think? Should we merge this and make a release, or do you want to have a shot at the other dataclass issues?

samuelcolvin:
That's very kind, thank you. No problem about the delay.

PR #4499, opened 2022-09-07
Fix field constraints with Collection and Sequence

AntoineD:
With pydantic 1.10.2, this model
```python
class Model(BaseModel):
    name: Collection[str] = Field(..., min_items=1)
```
gives an error
```bash
E   ValueError: On field "name" the following field constraints are set but not enforced: min_items. 
E   For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints
```

This PR extents the allowed constraints of `List` (*{min,max,unique}_items*) to `Collection` and `Sequence`.

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -10,6 +10,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    Collection,
     Dict,
     ForwardRef,
     FrozenSet,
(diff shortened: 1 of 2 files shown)


samuelcolvin:
This isn't new in v1.10, so it shouldn't be changed in a patch release.

you'll need to wait for V2 and check the equivalent logic works correctly.

Let me know if I'm missing something.

AntoineD:
Do you mean that this not considered as a bug fix?

Should I rebase on main the PR or wait for the release of V2?

AntoineD:
Hi @samuelcolvin,
I'm sorry to come back to this but I do not understand your answer: it seems to be a bug to me, couldn't this be fixed in a patch release?

samuelcolvin:
See https://pydantic-docs.helpmanual.io/blog/pydantic-v2/

There will not be another v1.X release, and V2 will involve rewriting pydantic in general, and this logic in particular, completely. Thus I'm afraid there's not really any value in rebasing this.

I agree that it's a bug and we'll make sure it gets fixed in v2. 

I really appreciate the contribution, I just don't want to waste your time.

I'd love more help on v2 once it's a bit further along and there are more obvious, independent tasks.

AntoineD:
Thank you for explaining :+1:

PR #4507, opened 2022-09-08
Enumerate the index of duplicated items

joshtemple:
When debugging uniqueness validation errors, it would be useful to know the index of duplicates (especially when the list being validated is very long). This change modifies the error message to include the index of the first duplicate encountered.


joshtemple:
Please review.

hramezani:
Thanks @joshtemple for this patch :+1: 

Unfortunately, we can not consider this patch because this isn't new in `1.10.x`.
You need to wait for `V2` and check it again.

PR #4520, opened 2022-09-15
Restore PyPI landing page

ofek:
Addresses https://github.com/pydantic/pydantic/pull/4473#discussion_r961944960 by using @hynek's [plugin](https://github.com/hynek/hatch-fancy-pypi-readme)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,14 +51,17 @@ Changelog = 'https://pydantic-docs.helpmanual.io/changelog'
 
 [tool.hatch.metadata.hooks.fancy-pypi-readme]
 content-type = 'text/markdown'
-fragments = [
-    { path = 'README.md' },
-    { path = 'HISTORY.md' },
-]
+
+[[tool.hatch.metadata.hooks.fancy-pypi-readme.fragments]]
+path = "README.md"
+
+[[tool.hatch.metadata.hooks.fancy-pypi-readme.fragments]]
+path = "HISTORY.md"
+pattern = "(.+?)<!-- package description limit -->"
 
 [[tool.hatch.metadata.hooks.fancy-pypi-readme.substitutions]]
-pattern = '#(\d+)'
-replacement = '[#\1](https://github.com/pydantic/pydantic/issues/\1)'
+pattern = '( +)#(\d+)'
+replacement = '\1[#\2](https://github.com/pydantic/pydantic/issues/\2)'
 
 [[tool.hatch.metadata.hooks.fancy-pypi-readme.substitutions]]
 pattern = '( +)@([\w\-]+)'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ['hatchling']
+requires = ['hatchling', 'hatch-fancy-pypi-readme>=22.5.0']
 build-backend = 'hatchling.build'
 
 [tool.hatch.version]
@@ -13,8 +13,6 @@ authors = [
     {name = 'Eric Jolibois', email = 'em.jolibois@gmail.com'},
     {name = 'Hasan Ramezani', email = 'hasan.r67@gmail.com'},
 ]
-license-files = { paths = ['LICENSE'] }
-readme = 'README.md'
 classifiers = [
     'Development Status :: 5 - Production/Stable',
     'Programming Language :: Python',
@@ -40,7 +38,7 @@ classifiers = [
 requires-python = '>=3.7'
 dependencies = [ 'typing-extensions>=4.1.0' ]
 optional-dependencies = { email = ['email-validator>=1.0.3'] }
-dynamic = ['version']
+dynamic = ['version', 'readme']
 
 entry-points.hypothesis = {_ = 'pydantic._hypothesis_plugin'}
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -51,6 +50,25 @@ Funding = 'https://github.com/sponsors/samuelcolvin'
...
 
+[tool.hatch.metadata.hooks.fancy-pypi-readme]
+content-type = 'text/markdown'
+fragments = [
+    { path = 'README.md' },
+    { path = 'HISTORY.md' },
+]
+
+[[tool.hatch.metadata.hooks.fancy-pypi-readme.substitutions]]
+pattern = '#(\d+)'
```
samuelcolvin, line 61: I think we need to require a space before the issue ID here like we do on usernames, or we could use a negative lookbehind to omit cases with `\w` directly before `#`. This caught us recently and the [second point on `v1.10.2`](https://pydantic-docs.helpmanual.io/changelog/) is currently messed up.

pyproject.toml
```diff
@@ -51,6 +50,25 @@ Funding = 'https://github.com/sponsors/samuelcolvin'
 Source = 'https://github.com/pydantic/pydantic'
 Changelog = 'https://pydantic-docs.helpmanual.io/changelog'
 
+[tool.hatch.metadata.hooks.fancy-pypi-readme]
+content-type = 'text/markdown'
+fragments = [
+    { path = 'README.md' },
+    { path = 'HISTORY.md' },
```
samuelcolvin, line 57: does this insert the new lines required?

samuelcolvin (commented):
Otherwise LGTM. I saw this plugin the other day and was thinking of using it.

Thanks @ofek.

And thanks @hynek for the plugin!

hynek:
Heh I live to serve. :) JFTR, your change log is getting a bit long, so if you need help to cut it to the latest release LMK.

samuelcolvin:
Ye, might be nice to include just releases after v1 or similar.

I think easiest solution would be to split `HISTORY.md`, an keep the older history just for https://pydantic-docs.helpmanual.io/changelog/.

Then we can add a link to the end of `HISTORY.md` along the lines of "For everything else, see here."

hynek:
Sure that works too, but it feels wrong to me to neuter the OG file. :) Check out how doc2dash does it – it's just a bunch of regexps:

- PyPI: https://pypi.org/project/doc2dash/
- changelog: https://github.com/hynek/doc2dash/blob/main/CHANGELOG.md
- the lines responsible: https://github.com/hynek/doc2dash/blob/f71bdc03fbb41fe51a92cd2843cc27f5c24a2513/pyproject.toml#L131-L145

samuelcolvin:
Makes sense. @ofek if you want to put a magic comment in `HISTORY.md` `<!-- package_description_limit -->` after V1, then ignore everything after that, that would be great.

pyproject.toml
ofek: Done!

pyproject.toml
ofek: Yup

ofek:
Address!

pyproject.toml
```diff
@@ -13,8 +13,6 @@ authors = [
     {name = 'Eric Jolibois', email = 'em.jolibois@gmail.com'},
     {name = 'Hasan Ramezani', email = 'hasan.r67@gmail.com'},
 ]
-license-files = { paths = ['LICENSE'] }
```
ofek, line 16: This is already matched by the PEP 639 defaults

samuelcolvin: Well what we had before was not working, see https://pypi.org/project/dnserver/ for an example.

samuelcolvin: That was with hatchling today.

ofek: `license = {file = 'LICENSE'}` refers to the raw text, `license-files` is for metadata inclusion. adding `license = 'MIT'` now though, thought it was defined

samuelcolvin:
Yes, but does this use the current actual license file, or a generic "MIT" label.

ofek:
In wheel it's under `pydantic-X.Y.Z.dist-info/licenses/`

ofek:
Capture

hynek:
FWIW these are separate problems:

A file called LICENSE (and some others which already [led to Fedora packagers pester me about it](https://github.com/hynek/hatch-fancy-pypi-readme/issues/8)) is always added courtesy of *wheel*: https://github.com/pypa/wheel/blob/5edc61d2c6535c380fa278bc83c578fbd5444167/src/wheel/bdist_wheel.py#L450

This is what's controlled by `license-files = { paths = ['LICENSE'] }` and `dnserver` wheel has the LICENSE file where it belongs:

Screenshot 2022-09-16 at 08 59 57@2x


---

The PyPI metadata is set using `project.license` and has only very recently been fixed by Hatchling (1.9 if I'm not mistaken).

You can see in https://pypi.org/project/doc2dash/ meta section that it shows correctly `License: MIT License (MIT)`. I have also [released a few packages with the full license in there](https://pypi.org/project/hatch-fancy-pypi-readme/22.4.0/).

The PEP-conformant way isn't `license = "MIT"` but `license = { text = "MIT" }`, otherwise e.g. VS Code will yell at you that `license` has to be a dict:

Screenshot 2022-09-16 at 08 58 11@2x

samuelcolvin:
Thanks so much for the thorough explanation @hynek.

@ofek, given all the subtleties here, let's go with the most vanilla and correct approach `license = { text = "MIT" }`.

ofek:
That's not what's happening here actually.

1. The only backend that uses https://github.com/pypa/wheel is setuptools. The code is here: https://github.com/pypa/hatch/blob/hatchling-v1.9.0/backend/src/hatchling/metadata/core.py#L614-L615
2. That section on PyPI is not what one may think. It's `"License: {license_classifier.split('::')[-1]} ({license['text']})"`. Notice a Hatch project: https://pypi.org/project/platformdirs/
3. Hatchling is the first to support [PEP 639](https://peps.python.org/pep-0639/) (specifically the end result of that Discourse thread) which allows for a string and deprecates the table. There are no more blockers for approval other than time.

hynek:
> 1. The only backend that uses https://github.com/pypa/wheel is setuptools. The code is here: https://github.com/pypa/hatch/blob/hatchling-v1.9.0/backend/src/hatchling/metadata/core.py#L614-L615

Ugh, back when I was on setuptools, they told me that they can't do anything because it's all in wheel. This will take a few years to untangle I guess… Anyhow, the effect stands since you're using the same patterns. :)

> 3. Hatchling is the first to support [PEP 639](https://peps.python.org/pep-0639/) (specifically the end result of that Discourse thread) which allows for a string and deprecates the table. There are no more blockers for approval other than time.

Cool! I'm not one to insist on packaging PEP peculiarities. I just noticed that my editor is yelling at me. :)

---

Anyhow, I don't mean to derail this any further.

ofek:
Shall I go back to using the new way then?

samuelcolvin:
I think this is good, but I need to review one more time. Will do on Tuesday.

samuelcolvin:
Thanks so much for this @ofek.

I've tweaked the setup slightly to make it a little more concise and IMO more readable, hope that's okay with you.

PR #4523, opened 2022-09-15
mypy plugin: Use annotation to determine if field is optional

hackedd:
## Change Summary

The mypy plugin would previously incorrectly mark fields as required when the `Field()` function was used but does not specify a default, resulting in _Missing named argument_ errors.

For example:

```python
class Model(BaseModel):
    bar: Optional[int] = Field()

m = Model()
```

## Checklist

### DIFF
--- a/changes/4523-hackedd.md
+++ b/changes/4523-hackedd.md
@@ -1,0 +1,1 @@
+Use annotation in mypy plugin to determine if field is optional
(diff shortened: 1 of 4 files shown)


hackedd:
My PR is ready, please review.

samuelcolvin:
This behaviour is really weird in V1.X, e.g. In the following case, `bar` is required:

```py

class Model(BaseModel):
    bar: int = Field()

m = Model()
```

In this situation, I'd rather not mess with the mypy extension in a patch release.

samuelcolvin:
Thanks for your PR, let's see what others think.

hramezani:
Agree with @samuelcolvin. It may break something else.

samuelcolvin:
Thanks so much @hackedd for the PR, I'd love you to contribute again in future.

However sadly, on this occasion, for the above reasons, I'm going to close this PR.

PR #4538, opened 2022-09-19
[1.10.X] Fix field regex with StrictStr type annotation

sisp:
There's an inconsistency between `constr(...)` and `ConstrainedStr`, which is the base class of `StrictStr`, with regard to the expected type of the `regex` argument/attribute which causes a field of type `StrictStr` with a `regex` constraint specified via the `Field(...)` class to raise an error:

```text
File "pydantic/main.py", line 340, in pydantic.main.BaseModel.__init__
File "pydantic/main.py", line 1076, in pydantic.main.validate_model
File "pydantic/fields.py", line 884, in pydantic.fields.ModelField.validate
File "pydantic/fields.py", line 1101, in pydantic.fields.ModelField._validate_singleton
File "pydantic/fields.py", line 1148, in pydantic.fields.ModelField._apply_validators
File "pydantic/class_validators.py", line 318, in pydantic.class_validators._generic_validator_basic.lambda13
File "pydantic/types.py", line 433, in pydantic.types.ConstrainedStr.validate
AttributeError: 'str' object has no attribute 'match'
```

I've fixed this inconsistency in a backwards compatible way by extending the type of `regex` in `ConstrainedStr` to also support a `str` value in addition to `Pattern[str]`.

I hope this fix will be considered despite the ongoing efforts towards Pydantic v2. :slightly_smiling_face:

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -403,16 +403,17 @@ class ConstrainedStr(str):
     min_length: OptionalInt = None
     max_length: OptionalInt = None
     curtail_length: OptionalInt = None
-    regex: Optional[Pattern[str]] = None
+    regex: Optional[Union[str, Pattern[str]]] = None
     strict = False
 
     @classmethod
     def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
+        regex = cls._regex()
         update_not_none(
             field_schema,
             minLength=cls.min_length,
             maxLength=cls.max_length,
-            pattern=cls.regex and cls.regex.pattern,
+            pattern=regex and regex.pattern,
         )
 
     @classmethod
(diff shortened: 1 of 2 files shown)


pydantic/types.py
```diff
@@ -429,12 +430,17 @@ def validate(cls, value: Union[str]) -> Union[str]:
         if cls.curtail_length and len(value) > cls.curtail_length:
             value = value[: cls.curtail_length]
 
-        if cls.regex:
-            if not cls.regex.match(value):
-                raise errors.StrRegexError(pattern=cls.regex.pattern)
+        regex = cls._regex()
+        if regex:
+            if not regex.match(value):
```
samuelcolvin, line 435: I'm afraid this might be slow since the compiled regex might not be cached.

Better to do the following which I think should work in both cases 

```suggestion
        if cls.regex:
            if not re.match(cls.regex, value):
```

samuelcolvin (commented):
otherwise happy to accept this.

pydantic/types.py
```diff
@@ -403,16 +403,17 @@ class ConstrainedStr(str):
     min_length: OptionalInt = None
     max_length: OptionalInt = None
     curtail_length: OptionalInt = None
-    regex: Optional[Pattern[str]] = None
+    regex: Optional[Union[str, Pattern[str]]] = None
     strict = False
 
     @classmethod
     def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
+        regex = cls._regex()
```
samuelcolvin, line 411: you'll therefore still need an `isinstance` check here.

sisp:
@samuelcolvin Thanks for your review feedback! I think I've made the changes accordingly which should improve regex caching. Happy to receive your feedback again. :slightly_smiling_face:

samuelcolvin (approved):
LGTM.

@hramezani please review and check you're happy with this.

hramezani (approved):
LGTM :+1: 

Thanks @sisp

PR #4542, opened 2022-09-20
Remove superfluous (???) files from sdist

samuelcolvin:
Will this, a bunch of files are removed from the sdist (`.tar.gz`) file which will be uploaded to PyPI.

## Question

* Is this correct?
* should tests still be included?
* I'm excluding `/docs/`, I think this is better, is there a strong argument against this?

With this changes, the following files are included in `pydantic-2.0.0.dev0.tar.gz`:
```
pydantic-2.0.0.dev0
├── .gitignore
├── HISTORY.md
├── LICENSE
├── Makefile
├── PKG-INFO
├── pydantic
│  ├── __init__.py
│  ├── _internal
│  │  ├── __init__.py
│  │  ├── ... 3 more files
│  ├── ... 23 more files
├── pyproject.toml
├── README.md
├── requirements
│  ├── 10 files
└── tests
   ├── mypy
   │  ├── 34 files
   ├── pyright
   │  ├── ... 2 files
   ├── ... 34 files
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,17 +2,18 @@
 requires = ['hatchling', 'hatch-fancy-pypi-readme>=22.5.0']
 build-backend = 'hatchling.build'
 
-[tool.hatch]
-version = { path = 'pydantic/version.py' }
-build = { exclude = [
+[tool.hatch.version]
+path = 'pydantic/version.py'
+
+[tool.hatch.build]
+exclude = [
     '/.github/**',
     '/changes/**',
     '/docs/**',
     '/mkdocs.yml',
     '/.pre-commit-config.yaml',
-    '/.gitignore',
     '/setup.py',
-] }
+]
 
 [project]
 name = 'pydantic'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,14 +5,16 @@ build-backend = 'hatchling.build'
 [tool.hatch.version]
 path = 'pydantic/version.py'
 
-[tool.hatch.build]
-exclude = [
-    '/.github',
-    '/changes',
-    '/docs',
-    '/mkdocs.yml',
-    '/.pre-commit-config.yaml',
-    '/setup.py',
+[tool.hatch.build.targets.sdist]
+# limit which files are included in the sdist (.tar.gz) asset,
+# see https://github.com/pydantic/pydantic/pull/4542
+include = [
+    '/README.md',
+    '/HISTORY.md',
+    '/Makefile',
+    '/pydantic',
+    '/tests',
+    '/requirements',
 ]
 
 [project]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,9 +7,9 @@ path = 'pydantic/version.py'
 
 [tool.hatch.build]
 exclude = [
-    '/.github/**',
-    '/changes/**',
-    '/docs/**',
+    '/.github',
+    '/changes',
+    '/docs',
     '/mkdocs.yml',
     '/.pre-commit-config.yaml',
     '/setup.py',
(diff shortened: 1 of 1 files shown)


samuelcolvin:
@ofek with your knowledge of packaging, does this look correct to you?

@davidbrochart any idea if this will work okay with conda-forge?

please review.

pyproject.toml
```diff
@@ -5,6 +5,16 @@ build-backend = 'hatchling.build'
...
 path = 'pydantic/version.py'
 
+[tool.hatch.build]
+exclude = [
+    '/.github/**',
+    '/changes/**',
+    '/docs/**',
+    '/mkdocs.yml',
+    '/.pre-commit-config.yaml',
+    '/setup.py',
```
samuelcolvin, line 15: excluding `.gitignore` seems to be ignored :shrug:.

davidbrochart:
> @davidbrochart any idea if this will work okay with conda-forge?

Yes, we will just need to add `hatchling` to the host requirements.

samuelcolvin:
great, thanks.

pyproject.toml
```diff
@@ -5,6 +5,16 @@ build-backend = 'hatchling.build'
 [tool.hatch.version]
 path = 'pydantic/version.py'
 
+[tool.hatch.build]
+exclude = [
+    '/.github/**',
+    '/changes/**',
+    '/docs/**',
```
ofek, line 12:
```suggestion
    '/.github',
    '/changes',
    '/docs',
```

samuelcolvin:
@mgorny (and anyone else who's involved in distributing pydantic downstream - linux distros etc.) is this okay with you?

Do you / would you, use the `docs/` directory?

mgorny:
We're not using sdists, we're fetching autogenerated archives from GitHub directly, so it shouldn't affect us.

samuelcolvin:
Interesting, thanks.

mgorny:
No problem, and thanks for asking. It's probably not something specific to pydantic, we're using GitHub archives whenever possible because sdists are missing tests too often. At this point, not even trying them saves us time. Not to mention that pypi URLs are just horrible.

samuelcolvin:
So (although gentoo doesn't use sdist), you would say we're right to include tests in the sdist and right (or at least it's okay) to omit docs?

mgorny:
If you asked me a year ago, I'd say that sdist should include both tests and docs (in source form) but I have non-Python background and I look at sdist as something akin to autotools' `make dist`, i.e. something you should be able to use to build, test and install everything's that useful about the package. As packagers, we find tests useful, and I know that many Gentoo users do appreciate offline documentation.

That said, I know that many people using Python see sdists as source material for `pip install` only, and prefer that they are as small as possible.

In other words, I see the points of both sides here and I don't really have a 100% good answer here. You can always try to satisfy everyone and e.g. push minimal sdist to pypi, and "fuller" sdist to GitHub ;-).

ofek:
> you would say we're right to include tests in the sdist and right (or at least it's okay) to omit docs?

That's what I do

pyproject.toml
```diff
@@ -5,6 +5,16 @@ build-backend = 'hatchling.build'
 [tool.hatch.version]
 path = 'pydantic/version.py'
 
+[tool.hatch.build]
+exclude = [
```
ofek, line 9: maybe instead:

```toml
include = [
    '/pydantic',
    '/tests',
    '/Makefile',
]
```

ofek, line 9: maybe instead:

```toml
[tool.hatch.build.targets.sdist]
include = [
    '/pydantic',
    '/tests',
    '/Makefile',
]
```

https://hatch.pypa.io/latest/plugins/builder/sdist/#default-file-selection

samuelcolvin: done.

samuelcolvin:
thanks everyone, merging this.

PR #4578, opened 2022-10-01
Use isinstance enum validator for 0-members enums

AtaiAmbus:
Bugfix for 1.10.2.

Tiny change to solve a left-over problem in from #1735 where "middle" enum subclasses used to override or provide methods instead of setting members could not be validated


AtaiAmbus:
It seems I was very much not thorough enough. Whoops.

PR #4589, opened 2022-10-05
Fix coercion of dict items with `Mapping` type

### LINKED ISSUE #4588: Incorrect coercion of dict items with `Mapping` type
### Initial Checks

### Description

When a model attribute is declared using the `Mapping` type and a list of tuples (i.e. `dict` items) is passed, the data is not coerced to a builtin `dict` but incorrectly to a list of the keys. When using a `Dict` type instead, it works as expected.

The example code below results in the following error:

```
>       assert repr(m) == "Model(x={'one': 1, 'two': 2}, y={'one': 1, 'two': 2})"
E       assert "Model(x={'on...one', 'two'])" == "Model(x={'on...1, 'two': 2})"
E         - Model(x={'one': 1, 'two': 2}, y={'one': 1, 'two': 2})
E         ?                                 ^     ---       ^^^^
E         + Model(x={'one': 1, 'two': 2}, y=['one', 'two'])
E         ?
```

### Example Code

```Python
from typing import Any, Dict, Mapping

def test_typing_coercion_dict_mapping_items():
    class Model(BaseModel):
        x: Dict[Any, Any]
        y: Mapping[Any, Any]

    v = [['one', 1], ['two', 2]]
    m = Model(x=v, y=v)
    assert repr(m) == "Model(x={'one': 1, 'two': 2}, y={'one': 1, 'two': 2})"
```

```Text
pydantic version: 1.10.2
       pydantic compiled: False
            install path: $HOME/dev/pydantic/pydantic
          python version: 3.9.13 (main, Sep  2 2022, 22:36:50)  [GCC 9.4.0]
                platform: Linux-5.13.0-48-generic-x86_64-with-glibc2.31
```

### Affected Components

sisp:
I've fixed a bug related to coercing dict items (a list of pairs) when using the `Mapping` type. It works fine with the `Dict` type but fails with the `Mapping` time which is incorrect and inconsistent.

Fixes #4588.

I hope this fix will be considered despite the ongoing efforts towards Pydantic v2. :slightly_smiling_face:

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -1042,7 +1042,7 @@ class ModelField(Representation):
         """
         original_cls = original.__class__
 
-        if original_cls == dict or original_cls == Dict:
+        if original_cls in {dict, Dict, list, List, tuple, Tuple}:
             return converted
         elif original_cls in {defaultdict, DefaultDict}:
             return defaultdict(self.type_, converted)
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -2107,6 +2107,16 @@ def test_typing_coercion_dict():
     assert repr(m) == "Model(x={'one': 1, 'two': 2})"
 
 
+def test_typing_coercion_dict_mapping_items():
+    class Model(BaseModel):
+        x: Dict[Any, Any]
+        y: Mapping[Any, Any]
+
+    v = [['one', 1], ['two', 2]]
+    m = Model(x=v, y=v)
+    assert repr(m) == "Model(x={'one': 1, 'two': 2}, y={'one': 1, 'two': 2})"
+
+
 def test_typing_non_coercion_of_dict_subclasses():
     KT = TypeVar('KT')
     VT = TypeVar('VT')


samuelcolvin:
Humm, not sure what's up with fastapi, is it related to this change?

I'm inclined to refuse this since I worry that it could have subtle side effects. Changes in patch releases should be as small as possible and fix significant, obvious bugs.

At the very least we need a bug where this is discussed and confirmation from others that this needs fixing

sisp:
Well, I think it's not intuitive, unexpected and inconsistent with how the `Dict` annotation behaves, and since `Mapping` is just an abstract and immutable type for, e.g., a concrete type `Dict`, they should behave identically. Wouldn't you agree? :thinking:

I've stumbled over this bug while trying to improve typing in https://github.com/copier-org/copier/pull/832 by preferring abstract container types over concrete ones.

sisp:
I don't think the failing FastAPI workflow is not related to this PR:

```
>       warnings.warn(TrioDeprecationWarning(msg), stacklevel=stacklevel)
E       trio.TrioDeprecationWarning: trio.MultiError is deprecated since Trio 0.22.0; use BaseExceptionGroup (on Python 3.11 and later) or exceptiongroup.BaseExceptionGroup (earlier versions) instead (https://github.com/python-trio/trio/issues/2211)
```

`pytest` is configured to treat warnings as errors unless they are explicitly ignored:

https://github.com/pydantic/pydantic/blob/da0b756e523ddb3db63f22d90ac34dc2fcf9f55f/setup.cfg#L4-L5

[Trio v0.22.0 has been released](https://trio.readthedocs.io/en/stable/history.html#trio-0-22-0-2022-09-28) about a week ago in which `MultiError` has been deprecated which matches the error we see in CI. Since FastAPI is installed with non-locked dependencies, [Trio is installed as a dependency of AnyIO](https://github.com/tiangolo/fastapi/blob/c6aa28bea2f751a91078bd8d845133ff83f352bf/pyproject.toml#L68) whose latest release v3.6.1 [installs Trio v0.16+](https://github.com/agronholm/anyio/blob/48efdec45e70a833cc939c1d2752f24e29d1bf0b/setup.cfg#L55) (i.e. v0.22.0 now). The FastAPI workflow apparently hadn't failed before because Trio v0.21 or less had been used which didn't emit that deprecation warning.

samuelcolvin:
I agree we would do it differently if we started from scratch (in v2 lists won't be allowed as inputs to either mapping or dicts), but that's not enough to change something in a patch release IMHO.

Look at the amount of stuff that would be different in python itself, but is not changed to avoid compatibility issues.

Some might rely on this quirk as a feature. I'm therefore inclined to refuse, but I'll wait for @PrettyWood and @hramezani's opinions?

PrettyWood:
I would keep this in mind for v2

sisp:
Can't wait to get my hands dirty with v2. 😜

samuelcolvin:
> Can't wait to get my hands dirty with v2. stuck_out_tongue_winking_eye

Me too, going as fast as I can :sweat:.

samuelcolvin:
I get what you mean entirely, no offence taken. Thanks for being understanding.

PR #4610, opened 2022-10-11
fix fastapi tests

samuelcolvin:
See https://github.com/pydantic/pydantic/pull/4589#issuecomment-1268063107, thanks @sisp for pointing this.


samuelcolvin:
replaced by https://github.com/tiangolo/fastapi/pull/5483

PR #4641, opened 2022-10-22
Allow dicts to have both patternProperties and additionalProperties

jparise:
Support for `patternProperties` was introduced in #332, but that logic unfortunately made `patternProperties` and `additionalProperties` mutually exclusive. JSON Schema supports their combination:

https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties

This prevented well-typed code generation for dictionaries that use regex-constrained string keys.

### DIFF
--- a/changes/4641-jparise.md
+++ b/changes/4641-jparise.md
@@ -1,0 +1,1 @@
+Allow dict schemas to have both `patternProperties` and `additionalProperties`
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -490,7 +490,7 @@ def field_type_schema(
             # Dict keys have a regex pattern
             # items_schema might be a schema or empty dict, add it either way
             f_schema['patternProperties'] = {regex.pattern: items_schema}
-        elif items_schema:
+        if items_schema:
             # The dict values are not simply Any, so they need a schema
             f_schema['additionalProperties'] = items_schema
     elif field.shape == SHAPE_TUPLE or (field.shape == SHAPE_GENERIC and not issubclass(field.type_, BaseModel)):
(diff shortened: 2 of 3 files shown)


jparise:
This patch is for the v1.10.x branch. The `main` (v2) patch is #4642.

jparise:
The failing FastAPI test in this branch appears to be unrelated to this change. It's dying in an SQLAlchemy code path (`sqlalchemy/engine/cursor.py:164: ValueError`).

matejetz:
i think this introduced inconsistency regarding the json schema vs pydantic validation behavior.

in the following test, pydantic rightfully raises a ValidationError
```python
    class Model(BaseModel):
        a: Dict[constr(regex="^text$"), str]

    json_schema = Model.schema()

        Model(a={'foo': 'bar'})  # correctly raises
```

however, the json schema produced by `Model.schema()` will now validate the following, as `'additionalProperties': { 'type': 'string'}`  just allows any string properties.
```json
{
  "a": {
    "foo": "bar"
  }
}
```

i also doubt this is the expected behavior when using a regex-constrained dict key. instead, imho, `additionalProperties` should be `false` by default and might possibly be set by passing some kwarg to `Field`.

samuelcolvin:
What version are you using? Might be worth trying on main.

We can't really make a change to this in v1, but we can get it right in V2.

matejetz:
sorry, i forgot to include that my comment concerns v1, which this PR was also targeting.
why do you think you can't change this, i'd almost go as far and classify this as a bug or at least unwanted/ unexpected behavior. 

i did not have a chance to take a look at v2 yet but will definitely do, looks very promising. love the lib! 👍

ElectronicRU:
Yes, this seems like a wrong change to me - it makes JSON schema more permissive than the actual schema is. Can @samuelcolvin elaborate why that's not changeable in v1?

PR #4642, opened 2022-10-22
Allow dicts to have both patternProperties and additionalProperties

jparise:
Support for `patternProperties` was introduced in #332, but that logic unfortunately made `patternProperties` and `additionalProperties` mutually exclusive. JSON Schema supports their combination:

https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties

This prevented well-typed code generation for dictionaries that use regex-constrained string keys.

### DIFF
--- a/changes/4642-jparise.md
+++ b/changes/4642-jparise.md
@@ -1,0 +1,1 @@
+Allow dict schemas to have both `patternProperties` and `additionalProperties`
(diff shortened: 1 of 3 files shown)


jparise:
This patch is for the `main` branch (v2). The v1.10.x patch is #4641.

samuelcolvin:
All this logic will be rewritten for V2, so we can't accept this. #4516 will be merged pretty soon, then we can start working on schema for V2.

Possibly we could accept this as a bug fix for v1, please create a new PR we'll review.

jparise:
> All this logic will be rewritten for V2, so we can't accept this. #4516 will be merged pretty soon, then we can start working on schema for V2.

Understood. I opened this PR for completeness after I opened #4641 for v1.

> Possibly we could accept this as a bug fix for v1, please create a new PR we'll review. 

Cool. I'm primarily interested in fixing this in v1 (#4641) and then helping to guard against a regression in v2.

PR #4647, opened 2022-10-23
chore: Update outdated LICENSE year

bluzir:
Updated the outdated copyright year to the present. 
So there will not be a need for any further license year updates.


samuelcolvin:
Thanks for the suggestion.

I have no idea about the legality or importance of this, but since, by definition if pydantic stops being updated, this will become wrong, I'd rather update it manually each year.

PR #4663, opened 2022-10-27
Support extracting attributes from the parent namespace

samuelcolvin:
Working on workaround for problems with python type hints namespaces, see #2678.

I'm not yet sure if this is a good idea, it's still confusing since intermediate parent namespaces are ignored.

Still this is an interesting idea.

Will need to be rebased to main once #4516 is merged.

### DIFF
--- a/tests/test_forward_ref.py
+++ b/tests/test_forward_ref.py
@@ -749,3 +749,98 @@ def test_force_rebuild():
     assert Foobar.__pydantic_model_complete__ is True
     assert Foobar.model_rebuild() is None
     assert Foobar.model_rebuild(force=True) is True
+
+
+def test_nested_annotation(create_module):
+    module = create_module(
+        # language=Python
+        """
+from __future__ import annotations
+from pydantic import BaseModel
+
+def nested():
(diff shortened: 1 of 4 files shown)


tests/test_forward_ref.py
```diff
@@ -749,3 +749,94 @@ class Foobar(BaseModel):
...
+    def module():
+        from annotated_types import Gt
+        from typing_extensions import Annotated
+
+        from pydantic import BaseModel
+
+        Foobar = Annotated[int, Gt(0)]  # noqa: F841
+
+        def nested():
+            Foobar = Annotated[int, Gt(10)]
```
hramezani, line 810:
```suggestion
            Foobar = Annotated[int, Gt(10)]  # noqa: F841
```

hramezani: @samuelcolvin I've just fixed the `CI / lint` problem by merging this.
Now it is green again.

samuelcolvin: great, thanks.

samuelcolvin:
might need #4735

PR #4679, opened 2022-10-28
fixed error in example code from documentation

rifatrakib:
floating point number inside a string cannot be converted to an integer directly in Python 3.9.6 because of which the example code snippet produces an error in the output.


samuelcolvin:
Pydantic converts strings to numbers where possible, this is tested thorough. If you've seen strange behavior, please create an issue.

Issue #4689, opened 2022-10-29
Add support for `postgresql+psycopg` as allowed scheme to use with PostgreDsn


morian:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this feature is missing

### Description

SQLAlchemy 2.0 (currently in beta) now supports [psycopg](https://www.psycopg.org/psycopg3/) (a.k.a. psycopg3).

This was introduced as a new dialect by the following commit:
https://github.com/sqlalchemy/sqlalchemy/commit/5eb407f84bdabdbcd68975dbf76dc4c0809d7373

The connection string appears to be `postgresql+psycopg://user:password@host:port/dbname[?key=value&key=value...]` but unfortunately it is not recognized by pydantic as a valid scheme by `PostgresDsn` from `networks.py`.

Would it be possible to add it as an allowed scheme for PostgresDsn?

By looking at the code and while not familiar with cockroachdb, it seems like it would also benefit from this support:
https://www.cockroachlabs.com/docs/stable/build-a-python-app-with-cockroachdb-psycopg3.html

Thanks!

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
- [ ] [Data validation/parsing](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://pydantic-docs.helpmanual.io/usage/exporting_models/) - `.dict()` and `.json()`
- [ ] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)
- [ ] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)
- [ ] [Model Config](https://pydantic-docs.helpmanual.io/usage/model_config/)
- [ ] [Field Types](https://pydantic-docs.helpmanual.io/usage/types/) - adding or changing a particular data type
- [ ] [Function validation decorator](https://pydantic-docs.helpmanual.io/usage/validation_decorator/)
- [ ] [Generic Models](https://pydantic-docs.helpmanual.io/usage/models/#generic-models)
- [ ] [Other Model behaviour](https://pydantic-docs.helpmanual.io/usage/models/) - `construct()`, pickling, private attributes, ORM mode
[...]

PR #4690, opened 2022-10-30
Add postgresql+psycopg as allowed scheme for PostgreDsn

morian:
Merge request for #4689.

Support is limited to `PostgreDsn` for now.
CockroachDsn was introduced in `pydantic` by #3830 based on what was supported by [sqlalchemy-cockroachdb](https://github.com/cockroachdb/sqlalchemy-cockroachdb).
This project does not show any sign of support for psycopg3 so far, so this was not included in this patch.


samuelcolvin:
I've also cherry picked this onto the v1.10 branch.

PR #4709, opened 2022-11-02
Move to `ruff` for linting

samuelcolvin:


pydantic/main.py
```diff
@@ -31,7 +31,7 @@
 
     AnyClassMethod = classmethod[Any]
     TupleGenerator = typing.Generator[tuple[str, Any], None, None]
-    Model = typing.TypeVar('Model', bound='BaseModel')
+    Model = typing.TypeVar("Model", bound='BaseModel')
```
samuelcolvin, line 34: @charliermarsh it looks like this incorrect quote example is not being picked up by ruff (at least for me locally).

Any idea what's going on?

charliermarsh: Will take a look!

charliermarsh: Ah, the `Q` codes aren't enabled by default -- only `E` and `F`.

You can change the `pyproject.toml` like this:

```
[tool.ruff]
line-length = 120
select = ["E", "F", "Q"]
flake8-quotes = {inline-quotes = 'single', multiline-quotes = 'double'}
```

Or, equivalently:

```
[tool.ruff]
line-length = 120
extend-select = ["Q"]
flake8-quotes = {inline-quotes = 'single', multiline-quotes = 'double'}
```

samuelcolvin: that worked, thanks. Might be worth adding that to your example of `flake8-quotes`.

hramezani (approved):
LGTM :+1:

pydantic/main.py
charliermarsh: Good call. (Ruff needs proper docs.)

charliermarsh:
Thanks @samuelcolvin and @hramezani! Ping whenever you have requests or run into issues. Happy to do custom work to support Pydantic as an early adopter!

PR #4714, opened 2022-11-03
replace U+2019 with U+0027 in docstrings

adriangb:


samuelcolvin:
No idea how that happened, good catch.

@charliermarsh (hope you don't mind me pinging you in a pr like this?) Do you think catching this and similar unicode mistakes would be a candidate for a ruff rule? If so I'd be happy to help with a list of common cases.

adriangb:
FWIW I only noticed because VSCode alerted me to it. I'm not sure where their list of characters to alert on is though.

charliermarsh:
@samuelcolvin - (No I don't mind at all!) I think it makes sense to include this in Ruff (perhaps even auto-fixable). Will track in https://github.com/charliermarsh/ruff/issues/565, thank you for filing.

PR #4734, opened 2022-11-08
Fix improper parsing of Enum document

mnixry:
To properly display document of an Enum, it requires to `inspect.cleandoc` before put into schema.
It had worked perfectly in 1.9.x, but it seems #4267 broke its behavior.

Expected behavior:

Actual behavior:

Enum code:
```python
class CommentType(IntEnum):
    """
    评论来源类型

    | **数值** | **含义** |
    |---|---|
    | 1 | 视频 |
    | 12 | 文章 |
    | 11 | 含图片的动态 |
    | 17 | 不含图片的动态 |
    | 14 | 音乐 |
    | 19 | 歌单 |
    """

    VIDEO = 1
    ARTICLE = 12
    DYNAMIC_PICTURE = 11
    DYNAMIC_TEXT = 17
    AUDIO = 14
    AUDIO_LIST = 19

```

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -1,3 +1,4 @@
+import inspect
 import re
 import warnings
 from collections import defaultdict
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -618,7 +618,7 @@ def enum_process_schema(enum: Type[Enum], *, field: Optional[ModelField] = None)
         'title': enum.__name__,
         # Python assigns all enums a default docstring value of 'An enumeration', so
         # all enums will have a description field even if not explicitly provided.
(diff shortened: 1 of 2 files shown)


samuelcolvin:
The inspect import needs to be local, not global.

samuelcolvin:
Otherwise lgtm.

samuelcolvin:
Oh, and you need to add a test.

samuelcolvin:
> I'm a bit curious about this, why `inspect` import needs to be local?

See #1132 and related discussions, e.g. on #1127 - basically we use relative imports to keep import time of pydantic to a minimum.

samuelcolvin:
thanks so much.

PR #4748, opened 2022-11-10
Use 3.10 for all reqs resolution

### LINKED ISSUE #4745: Invalid requirements
### Initial Checks

### Description

https://github.com/pydantic/pydantic/blob/4d773bb96b0ef23bd5258be22ac37875c48d98d8/requirements/testing.txt#L15
https://github.com/pydantic/pydantic/blob/4d773bb96b0ef23bd5258be22ac37875c48d98d8/requirements/docs.txt#L38

this makes all.txt impossible to instal

options:
- add a layer of indirection and use pip constraints (keeps pip-tools)
- Use a package manager with support for dependency groups and lockfiles

### Example Code

```Python
N/A
```

```Text
N/A
```

### Affected Components

samuelcolvin:
fix #4745


adriangb:
So it is a bit more complicated than this, and also not. Since I don't know what workflow you're using to generate these I assume you're manually running `pip-compile <file>.in` for each file. Here's the scenarios that play out:

## Locking without any existing `.txt` files
### Totally incompatible versions in `docs.in` and `tests.in`
For example say you go and add `importlib-metadata=~4` in `docs.in` and `importlib-metadata=~5` in `tests.in`. `pip-compile` will happily compile them individually, resulting in an invalid `all.txt` and you wouldn't know until you try to install it.
### Compatible versions
This works fine 👍🏻 

## Re-locking after an edit with an existing `.txt` file
This is quite likely to result in incompatible/diverging `.txt` files. Say both `tests.txt` and `docs.txt` have `importlib-metadata==4.2.0`. You go and add a dependency to `tests.in` that depends on `importlib-metadata>5` so when you run `pip-compile tests.in` you end up with `importlib-metadata==5.0.0` in `tests.txt`. Since `docs.txt` still has `importlib-metadata==4.2.0`, `all.txt` is once again unusable.


So TLDR this workflow will work as long as:
1. You make sure you never add incompatible dependencies, which you can easily test by trying to install `all.txt` after creating lock files.
2. You always delete the  generated `.txt` files before running `pip-compile`.

samuelcolvin:
So should we delete the `.txt` files in the `refresh-lockfiles` Makefile recipe before regenerating?

adriangb:
Yes. Can we also make it do `pip install --dry-run -r requirements/all.txt`?

samuelcolvin:
done.

adriangb (approved):

PR #4758, opened 2022-11-15
Renaming BaseModel

pmithrandir:
Hello

BaseModel is an import from pydantic.
Using that same name for the Base from sqlachemy can be misleading, and not aligned to the other examples.

I therefore suggest to use the same naming that the other example.

Pierre

# NOTICE


samuelcolvin:
thanks so much.

Issue #4670, opened 2022-10-27
V2: dataclasses


samuelcolvin:
dataclass validation should get **MASSIVELY** easier and more performant since we have the `arguments` validator in pydantic-core which means we can do validation before constructing the dataclass. Lots of examples in the docs [here](https://github.com/pydantic/pydantic-core/blob/main/tests/validators/test_arguments.py).

I think we should remove all the `__post_init_post_parse__` shit, and just use validation functions and the standard stuff offered by dataclasses.

@PrettyWood as the master of dataclasses, would be willing to work on this? If not, I quite understand, please remove yourself as assignee and hopefully someone else can take over.

Either way, I'd love your insights on what we should do.

dmontagu:
(Also, https://github.com/pydantic/pydantic/issues/5271 tracks a few remaining things.)

PR #4772, opened 2022-11-21
Handle dataclass initialization with alias

ddoyoon:
Hello! I initially came across this problem through a FastAPI issue (tiangolo/fastapi#5093) where `pydantic.dataclasses` did not work as expected when a field has an alias.
Basically when the field has an alias, users are unable to set the field using the field's name nor the alias.

I tried to create a fix for this but I am fairly new to pydantic - please let me know if this is not a bug but an intended behavior. Thank you!


samuelcolvin:
We won't suppor this in V1.10, but I believe this should be well supported in V2.

samuelcolvin:
Sorry, submitted too early.

Thanks so much for your contribution, you might be interested in #4670, if you have any feedback on how dataclasses should work in v2, best to add it there.

PR #4775, opened 2022-11-22
Fix code typo in docs

ventaquil:
Tiny typo fix.

skip change file check

### DIFF
--- a/docs/examples/postponed_annotations_broken.py
+++ b/docs/examples/postponed_annotations_broken.py
@@ -4,7 +4,7 @@ from pydantic.errors import ConfigError
 
 
 def this_is_broken():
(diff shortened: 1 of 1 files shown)


hramezani:
skip change file check

hramezani (approved):

hramezani:
Thanks @ventaquil for this patch.

Could you please create a PR against `main` and introduce this change there as well?

@samuelcolvin Do we have anything to cherry-pick to `main`?

samuelcolvin:
> @samuelcolvin Do we have anything to cherry-pick to main?

No, we only have cherry-picking from main onto the v1.10 branch, not the other way. Ideally PRs that apply to both should target main, but I guess we could add another cherry-pick setup for the other way :sleeping:.

PR #4777, opened 2022-11-23
ByteSize.human_readable() should not have decimal point with B suffix

lieryan:
Before this change:

```
>>> value = ByteSize(5)
>>> value.human_readable()
5.0B
```

This PR removes the `.0` suffix which does not make sense as ByteSize can only represent whole number of bytes anyway:

```
>>> value = ByteSize(5)
>>> value.human_readable()
5B
```

When there is SI or Binary suffix, the current behavior of having a single decimal point of precision remains the same.


linpan:
meaningful

samuelcolvin (approved):
LGTM.

@dgasmith I think you wrote the original code here, are you happy with this?

I'm not sure about why the failure affecting docs has started happening, I'll look into it.

dgasmith (approved):
Nice, LGTM!

hramezani (approved):

hramezani:
Thanks @lieryan

PR #4781, opened 2022-11-24
[1.10.x] Properly encode model and dataclass default for schema

### LINKED ISSUE #4774: `schema_json` can't encode default value of`dict[UUID, Any]` and raises `TypeError`
### Initial Checks

### Description

When a model has default value of `{uuid4(): "..."}`, and being a dafault value of another model, `json_schema` fails to generate schema with an error.

### Example Code

```Python
from uuid import uuid4

from pydantic import UUID4


class Service(BaseModel):
    urls: dict[UUID4, str] = {uuid4(): "https://example.com"}


class Settings(BaseModel):
    service: Service = Service()


print(Settings.schema_json(indent=4, ensure_ascii=False))
```

```py
  File "/Users/bobronium/Library/Application Support/JetBrains/PyCharm2022.3/scratches/scratch_131.py", line 15, in <module>
    print(Settings.schema_json(indent=4, ensure_ascii=False))
  File "pydantic/main.py", line 674, in pydantic.main.BaseModel.schema_json
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  [Previous line repeated 1 more time]
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/json/encoder.py", line 376, in _iterencode_dict
    raise TypeError(f'keys must be str, int, float, bool or None, '
TypeError: keys must be str, int, float, bool or None, not UUID
```

### This works, though

```py
from uuid import uuid4

from pydantic import UUID4


class Service(BaseModel):
    urls: dict[UUID4, str] = {uuid4(): "https://example.com"}


class Settings(BaseModel):
    service: Service  # <--------- Notice no default value here


print(Settings.schema_json(indent=4, ensure_ascii=False))
```
<details>

```json
{
    "title": "Settings",
    "type": "object",
    "properties": {
        "service": {
            "$ref": "#/definitions/Service"
        }
    },
    "required": [
        "service"
    ],
    "definitions": {
        "Service": {
            "title": "Service",
            "type": "object",
            "properties": {
                "urls": {
                    "title": "Urls",
                    "default": {
                        "ac762f77-1e77-4ff8-906a-fd63ca622938": "https://example.com"
                    },
                    "type": "object",
                    "additionalProperties": {
                        "type": "string"
                    }
                }
            }
        }
    }
}
```

</details>

```Text
pydantic version: 1.10.2
            pydantic compiled: True
                 install path: .venv/lib/python3.10/site-packages/pydantic
               python version: 3.10.1 (main, Jan 31 2022, 09:16:52) [Clang 13.0.0 (clang-1300.0.29.3)]
                     platform: macOS-12.4-arm64-arm-64bit
```

### Affected Components

- [X] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)

Bobronium:
Closes #4774

Fix was pretty straightforward, I guess we can have this in 1.10.x.

### DIFF
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -1,6 +1,7 @@
 import re
 import warnings
 from collections import defaultdict
+from dataclasses import is_dataclass
 from datetime import date, datetime, time, timedelta
 from decimal import Decimal
 from enum import Enum
@@ -971,7 +972,11 @@ def multitypes_literal_field_for_schema(values: Tuple[Any, ...], field: ModelFie
 
 
 def encode_default(dft: Any) -> Any:
-    if isinstance(dft, Enum):
+    from .main import BaseModel
+
+    if isinstance(dft, BaseModel) or is_dataclass(dft):
+        dft = cast(dict[str, Any], pydantic_encoder(dft))
+    elif isinstance(dft, Enum):
         return dft.value
     elif isinstance(dft, (int, float, str)):
         return dft
@@ -979,7 +984,8 @@ def encode_default(dft: Any) -> Any:
         t = dft.__class__
         seq_args = (encode_default(v) for v in dft)
         return t(*seq_args) if is_namedtuple(t) else t(seq_args)
-    elif isinstance(dft, dict):
+
+    if isinstance(dft, dict):
         return {encode_default(k): encode_default(v) for k, v in dft.items()}
     elif dft is None:
         return None
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -975,7 +975,7 @@ def encode_default(dft: Any) -> Any:
     from .main import BaseModel
 
     if isinstance(dft, BaseModel) or is_dataclass(dft):
-        dft = cast(dict[str, Any], pydantic_encoder(dft))
+        dft = cast('dict[str, Any]', pydantic_encoder(dft))
     elif isinstance(dft, Enum):
         return dft.value
     elif isinstance(dft, (int, float, str)):
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -976,6 +976,9 @@ def encode_default(dft: Any) -> Any:
 
     if isinstance(dft, BaseModel) or is_dataclass(dft):
         dft = cast('dict[str, Any]', pydantic_encoder(dft))
+
+    if isinstance(dft, dict):
+        return {encode_default(k): encode_default(v) for k, v in dft.items()}
     elif isinstance(dft, Enum):
         return dft.value
     elif isinstance(dft, (int, float, str)):
@@ -984,9 +987,6 @@ def encode_default(dft: Any) -> Any:
         t = dft.__class__
         seq_args = (encode_default(v) for v in dft)
         return t(*seq_args) if is_namedtuple(t) else t(seq_args)
-
-    if isinstance(dft, dict):
-        return {encode_default(k): encode_default(v) for k, v in dft.items()}
     elif dft is None:
         return None
     else:
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1528,7 +1528,7 @@ def test_model_default():
     from pydantic import BaseModel
 
     class Inner(BaseModel):
-        a: dict[Path, str] = {Path(): ''}
+        a: Dict[Path, str] = {Path(): ''}
 
     class Outer(BaseModel):
         inner: Inner = Inner()
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1525,8 +1525,6 @@ def test_dict_default():
 
 def test_model_default():
     """Make sure inner model types are encoded properly"""
-    from pydantic import BaseModel
-
     class Inner(BaseModel):
         a: Dict[Path, str] = {Path(): ''}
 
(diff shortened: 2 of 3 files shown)


Bobronium:
CI seems to be failing with installation error unrelated to the changes.

hramezani:
I created https://github.com/pydantic/pydantic/pull/4782 to fix the CI problem.
I am not sure about the solution

Bobronium:
Perhaps I can rebase `Bobronium:1.10.X-fix-schema-model-default` on `hramezani:setuptools`?

hramezani:
@Bobronium I would suggest waiting for @samuelcolvin to review and probably merge my PR. then you can rebase on top of main

hramezani:
FYI, I closed https://github.com/pydantic/pydantic/pull/4782 because it was on main and seems not a proper fix

hramezani:
As the problem got fixed in https://github.com/pydantic/pydantic/commit/78ca710fe769552d91c12adf972a9c88a2a1a90f, you can rebase your patch now!

Bobronium:
Please review :)

tests/test_schema.py
```diff
@@ -1523,6 +1523,39 @@ class UserModel(BaseModel):
     }
 
 
+def test_model_default():
+    """Make sure inner model types are encoded properly"""
+    from pydantic import BaseModel
+
```
hramezani, line 1529: I think we don't need this import here
```suggestion
```

pydantic/schema.py
```diff
@@ -971,15 +972,20 @@ def multitypes_literal_field_for_schema(values: Tuple[Any, ...], field: ModelFie
...
         return dft.value
     elif isinstance(dft, (int, float, str)):
         return dft
     elif isinstance(dft, (list, tuple)):
         t = dft.__class__
         seq_args = (encode_default(v) for v in dft)
         return t(*seq_args) if is_namedtuple(t) else t(seq_args)
-    elif isinstance(dft, dict):
+
+    if isinstance(dft, dict):
```
hramezani, line 988: Do we need this change here?

hramezani:
@Bobronium I left a suggestion and asked a question.

Please update

pydantic/schema.py
Bobronium: Yes. Notice the change above: we transform models and dataclasses to `dict` in first `if` and need this to be a separate one to work.

Bobronium:
Nice catch with suggestion, @hramezani! I answered the question as well.

Please review.

pydantic/schema.py
Bobronium: Any ideas how to make this more explicit and obvious, btw?

samuelcolvin: maybe just a comment with the above explanation.

samuelcolvin (commented):
Otherwise LGTM.

pydantic/schema.py
hramezani: I would add the first if as a separate if and move the `isinstance(dft, dict)` check to the top.

Here is my suggestion(I haven't tested it locally):
```python
def encode_default(dft: Any) -> Any:
    from .main import BaseModel

    if isinstance(dft, BaseModel) or is_dataclass(dft):
        dft = cast('dict[str, Any]', pydantic_encoder(dft))

    if isinstance(dft, dict):
        return {encode_default(k): encode_default(v) for k, v in dft.items()}
    elif isinstance(dft, Enum):
        return dft.value
    elif isinstance(dft, (int, float, str)):
        return dft
    elif isinstance(dft, (list, tuple)):
        t = dft.__class__
        seq_args = (encode_default(v) for v in dft)
        return t(*seq_args) if is_namedtuple(t) else t(seq_args)
    elif dft is None:
        return None
    else:
        return pydantic_encoder(dft)

```

Bobronium: Happy to accept suggestions on phrasing as well :)
```suggestion
    # we can end up here either from first condition or if none of above conditions were met
    if isinstance(dft, dict):
```

Bobronium: Thanks for suggestion, @hramezani, I like your approach! I think I was hesitant on reordering at first, but it seems like the best move here, indeed.

hramezani (approved):
LGTM :+1: 
Thanks @Bobronium

samuelcolvin:
Thanks so much.

I've also cherry-picked this into `main`, mostly so we don't forget this edge case when working on JSONSchema for V2.

PR #4782, opened 2022-11-24
Install setuptools before installing requirements in CI

hramezani:
Fixes the CI problem in https://github.com/pydantic/pydantic/pull/4781
Seems we need to install `setuptools` first.
I found the solution from https://github.com/pypa/packaging-problems/issues/573


hramezani:
close this one because the problem is on `1.10` and this seems not a proper fix

PR #4799, opened 2022-11-29
Remove trailing "```" from docs

Kludex:
I'm not sure how this interferes with the migration to V2... :eyes:

### DIFF
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -190,7 +190,3 @@ The callable is expected to mutate the schema dictionary *in-place*; the return 
 For example, the `title` key can be removed from the model's `properties`:
 
 {!.tmp_examples/schema_extra_callable.md!}
-
-
-
-```


hramezani (approved):

hramezani:
Thanks @Kludex

Kludex:
btw, the "This script is complete" and the "outputs" are in the same line, I'm not sure if this is intended...
 
image

hramezani:
I think yes, and the `Outputs` indicated the next block is the output of the execution.

Kludex:
But if you look at other pages, the "Outputs:" would have their own new line :eyes: 

Ref.: https://pydantic-docs.helpmanual.io/usage/validators/

Anyway... :wave: :runner: :dash:

Kludex:
> But I can't see Outputs on the page that you've referenced. I can only see (This script is complete, it should run "as is") (without Outputs:)

Based on this, I don't think I expressed myself very well.

> FYI, Outputs is used for presenting the JSON output of scripts. and will be appended to the script in

I got that, what I'm saying is that a new line is missing before `Outputs:`, because it looks like (based on the rest of the docs) that it's not intended to be in the same line as the "This script is complete..."

samuelcolvin:
Ye I think you're right, "Outputs:" should be a shown on a new line.

hramezani:
Thanks @Kludex for explaining. Yes, you are right.

@samuelcolvin I can take care of it.

PR #4802, opened 2022-11-30
Add newline before `Outputs:` when printing json output in docs

hramezani:
Fixes https://github.com/pydantic/pydantic/pull/4799#issuecomment-1330788495

Thanks @Kludex for reporting


samuelcolvin:
thanks so much.

Issue #4675, opened 2022-10-27
V2: Descriminated Unions


samuelcolvin:
Descriminated unions are getting lots of super powers in V2, probably best example right now is the tests in pydantic-core, see https://github.com/pydantic/pydantic-core/blob/main/tests/validators/test_tagged_union.py.

In particular, descriminated union desciminators can now be a path or a function.

How should we define these? I guess it can all be done with `Field` and `Annotated`?

smhavens:
Would it be alright for myself and @marcoo47 to request taking this one over as first time contributors? If so, what would you recommend for first steps, we've been reading over the repository code and documentation in the manual.

samuelcolvin:
Absolutely, have a go. You might also want to read through pydantic-core and understand how new descriminated unions work. I'll give some code examples of the final interface soon.

smhavens:
> 

Thank you so much!

marcoo47:
> 

Hi, my partner and I have read through most of the relevant documentation. Did you ever make the code examples for the final interface you mentioned?

samuelcolvin:
Interface should be as follows:

```py
from typing import Literal, Union, Annotated

from pydantic import BaseModel, Field
from pydantic.descriminators import Tag


class Cat(BaseModel):
    meows: int


class Dog(BaseModel):
    barks: float


class Lizard(BaseModel):
    scales: bool


class Model(BaseModel):
    pet: Union[
        Annotated[Cat, Tag('cat')],
        Annotated[Dog, Tag('dog')],
        Annotated[Lizard, Tag('lizard')],
    ] = Field(..., discriminator='pet_type')
    n: int
```

A few notes:
* `Tag` is just a dataclass used as the key in `tagged-union` `choices
* `Field(discriminator...)` can also now be a function as that's supported by pydantic-core
* There's a question over whether we can infer `Tag` in the case that all models have a literal with the same name, as we do now
* once https://github.com/pydantic/pydantic-core/pull/327 get's merged, `pydanticc-core` will support multiple keys to the same choice, but for now we need to either raise an error for that case and wait for that feature, or duplicate the union member with the different key - e.g. in the `pet_type: Literal['reptile', 'lizard']` case shown in the docs.

smhavens:
I was wondering how you would like paths and functions to work for discriminated unions. Would they be like the variables being used to determine class like above (a function within the list of classes/objects to be decided from) or an outside function that is run to determine the object instead?

[...]

PR #4827, opened 2022-12-09
Origin/discriminated union tests v2

marcoo47:
Half the tests are now passing for test_discriminated_union structure needs to be improveed to accomidate for nested oops of more than 1 level.

Due to time constraints we will need to pass the task on to someone else.

work in #4675

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -75,7 +75,7 @@ class ModelMetaclass(ABCMeta):
             validator_functions = _validation_functions.ValidationFunctions(bases)
             namespace['__pydantic_validator_functions__'] = validator_functions
 
-            # Errors for misuse of discriminated unions (check test_discriminated_unions)
+           # Errors for misuse of discriminated unions (check test_discriminated_unions)
             for name, value in namespace['__annotations__'].items():
                 #Ensure value is intended as a discriminated union
                 if isinstance(namespace.get(name), FieldInfo) and namespace[name].discriminator is not None:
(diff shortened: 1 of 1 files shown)


pydantic/main.py
```diff
@@ -75,6 +77,42 @@ def __pydantic_post_init__(self_: Any, **kwargs: Any) -> None:
             validator_functions = _validation_functions.ValidationFunctions(bases)
             namespace['__pydantic_validator_functions__'] = validator_functions
 
+           # Errors for misuse of discriminated unions (check test_discriminated_unions)
+            for name, value in namespace['__annotations__'].items():
```
samuelcolvin, line 81: this logic should be in 

https://github.com/pydantic/pydantic/blob/55fd51fbdfea44cc16b41bdda8e102c85aefcbfc/pydantic/_internal/_model_construction.py#L167

Also we shouldn't be looking at `namespace['__annotations__']` directly anymore, but instead using `get_type_hints` as done in the above function.

pydantic/main.py
```diff
@@ -75,6 +77,42 @@ def __pydantic_post_init__(self_: Any, **kwargs: Any) -> None:
...
+                        if get_origin(arg) is Annotated:
+                            arg = get_args(arg)[0]
+                            # arg = get_args(arg)
+                            for nested_type in get_args(arg):
+                                if not issubclass(nested_type, BaseModel) and not is_dataclass(nested_type):
+                                    raise TypeError(f"Type '{nested_type.__name__}' is not a valid `BaseModel` or `dataclass`")
+                        elif not issubclass(arg, BaseModel) and not is_dataclass(arg):
+                            raise TypeError(f"Type '{arg.__name__}' is not a valid `BaseModel` or `dataclass`")
+                    
+                    # Ensure discriminated value is present in all union options, if not raise error
```
samuelcolvin, line 99: this is not long required, the key doesn't need to exist in all versions of the union.

samuelcolvin (changes requested):
Thanks for the contribution, overall I think there's a lot more work to do here.

Let me know if you have any more questions, otherwise I'll try to take over this work soon.

samuelcolvin:
thanks for contributing, no response so closing this.

Issue #4840, opened 2022-12-12
Exporting models (recursively) without extra fields


mblo:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this feature is missing

### Description

Previous discussion https://github.com/pydantic/pydantic/discussions/3508

The `extra` model config offers great flexibility with options `ignore`, `allow`, or `forbid` that are used during model initialization. Also, there is the export option to specify fields to exclude from the returned dictionary using `exclude` parameter of `dict()` or `json()`.

**However, there is no option to recursively exclude extra fields while exporting a model.**

Example:

```py
class NestedModel(BaseModel):
    a: int = ...
    class Config:
        extra = Extra.allow

class Model(BaseModel):
    c: NestedModel = ...

model: Model = Model(c=NestedModel(a=20, b=22))

print(model.dict()) ## {'c': {'a': 20, 'b': 22}}
# how to exclude 'b' from the export without knowing the name of b upfront?
print()
```

### Suggested change: 
Add optional parameter `exclude_extra: bool = False` to `dict` and `json` which removes recursively all extra fields from the export when set to `True`.

Example:

```py
# Model from the previous snippet

print(model.dict(exclude_extra=True)) ## {'c': {'a': 20}}
```

Required code changes are [fairly simple](https://github.com/pydantic/pydantic/compare/main...mblo:pydantic:export-model-without-extra-fields). I'm aware of V2 plan, but wonder if this feature could be added.

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
- [ ] [Data validation/parsing](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage)
- [X] [Data serialization](https://pydantic-docs.helpmanual.io/usage/exporting_models/) - `.dict()` and `.json()`
- [ ] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)
- [ ] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)
- [ ] [Model Config](https://pydantic-docs.helpmanual.io/usage/model_config/)
- [ ] [Field Types](https://pydantic-docs.helpmanual.io/usage/types/) - adding or changing a particular data type
- [ ] [Function validation decorator](https://pydantic-docs.helpmanual.io/usage/validation_decorator/)
[...]

PR #4841, opened 2022-12-12
feat: add exclude_extra parameter for exporting models recursively without extra fields

mblo:
Related issue #4840


samuelcolvin:
Serialization is being completely re-written in rust, see https://github.com/pydantic/pydantic-core/pull/327, so there's not really much point in finishing this PR and merging it as it'll be deleted soon.

I think this is an interesting idea, in V2 I'm currently (in the PR linked above) ignoring extra unless `config.extra = 'allow'`, but I can see the use case for this too.

I'll comment on your issue.

PR #4835, opened 2022-12-11
added few missing type annotations to functions

jimkring:
# NOTICE

We're currently in the process of rewriting pydantic in preparation for V2, see
https://pydantic-docs.helpmanual.io/blog/pydantic-v2/.

Please see https://github.com/orgs/pydantic/projects/1 for the major tasks required for this migration.

To avoid wasting your time, it would be best if you only created significant PRs related to that project.

Thank you for your interest in pydantic, and your patience. :pray:

**Note:** if you're making a pull request to fix pydantic v1.10, please make it against the `1.10.X-fixes` branch.

Issue #4683, opened 2022-10-28
V2: config parameter to warning about missing annotations


samuelcolvin:
We have [logic](https://github.com/pydantic/pydantic/blob/eaa2ad14c71138600de301d5f0ebf8fda6775788/pydantic/_internal/_model_construction.py#L203) to deal with the case where an annotation is not defined when a model is constructed, at the moment this silently stops the schema/validator build and `__pydantic_model_complete__` is left `False` in the hope that the missing annotation(s) can be defined and the model completed via `model_rebuild`.

I think we should have a config flag, default True which raising a warning when this happens - it should be a `UserWarning` so it gets printed by default.

Needs #4516 to be merged before this can be worked on, but would be amazing if someone could take this on.

ybressler:
I'd be happy to contribute to this. Would be my first contribution to pydantic – do you think it's suitable?

samuelcolvin:
Yes absolutely, please submit a PR.

ybressler:
Sweet! I'll get to working on this over the weekend / next week. (Meanwhile, is it possible for you to `assign` to me? Easier for me to be organized. If this is against your management preference, that's ok too.

jimkring:
@samuelcolvin I was looking at #4683 and I could not create a unit test that demonstrates the problem you're describing.

What do you mean exactly by "the case where an annotation is not defined when a model is constructed"?

I tried a variety of different things and I could not make a test where `assert f.__pydantic_model_complete__` would fail.

for example:

```python
def test_warn_on_model_missing_annotation():
    
    Foo = ForwardRef('Foo')

    class Foo(BaseModel):
        a: int = 123
        b: Foo = None
    
    f = Foo().construct()
    
    assert f.__pydantic_model_complete__, 'Model should be complete'
```

Would you be willing to share a simple (xfail) test that demonstrates the desired pass conditions?

Thanks.

samuelcolvin:
I I think of you define `Foo` after `Bar` you should get the error.

Or definitely if `Foo` is another pydantic model with references `Bar`.

jimkring:
@samuelcolvin Thanks for that info. I spent a while looking through the code.  I just created a PR with an attempt to implement this feature.  I have some questions that I'll ask in the PR.

Issue #4672, opened 2022-10-27
V2: rename model methods


samuelcolvin:
See https://pydantic-docs.helpmanual.io/blog/pydantic-v2/#model-namespace-cleanup

This is mostly mechanical and fixing lots of tests, should definitely wait for #4516, but even after that, the biggest problem might be conflicts with other open PRs.

nefas777:
Hi, I would like to help the project and this issue looks like good to start with. Can I take it? Do I need to change any docs (blog, ...)? Are there any other blockers except https://github.com/pydantic/pydantic/pull/4516?

samuelcolvin:
Hi @nefas777, I'd love you to take this on. Thanks so much.

The docs should be updated, but don't worry too much since most of the docs will be rewritten, blog doesn't need to be updated.

I don't think there are any blockers on this, the only thing to clarify is names if there are any method names you're not sure about - probably best to start, create a draft PR and comment there if you aren't clear or have any questions.

Thanks so much for the offer. :raised_hands:

PR #4843, opened 2022-12-14
move xfail from module to individual failing tests

jimkring:
As requested by @samuelcolvin here: https://github.com/pydantic/pydantic/pull/4835#issuecomment-1346606879

> If you're interested in testing, one useful task would be to go through some of the test files with many tests passing but the entire module marked as xfail, remove the global xfail and mark just the tests which are actually failing as xfail. Let me know if you need more information.

### DIFF
--- a/tests/test_create_model.py
+++ b/tests/test_create_model.py
@@ -6,9 +6,8 @@ from pydantic import BaseModel, Extra, Field, ValidationError, create_model, err
 from pydantic.fields import ModelPrivateAttr
 from pydantic.generics import GenericModel
 
-pytestmark = pytest.mark.xfail(reason='working on V2', strict=False)
 
-
+@pytest.mark.xfail(reason='working on V2')
 def test_create_model():
     model = create_model('FooModel', foo=(str, ...), bar=123)
     assert issubclass(model, BaseModel)
@@ -20,6 +19,7 @@ def test_create_model():
     assert model.__module__ == 'pydantic.main'
 
 
+@pytest.mark.xfail(reason='working on V2')
 def test_create_model_usage():
     model = create_model('FooModel', foo=(str, ...), bar=123)
     m = model(foo='hello')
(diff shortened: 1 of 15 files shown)


jimkring:
@samuelcolvin @hramezani I believe this PR is ready for review. All checks are passing, after I fixed a formatting/linting issue.

hramezani:
Thanks @jimkring Thanks for your effort :+1: 

I can see there are some files that you marked all the tests as [tests/test_validators_dataclass.py](https://github.com/pydantic/pydantic/pull/4843/files#diff-26712e49764150ea2e2d2a66cf513695e55ae4691f5d65be6a1d3ba1e7c5f6a8), and removed the global `xfail`. @samuelcolvin do you want to have these files (with all failing tests) like this?

Other than this, LGTM

tests/test_create_model.py
```diff
@@ -260,6 +271,7 @@ def _some_func(self):
         assert a._some_func == 2
 
 
+@pytest.mark.xfail(reason='working on V2', strict=False)
```
samuelcolvin, line 274: Lines like this should be removed - e.g. if the test is now passing, we should have no `xfail` decorator.

samuelcolvin (commented):
Otherwise, this looks great to me.

Thank you so much for the contribution.

tests/test_create_model.py
jimkring: @samuelcolvin thanks for catching this. I've removed the xfail marking from test_create_model_with_slots and pushed to the PR.  Hopefully that's the last issue and this is now ready for merge.

jimkring:
@samuelcolvin @hramezani is this PR ready for merge? I'm excited to brag to my friends and colleagues that I'm a contributor to Pydantic V2 :)

hramezani (approved):
Thanks!

samuelcolvin:
Thanks so much for this @jimkring 🎉 

I'd love you to contribute more to V2, a few suggestions:
* #4683 - this should be fairly small
* #4675 - there's quite a lot of details on the issue, there's also a pending PR for this which I've commented on (#4827), but really you would probably be better to start from scratch
* #4672 - big code change, but shouldn't be too complicated

Issue #4845, opened 2022-12-14
Better jsonschema output compatibility


cmcconomyfwig:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this feature is missing

### Description

I would like to submit a PR to enhance the abilities for jsonschema generation; when generating a json, the following fields are currently missing:
* `$id` - typically a URL pointing to a particular repository where one may find this jsonschema instance hosted
* `$schema` - indicates the compatibility level used for jsonschema generation
* `description` - objects may contain a description field, which currently can only be provided by the docstring. If the docstring is to be used for other purposes, then you end up stuck

The Description approach contrasts to the Title field, which defaults to the classname but can be overridden using `Config.title`

To achieve the above features, I propose the following changes:
1. Calling `.schema_json()` now offers an optional kwarg, "flavor"; I only intend to support "flavor='jsonschema'". The default is `None` and doesn't prepend the optional `$` fields.
2. two new fields are added to Config: `uri`, an Optional[str] field which represents the URI to point to with `$schema`; and `description`, which is an Optional[str] which may be used to override the docstring for description.

PR to follow, suggestions welcome.

(PS. I tried to compile this with python 3.8, and `make docs` failed because of a structural matching example - I updated to 3.10 and of course it worked fine.)

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
- [ ] [Data validation/parsing](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://pydantic-docs.helpmanual.io/usage/exporting_models/) - `.dict()` and `.json()`
- [X] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)
- [ ] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)
- [ ] [Model Config](https://pydantic-docs.helpmanual.io/usage/model_config/)
- [ ] [Field Types](https://pydantic-docs.helpmanual.io/usage/types/) - adding or changing a particular data type
- [ ] [Function validation decorator](https://pydantic-docs.helpmanual.io/usage/validation_decorator/)
- [ ] [Generic Models](https://pydantic-docs.helpmanual.io/usage/models/#generic-models)
[...]

PR #4847, opened 2022-12-14
JsonSchema Enhancement

cmcconomyfwig:
Hi team,

I know that pydantic 2 is around the corner, and that you're requesting primarily bugfixes; I've made a fairly lightweight and I think pydantical update to support a more complete jsonschema output experience.

This PR is linked to the following issue:
https://github.com/pydantic/pydantic/issues/4845


hramezani:
@cmcconomyfwig Thanks for your patch :+1: 

I think you made the PR based on the wrong base.

Also, We are working on `V2` and I think we are not going to accept any big change on `1.10.x`.

cmcconomyfwig:
> @cmcconomyfwig Thanks for your patch 👍
> 
> I think you made the PR based on the wrong base.
> 
> Also, We are working on `V2` and I think we are not going to accept any big change on `1.10.x`.

I'm sorry, I don't understand what I should do next, could you please let me know?

Issue #2816, opened 2021-05-19
`BaseModel` runtime-warns about private class vars from an arbitrary type annotation


goodboy:
I tried to find something similar to this in the issue set but figured it was worth a new one:

This code triggers an odd `RuntimeWarning` at import time:

```python
from dataclasses import dataclass

from pydantic import BaseModel


@dataclass
class C:
    _private: str = 'yo'


class D(BaseModel):
    c: C = None
    
    class Config:
        arbitrary_types_allowed = True
        validate_all = False
        keep_untouched = (C,)
        underscore_attrs_are_private = False

```
running this code gives:
```
[nav] In [9]: run snippets/pydantic_warning.py
/home/goodboy/repos/piker/snippets/pydantic_warning.py:11: RuntimeWarning: fields may not start with an underscore, ignoring "_private"
  class D(BaseModel):
```
It doesn't seem to matter what `Config` settings I use; the warning stays.

For me this was quite odd since in my real use case, my annotation type was from another lib using a plain stdlib `@dataclass` and `pydantic` was reporting it as though those attrs were local to the class `D` here; so I was pretty confused tracking down the source. If I convert `c` -> `_c` the warning doesn't show up.

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
(3.9)  >>> python -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.8.2
            pydantic compiled: True
                 install path: /home/goodboy/repos/piker/3.9/lib/python3.9/site-packages/pydantic
               python version: 3.9.5 (default, May 12 2021, 17:14:51)  [GCC 10.2.0]
                     platform: Linux-5.12.4-arch1-2-x86_64-with-glibc2.33
     optional deps. installed: ['typing-extensions']
```

Hopefully this isn't just noise 😂

PrettyWood:
Hi @goodboy and thanks for reporting.
Yes it's actually "normal" because when you use a stdlib `dataclass` inside a `BaseModel`, _pydantic_ will convert you stdlib `dataclass` into a _pydantic_ `dataclass` to validate (and coerce if needed) input data.
Behind the scenes it creates a `BaseModel` based on your stdlib `dataclass` and will raise a `RunWarning` at this moment.

goodboy:
@PrettyWood cool.

So the right way to mask this warning is to make the attr private then or does this mean you have to leak `pydantic` config stuff to stdlib dataclasses as a user?

[...]

PR #4853, opened 2022-12-17
Remove annoying warning when using fields starting with an underscore in a dataclass (#2816)

misuzu:
Closes https://github.com/pydantic/pydantic/issues/2816
cc @PrettyWood

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -659,8 +659,6 @@ def create_model(
     annotations = {}
 
     for f_name, f_def in field_definitions.items():
-        if f_name.startswith('_'):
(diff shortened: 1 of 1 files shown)


samuelcolvin:
humm, not sure about this.

When dataclasses are supported in V2, they won't use `create_model` - create model will remain only as a public API, for that use-case I think the warning is probably better kept.

samuelcolvin:
This is not a bug fix, so don't be accepted in the v1.10.x patch I'm
afraid.

samuelcolvin:
deferring this until we've done dataclasses for V2 and know if we'll want it.

samuelcolvin:
Thanks for contributing this, as I said above, I think #2816 should be fixed by #5009.

PR #4881, opened 2022-12-27
Delete setup.py

Kludex:
I actually didn't research about it, but I think GitHub fixed the "Used by", since I can see it on FastAPI: https://github.com/tiangolo/fastapi.


samuelcolvin:
👍 Let's try it.

PR #4882, opened 2022-12-27
config variable to raise UserWarning if model not fully defined

### LINKED ISSUE #4683: V2: config parameter to warning about missing annotations

jimkring:
Raises a UserWarning if model not fully defined and will need rebuild.

fix #4683 -> **V2: config parameter to warning about missing annotations**

New config item added to BaseConfig:
`warn_on_undefined_types: bool = True`

Also added small block of code to `build_inner_schema()` that checks if any of the annotations are of type`ForwardRef` and raises a PydanticUndefinedAnnotation error, in the same way as would happen in the missing annotation were caught as an AttributeError when using `from __future__ import annotations` instead of a `ForwardRef`


jimkring:
@samuelcolvin I can't seem to get CI to pass since pre-commit wants to Pyupgrade `foo: Optional[Foo] = None` to `foo: Foo | None = None`, but that causes CI tests on older python versions to fail. Not sure best path forward.

pydantic/config.py
```diff
@@ -83,6 +83,7 @@ class BaseConfig:
     validate_assignment: bool = False
     error_msg_templates: Dict[str, str] = {}  # TODO remove
     arbitrary_types_allowed: bool = False  # TODO default True, or remove
+    warn_on_undefined_types: bool = True
```
ybressler, line 86: It's technically not an undefined type, rather, a non-annotated type. I guess this is pedantics (which is the whole point 😉 ), but we should choose a suitable name for this config arg.

tests/test_forward_ref.py
```diff
@@ -44,6 +44,9 @@ def module():
         class Foo(BaseModel):
             a: 'Bar' = None
 
+            class Config:
+                warn_on_undefined_types = False
+
```
ybressler, line 49: This is reused many times in the test. Perhaps create a new BaseClass with the proper config and inherit? (Could make tests dry-er)

pydantic/_internal/_model_construction.py
```diff
@@ -139,9 +139,11 @@ def complete_model_class(
     except PydanticUndefinedAnnotation as e:
         if raise_errors:
             raise
-        cls.__pydantic_validator__ = MockValidator(  # type: ignore[assignment]
-            f'`{name}` is not fully defined, you should define `{e}`, then call `{name}.model_rebuild()`'
-        )
+        warning_string = f'`{name}` is not fully defined, you should define `{e}`, then call `{name}.model_rebuild()`'
+        # print(cls.__config__)
```
ybressler, line 143: Vestige of development, remove:
```suggested
        warning_string = f'`{name}` is not fully defined, you should define `{e}`, then call `{name}.model_rebuild()`'
```

pydantic/_internal/_model_construction.py
```diff
@@ -202,6 +204,11 @@ def build_inner_schema(  # noqa: C901
                 raise
         raise PydanticUndefinedAnnotation(name) from e
 
+    # check for forward references and raise PydanticUndefinedAnnotation
+    for t in type_hints.values():
+        if isinstance(t, typing.ForwardRef):
+            raise PydanticUndefinedAnnotation(t.__forward_arg__)
```
ybressler, line 210: Is this necessary? Looks erroneous?

pydantic/config.py
ybressler, line 86: Also, is there documentation which needs updating? (Readme's etc.?)

tests/test_construction_missing_annotations.py
```diff
@@ -0,0 +1,56 @@
...
+
+    with pytest.raises(UserWarning, match='is not fully defined'):
+
+        Bar = ForwardRef('Bar')
+
+        class Foo(BaseModel):
+            a: Bar
+
+        class Bar(BaseModel):
+            foo: Optional[Foo] = None
```
ybressler, line 20: Which line exactly is expected to raise the error? Is it line 17? Or line 20? (Perhaps indicate so in the docstring of the test.)

tests/test_construction_missing_annotations.py
```diff
@@ -0,0 +1,56 @@
...
+    assert Foo.__pydantic_model_complete__ is False
+
+    class Bar(BaseModel):
+        foo: Optional[Foo] = None
+
+    assert Bar.__pydantic_model_complete__ is True
+
+    Foo.model_rebuild()
+    assert Foo.__pydantic_model_complete__ is True
+    assert Bar.__pydantic_model_complete__ is True
```
ybressler, line 42: This test should be revised to be a bit simpler. I gather that you are providing a reference in class `Foo` to `Bar`, the latter of which doesn't have an annotation. Thereafter, you provide `Bar` an annotation and expect to rebuild `Foo` without error. If yes, then building `Bar` to include `Foo` is unnecessarily complex. Why not provide `Bar` with simpler attributes, since this test just needs to validate that `Foo.__pydantic_model_complete__ is True` (Bar can be ignored...)

tests/test_construction_missing_annotations.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_user_warning():
```
ybressler, line 45: Function name should be a bit more specific IMO.

tests/test_construction_missing_annotations.py
```diff
@@ -0,0 +1,56 @@
...
+    Bar = ForwardRef('Bar')
+
+    class Foo(BaseModel):
+        b: Bar
+
+        class Config:
+            warn_on_undefined_types = False
+
+    class Bar(BaseModel):
+        foo: Optional[Foo] = None
```
ybressler, line 56: This test looks incomplete.

ybressler (changes requested):
Great work thus far. Left some feedback + some Q's for you.

ybressler:
> @samuelcolvin I can't seem to get CI to pass since pre-commit wants to Pyupgrade `foo: Optional[Foo] = None` to `foo: Foo | None = None`, but that causes CI tests on older python versions to fail. Not sure best path forward.

A simpler (more focused) test case would solve this issue for you IMO.

pydantic/_internal/_model_construction.py
samuelcolvin: why?

pydantic/config.py
samuelcolvin: I think "undefined" is the best term. This warning related to annotations which are not yet defined when creating the model.

This is not about non-annotated types.

samuelcolvin: yes, would be useful to add documentation for this now. Although the docs needs a lot of work for V2. 

@jimkring Please could you and something to the docs, and I'll take care of migrating it to the new documentation structure when the time comes.

tests/test_construction_missing_annotations.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import annotations
```
samuelcolvin, line 1: I think these tests can go in `test_forward_ref.py`.

tests/test_construction_missing_annotations.py
```diff
@@ -0,0 +1,56 @@
...
+import pytest
+
+from pydantic.main import BaseModel
+
+
+def test_missing_annotation_warning_defaults_to_true():
+
+    with pytest.raises(UserWarning, match='is not fully defined'):
+
+        Bar = ForwardRef('Bar')
```
samuelcolvin, line 14: Better to use `create_module` and `from __future__ import annotations`, then just have

```py

class Foobar(BaseModel):
    a: MissingType
```

samuelcolvin: you can add this self-referencing case as a separate test. Best to match against more of the error message to confirm where the error is happening.

pydantic/_internal/_model_construction.py
```diff
@@ -202,6 +204,11 @@ def build_inner_schema(  # noqa: C901
                 raise
         raise PydanticUndefinedAnnotation(name) from e
 
+    # check for forward references and raise PydanticUndefinedAnnotation
+    for t in type_hints.values():
```
samuelcolvin, line 208: This makes sense, but I don't think it needs to be a separate loop.

pydantic/_internal/_model_construction.py
samuelcolvin: See the PR description.

I think this case is tested, @jimkring please confirm.

pydantic/config.py
samuelcolvin: `undefined_types_warning` would be most concise.

samuelcolvin (commented):
This is looking great, a few more things to fix.

Please update.

pydantic/config.py
ybressler: Thanks for the clarification.

pydantic/_internal/_model_construction.py
ybressler: ^ I attempted to leave a code suggestion, but it didn't go. What my comment **meant** to say is >> "remove the print statement"

tests/test_construction_missing_annotations.py
ybressler: Curious if you feel strongly 1 way or the other @samuelcolvin

pydantic/_internal/_model_construction.py
jimkring: Oops, I deleted my old reply/comment somehow. The point is that I think it does need to be a separate loop, since there is really no "other loop" to put this in (that I can find).

jimkring (commented):
All CI tests are passing and I think I'm "done" with this round of edits. Look forward to your review.

tests/test_forward_ref.py
```diff
@@ -840,3 +859,95 @@ class Foo(BaseModel):
...
+    """
+
+    return pytest.raises(
+        UserWarning,
+        match=re.escape(
+            # split this f-string into two lines to pass linting (line too long)
+            f'`{defining_class_name}` is not fully defined, you should define `{missing_type_name}`, '
+            f'then call `{defining_class_name}.model_rebuild()`',
+        ),
+    )
```
jimkring, line 876: @samuelcolvin this is a reusable assertion that requires a block of code to raise an "undefined types warning". It requires matching the exact warning string, including the specific class names involved.

tests/test_forward_ref.py
```diff
@@ -840,3 +859,95 @@ class Foo(BaseModel):
...
+
+        class Foobar(BaseModel):
+            a: UndefinedType
+
+            class Config:
+                undefined_types_warning = False
+
+    # Since test is about suppressing the undefined_types_warning,
+    # we should also verify that our model was at least partially built
+    assert module.Foobar.__pydantic_model_complete__ is False
```
jimkring, line 953: @samuelcolvin @ybressler there are now 4 new unit tests which are basically a matrix of:

1) how to handle UndefinedType (import annotations vs import ForwardRef)
    a) use `from __future__ import annotations`
    b) use `UndefinedType = ForwardRef('UndefinedType')`
2) whether to suppress the undefined_types_warning (suppress warning or not)
    a) don't suppress it and then verify the expected warning and its detailed message
    b) do suppress it and then verify the model is partially built (without the warning being raised)

The test names reflect these 4 different possibilities of the two different parameters

tests/test_forward_ref.py
```diff
@@ -44,6 +45,9 @@ def module():
         class Foo(BaseModel):
             a: 'Bar' = None
 
+            class Config:
+                undefined_types_warning = False
+
```
jimkring, line 50: @samuelcolvin @ybressler there are several pre-existing tests that needed `undefined_types_warning = False` added to their config, in order for them to work as they did previously (since the previous behavior when using a `ForwardRef` was a silent fail instead of an undefined_types_warning).  We should make sure this makes sense for all the affected tests. You both (and/or the previous test authors) are more qualified (than I am) to make that determination.

samuelcolvin: Makes sense, I'll look.

tests/test_forward_ref.py
samuelcolvin: Great.

tests/test_forward_ref.py
```diff
@@ -840,3 +859,95 @@ class Foo(BaseModel):
...
+        create_module(
+            # language=Python
+            """
+from __future__ import annotations
+from pydantic import BaseModel
+
+class Foobar(BaseModel):
+    a: UndefinedType
+"""
+        )
```
ybressler, line 892: Fancy stuff. I like it!

tests/test_forward_ref.py
```diff
@@ -840,3 +859,95 @@ class Foo(BaseModel):
...
+
+class Foobar(BaseModel):
+    a: UndefinedType
+    class Config:
+        undefined_types_warning = False
+"""
+    )
+    # Since test is about suppressing the undefined_types_warning,
+    # we should also verify that our model was at least partially built
+    assert module.Foobar.__pydantic_model_complete__ is False
```
ybressler, line 930: Nice test. Looks very similar to the test `test_undefined_type_raises_warning()`. There are 2 general ways to make life simpler here:
1. Describe the relationship of these 2 tests in their docstrings. (Human talking to a human solution)
2. Parameterize these into a single test. In this case, you get tests which is slightly DRY-er, but slightly more more complex.

I personally am in favor of the 2nd option. Btw, feel free to delegate to a follow-up PR which I would be happy to contribute to – OR, I can contribute to this PR if you want support with this. (OR, choose option 1.) _OR..._ something else?

tests/test_forward_ref.py
```diff
@@ -840,3 +859,95 @@ class Foo(BaseModel):
...
+
+
+def test_disable_undefined_types_warning_if_missing_class_is_a_forwardref(create_module):
```
ybressler, line 933: Same feedback as above re: parameterizing tests (related to `test_undefined_type_raises_warning_if_forward_ref()`

tests/test_forward_ref.py
ybressler: Fantastic work!

pydantic/_internal/_model_construction.py
```diff
@@ -202,6 +205,11 @@ def build_inner_schema(  # noqa: C901
                 raise
         raise PydanticUndefinedAnnotation(name) from e
 
+    # Also raise PydanticUndefinedAnnotation if a type_hint is a ForwardRef type (i.e. not yet defined)
+    for t in type_hints.values():
+        if isinstance(t, typing.ForwardRef):
+            raise PydanticUndefinedAnnotation(t.__forward_arg__)
```
ybressler, line 211: I don't see a test which correctly raises this error.

In other words, if I were to check out your branch and remove this code (or change it to raise no error), I would want a test to fail.

docs/examples/types_undefined_warning.py
```diff
@@ -0,0 +1,50 @@
...
+    {
+      "title": "Python Crash Course",
+      "author": {
+        "name": "Eric Matthes"
+      }
+    }
+  ]
+}
+""".strip()
+)
```
ybressler, line 50: Instead of `jane_doe.json()` I'd rather `jane_doe.dict()`, allows neater evaluation of python objects.

ybressler (changes requested):
Thanks for your quick turnarounds! Some additional comments / feedback for you. I think we're almost done.

pydantic/_internal/_model_construction.py
jimkring: @ybressler when I comment out this block on my dev system (Python 3.10), the two tests fail:

- test_undefined_type_raises_warning_if_forward_ref
- test_disable_undefined_types_warning_if_missing_class_is_a_forwardref

This was the reason I added that block of code.  This makes me wonder if there are different behaviors on different Python versions or systems, since the importance of this block of code was initially questioned by you and @samuelcolvin.

Ah, yes... I think I found the source of the problem/difference:

Inside `_typing_extra.get_type_hints()` it checks if Python >= 3.10 and then uses:

Python >= 3.10 -> the built-in `typing.get_type_hints()` for Python >= 3.10
Python < 3.10 -> a special implementation with partially copied code from 3.10

_typing_extra.py lines 216-234
```python
if sys.version_info >= (3, 10):  # noqa C901
    get_type_hints = typing.get_type_hints

else:
    """
    For older versions of python, we have a custom implementation of `get_type_hints` which is a close as possible to
    the implementation in CPython 3.10.8.
    """
    fr_has_is_class = True

    def ForwardRefWrapper(arg: Any, is_argument: bool = True, *, is_class: bool = False) -> typing.ForwardRef:
        """
        Wrapper for ForwardRef that accounts for the `is_class` argument missing in older versions.
        The `module` argument is omitted as it breaks <3.9 and isn't used in the calls below.

        See https://github.com/python/cpython/pull/28560 for some background

        Implemented as EAFP with memory.
        """
```

Untangling this is probably important and is something I will defer to others :)

It's great that you noticed this @ybressler. My guess is that this will make sense to @samuelcolvin since the custom implementation of `get_type_hints` appears to be his work.

tests/test_forward_ref.py
jimkring: @ybressler I see what you mean, yet I'm not quite sure how to go about parameterizing a test that use `create_module`. Any code defined outside the `def module()` function is easily accessible inside the `module` that is the code under test.  Also, for tests where the `module` is calling `from __future__ import annotations` we need to `create_module()` from a string `# language=Python` otherwise it seems the tests don't run correctly (I'm not entirely sure why, but generally `from __future__ import annotations` has the be the first line of code in a file).  So, for the sake of symmetry and readability, I'm inclined the just sprinkle in some additional comments (1st option you recommended).  All that said, I'm very curious what changes you'd propose (since I'd probably learn a lot and you're probably able to see further around the obstacles than I can) -- I would probably defer that to a follow-up PR, once the smoke settles around the questions of ForwardRef handling (in Python < 3.10 vs. >=3.10).

pydantic/_internal/_model_construction.py
samuelcolvin: line 214 we iterate over `type_hints`

```py
for ann_name, ann_type in type_hints.items():
```

jimkring: It was hiding in plain sight :)
Thanks!

jimkring: fixed and pushed

docs/examples/types_undefined_warning.py
jimkring: Nice suggestion! I made the change (to what's shown below) and pushed:

```python
assert jane_doe.dict(exclude_unset=True) == {
    'name': 'Jane Doe',
    'books_read': [
        {
            'title': 'Python Crash Course',
            'author': {'name': 'Eric Matthes'},
        },
    ],
}
```

tests/test_forward_ref.py
jimkring: Thanks!

tests/test_forward_ref.py
ybressler: Good approach. Once comments are added, let's close this thread.

pydantic/_internal/_model_construction.py
ybressler: Agreed. Let's wait till Samuel reviews. I imagine a suitable approach is to create a follow up issue (perhaps backlogged) to untangle this functionality. Meanwhile, code is good as is.

docs/examples/types_undefined_warning.py
```diff
@@ -32,19 +32,12 @@ class Person(BaseModel):
...
-      }
-    }
-  ]
+assert jane_doe.dict(exclude_unset=True) == {
+    'name': 'Jane Doe',
+    'books_read': [
+        {
+            'title': 'Python Crash Course',
+            'author': {'name': 'Eric Matthes'},
+        },
```
ybressler, line 41: As long as there's one book, you won't have an issue. But if there are more books, you may get into an issue with ordering. That's not for now though. I think we're good to solve that problem later, should it present itself.

ybressler (approved):
Looks good to me! Let's resolve that one last thread about `python3.10` with `type_hints` before calling this done.

Great work, and very quick too!

tests/test_forward_ref.py
```diff
@@ -876,9 +876,19 @@ def pytest_raises_undefined_types_warning(defining_class_name, missing_type_name
...
+#   NOTE: the `undefined_types_warning` tests below are "statically parameterized" (i.e. have Duplicate Code).
+#   The initial attempt to refactor them into a single parameterized test was not strateforward, due to the use of the
+#   `create_module` fixture and the need for `from __future__ import annotations` be the first line in a module.
+#
+#   Test Parameters:
+#     1. config setting: (1a) default behavior vs (1b) overriding with Config setting:
+#     2. type checking approach: (2a) `from __future__ import annotations` vs (2b) `ForwardRef`
+#
+#   The parameter tags "1a", "1b", "2a", and "2b" are used in the test names below, to indicate which combination
+#   of conditions the test is covering.
```
ybressler, line 888: Probably the best inline documentation I've seen in a very long time. Well done.

ybressler (approved):

samuelcolvin:
I'll try and review tomorrow.

tests/test_forward_ref.py
```diff
@@ -840,3 +859,99 @@ class Foo(BaseModel):
...
+        class Foobar(BaseModel):
+            a: UndefinedType
+
+            # Suppress the undefined_types_warning
+            class Config:
+                undefined_types_warning = False
+
+    # Since we're testing the absence of a warning, it's important to confirm pydantic was actually run.
+    # The presence of the `__pydantic_model_complete__` is a good indicator of this.
+    assert module.Foobar.__pydantic_model_complete__ is False
```
jimkring, line 957: @ybressler I've added a block of comments/docs at the top of these 4 tests to describe:

- why the tests have duplicated code (and some context to help someone untangle it the future)
- what the tests are doing (with respect to the parameters/conditions we're trying to cover).
- some simple tags for each parameter state that can be used in the test name (and provide a reference to the comment block in order to to avoid duplication of the notes/comments within the tests themselves).

tests/test_forward_ref.py
jimkring: comments added. I'll close this thread and we can review the changes in that other thread.

tests/test_forward_ref.py
jimkring: thanks! I'll mark this as resolved

tests/test_forward_ref.py
ybressler: Excellent!

jimkring:
Thanks @samuelcolvin. I'm happy to discuss if you have any questions.

samuelcolvin:
Sorry, been busy, I'll review soon.

samuelcolvin:
Sorry, been busy, I'll review soon.

docs/examples/types_undefined_warning.py
```diff
@@ -0,0 +1,43 @@
...
+        undefined_types_warning = False
+
+
+class Person(BaseModel):
+    name: str
+    books_read: list[Book] | None = None
+
+
+# Now, we can rebuild the `Book` model, since the `Person` model is now defined.
+Book.model_rebuild()
```
samuelcolvin, line 26:
```suggestion
# Now, we can rebuild the `Book` model, since the `Person` model is now defined.
# Note: there's no need to call `model_rebuild()` on `Person`, it's already complete.
Book.model_rebuild()
```

samuelcolvin (approved):
otherwise LGTM. I'll add that suggestion and merge.

samuelcolvin:
Thanks so much @jimkring, this is great.

docs/examples/types_undefined_warning.py
```diff
@@ -0,0 +1,44 @@
...
+        undefined_types_warning = False
+
+
+class Person(BaseModel):
+    name: str
+    books_read: list[Book] | None = None
+
+
+# Now, we can rebuild the `Book` model, since the `Person` model is now defined.
+# Note: there's no need to call `model_rebuild()` on `Person`, it's already complete.
```
samuelcolvin, line 26:
```suggestion
# Note: there's no need to call `model_rebuild()` on `Person`, 
# it's already complete.
```

jimkring:
Thanks @samuelcolvin.
@ybressler: I just fixed a minor linting error (EOL) to get CI into a passing state.  It should be ready to merge.

Thank you both for taking the time to do the code reviews and offer suggestions. I learned a lot in the process and it feels great to support this impactful project.

samuelcolvin:
Thanks so much, merged.

PR #4911, opened 2023-01-04
fix field inheritance

### LINKED ISSUE #4909: Intended behavior of Field in Parent-Child-Inheritance
### Discussed in https://github.com/pydantic/pydantic/discussions/4896

<div type='discussions-op-text'>

<sup>Originally posted by **florianfischer91** January  2, 2023</sup>
Given the following code
```python
class Parent(BaseModel):
    a: int = 3
    # a: Annotated[int, Field(3)]

class Child(Parent):
    b: str = Field(alias='bbb')
```
the resulting field infos of `Parent` and `Child` are
```bash
{'a': FieldInfo(annotation=int, required=False, default=3)} # Parent
{'a': FieldInfo(annotation=int, required=True), 'b': FieldInfo(annotation=str, required=True, alias='bbb', alias_priority=2)} # Child
```

You can see that for `Child` the inherited field `a` has no default value and is required. Using `Annotated` and `Field` (see comment) leads to the (in my opinion) correct field infos
```bash
{'a': FieldInfo(annotation=int, required=False, default=3)}
{'a': FieldInfo(annotation=int, required=False, default=3), 'b': FieldInfo(annotation=str, required=True, alias='bbb', alias_priority=2)}
```

This behavior applies to every case where a 'value' is assigned to a field. I know that V2 is still in development but i just stumpled over this and want to now whether this is intended and one should just use `Annotated` or still has to be implemented?
</div>

florianfischer91:
Fixes #4909

Not sure whether my changes fixes the problem completely or not but at least some more tests are passing now. Let me know what you think and if i should change something.


florianfischer91:
please review

pydantic/_internal/_model_construction.py
```diff
@@ -202,6 +202,8 @@ def build_inner_schema(  # noqa: C901
                 raise
         raise PydanticUndefinedAnnotation(name) from e
 
+    # https://docs.python.org/3/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older
+    annotations = cls.__dict__.get('__annotations__', {})
```
samuelcolvin, line 206:
```suggestion
    # https://docs.python.org/3/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older
    # annotations is only used for finding fields in parent classes
    annotations = cls.__dict__.get('__annotations__', {})
```

pydantic/_internal/_model_construction.py
```diff
@@ -217,7 +219,12 @@ def build_inner_schema(  # noqa: C901
         try:
             default = getattr(cls, ann_name)
         except AttributeError:
-            fields[ann_name] = FieldInfo.from_annotation(ann_type)
+            # if field has no default value and is not in __annotations__ this means that it is
```
samuelcolvin, line 222:
```suggestion
            # if field has no default value and is not in `type_hints` this means that it is
```

pydantic/_internal/_model_construction.py
```diff
@@ -217,7 +219,12 @@ def build_inner_schema(  # noqa: C901
...
+            # defined in a base class and we can take it from there
+            if ann_name not in annotations:
+                fields[ann_name] = cls.__fields__[ann_name]
+            else:
+                fields[ann_name] = FieldInfo.from_annotation(ann_type)
```
samuelcolvin, line 227:
```suggestion
            if ann_name in annotations:
                fields[ann_name] = FieldInfo.from_annotation(ann_type)
            else:
                fields[ann_name] = cls.__fields__[ann_name]
```
Generally better to use to test for the "yes" case, not the "not yes" case.

samuelcolvin (commented):
Just a few tiny things, otherwise looks great.

Thanks so much for the contribution.

samuelcolvin:
please update.

pydantic/_internal/_model_construction.py
florianfischer91: Not sure about this change, because `get_type_hints` returns the annotations of inherited members too. I could write `annotations` without dunders, but perhaps i also missunderstand something here...

samuelcolvin: okay, ignore.

florianfischer91:
please review

samuelcolvin:
awesome, thank so much.

PR #4912, opened 2023-01-05
Support kwargs model configuration in Mypy plugin

mwgamble:
Previously, the Mypy plugin only supported drawing settings from a Config class. Now, it fully supports drawing settings from class kwargs too.


mwgamble:
I have a desire to see this change backported to v1.10. I _think_ it's going to be relatively straightforward, just a cherry-pick.

Is there a formal process for backporting fixes, or should I just submit a follow-up PR once this PR is merged?

samuelcolvin (approved):
LGTM, although the change looks fairly extensive.

Maybe @dmontagu might take a quick look at this as he wrote the original plugin>

WRT backporting, as @hramezani says we're only fixing critical bugs in V1.X now, so no backport.

mwgamble:
Thank you for such a fast response and review! I greatly appreciate it :heart:

May I ask how critical a bugfix needs to be to be considered for backporting? Without the changes in this PR, the mypy plugin is essentially unusable if you're using the kwargs-based model config approach.

dmontagu:
@samuelcolvin This seems fine to me.

If the plugin continues to require any significant amount of maintenance I think it will eventually make sense to improve the way the mypy plugin error snapshot-style tests work so it's a bit easier to review diffs (and relate them to source code changes). But not today.

samuelcolvin:
> May I ask how critical a bugfix needs to be to be considered for backporting? Without the changes in this PR, the mypy plugin is essentially unusable if you're using the kwargs-based model config approach.

Realistically we're only fixing regressions in v1.10 and security security bugs at this stage.

> If the plugin continues to require any significant amount of maintenance I think it will eventually make sense to improve the way the mypy plugin error snapshot-style tests work so it's a bit easier to review diffs (and relate them to source code changes). But not today.

Agreed, I guess we could just test with type-ignore including the error code. Either that or build a pytest plugin for this, but would be quite a lot of work.

I'll review and merge this in the week.

samuelcolvin:
please review

(just so I remember to review this soon)

samuelcolvin:
Thanks so much for this.

PR #4921, opened 2023-01-07
Make base config a dict

### LINKED ISSUE #4673: V2: Make `Config` a dict not a class
I would like to suggestion (strongly, but I will take feedback) that we change config from a nested class on a model, to a dict.

Rationale for this:
* config is already a dict if you set it via kwargs to a class on definition
* Config inheritance is currently magic, complicated and error prone
* when methods are defined on the config class, what do `self` or `cls` "mean", it's not clear to me, or to others
* we already have the weird distinction between `Model.__config__` and `Model.Config`, would be good to remove this
* Most of all, as more and more validation goes on without a model, it's not clear how config should be defined if it's a class

With that, I would suggest the default way of defining config on a model, is via kwargs, e.g:

```py

def my_alias_generate(..) ...:

class MyModel(BaseModel, strict=True, alias_generator=my_alias_generate, ...):
    ...
```

But, to support bigger configs, we could also support

```py
from pydantic import BaseModel, ConfigTypeDict

def my_alias_generate(..) ...:

class MyModel(BaseModel):
    model_config = ConfigTypeDict(strict=True, alias_generator=my_alias_generate, ...)
    ...
```

Then with the `validate` function #4669, config becomes

```py
from pydantic import validate

@validate(strict=True, alias_generator=...)
def my_method(...):
   ...
```

questions?

florianfischer91:
fix #4673

There are still some open `TODO`s related to the ConfigDict. Perhaps you can give me a hint what you want to do with it @samuelcolvin 
And i haven't adjusted the docs because i think you have mentioned somewhere the docs for v2 will be rewritten anyway. But let me know if i should update it anyway.


samuelcolvin:
Thanks, I'll review soon. Good if you can get tests passing first.

samuelcolvin:
please review

(so I remember to review this soon)

@kennethreitz maybe you could have a look at this?

pydantic/config.py
```diff
@@ -36,183 +34,159 @@ class Extra(str, Enum):
...
-        """
-        pass
-
-
-def get_config(config: Union[ConfigDict, Type[object], None]) -> Type[BaseConfig]:
+    post_init_call: Literal['before_validation', 'after_validation']  # TODO remove
+
+    # TODO where to impelement these two classmethods?
+    # @classmethod
+    # def get_field_info(cls, name: str) -> Dict[str, Any]:
```
samuelcolvin, line 69: delete it, we're removing `Config.fields`

pydantic/config.py
```diff
@@ -36,183 +34,159 @@ class Extra(str, Enum):
...
+    #     """
+    #     Optional hook to check or modify fields during model creation.
+    #     """
+    #     pass
+
+
+config_keys = set(ConfigDict.__annotations__.keys())
+
+
+def _default_config() -> ConfigDict:
```
samuelcolvin, line 104: This is my main question about this, PR I would argue that we shouldn't build full config for all models/validation

Instead we should have a static `default_config` dict, any then maybe a utility method, like `def get_config_param(config, name)` which falls back to `default_config`.

WDYT?

pydantic/config.py
```diff
@@ -36,183 +34,159 @@ class Extra(str, Enum):
...
+        json_loads=json.loads,
+        json_dumps=json.dumps,
+        json_encoders={},
+        allow_inf_nan=True,
+        strict=False,
+        copy_on_model_validation='shallow',
+    )
+
+
+def get_config(config: Union[ConfigDict, Type[object], None] = None) -> ConfigDict:
```
samuelcolvin, line 131:
```suggestion
def get_config(config: ConfigDict | None = None) -> ConfigDict | None:
```
As per my above comment, we can keep `config_dict` partial.

pydantic/config.py
```diff
@@ -36,183 +34,159 @@ class Extra(str, Enum):
...
 
-    config_bases = []
+    config_default = dict(_default_config())
+    config_bases = {}
+    configs_ordered = [config_default]
+    # collect all config options from bases
+    # to avoid that options will be overriden by default-values we just take value != default-value
     for base in bases:
-        config = getattr(base, 'Config', None)
+        config = getattr(base, 'model_config', {})
```
samuelcolvin, line 164:
```suggestion
        config = getattr(base, 'model_config', None)
```

pydantic/config.py
samuelcolvin: We could even for simplicity have something like

```py
class ConfigWrapper:
    title: Optional[str]
    anystr_lower: bool
    anystr_strip_whitespace: bool
    ...

    __slots__ = ('_config',)
    def __init__(self, config: ConfigDict | None):
        self._config = config
    
    def __getattr__(self, name: str) -> Any:
        # something like...
        if self._config:
            v = self._config.get(name)
            if v is not None:
                return v
        return default_config[name]
        ...
```

I know it would involve some duplication, but it would make usage through much more elegant.

samuelcolvin (commented):
Overall, looks good, main change is moving to config being internally `ConfigDict | None`.

Unfortunately, you'll have lots of conflicts due to #4912

pydantic/config.py
florianfischer91: We then assign an instance of ConfigWrapper to `model_config` when creating the class? It feels always a bit weird to me changing the value of a variable the user has set and changing the type even more. And if  we do that `model_config` would still have the type `ConfigDict` during type checking, right?

Another question is, do you rather want to write `model_config['extra']` or `model_config.extra` for getting the value.

I was also thinking about something like a `ChainMap`, but that behaves like a `TypedDict` for type checking.

samuelcolvin: I guess you're right, just keep it as a dict/None and use a utility to get the value.

pydantic/config.py
florianfischer91: Dropping `Type[object]` does also mean removing the ability to pass a class object into dataclass decorator?

```python
@dataclass_transform(kw_only_default=True, field_specifiers=(Field, FieldInfo))
def dataclass(
    _cls: Optional[Type[_T]] = None,
    *,
    init: bool = True,
    repr: bool = True,
    eq: bool = True,
    order: bool = False,
    unsafe_hash: bool = False,
    frozen: bool = False,
    config: ConfigDict | None = None, # before config: Union[ConfigDict, Type[object], None] = None,
    validate_on_init: Optional[bool] = None,
    kw_only: bool = False,
) -> Union[Callable[[Type[_T]], 'DataclassClassOrWrapper'], 'DataclassClassOrWrapper']:
```

samuelcolvin: Good point, thinking about this a bit more, I think it would probably be good to keep support for class defined Config everywhere, with a depreciation warning.

pydantic/config.py
florianfischer91: Done

pydantic/config.py
florianfischer91: Done

pydantic/config.py
samuelcolvin: I'm afraid this will make the implementation much uglier and require lots of the removed code to be added back.

Annoying I know, but I think worth it.

florianfischer91: What do you exactly mean? Adding back support for Config-Class in BaseModel?

florianfischer91: Any news here?

samuelcolvin: Sorry for the delay, yes we'll have to add back Config Class support and raise a deprecation warning.

Sorry about this, but I think it'll help people a lot.

samuelcolvin:
please update.

florianfischer91:
please review

pydantic/config.py
```diff
@@ -34,185 +35,175 @@ class Extra(str, Enum):
...
+
+if TYPE_CHECKING:
+
+    class ConfigDict(_ConfigDict):
+        ...
+
+else:
+
+    class ConfigDict(dict):
+        def __missing__(self, key):
```
samuelcolvin, line 82: :+1: this is smart.

pydantic/config.py
```diff
@@ -34,185 +35,175 @@ class Extra(str, Enum):
...
 
 class BaseConfig:
     title: Optional[str] = None
-    anystr_lower: bool = False  # TODO rename to str_to_lower
-    anystr_upper: bool = False  # TODO rename to str_to_upper
-    anystr_strip_whitespace: bool = False  # TODO rename to str_strip_whitespace
-    min_anystr_length: int = 0  # TODO rename to str_min_length
-    max_anystr_length: Optional[int] = None  # TODO rename to str_max_length
-    validate_all: bool = False  # TODO remove
+    str_to_lower: bool = False
```
samuelcolvin, line 115: I wonder if you can remove all these attributes from `BaseClass`, perhaps add a `__getattr__` method to with a warning?

pydantic/config.py
```diff
@@ -34,185 +35,175 @@ class Extra(str, Enum):
...
     if config is None:
-        return BaseConfig
+        return ConfigDict()
 
-    else:
-        config_dict = (
-            config
-            if isinstance(config, dict)
-            else {k: getattr(config, k) for k in dir(config) if not k.startswith('__')}
+    if not isinstance(config, dict):
```
samuelcolvin, line 143: I prefer to avoid unnecessary negatives, and also consolidate the `isinstance(config, dict)` checks here.

So something like 

```py
if isinstance(config, dict):
    config_dict = config
else:
    warnings.warn(...)
    config_dict = {k: getattr(config, k) for k in dir(config) if not k.startswith('__')}
```

?

samuelcolvin (commented):
overall this is looking great, I need to check it out and play with it a bit more, but looks pretty close.

I'm sorry about all the conflicts, that's my fault for merging a newer PR. If you like, I'll try to fix these for you.

pydantic/config.py
florianfischer91: You mean all attributes or just the ones which were renamed, so that you get a warning if you use f.i. `anystr_to_lower` instead of `str_to_lower` ? If we do it for every attribute then the user would get duplicate warnings when using BaseClass instead of ConfigDict and when using an attribute of the class... And you lose intellisense.
But if you wish i can change it.

florianfischer91:
> overall this is looking great, I need to check it out and play with it a bit more, but looks pretty close.
> 
> I'm sorry about all the conflicts, that's my fault for merging a newer PR. If you like, I'll try to fix these for you.

I will give it a try tomorrow

pydantic/config.py
samuelcolvin: I was suggesting removing all, usage is not recommended after all. But totally up to you.

florianfischer91: I was not able to get this working, because if you use `BaseConfig.str_to_lower` `__getattr__`  won't be called and you would probably need a metaclass which implements this method and i don't know if it's worth to do that and what implication this would have...

samuelcolvin: Ye, I thought we might need a custom metaclass. I'll look next week.

@dmontagu WDYT? On this specifically and the whole "config is a dict" thing.

samuelcolvin:
My opinion is that we should add a metaclass and `__getattr__` to `BaseConfig`, then remove the default attributes.

Once that's one, I think we merge this.

pydantic/config.py
```diff
@@ -116,34 +116,17 @@ def __missing__(self, key):
...
-    keep_untouched: Tuple[type, ...] = ()
-    json_loads: Callable[[str], Any] = json.loads
-    json_dumps: Callable[..., str] = json.dumps
-    json_encoders: Dict[Union[Type[Any], str, ForwardRef], Callable[..., Any]] = {}
-    allow_inf_nan: bool = True
-    strict: bool = False
-    copy_on_model_validation: Literal['none', 'deep', 'shallow'] = 'shallow'
-    post_init_call: Literal['before_validation', 'after_validation'] = 'before_validation'
+class ConfigMetaclass(type):
+    def __getattr__(self, item: str) -> Any:
```
samuelcolvin, line 120: can we add a deprecation warning here, and add a test for it's usage.

pydantic/config.py
```diff
@@ -116,34 +116,17 @@ def __missing__(self, key):
...
+        return _default_config.get(item)
```
samuelcolvin, line 121: I guess we should raise `AttributeError` if item is not found?

samuelcolvin (commented):
otherwise LGTM.

tests/test_config.py
```diff
@@ -0,0 +1,330 @@
...
+                ...
+
+        class MyModel(MyBaseModel):
+            ...
+
+        expected = _default_config.copy()
+        expected['title'] = 'MyTitle'
+        expected['frozen'] = True
+        for k, v in expected.items():
+            assert MyModel.model_config[k] == v
```
dmontagu, line 80: @samuelcolvin prior to my changes, this was:
```python
        expected = _default_config.copy()
        expected['title'] = 'MyTitle'
        expected['frozen'] = True
        assert MyModel.model_config == expected  # <-- affected line
```
Is the `.model_config` attribute supposed to hold all fields from `_default_config`? Considering `ConfigDict` has `total=False` I'm assuming it shouldn't need to have all fields present, but I wanted to make sure I understood the intent correctly.

tests/test_config.py
```diff
@@ -0,0 +1,330 @@
...
+        return all(_equals(a_, b_) for a_, b_ in zip(a, b))
+    else:
+        raise TypeError(f'arguments must be both strings or both lists, not {type(a)}, {type(b)}')
+
+
+@pytest.mark.filterwarnings('ignore:.* is deprecated.*:DeprecationWarning')
+class TestsBaseConfig:
+    def test_base_config_equality_defaults_of_config_dict_class(self):
+        for key, value in _default_config.items():
+            assert getattr(BaseConfig, key) == value
```
dmontagu, line 53: @samuelcolvin prior to my changes, there was an additional line in this test:
```python
assert _default_config.keys() == {k for k in dir(BaseConfig) if not k.startswith('__')}
```
I'm assuming it's okay that `dir(BaseConfig)` is now empty; I noticed that `dir(Config)` is used in the non-test code, but the relevant code seems to still work properly with some properties just not defined (instead of always defined with the default value).

Considering all the tests are passing I think it should be okay, I just want to make sure you aren't aware of anything I might be overlooking.

tests/test_config.py
samuelcolvin: That was my proposal - `MyModel.model_config` is really a partial of `ConfigDict`, but has special logic if the value is missing.

pydantic/config.py
```diff
@@ -34,189 +35,167 @@ class Extra(str, Enum):
...
+class BaseConfig(metaclass=ConfigMetaclass):
+    """
+    This class is only retained for backwards compatibility.
 
-    namespace['json_encoders'] = {
-        **getattr(parent_config, 'json_encoders', {}),
-        **getattr(self_config, 'json_encoders', {}),
-        **namespace.get('json_encoders', {}),
-    }
+    The preferred approach going forward is to assign a ConfigDict to the `model_config` attribute of the Model class.
```
dmontagu, line 128: @samuelcolvin Are you good with this docstring? Maybe there is a better way to phrase it. (Also I'm not actually sure if what it says is true, but it seems to match what I saw in some other tests that I assumed was the preferred approach..)

tests/test_config.py
samuelcolvin: yes, that's the idea.

We should just add one explicit test that `BaseConfig.something` returns what we expect (and yields a warning).

I guess we should also check that `BaseConfig().something` does the same, as I'm sure somewhere someone is initialising it.

tests/test_config.py
```diff
@@ -0,0 +1,382 @@
...
+
+            class Config:
+                frozen = True
+
+        m = Model(a=40, b=10)
+        assert m == m.copy()
+
+    def test_config_class_is_deprecated(self):
+        with pytest.warns(
+            DeprecationWarning, match='`BaseConfig` is deprecated and will be removed in a future version'
```
dmontagu, line 334: Even though we didn't discuss it, I added a DeprecationWarning any time you subclass `BaseConfig` (since I figure there's a good chance people will do that without ever accessing the attributes). Let me know if you think we should drop it.

The next test shows the deprecation warnings that are raised for access of any valid config attributes, and the test after that shows the errors raised for attempting to access "invalid" attributes.

samuelcolvin: LGTM.

pydantic/config.py
samuelcolvin: LGTM.

pydantic/config.py
```diff
@@ -34,189 +35,197 @@ class Extra(str, Enum):
...
-    elif self_config == parent_config:
-        base_classes = (self_config,)
-    else:
-        base_classes = self_config, parent_config
+    The preferred approach going forward is to assign a ConfigDict to the `model_config` attribute of the Model class.
+    """
+
+    def __getattr__(self, item: str) -> Any:
+        warnings.warn(
+            f'Support for "config" as "{type(self).__name__}" is deprecated and will be removed in a future version"',
```
samuelcolvin, line 142:
```suggestion
            f'Support for "config" as "{type(self).__name__}" is deprecated and will be removed in a future version',
```

pydantic/config.py
```diff
@@ -34,189 +35,197 @@ class Extra(str, Enum):
...
-    return type('Config', base_classes, namespace)
+def get_config(config: Union[ConfigDict, Dict[str, Any], Type[Any], None]) -> ConfigDict:
+    if config is None:
+        return ConfigDict()
+
+    if isinstance(config, dict):
+        config_dict = config
+    else:
+        warnings.warn(
+            f'Support for "config" as "{type(config).__name__}" is deprecated and will be removed in a future version"',
```
samuelcolvin, line 170:
```suggestion
            f'Support for "config" as "{type(config).__name__}" is deprecated and will be removed in a future version',
```

samuelcolvin (commented):
Couple of tiny things.

This Overall this looks great, and is ready to merge from my POC.

@florianfischer91 are you happy with it? If so, let's merge.

florianfischer91:
@samuelcolvin If you are happy, then i'm too ;-)
I can resolve the conflict in the evening if you want or you resolve it and press the button

samuelcolvin:
Thanks so much @florianfischer91, this is great.

PR #4933, opened 2023-01-11
Update email-validator optional dependency to >= 1.3.0

hramezani:
After updating `email-validator` to >- 1.2.0 in https://github.com/pydantic/pydantic/commit/3c1c3505d68e6d234e26d7ebea2308e262184988
Now we have another error in [Dependencies Check Action](https://github.com/pydantic/pydantic/actions/runs/3889484239/jobs/6637779176)


samuelcolvin:
:+1: thanks.

PR #4947, opened 2023-01-13
JSON Schema 2020-12 [work in progress]

### LINKED ISSUE #4666: V2: JSON Schema
I think we should try to target a specific version of JSON Schema, I guess ideally the latest version, e.g. currently `2020-12`.

See [this thread](https://twitter.com/samuel_colvin/status/1580877321279442945) on twitter for some pointers.

I've heard there are a set of test cases somewhere, would be good if we could test against all or some of them.

I know many people want OpenAPI support too. Maybe we have to build a simple converter or switch to work with the latest version of OpenAPI which I understand is very behind JSON Schema? Maybe a convert already exists?

In terms of how to implement this, might rough suggestion (once #4516 is merged) would be:
* delete everything in `pydantic/schema.py` and start again
* use the pydantic-core schema which is already being generated and saved on models as `__pydantic_validation_schema__` and process that recursively building a JSON Schema
* extra details required by JSON Schema, but not used in pydantic-core should be saved in the `'extra'` key which is already available in pydantic-core schema

You'll need to build something to respect the `__modify_schema__` method which existed in V2. We should rename this to perhaps `__pydantic_modify_json_schema__`, because:
* we're trying to be good (or good-ish) and note add dunder attributes/methods which don't start with `__pydantic`
* we need it to have a name that avoids confusion with pydantic-core schema

@tiangolo I know you originally implemented JSON Schema in pydantic when you were first working on fastapi :pray:, would you be able to implement it again? If not, please unassign yourself and hopefully someone else will take it on.

I've had some interactions with @Relequestual about JSON Schema, maybe he would be able to help too :question: :smile: there's also a JSON Schema slack org which might be able to help.

kennethreitz:
fix #4666.

Hello,

I am working on adding support for [JSON Schema 2020-12](https://json-schema.org/specification.html). This pull request is a work in progress.

Please note that there is a known issue regarding exclusiveMinimum / exclusiveMaximum, which should get resolved as development continues. 

### TODO

- [ ] ensure base-level functionality, and a roadmap moving forward
- [ ] support for arbitrary constraints
- [ ] ensure caching is still active (and desired — it's removed for development currently)
- [ ] think carefully about backwards compatibility re: method names / etc
- [ ] flesh out this list


pydantic/main.py
```diff
@@ -365,29 +367,45 @@ def copy(
...
+    #         default=pydantic_encoder,
+    #         **dumps_kwargs,
+    #     )
 
     @classmethod
-    def schema_json(
-        cls, *, by_alias: bool = True, ref_template: str = default_ref_template, **dumps_kwargs: Any
-    ) -> str:
-        from .json import pydantic_encoder
+    def json_schema(cls, *, ref_template=None):
```
samuelcolvin, line 394: I'll review more tomorrow, but just one quick note now: `json_schema` has been renamed to `model_json_schema`, see #4672

samuelcolvin (commented):
Looks like a good start, a lot more to do here.

Would be good if you addd type annotations and ran linting so it was easier to read.

pydantic/main.py
```diff
@@ -365,29 +365,33 @@ def copy(
...
+    def model_json_schema(cls, *, ref_template=None) -> typing.Dict[str, Any]:
 
-        return cls.__config__.json_dumps(
-            cls.model_json_schema(by_alias=by_alias, ref_template=ref_template),
-            default=pydantic_encoder,
-            **dumps_kwargs,
+        types_namespace = None
+
+        # cls.model_rebuild()
+        inner_schema, fields = _model_construction.build_inner_schema(
```
samuelcolvin, line 385: `cls.__pydantic_validation_schema__`

docs/examples/schema_custom.py
```diff
@@ -1,7 +1,7 @@
 # output-json
 import json
 from pydantic import BaseModel
-from pydantic.schema import schema
+from pydantic.json_schema import schema
```
spacether, line 4: Will openapi discriminator be supported?
If so then should this be openapi_schema?

samuelcolvin: Please see #4666 - JSON Schema 2020-12 is equivalent to OpenAPI 3.1, so that's what we're supporting.

spacether: So I read that and it did not say if discriminator is supported
per https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#properties
Openapi schema is a superset of json schema because it includes additional keywords like discriminator.
Will openapi discriminator be supported?

samuelcolvin: Maybe @Relequestual can give some guidance on this, ideally we would support discriminator or some equivalent but I'd also like to stick to vanilla JSON Schema 2020-12, any ideas?

jeremyfiel: 2020-12 JSON Schema supports `dialects` and `vocabularies` and OpenAPI have created their own dialect on top of the standard JSON Schema dialect to extend the usage of four additional keywords. `xml`, `discriminator`, `externalDocs` and `example`.  As long as you support dialects and vocabularies, any extension should be loaded correctly and validate. 

https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/dialect/base.schema.json

https://json-schema.org/understanding-json-schema/reference/schema.html#:~:text=A%20version%20of%20JSON%20Schema,new%20dialect%20of%20JSON%20Schema.

cc @handrews

jeremyfiel: well, in that case, if you want to support a user wanting to use `discriminator`, a dialect which supports that keyword should be used. In the case of OpenAPI, if they are using OAS3.1.x, the use of the OAS dialect should be preferred. I'm not familiar with another dialect using this keyword. Otherwise, creating a new dialect on the fly with the custom keywords could be another option.

handrews: Yes, you could generate schemas with:

```JSON
{
    "$schema": "https://spec.openapis.org/oas/3.1/dialect/base",
    ...
}
```
instead of the standard 2020-12 dialect meta-schema, which includes all of standard 2020-12 plus the four [OpenAPI dialect extension keywords](https://spec.openapis.org/oas/latest.html#dialectSchemaId): `discriminator`, `xml`, `externalDocs`, and `example` (singular, deprecated in favor of the standard `examples`, plural).

Of those four, `discriminator` might be the only one that's relevant to you, but it is used fairly heavily in OpenAPI.  It will likely be removed in OpenAPI 4 in favor of a standard JSON Schema keyword or keywords that will work a bit differently.  `discriminator` doesn't quite fit the standard JSON Schema processing model.

samuelcolvin:
replaced by #5029.

Issue #4665, opened 2022-10-27
V2: Generics


samuelcolvin:
I think @PrettyWood has some ideas on this.

Basically we want generic models to work on V2, at least as well as they do on V1.

@PrettyWood if you don't think you'll have time to work on this, please unassign yourself, and hopefully someone else can take this on.

Either way, it would be great if you could describe what you were thinking.

PrettyWood:
Now that we don't have cython, generic models should be handled easily following [PEP 484](https://peps.python.org/pep-0484/#user-defined-generic-types). That's all!
But I'll be more than happy to check out if everything works well (validation, schema generation...)

daviskirk:
As I'll probably be using these in various places, I will also try to iron out any kinks I might find

samuelcolvin:
Try to support #4908

samuelcolvin:
@dmontagu who implemented the original Generic functionality mentioned he might be able to work on this.

@PrettyWood hope that's okay, if you're really keen to take this one, let us know and I'm sure that's fine.

PrettyWood:
I'm more than happy to work less :D

caniko:
Please keep [PEP646](https://peps.python.org/pep-0646/), variadic generics, in mind! I haven't tested Pydantic with variadic generic yet, I'll look into it if noone else has.

dmontagu:
@samuelcolvin

It seems to me that if we want to drop the use of `GenericModel` (and I think we should), there are a handful of places we could put the relevant logic for doing different things to the class if `typing.Generic` is one of the bases:
* `BaseModel.__class_getitem__`  (as it currently is in `GenericModel`)
* `ModelMetaclass.__new__`
* `BaseModel.__init_subclass__`
* Somewhere else?

Before I go too far down the rabbit hole I wanted to see if you already had any thoughts about this choice. I know when we implemented `GenericModel`, compatibility with Cython dictated some of the awkwardness; since that's no longer an issue I wanted to make sure I reconsidered all possibilities. (I can't remember if some of the above approaches had other problems that were not related to Cython; guess I'll find out 😅 ).

I'm sure at least some of the logic will belong in `__class_getitem__` (since that's where the new class has to be created) but I was just thinking moving some of the class creation logic (specifically, the typevar substitution) to one of the other methods above _might_ be able to reduce the need for complex calls to `create_model`. Or maybe not, I haven't fully refreshed my recollection of how the typevar substitution happens.

[...]

Issue #4915, opened 2023-01-06
Protected namespaces


samuelcolvin:
See #4617.

We need to:
* add the concept of protected namespaces on models, by default just `model_`
* stop fields being added which conflict with them
* With `config.extra = 'allow'` ignore extra items which conflict with these namespaces - maybe not!

I think it's okay for users to add methods which conflict with namespaces, though we should warn in the docs that that could break in future.

adriangb:
I imagine you've considered this before but I want to bring it up here since it's directly related: could we avoiding protected namespaces altogether by making these methods free floating functions like they are for dataclasses?

```python
from pydantic import BaseModel, to_json


class MyModel(BaseModel):
    json: Any

to_json(MyModel(json={}))
```

I think the main cons would be:
1. User experience. More imports, possible name conflicts with other imports. But this is very common (pandas/numpy, dataclasses and many others are in similar boats) and solvable by doing `import pydantic as pyd` or other options.
2. Caching data / accessing private bits of models. Either `BaseModel` would have to expose internal bits as public or `to_json` would have to access private bits of a class it does not own.

samuelcolvin:
I've thought about this a lot, see #1001. I even pitched a slightly crazy idea at the language summit at pycon last year and got shot down, see [here](https://discuss.python.org/t/better-fields-access-and-allowing-a-new-character-at-the-start-of-identifiers/14529).

I really don't want to loose the great DX of `my_model.<tab>` -- shown list of methods, even though it causes a bunch of problems like this.

However we will need an equivalent of `to_json` and other methods to deal with the case where there's no pydantic model, see #4669, so you could perhaps use them with a `NamedTuple`, `TypedDict` or `dataclass`. If none of those types support extra properly, we could perhaps add a `BareBaseModel` with no methods but that still included `__fields_set__` and `config.extra`. WDYT?

samuelcolvin:
I agreed, also most customisation will be via `@validator` and `@serializer` decorators, not custom implementations of `model_json()` etc.

If we kept the names unchanged:
* `model_dump()`
* `model_dump_json()`
* `model_json_schema()`
* `model_validate_json()`
* etc.

Then it would be even easier for end users.

adriangb:
> I agreed

Sorry just to clarify, are you agreeing that top level methods with convenience aliases on the model class are the way to go?

samuelcolvin:
Yes exactly.

[...]

Issue #4146, opened 2022-06-09
X | Y union syntax breaks GenericModel


ntaylorwss:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.9.1
            pydantic compiled: True
                 install path: /home/paystone/.local/lib/python3.10/site-packages/pydantic
               python version: 3.10.2 (main, Jan 29 2022, 02:55:36) [GCC 10.2.1 20210110]
                     platform: Linux-5.18.2-arch1-1-x86_64-with-glibc2.31
     optional deps. installed: ['typing-extensions']
```

Python 3.10 incorporated [PEP 604](https://peps.python.org/pep-0604/), which added syntax for writing Unions as `X | Y`, instead of `typing.Union[X, Y]`. It seems that when this syntax is used in combination with TypeVars as part of a `GenericModel`, the binding of TypeVars to their types fails.

Below is first a working example using `typing.Union`, followed by a non-working example using `|` syntax, with the stack trace. The example seeks to construct a `Relationship` GenericModel, and a `BidirectionalMultiRelationship` model which contains a list of either `Relationship[A, B]` or `Relationship[B, A]` models:

```py
from typing import Union, Generic, TypeVar
from pydantic.generics import GenericModel

SourceT = TypeVar("SourceT")
TargetT = TypeVar("TargetT")

class Relationship(GenericModel, Generic[SourceT, TargetT]):
    source: SourceT
    target: TargetT

class BidirectionalMultiRelationship(GenericModel, Generic[SourceT, TargetT]):
    relationships: list[Union[Relationship[SourceT, TargetT], Relationship[TargetT, SourceT]]]
    
BidirectionalMultiRelationship[str, int]
```

```py
from typing import Generic, TypeVar
from pydantic.generics import GenericModel

SourceT = TypeVar("SourceT")
TargetT = TypeVar("TargetT")

class Relationship(GenericModel, Generic[SourceT, TargetT]):
    source: SourceT
    target: TargetT

class BidirectionalMultiRelationship(GenericModel, Generic[SourceT, TargetT]):
    relationships: list[Relationship[SourceT, TargetT] | Relationship[TargetT, SourceT]]
    
BidirectionalMultiRelationship[str, int]
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [2], in <cell line: 14>()
[...]

PR #4970, opened 2023-01-20
Initial work on generic models for pydantic v2

dmontagu:
Relevant issue: https://github.com/pydantic/pydantic/issues/4665

I have opened this as a draft pull request to get some initial feedback on:
* The inclusion of `typevars_map` as an argument to various schema-building functions
* The addition of some generics-related methods to `BaseModel`
* Moving some of the existing generics functionality to `pydantic.main` and some to `pydantic.generics_utils`. (Part of the motivation of creating a new file was to keep the PR small/easy-to-review for now. I am imagining that this file will eventually replace the current `pydantic.generics`.)
* The details of `pydantic.generics_utils.create_generic_submodel` and `BaseModel.__class_getitem__`

I am sure that there are still a variety of edge cases that are better-handled in pydantic v1 because I haven't tested them (e.g., anything depending on `_assigned_parameters`, which I haven't needed to migrate yet). 

If I get positive feedback on the approach in this PR, I will review those edge cases more thoroughly.

I have tried to bias toward copying less of the existing code when I didn't fully understand what it was used for, though I intend to check more thoroughly for any regressions soon (definitely before merging). Hopefully the mostly-disabled tests of `GenericModel` cover most of these..


pydantic/main.py
```diff
@@ -592,6 +603,51 @@ def __repr_args__(self) -> _repr.ReprArgs:
...
+    def __concrete_name__(cls: type[Any], params: typing.Tuple[type[Any], ...]) -> str:
+        """Compute class name for child classes.
+
+        :param params: Tuple of types the class . Given a generic class
+            `Model` with 2 type variables and a concrete model `Model[str, int]`,
+            the value `(str, int)` would be passed to `params`.
+        :return: String representing a the new class where `params` are
+            passed to `cls` as type variables.
+
+        This method can be overridden to achieve a custom naming scheme for GenericModels.
```
dmontagu, line 645: Obviously need to update these docs, but I wasn't sure how you were going to feel about keeping `__concrete_name__` as a method on `BaseModel`. Having the ability to modify the name-generation logic feels like a significant benefit, I just feel a bit averse to polluting the `BaseModel` namespace.

pydantic/_internal/_generate_schema.py
```diff
@@ -117,6 +124,26 @@ def generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
             if issubclass(obj, dict):
                 return self._dict_subclass_schema(obj)
             # probably need to take care of other subclasses here
+        elif isinstance(obj, typing.TypeVar):
+            # TODO: Add a check for recursion depth exception and an explanatory re-raise
```
dmontagu, line 128: (note to self -- need to delete this comment; the exception handling has already been added below)

pydantic/generics_utils.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Any
+
+TypeVarType = Any  # since mypy doesn't allow the use of TypeVar as a type
+Parametrization = typing.Mapping[TypeVarType, type[Any]]
+
+if typing.TYPE_CHECKING:
+    from pydantic import BaseModel
+
+    class GenericBaseModel(BaseModel):
+        __parameters__: typing.ClassVar[typing.Tuple[TypeVarType, ...]]
```
dmontagu, line 13: This type-checking-only `GenericBaseModel` class, along with the `is_generic_model` typeguard below, are just there to help address the fact that mypy doesn't like when I access `__parameters__` on a class that inherits from both `BaseModel` and `typing.Generic`. This approach seemed to handle it correctly with minimal mypy ignores.

pydantic/generics_utils.py
```diff
@@ -0,0 +1,58 @@
...
+        previous_caller_frame = sys._getframe(depth)
+    except ValueError as e:
+        raise RuntimeError('This function must be used inside another function') from e
+    except AttributeError:  # sys module does not have _getframe function, so there's nothing we can do about it
+        return None, False
+    frame_globals = previous_caller_frame.f_globals
+    return frame_globals.get('__name__'), previous_caller_frame.f_locals is frame_globals
+
+
+def is_generic_model(x: type['BaseModel']) -> typing.TypeGuard[type['GenericBaseModel']]:
```
dmontagu, line 57: Note that, where used, this type guard _is required_ at runtime (for the sake of generating an appropriate error message if you try to parameterize a non-generic BaseModel). In particular, even though the type guard provides typing benefits, and even though the type it is "guarding" isn't even "real", the runtime overhead of the check is still required.

pydantic/main.py
```diff
@@ -592,6 +603,51 @@ def __repr_args__(self) -> _repr.ReprArgs:
             if not k.startswith('_') and (k not in self.model_fields or self.model_fields[k].repr)
         ]
 
+    def __class_getitem__(cls, item: typing.Union[type[Any], typing.Tuple[type[Any], ...]]) -> type[Any]:
```
samuelcolvin, line 606: i think you can use `|` union syntax here.

pydantic/main.py
```diff
@@ -42,6 +43,10 @@
 # the `BaseModel` class, since that's defined immediately after the metaclass.
 _base_class_defined = False
 
+_generic_types_cache: _utils.LimitedDict[
+    typing.Tuple[type[Any], Any, typing.Tuple[Any, ...]], type[BaseModel]
```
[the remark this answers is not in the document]
samuelcolvin: It might be worth looking at #3829 #4551 to check we don't introduce any of those problems again - the memory leak issue should be solved by `LimitedDict`, but the `TypeError` could still exist I guess.

pydantic/main.py
```diff
@@ -19,6 +19,7 @@
 from .config import BaseConfig, Extra, build_config, inherit_config
 from .errors import PydanticUserError
 from .fields import Field, FieldInfo, ModelPrivateAttr
+from .generics_utils import create_generic_submodel, is_generic_model
```
samuelcolvin, line 22: to avoid private methods being publicly importable (e.g. with `from pydantic.main import create_generic_submodel`, better to add `_generics` to the `from ._internal import ...` line.

pydantic/main.py
```diff
@@ -592,6 +603,51 @@ def __repr_args__(self) -> _repr.ReprArgs:
...
+    def __class_getitem__(cls, item: typing.Union[type[Any], typing.Tuple[type[Any], ...]]) -> type[Any]:
+        def _cache_key(_params: Any) -> typing.Tuple[type[Any], Any, typing.Tuple[Any, ...]]:
+            return cls, _params, typing_extensions.get_args(_params)
+
+        cached = _generic_types_cache.get(_cache_key(item))
+        if cached is not None:
+            return cached
+
+        if not isinstance(item, tuple):
+            item = (item,)
```
samuelcolvin, line 615: is it worth having a more descriptive name, e.g.
```suggestion
            typevar_values = (item,)
        else:
            typevar_values = item
```
???

pydantic/main.py
```diff
@@ -592,6 +603,51 @@ def __repr_args__(self) -> _repr.ReprArgs:
...
+
+        if not is_generic_model(cls):
```
samuelcolvin, line 617: should this come first in the function.

pydantic/main.py
samuelcolvin: new rule for v2 to try and be better python citizens_ we want all dunder names to start with `__pydantic` (or at least include pydanitc, e.g. `__get_pydantic...`).

Should we use `__pydantic_concrete_name__` here?

(I assume `__concrete_name__` is not defined in the stdlib, or a pepe somewhere?)

Or, since this is not really private or magic, maybe we should just use `model_concrete_name`???

samuelcolvin: I I think I prefer a public name, e.g. `model_concrete_name`.

pydantic/main.py
```diff
@@ -592,6 +603,51 @@ def __repr_args__(self) -> _repr.ReprArgs:
...
+
+        :param params: Tuple of types the class . Given a generic class
+            `Model` with 2 type variables and a concrete model `Model[str, int]`,
+            the value `(str, int)` would be passed to `params`.
+        :return: String representing a the new class where `params` are
+            passed to `cls` as type variables.
+
+        This method can be overridden to achieve a custom naming scheme for GenericModels.
+        """
+        param_names = [_repr.display_as_type(param) for param in params]
```
samuelcolvin, line 647: if this is only used on generics, should we raise an error if the model is not generic?

samuelcolvin (commented):
Overall looks great.

Quite satisfying to see that the new core/core-schema approach has simplified this a lot, at least that's how I see it?!

Might be worth looking through the `1.10.X-fixes` branch and see if there are any bug fixes there that should be included here? E.g. #4551 mentioned below, you might just copy any new tests and check they pass.

Please update.

pydantic/main.py
dmontagu: @samuelcolvin I'm changing it to `model_concrete_name` for now, but one other alternative I'd like to propose: for public methods, using `model__` as a prefix, rather than `model_` (so `model__concrete_name` instead of `model_concrete_name`; similarly `model__json`, etc.).

This feels significantly less likely to run into collisions with user-desired fields (or methods), and to my eye almost *more* readable (as I can immediately tell it is a pydantic model method, since nothing else uses a format like that), with only one extra character. And it retains the various benefits of the method name being pseudo-public (e.g., IDEs including it at the top of autocomplete, etc.).

If you consider this an already-finalized decision, or just hate this proposal, let's just pretend this comment never happened :).

pydantic/_internal/_generate_schema.py
```diff
@@ -117,6 +124,26 @@ def generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
...
+                            f'The maximum recursion depth was exceeded while generating the schema for a TypeVar. '
+                            f'This likely indicates a cycle in the TypeVar substitutions map ({self.typevars_map=}). '
+                            'This may be resolved by using the TypeVars in the same order as the original '
+                            'parameterization, or by using entirely new ones.'
+                        ) from exc
+            if obj.__bound__:
+                return self.generate_schema(obj.__bound__)
+            if obj.__constraints__:
+                return self._union_schema(typing.Union[obj.__constraints__])
+            return core_schema.AnySchema(type='any')
```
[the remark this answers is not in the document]
dmontagu: I did think about this; I spent some time thinking through the implications of disallowing model classes with free typevars from being instantiated at all, and ultimately I think this approach is preferable.

I couldn't think of any cases where an error was merited that mypy wouldn't be able to catch with an appropriate configuration if you wanted to be aggressive about your types (i.e., using generic models with implicit Any parameter). And it felt to me that the flexibility benefits of being able to work with unspecified generics was worth more than other forms of strictness here.

-----

The only pattern that I could imagine being problematic would be something along the following lines:
```python
class MyGenericModel(BaseModel, Generic[T]):
    my_field: T

MyConcreteModel = MyGenericModel[MyType]
```
If there was a non-pathological example where `MyGenericModel(my_field=instance_of_my_type).my_field` could end up different than `MyConcreteModel(my_field=instance_of_my_type).my_field` due to (again, non-pathological) validator usage, etc., I think that would be concerning. I couldn't think of anything that would cause this though without going out of our way (e.g., by having non-idempotent validators or something like that).

pydantic/main.py
dmontagu: yeah, I'm in favor. I noticed that the `__class_getitem__` method is defined on `GenericModel` using `params` as the name of the argument; I renamed it to `item` here following the documentation of `__class_getitem__` but just realized that I think we used `params` as the name in `GenericModel` because that is what it is called in `typing.Generic`. I checked and confirmed that the function `__class_getitem__` is called using an arg, rather than a kwarg, so I was thinking we could maybe just call the argument to the `__class_getitem__` function `typevar_values`.

I think that would be my first preference, since you can then see the intent as soon as you see the function signature, but I could just use your suggestion here if you prefer. (I do think either way is an improvement.)

pydantic/main.py
dmontagu: I'm okay either way; I can't really imagine a circumstance where you'd want to use it for non-generic-models, if someone found a use case where it was worth using on non-generic models for some reason, they could always override the method to drop the check for generic and open an issue requesting it work for non-generic models..

pydantic/main.py
samuelcolvin: let's still with `model_concrete_name` here for consistency.

Happy to consider `model__` prefix, probably worth discussing on #4915 or in a new issue.

I dislike all proposals in this area, including my own - there's no elegant way around this problem.

pydantic/_internal/_generate_schema.py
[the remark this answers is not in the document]
samuelcolvin: okay, let's leave it.

pydantic/main.py
samuelcolvin: whichever you prefer, it's not a big issue.

dmontagu:
@samuelcolvin while investigating https://github.com/pydantic/pydantic/issues/4146#issuecomment-1399677854, I noticed the following unintuitive (to me) behavior:

```python

DataT = TypeVar("DataT")

class MyGenericModel(BaseModel, Generic[DataT]):
    generic: list[DataT] | None


class A(BaseModel):
    wrapper: MyGenericModel[int]

m = MyGenericModel(generic=None)
print(A(wrapper=m))
```

```
  File "scratch_66.py", line 14, in <module>
    print(A(wrapper=m))
          ^^^^^^^^^^^^
  File "pydantic/pydantic/main.py", line 155, in __init__
    values, fields_set = __pydantic_self__.__pydantic_validator__.validate_python(data)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyGenericModel[int]
wrapper
  Input should be a valid dictionary [type=dict_type, input_value=MyGenericModel(generic=None), input_type=MyGenericModel]
```

This is happening because `m` here is an instance of `MyGenericModel`, which is a _superclass_ of `MyGenericModel[int]`. There's nothing specific to generics with this error — this is basically the same as doing:
```python


class Inner(BaseModel):
    x: int


class SubInner(Inner):
    pass


class Outer(BaseModel):
    wrapper: SubInner


Outer(wrapper=Inner(x=1))
```

However, I can't help but feel it's a little awkward that you _have_ to add the annotation to the type when initializing it to get it to pass validation as a field.

I'm not sure I have a great proposal for a resolution to this (though I think we might be able to do something where we re-validate the data for concrete parametrizations of generic models, even if we don't do that for more general model superclasses). I don't think it needs to be resolved now (or possibly ever), as things are still plenty functional if you have to use the right parametrization when instantiating the class, it just feels a little awkward since you don't need to do that for non-pydantic generic classes.

Curious if you have any thoughts.

samuelcolvin:
Very easy to add a flag to either config or a model to skip the isinstancd check - just run validation against `.__dict__` and see what happens.

With this nascent flag set, any class instance with the right fields would pass, maybe that's okay.

The problem is when to enable it? I guess we could enable it for generics by default?

We could also do some kind of "is super class" check, which sounds weird but could be useful.

dmontagu:
> We could also do some kind of "is super class" check, which sounds weird but could be useful.

While addressing some of the more complex tests I realized I was running into some awkwardnesses with the class hierarchy being generated (M[int, T][str] would be two "layers" down from M, whereas M[int, str] is only one; M[int, T][str] was a different class than M[T, str][int] was a different class than M[int, str]).

To handle this, I made some more extensive changes to the implementation (adding fields like `__pydantic_generic_parent__` to hold a reference to the last class created the "normal" way, which then gets reparametrized using the full set of arguments when you do partial parametrizations). This means that no matter how you go about doing a partial parametrization, you should get class equality if the parametrization is the same, and parametrization only adds one layer to the class hierarchy no matter how many times `__class_getitem__` is called (with unsubstituted TypeVars).

I think it may be possible to leverage this approach for a more (appropriately) limited version of the "superclass" check you are describing in a straightforward way -- basically allowing instances of the unparametrized parent class. But, I'm not totally sure that this is something that makes sense.

Also, just to be clear, I'm concerned about the increase in complexity I've added but I think it's going to be necessary to handle some cases of deeply nested models with the same TypeVar used at multiple levels in the hierarchy; the more complex edge cases for which bugfixes and, thankfully, tests have been added over the last couple years :). I am pushing my progress directly here but I want to be clear that right now I'm mostly trying to find a path forward that works to handle the various edge cases, and will work on simplifying / reducing the amount of new/unnecessary fields once I figure out _something_ that works.

dmontagu:
One interesting thing I noticed -- when using a bounded or constrained TypeVar, we don't do class-creation-time validation of parameter compatibility.

In particular:
```python

T = TypeVar("T", bound=str)


    data: T

Model[int](data=1)  # no error due to TypeVar substitution with `int`, despite being a type system error
Model(data=1)  # error thanks to bound on TypeVar requiring an instance of `str`
```

This _would_ be caught by mypy, and I'm thinking we probably don't want to get into the business of being a runtime type **system** _in addition_ to being a runtime type validator, but just wanted to point this out.

-----

On a semi-related note, when it comes time to implement subclass checking, we _may_ want to use information about the covariance/contravariance of the TypeVars involved; I'm not sure whether 1.X does that currently (I'm assuming not, but I need to check how that all works). I guess it would come down to how hard it is to do runtime validation of subclass relationships of type parameters, writing that out makes me think probably not going to happen, but I'll look into it a bit.

dmontagu:
On the note of compatibility with runtime subclass checking of generics:
```python
from typing import Sequence

issubclass(Sequence[int], Sequence[float])

# TypeError: Subscripted generics cannot be used with class and instance checks
```

@samuelcolvin I'm wondering if disallowing subclass checks entirely with parameterized generics might be smart for reducing the amount of complex behavior we need to support; notice PyCharm even shows an error when you try (this is the case even if you introduce an alias for `A[int]` below):

image

I can imagine some advantages to supporting this, but I'm wondering if it's smarter to leave this to the type system and use a discriminated-union-esque approach if you really need to differentiate between specific generic parametrizations at runtime.

samuelcolvin:
We've agreed that we don't need subclass stuff to work since it doesn't work with typing generics.

caniko:
I don't see [PEP646](https://peps.python.org/pep-0646/) or variadic generics, being mentioned in this draft. Are you planning on adding this later?

dmontagu:
I've paused work on this temporarily while trying to work through JSON schema and some other smaller things, and trying to familiarize myself more with pydantic_core, where I want to move at least some of the generics logic to get around some of the more complex challenges (such as recursive generic models with long cycles).

That said, I am hoping to add support for variadic generics as part of this. If it turns out to be particularly complicated I may postpone that in favor of getting non-variadic generic models working, but I am (cautiously?) optimistic.

To be clear, the use case I have in mind would be something like:

```python
```

caniko:
This is a fine example:

> ```python
> class MyGeneric(BaseModel, Generic[*Shape]):
>     x: int
>     y: tuple[*Shape]
> ```

Moreover, in `>=3.11`: 
```python
Tuple[str, *Tuple[int, int], bool] -> Tuple[str, int, int, bool]
```
This is an extension of the logic from your example, I suggest that you implement with unpacking in mind.

Also keep in mind that normal types like `TypeVar` instances can precede a `TypeVarTuple` instance:
```python
T = TypeVar('T')    # or NewType('T')
Y = TypeVar('Y')    # or NewType('Y')
Shape = TypeVarTuple('Shape')

class MyGeneric(BaseModel, Generic[T, Y, *Shape]):
    x: T
    y: tuple[*Shape]
    z: list[Y, *Shape]
```

dmontagu:
Closing in favor of a new branch that I think is closer.

PR #4991, opened 2023-01-29
BaseSettings: Support `Field(env=...)` in nested sub models

### LINKED ISSUE #4982: BaseSettings: Environment variable is not loaded in nested model if set in `Field()`
### Initial Checks

### Description

I have a class `BaseSettings` with a nested `BaseModel`. Within the nested model (named `nested`), I defined a variable with `sub_var1: str = Field(..., env="SUB_VAR")` (as mentioned in the [docs](https://docs.pydantic.dev/usage/settings/#environment-variable-names)).

I would expect that the environment variable name specified in the nested `Field` is loaded from my `.env` file. But it's not (see the example below). My `.env` file looks as follows:

```
MAIN_VAR=main
SUB_VAR=sub
```

It only works with `env_nested_delimiter='__'` and an environment variable `NESTED__SUB_VAR1` but I would expect it to work with `SUB_VAR` (as it's set in the `Field`).

I'm not sure if this is a bug or expected behavior. If it's the latter I'm doing a feature request, I guess.

### Example Code

```Python
from pydantic import BaseSettings, BaseModel, Field


class SubModel(BaseModel):
    sub_var1: str = Field(..., env="SUB_VAR")
    sub_var2: str = "bar"


class Settings(BaseSettings):
    main_var: str = Field(..., env="MAIN_VAR")
    nested: SubModel

    class Config:
        env_file = '.env'
        env_nested_delimiter = '__'


print(Settings().dict())
```

```Text
pydantic version: 1.10.4
            pydantic compiled: True
                 install path: /home/rafael/miniconda3/envs/ds/lib/python3.8/site-packages/pydantic
               python version: 3.8.13 (default, Mar 28 2022, 11:38:47)  [GCC 7.5.0]
                     platform: Linux-5.15.0-58-generic-x86_64-with-glibc2.17
```

### Affected Components

RafaelWO:
Fixes #4982 

This MR fixes the issue of environment variable names defined in nested sub-models of a `BaseSettings` class via `Field(env=...)` not being recognised.

I implemented some recursion mechanism to get those environment variables specified in nested models. This may not be the best way but it works. Of course, I'm happy for any feedback and suggestions to make this fix cleaner.

#### Additional Details / Questions
I noticed that if the nested models are subclasses of `BaseSettings` (instead of `BaseModel`), the recursive call of `BaseSettings.Config.prepare_field()` would not be needed. I also saw people using nested `BaseSettings`, however, the docs only show this with `BaseModel`. Are both ways supposed to work? Or should we use `BaseSettings` only for the "root" class?

### DIFF
--- a/changes/4991-RafaelWO.md
+++ b/changes/4991-RafaelWO.md
@@ -1,0 +1,1 @@
+BaseSettings: Support the definition of environment variable via `Field(env=...)` in nested sub models 
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -91,6 +91,15 @@ class BaseSettings(BaseModel):
 
         @classmethod
         def prepare_field(cls, field: ModelField) -> None:
+            # Set 'env_names' for nested sub models of type 'BaseModel'
+            if (
+                isinstance(field.type_, type)
+                and issubclass(field.type_, BaseModel)
+                and not issubclass(field.type_, BaseSettings)
+            ):
+                for name, subfield in field.type_.__fields__.items():
+                    cls.prepare_field(subfield)
+
             env_names: Union[List[str], AbstractSet[str]]
             field_info_from_config = cls.get_field_info(field.name)
 
--- a/pydantic/env_settings.py
+++ b/pydantic/env_settings.py
@@ -91,12 +91,8 @@ class BaseSettings(BaseModel):
 
         @classmethod
         def prepare_field(cls, field: ModelField) -> None:
-            # Set 'env_names' for nested sub models of type 'BaseModel'
-            if (
-                isinstance(field.type_, type)
-                and issubclass(field.type_, BaseModel)
-                and not issubclass(field.type_, BaseSettings)
-            ):
+            # Set 'env_names' for nested sub models of type 'BaseModel' (except 'BaseSettings')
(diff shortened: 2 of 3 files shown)


samuelcolvin:
Humm, bit worried this could be considered a breaking change - we're really trying to limit changes in V1.10 to the simplest/most obvious bugs.

Definitely keen to fix this in V2, but settings has moved to [pydantic-settings](https://github.com/pydantic/pydantic-settings).

Happy to hear what others think ... @pydantic/pydantic-maintainers?

hramezani:
Same feeling @samuelcolvin. by checking the implementation it doesn't seem simple to me. I am afraid of introducing a hidden bug.

RafaelWO:
I understand your concerns. I have to admit that I'm not fully happy with my solution - although I can not think of a much simpler implementation for this based on the current logic of loading the environment variables. But it's also my first time touching the code here.

Maybe we should leave this for V2, i.e. implement it in pydantic-settings then? Not sure if there is a greater refactoring planned which potentially would change how env variables are loaded.

samuelcolvin:
> Maybe we should leave this for V2, i.e. implement it in pydantic-settings then? Not sure if there is a greater refactoring planned which potentially would change how env variables are loaded.

There is, see https://github.com/pydantic/pydantic-settings/pull/10.

I think we're agreed on this one.

PR #5007, opened 2023-02-05
Invalidate mypy cache if config changes

cdce8p:
Implement `Plugin.report_config_data` to return all plugin config data. For every run, mypy will compare if the data matches the cached one if not the cache is discarded.

`report_config_data` is called multiple times for every file. I would have liked to use `cached_property` but that requires `3.8`, unfortunately.

https://github.com/python/mypy/blob/v0.991/mypy/plugin.py#L539

_This is my first PR here, so I wasn't quite sure which branch to target as this is an issue in both `1.10.x` and `main`. Please let me know if I should change anything and I'll do so._

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -42,6 +42,7 @@ from mypy.plugin import (
     FunctionContext,
     MethodContext,
     Plugin,
+    ReportConfigContext,
     SemanticAnalyzerPluginInterface,
 )
 from mypy.plugins import dataclasses
@@ -98,6 +99,7 @@ def plugin(version: str) -> 'TypingType[Plugin]':
 class PydanticPlugin(Plugin):
     def __init__(self, options: Options) -> None:
         self.plugin_config = PydanticPluginConfig(options)
+        self._plugin_data = self.plugin_config.to_data()
         super().__init__(options)
 
     def get_base_class_hook(self, fullname: str) -> 'Optional[Callable[[ClassDefContext], None]]':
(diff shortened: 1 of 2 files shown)


pydantic/mypy.py
```diff
@@ -210,6 +215,9 @@ def __init__(self, options: Options) -> None:
                 setting = plugin_config.getboolean(CONFIGFILE_KEY, key, fallback=False)
                 setattr(self, key, setting)
 
+    def to_data(self) -> Dict[str, Any]:
```
samuelcolvin, line 218: Do we need a separate method for this?

pydantic/mypy.py
```diff
@@ -126,6 +128,9 @@ def get_class_decorator_hook(self, fullname: str) -> Optional[Callable[[ClassDef
             return dataclasses.dataclass_class_maker_callback  # type: ignore[return-value]
         return None
 
+    def report_config_data(self, ctx: ReportConfigContext) -> Dict[str, Any]:
```
samuelcolvin, line 131: Please add a docstring, the info in the pr body is great.

samuelcolvin (commented):

Please change the base to `main`, if we think it should be backported to v1.10, we can do that easily.

If you want it backported, please also add a change description.

pydantic/mypy.py
cdce8p: Not necessarily. Just though it might be best to let the `PluginConfig` class handle it.

Alternatives would be to move the dict build directly to `PydanticPlugin.__init__` or create a separate `data` attribute for `PluginConfig`. The later one is a bit tricky as it would also require a dedicated `fields` attribute then.

cdce8p:
> Please change the base to `main`, if we think it should be backported to v1.10, we can do that easily.
> 
> If you want it backported, please also add a change description.

Backporting it later would be great. I can open a separate PR for it after this one is merged.

The change description in `HISTORY.md`? Should I add it here already or only for the backport PR?

pydantic/mypy.py
samuelcolvin: Ok, keep as is.

samuelcolvin:
No, see https://github.com/pydantic/pydantic/tree/main/changes

We have a bot to backport.

samuelcolvin:
thanks so much.

Issue #5008, opened 2023-02-05
[v2.dev]  Pydantic does not support custom annotation types


ddneilson:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

[PEP 593](https://peps.python.org/pep-0593/) says "If a library (or tool) encounters a typehint Annotated[T, x] and has no special logic for metadata x, it should ignore it and simply treat the type as T."

The current in-development branch of Pydantic v2 throws an error if it encounters Annotated metadata that are not "instances of annotated_types.BaseMetadata or PydanticMetadata or a subclass of PydanticMetadata".

I believe this to be a bug. Expected behavior is that Pydantic preserves the added metadata, but ignores it for its own processing.

As a motivating use-case, consider an application that uses Pydantic v2 for parsing & validating a data model, but then wants to do some post-processing/understanding of the data provided. Such an application (like my own) may find that the ideal mechanism for that post-processing is a recursive traversal through the model with functionality during the traversal dictated by annotations on the fields of the model. These annotations are not for parsing & validation, and should be ignored by Pydantic.

### Example Code

```Python
import sys
from pydantic import BaseModel

if sys.version_info < (3,9):
    from typing_extensions import Annotated
else:
    from typing import Annotated

class MyAnnotation:
    pass

class TestModel(BaseModel):
    attr: Annotated[str, MyAnnotation]

# Traceback (most recent call last):
#   File "testannotation3.py", line 12, in <module>
#     class TestModel(BaseModel):
#   File "/home/dneilson/python/.venv/lib/python3.8/site-packages/pydantic/main.py", line 103, in __new__
#     _model_construction.complete_model_class(
#   File "/home/dneilson/python/.venv/lib/python3.8/site-packages/pydantic/_internal/_model_construction.py", line 141, in complete_model_class
#     inner_schema, fields = build_inner_schema(
#   File "/home/dneilson/python/.venv/lib/python3.8/site-packages/pydantic/_internal/_model_construction.py", line 251, in build_inner_schema
#     schema = model_fields_schema(
[...]

PR #5010, opened 2023-02-05
follow pep593 by preserving unknown metadata in annotation

ddneilson:
This is a PR for pydantic v2. 

It fixes [issue 5008](https://github.com/pydantic/pydantic/issues/5008) by allowing, and ignoring, Annotated metadata that is of type(s) that pydantic does nothing with.

### DIFF
--- a/tests/test_annotated.py
+++ b/tests/test_annotated.py
@@ -6,7 +6,6 @@ from annotated_types import Gt, Lt
 from typing_extensions import Annotated
 
 from pydantic import BaseModel, Field
-from pydantic.errors import PydanticSchemaGenerationError
 from pydantic.fields import Undefined
 
 NO_VALUE = object()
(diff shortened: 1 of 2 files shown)


tests/test_annotated.py
```diff
@@ -111,6 +105,22 @@ class M(BaseModel):
             x: hint = value
 
 
+@pytest.mark.parametrize(
+    'metadata',
+    [
+        0,
+        'foo',
+    ],
+)
```
samuelcolvin, line 114: nit-pick - you can remove the trailing commas and combine this onto one line.

tests/test_annotated.py
```diff
@@ -111,6 +105,22 @@ class M(BaseModel):
...
+def test_annotated_allows_unknown(metadata):
+    class M(BaseModel):
+        x: Annotated[int, metadata] = 5
+
+    field_info = M.model_fields['x']
+    assert len(field_info.metadata) == 1
+    assert metadata in field_info.metadata, 'Records the unknown metadata'
```
samuelcolvin, line 121: we should also check that `metadata` is included in `M.__annotations__` - that's where any other library would look for is.

samuelcolvin (commented):
LGTM, just a couple of tiny things.

please update.

PR #5011, opened 2023-02-06
Support One Time Password as a Type

yezz123:
```py
import random

from pydantic.types import OTP, OTP_ALPHABET


class GenerateOTP(BaseModel):
    """
    Generate an OTP
    """

    length: int = Field(6, ge=6, le=6)
    alphabet: str = Field(OTP_ALPHABET, min_length=6, max_length=32)

    def generate(self) -> OTP:
        return OTP(''.join(random.choices(self.alphabet, k=self.length)))
```

cc @samuelcolvin @hramezani what do you think about supporting OTP as a type?


samuelcolvin:
Closing, see #5012 - in short I think we should add this and others to a new `pydantic-extra` package - @yezz123 any chance you'd be willing to help maintain such a package.

yezz123:
> Closing, see #5012 - in short, I think we should add this and others to a new `pydantic-extra` package - @yezz123 any chance you'd be willing to help maintain such a package

Let's move with the idea of `pydantic-extra` I like it 🚀

Issue #5019, opened 2023-02-06
Support custom generic container types in v2


dmontagu:
In v2, currently the following code:
```python
from typing import TypeVar, List

from pydantic import BaseModel

T = TypeVar('T')


class GenericList(List[T]):
    pass


class Model(BaseModel):
    field: GenericList[int]
```
results in:
```
pydantic_core._pydantic_core.SchemaError: Invalid Schema:
typed-dict -> fields -> field -> schema
  Input tag 'genericlist' found using self-schema does not match any of the expected tags: 'any', 'none', 'bool', 'int', 'float', 'str', 'bytes', 'date', 'time', 'datetime', 'timedelta', 'literal', 'is-instance', 'is-subclass', 'callable', 'list', 'tuple-positional', 'tuple-variable', 'set', 'frozenset', 'generator', 'dict', 'function', 'function-wrap', 'function-plain', 'default', 'nullable', 'union', 'tagged-union', 'chain', 'lax-or-strict', 'typed-dict', 'model', 'arguments', 'call', 'recursive-ref', 'custom-error', 'json', 'url', 'multi-host-url' [type=union_tag_invalid, input_value={'type': 'genericlist', '...chema': {'type': 'int'}}, input_type=dict]
```

I believe this was addressed in v1.X by PR #2554; that PR referenced this comment: https://github.com/pydantic/pydantic/issues/2465#issuecomment-801497216

samuelcolvin:
Should be able to cover this with the same logic we use for deque.

dmontagu:
I think actually we _maybe_ shouldn't support this (though we should make sure we get a good error), but should instead raise an error message that links to documentation for how to implement `__get_pydantic_core_schema__` for the custom generic class.

The gist I link in this comment: https://github.com/pydantic/pydantic/issues/2274#issuecomment-1528082601 shows how to create a non-contrived pydantic core schema for a custom generic container, I think we should put something like this in the docs and ideally link to it in the error message.

(In this PR: https://github.com/pydantic/pydantic/pull/5629 I implement similar "error for custom generic containers" logic specifically for subclasses of `Sequence`)

adriangb:
It's a bit unclear to me what the original issue is asking for. I see two different things going on:
- Generic containers: we fully support these, at least for the common standard library types `dict`, `set`, `Counter`, etc.
- Subclasses of known types: these are supported, you just need to implement `__get_pydantic_core_schema__` and you'll get an error telling you to do so if you don't

[...]

PR #5020, opened 2023-02-06
Handle custom collections that are subclasses of list, set, frozenset

dmontagu:
Partially addresses #5019 

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -216,12 +216,21 @@ class GenerateSchema:
         elif obj == re.Pattern:
             return self._pattern_schema(obj)
         elif isinstance(obj, type):
+            if obj is dict:
+                return self._dict_schema(obj)
             if issubclass(obj, dict):
+                # TODO: We would need to handle generic subclasses of certain typing dict subclasses here
+                #   This includes subclasses of typing.Counter, typing.DefaultDict, and typing.OrderedDict
+                #   Note also that we may do a better job of handling typing.DefaultDict by inspecting its arguments.
                 return self._dict_subclass_schema(obj)
             # probably need to take care of other subclasses here
         elif isinstance(obj, typing.TypeVar):
             return self._unsubstituted_typevar_schema(obj)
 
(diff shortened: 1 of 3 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -312,22 +312,39 @@ def _generate_parameter_schema(
...
+                    break
+
+        schema: core_schema.CoreSchema = {  # type: ignore[misc,assignment]
+            'type': schema_type.__name__.lower(),
             'items_schema': self.generate_schema(get_first_arg(type_)),
         }
 
+        if is_proper_subclass:
+            # Ensure the validated value is converted back to the specific subclass type
+            return core_schema.function_after_schema(schema, lambda __input_value, **kwargs: type_(__input_value))
```
dmontagu, line 344: @samuelcolvin Is this along the lines you had in mind? I took the idea of the `function_after_schema` call from `deque_schema`, and otherwise just did modifications to the `_generic_collection_schema` function.

It might be possible to pull the handling of proper subclasses into a separate function if you don't want to muddle the `_generic_collection_schema` method as much; let me know what you think.

dmontagu:
This still needs handling of other collection types (e.g., Dict/Mapping/etc.), but I wanted to get some feedback before going farther down this path.

pydantic/_internal/_generate_schema.py
samuelcolvin: I started trying to explain what I thought through a comment, and decided it was easier to contribute directly to the PR.

I hope you don't mind.

I think this approach is easier, we re-use information we already had outside the function. Let me know what you think.

No idea if it works in all cases, but it works in more cases than what we had before this PR :-).

samuelcolvin:
Where are we on this?

Please update.

dmontagu:
I have done what I think is a fairly thorough job investigating how to handle custom subclasses of the generic types not handled in this PR (list, set, frozenset). They all have minor issues that make it so that handling them properly would need to be a bigger change, so I don't think it makes sense to fix that in this PR. I have documented the issues in comments, and have updated the relevant test to more thoroughly demonstrate the behavior for annotations of each custom generic type.

please review

Issue #1511, opened 2020-05-13
Decimal JSON encoder is lossy


lsorber:
## Version
```
             pydantic version: 1.5.1
            pydantic compiled: True
                 install path: 
               python version: 3.8.2 (default, May  6 2020, 02:49:43)  [Clang 4.0.1 (tags/RELEASE_401/final)]
                     platform: macOS-10.15.4-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions', 'email-validator']
```

## Bug

```py
from pydantic import BaseModel

class MyModel(BaseModel): 
    foo: Decimal

m = MyModel(foo="31231239131093781902381093810298310983123.12000000")
m.json()  # {"foo": 3.1231239131093784e+40} is lossy
```

## Proposed solution

Replace the current `Decimal: float` JSON encoder [1] with `Decimal: str`.

[1] https://github.com/samuelcolvin/pydantic/blob/master/pydantic/json.py#L27

samuelcolvin:
I think we used to have it like that but changed it since the current solutions is most user-friendly in most scenarios.

You can change this yourself on a model (and use a custom base model to extend the functionality to all your models):

```py
from decimal import Decimal

from devtools import debug
from pydantic import BaseModel


class MyModel(BaseModel):
    foo: Decimal

    class Config:
        json_encoders = {Decimal: str}


m = MyModel(foo='31231239131093781902381093810298310983123.12000000')
debug(m.json())
"""
> test.py:14 <module>
    m.json(): '{"foo": "31231239131093781902381093810298310983123.12000000"}' (str) len=61
"""
```

lsorber:
If I may challenge that decision: the primary use case for using Decimals is that you need to store and work with arbitrary precision decimals. Serialising these as floats will cause loss of information. If you don't care about storing all digits, then you might as well use `float` instead of `Decimal`.

In other words: if you all you need is convenience, we already have `float`. If you need arbitrary precision, you need `Decimal`. It would somewhat disappointing that users that need arbitrary precision also need a workaround to actually store those Decimals without loss of precision.

hassanselim0:
I agree with @lsorber here, the whole idea of using an uncommon datatype like `Decimal` is to get full precision like when dealing with monetary calculations. So having the default encoding be a JSON Number defeats that purpose.
Using `Decimal` is already a hassle, one that you even have to pass off to people consuming your API. I expect a convenient library like pydantic to not make my life harder here.

[...]

Issue #5013, opened 2023-02-06
auto-migration tool for Pydantic V2


samuelcolvin:
@dmontagu and I were just talking about making migration to V2 as easy as possible.

Some things will be complicated, but others should be mechanical, e.g. renaming methods `dict()` to `model_dict()` etc.

When I spoke to the team from sourcegraph last week, @jdorfman mentioned that they had a tool for fixing entire codebases. Would it be possible to collaborate with sourcegraph to make this available for anyone upgrading pydantic?

Alternatively, I think someone was talking on twitter about a tool to apply transforms to code, @adriangb do you remember who it was? Could it have been @Kludex?

jdorfman:
@samuelcolvin we (@beyang and I) will work on getting you a cloud account. That way you can run batch changes and save time.

Kludex:
Maybe because of https://github.com/Kludex/bump-testclient.

I'm always looking for excuses to build codemods. :sweat_smile:

Do you have the transformations that need to happen already? I can create a repo, you create an issue with the cases needed, and I can build it.

samuelcolvin:
Yes, that was it!

Let us do some digging and come up with a list of transforms. I'll reply here when things have settled a little and we have a list.

samuelcolvin:
Absolutely, we'll definitely have a thorough migration guide.

ghandic:
Perhaps the work going on in #5016 could inform this. There are a lot of scenarios we need to pick out with someone with the SME knowledge of where these objects will now reside or what the alternative is

jdorfman:
> Is there any way others could upgrade their code using your tool, or does that require a sourcegraph cloud account?

Yes, I can invite anyone you need to your org. Hopefully it isn't too late.

ghandic:
Will it be publicly available for any pydantic user to leverage?

jdorfman:
@ghandic [Batch Changes](https://docs.sourcegraph.com/batch_changes) will be only available for project maintainers i.e. anyone who has write access to the repo.

ghandic:
I think the requirement for this is for users in general, similar to the work @Kludex did for starlette to allow users to migrate with a simple command.

Give the widespread usage of pydantic there will be many users who could take advantage of the auto migration

jdorfman:
@ghandic oh, I see.

> similar to the work @Kludex did for starlette to allow users to migrate with a simple command.

Can you please give me some more context?

ghandic:
[...]

PR #5029, opened 2023-02-09
Add JSON schema support (v2)

### LINKED ISSUE #4666: V2: JSON Schema

dmontagu:
fix #4666

**Things left to resolve:**

* [ ] ~URLs~ Create an issue for this
	* [ ] Should minLength/maxLength be set on AnyUrl? (Class was moved from pydantic to pydantic_core)
	* [ ] Are we eliminating stricturl? Is the same functionality ported to pydantic_core in some manner?
		* What is the migration path?
	* **Tentative plan:** If it's possible to achieve the same functionality, drop stricturl and add notes to the migration guide

* [x] Json type
	* Is it okay to change Json[int]'s schema from {'type': 'number'} to {'type': 'string', 'format': 'json-string'}?
	* Change has been made

* [ ] ~Callable type~ Create an issue for this
	* How should we handle JSON schema generation for fields with type callable? (See, e.g., `tests.test_schema.test_callable_type`)
	* **Tentative plan:** Not sure.

* [ ] ~Decimal~ See issue #1511 and #5072
	* What should the Decimal schema be?
	* **Tentative plan:** Need to come to a consensus; https://github.com/pydantic/pydantic/issues/1511 may be relevant
		* Note: although serialization and parsing are different, I think for FastAPI/OpenAPI purposes it is important that the JSON schema be compatible with the serialization type. (Or else generated clients will produce type errors if an API returns a Decimal.) Happy to discuss this more.

* [ ] `__pydantic_modify_json_schema__`
	* [x] Do we retain backwards compatibility with the name `__modify_schema__`, or force migration to `__pydantic_modify_json_schema__`?
		* **Plan:** Backwards compatible with `__modify_schema__`; new syntax in `__pydantic_json_schema__`
		* Do signature introspection for passing kwargs
	* [ ] Do we support any args/kwargs for `__pydantic_modify_json_schema__` besides the schema itself?
		* Do we pass any info to it from FieldInfo? The CoreSchema? A "source" object? (Could be useful when working with custom generics)
		* **Tentative plan:**
		    * Because this isn't overriding anything, we can just do signature introspection for different sets of arguments
		    * Need to decide how to handle this on enums and fields. I think it's a plain classmethod on enums with similar signature to what it is for models.

* [x] `'definitions'` -> `'$defs'`
    * [x] Should we change the key `'definitions'` to `'$defs'` everywhere? This is in line with JSON schema 2020-12, but it would be nice to confirm this won't break any important tooling (e.g., openapi-generator).
        * Change has been made (see `tests.test_schema.test_schema_with_refs`; openapi doesn't actually use the definitions key anyway, and FastAPI will change out the `$defs` key as appropriate.)


dmontagu:
@samuelcolvin there's still a lot of work to be done before the JSON schema work is "finished", but I think I am starting to add too much to reasonably review. I think it would be better to stop new work on this branch, and just get everything it adds reviewed and merged before continuing.

There are already enough decisions in here that may be controversial that I think it's probably a mistake to continue building on them before establishing more thorough agreement.

I will be happy to remove things from this branch (such as my possibly-half-baked implementation of handling `discriminator`), even if it means more tests fail / etc., so that we can merge whatever fraction of it seems acceptable for now. This will make it easier to keep up-to-date, and make reviewing future improvements easier as well.

pydantic/_internal/_std_types_schema.py
```diff
@@ -66,26 +69,42 @@ def to_enum(__input_value: Any, **_kwargs: Any) -> Enum:
...
-            literal_schema,
-            to_enum,
-        )
+        return core_schema.function_after_schema(literal_schema, to_enum, metadata=metadata)
 
 
 @schema_function(Decimal)
 def decimal_schema(_schema_generator: GenerateSchema, _decimal_type: type[Decimal]) -> core_schema.FunctionSchema:
     decimal_validator = _validators.DecimalValidator()
+    metadata = build_core_metadata(update_schema=decimal_validator.__pydantic_update_schema__)
```
Kludex, line 106: Is this a good moment to mention https://github.com/pydantic/pydantic/issues/1511 ? i.e. change `float` and `int` by `str`.

pydantic/_internal/_core_metadata.py
```diff
@@ -0,0 +1,120 @@
...
+
+class CoreMetadataHandler:
+    """
+    Because the metadata field in pydantic_core is of type `Any`, we can't assume much about its contents.
+
+    This class is used to interact with the metadata field on a CoreSchema object in a consistent
+    way throughout pydantic.
+    """
+
+    def __init__(self, schema: CoreSchema | TypedDictField):
```
dmontagu, line 33: @samuelcolvin I find myself making use of the `CoreSchema | TypedDictField` in a few places (mostly related to the fact that json schema annotations can find their way onto CoreSchema _or_ TypedDictField, and often the result of producing the JSON schema for a TypedDictField is something you need to retain a JSON schema $ref to). The lack of a 'type' field on TypedDictField makes it a bit of a hassle.

It's not something we _need_ to change, but it would simplify some things if it had a unique `type` field, or even better were included in the CoreSchema union. (Maybe that wouldn't work for other reasons, but I feel like I should bring it up.)

pydantic/_internal/_core_metadata.py
```diff
@@ -0,0 +1,120 @@
...
+        elif not isinstance(metadata, dict):
+            raise ValueError(f'CoreSchema metadata should be a dict; got {metadata!r}.')
+
+    @property
+    def metadata(self) -> _CoreMetadata:
+        metadata = self.schema.get('metadata')
+        if metadata is None:
+            self.schema['metadata'] = metadata = {}
+        if not isinstance(metadata, dict):
+            raise ValueError(f'CoreSchema metadata should be a dict; got {metadata!r}.')
```
dmontagu, line 48: After playing around with the implications a bit, I felt like it was significantly cleaner to just raise errors if the CoreSchema metadata isn't at least a dict. Too many places things would get mishandled otherwise, and I think it's reasonable. I would vote that we change the type hint in `pydantic_core` from `Any` to `dict[str, Any] | None` (or even just `dict[str, Any]`, honestly, though if retaining None as a choice is more performant that's fine with me.)

pydantic/_internal/_typing_extra.py
```diff
@@ -68,6 +69,11 @@ def origin_is_union(tp: type[Any] | None) -> bool:
 
 
 NoneType = None.__class__
+if sys.version_info < (3, 10):
+    # NoneType = type(None)
```
dmontagu, line 73: I see there is `NoneType = None.__class__` above, I was thinking about changing things to handle `NoneType` the same as `EllipsisType` while I was at it since they were added at the same time/place/for the same reasons. But I guess the current definition is working? I'll drop these commented lines related to NoneType if you don't want to change it.

To be clear, I do get mypy errors if I don't use the import from `types` in python 3.11:
```
pydantic/_internal/_core_metadata.py:100: error: Variable "pydantic._internal._typing_extra.EllipsisType" is not valid as a type  [valid-type]
pydantic/_internal/_core_metadata.py:100: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

pydantic/main.py
```diff
@@ -362,18 +365,42 @@ def copy(
...
         cached = cls.__schema_cache__.get((by_alias, ref_template))
         if cached is not None:
             return cached
-        s = model_schema(cls, by_alias=by_alias, ref_template=ref_template)
+        s = schema_generator(by_alias=by_alias, ref_template=ref_template).generate(cls.__pydantic_core_schema__)
         cls.__schema_cache__[(by_alias, ref_template)] = s
         return s
 
+    @classmethod
+    def model_json_schema_misc(cls) -> JsonSchemaMisc | None:
```
dmontagu, line 386: I don't love making this a separate method from `model_json_schema`, BUT it kind of serves a different purpose, and reduces the amount of boilerplate that would be necessary to make many simple changes to the schema.

Open to renaming and/or replacing this method with a different way of achieving the same result if preferred.

pydantic/main.py
```diff
@@ -20,7 +21,8 @@
 from .errors import PydanticUserError
 from .fields import Field, FieldInfo, ModelPrivateAttr
 from .json import custom_pydantic_encoder, pydantic_encoder
-from .schema import default_ref_template, model_schema
+from .json_schema import DEFAULT_REF_TEMPLATE, GenerateJsonSchema
+from .json_schema_misc import JsonSchemaMisc
```
dmontagu, line 25: I don't like that `json_schema_misc` is a different file than `json_schema`, but I was having circular imports issues coming from the use of `JsonSchemaMisc` in the `CoreMetadataHandler`, and the `CoreMetadataHandler` in `GenerateJsonSchema`. Putting these all in the same file may have helped, but it felt wrong to put the metadata handling stuff in an otherwise-JSON-schema-specific file.

Happy to reorganize/rename files if there are any suggestions for improvement.

pydantic/_internal/_std_types_schema.py
```diff
@@ -66,26 +69,47 @@ def to_enum(__input_value: Any, **_kwargs: Any) -> Enum:
...
+        core_schema_override=literal_schema.copy(),
+        title=enum_type.__name__,
+        description=inspect.cleandoc(enum_type.__doc__ or 'An enumeration.'),
+    )
+    # TODO: Is 'enum_json_schema_misc' the best name? Is there a better way to handle this?
+    if hasattr(enum_type, 'enum_json_schema_misc'):
+        # The enum_json_schema_extra method should be a classmethod with signature:
+        #   def enum_json_schema_extra(cls, default: JsonSchemaMisc) -> JsonSchemaMisc: ...
+        # For convenience, the value generated above is passed to this function so you can use some/all of its defaults
+        json_schema_misc = getattr(enum_type, 'enum_json_schema_misc')(json_schema_misc)
```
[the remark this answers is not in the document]
dmontagu: Also, I think we could do something similar for arbitrary types, but I'm inclined not to unless there's a really good argument for it; just another thing to maintain. (I could easily be convinced to just drop the `enum_json_schema_misc` method thing entirely, but I feel it's worth putting the idea out there since I already added it.)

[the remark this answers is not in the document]
dmontagu: I realized __modify_schema__ (or whatever we rename it to) should work for this.

pydantic/main.py
dmontagu: Perhaps I should be trying to do more with `__modify_schema__`...

dmontagu:
@tiangolo @Julian

This PR implements an overhaul of the way we did JSON schema generation in v1, with an eye toward making it easier to maintain, and easier for end users to customize how their JSON schemas are generated.

This is still a work in progress, but I have put a significant amount of effort into it at this point and I think most of the structure is in place, subject to feedback.

I would very much appreciate any insights, suggestions, feedback, or requests for modification about the general approach here, or about any of the specifics.

@samuelcolvin I think you mentioned at least one other person that you suggested review this, if you could tag them here I would appreciate it.

pydantic/_internal/_generate_schema.py
```diff
@@ -75,7 +77,14 @@ def __init__(self, arbitrary_types: bool, types_namespace: dict[str, Any] | None
...
 
-    def generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
+    def generate_schema(self, obj: Any) -> core_schema.CoreSchema:
+        schema = self._generate_schema(obj)
+        modify_json_schema = getattr(obj, '__modify_schema__', None)
+        if modify_json_schema is not None:
+            CoreMetadataHandler(schema).merge_json_schema_misc(JsonSchemaMisc(modify_json_schema=modify_json_schema))
+        return schema
+
+    def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
```
dmontagu, line 87: @samuelcolvin I wrapped the old generate_schema method (now called _generate_schema) with something that reads the `__modify_schema__` off the input type and updates the metadata as appropriate, regardless of what the input was.

Let me know if you prefer this get handled differently.

dmontagu:
@Relequestual Samuel mentioned you would be a good person to ask for a review as well.

Kludex (commented):
Should https://json-schema.org/understanding-json-schema/reference/conditionals.html#id4 be included in this PR? It's likely going to be good to replace some common validators.

dmontagu:
@Kludex thanks for the review!

> Should https://json-schema.org/understanding-json-schema/reference/conditionals.html#id4 be included in this PR? It's likely going to be good to replace some common validators.

I think I'd like to get v2 working at parity with v1 first (though happy to make any changes that bring it in-line with 2020-12, including the ones you mentioned), and we could look into adding support for conditionals after that. If you have specific use cases in mind that would make use of JSON schema conditionals, I would very much appreciate it if you could open an issue with an example.

tests/test_schema.py
```diff
@@ -1677,14 +1679,14 @@ class Foo(BaseModel):
...
 
-@pytest.mark.xfail(reason='working on V2', strict=False)
 @pytest.mark.parametrize(
     'kwargs,type_,value',
     [
         ({'max_length': 5}, str, 'foo'),
         ({'min_length': 2}, str, 'foo'),
         ({'max_length': 5}, bytes, b'foo'),
-        ({'regex': '^foo$'}, str, 'foo'),
+        # TODO: Do we want to support regex keyword? It seems it should be pattern now..
```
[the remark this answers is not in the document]
dmontagu: I meant for backwards compatibility with v1 pydantic, not for compatibility with JSON schema. My question was whether, if someone has `my_match: str = Field(regex="^foo")` should that keep working, or should they be forced to change it to `my_match: str = Field(pattern="^foo")`. My possibly-misremembering understanding is we want to just force it to be changed to pattern everywhere, but I wanted to get explicit reconfirmation before I delete that comment / fix that test

pydantic/json_schema.py
```diff
@@ -0,0 +1,768 @@
...
+from typing_extensions import TypeGuard
+
+from pydantic._internal._core_metadata import CoreMetadataHandler
+from pydantic._internal._typing_extra import all_literal_values, is_namedtuple
+from pydantic._internal._utils import lenient_issubclass
+from pydantic.json import pydantic_encoder
+from pydantic.json_schema_misc import JsonSchemaValue
+from pydantic.networks import IPvAnyAddress, IPvAnyInterface, IPvAnyNetwork
+
+DEFAULT_REF_TEMPLATE = '#/definitions/{model}'
```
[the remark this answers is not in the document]
dmontagu: I'm happy to change it, but do you have a reference changing it to `defs`? All I can seem to find in [the official docs](https://json-schema.org/draft/2020-12/json-schema-core.html#name-schema-re-use-with-defs) uses `$defs` rather than just `defs`. Either way is fine with me, just want to do it "right".

tests/test_schema.py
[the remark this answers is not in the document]
Kludex: Ah, I like the "regex" on the `Field` 😅

pydantic/json_schema.py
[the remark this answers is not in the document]
dmontagu: I will probably keep this set as `definitions` for a little while to reduce the size of the diff of the tests, but I do agree we should change it to `$defs`

tests/test_schema.py
```diff
@@ -482,11 +485,14 @@ class Model(BaseModel):
...
 def test_json_schema():
     class Model(BaseModel):
-        a = b'foobar'
-        b = Decimal('12.34')
+        a: bytes = b'foobar'
+        b: Decimal = Decimal('12.34')
+
+    # TODO: What do we want the generated schema to be for Decimal? I'm thinking 'integer', 'number', _or_ 'str'
+    #     Decision: What we have in v1 is not bad enough to be worth changing
+    #     (i.e., keep it as only 'number'; maybe add a comment that other things could be okay)
```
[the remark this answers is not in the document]
samuelcolvin: See my comment on #1511, that's a different issue, but related via the discussion of whether we should allow generation of a different JSON Schema for `.model_json()` vs valiation.

pydantic/json_schema.py
```diff
@@ -0,0 +1,768 @@
...
+        values_schema.pop('title', None)  # don't give a title to the additionalProperties
+
+        json_schema: JsonSchemaValue = {'type': 'object'}
+        if values_schema:  # don't add additionalProperties if it's empty
+            json_schema['additionalProperties'] = values_schema
+
+        self.update_with_validations(json_schema, schema, self.ValidationsMapping.object)
+        return json_schema
+
+    source_class_to_schema: tuple[tuple[Any, dict[str, Any]], ...] = (
```
samuelcolvin, line 269: surely instead of doing, we should set `format` in `metadata` when creating the core-schema.

samuelcolvin (commented):
probably more points, but this is a start.

pydantic/_internal/_generate_schema.py
dmontagu: I'm not sure exactly what the concern is, but I do think the current recursion handles this properly; the following test passes:
```python
def test_modify_schema_dict_keys() -> None:
    class MyType:
        @classmethod
        def __modify_schema__(cls, schema):
            schema['test'] = 'passed'

    class MyModel(BaseModel):
        my_field: MyType

        model_config = dict(arbitrary_types_allowed=True)

    assert MyModel.model_json_schema() == {
        'properties': {'my_field': {'test': 'passed', 'title': 'My Field'}},  # <---- 'test': 'passed'
        'required': ['my_field'],
        'title': 'MyModel',
        'type': 'object',
    }
```

pydantic/json_schema_misc.py
```diff
@@ -0,0 +1,93 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
dmontagu: The main issue with that is that it was designed to be a part of the public API. I think maybe we should discuss this in more depth, perhaps a bigger refactor is merited.

pydantic/_internal/_std_types_schema.py
```diff
@@ -66,26 +69,42 @@ def to_enum(__input_value: Any, **_kwargs: Any) -> Enum:
...
-            to_enum,
-        )
+        return core_schema.function_after_schema(literal_schema, to_enum, metadata=metadata)
 
 
 @schema_function(Decimal)
 def decimal_schema(_schema_generator: GenerateSchema, _decimal_type: type[Decimal]) -> core_schema.FunctionSchema:
     decimal_validator = _validators.DecimalValidator()
+    metadata = build_metadata_dict(
+        update_core_schema=decimal_validator.__pydantic_update_schema__,
```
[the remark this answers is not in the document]
dmontagu: Yes, I'd like to go through the code and find all places where they are interacted with, and see if there can be some improvements. (Keeping this conversation unresolved as a reminder to do so.)

pydantic/_internal/_std_types_schema.py
```diff
@@ -66,26 +69,42 @@ def to_enum(__input_value: Any, **_kwargs: Any) -> Enum:
...
-            literal_schema,
-            to_enum,
-        )
+        return core_schema.function_after_schema(literal_schema, to_enum, metadata=metadata)
 
 
 @schema_function(Decimal)
 def decimal_schema(_schema_generator: GenerateSchema, _decimal_type: type[Decimal]) -> core_schema.FunctionSchema:
     decimal_validator = _validators.DecimalValidator()
+    metadata = build_metadata_dict(
```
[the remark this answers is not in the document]
dmontagu: The main issue with this is that there are already some things in the metadata that don't really feel like they belong with the JsonSchema stuff (namely -- the update_core_schema_function stuff).

But we can probably come up with something cleaner either way; at the very least I could create a method on JsonSchemaMisc that does produce a metadata dict if you don't need the other field

pydantic/json_schema.py
```diff
@@ -0,0 +1,768 @@
...
+
+
+class InvalidForJsonSchema(ValueError):
+    pass
+
+
+# There are three types of references relevant to building JSON schemas:
+#   1. core_schema "ref" values; these are not exposed as part of the JSON schema
+#       * these might look like the fully qualified path of a model, its id, or something similar
+CoreRef = NewType('CoreRef', str)
```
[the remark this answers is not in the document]
dmontagu: I think these are both reasonable proposals, though I'm also okay leaving as-is. Let's discuss this point soon

pydantic/json_schema.py
```diff
@@ -0,0 +1,768 @@
...
+        remaining_json_refs = self.get_json_refs(json_schema)
+        all_json_refs = list(self.json_to_defs_refs.keys())
+        for k in all_json_refs:
+            if k not in remaining_json_refs:
+                del self.definitions[self.json_to_defs_refs[k]]
+
+        if self.definitions:
+            json_schema['definitions'] = self.definitions
+
+        return json_schema
```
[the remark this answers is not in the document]
dmontagu: I will add _used

pydantic/json_schema.py
dmontagu: that... seems like an improvement :)

pydantic/json_schema.py
```diff
@@ -0,0 +1,768 @@
...
+    def __init__(self, by_alias: bool = True, ref_template: str = DEFAULT_REF_TEMPLATE):
+        self.by_alias = by_alias
+        self.ref_template = ref_template
+
+        self.core_to_json_refs: dict[CoreRef, JsonRef] = {}
+        self.core_to_defs_refs: dict[CoreRef, DefsRef] = {}
+        self.defs_to_core_refs: dict[DefsRef, CoreRef] = {}
+        self.json_to_defs_refs: dict[JsonRef, DefsRef] = {}
+
+        self.definitions: dict[DefsRef, JsonSchemaValue] = {}
```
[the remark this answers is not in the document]
dmontagu: * Regarding public vs. private: I kept everything public to convey the intent that it was okay for users to interact with them when overriding. I'm not sure if that implication is there, and even if it is, whether it's really a great idea to keep _everything_ public like this. But that's why I did it. Let me know if you prefer private despite this line of thinking.

* Regarding `__slots__` -- would that make it annoying for users overriding the class if they wanted to add some new attributes? I don't have a good sense of how `__slots__` works with inheritance.

pydantic/json_schema_misc.py
```diff
@@ -0,0 +1,93 @@
...
+from pydantic_core import CoreSchema
+
+from pydantic._internal._utils import dict_not_none
+
+JsonSchemaValue = Dict[str, Any]
+JsonValue = Dict[str, Any]
+
+
+@dataclass
+class JsonSchemaMisc:
```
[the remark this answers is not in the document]
dmontagu: I think that name is an improvement, and I'll change it. But let's also discuss whether there might be a better way to handle things that eliminates this class entirely. (I'm not sure there is, but I do find this class's existence kind of ugly.)

pydantic/json_schema_misc.py
```diff
@@ -0,0 +1,93 @@
...
+
+from dataclasses import asdict, dataclass, replace
+from typing import Any, Callable, Dict
+
+from pydantic_core import CoreSchema
+
+from pydantic._internal._utils import dict_not_none
+
+JsonSchemaValue = Dict[str, Any]
+JsonValue = Dict[str, Any]
```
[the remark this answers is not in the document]
dmontagu: I don't think that alias ended up being very useful anyway so I'm just going to delete it (it was only used in one place)

dmontagu:
For anyone following (or especially planning to review) this PR: I've updated the description at the top with the set of things I'm looking for help making decisions about.

tiangolo (approved):
First of all, so exciting to see you back around here @dmontagu! 🤩🎉

## PR Review

Now to the PR review:

I've been reviewing this for the last couple of days. Very clever all around. Amazing job, as always! 🚀

I added a couple of comments here and there. A couple of opinions in the places where I suspect my opinion could be relevant.

## Type Checks from VS Code - PyRight

I noticed a couple of errors with VS Code/PyRight in some places, I think mainly about getting a dict key form a dict where it is not required, or things like that.

Would it be valuable to point them out and suggest the fixes or is it just too much noise on top of the actual PR review?

I'm just adding a couple of those comments for now.

## Questions

About some of the questions in the description, my opinions inline.

Pretty much in favor with everything, for the rest I don't have a strong opinion or enough knowledge.

**Things left to resolve:**

* [x] Field kwargs
	* [x] Should we remove or keep the 'examples' kwarg on `Field`?
		* If not, what is the migration path?
		* **Tentative plan:** Re-add the examples keyword argument
			* [@tiangolo]: In favor with plan. All things equal, +1 to re-adding it. Unless there's something against it.
	* [x] Should we remove or keep the 'const' kwarg on `Field`? (This was used to indicate a field was required and the only allowed value was its default.)
		* If not, what is the migration path?
		* **Tentative plan:** Drop this argument in favor of `Literal` and/or validators as appropriate (i.e., keep as is on main branch)
			* [@tiangolo]: In favor with plan. If it's the same, I would think it's better to keep one single way to do it, and if there's now a "standard" way to do it with type annotations, I would use that. One single way reduces combinatorial logic, maintenance, testing. And it's easier for users to learn one single thing than two, even more if it's two competing ones.
	* [x] Should we change the `regex` kwarg to `pattern` in the `Field` signature?
		* **Tentative plan:** Change `regex` to `pattern` (i.e., keep as is on main branch)
			* [@tiangolo]: In favor with plan. As JSON Schema uses the same `pattern` keyword, I would vote for that. `regex` might be more intuitive from Python-land, but for me, `regex` is more ambiguous as to if it takes a regular expression object or a pattern string. Even though the Python docs also refer to a Pattern object, but anyway, I personally associate the word "pattern" more easily with the string.
	* [x] Do we want to do a backwards-compatibility handler thing with `**kwargs` — warning about now-invalid kwargs (whether or not we keep them working) on `Field`?
		* **Tentative plan:** Do not attempt to maintain backwards compatibility, but document in the migration guide
			* [@tiangolo] **Strongly** in favor with plan. More details below, and also inline on the PR.


* [x] ConfigDict fields
	* [x] Do we want to re-add 'schema_extra' as a field on model_config?
		* If not, should we drop the tests?
		* Either way, do we retain/drop support on a config _class_ in the model definition?
		* What is the migration path?
		* **Tentative plan:** Drop in favor of `__pydantic_modify_json_schema__`
			* [@tiangolo] In favor with plan.
	* [x] Do we want to re-add 'fields' as a field on model_config?
		* If not, should we drop the tests?
		* What is the migration path?
		* **Tentative plan:** Drop in favor of annotations and/or `= Field(...)`
			* [@tiangolo] In favor with plan.

* [ ] Default values
	* If a default value of a field is not JSON encodable, do we include anything in the schema indicating the default value? Do we error?
	* **Tentative plan:** Just don't include the value in the schema if json encoding fails, but keep the `required` field set appropriately. I believe this is perfectly compatible with JSON schema.
		* [@tiangolo] Soft in favor with plan. I'm not sure about the implications, but I think it should be fine.

* [ ] `__modify_schema__` (-> `__pydantic_modify_json_schema__`)
	* Do we retain backwards compatibility with the name `__modify_schema__`, or force migration to `__pydantic_modify_json_schema__`?
		* **Tentative plan:** Force the migration to `__pydantic_modify_json_schema__`
			* [@tiangolo] In favor with plan. This is the moment to break changes.
	* Do we support any args/kwargs for `__pydantic_modify_json_schema__` besides the schema itself?
		* Do we pass any info to it from FieldInfo? The CoreSchema? A "source" object? (Could be useful when working with custom generics)
		* **Tentative plan:** Not sure. Would potentially be useful to have access to the core schema and to the type "source" (e.g., to check generic parameter values, etc.)

* [x] FastAPI
	* Do we need to retain functions analogous to any of the following? (Note we have retained `model_schema`, and `schema`.)
		* get_flat_models_from_field/get_flat_models_from_fields
		* get_flat_models_from_model/get_flat_models_from_models
		* get_model_name_map
		* model_process_schema
	* If not, can we drop their tests?
	* **Tentative plan:** Could we instead build a CoreSchema in FastAPI that produces the desired JSON schema?
		* I'm inclined to put significant effort into helping with the migration of FastAPI to pydantic v2, and to implement any necessary analogous functions as a part of that effort. But if it's already clear that we will need them, happy to add them now.
			* [@tiangolo] In favor with plan. And thanks, as always! 🎉 Better not to add premature optimizations/abstractions. And more details in line in the PR.

* [ ] Callable type
	* How should we handle JSON schema generation for fields with type callable? (See, e.g., `tests.test_schema.test_callable_type`)
	* **Tentative plan:** Not sure.

* [ ] Decimal
	* What should the Decimal schema be?
	* **Tentative plan:** Need to come to a consensus; https://github.com/pydantic/pydantic/issues/1511 may be relevant
		* Note: although serialization and parsing are different, I think for FastAPI/OpenAPI purposes it is important that the JSON schema be compatible with the serialization type. (Or else generated clients will produce type errors if an API returns a Decimal.) Happy to discuss this more.
			* [@tiangolo] agreed that incompatible clients would be a non-obvious but important problem. Although I'm not sure I understand the problem well and what's the "serialization type" in this context.

## kwargs in `Field()`

Details about asking for no more kwargs in `Field()`:

I would request to **not** accept any extra kwargs. I asked for that here in this hidden comment: https://github.com/pydantic/pydantic/pull/4516#discussion_r1010870226

Copying here for context/completeness:

> This `**extra` has been used to add stuff to JSON Schema.

> But in some cases, people pass other stuff here expecting it to be used later or for other purposes, but then it also affects JSON Schema.

> I would like to ask to make it a single keyword argument `extra` instead of `**extra`, or alternatively, to have another single argument `json_schema_extra` so that extensions for JSON Schema can be apart from extra values passed here for other purposes.

> I haven't seen a use for this apart from JSON Schema, and the fact that it makes `Field()` accept any keyword arguments makes editors not complain about arguments that actually do nothing. I've seen this sometimes, e.g. recently with a `Field(required=True)`, but `required` is not a valid argument, if it's required or not is defined by the absence of a `default`. But the code seemed fine and linters passed, because it was taken by the catch all `**extra`.

> It's a breaking change, but I think it's totally worth it and this is the right time for those changes.

## Semi-off topic, highlights

**Note**: no action item here, just praise and highlights of non-obvious extra benefits. 🎉

Very clever implementation! 🚀

I like how the core schema refs, def refs, and json refs are stored and correlated, and then updated when necessary, but still stored. And how everything is stored in the object.

I think this will allow me to do something I've been wanting to do in FastAPI, to pre-generate the JSON Schema stuff for each endpoint, and when there's an error, to show the specific endpoint with an error instead of just saying "something is wrong".

## Wishlist - collect errors

I was checking the wishlist of things I wanted for Pydantic and almost everything is already done. 🎉

From the things not there yet, there's one single thing I would like to ask that is related to JSON Schema, although not sure if it could be a subsequent PR. It's also related to the point above.

### The problem

First the problem.

Currently, an example like this, rightfully breaks:

```Pyhton


class Car:
    def __init__(self, make: str, model: str, year: int):
        self.make = make
        self.model = model
        self.year = year

class Model(BaseModel):
    f1: int = 1
    f2: Car

    model_config = ConfigDict(arbitrary_types_allowed=True)


Model.model_json_schema()
```

But if `Model` is in a FastAPI path operation, and maybe `Car` is in a sub-model in a sub-field, in a large app, when calling the OpenAPI endpoint the whole thing will break, saying there's a problem with `Car`, but without saying where exactly... and more importantly, none of the OpenAPI will be generated. So, a single broken field breaks the whole OpenAPI and Swagger UI.

### Feature Request

Currently it immediately raises an error like:

```
Cannot generate a JsonSchema for X
```

...that stops the rest of the JSON Schema generation.

I would like instead to accumulate those errors and get them all together, but to be able to retrieve the rest of the JSON Schema.

### Possible Implementation

I think it's better to state the problem in case you see a better way to solve it, but I can also imagine possible implementations. For example:

```Python
schema = Model.model_json_schema()
```

...raises as normally, but:

```Python
schema, errors = Model.model_json_schema(keep_errors=True)
```

Doesn't raise, and returns a JSON Schema where anything with an error is replaced with `{}` and `errors` is a list of stored exceptions. Of course, `keep_errors` could use some bikeshedding on the parameter name, the same for the return interface, but you get the idea.

It could also be a context manager, maybe too complex, but that would keep the API interface (params, return) of `model_json_schema()` unmodified for both cases, don't know. And it would probably only be used by FastAPI and libraries like that, so not too bad on that side. But it would require contextvars and all that dance. E.g.

```Python
with ErrorCollector as c:
	schema = Model.model_json_schema()
print(e.errors)
```

Again, the specific implementation is up for discussion, but I would like to have a way to do this.

### Plan in FastAPI

My plan is also later to put this in FastAPI in a way that (if I manage to) when there's an error like this, the traceback would use Rich and point to the right file and line with the specific FastAPI *path operation* that uses that model. But I think that to be able to do that I first need to pre-compute and collect the JSON Schema errors per path-operation, and not only at the global level.

## Breaking Changes

About the breaking changes, mostly renaming things, in the small few places this happens, I think this is the right moment to do it and discard the cruft of backwards compatibility for those things.

An improvement in that migration, alternative to keep backwards compatibility, would be a tool to automatically update code to use the new things, for the simple ones (I can imagine, e.g. moving kwargs in `Field()` to `json_schema_extra`, or changing `regex` to `pattern`). Maybe that could also work for the `model_config`.

I haven't done it, but I know it's possible and feasible with LibCST. @Kludex has built a couple of CLI tools that do that type of thing, e.g. https://github.com/Kludex/bump-testclient

Julian:
Hey! Nice work -- I'm happy to try to review this -- though I think it's probably JSON Schema-specific questions that I'm best suited to opine on rather than pydantic-internal specific ones or otherwise implementation-specific ones here -- if there are any outstanding things that seem more like the former than the latter feel free to ping me on them specifically of course. Otherwise I can try to wholesale review this but I probably am missing lots of context on pydantic's current implementation so I'll probably find it hard to have opinions (or at least intelligent ones!) on how to best manage the transition.

pydantic/_internal/_generate_schema.py
```diff
@@ -189,15 +226,17 @@ def generate_field_schema(
         """
         assert field.annotation is not None, 'field.annotation should not be None when generating a schema'
         schema = self.generate_schema(field.annotation)
-        schema = apply_metadata(schema, field.metadata)
+        schema = apply_annotations(schema, field.metadata)
```
yezz123, line 229: We can extract this into a method would be fine I guess 🤔

pydantic/_internal/_generate_schema.py
```diff
@@ -303,7 +350,7 @@ def _generate_parameter_schema(
         field = FieldInfo.from_annotation(annotation)
         assert field.annotation is not None, 'field.annotation should not be None when generating a schema'
         schema = self.generate_schema(field.annotation)
-        schema = apply_metadata(schema, field.metadata)
+        schema = apply_annotations(schema, field.metadata)
```
yezz123, line 353: +1

yezz123 (commented):
Here is my sense at that case we can extract this in one method would be fine to use it I guess

```py
        schema = self._generate_field_definition(field)
```

and the method:

```py
    def _generate_field_definition(self, field):
        """
        Generate a schema for a field in a namedtuple, dataclass or function signature.
        """
        assert (
            field.annotation is not None
        ), 'field.annotation should not be None when generating a schema'
        result = self.generate_schema(field.annotation)
        result = apply_annotations(result, field.metadata)
        return result
```

What you think @dmontagu

pydantic/json_schema.py
```diff
@@ -0,0 +1,970 @@
...
+                    #   Maybe tell users to override this method if they want custom behavior here?
+                    #       (If populate by name is false)
+                    pass
+            field_json_schema = self.generate_inner(field).copy()
+            if 'title' not in field_json_schema and self.field_title_should_be_set(field):
+                title = self.get_title_from_name(name)
+                field_json_schema['title'] = title
+            field_json_schema = self.handle_ref_overrides(field_json_schema)
+            properties[name] = field_json_schema
+            if field['required']:
```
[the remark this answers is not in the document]
dmontagu: Thanks for pointing these out. They are real errors and I've been burned in other parts of the code by them :D. I figured out how to run pyright locally and will at least try to address any pyright-detected issues in the code of this PR.

Looks like there's a lot of (mostly-spurious?) things it is erroring about in other files, I'd be in favor of addressing those issues and possibly even integrating pyright into CI once v2 is released (if not before...)

pydantic/json_schema.py
```diff
@@ -0,0 +1,970 @@
...
+                if ref == old_json_ref:
+                    item['$ref'] = new_json_ref
+                return {k: walk_replace_json_schema_ref(v) for k, v in item.items()}
+            else:
+                return item
+
+        return walk_replace_json_schema_ref(json_schema)
+
+    def get_cache_defs_ref_schema(
+        self, core_ref: CoreRef, schema: CoreSchema | TypedDictField
```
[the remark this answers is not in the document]
dmontagu: Thanks for pointing this out.

I'm going to blame PyCharm for this one — it's largely incapable of functioning properly while working on pydantic (something related to inefficiencies in handling recursive type definitions, and how much pydantic_core uses such things), and I have clearly become dependent on it showing me when arguments are unused 😅. Originally I had this function actually storing the schema in the definitions, but clearly it doesn't still do that.

(For what it's worth, JetBrains is aware of the issue and have told us there should be a fix in the next major release.)

tests/test_schema.py
```diff
@@ -1212,81 +1216,88 @@ def create_testing_submodules():
...
+#     flat_models = get_flat_models_from_models([Baz, ModelA, ModelB, ModelC, ModelD])
+#     model_name_map = get_model_name_map(flat_models)
+#     assert model_name_map == {
+#         Foo: 'Foo',
+#         Bar: 'Bar',
+#         Baz: 'Baz',
+#         ModelA: 'pydantic_schema_test__modulea__modela__Model',
+#         ModelB: 'pydantic_schema_test__moduleb__modelb__Model',
+#         ModelC: 'pydantic_schema_test__modulec__modelc__Model',
+#     }
```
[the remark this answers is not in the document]
dmontagu: > ...and now I don't think there's any risk/fear that you would decide not to support features needed in Pydantic to integrate it with FastAPI properly. 

😅

tests/test_schema.py
```diff
@@ -2042,6 +2034,8 @@ class Model(BaseModel):
 @pytest.mark.xfail(reason='working on V2')
 def test_model_with_schema_extra():
     class Model(BaseModel):
+        # TODO: Do we want to keep schema_extra as a field in ConfigDict?
+        #   If not, should I delete this test?
         model_config = ConfigDict(schema_extra={'examples': [{'a': 'Foo'}]})
```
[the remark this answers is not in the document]
dmontagu: > To be honest, I don't remember this model_config, I guess this is equivalent to the previous inner class Config but with better typing support, etc. right?

Not sure if the typing support is better (😅), but yes, it is replacing the inner class Config.

There are new ways to update the schema, but I think they (at least as of the writing of this comment) involve a bit more ceremony than this (overriding methods, creating types, etc.).

In general I think I'm in favor of reducing the amount of stuff in model_config where appropriate, and this seems like a good opportunity; I think the main change I'll make is changing the JsonSchemaExtra dataclass I have added to just be a TypedDict. While I liked having the methods, I think it will be fine to move them elsewhere and I think using a TypedDict will make it easier for users to pick up (and with less boilerplate).

dmontagu:
@tiangolo what do you think of this? https://github.com/dmontagu/pydantic/pull/1

The basic idea is to just return errors "in place" instead of raising them, so that they can then be collected by a downstream function.

I think it should make some minor changes (possibly just changing JsonSchemaValue to `dict[Any, Any]`..) so I don't have to lie to the type system about the `_ErrorKey`, but it's a pretty minimal code change and I think powers exactly what you are looking for. (You can see a possible implementation of what I imagine the FastAPI error collector might look like in the test I added.)

pydantic/json_schema.py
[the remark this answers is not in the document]
tiangolo: Great! :rocket: 

Sometimes PyRight can get too noisy and in some cases spurious, it might be too annoying to have it in CI if there are still too many false positives, but if there aren't too many problems, I think that's a great idea.

pydantic/json_schema.py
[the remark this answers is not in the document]
tiangolo: He, I know that type of problem, here I noticed this just because VS Code grayed it out. :sweat_smile:

tiangolo:
@dmontagu ah, very clever!

And good point that subclassing `GenerateJsonSchema` might work (from the comment in the other PR). Although it's true that doing it in FastAPI would not allow users to intuitively subclass it themselves.

I think this idea of including errors in the schema with a flag should work. My only fear is including two types of things in the same object (schema and errors), but at the same time, this would only happen when enabling the flag, so whoever does that (me) would have to know what they are doing and know what to expect, so I think that might be enough.

And this is probably the simplest solution, not affecting anyone else nor changing much the return type annotations.

---

I was also thinking about an alternative implementation separating the errors and typing it with `@overload`, but thinking about it, your solution of including it in the schema would naturally preserve the exact spot of the error without needing any other tricks. So, the more I think about it, the more I get convinced your idea is better than what I was thinking originally. 🧠

---

Not sure if you would rather keep the conversation here in a central place, or there in the other PR to avoid my extra noise here, let me know if you woud prefer to switch over there!

tests/test_schema.py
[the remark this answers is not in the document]
samuelcolvin: I think we should just support `__pydantic_modify_json_schema__` (or whatever it end su being called) on a model, then anything else should only be supported for backwards compatibility.

samuelcolvin:
## `Field` - `example`, `const`, `regex` 

I agree we should move to remove/change them but provide backwards compatibility  with a deprecation warning, I guess we should also allow `**kwargs` and raise a warning for that too - I guess we should we add those `**kwargs` to JSON Schema for backwards compatibility

## `Config.schema_extra` 

I guess we should support it, with a warning

## `default` values 

I guess we should use the same logic as for fields with a type that can't be defined in JSON - e.g. `Callable` and `IsSubClass` - I don't really like `UserWarning`, but since we use it already I'm happy to keep it.

**Update:** @dmontagu I don't see this code, did we decide to remove it?

## `__pydantic_modify_json_schema__` 

I'm actually inclined here to change the behaviour and name to `def __pydantic_json_schema__(cls, schema: Dict[str, Any], info: Info) -> Dict[str, Any]`, e.g. remove the slightly odd "modify" behaviour to a much more obvious "take the value, return the new value" signature that matches what we're doing with `pydantic_core`.

I know this is another change, however:
* the method name has changed, we can continue to support the current behaviour with `__modify_schema__` albeit with a warning
* we can raise a warning or error if `__pydantic_modify_json_schema__` "returns None" and hence avoid mistakes

## `Decimal`

My preference would be to add a config setting for JSON Schema generation, something like `mode: Literal['validation', 'serialisation']` or `'input' | 'output'` which indicates whether we're build a JSON schema for what's required for `MyModel(**data)` vs. `my_model.model_json()` etc.

Then the JSON Schema for validation should be `number`, but should be `string` for serialisation where we're going to default
to returning a string as per #1511.

## @tiangolo errors feature request

My solution would be this:

* always collect the errors/warnings on the instance of `GenerateJsonSchema`, use a method on `GenerateJsonSchema` to create the error/warning, and another method to construct the JSON Schema value to use in these cases (both so they can be customised fairly easily)
* Change the default code to generate, then warn/raise an error if the errors exist, this would allow FastAPI to do something different

So

```py
s = schema_generator(by_alias=by_alias, ref_template=ref_template).generate(cls.__pydantic_core_schema__)
```

Becomes (by default)

```py
schema_generator = schema_generator_cls(by_alias=by_alias, ref_template=ref_template)
s = s.generate(cls.__pydantic_core_schema__)
if s.errors:
    raise PydanticInvalidForJsonSchema(...)
```

Then FastAPI have have it's own logic to generate JSON Schema with very little duplication.

## @tiangolo migration tool

see #5013

pydantic/json_schema.py
```diff
@@ -0,0 +1,1081 @@
...
+    ref_template: str = DEFAULT_REF_TEMPLATE,
+    schema_generator: type[GenerateJsonSchema] = GenerateJsonSchema,
+) -> dict[str, Any]:
+    model = get_model(model)
+    return model.model_json_schema(by_alias=by_alias, ref_template=ref_template, schema_generator=schema_generator)
+
+
+# TODO: It might be a good idea to move these functions to a separate file, whether in pydantic_core or not
+#   Alternatively, might be good to make them "public"
+def _is_typed_dict_field(schema: CoreSchema | TypedDictField) -> TypeGuard[TypedDictField]:
```
samuelcolvin, line 1076: can we move these into `_internal`?

pydantic/schema.py
```diff
@@ -277,7 +277,7 @@ def get_field_schema_validations(field: ModelField) -> Dict[str, Any]:
         f_schema['const'] = field.default
     if field.field_info.extra:
         f_schema.update(field.field_info.extra)
-    modify_schema = getattr(field.outer_type_, '__modify_schema__', None)
```
samuelcolvin, line 280: can we delete this file? Would make me so happy.

pydantic/json_schema.py
```diff
@@ -0,0 +1,1081 @@
...
+if TYPE_CHECKING:
+    from .dataclasses import Dataclass
+    from .main import BaseModel
+
+JsonSchemaValue = Dict[str, Any]
+
+# ##### JSON Schema Metadata Manipulation #####
+# Keys missing in a JsonSchemaMetadata should be treated the same as they would be if present with a value of None.
+# This is important to ensure that it is possible to "remove" an override by setting it to None.
+JsonSchemaMetadata = TypedDict(
```
samuelcolvin, line 45: is there a reason this can't be defined using the class syntax for `TypedDict`?

pydantic/json_schema.py
```diff
@@ -0,0 +1,1081 @@
...
+        # A final function to apply to the JSON schema after all other modifications have been applied
+        # If you want to force specific contents in the generated schema, you can use a function that ignores the
+        # input value and just return the schema you want.
+        'modify_js_function': Optional[Callable[[JsonSchemaValue], None]],
+    },
+    total=False,
+)
+
+
+def merge_js_metadata(
```
samuelcolvin, line 76: boring I know, but maybe we should be consistent about calling `json_schema`, `js` internally?

But not that bothered if you disagree.

samuelcolvin (approved):
I think this is looking awesome.

I'm in favour of merging it asap and creating separate issues for outstanding problems & discussions.

pydantic/json_schema.py
```diff
@@ -0,0 +1,1081 @@
...
+    we need to remove it from the object so that the handler doesn't recurse infinitely.
+    """
+    result = json_schema_metadata.copy()
+    result.pop('core_schema_override', None)
+    return result
+
+
+_STANDARD_FIELDS: list[
+    Literal['title', 'description', 'examples', 'deprecated', 'readOnly', 'writeOnly', '$comment']
+] = ['title', 'description', 'examples', 'deprecated', 'readOnly', 'writeOnly', '$comment']
```
samuelcolvin, line 109:
```suggestion
_STANDARD_FIELDS_TYPES = Literal['title', 'description', 'examples', 'deprecated', 'readOnly', 'writeOnly', '$comment']
_STANDARD_FIELDS: list[_STANDARD_FIELDS_TYPES] = list(_STANDARD_FIELDS_TYPES.__args__)
```

???

dmontagu: You get a mypy error trying to access `__args__` but I think we can use `_STANDARD_FIELDS = all_literal_values(_STANDARD_FIELD_TYPES)`. I'll make the change

pydantic/json_schema.py
dmontagu: I think there are cases where we should use `json_schema` or `JsonSchema` in full form for clarity; I'm most inclined to shorten it in private APIs where someone would be more likely to know that `js` is short for `json_schema`. I'll take a look though everywhere we use it in full form and make sure I think it's good.

pydantic/json_schema.py
dmontagu: The '$comment' key. We can drop it, but doing it like this allowed me to skip some key-remapping logic. If there is a way to use the class syntax with a different key name I would prefer that (like alias for pydantic model fields), I just couldn't find such a thing.

pydantic/schema.py
dmontagu: Lol yes I'll do that

pydantic/json_schema.py
samuelcolvin: agreed, ignore this one.

pydantic/json_schema.py
samuelcolvin: ok.

pydantic/json_schema.py
samuelcolvin: okay, keep.

pydantic/json_schema.py
dmontagu: Yeah, do you prefer them in pydantic instead of pydantic_core? I'll put it in an `_internal` module if in pydantic, just making sure.

tiangolo:
Thanks for the replies to all the points @samuelcolvin!

All looks good, and agreed it makes sense in subsequent PRs.

## `Field()` with `**kwargs`

About `Field()` supporting `**kwargs`, I would like a way to make mypy/editors complain about them, so that developers can avoid using things that are not supported while writing the code and not only at runtime with a warning.

I was playing around and found a way to achieve both, I think. This works at runtime (with a deprecation warning), but shows an error in editors. It's a bit of extra code duplication because of the overload, but I guess it's the cost of backwards compatibility. 😅

<details>
	<summary>Field() implementation</summary>

```Python
@overload
def Field(
    default: Any = Undefined,
    *,
    alias: str = None,
    title: str = None,
    description: str = None,
    exclude: typing.AbstractSet[int | str] | typing.Mapping[int | str, Any] | Any = None,
    include: typing.AbstractSet[int | str] | typing.Mapping[int | str, Any] | Any = None,
    gt: float = None,
    ge: float = None,
    lt: float = None,
    le: float = None,
    multiple_of: float = None,
    allow_inf_nan: bool = None,
    max_digits: int = None,
    decimal_places: int = None,
    min_items: int = None,
    max_items: int = None,
    min_length: int = None,
    max_length: int = None,
    frozen: bool = None,
    pattern: str = None,
    discriminator: str = None,
    repr: bool = True,
    strict: bool | None = None,
    json_schema_extra: dict[str, Any] | None = None,
) -> Any:
    ...

def Field(
    default: Any = Undefined,
    *,
    alias: str = None,
    title: str = None,
    description: str = None,
    examples: list[Any] = None,
    exclude: typing.AbstractSet[int | str] | typing.Mapping[int | str, Any] | Any = None,
    include: typing.AbstractSet[int | str] | typing.Mapping[int | str, Any] | Any = None,
    gt: float = None,
    ge: float = None,
    lt: float = None,
    le: float = None,
    multiple_of: float = None,
    allow_inf_nan: bool = None,
    max_digits: int = None,
    decimal_places: int = None,
    min_items: int = None,
    max_items: int = None,
    min_length: int = None,
    max_length: int = None,
    frozen: bool = None,
    pattern: str = None,
    discriminator: str = None,
    repr: bool = True,
    strict: bool | None = None,
    json_schema_extra: dict[str, Any] | None = None,
    **kwargs: Any,
) -> Any:
    """
    Used to provide extra information about a field, either for the model schema or complex validation. Some arguments
    apply only to number fields (``int``, ``float``, ``Decimal``) and some apply only to ``str``.

    :param default: since this is replacing the field's default, its first argument is used
      to set the default, use ellipsis (``...``) to indicate the field is required
    :param default_factory: callable that will be called when a default value is needed for this field
      If both `default` and `default_factory` are set, an error is raised.
    :param alias: the public name of the field
    :param title: can be any string, used in the schema
    :param description: can be any string, used in the schema
    :param examples: can be any list of json-encodable data, used in the schema
    :param exclude: exclude this field while dumping.
      Takes same values as the ``include`` and ``exclude`` arguments on the ``.dict`` method.
    :param include: include this field while dumping.
      Takes same values as the ``include`` and ``exclude`` arguments on the ``.dict`` method.
    :param gt: only applies to numbers, requires the field to be "greater than". The schema
      will have an ``exclusiveMinimum`` validation keyword
    :param ge: only applies to numbers, requires the field to be "greater than or equal to". The
      schema will have a ``minimum`` validation keyword
    :param lt: only applies to numbers, requires the field to be "less than". The schema
      will have an ``exclusiveMaximum`` validation keyword
    :param le: only applies to numbers, requires the field to be "less than or equal to". The
      schema will have a ``maximum`` validation keyword
    :param multiple_of: only applies to numbers, requires the field to be "a multiple of". The
      schema will have a ``multipleOf`` validation keyword
    :param allow_inf_nan: only applies to numbers, allows the field to be NaN or infinity (+inf or -inf),
        which is a valid Python float. Default True, set to False for compatibility with JSON.
    :param max_digits: only applies to Decimals, requires the field to have a maximum number
      of digits within the decimal. It does not include a zero before the decimal point or trailing decimal zeroes.
    :param decimal_places: only applies to Decimals, requires the field to have at most a number of decimal places
      allowed. It does not include trailing decimal zeroes.
    :param min_items: only applies to lists, requires the field to have a minimum number of
      elements. The schema will have a ``minItems`` validation keyword
    :param max_items: only applies to lists, requires the field to have a maximum number of
      elements. The schema will have a ``maxItems`` validation keyword
    :param min_length: only applies to strings, requires the field to have a minimum length. The
      schema will have a ``minLength`` validation keyword
    :param max_length: only applies to strings, requires the field to have a maximum length. The
      schema will have a ``maxLength`` validation keyword
    :param frozen: a boolean which defaults to True. When False, the field raises a TypeError if the field is
      assigned on an instance.  The BaseModel Config must set validate_assignment to True
    :param pattern: only applies to strings, requires the field match against a regular expression
      pattern string. The schema will have a ``pattern`` validation keyword
    :param discriminator: only useful with a (discriminated a.k.a. tagged) `Union` of sub models with a common field.
      The `discriminator` is the name of this common field to shorten validation and improve generated schema
    :param repr: show this field in the representation
    :param json_schema_extra: extra dict to be merged with the JSON Schema for this field
    :param strict: enable or disable strict parsing mode
    """
    current_json_schema_extra: dict[str, Any] | None = None
    if kwargs:
        print("what")
        warnings.warn(
            'Arbitrary Field keywords (**kwargs) have been deprecated, to extend the '
            'generated JSON Schema use the new dict parameter json_schema_extra, the '
            f'invalid keyword arguments are: {kwargs}',
            DeprecationWarning,
        )
        current_json_schema_extra = kwargs.copy()
    if current_json_schema_extra and json_schema_extra:
        current_json_schema_extra.update(json_schema_extra)
    else:
        current_json_schema_extra = json_schema_extra
    return FieldInfo.from_field(
        default,
        default_factory=default_factory,
        alias=alias,
        title=title,
        description=description,
        examples=examples,
        exclude=exclude,
        include=include,
        gt=gt,
        ge=ge,
        lt=lt,
        le=le,
        multiple_of=multiple_of,
        allow_inf_nan=allow_inf_nan,
        max_digits=max_digits,
        decimal_places=decimal_places,
        min_items=min_items,
        max_items=max_items,
        min_length=min_length,
        max_length=max_length,
        frozen=frozen,
        pattern=pattern,
        discriminator=discriminator,
        repr=repr,
        json_schema_extra=current_json_schema_extra,
        strict=strict,
    )
```

</details>

## errors feature request

I think that should work too, and it could probably be another PR on top. The main important thing is that the current approach doesn't make that impossible to achieve/add later. 🎉

dmontagu:
@samuelcolvin 

> My preference would be to add a config setting for JSON Schema generation, something like mode: Literal['validation', 'serialisation'] or 'input' | 'output' which indicates whether we're build a JSON schema for what's required for MyModel(**data) vs. my_model.model_json() etc.
>
> Then the JSON Schema for validation should be number, but should be string for serialisation where we're going to default

There is an important consideration here that I discuss in https://github.com/pydantic/pydantic/issues/5072, which is that when generating clients based on an OpenAPI spec (one of the biggest benefits of FastAPI imo, and one of the main reasons I have used it for years), models are frequently used both as _inputs_ (so will be "validated"), and as _outputs_ (so will be "serialized"). And unless you make two separate models (which comes with its own issues), they will share a schema. So it's not uncommon that you'll have to have one schema for the both inputs and outputs, and therefore need to resolve this.

And even if we _were_ okay with creating two separate schemas for the model based on whether it is an "input" or an "output" of the API — and I am not sure we should be — there's still the issue that we'd probably want a _single_ generator instance to produce the "output" format in some places and the "input" format in others, within the same schema (at least that's the closest to how FastAPI does it today, I think). So I'm thinking it may make more sense to somehow make it an annotation on the schema.

I'm not sure, but I'd be inclined to postpone addressing this in this PR and hopefully have some discussion in #5072

dmontagu:
> I was playing around and found a way to achieve both, I think.
@tiangolo Yeah this was my plan for how to do this. We can remove the overload down the line after people have had time to migrate, and it will still cause mypy/IDE errors for them now.

👍

dmontagu:
@samuelcolvin 

from your suggestion to @tiangolo's feature request:
```python
schema_generator = schema_generator_cls(by_alias=by_alias, ref_template=ref_template)
s = s.generate(cls.__pydantic_core_schema__)
if s.errors:
    raise PydanticInvalidForJsonSchema(...)
```

> Then FastAPI have have it's own logic to generate JSON Schema with very little duplication.

The main shortcoming I see with this is providing context about where the error was raised. Right now we don't have a good system for understanding the "path" that was taken through the crazy recursive function calls to produce the error. I am not opposed to implementing that logic in principle, but that logic would end up reflecting the structure of the CoreSchema more than the JsonSchema, which I could imagine leading to some confusion. Maybe you've thought of a good way to do it.

Another downside to always deferring error raising is that it makes it a lot more annoying to debug when you _want_ it to raise, and see where it was raised. If we add a deferred-error-collection mode I would suggest we still retain the ability to raise immediately (through one choice of an 'errors' kwarg or similar) so that we can easily get a stack trace to where the exception was raised when desired.

Either way, I think let's create an issue. (Actually I might just open a PR making the change I suggested above that at least retains information about where the error is within the json schema).

samuelcolvin:
Thanks so much for this @dmontagu, amazing to have this merged.

On the error stuff, my instinct is that the traceback might not be that useful anyway, we could even store the traceback with the errors if really necessary, but I'm also not that bothered about it.

It might be easier to just add a kwarg config setting to `GenerateJsonSchema` to do all the things with errors that @tiangolo wants, rather than aiming for absolute flexibility and making a horrible API.

dmontagu:
@Julian I've done some work on JSON schema for discriminated unions in #5051; I know it's more OpenAPI than JSON schema, so not sure if it's something you can help with, but I would appreciate any insight there. (And thanks for your offer to look at the JSON schema stuff, whether or not you can help in this particular case.)

PR #5039, opened 2023-02-10
Using `contextlib`'s suppress method to silence error

yezz123:
`context manager` slightly shortens the code and significantly clarifies the author's intention to ignore the specific errors.

[PEP 409 – Suppressing exception context](https://peps.python.org/pep-0409/)

cc @samuelcolvin

### DIFF
--- a/pydantic/_internal/_utils.py
+++ b/pydantic/_internal/_utils.py
@@ -3,8 +3,10 @@ Bucket of reusable internal utilities.
 
 This should be reduced as much as possible with functions only used in one place, moved to that place.
 """
+
 from __future__ import annotations as _annotations
 
+import contextlib
 import keyword
 import typing
 import weakref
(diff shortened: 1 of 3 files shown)


yezz123:
please review

samuelcolvin:
I don't think I agree with this, I suspect it has a non-zero overhead, and I'm not convinced it's clearer to the average user.

Happy to be contradicted.

dmontagu:
```python
import contextlib
import time

n = 5000000

t0 = time.time()
for i in range(n):
    with contextlib.suppress(TypeError, ValueError, RuntimeError):
        raise ValueError
t1 = time.time()
print(f"contextlib always error: {t1 - t0:.3f} seconds")

t0 = time.time()
for i in range(n):
    try:
        raise ValueError
    except (TypeError, ValueError, RuntimeError):
        pass
t1 = time.time()
print(f"try_except always error: {t1 - t0:.3f} seconds")

t0 = time.time()
for i in range(n):
    with contextlib.suppress(TypeError, ValueError, RuntimeError):
        if i % 100 == 0:
            raise ValueError
t1 = time.time()
print(f"contextlib rarely error: {t1 - t0:.3f} seconds")

t0 = time.time()
for i in range(n):
    try:
        if i % 100 == 0:
            raise ValueError
    except (TypeError, ValueError, RuntimeError):
        pass
t1 = time.time()
print(f"try_except rarely error: {t1 - t0:.3f} seconds")
```
Produces:
```
contextlib always error: 3.426 seconds
try_except always error: 1.616 seconds
contextlib rarely error: 2.512 seconds
try_except rarely error: 0.293 seconds
```

So, it seems to me that in basically all cases it is significantly more performant to use the try: except: approach, and it gets even more pronounced when exceptions are rare (which they probably will be in the modified code)

samuelcolvin:
Seems pretty clear.

yezz123:
Yes Seems good to not move with it 🙏🏻 

Kudos @dmontagu

PR #5048, opened 2023-02-13
Fix mishandling of longer reference cycles in recursive models

dmontagu:
This fixes an issue where recursive models that had "recursion cycles" of length longer than 2 resulted in infinite recursion while building the schema.

### DIFF
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -141,6 +141,7 @@ def complete_model_class(
         inner_schema, fields = build_inner_schema(
             cls, name, validator_functions, serialization_functions, bases, types_namespace
         )
+        inner_schema = _simplify_recursive_schema(inner_schema)
     except PydanticUndefinedAnnotation as e:
         if raise_errors:
             raise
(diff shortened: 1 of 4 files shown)


pydantic/_internal/_model_construction.py
```diff
@@ -266,9 +267,31 @@ def build_inner_schema(  # noqa: C901
...
+        if isinstance(s, dict):
+            if 'ref' in s and s.get('type') == 'typed-dict':
+                if s['ref'] in visited:
+                    return {'schema_ref': s['ref'], 'type': 'recursive-ref'}
+                else:
+                    visited.add(s['ref'])
+            return {k: replace_duplicate_refs(v) for k, v in s.items()}
+        elif isinstance(s, list):
+            return [replace_duplicate_refs(v) for v in s]
+        return s
```
dmontagu, line 290: This is a pretty awful hack, and I think a better approach would be to rework the original schema generation to not produce "duplicate" refs. This would probably amount to replacing some of the recursive function calls (to `build_inner_schema` in particular) with a generator class (similar to the GenerateJsonSchema one) that would keep track of the references as it goes (rather than relying on the local/global_ns stuff).

Given that, I'm thinking we probably shouldn't merge this PR as is (in particular, because this `_simplify_recursive_schema` function is pretty unsafe), and should probably do a bigger refactor. But I wanted to push the commit with this function to demonstrate the issue more concretely.

I would believe it is possible to resolve this issue with some smart changes to the `build_inner_schema` and related functions, but @samuelcolvin I'm curious if you think it might work better to rework the schema generation into a class that would retain the references in an attribute? I think it might not actually be too much work, but I'd like to get your sign off before going down that path.

dmontagu: Actually, I think this function may be useful, or even necessary, to handle a more core-schema-based approach to generics. Because if I want to perform a core-schema-substitution, and the substituted value has a ref in it, it might be good to remove all but one of the schemas with the same ref. I think I'll  make a more type-safe version of this function that has the sole purpose of replacing duplicate refs with recursive refs, and open a PR to pydantic-core to add it there. (If rejected there we can revisit this.)

pydantic/core_utils.py
```diff
@@ -0,0 +1,177 @@
+"""
+The goal is to move this file's contents into pydantic_core proper.
```
dmontagu, line 2: I put this stuff in regular pydantic for now to keep it simpler to get feedback, but I think all of this probably belongs in pydantic core if we find this useful.

pydantic/core_utils.py
```diff
@@ -0,0 +1,177 @@
...
+        schema = schema.copy()
+        if self.apply_before_recurse:
+            schema = self.f(schema)
+        method = self._schema_type_to_method[schema['type']]
+        schema = method(schema)
+        if not self.apply_before_recurse:
+            schema = self.f(schema)
+        return schema
+
+    def _handle_other_schemas(self, schema: core_schema.CoreSchema) -> core_schema.CoreSchema:
```
dmontagu, line 76: If we want to keep this file's code, I can replace this `_handle_other_schemas` method with the various type-specific methods it is currently covering, and we can add a test that ensures all schemas have methods defined (so adding new ones in the future would require us to update this class appropriately); I did it like this to reduce the amount of boilerplate I needed to test this approach out.

dmontagu: TODO: Make sure new types get handled properly

dmontagu:
This is now addressed in #5125

Issue #4483, opened 2022-09-05
Generic model parameterization raise `KeyError` after multiple calls 


uriyyo:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic  (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

After 200+ same generic model parameterization pydantic throws `KeyError`.
This issue is only reproducible when one generic model is inherited from another generic model and they use different instances of `TypeVar`.

The issue isn't reproducible in version `1.9.0`.

UPDATE: After investigation, I have found that [#4083](https://github.com/pydantic/pydantic/pull/4083) is the root cause of this issue.

### Example Code

```Python
from itertools import count
from typing import Generic, TypeVar

from pydantic import BaseModel
from pydantic.generics import GenericModel

T = TypeVar("T")


class Base(GenericModel, Generic[T]):
    a: T


R = TypeVar("R")


class Delivered(Base[R], Generic[R]):
    b: R


for i in count():
    class B(BaseModel):
        a: int


    m = Delivered[B]
```
```py
Traceback (most recent call last):
  File "/Users/yuriikarabas/my-projects/fastapi-pagination/temp.py", line 26, in <module>
    m = Delivered[B]
  File "/Users/yuriikarabas/my-projects/fastapi-pagination/.venv/lib/python3.9/site-packages/pydantic/generics.py", line 98, in __class_getitem__
    __base__=(cls,) + tuple(cls.__parameterized_bases__(typevars_map)),
  File "/Users/yuriikarabas/my-projects/fastapi-pagination/.venv/lib/python3.9/site-packages/pydantic/generics.py", line 219, in __parameterized_bases__
    yield from build_base_model(base_model, typevars_map)
  File "/Users/yuriikarabas/my-projects/fastapi-pagination/.venv/lib/python3.9/site-packages/pydantic/generics.py", line 185, in build_base_model
    parameterized_base = base_model.__class_getitem__(base_parameters)
  File "/Users/yuriikarabas/my-projects/fastapi-pagination/.venv/lib/python3.9/site-packages/pydantic/generics.py", line 98, in __class_getitem__
[...]

Issue #4131, opened 2022-06-02
`KeyError` in GenericModel.__parameterized_bases__.build_base_model


crazyscientist:
# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.9.1
            pydantic compiled: True
                 install path: /home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/pydantic
               python version: 3.9.12 (main, Apr 18 2022, 22:40:46)  [GCC 9.4.0]
                     platform: Linux-5.4.0-113-generic-x86_64-with-glibc2.31
     optional deps. installed: ['dotenv', 'typing-extensions']
```
I'm afraid reproduction of this _issue_ is a bit tricky as I was not able to reproduce it in a production scenario nor with a small code snippet. So far, I only noticed this in the test suite of this [project](https://github.com/SUSE/aimaas), which contains multiple sub-modules for tests; but only tests in two **can** fail, i.e. when the entire testsuite runs, they fail; if I explicitly only run tests from these modules, they pass. 

This behavior probably also indicates a problem within the testsuite, which I intend to look for, but the problem does not occur with version 1.9.0 of pydantic. So, I figured it's worth sharing.

In the project, we dynamically define API routes for FastAPI for dynamically generated models (i.e. model definitions are stored in DB), i.e. the models and routes get generated/updated at start-up or when something changes in the DB. The callback function used for this looks something like this:

```python
def route_get_entities(router: APIRouter, schema: Schema):
    # this is the callback
    factory = EntityModelFactory()
    filter_model = _filters_request_model(schema=schema)

    @router.get(
        f'/{schema.slug}',
        response_model=Page[factory(schema=schema, variant=ModelVariant.GET)],
        response_model_exclude_unset=True
    )
    def get_entities(
        filters: filter_model = Depends(),
        order_by: str = Query('name', description=''),
        ascending: bool = Query(True, description=''),
        db: Session = Depends(get_db),
        params: Params = Depends()
    ):
        # the actual API route
        ...
```
The exception raised in the testsuite looks like this:
```
_____________________________________ ERROR at setup of TestRouteGetEntityChanges.test_raise_on_change_doesnt_exist ______________________________________

dbsession = <sqlalchemy.orm.session.Session object at 0x7fca28575580>

    @pytest.fixture
    def client(dbsession):
>       app = create_app(session=dbsession)

backend/tests/conftest.py:338: 
[...]

PR #5052, opened 2023-02-14
fix: Fix broken parametrized bases with GenericModels

MarkusSintonen:
This [PR](https://github.com/pydantic/pydantic/pull/4083/files) broke the usage of parametrized bases with `GenericModel`s. Problem arises especially in applications having large number of generic data models (in our case a large FastAPI based application). The issue roots to the dictionary size limiting in `pydantic/generics.py`. When the limit is hit it starts to break usage of parametrized bases like here https://github.com/pydantic/pydantic/issues/4131#issuecomment-1429219280 and here https://github.com/pydantic/pydantic/issues/4483

There original PR tried to fix a memory leak by limiting the size of the internal dictionaries. But it seems its not correct way to do it as it can break generics usage. So instead this PR relies on weak dictionaries to handle any leaks with dynamic model usage.

### DIFF
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -17,6 +17,7 @@ from typing import (
     Union,
     cast,
 )
+from weakref import WeakKeyDictionary, WeakValueDictionary
 
 from typing_extensions import Annotated
 
@@ -25,7 +26,7 @@ from .fields import DeferredType
 from .main import BaseModel, create_model
 from .types import JsonWrapper
 from .typing import display_as_type, get_all_type_hints, get_args, get_origin, typing_base
-from .utils import LimitedDict, all_identical, lenient_issubclass
+from .utils import all_identical, lenient_issubclass
 
 if sys.version_info >= (3, 10):
     from typing import _UnionGenericAlias
@@ -35,13 +36,13 @@ TypeVarType = Any  # since mypy doesn't allow the use of TypeVar as a type
 
 Parametrization = Mapping[TypeVarType, Type[Any]]
 
-_generic_types_cache: LimitedDict[Tuple[Type[Any], Union[Any, Tuple[Any, ...]]], Type[BaseModel]] = LimitedDict()
+_generic_types_cache = WeakValueDictionary[Tuple[Type[Any], Union[Any, Tuple[Any, ...]]], Type[BaseModel]]()
 # _assigned_parameters is a Mapping from parametrized version of generic models to assigned types of parametrizations
 # as captured during construction of the class (not instances).
 # E.g., for generic model `Model[A, B]`, when parametrized model `Model[int, str]` is created,
 # `Model[int, str]`: {A: int, B: str}` will be stored in `_assigned_parameters`.
 # (This information is only otherwise available after creation from the class name string).
-_assigned_parameters: LimitedDict[Type[Any], Parametrization] = LimitedDict()
+_assigned_parameters = WeakKeyDictionary[Type[Any], Parametrization]()
 
 
 class GenericModel(BaseModel):


MarkusSintonen:
FYI with this change we are not observing memory leak as the original [issue](https://github.com/pydantic/pydantic/issues/3829) did. We are running a similar patch in production like the change here.

pydantic/generics.py
```diff
@@ -35,13 +36,20 @@
...
 
-_generic_types_cache: LimitedDict[Tuple[Type[Any], Union[Any, Tuple[Any, ...]]], Type[BaseModel]] = LimitedDict()
+if sys.version_info >= (3, 9):  # Typing for weak dictionaries available at 3.9
+    GenericTypesCache = WeakValueDictionary[Tuple[Type[Any], Any, Tuple[Any, ...]], Type[BaseModel]]
+    AssignedParameters = WeakKeyDictionary[Type[BaseModel], Parametrization]
+else:
+    GenericTypesCache = WeakValueDictionary
+    AssignedParameters = WeakKeyDictionary
+
+_generic_types_cache = GenericTypesCache()
```
samuelcolvin, line 46: can you add a comment explaining what this is about?

samuelcolvin (commented):
Humm, in principle this all looks good, we should definitely use this in V2.

I'm a bit of a wimp about making changes like this in V1.10 - I've seen too many have unexpected consequences.

@pydantic/pydantic-maintainers what do you think?

pydantic/generics.py
MarkusSintonen: Added a comment

samuelcolvin: thanks.

samuelcolvin:
@crazyscientist, would be amazing if you could review this and make sure it's working with your use case?

MarkusSintonen:
@adriangb it seems `pytest-memray` is bugged with parametrized tests having long names 😞 It works when not using `--memray-bin-path` param. But then we dont get the full result artifacts.

Its using the test name as the file name here https://github.com/bloomberg/pytest-memray/blob/1.4.0/src/pytest_memray/plugin.py#L139 Which breaks with long generated test names. Maybe it would be worth reporting to them 🤔

adriangb:
Thanks for tracking that down! 100% worth reporting. For now we can probably just put that test in it's own file and only run it on its own. What do you think?

adriangb:
Oh wait you fixed the issue. Never mind, I think this should work!

samuelcolvin:
We should definitely report this to mem ray.

I'll review this later.

crazyscientist:
@samuelcolvin thank you for the heads up. When I run our test suite with this fix, I unfortunately keep getting the `KeyError` raised as described in #4131.

samuelcolvin:
This is great, thanks all for your input.

In particular @MarkusSintonen thanks for your perseverance on this issue, I suspect you have the dubious honour of making the most complex use of pydantic generics.

We need this to be copied to `main`, probably worth waiting for #4970 to be merged. @MarkusSintonen would you be willing to contribute this?

Also, I'm yet again impressed with memray, I guess we should use it more.

MarkusSintonen:
> This is great, thanks all for your input. In particular @MarkusSintonen thanks for your perseverance on this issue, I suspect you have the dubious honour of making the most complex use of pydantic generics.

Haha no problem! 😊 

> We need this to be copied to main, probably worth waiting for https://github.com/pydantic/pydantic/pull/4970 to be merged. @MarkusSintonen would you be willing to contribute this?

Yeah I can open similar PR against `main`.

MarkusSintonen:
@samuelcolvin I already opened the V2 PR so I wont forget it: https://github.com/pydantic/pydantic/pull/5056

samuelcolvin:
awesome, thank you.

pablogsal:
> @adriangb it seems `pytest-memray` is bugged with parametrized tests having long names 😞 It works when not using `--memray-bin-path` param. But then we dont get the full result artifacts.
> 
> Its using the test name as the file name here https://github.com/bloomberg/pytest-memray/blob/1.4.0/src/pytest_memray/plugin.py#L139 Which breaks with long generated test names. Maybe it would be worth reporting to them 🤔

We are on it https://github.com/bloomberg/pytest-memray/issues/68

PR #5056, opened 2023-02-15
fix: V2 fix broken parametrized bases with GenericModels

MarkusSintonen:
Same as https://github.com/pydantic/pydantic/pull/5052 but for V2.

Requires https://github.com/pydantic/pydantic/pull/4970

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,42 @@ jobs:
         name: docs
         path: site
 
+  test-memray:
+    name: test memray
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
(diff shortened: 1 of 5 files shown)


dmontagu:
Thank you! I'll make sure this gets in with #4970

dmontagu:
I am pretty confident this is all handled now in #5125 (which I hope to merge today) but I will double check before closing this.

dmontagu:
There was one test in this PR that for some reason wasn't present in v1. Not sure if it was decided to remove it for some good reason there, but it was passing in v2 and I didn't see any good reason not to include it. So I added it to #5125 and will close this now, as I believe everything else from this PR is now in there. Thank you!

Issue #5071, opened 2023-02-18
Broken Link in the Documentation - pydantic/utils.py doesn't exist anymore


mrmammadov:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

There is a broken link in Pydantic Doc - [Data Binding part](https://docs.pydantic.dev/usage/models/#data-binding)
The link that refers to `GetterDict` in `utils.py` doesn't exist anymore in _main_ branch and seems like `utils.py` is moved to `_internal/_utils.py` but still `GetterDict` doesn't exist there as well. 

### Python, Pydantic & OS Version

```Text
pydantic version: 1.9.0
pydantic compiled: True
install path: /Users/gulumammadli/.pyenv/versions/3.10.0/lib/python3.10/site-packages/pydantic
python version: 3.10.0 (default, Jan 20 2022, 00:04:12) [Clang 12.0.5 (clang-1205.0.22.11)]
                     platform: macOS-11.6-x86_64-i386-64bit
optional deps. installed: ['dotenv', 'typing-extensions']
```

### Affected Components

- [X] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
- [ ] [Data validation/parsing](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://pydantic-docs.helpmanual.io/usage/exporting_models/) - `.model_dump()` and `.model_dump_json()`
- [ ] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)
- [ ] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)
- [ ] [Model Config](https://pydantic-docs.helpmanual.io/usage/model_config/)
- [ ] [Field Types](https://pydantic-docs.helpmanual.io/usage/types/) - adding or changing a particular data type
- [ ] [Function validation decorator](https://pydantic-docs.helpmanual.io/usage/validation_decorator/)
[...]

PR #5074, opened 2023-02-18
fix: Fix broken link in the docs (Pydantic Models - Data Binding)

mrmammadov:
Resolving [this issue](https://github.com/pydantic/pydantic/issues/5071)

skip change file check

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -155,7 +155,7 @@ Here a vanilla class is used to demonstrate the principle, but any ORM class cou
 ### Data binding
 
 Arbitrary classes are processed by *pydantic* using the `GetterDict` class (see
-[utils.py](https://github.com/pydantic/pydantic/blob/v1.10.5/pydantic/utils.py)), which attempts to
+[utils.py](https://github.com/pydantic/pydantic/blob/1.10.X-fixes/pydantic/utils.py)), which attempts to
 provide a dictionary-like interface to any class. You can customise how this works by setting your own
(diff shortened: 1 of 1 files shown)


docs/usage/models.md
```diff
@@ -155,7 +155,7 @@ Here a vanilla class is used to demonstrate the principle, but any ORM class cou
 ### Data binding
 
 Arbitrary classes are processed by *pydantic* using the `GetterDict` class (see
-[utils.py](https://github.com/pydantic/pydantic/blob/main/pydantic/utils.py)), which attempts to
+[utils.py](https://github.com/pydantic/pydantic/blob/v1.10.5/pydantic/utils.py)), which attempts to
```
hramezani, line 158: Why 
```suggestion
[utils.py](https://github.com/pydantic/pydantic/blob/1.10.X-fixes/pydantic/utils.py)), which attempts to
```

mrmammadov: @hramezani can you please explain me why? 
Shouldn't the links in docs target the latest stable release?

samuelcolvin: Because then we would have to change all such links on all releases, we could perhaps make that change in future, but that's a bigger issue.

mrmammadov: Okay, I see. Thanks for the clarification, will commit changes now.

hramezani (approved):

hramezani:
Thanks @mrmammadov

PR #5077, opened 2023-02-19
Fix mypy plugin for 1.1.0

cdce8p:
The next mypy release will add better support for `dataclass_transform`.

Since `ModelMetaclass` is decorated with it, it means that mypy would use that instead of the custom plugin. The plugin is quite good at parsing `BaseModel`, so the result would be worse. Especially with regards to the of default arguments as `pydantic.field.Field` uses a slightly different syntax.

With this PR, the plugin will basically pretend `ModelMetaclass` isn't decorated and thus continue to use the plugin for it.
I've also added a new (undocumented) option `debug_dataclass_transform` to be able to toggle that behavior and test new future versions of mypy without modifying the plugin again.

If accepted, it would be great if this PR could be backported to `1.10.X` and a new version released. The next mypy release which would otherwise break this behavior is currently scheduled for the end of February.

--
```py

class Event(BaseModel):
    id: int | None = Field(None, alias="id")

Event()
```

Without this PR mypy `1.1.0` would emit
```
error: Missing named argument "id" for "Event"  [call-arg]
```

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -74,6 +74,7 @@ except ImportError:  # pragma: no cover
 CONFIGFILE_KEY = 'pydantic-mypy'
 METADATA_KEY = 'pydantic-mypy-metadata'
 BASEMODEL_FULLNAME = 'pydantic.main.BaseModel'
+MODEL_METACLASS_FULLNAME = 'pydantic.main.ModelMetaclass'
 FIELD_FULLNAME = 'pydantic.fields.Field'
 DATACLASS_FULLNAME = 'pydantic.dataclasses.dataclass'
 
@@ -110,6 +111,11 @@ class PydanticPlugin(Plugin):
                 return self._pydantic_model_class_maker_callback
         return None
 
+    def get_metaclass_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
+        if fullname == MODEL_METACLASS_FULLNAME:
+            return self._pydantic_model_metaclass_marker_callback
+        return None
+
     def get_function_hook(self, fullname: str) -> 'Optional[Callable[[FunctionContext], Type]]':
         sym = self.lookup_fully_qualified(fullname)
         if sym and sym.fullname == FIELD_FULLNAME:
@@ -136,6 +142,20 @@ class PydanticPlugin(Plugin):
     def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
         transformer = PydanticModelTransformer(ctx, self.plugin_config)
         transformer.transform()
+
+    def _pydantic_model_metaclass_marker_callback(self, ctx: ClassDefContext) -> None:
+        """Reset dataclass_transform_spec attribute of ModelMetaclass.
+
+        Let the plugin handle it. This behavior can be disabled
+        if 'debug_dataclass_transform' is set to True', for testing purposes.
+        """
+        if self.plugin_config.debug_dataclass_transform:
+            return
+        info_metaclass = ctx.cls.info.declared_metaclass
+        assert info_metaclass
+        if getattr(info_metaclass.type, 'dataclass_transform_spec', None):
+            info_metaclass.type.dataclass_transform_spec = None  # type: ignore[attr-defined]
+        return None
 
     def _pydantic_field_callback(self, ctx: FunctionContext) -> 'Type':
         """
@@ -188,11 +208,18 @@ class PydanticPlugin(Plugin):
 
 
 class PydanticPluginConfig:
-    __slots__ = ('init_forbid_extra', 'init_typed', 'warn_required_dynamic_aliases', 'warn_untyped_fields')
+    __slots__ = (
+        'init_forbid_extra',
+        'init_typed',
+        'warn_required_dynamic_aliases',
+        'warn_untyped_fields',
+        'debug_dataclass_transform',
+    )
     init_forbid_extra: bool
     init_typed: bool
     warn_required_dynamic_aliases: bool
     warn_untyped_fields: bool
+    debug_dataclass_transform: bool  # undocumented
 
     def __init__(self, options: Options) -> None:
         if options.config_file is None:  # pragma: no cover
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -86,6 +86,9 @@ def parse_mypy_version(version: str) -> Tuple[int, ...]:
 MYPY_VERSION_TUPLE = parse_mypy_version(mypy_version)
 BUILTINS_NAME = 'builtins' if MYPY_VERSION_TUPLE >= (0, 930) else '__builtins__'
 
+# Increment version if plugin changes and mypy caches should be invalidated
+PLUGIN_VERSION = 1
+
 
 def plugin(version: str) -> 'TypingType[Plugin]':
     """
@@ -101,6 +104,7 @@ class PydanticPlugin(Plugin):
     def __init__(self, options: Options) -> None:
         self.plugin_config = PydanticPluginConfig(options)
         self._plugin_data = self.plugin_config.to_data()
+        self._plugin_data['version'] = PLUGIN_VERSION
         super().__init__(options)
 
     def get_base_class_hook(self, fullname: str) -> 'Optional[Callable[[ClassDefContext], None]]':
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -152,10 +152,9 @@ class PydanticPlugin(Plugin):
         if self.plugin_config.debug_dataclass_transform:
             return
         info_metaclass = ctx.cls.info.declared_metaclass
-        assert info_metaclass
+        assert info_metaclass, "callback not passed from 'get_metaclass_hook'"
         if getattr(info_metaclass.type, 'dataclass_transform_spec', None):
             info_metaclass.type.dataclass_transform_spec = None  # type: ignore[attr-defined]
-        return None
 
     def _pydantic_field_callback(self, ctx: FunctionContext) -> 'Type':
         """
--- a/changes/5077-cdce8p.md
+++ b/changes/5077-cdce8p.md
@@ -1,0 +1,1 @@
+Fix mypy plugin for v1.1.0


yezz123 (approved):
LGTM 👍🏻

dmontagu:
Looks like this PR is against the `main` branch; should we also do a backport to v1.10.X?

(Looks good to me as well.)

cdce8p:
> Looks like this PR is against the `main` branch; should we also do a backport to v1.10.X?

That would be great. The plugin wouldn't work properly for the next mypy release (1.1.0) without this change.

pydantic/mypy.py
```diff
@@ -137,6 +143,20 @@ def _pydantic_model_class_maker_callback(self, ctx: ClassDefContext) -> None:
...
+        Let the plugin handle it. This behavior can be disabled
+        if 'debug_dataclass_transform' is set to True', for testing purposes.
+        """
+        if self.plugin_config.debug_dataclass_transform:
+            return
+        info_metaclass = ctx.cls.info.declared_metaclass
+        assert info_metaclass
+        if getattr(info_metaclass.type, 'dataclass_transform_spec', None):
+            info_metaclass.type.dataclass_transform_spec = None  # type: ignore[attr-defined]
+        return None
```
hramezani, line 158: Do we need this `return` here?

hramezani (approved):
LGTM :+1: 

Thanks @cdce8p for this.

It seems there are two Prs in mypy:

- https://github.com/python/mypy/pull/14734 (Created by @cdce8p)
- https://github.com/python/mypy/pull/14695

Seems we need this fix whether your PR is merged or the other one

cdce8p:
> It seems there are two Prs in mypy:
> 
> * [python/mypy#14734](https://github.com/python/mypy/pull/14734) (Created by @cdce8p)
> * [python/mypy#14695](https://github.com/python/mypy/pull/14695)

Yeah, although both would address a different issue to this one. 

> Seems we need this fix in `main` and `1.10.x` whether your PR is merged or the other one

Exactly 👍🏻

hramezani:
Forgot to mention, I think we need a test for this change as well.

cdce8p:
> Forgot to mention, I think we need a test for this change as well.

That's a bit more complicated 😅
This PR is basically a proactive regression fix for a yet unreleased mypy version. There are already mypy tests in place and from a quick look I think those would fail without this PR if executed with the current mypy master.

Of course the obvious issue here is that the mypy change isn't released yet. So just in theory, there is a (very) slim chance the attribute could change which would require additional work here. However, with the `getattr` guard, it would just be a no-op and wouldn't cause any additional issues. That's also what would happen for mypy < 1.1 and what is already tested. 
Personally, I don't think that will happen and having a new pydantic version out would help better test mypy, so I already opened this one.

If you really want to have a test first, it might be better to wait for the actual mypy release and see it break first. Or we move forward with the PR and I come back later to update the mypy version used for the tests.

dmontagu:
@cdce8p @hramezani I think if the current mypy plugin tests fail if run using mypy 1.1.0, but pass with this change, then we shouldn't need a new test.

dmontagu:
@cdce8p Does this mean that if using mypy without the plugin, anyone using `Field` will start getting errors? I'm wondering if we'll need to drop the `@dataclass_transform` because of that..

cdce8p:
> @cdce8p Does this mean that if using mypy without the plugin, anyone using `Field` will start getting errors?

Not just with `Field` unfortunately. The pydantic syntax for arguments with defaults is incompatible with `dataclass_transform`. Thus, starting with mypy 1.1 basically everyone who is **not** using the plugin will get errors if they are missing some arguments even if they have default values. E.g.

```py

class Event(BaseModel):
    id: int | None = Field(None, alias="event")
    required_var: str | None = Field(...)
    optional_var: str | None

Event()
```

```
# mypy v1.0 (without plugin)
--

# mypy v1.0 (with plugin)
error: Missing named argument "required_var" for "Event"

# mypy v1.1.dev (without plugin)
error: Missing named argument "event" for "Event"  [call-arg]
error: Missing named argument "required_var" for "Event"  [call-arg]
error: Missing named argument "optional_var" for "Event"  [call-arg]

# mypy v1.1.dev (with current plugin v1.10.5)
error: Missing named argument "event" for "Event"  [call-arg]
error: Missing named argument "required_var" for "Event"  [call-arg]
error: Missing named argument "optional_var" for "Event"  [call-arg]

# mypy v1.1.dev (with plugin + this PR)
error: Missing named argument "required_var" for "Event"
```

> I'm wondering if we'll need to drop the `@dataclass_transform` because of that..

Tbh I don't know if other type checker depend on it, mainly pyright. I would image they also have custom logic for pydantic just because of the default value issue. However IMO, the better alternative would be to adopt the "default" syntax for custom Fields. I.e. any argument is required **unless** it is assigned `Field` with an keyword only `default` argument.
It would be a massive breaking change, although still possible as there is a transition path. `Field(default=<some value>)` is already valid.

> I think if the current mypy plugin tests fail if run using mypy 1.1.0, but pass with this change, then we might not need a new test.

Ran the mypy tests. `test_mypy_results[mypy-default.ini-success.py-None]` does fail without this change.

dmontagu:
@samuelcolvin

> Thus, starting with mypy 1.1 basically everyone who is not using the plugin will get errors if they are missing some arguments even if they have default values.

Want to bring this to your attention.

> I.e. any argument is required unless it is assigned Field with an keyword only default argument.
@cdce8p Can you run the same except where you set a default not using `Field`? (i.e., `other_var: str = "abc"`) — want to make sure that case still works okay.

I think it's reasonable to tell people they have to use `default=...` in Field to get it to work properly without the plugin. (Also, the Optional fields being still optional without a default value set is going to change in v2, so I'm a bit less concerned about that).

cdce8p:
> > Thus, starting with mypy 1.1 basically everyone who is not using the plugin will get errors if they are missing some arguments even if they have default values.
> 
> Want to bring this to your attention. ☝️

Just to clarify, some example
```py
class Event(BaseModel):
    # ok
    a: int | None = Field(default=None)
    b: int | None = None
    c: int
    d: int = Field(...)  # required -> only works without 'default' keyword

    # error
    e: int | None = Field(None)  # false-positive: not recognized as optional
    f: int | None                # false-positive: not recognized as optional
    g: int = ...                 # false-negative: not recognized as required

# Add empty constructor to show errors
Event()
```

cdce8p:
> I think it's reasonable to tell people they have to use `default=<value>` in `Field` to get it to work properly without the plugin, would be better if not but I suspect the benefits from `@dataclass_transform` might outweigh the costs. (Also, the `Optional` fields being still optional without a default value set is going to change in v2, so I'm a bit less concerned about that).

It might still makes sense to add this PR and not fall back to `dataclass_transform` entirely. One advantage of the plugin is that it parses the Model config, i.e. to determine if the `alias` should be used.

Never the less, syntax that more aligns with the stdlib `dataclass` would certainly help.

dmontagu:
Oh yeah I didn’t mean to indicate we shouldn’t merge this PR

dmontagu (approved):

samuelcolvin (approved):
Sorry, just catching up on this.

In principle this looks good (although I don't pretend to understand all the nuances).

I'll discuss with @dmontagu today and hopefully merge

A few (perhaps related) questions:
* should this be back-ported to the `1.10` pydantic branch?
* do the [changes to required logic in v2](https://docs.pydantic.dev/blog/pydantic-v2/#required-vs-nullable-cleanup) affect this at all?

Thanks so much @cdce8p  for digging into this and all your work.

samuelcolvin (approved):
Sorry, just catching up on this.

In principle this looks good (although I don't pretend to understand all the nuances).

I'll discuss with @dmontagu today and hopefully merge

A few (perhaps related) questions:
* should this be back-ported to the `1.10` pydantic branch?
* do the [changes to required logic in v2](https://docs.pydantic.dev/blog/pydantic-v2/#required-vs-nullable-cleanup) affect this at all?

Thanks so much @cdce8p  for digging into this and all your work.

cdce8p:
> * should this be back-ported to the `1.10` pydantic branch?

Yes

> * do the [changes to required logic in v2](https://docs.pydantic.dev/blog/pydantic-v2/#required-vs-nullable-cleanup) affect this at all?

No. The changes will improve the type checking experience for anyone `not` using the mypy plugin though. So they are definitely a step in the right direction. If I could suggest some more changes, I would recommend to
* Remove support of `g: int = ...` -> This is an assignment error in mypy already and required fields can be shorted to `g: int` anyway.
* Make `Field.default` a keyword only argument and remove support of `Field(...)`. This will match [PEP 681](https://peps.python.org/pep-0681/#field-specifier-parameters) and thus further improve the experience for anyone `not` using the plugin.

samuelcolvin:
on your points:
* the ellipsis is entirely optional (pun/confusion not intended) and has no effect, we're only supporting it for backwards compatibility, we won't use it in any V2 docs except the migration guide
* Again, I think we'll continue to support `Field(42)` rather than `Field(default=42)` for backwards compatibility reasons, but if you care about static typing, you can always use `Field(default=42)`
* I don't think `FieldInfo` is only public so you can inspect `.model_fields`, it's not designed to be used publicly otherwise - so I imagine mypy shouldn't care about it?

cdce8p:
> * the ellipsis is entirely optional (pun/confusion not intended) and has no effect (`g: int = ...` will mean exactly the the same thing as `g: int` in v2)

Not in the context of `dataclass_transform` unfortunately. Since `...` is technically a value, `g` would be interpreted as optional.

> * Again, I think we'll continue to support `Field(42)` rather than `Field(default=42)` for backwards compatibility reasons, but if you care about static typing, you can always use `Field(default=42)`

True, however that doesn't work if the model is part of a library that is used somewhere else. Especially if the maintainer of said library doesn't want to change it only to better support static typing.

> * `FieldInfo` is only public so you can inspect `.model_fields` (perhaps we should make it private and add a protocol to represent it), it's not designed to be used publicly otherwise - so I imagine mypy shouldn't care about it?

Didn't know that. Only saw it in the `dataclass_transform` call as `field_specifier`. It should probably be removed there then
https://github.com/pydantic/pydantic/blob/ae481b99b3ce85802751770b523e89c6e3f51c12/pydantic/main.py#L49-L50

https://github.com/pydantic/pydantic/blob/ae481b99b3ce85802751770b523e89c6e3f51c12/pydantic/dataclasses.py#L175-L176

--

Tbh those suggestions would be nice but the pydantic mypy plugin resolves all of these issues, once mypy knows to ignore the `dataclass_transform` with this PR. It would however help for other type checker which don't have custom code for pydantic or if users don't use the plugin (maybe because they don't know about it).

samuelcolvin:
@cdce8p we can add a recommendation to the docs to do those things "the right way", but we don't want to break V1 code unless we have to.

KotlinIsland:
Is this working with `dataclasses`? I'm seeing invalid errors:
```py
import pydantic

class A:
    a: str

```

cdce8p:
> Is this working with `dataclasses`? I'm seeing invalid errors:
> 
> ```python
> import pydantic
> 
> @pydantic.dataclasses.dataclass
> class A:
>     a: str
> 
> A("")  # error: Too many positional arguments for "A"  [misc]
> ```

Pydantic dataclasses are keyword only: https://docs.pydantic.dev/usage/dataclasses/
If that is wrong, the `dataclass_transform` decorator should be changed. So it's somewhat unrelated to the fix here for `ModelMetaclass`.
https://github.com/pydantic/pydantic/blob/6267ae3538cf7237ff6769174b4c340415e0f444/pydantic/dataclasses.py#L175-L176

KotlinIsland:
@cdce8p I don't think they are keyword only as that example works correctly at runtime and with the pydantic mypy plugin with v1.0.1. It's only with this change that they are now reporting errors.

cdce8p:
> @cdce8p I don't think they are keyword only as that example works correctly at runtime and with the pydantic mypy plugin with v1.0.1. It's only with this change that they are now reporting errors.

Are you sure that it's this change? As it doesn't change anything regarding `pydantic.dataclasses.dataclass` I wouldn't be so sure. What likely did trigger it was the mypy master branch itself as that starts parsing the `dataclass_transform` decorator.

KotlinIsland:
I did mean that, that mypy caused this defect originally. But also that, this PR doesn't seem to be a complete fix, as the functionality of dataclassess has regressed.

samuelcolvin:
pydantic dataclasses are not kwarg only, however we decorate them with `dataclass_transform(kw_only_default=True, ...`, see:

https://github.com/pydantic/pydantic/blob/314abcabbde0936ec2e8e5ff6532d64a221210c5/pydantic/dataclasses.py#L96

and other overloads. 

Not sure why we did that, but I guess it could be changed?

cdce8p:
Mypy `v1.1` was released a few hours ago. As expected it broke the current pydantic plugin. It would be awesome if the changes here could be released. The cherry-pick PR would be ready #5111.

It could make sense to include #5120 as well, although that was already broken.

https://mypy-lang.blogspot.com/2023/03/mypy-111-released.html

cdce8p:
> @cdce8p Do you mind if I merge the stuff from #5120 directly into this branch, along with some changes to tests to get everything behaving nicely? If so, I'll take care of it and I think we can have this ready to merge first thing tomorrow. (If I don't hear from you in the next few hours I might just start pushing..)

Not at all! Please let me know if I can help.

dmontagu:
(And to be clear, I meant the cherry-pick branch for 1.10.X, not _this_ branch.)

My current plan is to make it so that you can get strict/lenient behavior on mypy 1.1.1 by using the plugin, but without the plugin, when using mypy 1.1.1 it becomes "strict" in accordance with mypy's handling of `@dataclass_transform`. Anyone who **both** doesn't want to use the plugin **AND** wants mypy to be lenient, will basically just need to not upgrade to mypy 1.1.1.

@cdce8p would that work for you?

I will add some commits onto the cherry-pick branch, once I've done I'll notify you and it would be great if you could confirm if the changes work for you on Home Assistant.

cdce8p:
> My current plan is to make it so that you can get strict/lenient behavior on mypy 1.1.1 by using the plugin, but without the plugin, when using mypy 1.1.1 it becomes "strict" in accordance with mypy's handling of `@dataclass_transform`. Anyone who **both** doesn't want to use the plugin **AND** wants mypy to be lenient, will basically just need to not upgrade to mypy 1.1.1.
> 
> @cdce8p would that work for you?

That sounds like exactly what this PR / the cherry-pick one does, so yes. Basically the PR only restores the plugin behavior for mypy `1.1.1`, without the plugin the `dataclass_transform` is used.

> I will add some commits onto the cherry-pick branch, once I've done I'll notify you and it would be great if you could confirm if the changes work for you on Home Assistant.

Sounds good

--
Do you want to include a backport for #5120 as well? It also deals with the `dataclass_transform` decorator. The PR is ready. Just needs to be backported after it's merged.

dmontagu:
Yeah I realized that I didn't need to do anything because I wasn't going to make the tests actually pass on 1.1.1. So I just merged your changes in there.

dmontagu:
@andreagrandi See https://github.com/pydantic/pydantic/issues/5192 — I'll make sure that we address this once we get back to looking at the mypy plugin again (hopefully not too far out).

andreagrandi:
@dmontagu thank you for the quick reply and my bad I didn't check among existing issues 🙇🏻 I will stay tuned, thanks again!

PR #5085, opened 2023-02-21
Fix mypy plugin import for version >= 1.x.x

treasuryspring:
# NOTICE

### DIFF
--- a/changes/5085-treasuryspring.md
+++ b/changes/5085-treasuryspring.md
@@ -1,0 +1,1 @@
+Fix mypy plugin import for version >= 1.x.x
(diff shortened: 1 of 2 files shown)


hramezani:
Hi @treasuryspring,

Could you please explain what problem are you going to fix with this patch?
With `mypy 1.x.x` the `MYPY_VERSION_TUPLE` would be `(1, x, x)` for example `(1, 0, 1)` and the `MYPY_VERSION_TUPLE >= (0, 930)` should work. why do we need adding `len(MYPY_VERSION_TUPLE) > 2 or`?

Issue #5083, opened 2023-02-21
mypy tests are taking minutes (literally)


samuelcolvin:
What's going on with mypy tests, they're taking an incredibly long time, see [this recent build](https://twitter.com/wes321/status/1627718493691838478).

Slow tests:

```
62.41s call     tests/mypy/test_mypy.py::test_mypy_results[pyproject-plugin-strict.toml-plugin_success.py-plugin-success-strict.txt]
59.77s call     tests/mypy/test_mypy.py::test_mypy_results[mypy-plugin-strict-no-any.ini-dataclass_no_any.py-None]
56.15s call     tests/mypy/test_mypy.py::test_mypy_results[pyproject-default.toml-success.py-None]
54.15s call     tests/mypy/test_mypy.py::test_mypy_results[mypy-plugin-strict.ini-plugin_success.py-plugin-success-strict.txt]
51.63s call     tests/mypy/test_mypy.py::test_mypy_results[mypy-default.ini-success.py-None]
51.50s call     tests/mypy/test_mypy.py::test_mypy_results[pyproject-plugin.toml-plugin_success.py-None]
49.02s call     tests/mypy/test_mypy.py::test_mypy_results[mypy-plugin.ini-plugin_success.py-None]
1.41s call     tests/mypy/test_mypy.py::test_mypy_results[pyproject-default.toml-fail4.py-fail4.txt]
1.31s call     tests/mypy/test_mypy.py::test_mypy_results[mypy-plugin-strict.ini-plugin_fail_baseConfig.py-plugin-fail-strict-baseConfig.txt]
0.67s call     tests/mypy/test_mypy.py::test_mypy_results[pyproject-default.toml-fail3.py-fail3.txt]
```

This wasn't the case until a recent upgrade of mypy.

Any suggestions on how to fix it?

koxudaxi:
> see [this recent build](https://twitter.com/wes321/status/1627718493691838478).

:thinking: 
It's only 1 min.

koxudaxi:
What mypy version did you say?
I tested some versions without mypy cache in the main branch.
The latest version improved the speed.
`coverage run -m pytest --durations=10 tests/mypy `

```bash
tail -n3 mypy*.log 
==> mypy0.961.log <==
Results (136.28s):
        38 passed
         5 xfailed

==> mypy0.982.log <==
Results (286.70s):
        38 passed
         5 xfailed

==> mypy1.0.0.log <==
Results (75.57s):
        38 passed
[...]

PR #5086, opened 2023-02-21
Update mypy version to 1.0.1

koxudaxi:
The PR updates mypy version to 1.0.1 for performance.

## Related Issues
 https://github.com/pydantic/pydantic/issues/5083

## Motivation
The current CI test is a little slow for mypy tests.
Mypy 1.0.0 is improved performance.
from the blog 👇

https://mypy-lang.blogspot.com/2023/02/mypy-10-released.html
_Performance Improvements
Mypy 1.0 is up to 40% faster than mypy 0.991 when type checking the Dropbox internal codebase. We also set up a daily job to measure the performance of the most recent development version of mypy to make it easier to track changes in performance._

## Easy benchmarks
`coverage run -m pytest --durations=10 tests/mypy`

```bash
tail -n3 mypy*.log 
==> mypy0.961.log <==
Results (136.28s):
        38 passed
         5 xfailed

==> mypy0.982.log <==
Results (286.70s):
        38 passed
         5 xfailed

==> mypy1.0.0.log <==
Results (75.57s):
        38 passed
         5 xfailed

==> mypy1.0.1.log <==
Results (75.38s):
        38 passed
         5 failed
```


pydantic/_internal/_validators.py
```diff
@@ -157,7 +157,9 @@ def __call__(self, __input_value: int | float | str, **_kwargs: Any) -> Decimal:
                 raise PydanticKnownError('finite_number')
 
             if self.check_digits:
-                if exponent >= 0:
+                if isinstance(exponent, str):
+                    raise PydanticKnownError('finite_number')
+                elif exponent >= 0:
```
koxudaxi, line 162: mypy show  the type error of comparing `literal` and  `int`
`check_digits` feature doesn't work for Inf and NaN values.

pydantic/json.py
```diff
@@ -35,7 +35,8 @@ def decimal_encoder(dec_value: Decimal) -> Union[int, float]:
     >>> decimal_encoder(Decimal("1"))
     1
     """
-    if dec_value.as_tuple().exponent >= 0:
+    exponent = dec_value.as_tuple().exponent
+    if isinstance(exponent, int) and exponent >= 0:
         return int(dec_value)
```
koxudaxi, line 40: I can change `isinstance` to use try/exception for performance.

koxudaxi:
I added some conditions when the Decimal has `Inf` and `NaN`.
If we should separate the PR for the change, I will create it.

pydantic/_internal/_validators.py
samuelcolvin: i think this is fine, but please add a comment explaining why we need the above.

pydantic/json.py
samuelcolvin: I think this is fine, we'll need to change this code anyway.

samuelcolvin (approved):
otherwise LGTM.

This is great, thank you.

PR #5094, opened 2023-02-22
Add tests demonstrating an approach to JSON schema error collection

dmontagu:
I have opened this pull request for the sake of gathering feedback on a proposed approach to @tiangolo's feature request brought up in #5029.

I found a way to implement it that I believe is 100% compatible with the current `GenerateJsonSchema` implementation with no lying to mypy or anything, just based on adding overrides. (@tiangolo the main difference with my initial implementation I shared is that by creating a custom subclass of `str` we can return something that is a valid `JsonSchemaValue` and yet can be detected as definitely-created-by-the-overridden-method).

There are two main issues on the pydantic side:
* `'PydanticInvalidForJsonSchema'` was not in `pydantic.json_schema.__all__`; that should almost certainly be changed.
* `CoreMetadataHandler` is needed in the override method, but is in `pydantic._internal`. We have at least two options for resolving this:
    * We could make `CoreMetadataHandler` public (might want/need to do this for other reasons, though it would be nice to make the API cleaner if we did)
    * We could change the `handle_invalid_for_json_schema` to more narrowly wrap the raising of the error. The main reason I didn't already do that was because I was uncertain about whether my approach of not raising an error if there was a custom `__pydantic_modify_json_schema__` function was good/going to stand the test of time.

I don't intend to merge the bulk of this PR, but I'd be happy to transform it into something we do want to merge to resolve the two points above. Or, once @tiangolo is satisfied that the JSON schema generation is sufficiently flexible for FastAPI integration, we can close this PR and keep it around for (his?) reference, and I can open a separate PR resolving the issues mentioned above.

### DIFF
--- a/tests/test_fastapi_json_schema.py
+++ b/tests/test_fastapi_json_schema.py
@@ -1,0 +1,137 @@
+"""
+This file contains an initial proposal that can be scrapped and reworked if/when appropriate.
+Either way, this test file should probably be removed once the actual FastAPI implementation
+is complete and has integration tests with pydantic v2. However, we are including it here for now
+to get an early warning if this approach would require modification for compatibility with
+any future changes to the JSON schema generation logic, etc.
+
+See the original PR for more details: https://github.com/pydantic/pydantic/pull/5094
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+from pydantic_core import CoreSchema
+from pydantic_core.core_schema import TypedDictField
+
+from pydantic import BaseModel, ConfigDict
+from pydantic._internal._core_metadata import CoreMetadataHandler
+from pydantic.errors import PydanticInvalidForJsonSchema
+from pydantic.json_schema import GenerateJsonSchema, JsonSchemaValue
+
+
+class _ErrorKey(str):
+    pass
+
+
+class FastAPIGenerateJsonSchema(GenerateJsonSchema):
+    """
+    Idea: This class would be exported from FastAPI, and if users want to modify the way JSON schema is generated
+    in FastAPI, they should inherit from it and override it as appropriate.
+
+    In the JSON schema generation logic, FastAPI _could_ also attempt to work with classes that inherit directly from
+    GenerateJsonSchema by doing something like:
+
+        if UserGenerateJsonSchema.handle_invalid_for_json_schema is GenerateJsonSchema.handle_invalid_for_json_schema:
+            # The method has not been overridden; inherit from FastAPIGenerateJsonSchema
+            UserGenerateJsonSchema = type(
+                "UserGenerateJsonSchema", (FastAPIGenerateJsonSchema, UserGenerateJsonSchema), {}
+            )
+        else:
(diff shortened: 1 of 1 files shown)


tests/test_fastapi_json_schema.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if UserGenerateJsonSchema.handle_invalid_for_json_schema is GenerateJsonSchema.handle_invalid_for_json_schema:
+            # The method has not been overridden; inherit from FastAPIGenerateJsonSchema
+            UserGenerateJsonSchema = type(
+                "UserGenerateJsonSchema", (FastAPIGenerateJsonSchema, UserGenerateJsonSchema), {}
+            )
+        else:
+            raise TypeError(f"{UserGenerateJsonSchema.__name__} should inherit from FastAPIGenerateJsonSchema")
+
+    I'm not sure which approach is better.
```
dmontagu, line 35: I've included some notes about how I see this getting used in FastAPI above ☝️

tests/test_fastapi_json_schema.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def test_inheritance_detection() -> None:
+    class GenerateJsonSchema2(GenerateJsonSchema):
+        pass
+
+    assert GenerateJsonSchema2.handle_invalid_for_json_schema is GenerateJsonSchema.handle_invalid_for_json_schema
+    assert (
+        FastAPIGenerateJsonSchema.handle_invalid_for_json_schema
+        is not GenerateJsonSchema.handle_invalid_for_json_schema
```
dmontagu, line 80: this is just a quick proof that my suggestion above _does_ work for checking if a method was overridden.

dmontagu:
@samuelcolvin not sure if you feel this approach falls under "horrible API" as referenced here: https://github.com/pydantic/pydantic/pull/5029#issuecomment-1440802094, and would prefer to just handle it with an argument.

My opinion is this is pretty clean, but I'm open to alternative points of view 😄

tiangolo:
He, this is so clever. I enjoy seeing you work. :star_struck: :sunglasses: 

I'm pretty sure this would work, and it has the advantage of requiring little changes in Pydantic itself.

What I would do/suggest is keeping it on hold until more things are ready (not sure if even the first release(s)) and then finish it while developing the FastAPI side, both together. As many corner cases only show up during the actual implementation, and it might end up being evident that it requires more stuff, or less (for any reason).

dmontagu:
@tiangolo sounds good. I might merge in the change making the PydanticInvalidForJsonSchema error public, but I'll do that separately.

samuelcolvin:
@dmontagu happy to merge this.

dmontagu:
I'll add a note to the test file making it more clear that this is just an initial proposal that can be scrapped if/when appropriate, and probably should be removed once the acutal FastAPI implementation is complete and has integration tests with pydantic v2.

dmontagu:
@samuelcolvin I have added my above notes into the file, and generally added more description of why this file was added and when it should be removed. I'll merge this later today unless I hear otherwise from you.

Issue #5101, opened 2023-02-23
Attribute error for `functools.partial`


JensHeinrich:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

Using a `partialmethod`or `partial` as a `validator` or `root_validator` creates an unhelpful `AttributeError`:
```
tests/conftest.py:14: in <module>
    class TestClass(BaseModel):
tests/conftest.py:17: in TestClass
    _custom_validate = validator("name")(validate)
pydantic/class_validators.py:93: in pydantic.class_validators.validator.dec
    ???
pydantic/class_validators.py:150: in pydantic.class_validators._prepare_validator
    ???
E   AttributeError: 'functools.partial' object has no attribute '__qualname__'
```

The problem is rooted at [this code](https://github.com/pydantic/pydantic/blob/314abcabbde0936ec2e8e5ff6532d64a221210c5/pydantic/class_validators.py#L150).

IMO an try except block with a test for partial methods would be useful here, telling the User to supply the "allow_reuse" parameter for partials.

### Example Code

```Python
from functools import partial

from pydantic import validator, BaseModel
from pydantic.main import ModelMetaclass
from pydantic.utils import GetterDict

def custom_validator(additional_stuff: str, cls: ModelMetaclass, values: GetterDict):
    print(additional_stuff)
    return values

validate = partial(custom_validator, "TEXT")

class TestClass(BaseModel):
    name: str

    _custom_validate = validator("name")(validate)
```

### Python, Pydantic & OS Version

```Text
pydantic version: 1.10.5
            pydantic compiled: True
                 install path: XXX/.venv/lib/python3.10/site-packages/pydantic
               python version: 3.10.7 (main, Nov 24 2022, 19:45:47) [GCC 12.2.0]
                     platform: Linux-5.19.0-31-generic-x86_64-with-glibc2.36
     optional deps. installed: ['typing-extensions']
```

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
[...]

PR #5102, opened 2023-02-23
Add test for functoool.partial validators

JensHeinrich:
Add testcase for #5101

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -1,9 +1,9 @@
 from collections import deque
 from datetime import datetime
 from enum import Enum
-from functools import partial
+from functools import partial, partialmethod
 from itertools import product
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import pytest
 from typing_extensions import Literal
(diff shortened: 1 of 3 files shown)


tests/test_validators.py
```diff
@@ -1348,15 +1348,23 @@ class Model(BaseModel):
...
-@pytest.mark.xfail()
-def test_partialmethod_as_validator():
-    def custom_validator(additional_stuff: str, cls, values):
-        print(additional_stuff)
-        return values
+@pytest.mark.parametrize('func,allow_reuse', product([partialmethod, partial], [True, False]))
+def test_functool_as_validator(
+    func: Callable,
+    allow_reuse: bool,
+):
```
samuelcolvin, line 1355: I'm a bit confused by all the magic here, could we simplify this test?

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin:
Also I see tests are failing. 

Please update.

tests/test_validators.py
JensHeinrich: I could write the 4 cases as a list I guess

It's just testing both `functools.partial` and `functools.partialmethod` with `allow_reuse` on and off

JensHeinrich: Are the testcases easier to understand with the new ids?

tests/test_validators.py
```diff
@@ -1345,3 +1346,33 @@ class Model(BaseModel):
...
     ]
+
+
+@pytest.mark.parametrize(
+    'func,allow_reuse',
+    [
+        pytest.param(partial, False, id='`partial` and check for reuse'),
+        pytest.param(partial, True, id='`partial` and ignore reuse'),
+        pytest.param(partialmethod, False, id='`partialmethod` and check for reuse'),
+        pytest.param(partialmethod, False, id='`partialmethod` and ignore reuse'),
```
samuelcolvin, line 1357:
```suggestion
        pytest.param(partialmethod, True, id='`partialmethod` and ignore reuse'),
```
???

JensHeinrich: And that's the reason I used a generator at first: I'm really bad at copy paste

samuelcolvin (approved):
LGTM, @dmontagu are you happy with this?

tests/test_validators.py
samuelcolvin: I think this is much clearer, although there's a few "copied" lines.

For future reference, if you do want to do the "product" thing, imho you'd be better off using two decorators, one for each argument rather than the `product` function.

dmontagu:
Yes I'm fine with it.

If possible I think we should support functools validators in v2 (given my understanding of how validators will work in v2 though I think that should be straightforward).

dmontagu:
Does it make sense to close this in favor of #5126 ?

PR #5105, opened 2023-02-24
Fix up some of the test_edge_cases.py tests

dmontagu:
This is _mostly_ just fixes to tests that are necessary due to changes in properties available and/or error formats, but there are a couple behaviors I was wondering if we want to continue to support or not that would require some changes in pydantic_core. I will add comments in-line about these.

There are more tests in test_edge_cases.py that can be fixed, but this was as much as I could get to before I had to stop for a while, and it's probably better to keep the PR a bit smaller anyway.. 😅


tests/test_edge_cases.py
```diff
@@ -546,12 +556,22 @@ class Model(BaseModel):
...
 
-    # abstract set
-    assert m.model_dump(include={'a': 1}.keys()) == {'a': 1}
-    assert m.model_dump(exclude={'a': 1}.keys()) == {'b': 2, 'c': 3, 'd': 4, 'e': 5, 'f': 7}
-
-    assert m.model_dump(include={'a': 1}.keys(), exclude_unset=True) == {'a': 1}
-    assert m.model_dump(exclude={'a': 1}.keys(), exclude_unset=True) == {'b': 2, 'e': 5, 'f': 7}
+    # TODO: Do we still want to support the "abstract set" cases?
+    #   Right now, it causes: "TypeError: `include` argument must a set or dict." (and similar for exclude)
+    #   I opened https://github.com/pydantic/pydantic-core/pull/409 to address it, but that also allows list:
```
dmontagu, line 561: Link for convenience: https://github.com/pydantic/pydantic-core/pull/409

tests/test_edge_cases.py
```diff
@@ -570,6 +590,8 @@ class Model(BaseModel):
 
     m = Model(e='e', f=SubModel(c='foo', d=[SubSubModel(a='a', b='b'), SubSubModel(a='c', b='e')]))
 
+    # TODO: Do we still want to support this wrapping behavior on exclude?
+    #   If so, I am happy to implement it in pydantic_core. I got a good ways through but abandoned it until we discuss.
     assert m.model_dump(exclude={'f': {'c': ..., 'd': {-1: {'a'}}}}) == {
```
dmontagu, line 595: Other spot where I wasn't sure if the goal was to support this behavior or not..

tests/test_edge_cases.py
```diff
@@ -1427,31 +1512,35 @@ def validator(v):
...
         (List[Tuple[int, int]], 'List[Tuple[int, int]]'),
-        (Dict[int, str], 'Mapping[int, str]'),
+        (Dict[int, str], 'Dict[int, str]'),
         (FrozenSet[int], 'FrozenSet[int]'),
         (Tuple[int, ...], 'Tuple[int, ...]'),
-        (Optional[List[int]], 'Optional[List[int]]'),
+        (Optional[List[int]], 'Union[List[int], NoneType]'),
         (dict, 'dict'),
-        (DisplayGen[bool, str], 'DisplayGen[bool, str]'),
+        (DisplayGen[bool, str], ('DisplayGen[bool, str]', 'tests.test_edge_cases.DisplayGen[bool, str]')),
```
dmontagu, line 1531: Weirdly the display is different in Python<=3.9 and Python>=3.10, which is why I added two cases here.

Happy to change it back to only having one value and ensuring it's the same across python versions if that's preferable (or at least adding a case that differs for different `sys.version_info` values)

tests/test_edge_cases.py
```diff
@@ -570,6 +590,8 @@ class Model(BaseModel):
 
     m = Model(e='e', f=SubModel(c='foo', d=[SubSubModel(a='a', b='b'), SubSubModel(a='c', b='e')]))
 
+    # TODO: Do we still want to support this wrapping behavior on exclude?
+    #   If so, I am happy to implement it in pydantic_core. I got a good ways through but abandoned it until we discuss.
```
samuelcolvin, line 594: I suspect we probably should, it wouldn't be that hard and would make such migrations much easier.

tests/test_edge_cases.py
```diff
@@ -1306,10 +1377,12 @@ class Model(BaseModel):
         h: Union[Type[FooBar], Sequence[Type[FooBar]]] = FooBar
         i: Union[Type[FooBar], Sequence[Type[FooBar]]] = [FooBar]
 
-    assert Model.model_fields.keys() == {'b', 'c', 'e', 'f', 'g', 'h', 'i'}
+        model_config = dict(arbitrary_types_allowed=True)
+
+    # TODO: Why was this ever not the case? I don't understand why these weren't all fields
```
samuelcolvin, line 1382: because the value of `a` is `int` which meant it got omitted from fields.

But requiring typing annotations on all fields, we simplified the extraction of fields from `namepace` / `attributes` a lot.

tests/test_edge_cases.py
samuelcolvin: i think it's fine for it to be different, but if i'm being nitpicky about the tests, better to do

```
pytest.param(DisplayGen[bool, str], 'DisplayGen[bool, str]', mark=pytest.mark.skipif(sys.version_info < ...))),
```

And then keep the function body as-was.

samuelcolvin (approved):
otherwise LGTM, merge as and when you're happy. I'll review the `__contains__` pr now.

dmontagu:
@samuelcolvin Assuming this passes CI I think it's good to merge.

Issue #5070, opened 2023-02-17
Mypy plugin doesn't consider `allow_population_by_field_name`


jonathanslenders:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

Probably an issue with the Pydantic mypy plugin. It happened after the mypy 1.0.0 release.

Consider this code, and run `mypy --strict`:
```python
from pydantic import BaseModel, Field

class Model(BaseModel):
    field: str = Field(alias="alias")

    class Config:
        allow_population_by_field_name = True

Model(alias="")  # OK.
Model(field="")  # error: Unexpected keyword argument "field" for "Model"  [call-arg]
```

At runtime, this is fine, because of the `allow_population_by_field_name` setting, but the mypy plugin only allows instantiating the class through the alias.

### Python, Pydantic & OS Version

```Text
Python 3.9.13
Pydantic 1.10.5
Ubuntu on ARM, Linux 5.15.0-58
```

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
- [ ] [Data validation/parsing](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://pydantic-docs.helpmanual.io/usage/exporting_models/) - `.model_dump()` and `.model_dump_json()`
- [ ] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)
- [ ] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)
- [X] [Model Config](https://pydantic-docs.helpmanual.io/usage/model_config/)
- [ ] [Field Types](https://pydantic-docs.helpmanual.io/usage/types/) - adding or changing a particular data type
- [ ] [Function validation decorator](https://pydantic-docs.helpmanual.io/usage/validation_decorator/)
- [ ] [Generic Models](https://pydantic-docs.helpmanual.io/usage/models/#generic-models)
[...]

PR #5111, opened 2023-02-24
[cherry-pick] Fix mypy plugin for 1.1.0 (#5077)

cdce8p:
(cherry picked from commit 6267ae3538cf7237ff6769174b4c340415e0f444)

Manual backport to avoid and resolve merge conflict. Alternative to #5110.

### DIFF
--- a/changes/5077-cdce8p.md
+++ b/changes/5077-cdce8p.md
@@ -1,0 +1,1 @@
+Fix mypy plugin for v1.1.0
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -76,6 +76,7 @@ CONFIGFILE_KEY = 'pydantic-mypy'
 METADATA_KEY = 'pydantic-mypy-metadata'
 BASEMODEL_FULLNAME = 'pydantic.main.BaseModel'
 BASESETTINGS_FULLNAME = 'pydantic.env_settings.BaseSettings'
+MODEL_METACLASS_FULLNAME = 'pydantic.main.ModelMetaclass'
 FIELD_FULLNAME = 'pydantic.fields.Field'
 DATACLASS_FULLNAME = 'pydantic.dataclasses.dataclass'
 
@@ -86,6 +87,9 @@ def parse_mypy_version(version: str) -> Tuple[int, ...]:
 
 MYPY_VERSION_TUPLE = parse_mypy_version(mypy_version)
 BUILTINS_NAME = 'builtins' if MYPY_VERSION_TUPLE >= (0, 930) else '__builtins__'
+
+# Increment version if plugin changes and mypy caches should be invalidated
+PLUGIN_VERSION = 1
 
 
 def plugin(version: str) -> 'TypingType[Plugin]':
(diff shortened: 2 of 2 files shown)


KotlinIsland:
I couldn't test this with V2 because I don't know how to install it, but I think this is still broken on this branch (at least with `dataclass`es).

```toml
[tool.poetry]
name = "test-project"
version = "1"
description = ""
authors = []

[tool.poetry.dependencies]
python = "~3.11"
mypy = { git = "https://github.com/python/mypy.git", rev = "release-1.1" }
pydantic = { git = "https://github.com/cdce8p/pydantic", rev = "backport-plugin-fix"}

[tool.mypy]
plugins = ["pydantic.mypy"]
```

```py
import pydantic

class A:
    a: str

```

cdce8p:
> I couldn't test this with V2 because mypy just ran forever, but I think this is still broken on this branch (at least with `dataclass`es).

See https://github.com/pydantic/pydantic/pull/5077#issuecomment-1445703011

cdce8p:
@samuelcolvin This PR is ready. I guess the `change-file-checks` failure is because of the manual backport.

samuelcolvin:
we'll need to rename the change file so the link is right in the changelog, I'll do that when I get time.

cdce8p:
> we'll need to rename the change file so the link is right in the changelog, I'll do that when I get time.

I can do it, no problem. However, as it isn't renamed by the backport bot, I though it would be good here. It's basically the same PR as #5110, just without the merge conflict.
https://github.com/pydantic/pydantic/pull/5110/files#diff-7346c71569c422ee0520a5f2de506f6e959d587acb20aa72f90be6db4dfa7d82R1

samuelcolvin:
I guess we could either extend the backport bot to rename the change file, or change hooky to recognise the change file, but for now it, I think it's easiest just to rename it manually.

dmontagu:
Oh actually, @cdce8p could you check and make sure that with this branch there isn't anything else that could/should be fixed that I haven't already discussed in my previous comment?

cdce8p:
> Oh actually, @cdce8p could you check and confirm that, with this branch, there isn't anything else that could/should be fixed that I haven't already discussed in my previous comment?

AFAICT those are the same observations I saw during testing. So it should be good.
I saw one other issue with the mypy plugin, in particular with `allow_population_by_field_name` #5070. This can probably be address at a later point though as it's not related to mypy `1.1.1`.

> the only thing I'm not sure of is if there are any concerns with adding the `import dataclasses` line, as discussed here: https://github.com/pydantic/pydantic/pull/5120/files#r1118839536. It seems maybe this approach was agreed on as acceptable but I just want to confirm before merging.

Some background, this was still an open question in #5120. However, I noticed that other files already import `dataclasses`, so importing it in `pydantic.dataclasses` shouldn't effect the import time as it's already loaded. At least that's the case for `v2`. Searching `v1.10.x` it looks as if it would indeed be an additional import unfortunately. Not sure though it's worth changing it (and if we do, how).

dmontagu:
Okay, in the interest of keeping things simple how about I just change it back to the imports being the same. I think in that case it will be non-controversial.

dmontagu:
Okay actually I take it back, I see we are actually using it in a more meaningful way than I expected.

Considering how fast `dataclasses` is to import, I think it should be okay, but I'll let @samuelcolvin make the final call.

cdce8p:
@dmontagu Thanks for the quick release! Just confirmed with Home Assistant, everything is working as expected (when using the plugin).

Issue #5016, opened 2023-02-06
Add nice error message when attempting to import removed modules/functions in v2


dmontagu:
If a user attempts to import removed functions, from removed modules, or both, we should give a better error message than a plain ImportError. Ideally the error message will include some information about the migration path, or possibly a request to create an issue if the changes break something in their codebase in a way that is difficult for them to resolve.

As an example, it would be preferable to get such an error message when attempting to run `from pydantic.generics import replace_types`.

Probably anything that could be construed as part of the "public" api of v1.X (i.e., doesn't use leading underscores in the name or import path) should be handled neatly.

ghandic:
Keen to work on this! Will require a fair but of discussion on how we wish to inform the user, how long this will be persisted for etc.

I have an initial idea (see PR https://github.com/pydantic/pydantic/pull/5113) 


Using this PEP-562 standard, we could add a `__getattr__` method to all files (and the old file names - noting this will be all that is in them) 

Let me know your thoughts :)

ghandic:
We could also monkeypatch internals, this seems like a nice way to do it, but potentially risky... 

```shell
Traceback (most recent call last):
  File "/projects/pydantic/caller.py", line 1, in <module>
    from pydantic import Required
  File "/projects/pydantic/import_helper.py", line 30, in _handle_fromlist_override
    raise exception
pydantic.import_helper.V2MigrationRemoved: pydantic.Required has been removed in favour of ...
```

ghandic:
@samuelcolvin @dmontagu thoughts?

PR #5113, opened 2023-02-26
Add nice error message when attempting to import removed modules/functions in v2

ghandic:
Working towards #5016


samuelcolvin:
Please change the PR body to remove the default text and mention the issue you're closing.

ghandic:
Just an initial PoC, but this is what it looks like:

```shell
  File "/projects/pydantic/caller.py", line 1, in <module>
    from pydantic import Required
  File "<frozen importlib._bootstrap>", line 1075, in _handle_fromlist
  File "/projects/pydantic/__init__.py", line 115, in __getattr__
    inform(__name__, name)
  File "/projects/pydantic/import_helper.py", line 28, in inform
    raise exception
pydantic.import_helper.V2MigrationRemoved: pydantic.Required has been removed in favour of ...
```

caller.py
```diff
@@ -0,0 +1 @@
+from pydantic import Required
```
samuelcolvin, line 1: This file shouldn't be here.

pydantic/import_helper.py
```diff
@@ -0,0 +1,33 @@
+class V2MigrationRemoved(Exception):
+    ...
+
+
+class V2MigrationMoved(Exception):
```
samuelcolvin, line 5: If stuff is moved, it should continue to work, albeit with a warning.

pydantic/import_helper.py
```diff
@@ -0,0 +1,33 @@
...
+
+V2_MIGRATION_MAPPING = {
+    # Case: Removed
+    "pydantic.Required": V2MigrationRemoved("pydantic.Required has been removed in favour of ..."),
+    # Case: Moved path
+    # Case: Renamed object
+    # Case: Superseded
+}
+
+import importlib
```
samuelcolvin, line 25: I think clearer to just implement `__getattr__` in `__init__.py`.

samuelcolvin (commented):
Looks like a good start.

pydantic/import_helper.py
ghandic: The only was I could get `__getattr__` to work was on each file that was being accessed, so `from pydantic import Required` would be fine in `__init__.py` but `from pydantic.generics import replace_types` would require the `__getattr__` in `pydantic/generics.py` too

caller.py
ghandic: Will be moved into tests after initial feedback :)

pydantic/import_helper.py
```diff
@@ -0,0 +1,33 @@
...
+___handle_fromlist__ = importlib._bootstrap._handle_fromlist
```
samuelcolvin, line 26: I don't think this needs a dunder name.

pydantic/import_helper.py
```diff
@@ -0,0 +1,33 @@
...
+
+def _handle_fromlist_override(module, fromlist, import_, *, recursive=False):
+    if exception := V2_MIGRATION_MAPPING.get(f"{module.__name__}.{'.'.join(fromlist)}"):
```
samuelcolvin, line 29: We can't use the walrus operator, we have to support 3.7

caller.py
samuelcolvin: Best to start tests and not add this to the PR.

pydantic/import_helper.py
samuelcolvin: I think so this should go into a function and be called in `__init__.py`, not run on module import.

I'm still not sure about this route, is it supported by all supported versions and pypy?

ghandic: I'll go towards how it is now and start building up the scenarios, standards for messages etc and then can revisit this part if its not ideal. However, with the way it is at the moment it is isolated which has its advantages

samuelcolvin (commented):
B ready to get some tests added and ci passing, then we can review.

ghandic:
@samuelcolvin do you have tests that run against pypy?

samuelcolvin:
Only pydantic-core at the moment.

ghandic:
Looks like its passing - other tests dont currently pass on pypy though

samuelcolvin:
Okay, well adding pypy to tests will be more work than this, but good it works for this feature.

Remove pypy for now and we can add it in a separate pr.

pydantic/__init__.py
```diff
@@ -1,3 +1,5 @@
+import importlib
```
samuelcolvin, line 1: is there a particular reason this needs to be imported here?

pydantic/import_helper.py
```diff
@@ -0,0 +1,382 @@
+import warnings
+from importlib import import_module
+from types import ModuleType
+from typing import Any, Tuple
+
+
+class V2MigrationRemovedException(Exception):
+    ...
```
samuelcolvin, line 8: can we add docstings to these.

pydantic/import_helper.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+class V2MigrationSuperseded(Exception):
+    ...
+
+
+V2_MIGRATION_MAPPING = {
+    # Case: Removed
+    'pydantic.Required': V2MigrationRemovedException('pydantic.Required has been removed in favour of ...'),
+    'pydantic.create_model_from_namedtuple': V2MigrationRemovedException('TODO'),
```
samuelcolvin, line 26: I guess we can use a generic message for most of these.

pydantic/import_helper.py
```diff
@@ -0,0 +1,382 @@
...
+    ...
+
+
+class V2MigrationSuperseded(Exception):
+    ...
+
+
+V2_MIGRATION_MAPPING = {
+    # Case: Removed
+    'pydantic.Required': V2MigrationRemovedException('pydantic.Required has been removed in favour of ...'),
```
samuelcolvin, line 25: looking good. Did you use a script to find/generate these? If so might be worth including it in the PR in case it's needed again.

pydantic/import_helper.py
```diff
@@ -0,0 +1,382 @@
...
+    'pydantic.types.JsonWrapper': V2MigrationMovedWarning('TODO'),
+    'pydantic.types.JsonMeta': V2MigrationMovedWarning('TODO'),
+    'pydantic.types.ConstrainedDate': V2MigrationMovedWarning('TODO'),
+    # pydantic.typing # TODO
+    # pydantic.utils # TODO
+    # pydantic.validators # TODO
+    # pydantic.version
+    'pydantic.version.compiled': V2MigrationMovedWarning('TODO'),
+    # Case: Moved path
+    'pydantic.error_wrappers.ValidationError': (
```
samuelcolvin, line 307: for the moved cases we should raise a warning, but return the new correct thing.

samuelcolvin (commented):
looking good, I think this will be really helpful when migrating.

A few more things to do.

Please update.

pydantic/__init__.py
ghandic: For `patch_importlib_with_migration_info(importlib)` it needs the importlib to be in the root of the package unfortunately

pydantic/import_helper.py
ghandic: Not that sophisticated 😄 gave it a go, but didnt want to sink too much time into it... So I just looked at the git diff between each of the versions.

pydantic/import_helper.py
```diff
@@ -0,0 +1,382 @@
...
+    'pydantic.fields.ModelField': V2MigrationMovedWarning('TODO'),
+    'pydantic.fields.DeferredType': V2MigrationMovedWarning('TODO'),  # Moved
+    'pydantic.fields.is_finalvar_with_default_val': V2MigrationMovedWarning('TODO'),
+    # pydantic.main
+    'pydantic.main.validate_custom_root_type': V2MigrationMovedWarning('TODO'),
+    'pydantic.main.generate_hash_function': V2MigrationMovedWarning('TODO'),
+    'pydantic.main.ANNOTATED_FIELD_UNTOUCHED_TYPES': V2MigrationMovedWarning('TODO'),
+    'pydantic.main.UNTOUCHED_TYPES': V2MigrationMovedWarning('TODO'),
+    'pydantic.main.validate_model': V2MigrationMovedWarning('TODO'),
+    # pydantic.mypy NOTE: ERROR_ORM still exists, should it be renamed to ERROR_ATTRIBUTES
```
ghandic, line 228: @samuelcolvin for this, while looking through v1-> v2 I noticed that ERROR_ORM still exists, should it be renamed to ERROR_ATTRIBUTES during the migration ?

ghandic: Still present, just ignored at the moment in the tests, let me know your preference 👍🏼

samuelcolvin: Not really sure - I think we might just have not got there yet.

samuelcolvin:
replaced by #5480

ghandic:
Would have been good to have a discussion in open forum, feels like a waste of time duplicating this work.

Issue #5115, opened 2023-02-26
Add pypy to tests for V2. 


samuelcolvin:
See #5113.

ghandic:
Might be blocked on supporting pypy - due to dependence on [typed_ast](https://github.com/python/typed_ast) 

`Due to reliance on certain C APIs, this library does not build on and there are no plans to support PyPy.`

ghandic:
Took over 1 hour 20 for the tests to run on pypy 3.8/3.9 but they did pass excluding the known disabled issues that need more investigation

samuelcolvin:
fixed in #5133

PR #5116, opened 2023-02-27
Add pypy to tests for V2

ghandic:
Working towards #5115


NOTES.md
```diff
@@ -0,0 +1,12 @@
+# Problems with Pypy
```
samuelcolvin, line 1: it really doesn't help to create files like this.

If you have notes to make, please put them in on the PR.

samuelcolvin:
thanks so much for looking into this. I think it'll probably be easiest if I work on this as I know the changes which will likely be required.

PR #5120, opened 2023-02-27
Fix `dataclass_transform` decorator for pydantic dataclasses

### LINKED ISSUE #5117: (🐞) Mypy plugin doesn't work with positional arguments on `dataclasses`
On main [6267ae3](https://github.com/pydantic/pydantic/commit/6267ae3538cf7237ff6769174b4c340415e0f444)
```py
import pydantic

class A:
    a: str

```

See this: https://github.com/pydantic/pydantic/pull/5077#issuecomment-1445703011

cdce8p:
Remove `kw_only_default` and add additional `field_specifier` for stdlib dataclass fields.

This change should be backported to `1.10.X` since the next mypy release will break typing otherwise.

Ref https://peps.python.org/pep-0681/#dataclass-transform-parameters
Fixes #5117

### DIFF
--- a/changes/5120-cdce8p.md
+++ b/changes/5120-cdce8p.md
@@ -1,1 +1,1 @@
-Fix dataclass_transform decorator for pydantic dataclasses
+Fix `dataclass_transform` decorator for pydantic dataclasses
--- a/changes/5120-cdce8p.md
+++ b/changes/5120-cdce8p.md
@@ -1,0 +1,1 @@
+Fix dataclass_transform decorator for pydantic dataclasses
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -31,7 +31,6 @@ This means we **don't want to create a new dataclass that inherits from it**
 The trick is to create a wrapper around `M` that will act as a proxy to trigger
 validation without altering default `M` behaviour.
 """
-import dataclasses
 import sys
 from contextlib import contextmanager
 from functools import wraps
@@ -60,6 +59,8 @@ from .fields import Field, FieldInfo
 from .main import create_model
 
 if TYPE_CHECKING:
+    import dataclasses
+
     from .main import BaseModel
 
     NoArgAnyCallable = Callable[[], Any]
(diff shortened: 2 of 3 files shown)


samuelcolvin:
Not sure if this will conflict with #5009 which is a WIP implementation of dataclasses for V2 which I'm intending to complete this week?

cdce8p:
> Not sure if this will conflict with #5009 which is a WIP implementation of dataclasses for V2 which I'm intending to complete this week?

Just looked at it, as long as the public interface isn't changed (i.e. `dataclasses.field` and positional arguments still works) it shouldn't conflict. It might even be slightly helpful if you add a pydantic specific `field` method. That could / should be added to `dataclass_transform` then as well.

If you really want though, I could also change the target to `1.10.X`.

changes/5120-cdce8p.md
```diff
@@ -0,0 +1 @@
+Fix dataclass_transform decorator for pydantic dataclasses
```
samuelcolvin, line 1:
```suggestion
Fix `dataclass_transform` decorator for pydantic dataclasses
```

pydantic/dataclasses.py
```diff
@@ -31,6 +31,7 @@ class M:
 The trick is to create a wrapper around `M` that will act as a proxy to trigger
 validation without altering default `M` behaviour.
 """
+import dataclasses
```
samuelcolvin, line 34: due to import time problems in the past, we avoid doing module level import of stuff unless it's absolutely necessary.

I guess if we remove `dataclass` from `__init__.py`, maybe this would be fine, but we'll need to change this or on `v1.10`.

Not sure which you prefer?

samuelcolvin (commented):
otherwise LGTM.

pydantic/dataclasses.py
cdce8p: Found another way.  I moved the overloads and the import into `if TYPE_CHECKING` blocks.

It's actually enough to have one `dataclass_transform` decorator per overload case, even undefined behavior to have multiple. With the unnecessary ones removed, it still seem to works (tested with mypy and pyright).
https://peps.python.org/pep-0681/#undefined-behavior

pydantic/dataclasses.py
```diff
@@ -103,76 +105,74 @@ def __validate__(cls: Type['DataclassT'], v: Any) -> 'DataclassT':
...
-        frozen: bool = False,
-        config: Union[ConfigDict, Type[object], None] = None,
-        validate_on_init: Optional[bool] = None,
-    ) -> 'DataclassClassOrWrapper':
-        ...
-
-
-@dataclass_transform(kw_only_default=True, field_specifiers=(Field, FieldInfo))
+if TYPE_CHECKING:
+    if sys.version_info >= (3, 10):
```
samuelcolvin, line 109: sorry if I'm being dumb, why do we need different logic for 3.10 and before?

cdce8p: `3.10` adds an additional argument `kw_only`.

pydantic/dataclasses.py
```diff
@@ -103,76 +105,74 @@ def __validate__(cls: Type['DataclassT'], v: Any) -> 'DataclassT':
...
+
+        @dataclass_transform(field_specifiers=(dataclasses.Field, dataclasses.field, Field, FieldInfo))
```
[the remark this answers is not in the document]
cdce8p: Not sure about that tbh. The PEP mentioned both `dataclasses.Field` and `dataclasses.field` so I figured to use all four here.

> The standard dataclass behavior supports only one type of field specifier called Field plus a helper function (field) that instantiates this class, so if we were describing the stdlib dataclass behavior, we would provide the tuple argument (dataclasses.Field, dataclasses.field).

https://peps.python.org/pep-0681/#dataclass-transform-parameters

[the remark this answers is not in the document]
cdce8p: Removed `dataclassed.Field` and `FieldInfo`, after checking mypy and pyright. It's enough to specifier the classes / callables users should actually use.

pydantic/dataclasses.py
```diff
@@ -103,76 +105,74 @@ def __validate__(cls: Type['DataclassT'], v: Any) -> 'DataclassT':
...
+            order: bool = False,
+            unsafe_hash: bool = False,
+            frozen: bool = False,
+            config: Union[ConfigDict, Type[object], None] = None,
+            validate_on_init: Optional[bool] = None,
+        ) -> 'DataclassClassOrWrapper':
+            ...
+
+
 def dataclass(
```
KotlinIsland, line 176: Should the implementation keep the `dataclass_transform` for runtime introspection?

cdce8p: In that case we would need to import stdlib `dataclasses` at runtime. The goal was to avoid just that.
Unfortunately, it doesn't seem to be valid to use forward references as value for `field_specifiers`. Although mypy and pyright seems to know how to handle it already 🤔

Not sure how to processed here if I'm honest. I could take another look at the import situation but probably not until later this week unfortunately.

cdce8p:
> due to import time problems in the past, we avoid doing module level import of stuff unless it's absolutely necessary.
> 
> I guess if we remove `dataclass` from `__init__.py`, maybe this would be fine, but we'll need to change this or on `v1.10`.

`dataclasses` is already imported in other modules, e.g. `networks.py`. So it won't add any additional import time.

https://github.com/pydantic/pydantic/blob/6ab42070483907d70e62c2b2fea37832e7c2748d/pydantic/networks.py#L3

> Should the implementation keep the `dataclass_transform` for runtime introspection?

As the import time is no longer a concern, I reverted the change. With that `dataclass_transform` will be accessible at runtime again.

dmontagu:
Closing in favor of #5111

dmontagu:
Oh I realize this is for main, not 1.10.X, so I'll reopen it.

cdce8p:
I think it would make sense to move forward with this PR as well now that the changes have already been merged into `1.10.X`. (A backport is obviously no longer necessary). /CC: @dmontagu

dmontagu:
Thanks @cdce8p!

PR #5125, opened 2023-03-01
Add generics support for v2 (another attempt)

### LINKED ISSUE #5146: add `memray` tests
Probably here and in pydantic-core.

Do we add checks to all tests, or just a few tests designed specifically?

dmontagu:
Note: this requires an updated version of `pydantic_core`, but is currently passing all tests against `pydantic_core`'s main branch.

There is still one major failing test I haven't resolved yet, which is getting it to work properly with "deep" recursive generic models. Figuring that out is my goal for tomorrow, but I wanted to open this now for the purposes of review.

@samuelcolvin you mentioned that the dataclasses stuff might blow this up, I'll be happy to merge it all together when that's ready.

fix #5146


pydantic/_internal/_fields.py
```diff
@@ -45,11 +49,63 @@ def __init__(self, **metadata: Any):
...
+        updated_actions = cls.actions + [{'kind': 'class_getitem', 'item': item}]
+
+        class SelfType(cls):  # type: ignore[valid-type,misc]
+            actions = updated_actions
+
+        return SelfType
+
+    @classmethod
+    def replace_types(cls, typevars_map: Any) -> type[BaseSelfType]:
+        updated_actions = cls.actions + [{'kind': 'replace_types', 'typevars_map': typevars_map}]
```
dmontagu, line 81: Note to self -- I need to give these "actions" a `TypedDict` type.

pydantic/_internal/_core_utils.py
```diff
@@ -0,0 +1,195 @@
...
+    However, as an implementation detail, recursive generic models will emit a non-identical schema deeper in the
+    tree with a re-used ref, with the intent that that schema gets replaced with a recursive reference once the
+    specific generic parametrization to use can be determined.
+    """
+    refs = set()
+
+    def _replace_refs(s: core_schema.CoreSchema) -> core_schema.CoreSchema:
+        # Without the cast, get:
+        # pydantic/_internal/_core_utils.py:22: error: Incompatible types in assignment
+        # (expression has type "object", variable has type "Optional[str]")  [assignment]
```
dmontagu, line 29: Ignore this comment, it is out of date. (Note to self: need to delete this)

tests/test_generics.py
```diff
@@ -180,75 +208,159 @@ class Result(GenericModel, Generic[T]):
...
+    class MyGenericModel(BaseModel, Generic[T]):
+        x: T
+
+    Model = MyGenericModel[int]
+    assert len(GENERIC_TYPES_CACHE) == types_cache_size + 2
+    del Model
+    gc.collect()
+
+    # TODO: Need to figure out why this is failing
+    assert len(GENERIC_TYPES_CACHE) == types_cache_size
```
dmontagu, line 325: It seems the WeakValuesDictionary approach to the types cache (from #5052) isn't evicting items as expected. Definitely would appreciate any help investigating this. @adriangb

adriangb: In #5052 we validated it by creating a ton of generics and measuring memory usage, not by checking the size of the cache. I'm a bit surprised by your tests behavior but not _too_ surprised, I don't know if it's easy to measure these things at this fine of a level.

I did a sanity check:

```python
import gc
from typing import Any, Generic, TypeVar
from weakref import WeakValueDictionary

T = TypeVar("T")

class MyGeneric(Generic[T]):
    pass

Concrete = MyGeneric[int]

cache: WeakValueDictionary[Any, Any] = WeakValueDictionary()

cache[(MyGeneric, int)] = Concrete

del Concrete

gc.collect(0)
gc.collect(1)
gc.collect(2)

assert len(cache) == 0  # actually 1
```

So it seems this would not work in the first place!

adriangb: Ok I think I found the issue:

```python
import gc
from typing import Any, Generic, TypeVar
from weakref import WeakValueDictionary

T = TypeVar("T")

class MyGeneric(Generic[T]):
    pass

cache: WeakValueDictionary[Any, Any] = WeakValueDictionary()

Concrete = MyGeneric[int]
cache[(MyGeneric, int)] = Concrete
del Concrete

gc.collect(0)
gc.collect(1)
gc.collect(2)

referrer = next(iter(gc.get_referrers(next(cache.values()))))
print(referrer)  # <functools._lru_cache_wrapper object at 0x102dee560>
print(referrer.__module__)  # typing
```

So it looks like this cache inside of `_tp_cache` of `typing.py` may be holding onto a reference.
It has a `maxsize` of 128, which basically means that the `WeakValueDict` will act as a regular dict for the _last_ 128 things that got run through it!

That explains why the test that creates thousands of objects and measures memory consumption looks like it's not accumulating but this one measuring just 1 reference does.

adriangb: Another little test:

```python
import gc
from typing import Any, Generic, TypeVar
from weakref import WeakValueDictionary

T = TypeVar("T")

class MyGeneric(BaseModel, Generic[T]):
    x: T

cache: WeakValueDictionary[Any, Any] = WeakValueDictionary()

def create_and_del() -> None:
    Concrete = MyGeneric[int]
    cache[(MyGeneric, int)] = Concrete
    del Concrete

create_and_del()
gc.collect(0)
gc.collect(1)
gc.collect(2)

for r in gc.get_referrers(next(cache.values())):
    print(r, end="\n\n")
```

```
(<class '__main__.MyGeneric[int]'>, <class '__main__.MyGeneric'>, <class 'pydantic.main.BaseModel'>, <class 'pydantic._internal._repr.Representation'>, <class 'typing.Generic'>, <class 'object'>)

{'type': 'model', 'cls': <class '__main__.MyGeneric[int]'>, 'schema': {'type': 'typed-dict', 'fields': {'x': {'schema': {'type': 'int', 'metadata': {}}, 'required': True, 'metadata': {'pydantic_json_schema_metadata': {'title': None, 'description': None, 'examples': None, 'extra_updates': None}}}}, 'return_fields_set': True, 'ref': '__main__.MyGeneric:5611558880[int:4315709880]'}, 'config': {'title': 'MyGeneric[int]', 'typed_dict_extra_behavior': 'ignore', 'allow_inf_nan': True, 'populate_by_name': False, 'str_strip_whitespace': False, 'str_to_lower': False, 'str_to_upper': False, 'strict': False, 'ser_json_timedelta': 'iso8601', 'ser_json_bytes': 'utf8'}, 'metadata': {'pydantic_json_schema_metadata': {'title': 'MyGeneric[int]', 'description': None}}}
```

The first one seems like the MRO? The second is a pydantic-core schema.

dmontagu: > I'm a bit surprised by your tests behavior but not too surprised, I don't know if it's easy to measure these things at this fine of a level.

Just to be clear, I took this test directly from the v1 PR, so it should be working there.

samuelcolvin:
I've had a play and found an issue

```py
from devtools import debug


T = TypeVar("T")


class MyGenericModel(BaseModel, Generic[T]):
    foobar: list[MyGenericModel[T]] | None
    spam: T

debug(MyGenericModel[int](foobar=[{'foobar': [], 'spam': 1}], spam=1))
"""
> test.py:14 <module>
    MyGenericModel[int](foobar=[{'foobar': [], 'spam': 1}], spam=1): MyGenericModel[int](
        foobar=[
            (
                {
                    'foobar': [],
                    'spam': 1,
                },
                {
                    'spam',
                    'foobar',
                },
            ),
        ],
        spam=1,
    ) (MyGenericModel[int])
"""
```
E.g. the inter reference to `MyGenericModel` is pointing to the inner typed-dict, but is not being wrapped in a model

`debug(MyGenericModel.__pydantic_core_schema__)` is giving:

```py
    {
        'type': 'model',
        'cls': <class '__main__.MyGenericModel'>,
        'schema': {
            'type': 'typed-dict',
            'fields': {
                'foobar': {
                    'schema': {
                        'type': 'nullable',
                        'schema': {
                            'type': 'list',
                            'items_schema': {
                                'type': 'definition-ref',
                                'schema_ref': '__main__.MyGenericModel:5149425584',
                                'metadata': {},
                            },
                            'metadata': {},
                        },
                        'metadata': {},
                    },
                    'required': True,
                    'metadata': {
                        'pydantic_json_schema_metadata': {
                            'title': None,
                            'description': None,
                            'examples': None,
                            'extra_updates': None,
                        },
                    },
                },
                'spam': {
                    'schema': {
                        'type': 'any',
                        'metadata': {},
                    },
                    'required': True,
                    'metadata': {
                        'pydantic_json_schema_metadata': {
                            'title': None,
                            'description': None,
                            'examples': None,
                            'extra_updates': None,
                        },
                    },
                },
            },
            'return_fields_set': True,
            'ref': '__main__.MyGenericModel:5149425584',
        },
        'config': {
            'title': 'MyGenericModel',
            'typed_dict_extra_behavior': 'ignore',
            'allow_inf_nan': True,
            'populate_by_name': False,
            'str_strip_whitespace': False,
            'str_to_lower': False,
            'str_to_upper': False,
            'strict': False,
            'ser_json_timedelta': 'iso8601',
            'ser_json_bytes': 'utf8',
        },
        'metadata': {
            'pydantic_json_schema_metadata': {
                'title': 'MyGenericModel',
                'description': None,
            },
        },
    }
```

I'll look into it and see if I can fix it (mostly to force me to try and understand it).

pydantic/_internal/_generics.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+_GENERIC_TYPES_CACHE = GenericTypesCache()
+
+
+def _cache_key(cls: type[BaseModel], _params: Any) -> GenericTypesCacheKey:
+    # TODO: What is happening with args here?
+    #   Note: get_args always returns () if _params is a tuple, and
+    #   if _params is a type, _params and (_params,) should have the same meaning as arguments to __class_getitem__.
+    #   So something seems suspicious..
```
dmontagu, line 268: @samuelcolvin I'd be interested in any insight you might have into what I'm bringing up in this comment

pydantic/_internal/_repr.py
```diff
@@ -101,6 +101,13 @@ def display_as_type(obj: Any) -> str:
     elif isinstance(obj, Representation):
         return repr(obj)
 
+    if isinstance(obj, str):
+        # TODO: I'm hitting this for generic forward references. It used to return 'str' in this case. Affects:
+        #   * tests/test_json_schema.py::test_nested_generic
+        #   * tests/test_utils.py::test_display_as_type[foobar-str]
+        #   * tests/test_utils.py::test_display_as_type_310[<lambda>-str1]
+        return obj
```
dmontagu, line 109: @samuelcolvin Also interested to get your eyes on this

pydantic/main.py
```diff
@@ -99,6 +121,19 @@ def hash_func(self_: Any) -> int:
...
 
+            # TODO: cls.__concrete__ has been removed -- do we need to retain this for v1 compatibility?
+            # TODO: cls.__parameters__ may be removed -- do we need to retain this for v1 compatibility?
+            cls.__pydantic_generic_origin__ = __pydantic_generic_origin__
+            cls.__pydantic_generic_args__ = __pydantic_generic_args__
+            cls.__pydantic_generic_parameters__ = __pydantic_generic_parameters__ or getattr(
+                cls, '__parameters__', None
+            )
+            cls.__pydantic_deleted_attrs__ = {}  # TODO: yuck, get rid of this
+            cls.__pydantic_generic_typevars_map__ = (
```
dmontagu, line 132: Note to self: probably makes sense to remove this attribute since it is computed from the others.

.github/workflows/ci.yml
```diff
@@ -89,6 +89,40 @@ jobs:
         name: docs
         path: site
 
+# TODO: Get this working once the memory leak issue is resolved - https://github.com/pydantic/pydantic/issues/5136
```
dmontagu, line 92: I can confirm the memory leak issue is solved and the memray test is passing for me locally; not sure exactly how this should be modified for compatibility with v2.

pydantic/_internal/_generics.py
```diff
@@ -0,0 +1,327 @@
...
+        _GENERIC_TYPES_CACHE[_early_cache_key(parent, typevar_values[0])] = type_
+    if origin and args:
+        _GENERIC_TYPES_CACHE[_late_cache_key(origin, args, typevar_values)] = type_
+
+
+def get_cached_generic_type_early(parent: type[BaseModel], typevar_values: Any) -> type[BaseModel] | None:
+    return _GENERIC_TYPES_CACHE.get(_early_cache_key(parent, typevar_values))
+
+
+def get_cached_generic_type_late(
```
dmontagu, line 321: If you find this crazy _early and _late cache key business hard to understand, I can explain in more detail (probably makes sense to add comments to these functions but wanted to see if you ended up wanting to simplify it first).

This approach was necessary to get `Model[List[T]][int]` to be the same type as `Model[List[int]]`. It could be modified to not use two different cache keys at different points, but the `_early_cache_key` is optimized to be as quick to compute as possible (for repeated-access speed), and the `_late_cache_key` is optimized to be as "correct" as possible, so that two types that are ultimately the same after resolving the type arguments will always produce cache hits.

pydantic/_internal/_generics.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+def recursively_defined_type_refs() -> set[str]:
+    return set((_visit_counts_context.get() or {}).keys())
+
+
+_GENERIC_TYPES_CACHE = GenericTypesCache()
+
+
+def _union_orderings_key(typevar_values: Any) -> Any:
```
dmontagu, line 264: I did some digging into what was going on with the calls to get_args in the v1 cache key logic, and as far as I could tell the only thing it was really doing for us was helping us differentiate between the ordering of items in Unions (e.g., `Union[int, float]` vs. `Union[float, int]`). (It's definitely possible I'm missing other cases where it's relevant.)

I have pulled out this function to do exactly this check, and more "correctly" by applying to each item in a possible tuple (which is not currently done in v1), and ignoring the args for other types (besides Union) where order is not ignored and therefore not important (and differences in the result of get_args for other origin types can cause spurious cache misses).

Let me know if you know of any cases that also need handling besides when the origin is Union.

pydantic/main.py
dmontagu: I changed my mind as it is reused in a few places. Happy to revisit later.

pydantic/_internal/_repr.py
dmontagu: I'm optimistically removing this comment from the PR to get CI run through with the comment removed, in the hopes that you agree with the change. To make it easier to review, the comment that used to be above this line said:
```
        # TODO: I'm hitting this for generic forward references. It used to return 'str' in this case. Affects:
        #   * tests/test_json_schema.py::test_nested_generic
        #   * tests/test_utils.py::test_display_as_type[foobar-str]
        #   * tests/test_utils.py::test_display_as_type_310[<lambda>-str1]
```

dmontagu: I'll also note that if for any reason we _don't_ want to make this change to this function, I can change the `model_parametrized_name` method to have `param_names = [param if isinstance(param, str) else _repr.display_as_type(param) for param in params]`

pydantic/_internal/_generics.py
samuelcolvin: humm, but union or der can be important in pydantic, are we sure it's correct to ignore order?

pydantic/_internal/_generics.py
samuelcolvin: happy with two, let's add this to explanation to a docstring.

pydantic/_internal/_repr.py
samuelcolvin: happy to return here, but I wonder if we should return `f'ForwardRef({obj!r})'`?

Advtanges:
* it's hopefully clearer what we think is going on
* It would better match the behaviour of `List['SomeForwardRef']` which becomes `"List[ForwardRef('SomeForwardRef')]"` as per the test in `test_utils.py`

samuelcolvin (commented):
A few comments but on the whole I think we should merge this asap.

I still believe that `generic_recursion_self_type` and friends could be simplified if we adopt something like #5149. However:
1. I don't pretend to know enough to be sure
2. If those simplifications aren't possible, it's fine
3. this shouldn't block merging

pydantic/_internal/_generics.py
samuelcolvin: I don't think this should block merging - we can reconsider this later.

dmontagu: This is handling things the same as v1 currently, as far as I can tell. If something is now mishandled, I think it will be related to arguments in _non_-union types. There are no unit tests indicating this can cause issues so I'm inclined to leave like this for now but we can obviously fix it if we discover a use case that isn't addressed (and it would be good to add a unit test for it anyway).

dmontagu:
> I still believe that generic_recursion_self_type and friends could be simplified if we adopt something like https://github.com/pydantic/pydantic/pull/5149. However:

I agree it's hard to reason about, so I also don't pretend to be confident, but I think the main thing that has the potential to simplify the complex recursion stuff would be if the `BaseModel`'s validator was a `ModelValidator` rather than `TypedDictValidator`, as we discussed. Most of the weird have-to-go-deeper stuff comes from the fact that at the top level of the schema, you don't have a model_schema you can refer to. when building the inner typed_dict_schema.

PR #5126, opened 2023-03-01
Use getattr and default for qualname

### LINKED ISSUE #5101: Attribute error for `functools.partial`

JensHeinrich:
Was suggested in python/cpython#102323 by @sobolevn. Fixes #5101
Superceedes #5102


pydantic/class_validators.py
```diff
@@ -147,7 +147,11 @@ def _prepare_validator(function: AnyCallable, allow_reuse: bool) -> 'AnyClassMet
     """
     f_cls = function if isinstance(function, classmethod) else classmethod(function)
     if not in_ipython() and not allow_reuse:
-        ref = f_cls.__func__.__module__ + '.' + f_cls.__func__.__qualname__
+        ref = (
+            getattr(f_cls.__func__, '__module__', '<default_module>')
+            + '.'
+            + getattr(f_cls.__func__, '__qualname__', '<default name>')
```
dmontagu, line 153: Because uniqueness is required, and `__qualname__` was used here in part to provide uniqueness, I would replace `<default_name>` with something like `f'id:{id(f_cls.__func__)}'`

pydantic/class_validators.py
```diff
@@ -165,14 +169,14 @@ def get_validators(self, name: str) -> Optional[Dict[str, Validator]]:
         if name != ROOT_KEY:
             validators += self.validators.get('*', [])
         if validators:
-            return {v.func.__name__: v for v in validators}
+            return {getattr(v.func, '__name__', '<default name>'): v for v in validators}
```
dmontagu, line 172: Again, I'd try to return something that would be a bit more unique, using id seems reasonable to me if there is no `__name__`.

dmontagu:
Seems reasonable to me with the caveats above. Defer to @samuelcolvin on whether there might be any other reasons to be concerned about merging

pydantic/class_validators.py
JensHeinrich: TBH: I think less uniqueness is not that bad, when we are checking for uniqueness; at worst it will mark a function as duplicate that isn't; this should prevent the problem this check has been created for

dmontagu: The problem is it would raise an error any time you use two or more partials as validators, regardless of their relationship, unless you modify the model config to allow that. I don't have a particularly strong opinion on this, but I think using the `id` is more representative of current behavior, as things with a `__qualname__` will generally have the _same_ `__qualname__` if and only if they have the same `id`. (Maybe there are exceptions to that, but if there are, it's probably not "normal" usage.)

JensHeinrich: That's the reason I asked about allowing those only with `allow_reuse`.

~TBH I worry more on the other case~ That makes no sense as we are only using the id as fallback

PR #5128, opened 2023-03-01
Mark optional params as Optional

JensHeinrich:
Mark parameters as Optional if they are
skip change file check


JensHeinrich:
Yes, those annoy me because my editor is too angry

JonathanPlasse:
You should remove `implicit_optional = true` from `mypy` configuration.

JensHeinrich:
> You should remove `implicit_optional = true` from `mypy` configuration.

Pretty sure that won't work as my editor is using `pyright`

JensHeinrich:
> > You should remove `implicit_optional = true` from `mypy` configuration.
> 
> Pretty sure that won't work as my editor is using `pyright`

And I do like to do things correctly

JonathanPlasse:
It should be removed because you are removing implicit optional.
This is present because there were implicit optional.

JensHeinrich:
> It should be removed because you are removing implicit optional. This is present because there were implicit optional.

I just had a look for this option, and according to [this article](https://adamj.eu/tech/2022/10/18/python-type-hints-implicit-optional-types/) `mypy` disallows implicit optional since version 0.99.
The implicit optional type has been removed from PEP 484 in python/peps#689

dmontagu (approved):

yezz123 (approved):

pydantic/fields.py
```diff
@@ -223,29 +223,29 @@ def Field(
...
+    multiple_of: Optional[float] = None,
+    allow_inf_nan: Optional[bool] = None,
+    max_digits: Optional[int] = None,
+    decimal_places: Optional[int] = None,
+    min_items: Optional[int] = None,
+    max_items: Optional[int] = None,
+    unique_items: Optional[bool] = None,
+    min_length: Optional[int] = None,
+    max_length: Optional[int] = None,
+    allow_mutation: Optional[bool] = True,
```
hramezani, line 245: Is it really `Optional`? as the default value is `True`

hramezani (approved):

hramezani:
Thanks @JensHeinrich

PR #5132, opened 2023-03-02
Raise `ValidationError` for unhashable discriminator values

### LINKED ISSUE #4773: Unexpected Config Error with discriminator
### Initial Checks

### Description

Discriminator validation.

Expected: ValidationError when passing list or dict as discriminator.

Received: ConfigError.

### Example Code

```Python
from typing import Literal

import pydantic


class Model1(pydantic.BaseModel):
    target: Literal["t1"]
    a: int


class Model2(pydantic.BaseModel):
    target: Literal["t2"]
    b: int


class Foo(pydantic.BaseModel):
    foo: Model1 | Model2 = pydantic.Field(discriminator="target")


# Foo(**{"foo": {"target": []}})
Foo(**{"foo": {"target": {}}})
```

```Text
pydantic version: 1.10.2
            pydantic compiled: True
                 install path: /home/.../lib/python3.10/site-packages/pydantic
               python version: 3.10.8 (main, Oct 19 2022, 10:31:59) [GCC 9.4.0]
                     platform: Linux-5.15.0-53-generic-x86_64-with-glibc2.31
```

### Affected Components

kurtmckee:
As reported in #4773, unhashable discriminator values are incorrectly rejected with a `ConfigError`, rather than a `ValidationError`.

# The bug

The current code catches a `TypeError` at line 1121 and converts it to a `ConfigError`:

https://github.com/pydantic/pydantic/blob/7f3b754bea95fbb443772501b6176f2f4959f2a5/pydantic/fields.py#L1120-L1127

This can happen when `self.sub_fields_mapping` is `None`, in which case the `ConfigError` correctly suggests calling `.update_forward_refs()`. For reference, the exact error is:

```python
None[discriminator_value]  # TypeError: 'NoneType' object is not subscriptable
```

However, a `TypeError` can also be raised when `self.sub_fields_mapping` is a dictionary but `discriminator_value` is an unhashable type. For reference, the exact error is:

```python
{"valid": "dict"}[{}]  # TypeError: unhashable type: 'dict'
```

# The fix

This PR addresses the bug by first checking whether `self.sub_fields_mapping` is None. If it is, then a `ConfigError` is raised as expected. Then, both `KeyError` and `TypeError` are interpreted as an invalid discriminator value.

Please let me know if this needs to be separately forward-ported to `main`.

Thanks for your work on pydantic!

Fixes #4773

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -1117,15 +1117,18 @@ class ModelField(Representation):
             except (AttributeError, TypeError):
                 return v, ErrorWrapper(MissingDiscriminator(discriminator_key=self.discriminator_key), loc)
 
-        try:
-            sub_field = self.sub_fields_mapping[discriminator_value]  # type: ignore[index]
-        except TypeError:
+        if self.sub_fields_mapping is None:
             assert cls is not None
             raise ConfigError(
                 f'field "{self.name}" not yet prepared so type is still a ForwardRef, '
                 f'you might need to call {cls.__name__}.update_forward_refs().'
             )
-        except KeyError:
+
+        try:
+            sub_field = self.sub_fields_mapping[discriminator_value]  # type: ignore[index]
+        except (KeyError, TypeError):
+            # KeyError: `discriminator_value` is not in the dictionary.
+            # TypeError: `discriminator_value` is unhashable.
             assert self.sub_fields_mapping is not None
             return v, ErrorWrapper(
                 InvalidDiscriminator(
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -1125,7 +1125,7 @@ class ModelField(Representation):
             )
 
         try:
-            sub_field = self.sub_fields_mapping[discriminator_value]  # type: ignore[index]
+            sub_field = self.sub_fields_mapping[discriminator_value]
         except (KeyError, TypeError):
             # KeyError: `discriminator_value` is not in the dictionary.
             # TypeError: `discriminator_value` is unhashable.
--- a/tests/test_discrimated_union.py
+++ b/tests/test_discrimated_union.py
@@ -439,7 +439,5 @@ def test_discriminator_with_unhashable_type():
     class Foo(BaseModel):
         foo: Union[Model1, Model2] = Field(discriminator='target')
 
-    with pytest.raises(
-        ValidationError, match=re.escape("No match for discriminator 'target' and value {}")
-    ):
+    with pytest.raises(ValidationError, match=re.escape("No match for discriminator 'target' and value {}")):
         Foo(**{'foo': {'target': {}}})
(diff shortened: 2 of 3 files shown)


kurtmckee:
The GitHub workflow failed to start for this. If it's not possible to re-run the failed jobs, let me know and I'll push an update to re-trigger the workflow.

hramezani:
Thanks @kurtmckee for this.

The fix LGTM.

The only concern is that it is a change in old behavior(raising `ConfigError`) and I am afraid users depend on this(catching `ConfigError` on their code base).

> Please let me know if this needs to be separately forward-ported to main.

I think it's not needed.

@samuelcolvin, @dmontagu what do you think?

tests/test_discrimated_union.py
```diff
@@ -423,3 +423,23 @@ class Container(GenericModel, Generic[T]):
...
+    class Model2(BaseModel):
+        target: Literal['t2']
+        b: int
+
+    class Foo(BaseModel):
+        foo: Union[Model1, Model2] = Field(discriminator='target')
+
+    with pytest.raises(
+        ValidationError, match=re.escape("No match for discriminator 'target' and value {}")
+    ):
```
hramezani, line 444:
```suggestion
    with pytest.raises(ValidationError, match=re.escape("No match for discriminator 'target' and value {}")):
```

pydantic/fields.py
```diff
@@ -1117,15 +1117,18 @@ def _validate_discriminated_union(
...
+        if self.sub_fields_mapping is None:
             assert cls is not None
             raise ConfigError(
                 f'field "{self.name}" not yet prepared so type is still a ForwardRef, '
                 f'you might need to call {cls.__name__}.update_forward_refs().'
             )
-        except KeyError:
+
+        try:
+            sub_field = self.sub_fields_mapping[discriminator_value]  # type: ignore[index]
```
hramezani, line 1128:
```suggestion
            sub_field = self.sub_fields_mapping[discriminator_value]
```

hramezani:
@dmontagu 

> but I'd be a bit nervous to change the error handling in older versions unless this was causing issues that couldn't be properly handled by catching a ConfigError in the right place.

Agree!

kurtmckee:
It's my understanding that `ConfigError` would indicate a problem with the users' models, which is not the case here.

This appears to be a `ValidationError`, which is what users would anticipate having to catch when the discriminator key is invalid. I think that fixing this aligns the behavior with user expectations and code, and helps prevent production code from breaking due to an incorrect exception.

dmontagu:
@kurtmckee I agree that what you are describing is better behavior, the main point is that if someone is _already_ handling it as a ConfigError, it could end up breaking their code if they upgrade. If we were still planning additional releases to pydantic v1 this would _definitely_ make sense as an addition to a 1.11.0 release, but we aren't planning to do that now (instead, doing a 2.0.0).

That said, I recognize that this is arguably a bugfix; I'll defer to @samuelcolvin's opinion on what's the best path forward.

dmontagu:
Thought about this more and I think you are right about everything @kurtmckee. Going to merge this then do a new release.

kurtmckee:
Thank you @dmontagu, and thank you all for reviewing and discussing!

PR #5134, opened 2023-03-03
fix for small typo in orjson example

orgua:
typo is also present in current main-branch!


hramezani:
I think it is not a typo and the example uses `orjson_dumps` that defined above

orgua:
oh snap, you are right. sorry

PR #5137, opened 2023-03-04
Typo fix in metadata description function.

theonlykingpin:
# NOTICE

### DIFF
--- a/pydantic/_internal/_core_metadata.py
+++ b/pydantic/_internal/_core_metadata.py
@@ -56,7 +56,7 @@ class CoreMetadataHandler:
     def metadata(self) -> _CoreMetadata:
         """
         Retrieves the metadata dict off the schema, initializing it to a dict if necessary,
-        and erroring if it is not None and not a dict
+        and raise an error if it is not a dict.
         """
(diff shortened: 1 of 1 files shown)


pydantic/_internal/_core_metadata.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, schema: CoreSchema | TypedDictField):
     def metadata(self) -> _CoreMetadata:
         """
         Retrieves the metadata dict off the schema, initializing it to a dict if necessary,
-        and erroring if it is not None and not a dict
```
antonagestam, line 59: This doesn't seem like a good change to me, the method does handles `None`, so it doesn't make sense to remove that information.

Also nitpicking, changing from "erroring" to "raise an error" is the wrong verb tense. It should be "raises".

samuelcolvin:
Please replace the default body with a description of this change.

pydantic/_internal/_core_metadata.py
samuelcolvin: Agreed

PR #5138, opened 2023-03-04
Use RecursiveRefSchema to handle recursive TypedDict

### LINKED ISSUE #5058: Recrusive TypedDict crashes Pydantic
### Initial Checks

### Description

The following example causes a RecursionError

### Example Code

```Python
from typing import Optional, TypedDict

class RecursiveTypedDict(TypedDict):
    foo: Optional["RecursiveTypedDict"]

class Model(BaseModel):
    rec: RecursiveTypedDict
```

```Text
pydantic version: 1.10.4
            pydantic compiled: False
                 install path: /Users/adriangarciabadaracco/GitHub/pydantic/pydantic
               python version: 3.10.9 (main, Jan 27 2023, 18:06:58) [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.0-arm64-arm-64bit
```

### Affected Components

adriangb:
Closes #5058 

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -81,7 +81,6 @@ def generate_config(cls: type[BaseModel]) -> core_schema.CoreConfig:
 
 class GenerateSchema:
     __slots__ = 'arbitrary_types', 'types_namespace', 'typevars_map', '_recursion_cache'
-    _recursion_cache: dict[str, core_schema.DefinitionReferenceSchema]
 
     def __init__(
         self, arbitrary_types: bool, types_namespace: dict[str, Any] | None, typevars_map: dict[Any, Any] | None
@@ -89,7 +88,7 @@ class GenerateSchema:
         self.arbitrary_types = arbitrary_types
         self.types_namespace = types_namespace
         self.typevars_map = typevars_map
-        self._recursion_cache = {}
+        self._recursion_cache: dict[str, core_schema.DefinitionReferenceSchema] = {}
 
     def generate_schema(self, obj: Any) -> core_schema.CoreSchema:
         schema = self._generate_schema(obj)
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -80,7 +80,8 @@ def generate_config(cls: type[BaseModel]) -> core_schema.CoreConfig:
 
 
 class GenerateSchema:
-    __slots__ = 'arbitrary_types', 'types_namespace', 'typevars_map'
+    __slots__ = 'arbitrary_types', 'types_namespace', 'typevars_map', '_recursion_cache'
+    _recursion_cache: dict[str, core_schema.DefinitionReferenceSchema]
 
     def __init__(
         self, arbitrary_types: bool, types_namespace: dict[str, Any] | None, typevars_map: dict[Any, Any] | None
@@ -88,6 +89,7 @@ class GenerateSchema:
         self.arbitrary_types = arbitrary_types
         self.types_namespace = types_namespace
         self.typevars_map = typevars_map
+        self._recursion_cache = {}
 
     def generate_schema(self, obj: Any) -> core_schema.CoreSchema:
         schema = self._generate_schema(obj)
@@ -326,7 +328,9 @@ class GenerateSchema:
         assert expected, f'literal "expected" cannot be empty, obj={literal_type}'
         return core_schema.literal_schema(*expected)
 
-    def _typed_dict_schema(self, typed_dict_cls: Any) -> core_schema.TypedDictSchema:
+    def _typed_dict_schema(
+        self, typed_dict_cls: Any
+    ) -> core_schema.TypedDictSchema | core_schema.DefinitionReferenceSchema:
         """
         Generate schema for a TypedDict.
 
@@ -350,6 +354,13 @@ class GenerateSchema:
         fields: typing.Dict[str, core_schema.TypedDictField] = {}
         validation_functions = ValidationFunctions(())
         serialization_functions = SerializationFunctions(())
+
+        obj_ref = f'{typed_dict_cls.__module__}.{typed_dict_cls.__qualname__}:{id(typed_dict_cls)}'
+        if obj_ref in self._recursion_cache:
+            return self._recursion_cache[obj_ref]
+        else:
+            recursive_schema = core_schema.definition_reference_schema(obj_ref)
+            self._recursion_cache[obj_ref] = recursive_schema
 
         for field_name, annotation in _typing_extra.get_type_hints(typed_dict_cls, include_extras=True).items():
             required = field_name in required_keys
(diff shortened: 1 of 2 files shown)


adriangb:
What other types do we need to handle like this? I guess we technically could support something like this as long as we solve forward references:

```python
RecList = List[Union[None, "RecList"]]
```

pydantic/_internal/_generate_schema.py
```diff
@@ -288,19 +290,28 @@ def _literal_schema(self, literal_type: Any) -> core_schema.LiteralSchema:
...
+        self, typed_dict_cls: Any
+    ) -> core_schema.TypedDictSchema | core_schema.RecursiveReferenceSchema:
         """
         Generate schema for a TypedDict.
         """
         try:
             required_keys: typing.FrozenSet[str] = typed_dict_cls.__required_keys__
         except AttributeError:
-            raise TypeError('Please use `typing_extensions.TypedDict` instead of `typing.TypedDict`.')
+            raise TypeError('Please use `typing_extensions.TypedDict` instead of `typing.TypedDict` for Python < 3.11.')
```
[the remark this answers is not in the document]
dmontagu: It looks to me like `total` is still a thing in the Python 3.8 TypedDict — if that's right, I think we probably should reflect that properly (i.e., no keys required if total=False, all required if total=True). I'm not sure if it's possible to easily be more precise than that in python 3.8. I'd also be open to just changing the error message to `... for Python < 3.9`.

pydantic/_internal/_generate_schema.py
```diff
@@ -72,11 +72,13 @@ def generate_config(cls: type[BaseModel]) -> core_schema.CoreConfig:
 
 
 class GenerateSchema:
-    __slots__ = 'arbitrary_types', 'types_namespace'
+    __slots__ = 'arbitrary_types', 'types_namespace', '_seen'
+    _seen: Dict[str, core_schema.RecursiveReferenceSchema]
```
dmontagu, line 76: I think it would be worth adding a comment here or on the `__init__` describing what this attribute is used for; as of now, you have to read through the typed dict schema logic to understand.

This could also be a good candidate for replacing `str` with a `CoreRef` `NewType` as is done in the `GenerateJsonSchema`, not sure it's worth the effort at this stage though.

adriangb:
Tabling until #5009 / #4670 and #4665

samuelcolvin:
please review.

samuelcolvin:
please update.

pydantic/_internal/_generate_schema.py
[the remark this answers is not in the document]
adriangb: Resolved in #5142

adriangb:
please review

dmontagu:
Did you look at all into how this plays with recursive basemodels? Wondering if this approach could clean up some of the handling of recursive models from the generics PR.

Not sure if there's something special about TypedDict but I guess in an ideal world we'd use the same approach to handle recursion for both. If that's more involved than you want to get at this stage, I can look into that side of things once we feel good about everything else.

adriangb:
> Did you look at all into how this plays with recursive basemodels? Wondering if this approach could clean up some of the handling of recursive models from the generics PR.
> Not sure if there's something special about TypedDict but I guess in an ideal world we'd use the same approach to handle recursion for both. If that's more involved than you want to get at this stage, I can look into that side of things once we feel good about everything else.

I imagine there is quite a bit of overlap. I'm not sure generics (never mind recursive generics) work on TypedDict. I did open #5174 to tackle this in the future, e.g. if recursive generics don't work on TypedDict but we want them to that would be a good time to sit down and refactor to share logic with `BaseModel`. But for now this is simple enough and solves the somewhat common (I think) use case of recursion with TypedDict, so I would table merging logic with BaseModel for a future PR.

adriangb:
please review

tests/test_types_typeddict.py
```diff
@@ -378,3 +378,32 @@ class Model(BaseModel):
...
+    ls: List[T]
+
+
+class RecursiveGenTypedDictModel(BaseModel, Generic[T]):
+    rec: RecursiveGenTypedDict[T]
+
+
+def test_recursive_generic_typeddict():
+    data: RecursiveGenTypedDict[int] = {'foo': {'foo': None, 'ls': []}, 'ls': []}
+    assert RecursiveGenTypedDictModel[int](rec=data).rec == data
```
dmontagu, line 409: can you expand this test to show what happens if the "inner" _and_ "outer" values of `ls` are equal to `[1]`, and `['a']`? If it isn't validating properly, you can make it an xfail or add a TODO comment, and I'll get it working when I work on generic dataclasses.

dmontagu: Can you also add a TODO: comment indicating to make sure it works if `RecursiveGenTypedDict` was a forward reference? Don't worry about even writing the test, I'll do that

dmontagu:
If you just edit the test as indicated, I'd be happy to merge this. I think the logic might get changed later (or the logic for recursive generics might get changed .. :) but we can deal with any appropriate changes after dataclasses is done.

adriangb:
Ok handled, I'll let you merge @dmontagu

pydantic/_internal/_generate_schema.py
```diff
@@ -351,6 +354,13 @@ def _typed_dict_schema(self, typed_dict_cls: Any) -> core_schema.TypedDictSchema
         validation_functions = ValidationFunctions(())
         serialization_functions = SerializationFunctions(())
 
+        obj_ref = f'{typed_dict_cls.__module__}.{typed_dict_cls.__qualname__}:{id(typed_dict_cls)}'
```
samuelcolvin, line 357: please can we have a single function for this that's used in both places.

tests/test_types_typeddict.py
```diff
@@ -378,3 +378,111 @@ class Model(BaseModel):
             Model(d=input_value)
     else:
         assert Model(d=input_value).d == expected
+
+
+class RecursiveTypedDict(typing_extensions.TypedDict):
+    # TODO: See if we can get this working if defined in a function (right now, needs to be module-level)
```
samuelcolvin, line 384: you need to use `create_module` as is done lots in `test_forwards_refs.py`.

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin:
please update.

pydantic/_internal/_generate_schema.py
dmontagu: see `pydantic._internal._core_utils.get_type_ref`

tests/test_types_typeddict.py
dmontagu: That works, but I also think it _may_ be possible to get it working inside a function and not a module. I asked Adrian to add this comment so I could look at it later when I look at the other recursive/generic stuff. Module is a good idea though to prevent it from running on test execution (making it easier to prevent debugger from hitting unwanted code)

adriangb:
It looks like there was feedback 2h ago but this got merged this 17h ago. I'll create a new PR to address the feedback.

PR #5142, opened 2023-03-06
Enforce using typing_extensions.TypedDict for Python < 3.11

adriangb:
Extracting a bit of work out of #5058  and from discussion with @dmontagu

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -6,6 +6,7 @@ from __future__ import annotations as _annotations
 import collections.abc
 import dataclasses
 import re
+import sys
 import typing
 import warnings
 from typing import TYPE_CHECKING, Any
@@ -28,6 +29,9 @@ if TYPE_CHECKING:
     from ..main import BaseModel
 
 __all__ = 'model_fields_schema', 'GenerateSchema', 'generate_config'
+
+
+_SUPPORTS_TYPEDDICT = sys.version_info >= (3, 11)
 
 
 def model_fields_schema(
@@ -310,10 +314,18 @@ class GenerateSchema:
         """
         Generate schema for a TypedDict.
         """
-        try:
-            required_keys: typing.FrozenSet[str] = typed_dict_cls.__required_keys__
-        except AttributeError:
-            raise TypeError('Please use `typing_extensions.TypedDict` instead of `typing.TypedDict`.')
+        # It is not possible to track required/optional keys in TypedDict without __required_keys__
+        # since TypedDict.__new__ erases the base classes (it replaces them with just `dict`)
+        # and thus we can track usage of total=True/False
+        # __required_keys__ was added in Python 3.9 (https://github.com/miss-islington/cpython/blob/1e9939657dd1f8eb9f596f77c1084d2d351172fc/Doc/library/typing.rst?plain=1#L1546-L1548)  # noqa: E501
+        # however it is buggy (https://github.com/python/typing_extensions/blob/ac52ac5f2cb0e00e7988bae1e2a1b8257ac88d6d/src/typing_extensions.py#L657-L666)  # noqa: E501
+        # Hence to avoid creating validators that do not do what users expect we only
+        # support typing.TypedDict on Python >= 3.11 or typing_extension.TypedDict on all versions
+        is_typing_typeddict = type(typed_dict_cls).__module__ == 'typing'
+        if not _SUPPORTS_TYPEDDICT and is_typing_typeddict:
+            raise TypeError('Please use `typing_extensions.TypedDict` instead of `typing.TypedDict` on Python < 3.11.')
+
+        required_keys: typing.FrozenSet[str] = typed_dict_cls.__required_keys__
 
         fields: typing.Dict[str, core_schema.TypedDictField] = {}
         validation_functions = ValidationFunctions(())
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -321,8 +321,7 @@ class GenerateSchema:
         # however it is buggy (https://github.com/python/typing_extensions/blob/ac52ac5f2cb0e00e7988bae1e2a1b8257ac88d6d/src/typing_extensions.py#L657-L666)  # noqa: E501
         # Hence to avoid creating validators that do not do what users expect we only
         # support typing.TypedDict on Python >= 3.11 or typing_extension.TypedDict on all versions
-        is_typing_typeddict = type(typed_dict_cls).__module__ == 'typing'
-        if not _SUPPORTS_TYPEDDICT and is_typing_typeddict:
+        if not _SUPPORTS_TYPEDDICT and type(typed_dict_cls).__module__ == 'typing':
             raise TypeError('Please use `typing_extensions.TypedDict` instead of `typing.TypedDict` on Python < 3.11.')
 
         required_keys: typing.FrozenSet[str] = typed_dict_cls.__required_keys__
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -15,7 +15,7 @@ from annotated_types import BaseMetadata, GroupedMetadata
 from pydantic_core import SchemaError, SchemaValidator, core_schema
 from typing_extensions import Annotated, Literal, get_args, get_origin, is_typeddict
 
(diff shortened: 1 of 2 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -292,16 +292,24 @@ def _typed_dict_schema(self, typed_dict_cls: Any) -> core_schema.TypedDictSchema
...
+        type_hints = _typing_extra.get_type_hints(typed_dict_cls, include_extras=True)
+        required_keys: typing.FrozenSet[str]
         try:
-            required_keys: typing.FrozenSet[str] = typed_dict_cls.__required_keys__
+            required_keys = typed_dict_cls.__required_keys__
         except AttributeError:
-            raise TypeError('Please use `typing_extensions.TypedDict` instead of `typing.TypedDict`.')
+            # __required_keys__ is only available on Python >= 3.11
+            # if it is not present then _all_ keys are either jointly required (total=True)
+            # or optional (total=False)
```
dmontagu, line 302: do you know in pre-3.11 if you could achieve more nuanced logic as a function of base classes? I'm thinking the answer is yes. I defer to you/Samuel what we should do before __required_keys__ was available, but it is probably worth noting that technically this might not perfectly match the defined logic.

What I mean is:
```python
class MyTotalTypedDict(TypedDict):
    x: int

class MyLessTotalTypedDict(MyTotalTypedDict):
    y: int
```
Even in versions of python without `__required_keys__` I think `x` might be required and `y` not required.

adriangb: Let me take a look into the implementation of __required_keys__ in 3.11. Maybe we can just copy that

adriangb: But off the top of my head yes, you're right I think we actually need to traverse all the base classes to get the right logic. I expect implementation won't be that bad, but the test might be super ugly if we want to cover all the edge cases.

adriangb:
@samuelcolvin after much discussion @dmontagu and I came to the conclusion that it might be best to force users to use `typing_extensions.TypedDict` for < 3.11. There's possibly some improvements to how we check that to be made here but I wanted to get your perspective first. If you think things should stay as they are I can also pull out just the comments from here so that we at least don't loose this research in the future.

dmontagu:
Just want to add some color on the issues with `typing.TypedDict`:

Big picture, the issues stem from: 1, inheritance and the fact that you can't "walk" the base classes at runtime, and specifically 2, challenges with determining which keys are required (which depends on the `total`-ness of the base class in which each attribute is introduced, and potentially on the presence of the `Required` and `NotRequired` type wrappers).

* In python3.8, there is no `__required_keys__` or `__optional_keys__` attributes, and there is no way to obtain the base classes for a `typing.TypedDict`, so I think supporting requiredness-detection (in particular, reflecting the `total`-ness of base classes) for `typing.TypedDict` types using inheritance is out of the question in python3.8. This is such a significant issue that I don't think we should make _any_ attempt to support `typing.TypedDict` in python 3.8.
* Supposedly in python 3.9.0 and 3.9.1 there are some issues with the correctness of the `__required_keys__` and `__optional_keys__` attributes, and again, there is no way to "walk" the base classes. So I think we should not attempt to support `typing.TypedDict` on these python versions either.
    * Note: `__required_keys__` and `__optional_keys__` were only _officially_ added in 3.11, but have been present (just undocumented) since 3.9.
* In the `typing.TypedDict` from Python >=3.9.2,<3.11, I believe that the `__optional_keys__` and `__required_keys__` attributes are present, and correctly reflect the `total`-ness with inheritance/etc., BUT, unsurprisingly, do not respect `typing_extensions.NotRequired` or `typing_extensions.Required`. I believe with some elbow grease we could implement a function ourselves that computes which keys are _actually_ required (i.e., respecting `typing_extensions.Required` and `typing_extensions.NotRequired`), something like:

        def get_real_required_optional_keys(typed_dict_type: type[TypedDict]) -> tuple[FrozenSet[str], FrozenSet[str]]:
            required_keys = typed_dict_type.__required_keys__
            optional_keys = typed_dict_type.__optional_keys__
            for k, v in typed_dict_type.__annotations__.items():
                if getattr(v, '__origin__', None) is Required:
                    optional_keys = frozenset([x for x in optional_keys if x != k])
                    required_keys = frozenset([k, *required_keys])
                elif getattr(v, '__origin__', None) is NotRequired:
                    optional_keys = frozenset([k, *optional_keys])
                    required_keys = frozenset([x for x in required_keys if x != k])
            return required_keys, optional_keys

    But I'm not sure if there might be some surprising complexities (e.g., handling `Annotated[Required[X], ...]`). Given that, I'm not sure it's worth maintaining our own logic for this.

I think there are three obvious possible paths forward:

* Least effort: Don't support `typing.TypedDict` on Python <3.11 at all
* Middle effort: Support `typing.TypedDict` on Python >=3.9.2, but don't attempt to handle `typing_extensions.Required` and `typing_extensions.NotRequired`. (We could or could not emit a warning.)
* Most effort: Support `typing.TypedDict` on Python >=3.9.2, and make a best-effort attempt to handle `typing_extensions.Required` and `typing_extensions.NotRequired` ourselves (basically, manually "correcting" the `__required_keys__` and `__optional_keys__` attributes present on the type).

(To be clear, for each of the above approaches, I would propose we support `typing_extensions.TypedDict` on all python versions, and `typing.TypedDict` for Python >=3.11, both of which do not suffer from any of the issues described above as far as I'm aware.)

I think for reasons of migration it _may_ be a good idea to take the middle- or most-effort option, but I could be convinced about any of the above.

samuelcolvin:
I don't think the middle option works - could be too confusing.

We either go for least effort, or go the whole hog and respect both `Required`s.

Happy to do least effort now, then and full support later.

samuelcolvin:
Yes, please create an issue now. Assuming this is complete, I'll review it soon.

tests/test_types_typeddict.py
```diff
@@ -33,6 +34,9 @@ def fixture_typed_dict(TypedDictAll):
     class TestTypedDict(TypedDictAll):
         foo: str
 
+    if sys.version_info < (3, 11) and TypedDictAll.__module__ == 'typing':
+        pytest.xfail('typing.TypedDict does not track required keys correctly on Python < 3.11')
```
samuelcolvin, line 38: might be better to use `skip`?

The plan is to remove all `xfail`s before v2, and this code will remain I guess.

samuelcolvin (commented):
otherwise LGTM.

Please update.

pydantic/_internal/_generate_schema.py
```diff
@@ -292,10 +296,18 @@ def _typed_dict_schema(self, typed_dict_cls: Any) -> core_schema.TypedDictSchema
...
-            raise TypeError('Please use `typing_extensions.TypedDict` instead of `typing.TypedDict`.')
+        # It is not possible to track required/optional keys in TypedDict without __required_keys__
+        # since TypedDict.__new__ erases the base classes (it replaces them with just `dict`)
+        # and thus we can track usage of total=True/False
+        # __required_keys__ was added in Python 3.9 (https://github.com/miss-islington/cpython/blob/1e9939657dd1f8eb9f596f77c1084d2d351172fc/Doc/library/typing.rst?plain=1#L1546-L1548)  # noqa: E501
+        # however it is buggy (https://github.com/python/typing_extensions/blob/ac52ac5f2cb0e00e7988bae1e2a1b8257ac88d6d/src/typing_extensions.py#L657-L666)  # noqa: E501
+        # Hence to avoid creating validators that do not do what users expect we only
+        # support typing.TypedDict on Python >= 3.11 or typing_extension.TypedDict on all versions
+        is_typing_typeddict = type(typed_dict_cls).__module__ == 'typing'
+        if not _SUPPORTS_TYPEDDICT and is_typing_typeddict:
```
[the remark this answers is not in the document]
adriangb: No big pushback but generally as soon as I have more than 1 boolean I like to give each condition a name so that the logic branch is more readable i.e. `if not we_support_typeddict_on_this_python_version and typed_dict_class_is_from_the_typing_module:` is more readable to me than `if sys.version_info < (3, 11) and type(typed_dict_cls).__module__ == 'typing':`. But that's a bit of personal preference, especially for 2 items. It think it is almost required if we're talking about 4+ items or combining and/or conditions.

adriangb:
All feedback addressed. This is ready for a merge.

Issue #4669, opened 2022-10-27
V2: The `validate` general method


samuelcolvin:
Pydantic V2 will do a massively better job of validating arbitrary objects.

To accomplish this without many methods, we should provide one function which can:
* be used as a decorator on a function to validate it's argument, and optionally return type
* be used as a decorator on `dataclass`, `NamedTuple`, `TypedDict`
* be used to create a "validated" version of anything, e.g. `ValidatedTuple = validate(tuple[int, int, int])`

With this, the pydantic version of the `dataclass` decorator, will just become effectively:

```py
def dataclass(*args, **kwargs):
    return validate(dataclasses.dataclass(*args, **kwargs))
```

Dataclasses need work (#4670), but validation and generate validation schemas for the rest of these types should already work.

samuelcolvin:
So usage would be

```py
from pydantic import validate

@validate(...config)
def my_method(...):
   ...

@validate(...config)
@dataclass
class MyDataclass:
   ...

@validate(...config)
class MyTypeDict(TypedDict):
   ...
```

Main question: How do we define validators, do we continue to extract them as methods, or do we let them be provided as kwargs to `validate`, or both?

marcoo47:
Is this issue still open? if so I was wondering if it would be a reasonable first contribution for me and my partner @smhavens. thanks!

samuelcolvin:
I'm afraid this is probably quite a challenging first task.

How about something like #4675? It's still pretty hard and complicated, but not as wide ranging as this.

gavindsouza:
Q: Would the `validate` API do more than raise exceptions? Would it try to parse/coerce objects as `parse_obj_as` (eg: parse_obj_as(str, 1) => "1") does now?

samuelcolvin:
(added a few more examples in the above usage example)

So the question is, what does `validate` return.

Well, it has to depend on what it's called with:
* if it's used as a decorator on a dataclass, it needs to return a valid dataclass, with a custom `__init__` method
* similar if it's used as a decorator on a `NamedTuple` or `TypedDict`
* if it's used as as decorator on a function, it can simply return another function - this is the easiest case
* if it's called with a `Union`, I'm not sure if we can create something that looks like a valid `typing.Union`, otherwise it should just return a function
* if it's called on a tuple, like `validate(tuple[int, int, int])`, hopefully we can create a tuple subclass, same for `list` and `dict` etc.
[...]

PR #5145, opened 2023-03-07
Introduce Validator for arbitrary non-model types

adriangb:
Part of #4669


pydantic/_internal/_model_construction.py
```diff
@@ -175,16 +175,19 @@ def complete_model_class(
...
         cls,
         inner_schema,
         config=core_config,
         call_after_init=model_post_init,
         metadata=build_metadata_dict(js_metadata=js_metadata),
     )
+    validation_schema = core_schema.union_schema(inner_schema, outer_schema)
+    cls.__pydantic_validator__ = SchemaValidator(validation_schema, core_config)
+    cls.__pydantic_core_schema__ = outer_schema
+    cls.__pydantic_core_validation_schema__ = validation_schema
```
adriangb, line 190: This is so that `Validator(SomeModel)` accepts and instance of `SomeModel`.

pydantic/main.py
```diff
@@ -721,3 +723,41 @@ def create_model(
...
+
+T = TypeVar('T')
+
+
+class Validator(Generic[T]):
+    @overload
+    def __init__(self, __type: Type[T], *, config: CoreConfig | None = None) -> None:
+        ...
+
+    @overload
```
adriangb, line 736:
```suggestion
    # Adding this overload ensures you can use special forms without getting mypy errors.
    # For example:
    #   v: Validator[int | str] = Validator(int | str)
    # Type checkers don't consider special forms like `int | str` (or `Union[int, str]`) to satisfy a Type[T]
    @overload
```

samuelcolvin:
please update.

tests/test_validator.py
```diff
@@ -0,0 +1,94 @@
...
+        (Tuple[str], ('1',), ('1',)),
+        (Tuple[str, int], ('1', 1), ('1', 1)),
+        (Tuple[str, ...], ('1',), ('1',)),
+        (Dict[str, int], {'foo': 123}, {'foo': 123}),
+        (Union[int, str], 1, 1),
+        (Union[int, str], '2', '2'),
+        (GenericPydanticModel[int], {'x': [[1]]}, GenericPydanticModel[int](x=[[1]])),
+        (GenericPydanticModel[int], {'x': [['1']]}, GenericPydanticModel[int](x=[[1]])),
+        (NestedList[int], [[1]], [[1]]),
+        (NestedList[int], [['1']], [[1]]),
```
adriangb, line 50: @dmontagu this "just worked", I was honestly kinda surprised. Should I be doing more in-depth testing here or if this works does it mean that all of our generics support works?

pydantic/main.py
```diff
@@ -832,3 +843,47 @@ def create_model(
...
+        # On the other hand Validator() can be called with arbitrary objects,
+        # including type aliases where __module__ (always `typing.py`) is not useful
+        # So instead we look at the globals in our parent stack frame
+        # This works for the case where Validator() is called in a module that
+        # has the target of forward references in it's scope but obviously
+        # does not work for more complex cases
+        global_ns = sys._getframe(1).f_globals.copy()
+        global_ns.update(local_ns or {})
+        gen = _generate_schema.GenerateSchema(
+            arbitrary_types=arbitrary_types, types_namespace=global_ns, typevars_map={}
```
adriangb, line 883: @dmontagu should I be passing anything in here for `typevars_map`? Seems to work with an empty dict but not sure.

adriangb:
please review

tests/test_validator.py
adriangb: I'm going to add a similar test in another branch. The following doesn't work currently (unrelated to this PR):

```python
from typing import Generic, List, TypeVar

ItemType = TypeVar('ItemType')

class GenericPydanticModel(BaseModel, Generic[ItemType]):
    x: ItemType

GenericPydanticModelList = List[GenericPydanticModel[ItemType]]
GenericPydanticModelList[int]
```

adriangb: #5183

pydantic/main.py
```diff
@@ -832,3 +843,47 @@ def create_model(
...
+        }
+        arbitrary_types = bool((config or {}).get('arbitrary_types_allowed', False))
+        local_ns = _typing_extra.parent_frame_namespace(parent_depth=2)
+        # BaseModel uses it's own __module__ to find out where it was defined
+        # and then look for symbols to resolve forward references in those globals
+        # On the other hand Validator() can be called with arbitrary objects,
+        # including type aliases where __module__ (always `typing.py`) is not useful
+        # So instead we look at the globals in our parent stack frame
+        # This works for the case where Validator() is called in a module that
+        # has the target of forward references in it's scope but obviously
```
dmontagu, line 878:
```suggestion
        # has the target of forward references in its scope
```

I would probably drop the `but obviously ...` part of this comment unless there is an easily explained example of how this can fail that you think is a risk could be used in otherwise-reasonable code, in which case, I'd probably include a little detail about that example here if possible.

dmontagu:
```
has the target of forward references in its scope, but could fail, for example,
if calling Validator on an imported type which contains forward references to
other types only defined in the module from which it was imported
```

Another possibility given the example you shared on slack☝️

dmontagu (approved):

PR #5149, opened 2023-03-07
use definitions to define all models

samuelcolvin:
experiment with using defininitions to define models, also some refactoring.


dmontagu:
@samuelcolvin I'm guessing this is superseded by #5285?

samuelcolvin:
replaced by #5302 and #5305

PR #5151, opened 2023-03-08
Fix up more of the test_edge_cases.py tests

dmontagu:
I have tried to update the remaining tests in this file to be more compatible with v2. This entails:

* Updating the expected errors/etc. and removing the xfail for some tests
* Updating the expected errors/etc. where possible but still leaving the xfail for some tests where I could update some things but not quite get it to pass
* Add some notes about why tests are failing and/or asking how they should be resolved

In several cases, I was unsure if we intended to preserve the v1 behavior and/or whether there was a plan for how we intend to achieve it. I tried to add a TODO: note with a question in each such instance. I'm happy whether we try to resolve those questions before merging, or after (down the road).

Also, on a tangential note — a significant fraction of the remaining failures in this file relate to validators — @adriangb for your context I've been trying to add ` - validators` in the xfail reason for tests failing for things related to validators (both here and in other non-`test_validators.py` files); searching for that might help you find other tests to get passing that aren't in the `test_validators.py` file.

-----

Update: I have added comments below indicating everywhere that I wasn't 100% sure what the right thing to do in v2 was.

Selected Reviewer: @adriangb


tests/test_edge_cases.py
```diff
@@ -1078,6 +1078,7 @@ class C(A):
         class D(A):
             integer = 'G'
 
+    # TODO: Do we want any changes to this behavior in v2? (Currently, no error is raised)
     assert str(exc_info.value) == (
         'The type of D.integer differs from the new default value; '
         'if you wish to change the type of this field, please use a type annotation'
```
dmontagu, line 1084: Question 1 ☝️

tests/test_edge_cases.py
```diff
@@ -1166,6 +1167,7 @@ def test_unable_to_infer():
         class InvalidDefinitionModel(BaseModel):
             x = None
 
+    # TODO: Do we want any changes to this behavior in v2? (Currently, no error is raised)
     assert exc_info.value.args[0] == 'unable to infer type for attribute "x"'
```
dmontagu, line 1171: Question 2 ☝️

tests/test_edge_cases.py
```diff
@@ -1340,6 +1342,8 @@ class TopModel(BaseModel):
         self: str
         nest: NestedModel
 
+    # TODO: Do we want any changes to this behavior in v2? (Currently the __init__-override is not called)
+    #   We could detect if `__init__` has been overridden if necessary...
     m = TopModel.model_validate(dict(self='Top Model', nest=dict(self='Nested Model', modified_number=0)))
```
dmontagu, line 1347: Question 3 ☝️

tests/test_edge_cases.py
```diff
@@ -1933,41 +1975,53 @@ def __init__(self, t1: T1, t2: T2):
...
             gen: MyGen[str, bool]
 
 
-@pytest.mark.xfail(reason='working on V2')
 def test_hashable_required():
     class Model(BaseModel):
         v: Hashable
 
+        # TODO: Should arbitrary_types_allowed be necessary for Hashable?
+        model_config = dict(arbitrary_types_allowed=True)
```
dmontagu, line 1994: Question 4 ☝️

tests/test_edge_cases.py
```diff
@@ -2135,6 +2157,7 @@ def __new__(cls, data: int):
             return self
 
     m = MyModel(my_int=IntSubclass(123))
+    # TODO: Is this still the behavior we want in v2? (Currently m.my_int.__class__ is int)
     assert m.my_int.__class__ == IntSubclass
```
dmontagu, line 2161: Question 5 ☝️

tests/test_edge_cases.py
```diff
@@ -2161,6 +2184,12 @@ def test_long_int():
     class Model(BaseModel):
         x: int
 
+    # TODO: The next line now raises the following error:
+    #     E       pydantic_core._pydantic_core.ValidationError: 1 validation error for Model
+    #     E       x
+    #     E         Input should be a finite number [type=finite_number,
+    #     input_value='111111111111111111111111...11111111111111111111111', input_type=str]
+    #   Do we need to resolve this? How hard would that be in pydantic_core? Is it worth it?
     assert Model(x='1' * 4_300).x == int('1' * 4_300)
```
dmontagu, line 2193: Question 6 ☝️

tests/test_edge_cases.py
samuelcolvin: I think it should be an error to redefine any field without an annotation - that way we don't need to start trying to infer the type of the default value.

tests/test_edge_cases.py
samuelcolvin: `x` definitely shouldn't be a field, I guess an error would be best, but might be hard to identify "non-field attributes reliable"?

tests/test_edge_cases.py
samuelcolvin: I guess this should be an error or warning. If you want to do stuff on init, you should use `model_post_init`

tests/test_edge_cases.py
samuelcolvin: ideally I guess we should have a validator for this.

tests/test_edge_cases.py
samuelcolvin: yes, because in `pydantic-core` we cast the value to a rust `i64`, so the sub-type information is lost.

If you want to keep subtype information you should use an `is-instance` validator - I guess we need a python type structured as an annotation (like `ImportString`) to allow use of the `is-instance` validator, so the syntax here would be something like

```py
class MyModel(BaseModel):
    my_int: IsInstance[int]
    # which is just shorthand for
    my_int: Annotated[int, CheckIsInstance]
```

(`CheckIsInstance` name needs thought)

tests/test_edge_cases.py
samuelcolvin: in `pydantic-core` we use an `i64`, which constrains the max and min values. Since that's massively more performant, and there are very few real world uses for int > `i64:MAX`, the error is correct.

Again, if you really want a bigger int, use the `is-instance` validator and your own validator functions - I'll add this to the migrations guide points now.

samuelcolvin: https://github.com/orgs/pydantic/projects/2

samuelcolvin (commented):
otherwise this is looking great.

samuelcolvin:
please update.

tests/test_edge_cases.py
dmontagu: What if you want to do something in `__init__` _prior_ to the call to `super().__init__()`?

An example might be adding an extra argument that is used to do something to the other input arguments (e.g., something like having a flag to indicate whether they should be manually injected from environment if not present).

I'm not saying this is necessarily something we should intend to support, but I could see it posing a migration challenge for users who are using this functionality with their existing (v1) BaseModel classes.

If there's a way to migrate such logic I think it's fine to require changes, just seems like a good idea to make sure it's possible. (Seems like something a root-validator would be able to do, but not sure what the status of those is in v2).

dmontagu:
I updated all the TODO comments I asked for more detail about; let's get this merged now as it fixes up otherwise-working tests and address the issues in separate PRs.

dmontagu:
please review

tests/test_edge_cases.py
adriangb: Would the error be that `x` is a non-field attribute? Wouldn't the error just be something along the lines of `x does not have an annotation`?

tests/test_edge_cases.py
adriangb: I'm curious as to why this doesn't work, I assume it's a technical reason?

tests/test_edge_cases.py
```diff
@@ -1552,7 +1566,7 @@ class MyModel(BaseModel):
     assert dict(m) == {'foo': None}
 
 
-@pytest.mark.xfail(reason='working on V2')
+@pytest.mark.xfail(reason='working on V2 - typevar-handling will be improved after merging generics')
```
adriangb, line 1569: Hasn't generics been merged? Maybe this needs a rebase / merge?

tests/test_edge_cases.py
```diff
@@ -1795,22 +1819,38 @@ class Model(BaseModel):
     with pytest.raises(ValidationError) as exc_info:
         Model()
     assert exc_info.value.errors() == [
-        {'loc': ('nullable1',), 'msg': 'field required', 'type': 'value_error.missing'},
-        {'loc': ('nullable2',), 'msg': 'field required', 'type': 'value_error.missing'},
+        {'input': {}, 'loc': ('optional1',), 'msg': 'Field required', 'type': 'missing'},
+        {'input': {}, 'loc': ('nullable1',), 'msg': 'Field required', 'type': 'missing'},
+        {'input': {}, 'loc': ('nullable2',), 'msg': 'Field required', 'type': 'missing'},
```
adriangb, line 1824: maybe not for this PR, but is `optional2` relying on inference of field types from default arguments? If we're getting rid of that we'll need to edit this test to something like:

````
    class Model(BaseModel):
        optional1: Any
        optional2: Optional[Any] = ...
        nullable1: Any = ...
        nullable2: Any = Field(...)
```

tests/test_edge_cases.py
adriangb: Agreed. We should be able to at least check at build time if `__hash__` and `__eq__` are defined (and maybe that they are not `object.__hash__/__eq__`).

tests/test_edge_cases.py
adriangb: Makes sense to me

adriangb (approved):
Some minor feedback, feel free to address or table, but let's move forward with this.

tests/test_edge_cases.py
dmontagu: yes, it should be changed to something like that; however I don't want to update the expected error message here, I think that can be done in whatever pull request makes it actually raise an error.

tests/test_edge_cases.py
dmontagu: > why this doesn't work

from `pydantic.main`:
```python
    @classmethod
    def model_validate(cls: type[Model], obj: Any) -> Model:
        values, fields_set = cls.__pydantic_validator__.validate_python(obj)
        m = cls.__new__(cls)
        _object_setattr(m, '__dict__', values)
        _object_setattr(m, '__fields_set__', fields_set)
        if hasattr(cls, '__pydantic_post_init__'):
            cls.__pydantic_post_init__(context=None)  # type: ignore[attr-defined]
        return m
```
You can see it doesn't actually call the `__init__` function.

tests/test_edge_cases.py
dmontagu: good catch — I think this can probably be tweaked to fix these now. I'll do that and resolve this once it's done.

tests/test_edge_cases.py
dmontagu: good catch. I'm thinking based on this it might make sense to change it to:
```python
    class Model(BaseModel):
        optional1: Any
        optional2: Any = None
        optional3: Optional[Any] = None
        nullable1: Any = ...
        nullable2: Any = Field(...)
        nullable3: Optional[Any] = ...
```

adriangb: That sounds good to me

PR #5152, opened 2023-03-08
Fix tests in test_validators.py

adriangb:


samuelcolvin:
merging as this LGTM, in future good to add an explicit, note saying "merge if you're happy".

Issue #5068, opened 2023-02-17
Validator with multiple field inputs not working [V2]


XiaoHuiHui233:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

I'm not sure if this is a Pydantic V2-only bug, but I discovered it while using V2.

I have a BaseModel with two `int` attributes and a validator method. I want to apply the validator to the two attributes. But no matter how many fields I give, only the first one will trigger the validator. And when I split it into two validators and functions, each specifying one attribute, it works fine. But this will generate a lot of redundant code, and this is the syntax introduced in the documentation.

Since I haven't tested in v1, I'm not sure if this feature is a bug or not. V2 hasn't been released yet, so this may be a feature that hasn't been done yet, if so please forgive me for reporting a bug when using a version that wasn't released at all. But I am looking forward to the release of V2, it has too many features that I want, thank you very much.

### Example Code

```Python
from pydantic import BaseModel, validator

class MyModel(BaseModel):
    a: int
    b: int
    
    @validator("a", "b", allow_reuse=True)
    def check_num(cls, v, **kwargs):
        print("hello")
        return 2

print(MyModel.model_validate_json('{"a": 0, "b": 1}'))

# The output:
# hello
# a=2 b=1

# For comparison, the following code is my expected behavior.

from pydantic import BaseModel, validator

class MyModel(BaseModel):
    a: int
    b: int
    
    @validator("a")
    def check_num(cls, v, **kwargs):
        print("hello")
        return 2

    @validator("b")
    def check_num2(cls, v, **kwargs):
        print("hello")
        return 2

print(MyModel.model_validate_json('{"a": 0, "b": 1}'))

# The output:
# hello
# hello
# a=2 b=2
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0.0.dev0
install path: /.../miniconda3/envs/.../lib/python3.10/site-packages/pydantic
python version: 3.10.9 (main, Jan 11 2023, 15:21:40) [GCC 11.2.0]
[...]

Issue #4707, opened 2022-11-02
Applying validators to child items of lists


adriangb:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this feature is missing

### Description

Followup to https://github.com/pydantic/pydantic/pull/4516#discussion_r1012045834

We are currently on the path to allow the following in v2:

```python
data: Annotated[list[int], Field(min_items=3, gt=0)]
# or
data: list[int] = Field(min_items=3, gt=0)
```

Since `gt=0` doesn't apply to `list` we'd automatically apply it the items (`int`).

I think this behavior might be confusing and lead to issues down the road so I would like to propose we remove it for now and add it later once v2 is more stable and if users request it. The alternative, which is just a bit more boilerplate, is:

```python
data: Annotated[list[Annotated[int, Field(gt=0)], Field(min_items=3)]
# or the more legible option imo
PositiveInt = Annotated[int, Field(gt=0)]
data: Annotated[list[PositiveInt], Field(min_items=3)]
# but also possibly
PositiveInt = Annotated[int, Field(gt=0)]
T = TypeVar("T")
ListWithAtLeast3Items = Annotated[list[T], Field(min_items=3)]
data: ListWithAtLeast3Items[PositiveInt]
# which allows m x n composability
PhoneNumber = Annotated[str, Field(regex=r"\d{9})]
data2: ListWithAtLeast3Items[PhoneNumber]
```

I think the advantages of this style are:
1. Straightforward interpretation. There is no confusion as to what things constraints apply to, how constraints interact with generics, etc.
2. Easier to implement (I think). We don't need any special casing, we always apply constraints to the type at the level the annotations is at.
3. More re-usable code. When using type aliases, this style naturally leads to re-usable code.
4. More legible code. Using type aliases with meaningful names helps understand the code. If you read `ListWithAtLeast3Items[PhoneNumber]` you probably don't even have to go look at the constraints to understand what is valid and what is not.

The disadvantages are:
1. More code and if using type aliases more _lines_ of code. I think this is mitigated by the fact that the magic pushdown approach doesn't scale beyond a single level of nesting. If you want to do `list[list[int]]` you still have to use the nested annotated approach.

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
[...]

PR #5153, opened 2023-03-08
Add ValidationInfo param in `@validator` & add backwards compat shims

### LINKED ISSUE #5045: Modelattribute with default value and validator will not be created (v2)
### Initial Checks

### Description

This is related to pydantic v2.

Having a BaseModel with an attribute with a default value and a validator method does not create the attribute when the model gets instantiated. Output of the example below:

```bash
  File "...", line 10, in <module>
    assert FooModel().foo == 42
AttributeError: 'FooModel' object has no attribute 'foo'
```

If you remove the validator or pass a value for `foo` to `__init__` then everything works as expected.
Is this a known issue with needs to be fixed or should it actually already work?
I also tried to directly create the schemaValidator which does return an empty dictionary, so perhaps it is more related to pydantic-core than pydantic.

### Example Code

```Python

class FooModel(BaseModel):
    foo: int = 42
    
    @validator("foo")
    def check_foo(cls, v, **kwargs):
        return v

assert FooModel().foo == 42
```

```Text
pydantic version: 2.0.0.dev0
install path: ...
python version: 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]
platform: Windows-10-10.0.19044-SP0
```

### Affected Components

adriangb:
Digging around through tests that we can make pass. A good starting point to be able to tackle #5068 as well.

Fixes #5045

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -4,9 +4,12 @@ Logic related to validators applied to models etc. via the `@validator` and `@ro
 from __future__ import annotations as _annotations
 
 import warnings
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Generic, TypeVar
+from functools import wraps
+from inspect import Parameter, signature
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Generic, List, TypeVar, Union, cast
 
-from pydantic_core.core_schema import JsonReturnTypes, WhenUsed
+from pydantic_core.core_schema import JsonReturnTypes, ValidationInfo, ValidatorFunction, WhenUsed
+from typing_extensions import Protocol
 
 from ..errors import PydanticUserError
 
@@ -133,7 +136,15 @@ class DecoratorFunctions(Generic[DecFunc]):
         Set functions in self._decorators, now that the class is created and functions are bound.
         """
         for name, decorator in self._decorators.items():
-            decorator.function = getattr(cls, name)
+            func = getattr(cls, name)
+            if isinstance(decorator, Validator):
+                try:
+                    decorator.function = make_generic_validator(func, decorator.mode)
+                except Exception as e:
+                    print(e)
+                    raise
+            else:
+                decorator.function = func
 
     def get_root_decorators(self) -> list[DecFunc]:
         return [self._decorators[name] for name in self._root_decorators]
@@ -218,3 +229,106 @@ def in_ipython() -> bool:
         return False
     else:  # pragma: no cover
         return True
+
+
+class OnlyValueValidator(Protocol):
+    def __call__(self, __value: Any) -> Any:  # pragma: no cover
+        ...
+
+
+class V1ValidatorWithValues(Protocol):
+    def __call__(self, __value: Any, values: dict[str, Any]) -> Any:  # pragma: no cover
+        ...
+
+
+class V1ValidatorWithValuesKwOnly(Protocol):
+    def __call__(self, __value: Any, *, values: dict[str, Any]) -> Any:  # pragma: no cover
+        ...
+
+
+class V1ValidatorWithKwargs(Protocol):
+    def __call__(self, __value: Any, **kwargs: Any) -> Any:  # pragma: no cover
+        ...
+
+
+class V1ValidatorWithKwargsAndValue(Protocol):
+    def __call__(self, __value: Any, values: dict[str, Any], **kwargs: Any) -> Any:  # pragma: no cover
+        ...
+
+
+class V1ValidatorWithValuesAndKwargs(Protocol):
+    def __call__(self, __value: Any, values: dict[str, Any], **kwargs: Any) -> Any:  # pragma: no cover
+        ...
+
+
+V1Validator = Union[
+    V1ValidatorWithValues, V1ValidatorWithValuesKwOnly, V1ValidatorWithKwargs, V1ValidatorWithValuesAndKwargs
+]
+
+
+def make_generic_validator(
+    validator: V1Validator | OnlyValueValidator | ValidatorFunction, mode: str
+) -> ValidatorFunction:
+    """
+    Make a generic function which calls a validator with the right arguments.
+    """
+    sig = signature(validator)
+
+    positional_params: List[str] = []
+    keyword_only_params: List[str] = []
+    accepts_kwargs = False
+    for param_name, parameter in sig.parameters.items():
+        if param_name in ('field', 'config'):
+            raise TypeError(
+                'The `field` and `config` parameters are not available in Pydantic V2.'
+                ' Please use the `info` parameter instead.'
+                ' You can access the configuration via `info.config`,'
+                ' but it is a dictionary instead of an object like it was in Pydantic V1.'
+                ' The `field` argument is unavailable.'
+            )
+        if parameter.kind in (Parameter.POSITIONAL_ONLY, Parameter.POSITIONAL_OR_KEYWORD):
+            positional_params.append(param_name)
+        elif parameter.kind is Parameter.KEYWORD_ONLY:
+            keyword_only_params.append(param_name)
+        elif parameter.kind is Parameter.VAR_KEYWORD:
+            accepts_kwargs = True
+
+    accepts_values_kw = (keyword_only_params == ['values'] and len(positional_params) == 1) or (
+        len(positional_params) == 2 and positional_params[1] == 'values'
+    )
+
+    if accepts_kwargs and len(positional_params) == 1:
+        # although this could be compatible with the V2 signature we want to discourage it,
+        # so we treat it as a backwards compatible validator
+        validator = cast(V1ValidatorWithKwargs, validator)
+
+        @wraps(validator)
+        def _wrapper1(value: Any, info: ValidationInfo) -> Any:
+            return validator(value, values=info.data)  # type: ignore[call-arg]
+
+        return _wrapper1
+    if len(positional_params) == 1 and keyword_only_params == []:
+        validator = cast(OnlyValueValidator, validator)
+
+        @wraps(validator)
+        def _wrapper2(value: Any, info: ValidationInfo) -> Any:
+            return validator(value)  # type: ignore[call-arg]
+
+        return _wrapper2
+    elif len(positional_params) in (1, 2) and accepts_values_kw:
+        validator = cast(V1ValidatorWithValues, validator)
+
+        @wraps(validator)
+        def _wrapper3(value: Any, info: ValidationInfo) -> Any:
+            return validator(value, values=info.data)  # type: ignore[call-arg]
+
+        return _wrapper3
+    elif keyword_only_params == [] and len(positional_params) == 2:
+        validator = cast(ValidatorFunction, validator)
+        return validator
+    raise TypeError(
+        f'Unsupported signature for {mode} validator {validator}: {sig} is not supported.'
+        ' Validators must be compatible with one of the following two signatures:\n'
+        ' - (__value: Any, __info: pydantic.ValidationInfo) -> Any\n'
+        ' - (__value: Any, values: dict[str, Any]) -> Any'
+    )
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -3,6 +3,7 @@ Logic related to validators applied to models etc. via the `@validator` and `@ro
 """
 from __future__ import annotations as _annotations
 
+import inspect
 import warnings
 from functools import wraps
 from inspect import Parameter, signature
@@ -203,20 +204,26 @@ class SerializationFunctions(DecoratorFunctions[Serializer]):
 _FUNCS: set[str] = set()
 
 
-def prepare_decorator(function: Callable[..., Any], allow_reuse: bool) -> classmethod[Any]:
+def prepare_decorator(function: Callable[..., Any], allow_reuse: bool) -> Any:
     """
     Convert the function to a classmethod if it isn't already.
 
     Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
     """
-    f_cls = function if isinstance(function, classmethod) else classmethod(function)
+    first_param = next(iter(inspect.signature(function).parameters.values()))
+    ret: Any
+    if first_param.name == 'cls':
+        ret = function if isinstance(function, classmethod) else classmethod(function)
+        function = ret.__func__
+    else:
+        ret = function
     if not allow_reuse and not in_ipython():
-        ref = f'{f_cls.__func__.__module__}::{f_cls.__func__.__qualname__}'
+        ref = f'{function.__module__}::{function.__qualname__}'
         if ref in _FUNCS:
             warnings.warn(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
         _FUNCS.add(ref)
-    return f_cls
+    return ret
 
 
 def in_ipython() -> bool:
@@ -274,6 +281,23 @@ def make_generic_validator(
     """
     sig = signature(validator)
 
+    def _warn_v1_validator() -> None:
+        warnings.warn(
+            'Validator signatures using the `values` keyword argument or `**kwargs` are no longer supported.'
+            ' Please use an `info: pydantic.ValidationInfo` as the second positional argument instead.'
+            ' This compatibility shim may be removed in a future minor release of Pydantic v2.X',
+            DeprecationWarning,
+            # The stacklevel parameter makes the warning show up as coming from the User's code instead
+            # of our internal implementation
+            # Since this just goes up the stack the source location will appear as
+            # class TheModelName(BaseModel):
+            # Which is good enough for now in terms of helping users locate the issue
+            # In the future maybe we can capture the calling module and line number in the
+            # @validator decorator and use that here to be more accurate about where the issue is
+            # But that may not be 100% reliable, so tabling for now.
+            stacklevel=6,
+        )
+
     positional_params: List[str] = []
     keyword_only_params: List[str] = []
     accepts_kwargs = False
@@ -302,6 +326,8 @@ def make_generic_validator(
         # so we treat it as a backwards compatible validator
         validator = cast(V1ValidatorWithKwargs, validator)
 
+        _warn_v1_validator()
+
         @wraps(validator)
         def _wrapper1(value: Any, info: ValidationInfo) -> Any:
             return validator(value, values=info.data)  # type: ignore[call-arg]
@@ -317,6 +343,8 @@ def make_generic_validator(
         return _wrapper2
     elif len(positional_params) in (1, 2) and accepts_values_kw:
         validator = cast(V1ValidatorWithValues, validator)
+
+        _warn_v1_validator()
 
         @wraps(validator)
         def _wrapper3(value: Any, info: ValidationInfo) -> Any:
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -231,7 +231,7 @@ def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool)
         ret = function if isinstance(function, classmethod) else classmethod(function)
         function = ret.__func__
     elif first_param.name == 'self':
-        raise TypeError('Validators cannot be instance methods (they should not accept self as an argument)')
+        raise TypeError('Validators cannot be instance methods (they should not accept `self` as an argument)')
     else:
         ret = function if isinstance(function, staticmethod) else staticmethod(function)
         function = ret.__func__
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -230,7 +230,7 @@ def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool)
 
 def prepare_serializer_decorator(function: Callable[..., Any], allow_reuse: bool) -> Any:
     """
-    Convert the function to a classmethod if it isn't already.
+    Register the serializer so we can warn if it is re-used.
 
     Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -26,6 +26,7 @@ __all__ = (
     'ValidationFunctions',
     'SerializationFunctions',
     'prepare_validator_decorator',
+    'prepare_serializer_decorator',
 )
 FIELD_VALIDATOR_TAG = '_field_validator'
 ROOT_VALIDATOR_TAG = '_root_validator'
@@ -200,6 +201,21 @@ class SerializationFunctions(DecoratorFunctions[Serializer]):
 _FUNCS: set[str] = set()
 
 
+def prepare_serializer_decorator(function: Callable[..., Any], allow_reuse: bool) -> classmethod[Any]:
+    """
+    Convert the function to a classmethod if it isn't already.
+    Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
+    which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
+    """
+    f_cls = function if isinstance(function, classmethod) else classmethod(function)
+    if not allow_reuse and not in_ipython():
+        ref = f'{f_cls.__func__.__module__}::{f_cls.__func__.__qualname__}'
+        if ref in _FUNCS:
+            warnings.warn(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
+        _FUNCS.add(ref)
+    return f_cls
+
+
 def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool) -> Any:
     """
     Convert the function to a classmethod if it isn't already.
@@ -226,24 +242,6 @@ def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool)
             warnings.warn(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
         _FUNCS.add(ref)
     return ret
-
-
-def prepare_serializer_decorator(function: Callable[..., Any], allow_reuse: bool) -> Any:
-    """
-    Register the serializer so we can warn if it is re-used.
-
-    Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
-    which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
-    """
-    fn = function
-    if isinstance(fn, staticmethod):
-        fn = fn.__func__
-    if not allow_reuse and not in_ipython():
-        ref = f'{fn.__module__}::{fn.__qualname__}'
-        if ref in _FUNCS:
-            warnings.warn(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
-        _FUNCS.add(ref)
-    return function
 
 
 def in_ipython() -> bool:
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -218,9 +218,8 @@ def prepare_serializer_decorator(function: Callable[..., Any], allow_reuse: bool
 
 def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool) -> Any:
     """
-    Convert the function to a classmethod if it isn't already.
-
-    Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
+    Apply the @classmethod or @staticmethod decorator to @validator functions if it was not applied already.
+    Warn about validators with duplicated names since without this, they can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
     """
     first_param = next(iter(inspect.signature(function).parameters.values()))
@@ -229,7 +228,7 @@ def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool)
         ret = function if isinstance(function, classmethod) else classmethod(function)
         function = ret.__func__
     elif first_param.name == 'self':
-        ret = function
+        raise TypeError('Validators cannot be instance methods (they should not accept self as an argument)')
     else:
         ret = function if isinstance(function, staticmethod) else staticmethod(function)
         function = ret.__func__
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -222,7 +222,10 @@ def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool)
     Warn about validators with duplicated names since without this, they can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
     """
-    first_param = next(iter(inspect.signature(function).parameters.values()))
+    sig = inspect.signature(function)
+    first_param = next(iter(sig.parameters.values()), None)
+    if first_param is None:
+        raise TypeError(f'Unrecognized validator signature {sig} for {function}')
     ret: Any
     if first_param.name == 'cls':
         ret = function if isinstance(function, classmethod) else classmethod(function)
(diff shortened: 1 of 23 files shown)


pydantic/_internal/_decorators.py
```diff
@@ -218,3 +226,107 @@ def in_ipython() -> bool:
...
+            continue
+        if parameter.kind == Parameter.VAR_KEYWORD:
+            accepts_variable_kwargs = True
+            break
+        elif parameter.kind == Parameter.VAR_POSITIONAL:
+            raise TypeError  # TODO: Error message
+        if parameter_name in {'values', 'config', 'context'}:
+            kwargs.add(parameter_name)
+        else:
+            raise TypeError  # TODO: Error message
```
adriangb, line 261: Note to self: complete these TODOs

tests/test_json_schema.py
```diff
@@ -1741,7 +1741,7 @@ class Model(BaseModel):
         something: Optional[int] = None
 
         @validator('something')
-        def check_field(cls, v, *, values, config, field):
+        def check_field(cls, v, *, values, config, context):
```
adriangb, line 1744: Is it possible to get the field that is being validated in V2? Does it come in `context`?

adriangb (commented):
I assumed we are keeping `values` and not cutting over to `data` (which is what `pydantic-core` is using). I also allowed `data`, `config` and `context` to all be keyword-only or keyword-or-positional, even though `pydantic-core` wants them to be keyword only ([here](https://github.com/pydantic/pydantic-core/blob/ee3da4c693f2b7b6a2a6ff2fdb933457014986fe/pydantic_core/core_schema.py#L1434), should we change that? The implementation doesn't require it).

dmontagu:
>I assumed we are keeping values and not cutting over to data

My weakly-held preference would be to keep things as backwards-compatible as possible, so assuming this is possible and helps with backwards compatibility, that would be my vote.

adriangb:
This PR will need an update after https://github.com/pydantic/pydantic-core/issues/422 to account for the new signature, but the approach will be generally the same.

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +226,107 @@ def in_ipython() -> bool:
...
+        ) -> Any:
+            return validator(__input_value, context=context, config=config)
+
+        return wrapper
+
+    # kwargs == {'values', 'field', 'config'}
+    def wrapper(  # type: ignore[no-redef]
+        __input_value: Any, *, data: Any, config: CoreConfig | None, context: Any, **future_kwargs: Any
+    ) -> Any:
+        return validator(__input_value, values=data, context=context, config=config)
```
adriangb, line 330: We need to make this an `else` and then handle the deprecated cases (`field` primarily)`.

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +226,107 @@ def in_ipython() -> bool:
...
+            return validator(__input_value, context=context)
+
+    elif kwargs == {'config'}:
+
+        def wrapper(
+            __input_value: Any, *, data: Any, config: CoreConfig | None, context: Any, **future_kwargs: Any
+        ) -> Any:
+            return validator(__input_value, config=config)
+
+        return wrapper
```
adriangb, line 300: Do we want to raise an error for `config`? Or should we name this argument something else (`core_config`?) since it is not the same object as V1? Not sure how we handled this breaking change elsewhere with `Config` (e.g. `model.model_config`)?

dmontagu: Good question.

I am leaning toward probably a good idea to change it, just to reduce confusion during migration.

Another thing to note — we should probably ensure that the typing is set up in a way that the validator decorator will error if the item whose signature it is wrapping has the wrong signature. I think this should be possible by annotating the type of the input to `Validator` with a [callback protocol](https://mypy.readthedocs.io/en/stable/protocols.html#callback-protocols).

Will it be possible to maintain (deprecated) backwards compatibility with validators that currently define the config argument? If so, I think that would be an even stronger argument for giving it a different name here (and `core_config` doesn't seem unreasonable to me, though `cfg` might be nicer for reducing typing without losing clarity).

adriangb: > Will it be possible to maintain (deprecated) backwards compatibility with validators that currently define the config argument?

yes in the sense that things won't crash calling the function. no in the sense that if they used the `config` argument at all things will fail since it's a different object. so yeah I'm leaning towards not making the signature backwards compatible.

We wouldn't add a new `cfg` parameter, we'd say use `info`.

dmontagu: > Or should we name this argument something else (core_config?)

If we are considering renaming to `core_config`, I'm just proposing that `cfg` might be a reasonable (and less verbose) alternative. If we'd say 'use `info`' instead of having a `cfg` argument, I'd probably suggest we error instead of changing to `core_config` as an argument.

Don't have a strong opinion though.

adriangb: I went with explicitly erroring if `config` is an acceptable keyword argument.

pydantic/_internal/_decorators.py
adriangb: Done

tests/test_generics.py
```diff
@@ -437,16 +437,15 @@ class Result(BaseModel, Generic[data_type, error_type]):
         positive_number: int
 
         @validator('error')
-        def validate_error(cls, v: Optional[error_type], **kwargs) -> Optional[error_type]:
```
adriangb, line 440: @dmontagu can you confirm this change is right since you just wrote the test?

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +229,106 @@ def in_ipython() -> bool:
...
+    keyword_only_params: List[str] = []
+    accepts_kwargs = False
+    for param_name, parameter in sig.parameters.items():
+        if param_name in ('field', 'config'):
+            raise TypeError(
+                'The `field` and `config` parameters are not available in Pydantic V2.'
+                ' Please use the `info` parameter instead.'
+                ' You can access the configuration via `info.config`,'
+                ' but it is a dictionary instead of an object like it was in Pydantic V1.'
+                ' The `field` argument is unavailable.'
```
adriangb, line 287: We can edit this in the future if we add `field`

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +229,106 @@ def in_ipython() -> bool:
...
+
+        return _wrapper1
+    if len(positional_params) == 1 and keyword_only_params == []:
+        validator = cast(OnlyValueValidator, validator)
+
+        @wraps(validator)
+        def _wrapper2(value: Any, info: ValidationInfo) -> Any:
+            return validator(value)  # type: ignore[call-arg]
+
+        return _wrapper2
```
adriangb, line 317: I think this is a candidate for passing an argument along the lines of `accepts_info_param` to Rust just to cut out one more Python function call. Either way, I'd keep this signature option

samuelcolvin: Yup, I'm up for `accepts_info_param` being added to the core-schema.

tests/test_generics.py
dmontagu: as far as I can tell this looks good to me, with one exception (that I don't think is your fault): below, it copies this twice:
```python

        Result[Data, Error](data=[Data(number=1, text='a')], error=Error(message='error'), positive_number=1)
        {
            'type': 'value_error',
            'loc': ('error',),
            'msg': 'Value error, Must provide data or error',
            'input': Error(message='error'),
            'ctx': {'error': 'Must provide data or error'},
        }
    ]
```

I think the second time, it should be more like:
```python

        Result[Data, Error](positive_number=1)
        {
            'type': 'value_error',
            'loc': ('error',),
            'msg': 'Value error, Must not provide both data and error',
            'input': Error(message='error'),
            'ctx': {'error': 'Must not provide both data and error'},
        }
    ]
```

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +229,106 @@ def in_ipython() -> bool:
...
+
+
+class V1ValidatorWithValuesAndKwargs(Protocol):
+    def __call__(self, __value: Any, values: dict[str, Any], **kwargs: Any) -> Any:  # pragma: no cover
+        ...
+
+
+V1Validator = Union[
+    V1ValidatorWithValues, V1ValidatorWithValuesKwOnly, V1ValidatorWithKwargs, V1ValidatorWithValuesAndKwargs
+]
```
dmontagu, line 266: Are we emitting any kind of deprecation warning if a V1 signature is used? We'll probably want to eliminate the big union at some point if it makes mypy errors highly verbose, just want to make sure we put something in place to let us do that. (Or decide we intend to support these signatures long term, I guess)

dmontagu: (Can add the warning later if you just want to get this merged sooner and don't want the distraction — assuming you and Samuel agree on that point anyway.)

pydantic/_internal/_std_types_schema.py
```diff
@@ -169,7 +169,7 @@ def _deque_ser_schema(inner_schema: core_schema.CoreSchema | None = None) -> cor
 
 def _deque_any_schema() -> core_schema.FunctionWrapSchema:
     return core_schema.function_wrap_schema(
-        _validators.deque_any_validator, core_schema.list_schema(), serialization=_deque_ser_schema()
+        _validators.deque_any_validator, core_schema.list_schema(allow_any_iter=True), serialization=_deque_ser_schema()
```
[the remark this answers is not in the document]
adriangb: Hmm. Well there was a test for a set. Alternatively I can make a union schema.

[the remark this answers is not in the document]
adriangb: union of list, tuple and set worked

pydantic/_internal/_generate_schema.py
```diff
@@ -247,11 +247,12 @@ def generate_field_schema(
         schema = self.generate_schema(field_info.annotation)
         schema = apply_annotations(schema, field_info.metadata)
 
+        schema = apply_validators(schema, validator_functions.get_field_decorators(name))
```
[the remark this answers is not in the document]
adriangb: There is a reason. Consider this pydantic-core example:

```python
from typing import Any
from pydantic_core import core_schema, SchemaValidator


def val(v: Any, info: core_schema.ValidationInfo) -> Any:
    return v


schema1 = core_schema.typed_dict_schema(
    fields={
        "a": core_schema.typed_dict_field(
            schema=core_schema.function_after_schema(
                function=val,
                schema=core_schema.with_default_schema(
                    core_schema.nullable_schema(core_schema.int_schema()), default=None
                ),
            ),
            required=False
        )
    }
)

v = SchemaValidator(schema1)
assert v.validate_python({}) == {}


schema2 = core_schema.typed_dict_schema(
    fields={
        "a": core_schema.typed_dict_field(
            schema=core_schema.with_default_schema(
                schema=core_schema.function_after_schema(
                    function=val, schema=core_schema.nullable_schema(core_schema.int_schema())
                ),
                default=None,
            ),
            required=False
        )
    }
)

v = SchemaValidator(schema2)
assert v.validate_python({}) == {"a": None}
```

`schema1` is what gets generated in main and `schema2` in this branch. Also corresponds to this example:

```python
class Model(BaseModel):
    x: Optional[int] = None

    @validator("x")
    def val_x(cls, v: Any) -> Any:
        return v

print(Model())
```

On `main` it gives an attribute error.

I didn't dig too deep but I'm guessing it's because the `required=False` option to the field only works if the inner schema supports a default value, which the `function` schemas do not.

Is this okay to keep then or do you consider this is a bug in `pydantic-core`? I think this could be it's own PR but since it depends on `@validator` it might be complicated to split out.

pydantic/_internal/_decorators.py
```diff
@@ -133,7 +136,15 @@ def set_bound_functions(self, cls: type[BaseModel]) -> None:
         Set functions in self._decorators, now that the class is created and functions are bound.
         """
         for name, decorator in self._decorators.items():
-            decorator.function = getattr(cls, name)
+            func = getattr(cls, name)
+            if isinstance(decorator, Validator):
+                try:
+                    decorator.function = make_generic_validator(func, decorator.mode)
+                except Exception as e:
+                    print(e)
```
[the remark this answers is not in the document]
adriangb: yep got rid of it hehe

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +229,106 @@ def in_ipython() -> bool:
         return False
     else:  # pragma: no cover
         return True
+
+
+class OnlyValueValidator(Protocol):
+    def __call__(self, __value: Any) -> Any:  # pragma: no cover
```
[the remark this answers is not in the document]
adriangb: done

pydantic/_internal/_decorators.py
adriangb: Deprecation warnings added

pydantic/_internal/_decorators.py
adriangb: tabling this for a future PR, it's really just a performance opt

tests/test_generics.py
adriangb: removed the duplication

pydantic/decorators.py
```diff
@@ -85,26 +182,31 @@ def root_validator(
...
-        f_cls = _decorators.prepare_decorator(f, allow_reuse)
+        f_cls = _decorators.prepare_validator_decorator(f, allow_reuse)
         setattr(f_cls, _decorators.ROOT_VALIDATOR_TAG, _decorators.Validator(mode=mode))
         return f_cls
 
     return dec
 
 
+_SerializerFunctionType = TypeVar(
+    '_SerializerFunctionType', bound=Union[_core_schema.SerializeWrapFunction, _core_schema.SerializePlainFunction]
```
adriangb, line 198: Important to note: I am explicitly prohibiting instance method serializers, at least for now, since they’re broken and if we want to do that right I think we need to rethink some things

tests/test_validators.py
```diff
@@ -111,30 +173,30 @@ class Model(BaseModel):
         b: int
         a: List[int]
 
+        # TODO: do we support `each_item`?
```
dmontagu, line 176: Does this currently work? Just curious

tests/test_validators.py
```diff
@@ -236,9 +310,9 @@ class ModelTwo(BaseModel):
         b: int
 
         @validator('b')
-        def validate_b(cls, b, values):
-            if 'm' in values:
-                return b + values['m'].a  # this fails if values['m'] is a dict
+        def validate_b(cls, b, info: ValidationInfo):
+            if 'm' in info.data:
+                return b + info.data['m'].a  # this fails if values['m'] is a dict
```
dmontagu, line 315: super minor, probably makes sense to update the comment

tests/test_validators.py
```diff
@@ -344,8 +412,9 @@ def test_validate_always():
     class Model(BaseModel):
         a: str = None
 
-        @validator('a', pre=True, always=True)
-        def check_a(cls, v):
+        # TODO: do we support always? I think we now always validate always
```
dmontagu, line 415: I don't have the answer to this question, but I'll draw attention to the question ☝️ @samuelcolvin

tests/test_validators.py
```diff
@@ -1020,11 +1085,11 @@ class Model(BaseModel):
         b: str
 
         @validator('a', allow_reuse=allow_1)
-        def duplicate_name(cls, v):
+        def duplicate_name(cls, v: Any):
```
dmontagu, line 1088: I've noticed you've replaced `v` with `v: Any` in most of these signatures — is that necessary now, or just something you are in the habit of doing? I don't object or anything, just curious

tests/test_validators.py
```diff
@@ -1404,3 +1467,120 @@ def set_area(cls, values):
...
+    def _check_x(y: Any, values: Dict[str, Any], **kwargs: Any) -> int:
+        raise NotImplementedError
+
+    with pytest.warns(DeprecationWarning, match=V1_VALIDATOR_DEPRECATION_MATCH) as w:
+
+        class Point(BaseModel):
+            y: int
+            x: int
+
+            check_x = validator('x')(_check_x)
```
dmontagu, line 1562: Do you have tests that these deprecated validator signatures work properly? It looks like this is just testing that the warning is emitted, but not that the validator actually works as intended. Seems to me to be worth adding tests that the v1 validator behavior still works with the deprecated signature, at least as long as we intend for it to. (And I'm assuming we do for the purpose of migration-friendliness.)

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +256,127 @@ def in_ipython() -> bool:
...
+    def __call__(self, __value: Any, **kwargs: Any) -> Any:
+        ...
+
+
+class V1ValidatorWithKwargsAndValue(Protocol):
+    def __call__(self, __value: Any, values: dict[str, Any], **kwargs: Any) -> Any:
+        ...
+
+
+class V1ValidatorWithValuesAndKwargs(Protocol):
```
dmontagu, line 286: I noticed all of these Protocols seem to be defined in _both_ `pydantic._internal._decorators` and `pydantic.decorators` — any chance we can unify and get them defined in just one place?

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +256,127 @@ def in_ipython() -> bool:
...
+    validator: V1Validator | OnlyValueValidator | ValidatorFunction, mode: str
+) -> ValidatorFunction:
+    """
+    Make a generic function which calls a validator with the right arguments.
+    """
+    sig = signature(validator)
+
+    def _warn_v1_validator() -> None:
+        warnings.warn(
+            'Validator signatures using the `values` keyword argument or `**kwargs` are no longer supported.'
```
dmontagu, line 306:
```suggestion
            'One of the validators in this class has a signature that uses the `values` keyword argument or `**kwargs`, which are no longer supported.'
```

This is perhaps splitting hairs a bit, but I'm thinking that if (according to the note below) the warning will just show that it is coming from the class, and not a specific line, it might be worth the extra bit of clarity in this warning.

Actually — is it the case that this warning can only be produced in the context of defining a class? If so, I think my suggestion might be sensible, but if it can be produced in other contexts it might be worth confirming that the `stacklevel` keyword argument is still doing something reasonable.

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +256,127 @@ def in_ipython() -> bool:
...
+    keyword_only_params: List[str] = []
+    accepts_kwargs = False
+    for param_name, parameter in sig.parameters.items():
+        if param_name in ('field', 'config'):
+            raise TypeError(
+                'The `field` and `config` parameters are not available in Pydantic V2.'
+                ' Please use the `info` parameter instead.'
+                ' You can access the configuration via `info.config`,'
+                ' but it is a dictionary instead of an object like it was in Pydantic V1.'
+                ' The `field` argument is unavailable.'
```
dmontagu, line 331:
```suggestion
                ' The `field` argument is no longer available.'
```
I think this makes it more clear that it was purposely removed. I think we will need to document how to migrate if people were using this in their own v1 validators — is there a migration path for this? If not, do we need to add this back in?

pydantic/_internal/_decorators.py
```diff
@@ -192,20 +200,50 @@ class SerializationFunctions(DecoratorFunctions[Serializer]):
...
         if ref in _FUNCS:
             warnings.warn(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
         _FUNCS.add(ref)
-    return f_cls
+    return ret
+
+
+def prepare_serializer_decorator(function: Callable[..., Any], allow_reuse: bool) -> Any:
+    """
+    Convert the function to a classmethod if it isn't already.
```
dmontagu, line 233: Does this docstring need changing? It looks like a copy of the above, and I don't see anything happening below that looks like conversion into a classmethod, but maybe I'm missing something.

Maybe this is related to what we talked about with not forcing serializer methods to be classmethods; either way probably makes sense to at least add a TODO in this docstring saying to update it once we have more clarity.

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +256,127 @@ def in_ipython() -> bool:
...
+V1Validator = Union[
+    V1ValidatorWithValues, V1ValidatorWithValuesKwOnly, V1ValidatorWithKwargs, V1ValidatorWithValuesAndKwargs
+]
+
+
+def make_generic_validator(
+    validator: V1Validator | OnlyValueValidator | ValidatorFunction, mode: str
+) -> ValidatorFunction:
+    """
+    Make a generic function which calls a validator with the right arguments.
```
dmontagu, line 300: I personally would appreciate a little more description about why this is necessary; I looked at the use site and it wasn't immediately obvious. (Not suggesting this function isn't necessary — just that it might help me, or others reading this code, get our bearings to have a little more context about what's happening.)

I think it could say something to the effect of:
```
In order to support different signatures, including deprecated validator signatures from v1,
we introspect the function signature and wrap it in a parent function that has a signature
compatible with pydantic_core
```
(Or something similar that is more detailed/accurate, if the above is wrong)

dmontagu (commented):
Looking good, most of my notes here are more about documentation than much of substance (I guess with the exception of the point about duplicated decorators, and about adding tests for the deprecated validator signatures).

I think it's worth merging this soon to keep `main` up to date even if there's more to do, or other changes that we know we'll want to make.

pydantic/_internal/_decorators.py
dmontagu: I just read https://github.com/pydantic/pydantic/issues/5170 — I think if we adopt the approach Samuel described there, it's probably worth updating this message to indicate how to access the relevant data (i.e., using `cls.model_fields`).

tests/test_validators.py
adriangb: It doesn't work. `@validator` didn't accept an `each_item` param before this PR and this PR is not adding it. As discussed in https://github.com/pydantic/pydantic/issues/4707 I think we should actually not support this and instead make you apply validators to child items using `Annotated`.

tests/test_validators.py
adriangb: Not necessary put since `@validator` has better typing now Pyright will complain at you if the signature is unknown. So I'm mainly adding it for our own sanity.

pydantic/_internal/_decorators.py
adriangb: Unfortunately, no. The ones in `pydantic.decorators` add a `cls` argument since we do the magic classmethod conversion.

Which does bring up the topic: @dmontagu you had said you backed the idea of removing this magic and instead requiring users to add an `@classmethod` themselves.

I do think maybe we could add more overloads and say "if you give me a function with `cls` as the first argument I will give you back a classmethod, otherwise I will give you back the same thing I got.

pydantic/_internal/_decorators.py
adriangb: I think it might show up as well if you do `create_model`. In that case it would look like it's coming from within `create_model`, which I think is fine enough...

pydantic/_internal/_decorators.py
adriangb, line 300:
```suggestion
    In order to support different signatures, including deprecated validator signatures from v1,
    we introspect the function signature and wrap it in a parent function that has a signature
    compatible with pydantic_core
```

adriangb: Done

tests/test_validators.py
dmontagu: Oops I missed the skipped lines and thought this test had its `xfail` removed.

I added a note about this to the migration guide points ("Describe how to replicate the functionality of each_item validators").

Does applying the validators to the child items using `Annotated` work now? Just thinking we could update this test to do that, or at least update the `TODO:` comment to say that the test should eventually be updated to do something like that.

pydantic/_internal/_decorators.py
adriangb, line 233:
```suggestion
    Register the serializer so we can warn if it is re-used.
```

tests/test_validators.py
adriangb: Not yet. That needs us to create the `PlainValidator` and others metadata.

One important thing to note is that the item validators won't have access to `values: Dict[str, Any]` or the model class. I think that seems like a reasonable limitation. We have the validation context where if you want to do something weird like that you can.

pydantic/_internal/_decorators.py
dmontagu: The main reason I don't like the idea of converting methods into classmethods using our own decorator is that I think it won't play nice with editors (I've found myself annoyed about that in the past).

I'm not sure whether that is relevant to the point about a function-with-cls-as-first-argument or not, I think I could be convinced about any behavior with free-standing functions.

pydantic/_internal/_decorators.py
dmontagu: For what it's worth, my understanding was that the plan was to get rid of `create_model`. Not 100% sure about that though.

adriangb: Even if we don't get rid of it I'd still consider it a somewhat advanced use case and at that point you should be able to figure out that the warning is caused by you 1 stack frame up.

pydantic/_internal/_decorators.py
adriangb: I don't like it either, but we need it for V1 compatibility. I organized the types so that we can easily deprecate that approach in the future. I didn't add a deprecation warning here but we can. The adding overloads point was about maybe making the situation with IDEs better. FWIW though Pylance seems to be very happy with what this PR is currently doing.

tests/test_validators.py
adriangb: https://github.com/pydantic/pydantic/blob/73088c94368f7ce29e779f640b892c563b774b88/tests/test_validators.py#L1495-L1586

dmontagu: These tests aren't checking that the validator is actually called though, right? Am I missing something? I meant something like:
```python
def _check_x(v):
    return v * 2

assert Point(x=1, y=1).model_dump() == {'x': 2, 'y': 1}
```
(Mostly just making sure it actually calls the validator during validation without erroring.)

If you think it's unnecessary I'm fine, just felt like it was possible that these tests could pass and yet the actual validation cause errors for other reasons.

tests/test_validators.py
dmontagu: Yeah I think it's reasonable for us to say that if you want to access that stuff, you just have to define the function in a way that it applies to the "full" field value, rather than individual items within it.

tests/test_validators.py
adriangb: No it’s a good idea! I’ll add them

adriangb: Donzo, thank you for the good idea!

adriangb:
I'm going to hold this off until https://github.com/pydantic/pydantic-core/pull/437 is ready to reduce the diff.

pydantic/_internal/_generate_schema.py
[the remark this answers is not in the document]
adriangb: Indeed the logic in `pydantic-core` seems hardcoded to check if the innner validator to `typed-dict-field` has a default value. Since this is not recursive and `function_after_schema` does not have a default value it just doesn't use one in the first case. It makes sense that switching the order fixes the problem.

dmontagu (approved):

pydantic/_internal/_generate_schema.py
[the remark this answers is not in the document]
samuelcolvin: I think this is the same as #5045 - I was assuming we'd have to fix that in pydantic-core.

Please can you check if you're fix here resolves #5045, if so mark at issue to be fixed by this PR.

If not, I'll try to work on that soon - or alternatively, explain how I would fix it.

tests/test_validators.py
samuelcolvin: `always` was weird on V1, e.g. we didn't run validation if the input value was `None` unless `always=True`, but it also caused validation to be fun on the default value.

Validating the default in pydantic-core, see https://github.com/pydantic/pydantic-core/issues/358

samuelcolvin (commented):
Overall this is looking great.

The many thing it makes me think is that we should allow pydantic-core to omit the `info` argument ASAP.

A few small things to fix.

pydantic/_internal/_generate_schema.py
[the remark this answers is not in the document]
adriangb: Yep this does fix #5045!

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +216,34 @@ def prepare_decorator(function: Callable[..., Any], allow_reuse: bool) -> classm
...
+    Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
+    which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
+    """
+    first_param = next(iter(inspect.signature(function).parameters.values()))
+    ret: Any
+    if first_param.name == 'cls':
+        ret = function if isinstance(function, classmethod) else classmethod(function)
+        function = ret.__func__
+    elif first_param.name == 'self':
+        ret = function
```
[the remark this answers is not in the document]
adriangb: Done

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +216,34 @@ def prepare_decorator(function: Callable[..., Any], allow_reuse: bool) -> classm
     return f_cls
 
 
+def prepare_validator_decorator(function: Callable[..., Any], allow_reuse: bool) -> Any:
+    """
+    Convert the function to a classmethod if it isn't already.
```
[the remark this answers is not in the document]
adriangb: updated

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +216,34 @@ def prepare_decorator(function: Callable[..., Any], allow_reuse: bool) -> classm
...
+
+    Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
+    which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
+    """
+    first_param = next(iter(inspect.signature(function).parameters.values()))
```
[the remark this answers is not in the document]
adriangb: handled and added a test

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +254,129 @@ def in_ipython() -> bool:
...
+
+        return _wrapper3
+    elif keyword_only_params == [] and len(positional_params) == 2:
+        validator = cast(ValidatorFunction, validator)
+        return validator
+    raise TypeError(
+        f'Unsupported signature for {mode} validator {validator}: {sig} is not supported.'
+        ' Validators must be compatible with one of the following two signatures:\n'
+        ' - (__value: Any, __info: pydantic.ValidationInfo) -> Any\n'
+        ' - (__value: Any, values: dict[str, Any]) -> Any'
```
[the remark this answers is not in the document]
adriangb: Okay I'll remove it

pydantic/_internal/_decorators.py
```diff
@@ -218,3 +254,129 @@ def in_ipython() -> bool:
...
+
+        return _wrapper2
+    elif len(positional_params) in (1, 2) and accepts_values_kw:
+        validator = cast(V1ValidatorWithValues, validator)
+
+        _warn_v1_validator()
+
+        @wraps(validator)
+        def _wrapper3(value: Any, info: ValidationInfo) -> Any:
+            return validator(value, values=info.data)  # type: ignore[call-arg]
```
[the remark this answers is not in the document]
adriangb: Unless you tell me otherwise, I believe they are the same. There are tests using `values` that seem to be passing using `info.data`

tests/test_validators.py
adriangb: Got it, seems like we're not going to support it in V2 but I'll leave this test commented out until we https://github.com/pydantic/pydantic-core/issues/358 gets done

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +216,36 @@ def prepare_decorator(function: Callable[..., Any], allow_reuse: bool) -> classm
...
+    sig = inspect.signature(function)
+    first_param = next(iter(sig.parameters.values()), None)
+    if first_param is None:
+        raise TypeError(f'Unrecognized validator signature {sig} for {function}')
+    ret: Any
+    if first_param.name == 'cls':
+        ret = function if isinstance(function, classmethod) else classmethod(function)
+        function = ret.__func__
+    elif first_param.name == 'self':
+        raise TypeError('Validators cannot be instance methods (they should not accept self as an argument)')
```
samuelcolvin, line 234:
```suggestion
        raise TypeError('Validators cannot be instance methods (they should not accept `self` as an argument)')
```

samuelcolvin (approved):
Otherwise LGTM.

samuelcolvin:
thanks so much, this is awesome.

Issue #5161, opened 2023-03-09
mypy error "Unexpected keyword argument" on pydantic dataclass InitVar field


brianmedigate:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

Since upgrading to pydantic 1.10.6, mypy started complaining about a kwarg being passed to the constructor of a dataclass if the field is an `InitVar`.

I'm guessing it might be related to https://github.com/pydantic/pydantic/pull/5111

```
$ mypy test.py
test.py:10: error: Unexpected keyword argument "foo" for "MyDataclass"  [call-arg]
Found 1 error in 1 file (checked 1 source file)
```

### Example Code

```Python
from pydantic.dataclasses import dataclass
from dataclasses import InitVar

@dataclass
class MyDataclass:
    foo: InitVar[str]
    bar: str

MyDataclass(foo="foo", bar="bar")
```

### Python, Pydantic & OS Version

```Text
pydantic version: 1.10.6
            pydantic compiled: True
                 install path: /home/brian/repos/medigator/venv/lib/python3.8/site-packages/pydantic
               python version: 3.8.10 (default, Nov 14 2022, 12:59:47)  [GCC 9.4.0]
                     platform: Linux-5.15.0-67-generic-x86_64-with-glibc2.29
     optional deps. installed: ['dotenv', 'email-validator', 'typing-extensions']
```

### Affected Components

- [ ] [Compatibility between releases](https://docs.pydantic.dev/changelog/)
- [ ] [Data validation/parsing](https://docs.pydantic.dev/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://docs.pydantic.dev/usage/exporting_models/) - `.dict()` and `.json()`
- [ ] [JSON Schema](https://docs.pydantic.dev/usage/schema/)
- [X] [Dataclasses](https://docs.pydantic.dev/usage/dataclasses/)
- [ ] [Model Config](https://docs.pydantic.dev/usage/model_config/)
- [ ] [Field Types](https://docs.pydantic.dev/usage/types/) - adding or changing a particular data type
[...]

PR #5162, opened 2023-03-09
Don't apply dataclass transform twice with plugin + mypy 1.1.1

cdce8p:
## Change Summary
Mypy `1.1.1` added support for `@dataclass_transform`. With the plugin active, the mypy dataclass transformer would be executed twice. As the first iteration already removes the `InitVar` annotation, the second one results in an error message.

The PR targets the `1.10.X-fixes` branch as that is much easier to test. If excepted, I'll also open a PR for `main`.

Fixes: #5161

## Checklist

### DIFF
--- a/changes/5162-cdce8p.md
+++ b/changes/5162-cdce8p.md
@@ -1,0 +1,1 @@
+Fix `InitVar` usage with pydantic dataclasses, mypy version `1.1.1` and the custom mypy plugin
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -89,7 +89,7 @@ MYPY_VERSION_TUPLE = parse_mypy_version(mypy_version)
 BUILTINS_NAME = 'builtins' if MYPY_VERSION_TUPLE >= (0, 930) else '__builtins__'
 
 # Increment version if plugin changes and mypy caches should be invalidated
-PLUGIN_VERSION = 1
+PLUGIN_VERSION = 2
 
 
 def plugin(version: str) -> 'TypingType[Plugin]':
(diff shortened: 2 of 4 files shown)


dmontagu:
Would it be possible to add a test demonstrating the changed behavior? I recognize that we aren't actually running tests against 1.1.1 but I would still like to add some code that _would_ be affected if we were (if only for future's sake). If such code already exists, I would appreciate if you could point me at it.

If you aren't comfortable with the mypy plugin testing infrastructure, if you can just show me how to reproduce the issue that this resolves, that might be enough for me to add it to the tests.

dmontagu:
There's an issue with the `type[...]` annotation due to the fact that this module doesn't have the `__future__` import. I'll fix that and push a commit after I merge the generics PR. (Killing the CI run now since it's going to fail)

dmontagu:
Just FYI -- I'm getting test failures on this branch with mypy 0.971. I'm thinking that needs to be resolved before merging.

cdce8p:
Pushed some more changes. Should be ready now
* Replaced the `PydanticPlugin.mypy_version` class variable with the existing `MYPY_VERSION_TUPLE`.
* Renamed `PLUGIN_VERSION` to `__version__`. After looking at the mypy code for plugins a bit more, I saw that mypy actually uses multiple different things to determine if a plugin has changed
  * `report_config_data` for plugin config data
  * A hash of the plugin module file contents
  * The `__version__` attribute defined by the plugin module itself.
  => I.e. mypy should notice a plugin change already even without `PLUGIN_VERSION`. However, we can still keep it around, but don't need to write it to `_plugin_data` by renaming it to `__version__`
* Fixed the mypy test by adding the new errors. Those are emitted with the default mypy config (without the plugin), i.e. those are expected. _This only changes once mypy `1.1.1` is tested as it supports the `@dataclass_transform` decorator.

dmontagu:
I ran this in 1.1.1 and I didn't see the errors, so I'm going to merge. Thanks @cdce8p — I was going to say I wasn't able to get to this today, but great work figuring this all out, very much appreciated!

samuelcolvin:
Looks great, thank you both for sorting this.

PR #5167, opened 2023-03-10
Don't apply dataclass transform twice with plugin + mypy 1.1.1 (#5162)

### LINKED ISSUE #5161: mypy error "Unexpected keyword argument" on pydantic dataclass InitVar field

cdce8p:
(cherry picked from commit 355e6dac6b1881a12cf48783f42dd8a3991437b2)

## Change Summary
Cherry picked fix from `1.10.X-fixes` branch (#5162) onto `main`.
/CC: @dmontagu 

Closes #5161

Selected Reviewer: @hramezani


samuelcolvin:
please update (just so we know who's assigned).

cdce8p:
Seems like `InitVar` causes the crash with pydantic `v2` and mypy. Not sure why

dmontagu:
It seems to me that the inclusion of `InitVar` on any dataclass is causing `get_class_decorator_hook` to get called infinitely.

I'm thinking either this affects all dataclass-oriented mypy plugins, or there is something funny going on with how we've defined our dataclasses.

Marshmallow has a dataclasses plugin that I think we could use as a reference to see if they have issues handling `InitVar` with the same versions of mypy/etc. If so, I think it will be easier for us to go to the mypy devs (and/or dataclasses devs?) and explain that this seems broken; if not, hopefully that will give us some more insight into what could be going wrong.

However, before digging into this more deeply, I'd like to wait for @samuelcolvin's work on v2 dataclasses to come through in the case that it ends up having any effect on the mypy side of things. (I suspect this is unlikely as it seems current work is mostly focused on the `pydantic_core` side of things, but, stil..)

dmontagu:
Update: I did a bit more digging, and noticed that there must be something more complicated going on; a much simpler file that can reproduce the infinite calling to the dataclass transform decorator:

```python
from dataclasses import InitVar  # noqa E402

class Model(BaseModel):
    x: float
    y: str

    model_config = ConfigDict(from_attributes=True)

    not_config = ConfigDict(frozen=True)  # <--------- this line


@dataclass
class MyDataClass:
    foo: InitVar[str]
    bar: str
```

Very interestingly, if I delete the line with the `# <--------- this line` comment above, the issue goes away(!). I'll do a bit more investigation now given this is a pretty promising lead.

dmontagu:
Okay, I think this is good to merge. The main change I made is dropping the analysis of non-annotated fields.

I'm not sure why that was causing infinite loops but it's no longer valid in v2 anyway, so I dropped it as a config option and made it always error and stop analyzing further. That seems to have resolved the issue.

@samuelcolvin If you are okay with the docs changes (presumably we'll do a bigger pass eventually, but I wanted to update this now so I don't forget), and with the high-level ideas behind the changes, then I think this is good to merge.

dmontagu:
please review

samuelcolvin (commented):
LGTM, although i haven't read every line.

PR #5177, opened 2023-03-13
Add strict schemas for types defined in Python

### LINKED ISSUE #4664: V2: Strict mode for types defined in python
See [the "strict mode" section of the blog post](https://pydantic-docs.helpmanual.io/blog/pydantic-v2/#strict-mode).

#4516 implements types for all/most of the types supported in V1.

But it doesn't touch the issue of "strict mode" at all. We need to:
* modify all types defined in pydantic to make them work correctly in strict mode
* add thorough tests for these

pydantic-core can switch from strict to lax mode at validation time, it doesn't require separate validators, however unfortunately I very much doubt that will be possible in pydantic - many of the schemas will need to be different for strict and lax mode.

I guess we define models as strict or lax, then add class methods:
* `model_lax()` which creates a lax version of the model
* `model_strict()` which creates a strict version of the model

The other option is to build (or optionally) build (perhaps lazily) two validators for a model.

adriangb:
Closes #4664

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_std_types_schema.py
+++ b/pydantic/_internal/_std_types_schema.py
@@ -16,8 +16,8 @@ from pathlib import PurePath
 from typing import Any, Callable
 from uuid import UUID
 
-from pydantic_core import MultiHostUrl, PydanticCustomError, Url, core_schema
-from typing_extensions import get_args
+from pydantic_core import CoreSchema, MultiHostUrl, PydanticCustomError, Url, core_schema
+from typing_extensions import Literal, get_args
 
 from ..json_schema import JsonSchemaMetadata
 from . import _serializers, _validators
@@ -82,33 +82,41 @@ def enum_schema(_schema_generator: GenerateSchema, enum_type: type[Enum]) -> cor
     )
     metadata = build_metadata_dict(js_metadata=js_metadata)
 
+    lax: CoreSchema
+    json_type: Literal['int', 'float', 'str']
     if issubclass(enum_type, int):
         # this handles `IntEnum`, and also `Foobar(int, Enum)`
         js_metadata['extra_updates'] = {'type': 'integer'}
-        return core_schema.chain_schema(
+        lax = core_schema.chain_schema(
             core_schema.int_schema(),
             literal_schema,
             core_schema.general_plain_validator_function(to_enum),
             metadata=metadata,
         )
+        json_type = 'int'
     elif issubclass(enum_type, str):
         # this handles `StrEnum` (3.11 only), and also `Foobar(str, Enum)`
         # TODO: add test for StrEnum in 3.11, and also for enums that inherit from str/int
         js_metadata['extra_updates'] = {'type': 'string'}
-        return core_schema.chain_schema(
+        lax = core_schema.chain_schema(
             core_schema.str_schema(),
             literal_schema,
             core_schema.general_plain_validator_function(to_enum),
             metadata=metadata,
         )
+        json_type = 'str'
     else:
-        return core_schema.general_after_validator_function(to_enum, literal_schema, metadata=metadata)
+        lax = core_schema.general_after_validator_function(to_enum, literal_schema, metadata=metadata)
+        json_type = 'str'
+    return core_schema.lax_or_strict_schema(
+        lax_schema=lax,
+        strict_schema=core_schema.is_instance_schema(enum_type, json_types={json_type}),
+        metadata=metadata,
+    )
 
 
 @schema_function(Decimal)
-def decimal_schema(
-    _schema_generator: GenerateSchema, _decimal_type: type[Decimal]
-) -> core_schema.AfterValidatorFunctionSchema:
+def decimal_schema(_schema_generator: GenerateSchema, _decimal_type: type[Decimal]) -> core_schema.LaxOrStrictSchema:
     decimal_validator = _validators.DecimalValidator()
     metadata = build_metadata_dict(
         update_cs_function=decimal_validator.__pydantic_update_schema__,
@@ -117,7 +125,7 @@ def decimal_schema(
             core_schema_override=lambda: decimal_validator.json_schema_override_schema()
         ),
     )
-    return core_schema.general_after_validator_function(
+    lax = core_schema.general_after_validator_function(
         decimal_validator,
         core_schema.union_schema(
             core_schema.is_instance_schema(Decimal, json_types={'int', 'float'}),
@@ -126,20 +134,28 @@ def decimal_schema(
             core_schema.str_schema(strip_whitespace=True),
             strict=True,
         ),
-        metadata=metadata,
     )
+    strict = core_schema.custom_error_schema(
+        core_schema.general_after_validator_function(
+            decimal_validator,
+            core_schema.is_instance_schema(Decimal, json_types={'int', 'float'}),
+        ),
+        custom_error_type='decimal_type',
+        custom_error_message='Input should be a valid Decimal instance or decimal string in JSON',
+    )
+    return core_schema.lax_or_strict_schema(lax_schema=lax, strict_schema=strict, metadata=metadata)
 
 
 @schema_function(UUID)
-def uuid_schema(_schema_generator: GenerateSchema, uuid_type: type[UUID]) -> core_schema.UnionSchema:
+def uuid_schema(_schema_generator: GenerateSchema, uuid_type: type[UUID]) -> core_schema.LaxOrStrictSchema:
     metadata = build_metadata_dict(
         js_metadata=JsonSchemaMetadata(
             source_class=UUID, type='string', format='uuid', modify_js_function=lambda schema: schema.pop('anyOf', None)
         )
     )
     # TODO, is this actually faster than `function_after(union(is_instance, is_str, is_bytes))`?
-    return core_schema.union_schema(
-        core_schema.is_instance_schema(uuid_type),
+    lax = core_schema.union_schema(
+        core_schema.is_instance_schema(uuid_type, json_types={'str'}, metadata=metadata),
         core_schema.general_after_validator_function(
             _validators.uuid_validator,
             core_schema.union_schema(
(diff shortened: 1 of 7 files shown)


pydantic/_internal/_std_types_schema.py
```diff
@@ -219,71 +261,92 @@ def ordered_dict_schema(schema_generator: GenerateSchema, obj: Any) -> core_sche
...
 @schema_function(IPv6Network)
-def ip_v6_network_schema(_schema_generator: GenerateSchema, _obj: Any) -> core_schema.FunctionPlainSchema:
+def ip_v6_network_schema(_schema_generator: GenerateSchema, _obj: Any) -> core_schema.LaxOrStrictSchema:
     metadata = build_metadata_dict(
         js_metadata=JsonSchemaMetadata(source_class=IPv6Network, type='string', format='ipv6network')
     )
-    return core_schema.function_plain_schema(
-        _validators.ip_v6_network_validator, serialization=core_schema.to_string_ser_schema(), metadata=metadata
+    return core_schema.lax_or_strict_schema(
+        lax_schema=core_schema.function_plain_schema(_validators.ip_v6_network_validator, metadata=metadata),
```
dmontagu, line 347: Should the `metadata` (which is currently just used for JSON schema generation) be living on the parent `lax_or_strict_schema` rather than just the `lax` schema?

adriangb: Turns out, that doesn't work. I think we'd need to make some changes to our json schema generation helper functions, but that would greatly increase the scope of this PR, so I'm going to leave it as-is.

dmontagu: okay, I'll be happy to take care of that after merging this

tests/test_types.py
```diff
@@ -3329,8 +3429,30 @@ class Model(BaseModel):
...
+            ),
+        ),
+        pytest.param(
+            Tuple[int, int],
+            {1: 10, 2: 20, 3: 30}.items(),
+            deque([(1, 10), (2, 20), (3, 30)]),
+            marks=pytest.mark.skipif(
+                condition=python_implementation() == 'PyPy', reason='PyPy does not implement a dict_items type'
+            ),
+        ),
```
adriangb, line 3455: I had to skip these tests on PyPy. I bet that if we replaced https://github.com/pydantic/pydantic-core/blob/7a4666490d825dd766dc98f506be5ea526eb5cfb/src/input/input_python.rs#L11 with a GILOnceCell that gets `type({}.keys())` and such it would work on PyPy.

pydantic/_internal/_std_types_schema.py
```diff
@@ -82,25 +82,32 @@ def to_enum(__input_value: Any, _: core_schema.ValidationInfo) -> Enum:
...
+        lax = core_schema.chain_schema(
+            core_schema.str_schema(), literal_schema, core_schema.function_plain_schema(to_enum)
         )
     else:
-        return core_schema.function_after_schema(literal_schema, to_enum, metadata=metadata)
+        lax = core_schema.function_after_schema(literal_schema, to_enum)
+
+    return core_schema.lax_or_strict_schema(
+        lax_schema=lax,
+        strict_schema=core_schema.is_instance_schema(enum_type),
```
samuelcolvin, line 104: either here or in a separate issue we need to think about the validation error you'd get.

I guess ideally the error should be the same or consistent between lax and strict.

We could use something like

```suggestion
        strict_schema=core_schema.custom_error_schema(
            core_schema.is_instance_schema(enum_type),
            'literal_error',
            custom_error_context=', '.join(... enum members),
        ),
```

not a blocker for this, but if we don't change now, we should create an issue to remember about this.

pydantic/_internal/_std_types_schema.py
```diff
@@ -109,67 +116,82 @@ def decimal_schema(_schema_generator: GenerateSchema, _decimal_type: type[Decima
...
+                core_schema.int_schema(),
+                core_schema.float_schema(),
+                core_schema.str_schema(strip_whitespace=True),
+                strict=True,
+            ),
+            decimal_validator,
+            metadata=metadata,
         ),
-        decimal_validator,
+        strict_schema=core_schema.is_instance_schema(Decimal),
```
samuelcolvin, line 131: to support max/min/nan checks, I think this needs to be
```suggestion
        strict_schema=core_schema.function_after_schema(
            core_schema.is_instance_schema(Decimal, json_types={'int', 'float'}),
            decimal_validator,
            metadata=metadata,
        ),
```

pydantic/_internal/_std_types_schema.py
samuelcolvin: also just realised that this won't work for JSON. 

I guess we'll need to use a function after validator with

```py
core_schema.is_instance_schema(enum_type, json_types={'int', 'float', 'str'})
```

I guess ultimately we might need a `core_schema.python_or_json_schema` type so we can have a custom schema for json data, but I think that's not necessary now.

pydantic/_internal/_std_types_schema.py
```diff
@@ -109,67 +116,82 @@ def decimal_schema(_schema_generator: GenerateSchema, _decimal_type: type[Decima
...
+                metadata=metadata,
+            ),
+            custom_error_type='path_type',
+            custom_error_message='Input is not a valid path',
+            strict=True,
         ),
-        custom_error_type='path_type',
-        custom_error_message='Input is not a valid path',
-        strict=True,
+        strict_schema=core_schema.is_instance_schema(path_type),
```
samuelcolvin, line 179: again we can add `json_types={'str'}`, but we will need to add an after validator to validate the input is valid if it came from JSON.

pydantic/_internal/_std_types_schema.py
```diff
@@ -109,67 +116,82 @@ def decimal_schema(_schema_generator: GenerateSchema, _decimal_type: type[Decima
...
-def _deque_any_schema() -> core_schema.FunctionWrapSchema:
-    return core_schema.function_wrap_schema(
-        _validators.deque_any_validator, core_schema.list_schema(allow_any_iter=True), serialization=_deque_ser_schema()
+def _deque_any_schema() -> core_schema.LaxOrStrictSchema:
+    return core_schema.lax_or_strict_schema(
+        lax_schema=core_schema.function_wrap_schema(
+            _validators.deque_any_validator,
+            core_schema.list_schema(),
+        ),
+        strict_schema=core_schema.is_instance_schema(deque),
```
samuelcolvin, line 193: `json_types=['array']`?

pydantic/_internal/_std_types_schema.py
```diff
@@ -191,15 +213,28 @@ def deque_schema(schema_generator: GenerateSchema, obj: Any) -> core_schema.Core
...
-            _validators.deque_typed_validator,
+        return core_schema.lax_or_strict_schema(
+            lax_schema=core_schema.function_after_schema(
+                core_schema.list_schema(inner_schema),
+                _validators.deque_typed_validator,
+                serialization=_deque_ser_schema(inner_schema),
+            ),
+            strict_schema=core_schema.function_after_schema(
+                core_schema.chain_schema(
+                    core_schema.is_instance_schema(deque),
```
samuelcolvin, line 224:
```suggestion
                    core_schema.is_instance_schema(deque, json_types=['array']),
```

pydantic/_internal/_std_types_schema.py
```diff
@@ -191,15 +213,28 @@ def deque_schema(schema_generator: GenerateSchema, obj: Any) -> core_schema.Core
...
         # `Deque[Something]`
         inner_schema = schema_generator.generate_schema(arg)
-        return core_schema.function_after_schema(
-            core_schema.list_schema(inner_schema),
-            _validators.deque_typed_validator,
+        return core_schema.lax_or_strict_schema(
+            lax_schema=core_schema.function_after_schema(
+                core_schema.list_schema(inner_schema),
+                _validators.deque_typed_validator,
+                serialization=_deque_ser_schema(inner_schema),
```
samuelcolvin, line 220: no idea if this is required when we have it below too.

pydantic/_internal/_std_types_schema.py
```diff
@@ -191,15 +213,28 @@ def deque_schema(schema_generator: GenerateSchema, obj: Any) -> core_schema.Core
...
             serialization=_deque_ser_schema(inner_schema),
         )
 
 
-def _ordered_dict_any_schema() -> core_schema.FunctionWrapSchema:
-    return core_schema.function_wrap_schema(_validators.ordered_dict_any_validator, core_schema.dict_schema())
+def _ordered_dict_any_schema() -> core_schema.LaxOrStrictSchema:
+    return core_schema.lax_or_strict_schema(
+        lax_schema=core_schema.function_wrap_schema(_validators.ordered_dict_any_validator, core_schema.dict_schema()),
+        strict_schema=core_schema.is_instance_schema(OrderedDict),
```
samuelcolvin, line 236:
```suggestion
        strict_schema=core_schema.is_instance_schema(OrderedDict, json_types=['object']),
```

samuelcolvin (commented):
I've reviewed partially, but I think we need to add either:
1. `schema_for_json` field to `is-instance` validators
2. or, a `python-or-json` schema type as described below

I think my preference is for 1., but I'm fairly easy.

samuelcolvin:
please update.

pydantic/_internal/_std_types_schema.py
adriangb: Here's what I'm getting with a custom error:

```
{
    'type': 'decimal_parsing',
    'loc': (
        'a',
        'lax-or-strict[lax=function-after[DecimalValidator(allow_inf_nan=False, check_digits=False, strict=False)(), union[is-instance[Decimal],int,float,constrained-str]],strict=custom-error[function-after[DecimalValidator(allow_inf_nan=False, check_digits=False, strict=False)(), is-instance[Decimal]]]]',
    ),
    'msg': 'Input should be a valid decimal',
    'input': 'foobar',
}
```

In particular `lax-or-strict` should never show up in an error loc. Unlike a union, we are either lax or strict for the entire validation. So `lax-or-strict` should just surface up the lax or strict error depending on which mode is being used.

adriangb:
please review

PR #5181, opened 2023-03-13
Update some more old tests with notes and/or fixes

dmontagu:
Similar in spirit to #5151 — fixes some existing tests, and attempts to get some clarity on how we want to resolve some of the other failures. Depending on the responses, I can either update/remove tests where appropriate, or, if fixing them is too involved to make sense in this PR, I'll update the TODO: comments in them to reflect what should be done.

### DIFF
--- a/pydantic/__init__.py
+++ b/pydantic/__init__.py
@@ -1,5 +1,5 @@
 from pydantic_core import ValidationError
-from pydantic_core.core_schema import ValidationInfo
+from pydantic_core.core_schema import ModelFieldValidationInfo, ValidationInfo
 
 from . import dataclasses
 from .config import BaseConfig, ConfigDict, Extra
@@ -25,6 +25,7 @@ __all__ = [
     'validator',
     'serializer',
     'ValidationInfo',
+    'ModelFieldValidationInfo',
     # config
     'BaseConfig',
     'ConfigDict',
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -207,7 +207,7 @@ class WalkAndApply:
         return schema
 
     def handle_function_schema(
-        self, schema: core_schema.FunctionSchema | core_schema.FunctionPlainSchema | core_schema.FunctionWrapSchema
+        self, schema: core_schema.FunctionSchema | core_schema.PlainFunctionSchema | core_schema.WrapFunctionSchema
     ) -> CoreSchema:
         if schema['mode'] == 'plain':
             return schema
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -9,7 +9,7 @@ from functools import wraps
 from inspect import Parameter, signature
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Generic, List, TypeVar, Union, cast
 
-from pydantic_core.core_schema import JsonReturnTypes, ValidationInfo, ValidatorFunction, WhenUsed
+from pydantic_core.core_schema import GeneralValidatorFunction, JsonReturnTypes, ValidationInfo, WhenUsed
 from typing_extensions import Protocol
 
 from ..errors import PydanticUserError
@@ -39,12 +39,13 @@ class Validator:
     Store information about field and root validators.
     """
 
-    __slots__ = 'function', 'mode', 'sub_path', 'check_fields'
+    __slots__ = 'function', 'mode', 'sub_path', 'check_fields', 'is_field_validator'
 
     def __init__(
         self,
         *,
         mode: Literal['before', 'after', 'wrap', 'plain'],
+        is_field_validator: bool,
         sub_path: tuple[str | int, ...] | None = None,
         check_fields: bool | None = None,
     ):
@@ -53,6 +54,7 @@ class Validator:
         self.mode = mode
         self.sub_path = sub_path
         self.check_fields = check_fields
+        self.is_field_validator = is_field_validator
 
 
 class Serializer:
@@ -294,8 +296,8 @@ V1Validator = Union[
 
 
 def make_generic_validator(
-    validator: V1Validator | OnlyValueValidator | ValidatorFunction, mode: str
-) -> ValidatorFunction:
+    validator: V1Validator | OnlyValueValidator | GeneralValidatorFunction, mode: str
+) -> GeneralValidatorFunction:
     """
     In order to support different signatures, including deprecated validator signatures from v1,
     we introspect the function signature and wrap it in a parent function that has a signature
@@ -374,7 +376,7 @@ def make_generic_validator(
 
         return _wrapper3
     elif keyword_only_params == [] and len(positional_params) == 2:
-        validator = cast(ValidatorFunction, validator)
+        validator = cast(GeneralValidatorFunction, validator)
         return validator
     raise TypeError(
         f'Unsupported signature for {mode} validator {validator}: {sig} is not supported.'
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -564,7 +564,7 @@ class GenerateSchema:
         from ._validators import mapping_validator
 
         # TODO could do `core_schema.chain_schema(core_schema.is_instance_schema(dict_subclass), ...` in strict mode
-        return core_schema.function_wrap_schema(
+        return core_schema.general_wrap_validation_function(
             mapping_validator,
             core_schema.dict_schema(
                 keys_schema=self.generate_schema(arg0),
@@ -581,12 +581,12 @@ class GenerateSchema:
         from ._validators import construct_counter
 
         # TODO could do `core_schema.chain_schema(core_schema.is_instance_schema(Counter), ...` in strict mode
-        return core_schema.function_after_schema(
+        return core_schema.general_after_validation_function(
+            construct_counter,
             core_schema.dict_schema(
                 keys_schema=self.generate_schema(arg),
                 values_schema=core_schema.int_schema(),
             ),
-            construct_counter,
         )
 
     def _mapping_schema(self, mapping_type: Any) -> core_schema.CoreSchema:
@@ -600,7 +600,7 @@ class GenerateSchema:
         else:
             from ._validators import mapping_validator
 
-            return core_schema.function_wrap_schema(
+            return core_schema.general_wrap_validation_function(
                 mapping_validator,
                 core_schema.dict_schema(
                     keys_schema=self.generate_schema(arg0),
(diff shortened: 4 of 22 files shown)


tests/test_aliases.py
```diff
@@ -106,12 +119,19 @@ class Model(BaseModel):
...
+            'loc': ('last_updated_by',),
+            'msg': 'Extra inputs are not permitted',
+            'type': 'extra_forbidden',
+        }
     ]
 
 
 @pytest.mark.xfail(reason='working on V2')
 def test_alias_child_precedence():
+    # TODO: Need to decide what we are doing with fields in model_config
```
dmontagu, line 133: Question 2: Re-confirming this: Are we sure we want to drop `model_config['fields']`/`Config.fields`?

In v1, it seems you can do arbitrary updates to the `FieldInfo`s for the fields by setting this attribute; there are a good number of tests using this to set JSON schema data, and using it to affect aliases, especially for fields defined in parent classes.

I think if we want to drop this, we need to either describe to users how they should override the field info from fields defined in parent classes (which I'm not sure how to do now), or we need to tell people that we no longer support that (which I think _may_ be a pain point for anyone who is currently using that functionality).

tests/test_aliases.py
```diff
@@ -178,6 +198,10 @@ class Child(Parent):
 
 @pytest.mark.xfail(reason='working on V2')
 def test_low_priority_alias():
+    # TODO: alias_priority has been removed from `Field` — what should happen to this test?
+    #   Should we re-add alias_priority as a keyword argument to `Field`?
+    #   Is there something new that can be used to replicate this functionality?
+    #   Either way, if we don't re-add it to `Field`, remember to update the migration guide
```
dmontagu, line 204: Question 3: Should we re-add `alias_priority` to `Field`? I am inclined to say yes for backwards compatibility's sake since it doesn't seem to have much impact, but I don't have to if we want to remove it.

tests/test_forward_ref.py
```diff
@@ -644,6 +644,8 @@ class Model(BaseModel):
     )
 
     m = module.Model(foo_user={'x': 'user1'}, user={'y': 'user2'})
+    # TODO: Do we need to support models_as_dict=False?
+    #   If not, is there a way to replicate this custom-encoder functionality?
     assert m.model_dump_json(models_as_dict=False) == '{"foo_user": {"x": "user1"}, "user": "User(user2)"}'
```
dmontagu, line 649: Question 4: Without something like `models_as_dict=False`, I don't think see a good way to customize the encoding for the models; is there a way to handle this in v2?

tests/test_networks.py
```diff
@@ -647,13 +647,13 @@ class Model2(BaseModel):
...
 
-@pytest.mark.xfail(reason='URL to JSON not yet supported')
 def test_json():
     class Model(BaseModel):
         v: HttpUrl
 
     m = Model(v='http://foo@example.net')
-    assert m.model_dump_json() == '{"v": "http://foo@example.net"}'
+    # TODO: Is it okay that the trailing '/' gets added?
+    assert m.model_dump_json() == b'{"v":"http://foo@example.net/"}'
```
dmontagu, line 656: Question 5: Note this behavior has changed relative to v1, is this acceptable? (I.e., adding the trailing `/`)

tests/test_parse.py
```diff
@@ -114,18 +113,22 @@ class MyModel(BaseModel):
     assert isinstance(m.__root__.__root__[0], str)
 
 
-@pytest.mark.xfail(reason='working on V2')
 def test_json():
-    assert Model.parse_raw('{"a": 12, "b": 8}') == Model(a=12, b=8)
+    # TODO: Are we planning to keep model_validate_json?
+    assert Model.model_validate_json('{"a": 12, "b": 8}') == Model(a=12, b=8)
```
dmontagu, line 118: Question 6: Since we dropped all the other forms of parse_raw, just wanted to confirm that we want to keep model_validate_json around

tests/test_parse.py
```diff
@@ -114,18 +113,22 @@ class MyModel(BaseModel):
...
 
 
 @pytest.mark.xfail(reason='working on V2')
 def test_json_ct():
+    # TODO: Should this test be dropped now that parse_raw is gone?
```
dmontagu, line 123: Question 7: This seemed to be checking specifically for the behavior of the `content_type` arg; since `parse_raw` is gone and `model_validate_json` doesn't have this arg, can we drop this test completely?

tests/test_parse.py
```diff
@@ -114,18 +113,22 @@ class MyModel(BaseModel):
...
     assert Model.parse_raw('{"a": 12, "b": 8}', content_type='application/json') == Model(a=12, b=8)
 
 
 @pytest.mark.xfail(reason='working on V2')
 def test_pickle_ct():
+    # TODO: It seems model_validate_json has replaced parse_raw, except that it can _only_ be used with json
+    #   Are we just dropping the content_type/proto functionality? If so, should we drop these tests?
+    #   Do we need to document and/or create a suggested migration path?
     data = pickle.dumps(dict(a=12, b=8))
     assert Model.parse_raw(data, content_type='application/pickle', allow_pickle=True) == Model(a=12, b=8)
```
dmontagu, line 133: Question 8: Seems we've decided to not support other kinds of content_type any more, should this test just be dropped now? Same question applies to the "proto" argument.

tests/test_parse.py
```diff
@@ -148,6 +151,7 @@ def test_bad_proto():
 
 @pytest.mark.xfail(reason='working on V2')
 def test_file_json(tmpdir):
+    # TODO: Should the parse_file tests be dropped?
```
dmontagu, line 154: Question 9: It seems we no longer want to provide any built-in support for parsing files directly; should I just drop all these tests?

tests/test_parse.py
```diff
@@ -189,20 +193,3 @@ def test_file_pickle_no_ext(tmpdir):
...
-
-    class SubModelB(BaseModel):
-        key: str = Field('B', const=True)
-        foo: int
-
-    class Model(BaseModel):
-        a: Union[SubModelA, SubModelB]
-
-    m = Model.model_validate({'a': {'key': 'B', 'foo': 3}})
-    assert isinstance(m.a, SubModelB)
```
dmontagu, line 208: Question 10: Since we no longer support `const` on `Field`, and we have plenty of tests of `Literal`, I don't think this test is necessary any more, but let me know if it should somehow be retained.

If we were okay updating it to use Literal, we could do that; if we wanted to keep these type hints, we'd need to use field validators. However, I think both approaches are thoroughly tested elsewhere, which is why I want to drop this test.

dmontagu (commented):
Including this "review" to direct attention to the things I had questions about.

dmontagu:
please review

tests/test_networks.py
adriangb: Personally I think this is okay

tests/test_parse.py
adriangb: Isn't `model_validate_json` the "right" way to do this in V2?

tests/test_parse.py
adriangb: I am horrified that we unpickled random data previously, I'm glad we're dropping it. I would vote for removing this test.

tests/test_parse.py
adriangb: I would, but maybe the PR that drops them can update the migration guide?

tests/test_parse.py
adriangb: That is fine with me

adriangb (commented):
I responded to some questions but most I don't think I can give input on so re-assigning to @samuelcolvin

tests/test_parse.py
samuelcolvin: yes.

tests/test_parse.py
samuelcolvin: yes, drop test.

tests/test_parse.py
samuelcolvin: well we had the required `allow_pickle` argument as a guard, but yes, I agree. :see_no_evil: 

Remove the test.

tests/test_parse.py
samuelcolvin: yes, remove.

tests/test_parse.py
samuelcolvin: remove.

samuelcolvin (commented):
otherwise looks great.

please update.

tests/test_aliases.py
dmontagu: In this case, it provides a way to ensure that a field alias gets changed by an alias generator.

Since we are getting rid of the ability to change aliases of fields defined in parent classes using `Config.fields`, we should be thoughtful about whether we choose to support this, because it may result in people using `alias_generator` as a way to retain their ability to specify aliases in subclasses.

I see arguments for and against:
* For: this is a migration path for overriding parent class aliases, even if it ends up deprecated
* Against: alias_generator is treated as opaque by the mypy plugin (that said, I'm not sure if that's also the case for `Config.fields`, but I suspect the mypy plugin does inspect its values). Because of that, supporting this as a viable alternative might push people in a direction where they end up frustrated by this as a regression.

If there was any way to "redefine" a field on a child class without emitting a warning about shadowing I think that would address this as an issue, but I'd also be open to just telling people "you can't override aliases in submodels", and tell them if they want different aliases at different points in the class hierarchy they need to create parent models without aliases and subclass them with alias generators (not that that's a particularly elegant solution, but I think it would work, modulo inheritance wonkiness that's probably best to avoid anyway).

tests/test_construction.py
```diff
@@ -81,23 +81,18 @@ def test_simple_copy():
     assert m.model_fields == m2.model_fields
 
 
-@pytest.fixture(scope='session', name='ModelTwo')
-def model_two_fixture():
-    class ModelTwo(BaseModel):
-        __foo__ = PrivateAttr({'private'})
+class ModelTwo(BaseModel):
```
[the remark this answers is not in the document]
dmontagu: The reason I did this was that pickling requires models to be defined at module level, and the last time these tests were passing was when `ModelTwo` was defined at the module level; I assumed it was moved into a fixture to prevent failures at import time and now that it didn't generate those it could just be moved back.

But in accordance with your suggestion, I have reverted this and used `create_module` to handle the pickle tests.

tests/test_forward_ref.py
dmontagu: Is the idea that you'd wrap the model with a root serializer through an annotation?

If you have to add the root serializer to the model class itself then you'd lose the ability to change the behavior for different parent classes / the root class itself, which I suspect would be a pain point for migration. But as long as you can change the behavior through an annotation then I think it should be okay.

dmontagu:
please review

tests/test_forward_ref.py
samuelcolvin: either an annotation, or a field serialiser - e.g. the `@serialiser` decorator.

samuelcolvin:
great, thank you.

Only remaining thing to do might be creating some issues on extra stuff to do.

PR #5186, opened 2023-03-13
Use Validator in parse_obj_as

dmontagu:
The main question I still have: should we add a DeprecationWarning to `parse_obj_as` (telling people to use `Validator` instead)?

If I recall correctly, achieving similar functionality was the main reason this function was introduced.

### DIFF
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -1,6 +1,8 @@
+import warnings
 from functools import lru_cache
 from typing import Any, Callable, Optional, Type, TypeVar, Union
 
+from . import Validator
 from ._internal import _repr
 
 __all__ = 'parse_obj_as', 'schema_of', 'schema_json_of'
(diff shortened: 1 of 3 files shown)


pydantic/tools.py
```diff
@@ -26,9 +27,8 @@ def _get_parsing_type(type_: Any, *, type_name: Optional[NameFactory] = None) ->
 T = TypeVar('T')
 
 
-def parse_obj_as(type_: Type[T], obj: Any, *, type_name: Optional[NameFactory] = None) -> T:
-    model_type = _get_parsing_type(type_, type_name=type_name)  # type: ignore[arg-type]
-    return model_type(__root__=obj).__root__
+def parse_obj_as(type_: Type[T], obj: Any) -> T:
```
dmontagu, line 30: * Should we retain the `type_name` argument for backwards compatibility and just ignore its value? (I think I'm for this, unless we plan to keep the function around long term, in which case maybe we at least add a deprecation warning for using that argument)
* Should we make some effort to massage the error message into looking _exactly_ the same? (I'm against this)

adriangb (approved):

PR #5200, opened 2023-03-16
Test using a pydantic-core-stubs package

adriangb:


adriangb:
This is a bit suspicious: https://github.com/pydantic/pydantic/actions/runs/4434381419/jobs/7780305802#step:10:24
> Found 1 source file

adriangb:
TODO:
- Can we now run without `--disable-recursive-aliases` since `mypy` should not go down into `pydantic-core` while running `test_mpypy.py`?
- Should we switch the type checker we use for _our_ code (in the `lint` step, in `pre-commit`) from mypy to pyright?

adriangb:
Closing in favor of #5206

PR #5204, opened 2023-03-16
Fix issue where recursion tracking contextvar doesn't get reset if an exception happens

dmontagu:


dmontagu:
Because this is just adding a test for already-merged code, and "fixing" a test that was previously mis-implemented (and adding a note about strange behavior that still needs resolution), I'm just going to auto-merge.

PR #5206, opened 2023-03-16
skip analyzing pydantic-core via mypy configs

adriangb:


.github/workflows/ci.yml
```diff
@@ -192,9 +192,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # test recent mypy versions on 3.10 and mypy latest on all supported python versions
-        # (note pydantic V2 only supports mypy>1 and we don't yet support mypy 1.1.1)
-        mypy-version: ['1.0.1']
+        # test recent mypy versions on 3.10 and mypy 1.0.1 on all supported python versions
+        mypy-version: ['0.971', '0.982', '0.991', '1.0.1']
```
samuelcolvin, line 196: we still want to remove all `mypy` before `1.0.1` - we're not going to support older mypy in V2.

samuelcolvin (commented):
otherwise LGTM.

Great find!

samuelcolvin:
please update.

.github/workflows/ci.yml
samuelcolvin: merge once this is done.

adriangb:
It is meant to be a solution for users that just happens to speed up our CI. I realize that it is unfortunately not automatic but if the cost is having to respond and close issues that get opened by linking to the docs that seems like a small price to pay for us. I am concerned about users that just suffer through the pain, but it is so bad that I don’t know that any sane person will do that. There is also a chance that mypy fixes this on their end before V2 adoption ramps up fully so there’s something to be said for making this our best solution for now and revisiting later.

.github/workflows/ci.yml
adriangb: Really? I thought we removed it primarily because of the slowdown. If we solve that with this config, wouldn’t we want to support older versions of mypy to avoid forcing users to migrate both packages at the same time? Upgrading mypy may produce a ton of new type errors in users code, I shiver at the thought of the huge PRs users would have to do to update to V2

adriangb:
I did a bit more testing. I don't seem to be able to poke a hole in this: results are exactly the same with or without these settings.

.github/workflows/ci.yml
samuelcolvin: Humm, it depends how much extra code we require to keep old versions working.

I think the big problem will be mypy 1.1.1 and 1.2.

But you have a point about big updates to support newer mypy.

samuelcolvin:
This is great, thanks so much, merging.

We can discuss mypy versions supported later.

PR #5199, opened 2023-03-15
Fix classvars

dmontagu:
Will change the merge target to `main` after fix-private-attrs gets merged

Selected Reviewer: @samuelcolvin

PR #5215, opened 2023-03-17
Dunder copy

### LINKED ISSUE #4986: Depreciate `.copy()`, instead implement `__copy__`
As suggested by @PrettyWood we should remove `.copy()` and all it's "powerful" arguments as they're somewhat broken, and instead implement `__copy__`.

I guess we should leave `copy()` in place for now and add a deprecation warning.

@tiangolo @hramezani @adriangb @dmontagu anyone else, what do you think about this - is it going to cause a revolution?

I guess if we really want to keep `.copy()`, we could just use `.model_dump()` (née `.dict()`) and then run validation, to make sure the arguments all behave correctly.

Either way we should definitely implement `__copy__` and `__deepcopy__`. @ybressler you were asking about PRs to work on, maybe you could work on this bit while we discuss removing `.copy()`?

dmontagu:
This is initial work on #4986. Looking for feedback on these definitions before I press onward with the deprecation warnings, etc.

(Note this is currently against the branch from PR #5199.)

Closes #4986

Selected Reviewer: @adriangb


pydantic/main.py
```diff
@@ -658,6 +658,40 @@ def __eq__(self, other: Any) -> bool:
...
+        return m
+
+    def __deepcopy__(self, memo: dict[int, Any] | None = None) -> typing_extensions.Self:
+        """
+        Returns a deep copy of the model
+        """
+        cls = type(self)
+        m = cls.__new__(cls)
+        _object_setattr(m, '__dict__', deepcopy(self.__dict__, memo=memo))
+        _object_setattr(m, '__fields_set__', copy(self.__fields_set__))
```
dmontagu, line 688: this doesn't need to be a deepcopy because `__fields_set__` is a `set[str]`. I could use `deepcopy` for symmetry, but I figure that might be less performant in principle because python will probably attempt to deepcopy all the items from inside `__fields_set__` when we know that shouldn't be necessary.

I can add a comment explaining this if preferred.

dmontagu:
@samuelcolvin looking for your agreement on the implementation of `__copy__` and `__deepcopy__` here. If you are fine with them, just have hooky assign it back to me and I'll keep going.

pydantic/main.py
dmontagu: Added the comment; I'll mark this as resolved

dmontagu:
I'll note https://stackoverflow.com/a/54467864 — it seems in principle deepcopy might be much faster through rust than through python if we could get the logic to be equivalent.

However, I'm a bit concerned that we'll find ourselves in edge-case-hell if we try to perform arbitrary deepcopying outside python. Maybe it's simpler than I'm thinking, but the nice thing about this approach is that it should work for arbitrary inputs with minimal code.

pydantic/_internal/_dataclasses.py
```diff
@@ -65,7 +65,7 @@ def prepare_dataclass(
     self_schema, model_ref = get_dc_self_schema(cls)
     types_namespace = {**(types_namespace or {}), name: PydanticForwardRef(self_schema, cls)}
     try:
-        fields = collect_fields(cls, bases, types_namespace, is_dataclass=True, dc_kw_only=kw_only)
+        fields, _class_vars = collect_fields(cls, bases, types_namespace, is_dataclass=True, dc_kw_only=kw_only)
```
samuelcolvin, line 68:
```suggestion
        fields, _ = collect_fields(cls, bases, types_namespace, is_dataclass=True, dc_kw_only=kw_only)
```

clearer to me.

pydantic/_internal/_fields.py
```diff
@@ -104,7 +104,7 @@ def collect_fields(  # noqa: C901
     typevars_map: dict[str, Any] | None = None,
     is_dataclass: bool = False,
     dc_kw_only: bool | None = None,
-) -> dict[str, FieldInfo]:
+) -> tuple[dict[str, FieldInfo], set[str]]:
```
samuelcolvin, line 107: can you add something to the docstring about what's returned.

pydantic/_internal/_generate_schema.py
```diff
@@ -716,7 +716,7 @@ def _dataclass_schema(self, dataclass: type[StandardDataclass]) -> core_schema.C
         Generate schema for a dataclass.
         """
         # FIXME we need a way to make sure kw_only info is propagated through to fields
-        fields = collect_fields(
+        fields, _class_vars = collect_fields(
```
samuelcolvin, line 719:
```suggestion
        fields, _ = collect_fields(
```

pydantic/_internal/_model_construction.py
```diff
@@ -101,7 +104,7 @@ def deferred_model_get_pydantic_validation_schema(
     """
     self_schema, model_ref = get_model_self_schema(cls)
     types_namespace = {**(types_namespace or {}), cls.__name__: PydanticForwardRef(self_schema, cls)}
-    fields = collect_fields(cls, cls.__bases__, types_namespace)
+    fields, class_vars = collect_fields(cls, cls.__bases__, types_namespace)
```
samuelcolvin, line 107: is `class_vars` used here?

samuelcolvin (commented):
otherwise LGTM, do you want to do the extra work on `copy()` elsewhere?

pydantic/_internal/_model_construction.py
dmontagu: nope, changed to underscore

pydantic/main.py
```diff
@@ -233,6 +236,8 @@ def model_post_init(self, **kwargs: Any) -> None:
 
     @typing.no_type_check
     def __setattr__(self, name, value):
+        if name in self.__class_vars__:
+            raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')
```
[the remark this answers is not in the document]
dmontagu: I'll make it do this, as I think it's an improvement, though I'll note that there was a test checking for this exact behavior in v1

dmontagu:
@samuelcolvin I think most of your comments were more related things that had already been merged but where this branch was out of date. I have created a new pull request to address those (#5238) and I will proceed with the `model_copy` stuff here.

pydantic/main.py
```diff
@@ -577,11 +589,15 @@ def _iter(
     ) -> TupleGenerator:
         # Merge field set excludes with explicit exclude parameter with explicit overriding field set options.
         # The extra "is not None" guards are not logically necessary but optimizes performance for the simple case.
-        # if exclude is not None or self.__exclude_fields__ is not None:
-        #     exclude = _utils.ValueItems.merge(self.__exclude_fields__, exclude)
-        #
-        # if include is not None or self.__include_fields__ is not None:
-        #     include = _utils.ValueItems.merge(self.__include_fields__, include, intersect=True)
+        if exclude is not None:
```
dmontagu, line 592: Uncommenting these lines (with some modifications) allowed me to get tests passing, and this method seems to _only_ be used in (deprecated) `copy`.

I tried changing `copy` method to just do:
```
        data = self.model_dump(include=include, exclude=exclude, round_trip=True)
        data = {**data, **(update or {})}
        return self.model_validate(data)
```
but this doesn't work because 1, it actually validates, which makes many things not work with include/exclude, and 2, it doesn't actually do shallow copies.

For now, I'd be inclined to keep the logic like this (since all unit tests pass) and plan to remove `_iter` at the same time we remove `copy`.

dmontagu:
please review

dmontagu:
@JacobHayes made a good point in https://github.com/pydantic/pydantic/pull/5226#issuecomment-1480123314 that one thing that the current `copy` method makes easy (and in particular, possible) is getting a copy of a _frozen_ model with a field changed. As this PR currently exists, there is no way to modify fields while producing the copy.

Initially I wanted to propose adding it into the existing MR/`model_copy` function, but thinking about it more, I think we should probably just make a new issue for this (if one doesn't already exist), rather than necessarily addressing it in this PR, unless @samuelcolvin you think it's a good idea to re-add `update` as a keyword argument to `model_copy`. (The `update` keyword is certainly less complicated to support than include/exclude.)

samuelcolvin:
If it's really as simple as adding the update and calling `.update(update)`, I think we may as well as it here.

But if there are complex interactions with include/exclude, I'd prefer a new method.

pydantic/main.py
```diff
@@ -646,6 +664,51 @@ def __eq__(self, other: Any) -> bool:
...
+        :param update: values to change/add in the new model. Note: the data is not validated before creating
+            the new model: you should trust this data
+        :param deep: set to `True` to make a deep copy of the model
+        :return: new model instance
+        """
+        copied = self.__deepcopy__() if deep else self.__copy__()
+        if update:
+            copied.__dict__.update(update)
+            copied.__fields_set__.update(update.keys())
+        return copied
```
dmontagu, line 680: > If it's really as simple as adding the update and calling .update(update), I think we may as well as it here.
>
> But if there are complex interactions with include/exclude, I'd prefer a new method.

@samuelcolvin is this simple enough for you? (I think it meets the bar.) It passes all existing tests of copy that don't use include/exclude. (Including checks that you can use this to obtain frozen model instances with modified values.)

I think we should eventually try to get mypy to typecheck the `model_copy` method, but that can come later.

samuelcolvin (approved):
LGTM

pydantic/main.py
lig: @JoshYuJump `model_` is a protected namespace Pydantic V2 uses for its needs. This allows you to have a model that have an attribute named `copy` and avoid clashes with Pydantic public methods.

PR #5208, opened 2023-03-16
Fix mypy plugin crash (1.10.X-fixes)

dmontagu:
Closes #5191

Selected Reviewer: @hramezani

Issue #5191, opened 2023-03-14
Mypy plugin crashes with `explicit_package_bases` set as true


marcelopgc:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

After updating mypy to v1.1.1, I added the following `pyproject.toml` config, as descripted in the [documentation](https://docs.pydantic.dev/mypy_plugin/#plugin-settings), to prevent mypy errors on classes with Optional attributes:

``` txt
[tool.mypy]
strict = true
files = "src/app"
mypy_path = "src/"
namespace_packages = true
explicit_package_bases = true
follow_imports = "silent"
ignore_missing_imports = true
exclude = "(resources/.*)"

plugins = [
    "pydantic.mypy"
]

warn_redundant_casts = true
warn_unused_ignores = true
disallow_any_generics = true
check_untyped_defs = true
no_implicit_reexport = true

# for strict mypy: (this is the tricky one :-))
disallow_untyped_defs = true

[tool.pydantic-mypy]
init_forbid_extra = true
init_typed = true
warn_required_dynamic_aliases = true
warn_untyped_fields = true
```

Using that config, I found an internal error while running mypy
```
src/app/models/stages.py:152: error: INTERNAL ERROR -- Please try using mypy master on GitHub:
https://mypy.readthedocs.io/en/stable/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 1.1.1
Traceback (most recent call last):
  File "mypy/semanal.py", line 6261, in accept
  File "mypy/nodes.py", line 1123, in accept
  File "mypy/semanal.py", line 1561, in visit_class_def
  File "mypy/semanal.py", line 1641, in analyze_class
  File "mypy/semanal.py", line 1668, in analyze_class_body_common
  File "mypy/semanal.py", line 1751, in apply_class_plugin_hooks
  File "pydantic/mypy.py", line 150, in pydantic.mypy.PydanticPlugin._pydantic_model_class_maker_callback
[...]

PR #5216, opened 2023-03-18
Fix mypy plugin crash (main)

dmontagu:
Reported in issue #5191; this is essentially a manual cherry pick of the MR for the `1.10.X-fixes` branch (#5208).

I'll copy the same comment I made there:

> The only way I know how to reproduce this issue is to have circular imports across modules (see example codebase linked in https://github.com/pydantic/pydantic/issues/5191 — thank you @marcelopgc for sharing).
> 
> Given that, I'd like to skip adding a test for this for now (so we can get a fixed release out), as it will be significantly more complex than our existing mypy tests, but I have added a comment linking to issue https://github.com/pydantic/pydantic/issues/5191 which shows how to reproduce it. That should be enough to get better coverage once we attempt to get 100% coverage there.
> 
> @samuelcolvin let me know if you aren't comfortable merging this without a covering test.

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -542,13 +542,26 @@ class PydanticModelTransformer:
 
         This is the same approach used by the attrs and dataclasses plugins.
         """
-        info = self._ctx.cls.info
+        ctx = self._ctx
+        info = ctx.cls.info
         for field in fields:
             sym_node = info.names.get(field.name)
             if sym_node is not None:
                 var = sym_node.node
-                assert isinstance(var, Var)
-                var.is_property = frozen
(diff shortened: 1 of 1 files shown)


dmontagu:
I'll also note that while this was reported as being due to `explicit_package_bases=True`, I'm not seeing anything indicating that that is connected to the crash, and I think this is definitely an improvement either way as, even if new errors are hit in this logic path, this should just produce a "normal" mypy error (that can be `# type: ignore`'ed) with debug information for us rather than producing an internal error in the plugin.

dmontagu:
please review

PR #5221, opened 2023-03-18
Support instance method serializers

adriangb:


adriangb:
please review

pydantic/__init__.py
```diff
@@ -1,5 +1,12 @@
 from pydantic_core import ValidationError
-from pydantic_core.core_schema import FieldValidationInfo, ValidationInfo
+from pydantic_core.core_schema import (
+    CallableValidator as ValidateWrapHandler,
```
adriangb, line 3: Leave name as is in pydantic-core and maybe rename in pydantic-core later

pydantic/__init__.py
```diff
@@ -1,5 +1,12 @@
...
+    FieldSerializationInfo,
+    FieldValidationInfo,
+    SerializationInfo,
+    SerializeWrapHandler,
+    ValidationInfo,
+)
```
adriangb, line 9: Export from `pydantic.validators` and `pydantic.serializers`?

adriangb: Tabling for future

pydantic/_internal/_generate_schema.py
```diff
@@ -768,23 +769,66 @@ def apply_validators(schema: core_schema.CoreSchema, validators: list[Validator]
...
+    'field-wrap',
+    'general-wrap',
+    'field-plain',
+    'general-plain',
+]
+
+
+def _is_unbound_instance_method(sig: Signature) -> bool:
+    if len(sig.parameters) < 1:
+        raise TypeError
```
adriangb, line 782: Will take care of these TODOs in a minute

samuelcolvin, line 782:
```suggestion
```
Looks like we don't need this from the usage below.

pydantic/_internal/_generate_schema.py
```diff
@@ -768,23 +769,66 @@ def apply_validators(schema: core_schema.CoreSchema, validators: list[Validator]
...
     if serializers:
-        # user the last serializser to make it easy to override a serializer set on a parent model
+        # user the last serializer to make it easy to override a serializer set on a parent model
         serializer = serializers[-1]
         assert serializer.sub_path is None, 'serializer.sub_path is not yet supported'
         function = typing.cast(typing.Callable[..., Any], serializer.function)
-        if serializer.wrap:
+        type_ = _infer_serializer_type_from_signature(function)
+        if serializer.wrap and type_ not in ('general-wrap', 'field-wrap'):
+            raise TypeError
```
samuelcolvin, line 814: I guess we need an error description here.

pydantic/_internal/_generate_schema.py
```diff
@@ -768,23 +769,66 @@ def apply_validators(schema: core_schema.CoreSchema, validators: list[Validator]
...
+    raise TypeError(f'Unrecognized serializer signature for {func.__name__} {sig}')
+
+
 def apply_serializers(schema: core_schema.CoreSchema, serializers: list[Serializer]) -> core_schema.CoreSchema:
     """
     Apply serializers to a schema.
     """
     if serializers:
-        # user the last serializser to make it easy to override a serializer set on a parent model
+        # user the last serializer to make it easy to override a serializer set on a parent model
```
samuelcolvin, line 808:
```suggestion
        # use the last serializer to make it easy to override a serializer set on a parent model
```

tests/test_serialize.py
```diff
@@ -85,3 +85,44 @@ class MyModel(BaseModel):
...
+        return f'{nxt(v):,}'
+
+    class MyModel(BaseModel):
+        f1: int
+        f2: int
+        f3: int
+        f4: int
+
+        @serializer('f1')
+        def ser_f1(self, v: Any, info: FieldSerializationInfo) -> Any:
```
samuelcolvin, line 104: Maybe I was confused by I thought we were changing the signature to `ser_f1(self, info: FieldSerializationInfo)` - e.g. not pass the value?

samuelcolvin (commented):
Looks good apart from a few small points, and question about whether we make field serialisers not take the value, and instead just use `self`?

please update.

pydantic/_internal/_decorators.py
```diff
@@ -203,19 +203,20 @@ class SerializationFunctions(DecoratorFunctions[Serializer]):
...
+def prepare_serializer_decorator(function: _SerializerType, allow_reuse: bool) -> _SerializerType:
     """
-    Convert the function to a classmethod if it isn't already.
     Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
     """
-    f_cls = function if isinstance(function, classmethod) else classmethod(function)
     if not allow_reuse and not in_ipython():
-        ref = f'{f_cls.__func__.__module__}::{f_cls.__func__.__qualname__}'
+        ref = f'{function.__module__}::{function.__qualname__}'
```
dmontagu, line 215: This caused an issue elsewhere with partial/etc.

I think here and other places in the current v2 code we should probably not be using `__qualname__` without a fallback (if I recall correctly, the python core devs said on some issue we commented on that you shouldn't assume this exists for everything callable).

This was a v1 PR where we improved support for `partial` and similar things, probably don't want to add a regression: https://github.com/pydantic/pydantic/pull/5126/files

If you don't want to fix this as part of this PR I don't think we necessarily have to do it here, BUT then we should make sure we are tracking this somewhere.

tests/test_serialize.py
adriangb: I ended up opting not to. It makes it much harder to have the following work:

```python
@serilizer('x', 'y')
def ser_xy(self, v: Any, info: FieldSerializationInfo) -> Any:
    lang = self.lang
    return internationalize(v, lang)
````

The main tradeoff is that even if you don't want to use the value you still have to accept it as a parameter.

It also made the signature introspection a bit easier because we can use the number of arguments as a conditional.

File: pydantic/_internal/_decorators.py (hunk 7 above)
Comment (adriangb) [reply pydantic/_internal/_decorators.py:215]: I agree we may need to be more careful but the reason I am removing this in this PR is that previously this function was getting wrapped with `classmethod` right above, and `classmethod` exposes the original function via `__func__`. So this change makes us no better or worse at handling `partial` than we were before.

File: pydantic/_internal/_generate_schema.py (hunk 4 above)
Comment (adriangb) [reply pydantic/_internal/_generate_schema.py]: done

File: pydantic/_internal/_generate_schema.py (hunk 5 above)
Comment (adriangb) [reply pydantic/_internal/_generate_schema.py]: fixed

File: pydantic/_internal/_generate_schema.py (hunk 3 above)
Comment (adriangb) [reply pydantic/_internal/_generate_schema.py]: Well we are doing `next(iter(thing))` here so I like having the check that we have at least 1 item here to justify that we can call `next(iter(...))` without catching the failure for an empty iterable

File: pydantic/_internal/_generate_schema.py [hunk 8]
```diff
@@ -768,23 +769,68 @@ def apply_validators(schema: core_schema.CoreSchema, validators: list[Validator]
...
+        elif type_ == 'field-wrap':
+            schema['serialization'] = core_schema.field_function_wrap_ser_schema(
+                function, schema.copy(), json_return_type=serializer.json_return_type, when_used=serializer.when_used
+            )
+        elif type_ == 'general-plain':
             schema['serialization'] = core_schema.general_function_plain_ser_schema(
                 function, json_return_type=serializer.json_return_type, when_used=serializer.when_used
             )
+        else:
+            # field-plain
```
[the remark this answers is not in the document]
adriangb: 45aca51

pydantic/_internal/_generate_schema.py
```diff
@@ -768,23 +769,68 @@ def apply_validators(schema: core_schema.CoreSchema, validators: list[Validator]
...
+        return 'general-plain'
+    elif len(sig.parameters) == 3:
+        # (self, value, info) -> Any or (value, nxt, info)
+        if _is_unbound_instance_method(sig, func):
+            return 'field-plain'
+        return 'general-wrap'
+    elif len(sig.parameters) == 4 and _is_unbound_instance_method(sig, func):
+        # (self, value, nxt, info) -> Any
+        return 'field-wrap'
+    raise TypeError(f'Unrecognized serializer signature for {func.__name__}: {sig}')
```
dmontagu, line 799: I think we should include information in this error for users about how they can find what the "recognized" serializer signatures are. Something like: `See pydantic.SerializerCallableProtocol` (or whatever it is called)

dmontagu (approved):
Other than awaiting responses to the other comments I made, this looks good to me.

pydantic/_internal/_generate_schema.py
adriangb: I added text describing the signatures. We don't have a single type, it's a union, and it's exported from `pydantic-core`. The whole callable protocol thing is also a bit confusing. So I decided to add a text explanation rather than pointing at a type. In the future when we have this in our docs we should just link to that.

dmontagu: > I added text describing the signatures.

Has this been pushed yet?

I think the important thing for me as a user would be that if this TypeError gets raised, I immediately know what I need to do to resolve it (don't need to start opening pydantic source code or Google). I don't know what the best description is, understand if it's not referencing the Protocol stuff, but what I'm getting at is that a user should be able to look at just a stacktrace and know how to proceed from there.

adriangb: https://github.com/pydantic/pydantic/blob/6cf49f95f453e24241f4fda744b51af523b1a177/pydantic/_internal/_generate_schema.py#L780-L806

dmontagu: Sorry, I think github was confusing me by showing me older state when I clicked on some item above, and I didn't realize I wasn't seeing the latest. Anyway this definitely looks good.

PR #5223, opened 2023-03-19
Fix custom constr class schema

matejetz:
## Change Summary

Currently, schema creation fails if a model uses a ConstrainedStr Model as keys in a Dict:
```python
class CustomStr(ConstrainedStr):
    regex = '^text$'

class Model(BaseModel):
    a: Dict[CustomStr, Any]

json_schema = Model.schema()
# AttributeError: 'str' object has no attribute 'pattern'
```

ConstrainedStr `regex` field is typed as `Optional[Union[str, Pattern[str]]`, when accessing pattern generating the schema for patternProperty, regex.pattern is directly accessed.

This PR contains following changes:
- Move `ConstrainedStr._get_pattern` to `get_pattern` in `utils.py`
- Use it in schema.py when accessing the pattern
- Add a test case for the described scenario

## Related issue number
-

## Checklist

* [] Tests pass on CI and coverage remains at 100%
* [] Documentation reflects the changes where applicable
* [] `changes/<pull request or issue id>-<github username>.md` file added describing change
* [] My PR is ready to review, **please add a comment including the phrase "please review" to assign reviewers**

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -30,7 +30,7 @@ from weakref import WeakSet
 
 from . import errors
 from .datetime_parse import parse_date
-from .utils import import_string, update_not_none
+from .utils import get_pattern, import_string, update_not_none
 from .validators import (
     bytes_validator,
     constr_length_validator,
@@ -412,7 +412,7 @@ class ConstrainedStr(str):
             field_schema,
             minLength=cls.min_length,
             maxLength=cls.max_length,
-            pattern=cls.regex and cls._get_pattern(cls.regex),
+            pattern=cls.regex and get_pattern(cls.regex),
         )
 
     @classmethod
@@ -431,13 +431,9 @@ class ConstrainedStr(str):
 
         if cls.regex:
             if not re.match(cls.regex, value):
-                raise errors.StrRegexError(pattern=cls._get_pattern(cls.regex))
+                raise errors.StrRegexError(pattern=get_pattern(cls.regex))
 
         return value
-
-    @staticmethod
-    def _get_pattern(regex: Union[str, Pattern[str]]) -> str:
-        return regex if isinstance(regex, str) else regex.pattern
 
 
 def constr(
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -820,6 +820,23 @@ def test_str_basic_types(field_type, expected_schema):
     assert Model.schema() == base_schema
 
 
+def test_constrained_str_class_dict():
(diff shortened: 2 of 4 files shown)


matejetz:
please review :)

pydantic/types.py
```diff
@@ -431,14 +431,10 @@ def validate(cls, value: Union[str]) -> Union[str]:
 
         if cls.regex:
             if not re.match(cls.regex, value):
-                raise errors.StrRegexError(pattern=cls._get_pattern(cls.regex))
+                raise errors.StrRegexError(pattern=get_pattern(cls.regex))
 
         return value
 
-    @staticmethod
-    def _get_pattern(regex: Union[str, Pattern[str]]) -> str:
```
samuelcolvin, line 439: We can't remove this function in case people are using it.

People could also have overridden the method to customise behaviour. 

Better to keep this and use it as we were before.

tests/test_schema.py
```diff
@@ -1702,27 +1703,36 @@ class Foo(BaseModel):
     }
 
 
-def test_schema_dict_constr():
+class TestSchemaDictConstr:
```
samuelcolvin, line 1706: Please don't change these test, just add a new one.

samuelcolvin (commented):
We're trying to absolutely minimise the footprint in changes to v1, hence my suggestions.

Please update.

matejetz:
updated towards minimal footprint @samuelcolvin

changes/5223-matejetz.md
```diff
@@ -0,0 +1 @@
+Fix creating schema from model using ConstrainedStr with regex as dict key
```
samuelcolvin, line 1:
```suggestion
Fix creating schema from model using `ConstrainedStr` with `regex` as dict key
```

samuelcolvin (approved):
otherwise LGTM.

@dmontagu please review and merge if you're happy.

dmontagu:
LGTM. I don't like using the "private" `ConstrainedStr._get_pattern` in a separate module, but I think it's better than duplicating the (admittedly fairly simple) logic (since if it changed it would be easy to miss that it needed an update here), and I don't think it makes sense to make the method public.

If necessary, we can be more thoughtful about the method naming, etc., in v2

dmontagu:
Yep, it looks like the code in v2 has been refactored so this particular failure mode can't even happen. So I don't think there's anything to do in terms of implementing a similar fix in v2.

PR #5224, opened 2023-03-19
[WIP] Prepare to Drop types from Pydantic

yezz123:
** ⚠️ DON'T MERGE**

ref: https://github.com/pydantic/pydantic-extra-types/issues/18


samuelcolvin:
We can drop these types from pydantic, instead we need to add warnings and we'll drop them completely in V3.

PR #5228, opened 2023-03-20
Skip analyzing pydantic_core for mypy run via pre-commit

adriangb:


adriangb:
With this change pre-commit takes <5s for me even with some changes.

PR #5233, opened 2023-03-20
Get more JSON schema tests passing

dmontagu:
* Adds new warning-emitting logic to address some TODOs from the first pass, and makes it easier for users to override.
* Handles aliases properly when producing JSON schema for a discriminated union
* Fixes a bug with the json schema generation for tuple-variable schemas (this was not caught by mypy due to recursive types 😢)
* Chooses not to set the max length for a Url in the JSON schema by default
* Gets some tests working by updating them to v2 methods/syntax/etc.
* Removes some tests using `__root__`, which will be removed
* Removes some other dead code from the `test_json_schema.py` file

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -2,13 +2,14 @@ from __future__ import annotations
 
 import math
 import re
+import warnings
 from dataclasses import is_dataclass
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Callable, Counter, Dict, Iterable, List, NewType, Sequence, Tuple, Union, cast
 
 from pydantic_core import CoreSchema, CoreSchemaType, core_schema
 from pydantic_core.core_schema import TypedDictField
-from typing_extensions import TypedDict
+from typing_extensions import Literal, TypedDict
 
 from ._internal import _core_metadata, _core_utils, _typing_extra, _utils
 from .errors import PydanticInvalidForJsonSchema, PydanticUserError
@@ -95,6 +96,12 @@ _FIELDS_MAPPING = {
     'format': 'format',
 }
 
+JsonSchemaWarningKind = Literal['skipped-choice', 'non-serializable-default']
+
+
+class PydanticJsonSchemaWarning(UserWarning):
+    pass
+
 
 def _apply_js_metadata(js_metadata: JsonSchemaMetadata, schema: JsonSchemaValue) -> None:
     """
@@ -137,6 +144,7 @@ JsonRef = NewType('JsonRef', str)
 class GenerateJsonSchema:
     # See https://json-schema.org/understanding-json-schema/reference/schema.html#id4 for more info about dialects
     schema_dialect = 'https://json-schema.org/draft/2020-12/schema'
+    ignored_warning_kinds: set[JsonSchemaWarningKind] = {'skipped-choice'}
 
     def __init__(self, by_alias: bool = True, ref_template: str = DEFAULT_REF_TEMPLATE):
         self.by_alias = by_alias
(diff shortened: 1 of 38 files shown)


dmontagu:
please review

pydantic/json_schema.py
```diff
@@ -95,6 +96,12 @@ def without_core_schema_override(json_schema_metadata: JsonSchemaMetadata) -> Js
     'format': 'format',
 }
 
+JsonSchemaWarningKind = Literal['skipped-choice', 'non-serializable-default']
+
+
+class PydanticJsonSchemaWarning(UserWarning):
+    pass
```
samuelcolvin, line 103: docstring here explaining what's going on would be good.

pydantic/json_schema.py
```diff
@@ -372,7 +380,7 @@ def tuple_positional_schema(self, schema: core_schema.TuplePositionalSchema) ->
     def tuple_variable_schema(self, schema: core_schema.TupleVariableSchema) -> JsonSchemaValue:
         json_schema: JsonSchemaValue = {'type': 'array', 'items': {}}
         if 'items_schema' in schema:
-            json_schema['items'] = [self.generate_inner(s) for s in schema['items_schema']]
+            json_schema['items'] = [self.generate_inner(schema['items_schema'])]
```
samuelcolvin, line 383: should items really be a list? looks to me like this suggests the array should only have one element?

pydantic/json_schema.py
```diff
@@ -445,20 +453,20 @@ def default_schema(self, schema: core_schema.WithDefaultSchema) -> JsonSchemaVal
...
 
         try:
             encoded_default = self.encode_default(default)
         except TypeError:
-            # This happens if the default value is not JSON serializable; in this case, just return the inner schema
-            # Note: We could update the '$comment' field to indicate that the default value was not JSON serializable.
-            #   This would have the upside that there would be some positive indication that the default value was not
-            #   valid, but would have the downside that it would make some assumptions about how users are using the
-            #   '$comment' field. For now, I have decided not to do this.
+            self.emit_warning(
```
samuelcolvin, line 465: i guess this is where we were going to use a method from pydantic-core, do you still need that?

pydantic/json_schema.py
```diff
@@ -513,20 +521,59 @@ def tagged_union_schema(self, schema: core_schema.TaggedUnionSchema) -> JsonSche
...
+            return schema['discriminator']
+
+        if isinstance(schema['discriminator'], list):
+            # When an alias is used, the discriminator will be a list of single str lists, one for the attribute
+            # and one for the actual alias. This logic will work even if there is more than one possible attribute,
+            # and looks for whether a single alias choice is present as a documented property on all choices.
+            # If so, that property will be used as the OpenAPI discriminator.
+            for alias_path in schema['discriminator']:
+                if not isinstance(alias_path, list):
+                    break  # this means that the discriminator is not a list of alias paths
```
samuelcolvin, line 557: surely in this case we should set `openapi_discriminator = alias_path`? at least if `alias_path` is a string?

samuelcolvin (commented):
otherwise looking good.

please update.

pydantic/json_schema.py
dmontagu: I realize I didn't handle the case of a single-element list properly. But if it was a list of more than one element it would be a path, which can't be handled. I will add handling of the one string element case. Let me know if you think there's more to it than that

pydantic/json_schema.py
dmontagu: I'm not sure I understand the question. I added `self.emit_warning` to make it easier to override the warning behaviors, between that and `ignored_warning_kinds` I might have gone overboard though.

pydantic/json_schema.py
samuelcolvin: I was more thinking about `alias = ['foobar', [0, 'foo', 'bar']`, e.g. a choice of either `foobar` or a path, but maybe that's not allowed.

dmontagu: My understanding was that that would be a malformed value of `alias` — I would expect that to look like `[['foobar'], [0, 'foo', 'bar']]`.

To be clear, it wouldn't be compatible with the type hint for discriminator: `str | list[str | int] | list[list[str | int]] | Callable[[Any], str | int | None]`

dmontagu: type hint for validation_alias: `Union[str, List[Union[str, int]], List[List[Union[str, int]]]]` (same point stands)

pydantic/json_schema.py
dmontagu: I added a TODO: comment about this inside `self.encode_default`.

dmontagu:
please review

PR #5235, opened 2023-03-21
Fix validator tests

### LINKED ISSUE #5246: Fix `each_item` in `@validator`
After #5235

adriangb:
- Fix most of the remaining`@validator` tests
- Fix most of the remaining `@root_validator` tests
- Introduces `@field_validator` for V2 validators and emit a warning for V1 `@validator`

Closes #5246

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -3,202 +3,209 @@ Logic related to validators applied to models etc. via the `@validator` and `@ro
 """
 from __future__ import annotations as _annotations
 
-import inspect
 import warnings
-from functools import wraps
 from inspect import Parameter, signature
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Generic, TypeVar, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Generic,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
 
-from pydantic_core.core_schema import GeneralValidatorFunction, JsonReturnTypes, ValidationInfo, WhenUsed
-from typing_extensions import Protocol
-
-from ..errors import PydanticUserError
+from pydantic_core.core_schema import (
+    FieldValidationInfo,
+    FieldValidatorFunction,
+    FieldWrapValidatorFunction,
+    JsonReturnTypes,
+    ValidationInfo,
+    WhenUsed,
+)
+from typing_extensions import Protocol, TypeAlias
 
 if TYPE_CHECKING:
     from typing_extensions import Literal
 
-    from ..main import BaseModel
 
-__all__ = (
-    'FIELD_VALIDATOR_TAG',
-    'ROOT_VALIDATOR_TAG',
-    'Validator',
-    'ValidationFunctions',
-    'SerializationFunctions',
-    'prepare_validator_decorator',
-    'prepare_serializer_decorator',
-)
 FIELD_VALIDATOR_TAG = '_field_validator'
 ROOT_VALIDATOR_TAG = '_root_validator'
 
 FIELD_SERIALIZER_TAG = '_field_serializer'
 
 
-class Validator:
+class ValidatorDecoratorInfo:
     """
-    Store information about field and root validators.
+    Store information about field validators created via `@validator`
     """
 
-    __slots__ = 'function', 'mode', 'sub_path', 'check_fields', 'is_field_validator'
+    model_attribute: ClassVar[str] = '__pydantic_validator_functions__'
 
     def __init__(
         self,
         *,
+        fields: tuple[str, ...],
         mode: Literal['before', 'after', 'wrap', 'plain'],
-        is_field_validator: bool,
+        type: Literal['unbound', 'field'],
         sub_path: tuple[str | int, ...] | None = None,
         check_fields: bool | None = None,
-    ):
-        # function is set later after the class is created and functions are bound
-        self.function: Callable[..., Any] | None = None
+    ) -> None:
+        """
+        :param mode: the pydantic-core validator mode.
+        :param type: either 'unbound' or 'field' indicating if this validator should have
+            access to the model itself.
+        :param mode: Not yet supported.
+        :param check_fields: whether to check that the fields actually exist on the model.
+        :param wrap: a callback to apply V1 compatibility shims or allow extra signatures
+            that pydantic-core does not recognize.
+        """
+        self.fields = fields
         self.mode = mode
         self.sub_path = sub_path
         self.check_fields = check_fields
-        self.is_field_validator = is_field_validator
+        self.type = type
 
 
-class Serializer:
+class RootValidatorDecoratorInfo:
     """
-    Store information about field serializers.
+    Store information about root validators created via `@root_validator`
     """
 
-    __slots__ = 'function', 'sub_path', 'wrap', 'json_return_type', 'when_used', 'check_fields'
+    model_attribute: ClassVar[str] = '__pydantic_root_validator_functions__'
 
     def __init__(
         self,
         *,
-        wrap: bool = False,
+        mode: Literal['before', 'after'],
+    ) -> None:
+        """
+        :param mode: the pydantic-core validator mode
+        """
+        self.mode = mode
+
+
+class SerializerDecoratorInfo:
+    """
+    A container for data from `@serializer` so that we can access it
+    while building the pydantic-core schema.
+    """
+
+    model_attribute: ClassVar[str] = '__pydantic_serializer_functions__'
+
+    def __init__(
+        self,
+        *,
+        fields: tuple[str, ...],
+        mode: Literal['plain', 'wrap'],
         json_return_type: JsonReturnTypes | None = None,
         when_used: WhenUsed = 'always',
         sub_path: tuple[str | int, ...] | None = None,
         check_fields: bool | None = None,
     ):
-        # arguments match core_schema.general_plain_serializer_function_ser_schema or
-        # core_schema.general_wrap_serializer_function_ser_schema
-        # function is set later after the class is created and functions are bound
-        self.function: Callable[..., Any] | None = None
+        self.fields = fields
         self.sub_path = sub_path
-        self.wrap = wrap
+        self.mode = mode
         self.json_return_type = json_return_type
         self.when_used = when_used
         self.check_fields = check_fields
 
 
-DecFunc = TypeVar('DecFunc', Validator, Serializer)
+DecoratorInfo = Union[ValidatorDecoratorInfo, RootValidatorDecoratorInfo, SerializerDecoratorInfo]
+
+ReturnType = TypeVar('ReturnType')
+DecoratedType: TypeAlias = 'Union[classmethod[ReturnType], staticmethod[ReturnType], Callable[..., ReturnType]]'
 
 
-class DecoratorFunctions(Generic[DecFunc]):
-    __slots__ = (
-        '_decorators',
-        '_field_decorators',
-        '_direct_field_decorators',
-        '_all_fields_decorators',
-        '_root_decorators',
-        '_used_decorators',
-    )
-    model_attribute: ClassVar[str]
-    _field_tag: ClassVar[str]
-    _root_tag: ClassVar[str | None]
+class PydanticDecoratorMarker(Generic[ReturnType]):
+    """
+    Wrap a classmethod, staticmethod or unbound function
+    and act as a descriptor that allows us to detect decorated items
+    from the class' attributes.
 
-    def __init__(self, bases: tuple[type[Any], ...]) -> None:
-        self._decorators: dict[str, DecFunc] = {}
-        self._field_decorators: dict[str, list[str]] = {}
-        self._direct_field_decorators: set[str] = set()
-        self._all_fields_decorators: list[str] = []
-        self._root_decorators: list[str] = []
-        self._used_decorators: set[str] = set()
-        self._inherit(bases)
+    This class' __get__ returns the wrapped item's __get__ result,
+    which makes it transparent for classmethods and staticmethods.
+    """
 
-    def extract_decorator(self, name: str, value: Any) -> bool:
-        """
-        If the value is a field or root decorator, add it to the appropriate group of decorators.
+    def __init__(
+        self,
+        wrapped: DecoratedType[ReturnType],
+        decorator_info: DecoratorInfo,
+        shim: Callable[[Callable[..., Any]], Callable[..., Any]] | None,
+    ) -> None:
+        self.wrapped = wrapped
+        self.decorator_info = decorator_info
+        self.shim = shim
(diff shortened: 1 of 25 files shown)


pydantic/_internal/_decorators.py
```diff
@@ -116,24 +116,33 @@ def extract_decorator(self, name: str, value: Any) -> bool:
...
             self._decorators[name] = decorator
             for field_name in fields:
                 this_field_decorators = self._field_decorators.get(field_name)
                 if this_field_decorators:
-                    this_field_decorators.append(name)
+                    # if a function with the same name as the current function
+                    # is already registered replace it instead of calling
+                    # a function by that name twice (which in many cases
+                    # would result in calling the last added function multiple
+                    # times)
```
dmontagu, line 130: I think this comment should be updated now based on the logic below. I'd suggest:
```
Only append the name of a new validator method if the name wasn't already present.
We've implemented it like this so that if you re-decorate a new method that has
the same name, the behavior is similar to what you'd get from inheritance/class
namespaces, and lets you override the original validator's logic.
```

tests/test_validators.py
```diff
@@ -22,6 +22,9 @@
 )
 from pydantic.decorators import root_validator
 
+# TODO: do we really want this API? Can we change it?
+RootValidatorValuesAndFieldsSet = Tuple[Dict[str, Any], Set[str]]
```
dmontagu, line 26: I would suggest we keep it for now and get this merged, but revisit before the alpha release.

tests/test_validators.py
```diff
@@ -638,37 +635,92 @@ class Child(Parent):
...
+            v.append('child after')
+            return v
+
+    assert Parent(a=[]).a == ['parent before', 'parent', 'parent after']
+    assert Child(a=[]).a == ['parent before', 'child', 'parent after', 'child before', 'child after']
 
 
-def test_inheritance_new():
+def test_inheritance_replace_root_validator():
+    """We promise that if you add a validator
```
dmontagu, line 683:
```suggestion
    """
    We promise that if you add a validator
```

tests/test_validators.py
```diff
@@ -683,6 +735,7 @@ def check_foobar(cls, v: Any):
     assert Model(foobar={1: 1}).foobar == {1: 2}
 
 
+# TODO: fairly certain
```
dmontagu, line 738: me too

tests/test_validators.py
```diff
@@ -987,15 +1052,14 @@ def root_validator(self, values):
                 return values
 
 
-@pytest.mark.xfail(reason='working on V2')
 def test_root_validator_extra():
-    with pytest.raises(errors.PydanticUserError) as exc_info:
+    with pytest.raises(TypeError, match=r'Unsupported signature for after validator') as exc_info:
```
dmontagu, line 1056: Not sure if the train has already left the station, but I would really prefer it somehow make it clear that it means a `mode='after'`-validator; right now trying to put myself in the shoes of a new-to-v2 user, I read `Unsupported signature for after validator` and think it's a signature coming after a validator or something else I don't understand.

adriangb:
I ended up having to rework the internal machinery due, in part, to the existing implementation relying on the fact that it always wrapped the method with `@classmethod` internally and thus had a new object to monkey patch onto every time it was called.

The new implementation returns an explicit descriptor object instead of monkey patching `classmethod`. I feel that this also had the advantage of cleaning up the implementation a bit, making it more strongly typed and easier to follow, but that may be just my personal bias.

tests/test_validators.py
adriangb: Since we now have `@validator` and `@field_validator` using `@validator` always warns and just tells you to use V2 validators.

pydantic/_internal/_generate_schema.py
```diff
@@ -318,11 +356,24 @@ def generate_td_field_schema(
         """
         assert field_info.annotation is not None, 'field_info.annotation should not be None when generating a schema'
         schema = self.generate_schema(field_info.annotation)
+        this_field_validators = filter_field_decorator_info_by_field(decorators.validator.values(), name)
```
dmontagu: It looks like `this_field_validators` is part of the v1 compatibility shim; if that's right, can we move this line below the comment below?

pydantic/decorators.py
```diff
@@ -83,46 +88,109 @@ def __call__(
...
+    check_fields: bool | None = None,
     allow_reuse: bool = False,
-) -> Callable[[V2Validator | V1Validator], classmethod[Any]]:
-    ...
+    each_item: bool = False,
+) -> Callable[[_V1ValidatorType], _V1ValidatorType]:
+    """
+    Decorate methods on the class indicating that they should be used to validate fields
+    :param fields: which field(s) the method should be called on
+    :param sub_path: TODO
```
dmontagu, line 130: I noticed sub_path isn't there, but it's still in this docstring. Should this docstring be updated?

pydantic/main.py
```diff
@@ -170,8 +161,7 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         __pydantic_validator__: typing.ClassVar[SchemaValidator]
         __pydantic_core_schema__: typing.ClassVar[CoreSchema]
         __pydantic_serializer__: typing.ClassVar[SchemaSerializer]
-        __pydantic_validator_functions__: typing.ClassVar[_decorators.ValidationFunctions]
-        __pydantic_serializer_functions__: typing.ClassVar[_decorators.SerializationFunctions]
+        __pydantic_decorators__: typing.ClassVar[_decorators.DecoratorInfos]
```
dmontagu, line 164: I think it would be helpful to add a comment indicating that this thing will hold information about validators and serializers (in addition to the docstrings I mentioned elsewhere)

tests/test_validators.py
```diff
@@ -154,48 +146,38 @@ def test_validate_whole():
...
+        @classmethod
+        def check_a2(cls, v: List[int]) -> List[Any]:
             v.append(456)
             return v
 
     assert Model(a=[1, 2]).a == [1, 2, 123, 456]
 
 
-@pytest.mark.xfail(reason='working on V2')
-def test_validate_kwargs():
```
dmontagu, line 171: did you retain any each_item tests? If not, I'm thinking we should keep this with a `with pytest.warns(DeprecationWarning)` wrapper or similar

pydantic/decorator.py
```diff
@@ -6,7 +6,7 @@
 
 from ._internal import _typing_extra, _utils
 from .config import Extra, get_config
-from .decorators import validator
+from .decorators import field_validator
```
dmontagu, line 9: Do we need both decorator.py _and_ decorators.py? Lol. Maybe changing that isn't a this-PR thing haha.

pydantic/decorator.py
```diff
@@ -246,7 +249,8 @@ def check_positional_only(cls, v: Optional[List[str]]) -> None:
                 keys = ', '.join(map(repr, v))
                 raise TypeError(f'positional-only argument{plural} passed as keyword argument{plural}: {keys}')
 
-            @validator(V_DUPLICATE_KWARGS, check_fields=False, allow_reuse=True)
+            @field_validator(V_DUPLICATE_KWARGS, check_fields=False, allow_reuse=True)
+            @classmethod
             def check_duplicate_kwargs(cls, v: Optional[List[str]]) -> None:
```
dmontagu, line 254: Do you know what's going on here? No tests currently cover this function

pydantic/_internal/_decorators.py
```diff
@@ -222,34 +284,66 @@ def prepare_serializer_decorator(function: _SerializerType, allow_reuse: bool) -
...
+def unwrap_unbound_methods(function: Callable[..., Any] | classmethod[Any] | staticmethod[Any]) -> Callable[..., Any]:
+    """
+    Unwrap unbound classmethods and staticmethods
+    """
+    if isinstance(function, (classmethod, staticmethod)):
+        return function.__func__
+    return function
+
+
+def is_v1_validator(function: Callable[..., Any] | staticmethod[Any] | classmethod[Any]) -> bool:
```
dmontagu, line 296: It seems no tests currently cover this

pydantic/_internal/_decorators.py
```diff
@@ -299,33 +392,36 @@ def __call__(self, __value: Any, values: dict[str, Any], **kwargs: Any) -> Any:
...
+    @validator('x')
+    @staticmethod  # required
+    def val_x4(value: Any, values: Dict[str, Any]) -> Any: ...
+
+    val_x5 = validator('x')(f1)
+    val_x6 = validator('x')(f2)
+"""
+
+
+def make_generic_v1_field_validator(validator: V1Validator) -> FieldValidatorFunction:
```
dmontagu, line 423: Also, much of this function seems uncovered by tests. I'm guessing it wouldn't be hard to change that by re-adding tests with the old stuff and wrapping them in deprecation-warning-catchers. You might be able to do something like what I did with `tests.test_construction.copy_method` where it parametrizes over both, and ensures a deprecation warning is raised for the deprecated one (you'd only use it where it was runtime-viable to use either of the validator or field_validator decorators)

pydantic/decorators.py
```diff
@@ -131,67 +199,107 @@ def validator(
...
+            ' you MUST specify `skip_on_failure=True`.'
+            ' The `skip_on_failure=False` option is no longer available.'
+            ' If you were not trying to set `skip_on_failure=False` you'
+            ' can safely set `skip_on_failure=True`.'
+            ' If you do, this root validator will no longer be called'
+            ' if validation fails for any of the fields.'
+            ' Please see the migration guide for more details.'
+        )
+    wrap = partial(_decorators.make_v1_generic_root_validator, pre=pre)
+    if __func is not None:
```
dmontagu, line 286: this branch is not covered in any tests

pydantic/decorators.py
```diff
@@ -83,46 +88,109 @@ def __call__(
...
+        ' see the migration guide for more details',
+        DeprecationWarning,
+        stacklevel=2,
+    )
+
+    mode: Literal['before', 'after'] = 'before' if pre is True else 'after'
+
+    def dec(f: Any) -> _decorators.PydanticDecoratorMarker[Any]:
+        if _decorators.is_instance_method_from_sig(f):
+            raise TypeError('`@validator` cannot be applied to instance methods')
```
dmontagu, line 147: this TypeError is not covered by any tests

dmontagu (changes requested):
Looking good

pydantic/_internal/_decorators.py
```diff
@@ -3,202 +3,264 @@
...
+        :param mode: the pydantic-core validator mode
+        """
+        self.mode = mode
+
+
+class SerializerDecoratorInfo(Representation):
     """
-    Store information about field serializers.
+    A container for data from `@serializer` so that we can access it
+    while building the pydantic-core schema.
```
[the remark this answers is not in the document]
adriangb: They are all used the same, I made them all match the `SerializerDecoratorInfo` one.

pydantic/decorator.py
adriangb: Nope, I don't really know (yet) what is going on in `decorator.py` / `@validate_arguments`. All I know is that it uses `@validator` which is now deprecated, so I am replacing it with the new version.

pydantic/_internal/_decorators.py
adriangb: I'll add tests to cover these cases.

pydantic/decorators.py
adriangb: thanks, I should have looked at coverage, removed now :)

pydantic/_internal/_decorators.py
```diff
@@ -3,202 +3,264 @@
...
-                self._decorators.update(parent_vf._decorators)
-                for k, v in parent_vf._field_decorators.items():
-                    existing = self._field_decorators.get(k)
-                    if existing:
-                        existing.extend(v)
-                    self._field_decorators[k] = v[:]
-                self._all_fields_decorators.extend(parent_vf._all_fields_decorators)
-                self._root_decorators.extend(parent_vf._root_decorators)
 
+class Decorator(Generic[DecoratorInfoType], Representation):
```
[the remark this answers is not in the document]
adriangb: added a docstring

pydantic/decorators.py
adriangb: added tests

pydantic/decorators.py
adriangb: thanks removed it

pydantic/_internal/_decorators.py
```diff
@@ -3,202 +3,274 @@
...
-        # function is set later after the class is created and functions are bound
-        self.function: Callable[..., Any] | None = None
+        sub_path: tuple[str | int, ...] | None,
+        check_fields: bool | None,
+    ) -> None:
+        """
+        :param mode: the pydantic-core validator mode.
+        :param type: either 'unbound' or 'field' indicating if this validator should have
+            access to the model itself.
+        :param mode: Not yet supported.
```
dmontagu, line 88: This looks like a typo since it's repeated twice

tests/test_validators.py
adriangb: I did retain the `each_item` tests. This test is weird because it's supposed to be testing `**kwargs` (which is no longer supported and in fact we have / will have in a minute tests proving errors) but is also using `each_item` for no apparent reason. So I'm removing this test in favor of tests that test each one of those things on their own. Unless you see a relationship between `**kwargs` and `each_item` that I am missing.

pydantic/_internal/_decorators.py
```diff
@@ -3,202 +3,274 @@
...
+        *,
+        fields: tuple[str, ...],
+        # pre=True/False in v1 should be converted to mode='before'/'after' in v2
+        mode: Literal['before', 'after'],
+        check_fields: bool | None,
+        each_item: bool,
+    ) -> None:
+        """
+        :param mode: the pydantic-core validator mode.
+        :param mode: Not yet supported.
```
dmontagu, line 61: Is mode supposed to be listed here twice like this?

pydantic/decorator.py
adriangb: I definitely think we don't need both, but like you say not something to change in this PR.

tests/test_validators.py
```diff
@@ -891,41 +941,52 @@ def example_root_validator2(cls, values):
 
     with pytest.raises(ValidationError) as exc_info:
         Model(b='snap dragon', c='snap dragon2')
+    # TODO: we seem to stop early with root validators
+    # In V1 we ran them all (as per this test)
+    # Both V1 and V2 do stop early for function validators _on the same field_
+    # Is this intentional?
```
dmontagu, line 947: Should this comment be deleted now?

pydantic/_internal/_decorators.py
adriangb: thanks, fixed

pydantic/_internal/_decorators.py
adriangb: no, fixed

PR #5238, opened 2023-03-21
Improve classvars collection code/documentation/error messages

dmontagu:
I think these changes snuck their way into main due to too many things happening at once (there was a ClassVars PR already), but I've addressed feedback from #5215 that was fortunately visible due to not being up-to-date with the main branch.

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -237,7 +237,10 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
     @typing.no_type_check
     def __setattr__(self, name, value):
         if name in self.__class_vars__:
-            raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')
+            raise AttributeError(
+                f'"{name}" is a ClassVar of `{self.__class__.__name__}` and cannot be set on an instance. '
+                f'If you want to set a value on the class, use `{self.__class__.__name__}.{name} = value`.'
+            )
         if name.startswith('_'):
             _object_setattr(self, name, value)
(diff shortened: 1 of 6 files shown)


pydantic/main.py
```diff
@@ -237,7 +237,10 @@ def model_post_init(self, **kwargs: Any) -> None:
     @typing.no_type_check
     def __setattr__(self, name, value):
         if name in self.__class_vars__:
-            raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')
+            raise AttributeError(
```
dmontagu, line 240: I'll note that producing this error message is the _only_ reason we collect the class_vars in `collect_fields`, and they are only used in the `pydantic._internal._model_construction.complete_model_class` call to `collect_fields` (one of four calls).

That said, I still think it makes sense to collect them there, and to produce this error message.

tests/test_private_attributes.py
```diff
@@ -131,7 +131,13 @@ class Model(BaseModel):
     m._foo = None
     assert m._foo is None
 
-    with pytest.raises(ValueError, match='"Model" object has no field "_bar"'):
```
dmontagu, line 134: This test was a hold-over from v1, where it checked that classvars did not produce fields. I think adding a clearer error message is an improvement though, and clearly it's not a challenge to do so, which, at least when this test was created, I suspect is related to why it wasn't a separate error message in v1.

dmontagu:
please review

adriangb (approved):
Looks good to me

PR #5239, opened 2023-03-22
Bump pydantic core

dmontagu:


pydantic/_internal/_model_construction.py
```diff
@@ -35,7 +35,7 @@
 object_setattr = object.__setattr__
 
 
-def init_private_attributes(self_: Any, **_kwargs: Any) -> None:
+def init_private_attributes(self_: Any, context: Any) -> None:
```
dmontagu, line 38: Note the `__pydantic_post_init__` function is called with `context=None` in various places, but it is now also called with positional arguments from Rust, so we _have_ to name this argument `context` and accept it positionally.

If we want to change the way `init_private_attributes` is called from python (e.g., in `pydantic.main.BaseModel.__init__`, with `context=None`), I'm okay either way. Doesn't need to happen in this PR, but it can.

pydantic/main.py
```diff
@@ -83,9 +83,9 @@ def __new__(
                     # in a single function
                     namespace['_init_private_attributes'] = _model_construction.init_private_attributes
 
-                    def __pydantic_post_init__(self_: Any, **kwargs: Any) -> None:
-                        self_._init_private_attributes()
-                        self_.model_post_init(**kwargs)
+                    def __pydantic_post_init__(self_: Any, context: Any) -> None:
```
dmontagu, line 86: Here is another place where the signature is dictated by the fact that it needs to accept positional arguments and be callable via keyword from python. (Not sure if we'd prefer this signature or `*args` if it wasn't called with keyword arguments anywhere)

pydantic/main.py
```diff
@@ -243,9 +243,9 @@ def __setattr__(self, name, value):
         elif self.model_config['frozen']:
             raise TypeError(f'"{self.__class__.__name__}" is frozen and does not support item assignment')
         elif self.model_config['validate_assignment']:
-            values, fields_set = self.__pydantic_validator__.validate_assignment(name, value, self.__dict__)
+            values = self.__pydantic_validator__.validate_assignment(self.__dict__, name, value)
             _object_setattr(self, '__dict__', values)
-            self.__fields_set__ |= fields_set
+            self.__fields_set__ |= {name}
```
dmontagu, line 248: I don't know if this is the desired logic here now that `validate_assignment` no longer returns `fields_set`.

I will say the current set of unit tests pass with or without this line; not sure if that's expected due to changes in the logic in `validate_assignment`...

tests/test_main.py
```diff
@@ -245,7 +245,7 @@ class Model(BaseModel):
         a: float
 
     model = Model(a=0.2)
-    with pytest.raises(ValidationError, match='Extra inputs are not permitted'):
+    with pytest.raises(ValidationError, match="Object has no attribute 'b'"):
```
dmontagu, line 248: I'm assuming this is on purpose; I think the error message is nicer now.

dmontagu (commented):
Overall, I do feel the mass of naming changes is noticeable improvement to consistency.

I added comments below everywhere I could see that I had made changes that weren't just renaming following the changes merged in https://github.com/pydantic/pydantic-core/pull/467.

dmontagu:
@samuelcolvin I've just assigned you explicitly to make sure I get your eyes on the few points not related to name changes made above.

dmontagu:
I realized most of the interesting stuff was already part of the init_self PR... Merged that, updated this

PR #5247, opened 2023-03-22
Add DeepChainMap and re-add LimitedDict

dmontagu:
This gets more consistent behavior out of recursive generics, making it substantially more difficult for issues with generic caching to cause problems.

This was necessary to resolve an issue preventing #5235 from passing CI, which I was able to trace back to different behavior in a patch version of python 3.8. (Changing WeakValuesDict to dict in Python 3.8.16 makes the tests pass, and otherwise they fail; this is not necessary in Python 3.8.12.)

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generics.py
+++ b/pydantic/_internal/_generics.py
@@ -3,11 +3,11 @@ from __future__ import annotations
 import sys
 import types
 import typing
-from collections import defaultdict
+from collections import ChainMap, defaultdict
 from contextlib import contextmanager
 from contextvars import ContextVar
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

adriangb (approved):
Approving, since we need this to keep moving forward, but I do think we should revisit in the future

dmontagu:
I'll merge this for now but @samuelcolvin if you have any misgivings I'll be happy to revert / otherwise address feedback when it rolls in.

PR #5249, opened 2023-03-22
Downgrade mypy

adriangb:


dmontagu:
@adriangb I believe you did this because the existing mypy tests were failing in v1.1.1 but not in v1.0.1.

I will note that there may be a good reason _not_ to do this downgrade that we didn't discuss: it may be faster/more thorough about catching type-checking issues outside of the tests.

I'll work on getting the mypy tests to pass somehow when run with mypy 1.1.1.

adriangb:
Agreed it would be great to get 1.1.1 working but for now at least I want local dev and CI to be in sync, and 1.1.1 is failing

PR #5250, opened 2023-03-22
Use constraints files for requirements

adriangb:
Use the `all.txt` as a constraints file for the other lockfiles. This ensures that there is no incompatibilities between the different environments.

Selected Reviewer: @dmontagu


samuelcolvin:
conflicts, please update.

adriangb:
please review

dmontagu:
And to be clear, as far as I'm concerned, this is fine if it's working (and CI seems to indicate it is).

I guess the only thing I would request is adding a brief comment (one line is enough) in the Makefile and/or shell scripts that explains the difference between `refresh_lockfiles` and `rebuild_lockfiles`. I see this is described in the README.md but I'm less likely to find it there when trying to remember what does what.

adriangb:
I added some comments explaining what the commands do and the check for `pip-tools` being installed that you suggested in Slack

samuelcolvin:
This is great, thank you.

PR #5254, opened 2023-03-23
Get nearly all remaining tools tests passing

dmontagu:
I don't know whether we should include these `schema_of` and `schema_json_of` functions in v2 or remove them, but I got the existing unit tests for them passing.


tests/test_parse.py
```diff
@@ -28,14 +28,19 @@ def test_model_validate_submodel():
     assert m.model_dump() == {'a': 10.2, 'b': 10}
 
 
-@pytest.mark.xfail(reason='working on V2')
 def test_model_validate_wrong_model():
```
dmontagu, line 31: This one's not really related, but it was another one that just seemed to need an update to the output to be in v2 convention.

dmontagu:
please review

pydantic/tools.py
```diff
@@ -1,12 +1,18 @@
+from __future__ import annotations
```
adriangb, line 1: Just a nit: was this necessary to get tests passing? If not I would have tables it

pydantic/tools.py
```diff
@@ -15,7 +21,7 @@ def _generate_parsing_type_name(type_: Any) -> str:
 
 
 @lru_cache(maxsize=2048)
```
adriangb, line 23: I wonder if this or other random caches we probably have might be messing with our references and causing those weak ref issues

tests/test_tools.py
```diff
@@ -78,22 +74,21 @@ def test_parse_mapping_as():
...
         '{\n'
-        '  "title": "IntOrStr",\n'
         '  "anyOf": [\n'
         '    {\n'
         '      "type": "integer"\n'
         '    },\n'
         '    {\n'
         '      "type": "string"\n'
         '    }\n'
-        '  ]\n'
```
adriangb, line 97: Are these sorts of strict outputs deterministic? If they’re not, we probably should have some way to at least actually enable it, similar to the sort_keys option in json dumps

adriangb (approved):

pydantic/tools.py
dmontagu: I don't think so, as this isn't used anywhere else.

pydantic/tools.py
dmontagu: I think what happened was, at some point I was importing something for type checking purposes that was defined in an `if TYPE_CHECKING:` block in its file, which was super annoying. Looks like that's gone now, but it seemed reasonable to keep it like this.

adriangb: Yeah no worries that’s a good enough reason. I only ask because I’ve done it accidentally

tests/test_tools.py
dmontagu: Yes, just the order changed due to changes in how the schema is generated in v2 (everything is added as dict keys, so order is preserved). I think there are some minor improvements we _could_ do here (even if just taking more care to get a field ordering we like in GenerateJsonSchema), but I think it's worth ignoring for now.

PR #5256, opened 2023-03-23
(V1) Fix issue where constraints were not allowed on generic models

### LINKED ISSUE #2581: Cannot construct parameterised type that includes validation
### Checks

# Bug

```
             pydantic version: 1.8.1
            pydantic compiled: False
                 install path: /home/chris/projects/external/pydantic/pydantic
               python version: 3.9.2 (default, Mar 24 2021, 22:11:22)  [GCC 9.3.0]
                     platform: Linux-5.4.72-microsoft-standard-WSL2-x86_64-with-glibc2.31
```

# Test Case

```py
@skip_36
def test_construct_parameterized_type_with_validation():
    T = TypeVar('T')

    class Page(GenericModel, Generic[T]):
        page: int = Field(ge=1)
        items: Sequence[T]

    # Check we can perform this assignemnt, this is the actual test
    response_model = Page[str]
    assert response_model.__name__== 'Page[str]'

    # Sanity check the resulting type works as expected
    valid = response_model(page=1, items=[])
    assert valid.page == 1

        response_model(page=0, items=[])
    assert exc_info.value.errors() == [{'loc': ('__root__', 'page'), 'msg': 'sum too large', 'type': 'value_error'}]
```

# Output
```
E           ValueError: On field "page" the following field constraints are set but not enforced: ge. 
E           For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints

pydantic/schema.py:926: ValueError
```

# Description

I have a generic page type that I use for responses in FastAPI. To detect bugs the invariants of this response type are configured as standard validations such as `page: int = Field(ge=1)`.

The type can be used normally, and will have the correct paramerized type when instantiating it, such as `Page[str](page=1, ...)`.

However, when trying to assign the `Type[Page[str]]` to the `response_model` parameter in FastAPI pydantic throws a `ValueError` with the message "On field "page" the following field constraints are set but not enforced: ge."

```
@app.get("/", response_model=Page[str])
def index():
    return Page(page=1, items=["foo", "bar"])
```

Without the assignment to `response_model` this function will work as intended.

While debugging the pydantic code it seems during the evaluation of the assignment to `response_model`, the type being evaluated `annotation` at `pydantic/schema.py:926` is an instance of type `pydantic.fields.DeferredType`.

### LINKED ISSUE #3358: "Field constraints are set but not enforced" with generics
### Checks

# Bug

```
             pydantic version: 1.8.2
            pydantic compiled: False
                 install path: C:\Users\seragiottoc\Miniconda3\envs\domino-prototype\Lib\site-packages\pydantic
               python version: 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:15:42) [MSC v.1916 64 bit (AMD64)]
                     platform: Windows-10-10.0.18363-SP0
```

```py

from pydantic import Field

T = TypeVar("T")


class A(GenericModel, Generic[T]):
    t: list[int] = Field(..., min_items=1)


A[int](t=[1])
...
```
_…
ValueError: On field "t" the following field constraints are set but not enforced: min_items. 
For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints_

dmontagu:
This is a slightly simplified, brought-up-to-date implementation of #3593 (and closes #3358 and #2581).

fix #3358, fix #2581.

I think this should be safe to merge.

I have verified that the added tests are already behaving as desired in v2. (Actually they work better than in v1 — you can add a constraint on a `PositiveInt` without an error in v2, unlike what's happening in the tests here, and which happens on non-generic models in v1 as well.)

If we merge this I will open a separate PR to v2 adding similar tests, but no changes are necessary to the v2 source code for it to handle this all correctly.

skip change file check

### DIFF
--- a/changes/2581-MaratBR.md
+++ b/changes/2581-MaratBR.md
@@ -1,2 +1,2 @@
-Fixes [#2581](/samuelcolvin/pydantic/issues/2581). Using constraints such as `gt` or `le` 
+Allow constraints such as `gt` or `le` on generic models.
 no longer raises `ValueError` due to unenforced constraint.
--- a/changes/2581-MaratBR.md
+++ b/changes/2581-MaratBR.md
@@ -1,0 +1,2 @@
+Fixes [#2581](/samuelcolvin/pydantic/issues/2581). Using constraints such as `gt` or `le` 
+no longer raises `ValueError` due to unenforced constraint.
--- a/changes/2581-MaratBR.md
+++ b/changes/2581-MaratBR.md
@@ -1,2 +1,1 @@
 Allow constraints such as `gt` or `le` on generic models.
-no longer raises `ValueError` due to unenforced constraint.
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -24,6 +24,7 @@ from typing_extensions import Annotated
 from .class_validators import gather_all_validators
 from .fields import DeferredType
 from .main import BaseModel, create_model
+from .schema import get_annotation_from_field_info
 from .types import JsonWrapper
 from .typing import display_as_type, get_all_type_hints, get_args, get_origin, typing_base
 from .utils import all_identical, lenient_issubclass
@@ -383,7 +384,12 @@ def _prepare_model_fields(
         assert field.type_.__class__ is DeferredType, field.type_.__class__
 
         field_type_hint = instance_type_hints[key]
-        concrete_type = replace_types(field_type_hint, typevars_map)
+        concrete_type = get_annotation_from_field_info(
+            replace_types(field_type_hint, typevars_map),
+            field.field_info,
+            field.name,
+            created_model.__config__.validate_assignment,
+        )
         field.type_ = concrete_type
         field.outer_type_ = concrete_type
         field.prepare()
--- a/pydantic/schema.py
+++ b/pydantic/schema.py
@@ -43,6 +43,7 @@ from .fields import (
     SHAPE_SINGLETON,
     SHAPE_TUPLE,
     SHAPE_TUPLE_ELLIPSIS,
+    DeferredType,
     FieldInfo,
     ModelField,
 )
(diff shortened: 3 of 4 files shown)


dmontagu:
(Using the change file from #3593 as @MaratBR deserves the credit)

dmontagu:
please review

changes/2581-MaratBR.md
```diff
@@ -0,0 +1,2 @@
+Fixes [#2581](/samuelcolvin/pydantic/issues/2581). Using constraints such as `gt` or `le`
```
samuelcolvin, line 1:
```suggestion
Allow constraints such as `gt` or `le` on generic models.
```

You could also update the file name to use your handle.

samuelcolvin (commented):
honestly, I just don't trust V1 enough to be sure about these things, but if you're sure this is a good idea, happy to go ahead.

samuelcolvin:
please update.

changes/2581-MaratBR.md
dmontagu: I wanted to give MaratBR credit as it was very insightful, I would not have fixed this without that PR for reference (even if mostly because I don't want to be thinking about v1).

dmontagu:
> honestly, I just don't trust V1 enough to be sure about these things, but if you're sure this is a good idea, happy to go ahead.

I feel pretty confident this shouldn't introduce any new misbehaviors, but I'm also happy to be responsible for cleaning them up if they arise. Worst case I would revert and go back to telling people it won't be fixed in v1. But I do think it should be okay.

Sadly I agree with you its practically impossible to be 100% sure but this logic change seems sound to me.

dmontagu:
Now that we have an alpha release of v2 out, I'm inclined to close this as a won't fix, as I don't want to deal with any potential fallout caused by merging this. But the code is here and ready to merge in case we change our mind and/or users demand it.

alita-moore:
Particularly in machine learning, there are libraries that do and will likely continue to only support pydantic v1 for a long time. So it would be helpful if this could be supported in V1.

PR #5257, opened 2023-03-23
Improve implementation of dunder equals

dmontagu:
Inspired by #3066, with tweaks for v2 and special handling of generic models to simplify possible edge cases.

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -661,10 +661,24 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         return keys
 
     def __eq__(self, other: Any) -> bool:
-        if isinstance(other, BaseModel):
-            return self.model_dump() == other.model_dump()
-        else:
-            return self.model_dump() == other
+        if not isinstance(other, BaseModel):
+            return False
+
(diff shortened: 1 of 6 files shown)


dmontagu:
please review

adriangb:
The new behavior looks good to me. I think this is the sort of thing where we should just make the breaking change 👍

pydantic/main.py
```diff
@@ -652,10 +652,27 @@ def _calculate_keys(
...
+        # only require their generic origin types to be equal, rather than exact type equality.
+        # This prevents headaches like MyGeneric(x=1) != MyGeneric[Any](x=1).
+        self_type = getattr(self, '__pydantic_generic_origin__', None) or self.__class__
+        other_type = getattr(other, '__pydantic_generic_origin__', None) or other.__class__
+
+        if self_type != other_type:
+            return False
+
+        if self.__dict__ != other.__dict__:
+            return False
```
dmontagu, line 668: Previously I had this as a check of a tuple, but that was repeatably about 10% slower in the case where the types were different. The difference when the types were the same was negligible.

samuelcolvin:
awesome, thank you.

PR #5260, opened 2023-03-23
Fix more validator tests

adriangb:


pydantic/_internal/_decorators.py
```diff
@@ -209,10 +209,12 @@ def __init__(
         self,
         cls_var_name: str,
         func: Callable[..., Any],
+        unwapped_func: Callable[..., Any],
```
dmontagu, line 212: Typo

dmontagu (approved):
Other than this typo, looks good to me, I'd suggest auto merge after fixing the typo

pydantic/_internal/_decorators.py
adriangb: resolved

PR #5262, opened 2023-03-23
Fix tests in test_validators_dataclass

dmontagu:


tests/test_validators_dataclass.py
```diff
@@ -41,9 +43,7 @@ def check_a2(cls, v: List[int]) -> List[int]:
     assert MyDataclass(a=[1, 2]).a == [1, 2, 123, 456]
 
 
-@pytest.mark.xfail(reason='working on V2')
 def test_validate_multiple():
-    # also test TypeError
```
dmontagu, line 46: Added a test below demonstrating that type errors aren't handled by the validators

adriangb (approved):
Reviewed, looks like a lot of it was just removing kwargs

tests/test_validators_dataclass.py
```diff
@@ -53,20 +53,50 @@ class MyDataclass:
...
 
-# @pytest.mark.xfail(reason='working on V2')
+def test_type_error():
+    @dataclass
+    class MyDataclass:
+        a: str
+        b: str
+
+        @field_validator('a', 'b')
+        @classmethod
```
adriangb, line 88: This isn't necessary. But I think we should, eventually, add it everywhere and pretend as if it is to encourage users to do the same. But we should do that, and just that, in a future PR to avoid polluting diffs. But this is fine here for now, no need to change it.

adriangb (approved):

PR #5263, opened 2023-03-23
Add warning/errors for non-annotated fields

dmontagu:
I renamed `keep_untouched` to `non_field_types` as I feel this is more descriptive. (See the minor doc updates.)

Open to changing the logic here, and adding unit tests to ensure any scenarios we can think of are handled properly.


dmontagu:
please review

pydantic/_internal/_model_construction.py
```diff
@@ -69,19 +77,28 @@ def inspect_namespace(namespace: dict[str, Any]) -> dict[str, ModelPrivateAttr]:
...
+                    f'Field {var_name!r} defined on a base class was overridden by a non-annotated attribute. '
+                    f'All field definitions, including overrides, require a type annotation.',
+                )
+            elif isinstance(value, FieldInfo):
+                raise PydanticUserError(f'Field {var_name!r} requires a type annotation')
+            else:
+                raise PydanticUserError(
+                    f'A non-annotated attribute was detected: `{var_name} = {value!r}`. All model fields require a '
+                    f'type annotation; if {var_name!r} is not meant to be a field, you may be able to suppress this '
+                    f'warning by annotating it as a ClassVar or updating model_config["non_field_types"].',
```
adriangb, line 100:
```suggestion
                    f'error by annotating it as a ClassVar or updating model_config["non_field_types"].',
```

pydantic/_internal/_model_construction.py
```diff
@@ -30,8 +30,7 @@
 
 __all__ = 'object_setattr', 'init_private_attributes', 'inspect_namespace', 'complete_model_class', 'MockValidator'
 
-
-IGNORED_TYPES: tuple[Any, ...] = (FunctionType, property, type, classmethod, staticmethod)
+IGNORED_TYPES: tuple[Any, ...] = (FunctionType, property, type, classmethod, staticmethod, PydanticDecoratorMarker)
```
adriangb, line 33: What’s the difference between this and PYDANTIC_NON_FIELD_TYPES?

tests/test_edge_cases.py
```diff
@@ -926,11 +926,29 @@ def test_inheritance():
...
+
+    assert Bar2().model_dump() == {'x': 12.3, 'a': 123.0}
+
+    class Bar3(Foo):
         x: float = 12.3
-        a = 123.0
+        a: float = Field(default=123.0)
 
-    assert Bar().model_dump() == {'x': 12.3, 'a': 123.0}
+    assert Bar3().model_dump() == {'x': 12.3, 'a': 123.0}
```
adriangb, line 951: Can you make those numbers different for every Bar? I just want to make sure there’s no way we’re accidentally modifying the base class and getting the values from there

adriangb (approved):

pydantic/_internal/_model_construction.py
samuelcolvin: can we use `ignored_types` instead of `non_field_types`?

pydantic/main.py
```diff
@@ -71,9 +73,20 @@ def __new__(
     ) -> type:
         if _base_class_defined:
             class_vars: set[str] = set()
+            base_private_attributes: dict[str, ModelPrivateAttr] = {}
```
samuelcolvin, line 76: can we move this into a separate function?

samuelcolvin: no worries if it doesn't make sense.

pydantic/main.py
```diff
@@ -134,6 +142,14 @@ def hash_func(self_: Any) -> int:
             )
 
             cls.__pydantic_model_complete__ = False  # Ensure this specific class gets completed
+
+            # preserve `__set_name__` protocol defined in https://peps.python.org/pep-0487
```
samuelcolvin, line 146: is it okay to remove this?

samuelcolvin (commented):
otherwise LGTM.

pydantic/_internal/_model_construction.py
dmontagu: `PrivateAttr` wouldn't be allowed here specifically because you only get this error if the field name doesn't start with an `_`. 

However, while investigating this, I did discover that currently on this branch if you do:
```python

class WarningModel(BaseModel):
    _x = 1
```
You don't get an error _and_ `_x` doesn't become a `PrivateAttr`.

@samuelcolvin I'm assuming we want to make this an error with a similar message except that it also calls out `PrivateAttr` as an option. Does that make sense to you? Would you prefer we make note that `PrivateAttr` is another option if they change the name of the field?

pydantic/main.py
dmontagu: (I'm assuming you mean the actual code, not just this comment.)

There was an issue where it was added where someone specifically wanted to be able to use custom descriptors. (See https://github.com/pydantic/pydantic/pull/4407 and https://github.com/pydantic/pydantic/pull/4410). And there is a test below that requires it (`test_set_name` in test_create_model.py). And I tried adding `model_config = dict(non_field_types=(class_deco,))` to that test and it still failed without this code.

Is that a good enough reason to keep it, or would you still like to remove it?

pydantic/main.py
dmontagu: sure, does this seem good to you?
```python
def _collect_bases_data(bases) -> tuple[set[str], set[str], dict[str, ModelPrivateAttr]]:
    field_names: set[str] = set()
    class_vars: set[str] = set()
    private_attributes: dict[str, ModelPrivateAttr] = {}
    for base in bases:
        if _base_class_defined and issubclass(base, BaseModel) and base != BaseModel:
            # model_fields might not be defined yet in the case of generics, so we use getattr here:
            field_names.update(getattr(base, 'model_fields', {}).keys())
            class_vars.update(base.__class_vars__)
            private_attributes.update(base.__private_attributes__)
    return field_names, class_vars, private_attributes
```

that changes this to:
```python
        if _base_class_defined:
            base_field_names, class_vars, base_private_attributes = _collect_bases_data(bases)
            config_new = build_config(cls_name, bases, namespace, kwargs)
            ...
```

(Let me know if you want a different function name)

tests/test_create_model.py
```diff
@@ -3,24 +3,27 @@
...
 from pydantic import BaseModel, ConfigDict, Extra, Field, ValidationError, create_model, errors
-from pydantic.decorators import field_validator
+from pydantic.decorators import field_validator, validator
 from pydantic.fields import ModelPrivateAttr
 
 
-@pytest.mark.xfail(reason='working on V2')
 def test_create_model():
-    model = create_model('FooModel', foo=(str, ...), bar=123)
+    model = create_model('FooModel', foo=(str, ...), bar=(int, 123))
```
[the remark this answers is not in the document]
dmontagu:
```
pydantic.errors.PydanticUserError: A non-annotated attribute was detected: `bar = 123`. All model fields require a type annotation; if 'bar' is not meant to be a field, you may be able to resolve this error by annotating it as a ClassVar or updating model_config["non_field_types"].
```
Same error you'd get for a non-annotated field outside of `create_model`. Do you want to add more detail if this happens in the context of `create_model`? If you want changes, any detail about how you'd like to handle it would be appreciated.

Either way, I can add a test for this in this file.

pydantic/_internal/_model_construction.py
dmontagu: looks like I added `PYDANTIC_NON_FIELD_TYPES` then refactored things to stop using it; I will drop it.

I will also change `non_field_types` to `ignored_types`.

pydantic/_internal/_model_construction.py
samuelcolvin: don't really care, if it's easy, match V1.

tests/test_create_model.py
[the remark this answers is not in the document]
samuelcolvin: that's fine I think.

PR #5265, opened 2023-03-24
Improve xfail reasons for test_decorators tests (and get many passing)

dmontagu:
Selected Reviewer: @hramezani

### DIFF
--- a/tests/test_decorator.py
+++ b/tests/test_decorator.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from typing import List
 
 import pytest
+from dirty_equals import IsInstance
 from typing_extensions import Annotated, TypedDict
 
 from pydantic import BaseModel, Extra, Field, ValidationError, validate_arguments
@@ -15,7 +16,6 @@ from pydantic.errors import PydanticUserError
 skip_pre_38 = pytest.mark.skipif(sys.version_info < (3, 8), reason='testing >= 3.8 behaviour only')
 
 
-@pytest.mark.xfail(reason='working on V2')
 def test_args():
     @validate_arguments
     def foo(a: int, b: int):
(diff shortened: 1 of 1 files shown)


tests/test_decorator.py
```diff
@@ -28,47 +28,34 @@ def foo(a: int, b: int):
...
+        {
+            'input': 'x',
+            'loc': ('b',),
+            'msg': 'Input should be a valid integer, unable to parse string as an ' 'integer',
+            'type': 'int_parsing',
+        }
     ]
 
-    with pytest.raises(ValidationError) as exc_info:
+    with pytest.raises(TypeError, match='2 positional arguments expected but 3 given'):
```
dmontagu, line 47: My understanding is that we no longer want to produce ValidationError's for these signature-related issues. If that's right, I think this should be good. I could also drop these tests that only have TypeError checks now if desired, or if we _do_ still want to produce ValidationError's for these, I can undo these changes and put them back to xfailing. 

(There are also a number of tests here that were passing but just needed updates to the error formats; we should at least merge those.)

adriangb: > My understanding is that we no longer want to produce ValidationError's for these signature-related issues

Are we just calling the function and letting it fail or something else?

tests/test_decorator.py
```diff
@@ -417,7 +376,7 @@ def func(s: str, count: int, *, separator: bytes = b''):
     stub.assert_not_called()
 
 
-@pytest.mark.xfail(reason='working on V2')
+@pytest.mark.xfail(reason='Annotated does not seem to be respected')
```
adriangb, line 379: Is it the `@validate_arguments` decorator that is not respecting `Annotated`? Or are our models not respecting it? Asking because `@validate_arguments` builds a model internally.

tests/test_decorator.py
```diff
@@ -485,7 +444,7 @@ def foo(dt: datetime = Field(default_factory=lambda: 946684800), /):
     assert module.foo(0) == datetime(1970, 1, 1, tzinfo=timezone.utc)
 
 
-@pytest.mark.xfail(reason='working on V2')
+@pytest.mark.xfail(reason='config["extra"] does not seem to be respected')
```
adriangb, line 447: This is kind of strange to me in the first place. If we think of the decorator as creating a model and the parameters as its fields, then why would the config (which is at the model level) apply to one (all?) of the fields?

I do think that in general we need to figure out a way to apply a config (or the portion of it that is applicable) to a `TypedDict` model.

adriangb (approved):
Left a couple comments but looks good overall

tests/test_decorator.py
dmontagu: Yeah I think these type errors are just what you'd get from calling the function directly.

tests/test_decorator.py
dmontagu: Annotated definitely works on BaseModel fields, and I checked it seems to work for this sort of case:
```python
from typing import Annotated

class Model(BaseModel):
    x: Annotated[int, Field(alias='b')]

print(Model.model_fields)
# {'x': FieldInfo(annotation=int, required=True, alias='b', alias_priority=2)}
```

tests/test_decorator.py
dmontagu: Yeah, I think probably if we are going to raise type errors for signature issues we should not support "extra" in the validate_arguments decorator. I'll admit I haven't ever used this decorator though, so I don't know if people have good reasons to want this sort of thing.

tests/test_decorator.py
adriangb: We honestly should probably just rework `@validate_arguments` internally to create a `Validator` for each argument instead of a model, or something like that.

PR #5272, opened 2023-03-24
Allow serializers to ignore the info argument

adriangb:

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -627,7 +627,7 @@ def ser_x(self, value: Any, handler: pydantic.SerializerFunctionWrapHandler): ..
 """
 
 
-def make_generic_v2_field_serializer(
+def make_generic_field_serializer(
     serializer: AnySerializerFunction, mode: Literal['plain', 'wrap']
 ) -> AnyCoreSerializer:
     """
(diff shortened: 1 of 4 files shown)


tests/test_serialize.py
```diff
@@ -177,22 +177,8 @@ def no_args(not_self: Any, value: Any, nxt: Any, info: Any) -> Any:  # pragma: n
...
-
-        class _(BaseModel):
-            x: int
-
-            # note that type checkers won't pick this one up
-            # we could fix it but it would require some fiddling with the self argument in the
-            # callable protocols
-            @serializer('x', mode='wrap')
-            def no_args(self, value: Any, info: Any) -> Any:  # pragma: no cover
-                ...
```
adriangb, line 191: Note that we can't catch this at runtime or statically anymore because there's no way to distinguish it from `(value, handler)` unless we want to (1) introspect types or (2) force some of these to be keyword arguments (but then we have to construct kwargs in rust).

adriangb:
please review

pydantic/_internal/_decorators.py
```diff
@@ -563,3 +571,105 @@ def _wrapper2(
...
+
+@serializer('x')
+def ser_x(self, value: Any): ...
+
+@serializer('x', mode='wrap')
+def ser_x(self, value: Any, handler: pydantic.SerializerFunctionWrapHandler): ...
+"""
+
+
+def make_generic_v2_field_serializer(
```
samuelcolvin, line 630: don't think we need v2 in the name.

pydantic/_internal/_decorators.py
```diff
@@ -563,3 +571,105 @@ def _wrapper2(
...
+]
+
+AnySerializerFunction = Union[
+    PlainSerializerWithoutInfo,
+    WrapSerializerWithoutInfo,
+    AnyCoreSerializer,
+]
+
+
+_VALID_SERIALIZER_SIGNATURES = """\
```
samuelcolvin, line 593: find with this here for now, but it should end up in the docs.

pydantic/_internal/_decorators.py
```diff
@@ -563,3 +571,105 @@ def _wrapper2(
...
+
+            return _wrap1
+        if n_positional != 2:
+            raise TypeError(
+                f'Unrecognized serializer signature for {serializer} with `mode={mode}`:{sig}\n'
+                f' {_VALID_SERIALIZER_SIGNATURES}'
+            )
+        func = cast(AnyCoreSerializer, serializer)
+        return func
+    assert mode == 'wrap'
```
samuelcolvin, line 661: new line or `else:` would make this easier to read.

samuelcolvin (commented):
otherwise LGTM - though I haven't gone into complete detail.

please update.

PR #5273, opened 2023-03-24
Update remaining test xfail reasons

dmontagu:
The only place where I've left non-updated xfail tests is in `test_orm_mode.py`, which I'm not quite sure what we're going to do with yet, but clearly it's going to get an overhaul.

I was able to fix at least a handful of tests along the way here. There was one place where I changed an error message to make things a little more clear, in a way that I felt was good enough to change a test from xfailing to passing. (I'll point this out below.)

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -88,7 +88,7 @@ class CustomValidator(ABC):
         attrs = attrs or set(self.__slots__)  # type: ignore[attr-defined]
         for k, v in constraints.items():
             if k not in attrs:
-                raise TypeError(f'{self.__class__.__name__} has no attribute {k!r}')
+                raise TypeError(f'{k!r} is not a valid constraint for {self.__class__.__name__}')
             setattr(self, k, v)
 
 
--- a/tests/test_dataclasses.py
+++ b/tests/test_dataclasses.py
@@ -480,7 +480,7 @@ def test_default_factory_singleton_field():
     assert Foo().singleton is Foo().singleton
 
 
-@pytest.mark.xfail(reason='JSON schema')
+@pytest.mark.xfail(reason='dataclasses JSON schema')
 def test_schema():
     @pydantic.dataclasses.dataclass
     class User:
@@ -522,7 +522,7 @@ def test_schema():
     }
 
 
-@pytest.mark.xfail(reason='JSON schema')
+@pytest.mark.xfail(reason='dataclasses JSON schema')
 def test_nested_schema():
     @pydantic.dataclasses.dataclass
     class Nested:
@@ -771,7 +771,7 @@ def test_override_builtin_dataclass_nested():
     assert foo.file.meta.seen_count == 7
 
 
-@pytest.mark.xfail(reason='JSON schema')
+@pytest.mark.xfail(reason='dataclasses JSON schema')
 def test_override_builtin_dataclass_nested_schema():
     @dataclasses.dataclass
     class Meta:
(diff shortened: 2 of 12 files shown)


pydantic/_internal/_fields.py
```diff
@@ -88,7 +88,7 @@ def _update_attrs(self, constraints: dict[str, Any], attrs: set[str] | None = No
         attrs = attrs or set(self.__slots__)  # type: ignore[attr-defined]
         for k, v in constraints.items():
             if k not in attrs:
-                raise TypeError(f'{self.__class__.__name__} has no attribute {k!r}')
+                raise TypeError(f'{k!r} is not a valid constraint for {self.__class__.__name__}')
```
dmontagu, line 91: This is the only change to non-test code in the PR

dmontagu:
please review

tests/test_edge_cases.py
```diff
@@ -1910,27 +1909,25 @@ def __init__(self, t1: T1, t2: T2):
...
+
+            t1_f = Validator(args[0])
+            t2_f = Validator(args[1])
+
+            def validate(v, info):
+                if not args:
+                    return v
+                # TODO: Collect these errors, rather than stopping early, and modify the loc to make the test pass
+                t1_f(v.t1)
+                t2_f(v.t2)
```
adriangb, line 1927: Since we want to allow validators to raise a `ValidationError`, I wonder if the solution here is to let validators raise a `ValueError[List[ValLineError]]` (or similar, whatever we want the Python API to be)? Then they would also be able to change the `loc`. We can include the `loc` that was used to get to them via `info`.

tests/test_dataclasses.py
```diff
@@ -1434,7 +1434,7 @@ def double_b(cls, v, _):
     assert d.b == 5.0
 
 
-@pytest.mark.xfail(reason='working on V2 - validator in child not applied')
+@pytest.mark.xfail(reason='validator in child not applied')
```
adriangb, line 1437: If this means that `B`'s `validate_a` method is not called it sounds like a bug in the collection of validators and should be an easy fix.

tests/test_edge_cases.py
```diff
@@ -1149,7 +1149,7 @@ class Model(BaseModel):
     assert exc_info.value.errors() == [{'input': {}, 'loc': ('bar',), 'msg': 'Field required', 'type': 'missing'}]
 
 
-@pytest.mark.xfail(reason='working on V2 - validators')
+@pytest.mark.xfail(reason='items yielded by __get_validators__ are not inspected for valid signatures')
```
adriangb, line 1152: Should we be inspecting them? I was thinking that the `@validator` versions would get inspected and support several signatures for brevity but the versions defined by `__get_validators__` would only support the exact pydantic-core signauture and be enforced only via typing. Reason being that `@validator` gets used a lot more frequently so it's nice to be succint and get more errors caught early but `__get_validators__` is more rare thus it's okay to force users to be more verbose.

adriangb (approved):

adriangb:
Great stuff!

tests/test_edge_cases.py
dmontagu: yeah I'm not sure what's best but I think a lot of things could work and that proposal sounds reasonable

tests/test_edge_cases.py
dmontagu: Not sure, but that _is_ the reason the test is failing 😄

adriangb: Yep next pass may be fixing any signatures that aren’t V2 compliant and deleting this test…

PR #5274, opened 2023-03-24
Get generic typed dicts working

dmontagu:
Also did some cleanup of namespace inference that I think will resolve issues for some other types, I'll point out in-line

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -22,6 +22,22 @@ if TYPE_CHECKING:
     from ..fields import FieldInfo
 
 
+def get_type_hints_infer_globalns(
+    obj: Any,
+    localns: dict[str, Any] | None = None,
+    include_extras: bool = False,
+) -> dict[str, Any]:
+    module_name = getattr(obj, '__module__', None)
+    globalns: dict[str, Any] | None = None
+    if module_name:
+        try:
+            globalns = sys.modules[module_name].__dict__
+        except KeyError:
+            # happens occasionally, see https://github.com/pydantic/pydantic/issues/2363
+            pass
+    return get_type_hints(obj, globalns=globalns, localns=localns, include_extras=include_extras)
+
+
 class _UndefinedType:
     """
     Singleton class to represent an undefined value.
@@ -123,18 +139,9 @@ def collect_fields(  # noqa: C901
     """
     from ..fields import FieldInfo
 
-    module_name = getattr(cls, '__module__', None)
-    global_ns: dict[str, Any] | None = None
-    if module_name:
-        try:
-            global_ns = sys.modules[module_name].__dict__
-        except KeyError:
-            # happens occasionally, see https://github.com/pydantic/pydantic/issues/2363
-            pass
-
     # get type hints and raise a PydanticUndefinedAnnotation if any types are undefined
     try:
-        type_hints = get_type_hints(cls, global_ns, types_namespace, include_extras=True)
+        type_hints = get_type_hints_infer_globalns(cls, types_namespace, include_extras=True)
     except NameError as e:
         try:
             name = e.name
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -30,7 +30,7 @@ from ._decorators import (
     SerializerDecoratorInfo,
     ValidatorDecoratorInfo,
 )
-from ._fields import PydanticGeneralMetadata, PydanticMetadata, Undefined, collect_fields
+from ._fields import PydanticGeneralMetadata, PydanticMetadata, Undefined, collect_fields, get_type_hints_infer_globalns
 from ._forward_ref import PydanticForwardRef
 from ._generics import replace_types
 
@@ -285,7 +285,7 @@ class GenerateSchema:
         elif _typing_extra.is_literal_type(obj):
             return self._literal_schema(obj)
         elif is_typeddict(obj):
-            return self._typed_dict_schema(obj)
+            return self._typed_dict_schema(obj, None)
         elif _typing_extra.is_namedtuple(obj):
             return self._namedtuple_schema(obj)
         elif _typing_extra.is_new_type(obj):
@@ -345,6 +345,8 @@ class GenerateSchema:
             return self._counter_schema(obj)
         elif origin in (typing.Dict, dict):
             return self._dict_schema(obj)
+        elif is_typeddict(origin):
+            return self._typed_dict_schema(obj, origin)
         elif issubclass(origin, typing.Dict):
             # Subclasses of typing.Dict may be handled as subclasses of dict; see note above
             return self._dict_subclass_schema(obj)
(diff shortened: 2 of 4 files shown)


pydantic/_internal/_fields.py
```diff
@@ -22,6 +22,22 @@
...
+) -> dict[str, Any]:
+    module_name = getattr(obj, '__module__', None)
+    globalns: dict[str, Any] | None = None
+    if module_name:
+        try:
+            globalns = sys.modules[module_name].__dict__
+        except KeyError:
+            # happens occasionally, see https://github.com/pydantic/pydantic/issues/2363
+            pass
+    return get_type_hints(obj, globalns=globalns, localns=localns, include_extras=include_extras)
```
dmontagu, line 38: This logic should be reused in various places (anywhere that you know that `obj` would definitely have its module be valid as the globalns for its type hints; this will be the case any time `obj` is a user-defined type, such as a BaseModel subclass not created with `create_model`, typeddict, namedtuple)

pydantic/_internal/_generate_schema.py
```diff
@@ -345,6 +345,8 @@ def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
             return self._counter_schema(obj)
         elif origin in (typing.Dict, dict):
             return self._dict_schema(obj)
+        elif is_typeddict(origin):
```
dmontagu, line 348: the trick is that a _parametrized_ generic typeddict will be a _GenericAlias, which returns False from is_typeddict, but its _origin_ will return True.

pydantic/_internal/_generate_schema.py
```diff
@@ -522,6 +524,12 @@ def _typed_dict_schema(
         Hence to avoid creating validators that do not do what users expect we only
         support typing.TypedDict on Python >= 3.11 or typing_extension.TypedDict on all versions
         """
+        if origin is not None:
+            typeddict_typevars_map = dict(zip(origin.__parameters__, typed_dict_cls.__args__))
+            typed_dict_cls = origin
+        else:
+            typeddict_typevars_map = {}
```
dmontagu, line 531: this is a little hacky where I replace the typed_dict_cls with the origin, but it made for a small diff here and I think it's clear enough. (What's happening is I'm pulling the typevars mapping off of the origin+_GenericAlias pair, and then applying that mapping down below to each field in turn)

pydantic/_internal/_generate_schema.py
```diff
@@ -538,7 +546,10 @@ def _typed_dict_schema(
             recursive_schema = core_schema.definition_reference_schema(obj_ref)
             self._recursion_cache[obj_ref] = recursive_schema
 
-        for field_name, annotation in _typing_extra.get_type_hints(typed_dict_cls, include_extras=True).items():
+        for field_name, annotation in get_type_hints_infer_globalns(
+            typed_dict_cls, localns=self.types_namespace, include_extras=True
```
dmontagu, line 550: I confirmed that when you define a custom typeddict class, the module is the module in which it was defined, not the module of TypedDict, so this use of `get_type_hints_infer_globalns` seems sound to me.

pydantic/_internal/_generate_schema.py
```diff
@@ -568,7 +579,9 @@ def _namedtuple_schema(self, namedtuple_cls: Any) -> core_schema.CallSchema:
         """
         Generate schema for a NamedTuple.
         """
-        annotations: dict[str, Any] = _typing_extra.get_type_hints(namedtuple_cls, include_extras=True)
+        annotations: dict[str, Any] = get_type_hints_infer_globalns(
```
dmontagu, line 582: It seems to me the same infer-globalns logic deserved to apply here as well. (I found this by doing find-everywhere search for `get_type_hints`.)

tests/test_types_typeddict.py
```diff
@@ -470,15 +467,14 @@ def module():
 
         T = TypeVar('T')
 
+        class RecursiveGenTypedDictModel(BaseModel, Generic[T]):
+            rec: 'RecursiveGenTypedDict[T]'
+            model_config = dict(undefined_types_warning=False)
+
         class RecursiveGenTypedDict(TypedDict, Generic[T]):
-            # TODO: See if we can get this working if defined in a function (right now, needs to be module-level)
```
dmontagu, line 474: I'm proud I was able to get this working — this was a matter of passing the global/local namespaces in the right places

dmontagu:
please review

dmontagu:
@samuelcolvin  @adriangb hooky assigned @hramezani, so I'll just assign both of you and you can decide who should review

pydantic/_internal/_fields.py
adriangb: This is great. I think it's fine here for now, it's where it's being used.

pydantic/_internal/_generate_schema.py
adriangb: Makes sense. I'm not sure how we handle it for BaseModel / dataclasses (the latter I think the answer is we don't, yet?) but we should probably make them all mirror images of each other?

pydantic/_internal/_generate_schema.py
adriangb: Let's hope they don't decided to erase that in the future like they do with the bases...

pydantic/_internal/_generate_schema.py
```diff
@@ -568,7 +579,9 @@ def _namedtuple_schema(self, namedtuple_cls: Any) -> core_schema.CallSchema:
...
+            namedtuple_cls, include_extras=True, localns=self.types_namespace
```
adriangb, line 583: It seems like `localns` is always `self.types_namespace`. Should we rename `self.types_namespace` to `localns` elsewhere? Or should we have some sort of explanation in the docstrings of `GenerateSchema` explaining what it expects `types_namespace` to be and how you should get it (`sys._getframe(1).f_globals.copy()` of wherever you are called?).

adriangb (approved):

pydantic/_internal/_generate_schema.py
dmontagu: probably; I haven't really looked into how it gets used, when wrestling with generics it has mostly "just worked" for me. I think when we get generic dataclasses working and/or when we refactor generics to drop the contextvar, it will clear things up a bit.

pydantic/_internal/_generate_schema.py
dmontagu: I agree — I'll just note that it won't surprise me at all if this needs (simplicity-reducing) changes to work in the most general cases (like recursive nested generics that involve chains of generic basemodels, dataclasses, and typeddicts sharing parameters etc.).

But considering this handles what are likely to be by far the most common cases, I think it's worth doing this now even if we never perfectly handle the most complicated things imaginable.

I do plan to do another unifying pass over the generics stuff at some point though, just to be clear.

PR #5276, opened 2023-03-25
Fix dataclass decorators

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -249,10 +249,15 @@ def collect_fields(  # noqa: C901
                 if cls.__pydantic_generic_parameters__:  # model can be parametrized
                     assert cls.__pydantic_generic_defaults__ is not None
                     cls.__pydantic_generic_defaults__[ann_name] = default
-                if not (is_dataclass and not isinstance(default, (dataclasses.Field, FieldInfo))):
+                delattr(cls, ann_name)
+                if (
+                    is_dataclass
+                    and isinstance(default, (dataclasses.Field, FieldInfo))
+                    and default.default not in (Undefined, dataclasses.MISSING)
+                ):
                     # for dataclasses we only delete the attribute if it is an instance of
                     # field, e.g. `a: int = 1` should not get deleted
-                    delattr(cls, ann_name)
+                    setattr(cls, ann_name, default.default)
             except AttributeError:
                 pass  # indicates the attribute was on a parent class
 
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -255,8 +255,9 @@ def collect_fields(  # noqa: C901
                     and isinstance(default, (dataclasses.Field, FieldInfo))
                     and default.default not in (Undefined, dataclasses.MISSING)
                 ):
-                    # for dataclasses we only delete the attribute if it is an instance of
-                    # field, e.g. `a: int = 1` should not get deleted
+                    # for dataclasses we preserve the default value if it is set
+                    # field, e.g. `a: int = 1` gets kept as is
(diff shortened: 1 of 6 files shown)


adriangb:
please review

pydantic/_internal/_fields.py
```diff
@@ -246,10 +246,13 @@ def collect_fields(  # noqa: C901
...
             # 2. To avoid false positives in the NameError check above
             try:
-                delattr(cls, ann_name)
                 if cls.__pydantic_generic_parameters__:  # model can be parametrized
                     assert cls.__pydantic_generic_defaults__ is not None
                     cls.__pydantic_generic_defaults__[ann_name] = default
+                if not (is_dataclass and not isinstance(default, (dataclasses.Field, FieldInfo))):
+                    # for dataclasses we only delete the attribute if it is an instance of
+                    # field, e.g. `a: int = 1` should not get deleted
+                    delattr(cls, ann_name)
```
adriangb, line 255: Before this, the following didn't work:

```python
import pydantic

class Parent:
    a: int = 1

print(Parent.a)  # AttributeError: type object 'Parent' has no attribute 'a'
```

This fixes it and doesn't break anything else, _but_ the example does currently work for `BaseModel` and I can't tell why. So `BaseModel` must be doing something slightly different, which I don't like because I would like them to both behave the same and share the same implementation in this regard at least.

pydantic/_internal/_generate_schema.py
```diff
@@ -450,13 +450,14 @@ def generate_dc_field_schema(
...
-            schema = wrap_default(field_info, schema)
-
         schema = apply_validators(
             schema, filter_field_decorator_info_by_field(decorators.field_validator.values(), name)
         )
         schema = apply_validators(schema, filter_field_decorator_info_by_field(decorators.validator.values(), name))
+
+        if not field_info.is_required():
+            schema = wrap_default(field_info, schema)
+
```
adriangb, line 460: This is the same issue we've run into before where if you put a function validator outside of a WithDefault then the WithDefault will not work. This fixes it and doesn't seem to break anything, but there is also the option of making WithDefault recursive in pydantic-core (i.e. the function validator would call into the schema it wraps to see if it has a default). This would make sense for function validators, but I'm not sure where else, if it even makes sense in general or if there are even any cons to just forcing WithDefault to be the outermost schema.

pydantic/_internal/_fields.py
dmontagu: As noted on slack — let's just modify this so that if it is a dataclasses.Field or FieldInfo that has a default, it does `setattr(cls, ann_name, default)` (same as it does in the standard dataclasses logic).

adriangb: [8c98fae](https://github.com/pydantic/pydantic/pull/5276/commits/8c98fae49b0c86de055449b4fe50009750e98ae1)

pydantic/_internal/_fields.py
```diff
@@ -246,10 +246,18 @@ def collect_fields(  # noqa: C901
...
                     cls.__pydantic_generic_defaults__[ann_name] = default
+                delattr(cls, ann_name)
+                if (
+                    is_dataclass
+                    and isinstance(default, (dataclasses.Field, FieldInfo))
+                    and default.default not in (Undefined, dataclasses.MISSING)
+                ):
+                    # for dataclasses we only delete the attribute if it is an instance of
+                    # field, e.g. `a: int = 1` should not get deleted
+                    setattr(cls, ann_name, default.default)
```
dmontagu, line 260: I think the comment above this line has become a bit confusing now that it is right above where you setattr, with all of the above conditions. Not a big deal if you don't want to make more changes

dmontagu (approved):

PR #5287, opened 2023-03-27
Use pydantic core 0.17.1

dmontagu:
(This is going to fail CI until the 0.17.0 release is live, but I'll open the PR now anyway.)


dmontagu:
Will go straight to 0.17.1 instead of 0.17.0

PR #5289, opened 2023-03-27
Validate always

dmontagu:
Implements the handling of `validate_default` for fields and the `always=True` argument to v1-style `@validator` decorator. Also adds the `validate_default` as a keyword argument to model_config.

### DIFF
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -336,6 +336,8 @@ In this case you should set `check_fields=False` on the validator.
 
 Validators also work with *pydantic* dataclasses.
 
+# TODO: Change this example so that it _should_ use a validator; right now it would be better off with default_factory..
+
 ```py
 from datetime import datetime
 
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -336,16 +336,16 @@ In this case you should set `check_fields=False` on the validator.
 
 Validators also work with *pydantic* dataclasses.
 
-```py test="xfail - currently decorators are messing up dataclass field defaults #5271"
+```py
 from datetime import datetime
 
-from pydantic import field_validator
+from pydantic import Field, field_validator
 from pydantic.dataclasses import dataclass
 
 
 @dataclass
 class DemoDataclass:
-    ts: datetime = None
+    ts: datetime = Field(None, validate_default=True)
 
     @field_validator('ts', mode='before')
     def set_ts_now(cls, v):
@@ -353,5 +353,7 @@ class DemoDataclass:
 
 
 print(DemoDataclass())
+#> DemoDataclass(ts=datetime.datetime(2032, 1, 2, 3, 4, 5, 6))
 print(DemoDataclass(ts='2017-11-08T14:00'))
+#> DemoDataclass(ts=datetime.datetime(2017, 11, 8, 14, 0))
 ```
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -59,8 +59,9 @@ class ValidatorDecoratorInfo(Representation):
         fields: tuple[str, ...],
         # pre=True/False in v1 should be converted to mode='before'/'after' in v2
         mode: Literal['before', 'after'],
-        check_fields: bool | None,
         each_item: bool,
+        always: bool,
+        check_fields: bool | None,
     ) -> None:
         """
         :param mode: the pydantic-core validator mode.
@@ -70,8 +71,9 @@ class ValidatorDecoratorInfo(Representation):
         """
         self.fields = fields
         self.mode = mode
-        self.check_fields = check_fields
         self.each_item = each_item
+        self.always = always
+        self.check_fields = check_fields
 
 
 class FieldValidatorDecoratorInfo(Representation):
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -13,7 +13,8 @@ from typing import TYPE_CHECKING, Any, Callable, ForwardRef, Iterable, Mapping, 
 
 from annotated_types import BaseMetadata, GroupedMetadata
 from pydantic_core import SchemaError, SchemaValidator, core_schema
-from typing_extensions import Annotated, Literal, get_args, get_origin, is_typeddict
+from pydantic_core.core_schema import dict_not_none
+from typing_extensions import Annotated, Literal, Required, TypedDict, get_args, get_origin, is_typeddict
 
 from ..errors import PydanticSchemaGenerationError, PydanticUserError
 from ..fields import FieldInfo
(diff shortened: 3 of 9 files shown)


pydantic/_internal/_decorators.py
```diff
@@ -59,8 +59,9 @@ def __init__(
         fields: tuple[str, ...],
         # pre=True/False in v1 should be converted to mode='before'/'after' in v2
         mode: Literal['before', 'after'],
-        check_fields: bool | None,
         each_item: bool,
+        always: bool,
+        check_fields: bool | None,
```
dmontagu, line 64: While adding always, I just put the arguments in the same order here as they are in the actual decorator definition

docs/usage/validators.md
```diff
@@ -336,22 +336,26 @@ In this case you should set `check_fields=False` on the validator.
 
 Validators also work with *pydantic* dataclasses.
 
-```py test="xfail - currently decorators are messing up dataclass field defaults #5271"
+# TODO: Change this example so that it _should_ use a validator; right now it would be better off with default_factory..
```
dmontagu, line 339: Let me know if I should change this test in a different way, or if I should indicate the need to change it in another way than this TODO comment.

pydantic/_internal/_generate_schema.py
```diff
@@ -53,6 +54,31 @@
 ]
 
 
+class _CommonField(TypedDict, total=False):
```
dmontagu, line 57: I'm guessing you guys aren't going to like this approach; I'm happy to rework it in whatever way makes sense, I just want to make sure that we can share as much of the field-building logic as possible between typed dict / model fields and dataclass fields

pydantic/decorators.py
```diff
@@ -116,21 +117,51 @@ def __call__(self, __cls: Any, __values: _decorators.RootValidatorValues) -> _de
...
     :param fields: which field(s) the method should be called on
+    :param pre: whether or not this validator should be called before the standard validators (else after)
+    :param each_item: for complex objects (sets, lists etc.) whether to validate individual elements rather than the
+      whole object
+    :param always: whether this method and other validators should be called even if the value is missing
     :param check_fields: whether to check that the fields actually exist on the model
     :param allow_reuse: whether to track and raise an error if another validator refers to the decorated function
     """
-    fields = tuple((__field, *fields))
+    if not fields:
```
dmontagu, line 152: I re-added all these errors, mostly because I ended up getting surprising behavior while dev'ing on this because I did something that was type-invalid but it didn't raise any runtime errors and just silently didn't validate anything.

Considering this is only executed during the decorator call (not during actual validation) I feel like these checks (and error messages at point-of-failure) are worth including. I've also added them to `field_validator`, and re-added the v1 tests for this behavior that had been removed.

I am happy to undo this if preferred, but it would have helped me during development if they were there.

dmontagu: Also, I left them as `TypeError` for now; I'll convert them into some form of PydanticError with proper error codes etc. as part of that work.

dmontagu:
please review

pydantic/decorators.py
adriangb: Personally I think these are fine as TypeError. They should be quite self explanatory.

pydantic/decorators.py
```diff
@@ -116,21 +117,51 @@ def __call__(self, __cls: Any, __values: _decorators.RootValidatorValues) -> _de
 )
 
 
+# Use a single overload here to get a different signature that requires at least one field be specified
```
adriangb, line 120: Why don’t we do the same thing just without an overload? Or at least using TYPE_CHECKING? A single overload is wrong

pydantic/decorators.py
```diff
@@ -116,21 +117,51 @@ def __call__(self, __cls: Any, __values: _decorators.RootValidatorValues) -> _de
...
     """
-    fields = tuple((__field, *fields))
+    if not fields:
+        raise TypeError('validator with no fields specified')
+    elif isinstance(fields[0], FunctionType):
+        raise TypeError(
+            'validators should be used with fields and keyword arguments, not bare. '
+            "E.g. usage should be `@validator('<field_name>', ...)`"
+        )
+    elif not all(isinstance(field, str) for field in fields):
```
adriangb, line 159: The other ones I agree with, this one I think is overkill

adriangb (changes requested):

pydantic/_internal/_generate_schema.py
adriangb: When I first saw this, I didn't like it. But seeing the actual usage I'm more okay with it. I think what would really help is a docstring / comment giving some background and explaining why it's beneficial to have this class/function.

docs/usage/validators.md
samuelcolvin: i agree this case should use a `default_factory`.

pydantic/_internal/_generate_schema.py
samuelcolvin: i feel vaguely the same as @adriangb, but can we move it to the bottom (or near the bottom) of the file, rather than up here.

pydantic/decorators.py
samuelcolvin: I think worth use `PydanticTypeError` down the road.

samuelcolvin (commented):
otherwise looking good.

please update.

pydantic/decorators.py
dmontagu: > A single overload is wrong

I did this for two main reasons:

* I prefer the IDE behavior of a single overload, because jump-to-definition takes me to the implementation, rather than the `@overload` type signature. With `if TYPE_CHECKING:`, it will take me to the type-checking signature rather than the implementation, which I usually don't want (cough cough ConfigDict, 😩).

* I wanted it to be visually similar to the `@field_validator` decorator. In that case, because there are currently two signatures, the `@overload` is allowed. I'll note that even there, in principle we don't need two `@overload`s — with 
  ```python
  _V2ValidatorType = TypeVar(
      '_V2ValidatorType', bound=Union[V2WrapValidator, V2Validator, 'classmethod[Any]', 'staticmethod[Any]']
  )
  ```
  it could be expressed with one:
  ```python
  def field_validator(...) -> Callable[[_V2ValidatorType], _V2ValidatorType]:
  ```
  That's not to say that I'd suggest we change it — I prefer that it use @overload because it lets us have the different (compatible) signature in the implementation without resorting to the (in my opinion) less ergonomic `if TYPE_CHECKING:`.

-----

At any rate, I am happy using any of the following approaches:

* Single `@overload` (current approach)
* Changing the single `@overload` to `if TYPE_CHECKING:`
* Just dropping the `@overload` signature (and losing the mypy error for not specifying even one field name)
* Changing the implementation signature to `(__field, *fields, ...` and relying on the missing-positional-argument TypeError produced by python, rather than the more explicit error message.

Just let me know which you prefer.

adriangb: > Changing the implementation signature to (__field, *fields, ... and relying on the missing-positional-argument TypeError produced by python, rather than the more explicit error message.

I prefer this version. You’ll also have to check that __field is not a function if you want to give a runtime error for calling with no args

pydantic/decorators.py
dmontagu: if we drop the check that the positional items being passed to @field_validator are of type str, this test:
```python
        class Model(BaseModel):
            a: str
            b: str

            @field_validator(['a', 'b'])
            def check_fields(cls, v):
                return v
```
goes from raising: `TypeError: field_validator fields should be passed as separate string args.`
to raising: `TypeError: unhashable type: 'list'`

The docstring of the test (`test_field_validator_bad_fields_throws_configerror`) explicitly says the goal was to raise a custom error in the case of passing a list of str instead of multiple string args. (Admittedly, we don't do that now, but the plan would be to make it raise a pydantic-specific error in those cases as part of a separate PR.) Even given that, do you want to drop this error case?

pydantic/decorators.py
adriangb: Also, we definitely need the overload on field_validator to enforce the relationship between mode and the function signature

dmontagu: Missed that, makes sense

dmontagu: I've changed it back to using `__field, *fields` in the implementation and dropped the single overload.

adriangb (approved):

PR #5297, opened 2023-03-28
Fix more dataclass xfail tests

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -101,7 +101,7 @@ def prepare_dataclass(
 
     extra = config.get('extra', Extra.forbid)
     if extra != Extra.forbid:
-        raise ValueError(f'extra=Extra.{extra} is not allowed for dataclasses. Only Extra.forbid is allowed.')
+        raise ValueError(f'extra=Extra.{extra.value} is not allowed for dataclasses. Only Extra.forbid is allowed.')
 
     if config.get('validate_assignment', False) is True:
 
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -10,6 +10,8 @@ from typing import Any, Callable, ClassVar
 
 from pydantic_core import ArgsKwargs, SchemaSerializer, SchemaValidator, core_schema
 
+from pydantic.config import Extra
+
 from ..errors import PydanticUndefinedAnnotation
 from ..fields import FieldInfo
 from . import _decorators
@@ -96,6 +98,10 @@ def prepare_dataclass(
     cls.__pydantic_validator__ = validator = SchemaValidator(schema, core_config)
     # this works because cls has been transformed into a dataclass by the time "cls" is called
     cls.__pydantic_serializer__ = SchemaSerializer(schema, core_config)
+
+    extra = config.get('extra', Extra.forbid)
+    if extra != Extra.forbid:
+        raise ValueError(f'extra=Extra.{extra} is not allowed for dataclasses. Only Extra.forbid is allowed.')
 
     if config.get('validate_assignment', False) is True:
 
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -106,7 +106,7 @@ def prepare_dataclass(
             ' Only Extra.ignore (the default) and Extra.forbid are allowed.'
         )
 
-    if config.get('validate_assignment', False) is True:
+    if config.get('validate_assignment'):
 
         @wraps(cls.__setattr__)
         def validated_setattr(instance: Any, __field: str, __value: str) -> None:
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -5,6 +5,7 @@ from __future__ import annotations as _annotations
 
 import typing
 import warnings
+from functools import wraps
 from typing import Any, Callable, ClassVar
 
 from pydantic_core import ArgsKwargs, SchemaSerializer, SchemaValidator, core_schema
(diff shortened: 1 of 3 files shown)


adriangb:
please review

pydantic/_internal/_dataclasses.py
```diff
@@ -92,10 +95,22 @@ def prepare_dataclass(
...
     core_config = generate_config(config, cls)
     cls.__pydantic_fields__ = fields
-    cls.__pydantic_validator__ = SchemaValidator(schema, core_config)
+    cls.__pydantic_validator__ = validator = SchemaValidator(schema, core_config)
     # this works because cls has been transformed into a dataclass by the time "cls" is called
     cls.__pydantic_serializer__ = SchemaSerializer(schema, core_config)
 
+    extra = config.get('extra', Extra.forbid)
+    if extra != Extra.forbid:
+        raise ValueError(f'extra=Extra.{extra.value} is not allowed for dataclasses. Only Extra.forbid is allowed.')
```
dmontagu, line 104: I'm curious the reasoning behind removing support for this.

I think it's fairly common in FastAPI world to want to allow extra fields to be passed into your API (e.g. so you can remove ignored fields from your model without it causing errors), and there have been many issues over the years of people wanting to use dataclasses instead of BaseModel in their fastapi endpoints. 

And today, you can allow extra for dataclasses in FastAPI:
```
import pydantic.dataclasses

from fastapi import FastAPI
from fastapi.testclient import TestClient

app = FastAPI()


class MyBody:
    x: int


@app.post("/items/", include_in_schema=False)
async def read_items(body: MyBody):
    return {"x": body.x}


client = TestClient(app)

response = client.post("/items/", json={'x': 1, "y": 2})
assert response.status_code == 200, response.text
assert response.json() == {"x": 1}, response.json()
```

I would be amenable to the idea of dropping support for this if it causes serious issues (or if the plan is to do something different for FastAPI validation, somehow). I think if we want to make this change though and won't support extra fields for FastAPI body arguments annotated with dataclasses, we should probably run that by Sebastian for feedback.

dmontagu: I realized I wasn't clear in some of my wording above: I think it is important to support `extra='ignore'`, I am fine dropping support for `extra='allow'` for dataclasses.

dmontagu:
please update

pydantic/_internal/_dataclasses.py
adriangb: Made the default `ignore` and prohibited just `allow`. Thanks folks.

adriangb:
please review

pydantic/_internal/_dataclasses.py
```diff
@@ -92,10 +95,25 @@ def prepare_dataclass(
...
     core_config = generate_config(config, cls)
     cls.__pydantic_fields__ = fields
-    cls.__pydantic_validator__ = SchemaValidator(schema, core_config)
+    cls.__pydantic_validator__ = validator = SchemaValidator(schema, core_config)
     # this works because cls has been transformed into a dataclass by the time "cls" is called
     cls.__pydantic_serializer__ = SchemaSerializer(schema, core_config)
 
+    extra = config.get('extra', Extra.ignore)  # ignore is the default in V1
+    if extra is Extra.allow:
+        raise ValueError(
```
dmontagu, line 104: Super minor, but should this be a TypeError? (I'm assuming we'll give a custom error type in the future anyway.)

If you want to leave as is and just get it merged I'm fine with that.

adriangb: No it should be a ValueError. It's the right type but the value isn't allowed.

pydantic/_internal/_dataclasses.py
```diff
@@ -92,10 +95,25 @@ def prepare_dataclass(
...
+            'extra=Extra.allow is not allowed for dataclasses.'
+            ' Only Extra.ignore (the default) and Extra.forbid are allowed.'
+        )
+
+    if config.get('validate_assignment', False) is True:
```
dmontagu, line 109:
```suggestion
    if config.get('validate_assignment'):
```
I'm also fine keeping as is if you think that's clearer

adriangb: Good catch. I've been trying to use `get` because (1) it plays nicer with my type checker / IDE and (2) we've talked about getting rid of the magic defaults so I'm just not relying on them.

pydantic/_internal/_dataclasses.py
dmontagu: I meant it as a misconfiguration of the type you are creating, but your explanation makes sense.

tests/test_dataclasses.py
```diff
@@ -131,20 +133,6 @@ def double_a(cls, v):
...
-    class MyDataclass:
-        a: int
-
-    d = MyDataclass(1)
-    assert d.a == 1
-
-    d.extra_field = 1.23
-    assert d.extra_field == 1.23
-    d.extra_field = 'bye'
-    assert d.extra_field == 'bye'
```
dmontagu, line 145: I think this was removed because you can't set missing fields with validate_assignment=True, but I would suggest we add a test of the new behavior:
```python
def test_validate_assignment_extra_missing_fields():
    @pydantic.dataclasses.dataclass(config=dict(validate_assignment=True), frozen=False)
    class MyDataclass:
        a: int

    d = MyDataclass(1)
    assert d.a == 1

        d.extra_field = 1.23
        {
            'ctx': {'attribute': 'extra_field'},
            'input': 1.23,
            'loc': ('extra_field',),
            'msg': "Object has no attribute 'extra_field'",
            'type': 'no_such_attribute',
        }
    ]
```

dmontagu: Whoops, no, those are only for BaseModel not dataclass

dmontagu (approved):
Looks good to me other than I think it makes sense to add the test I shared above.

tests/test_dataclasses.py
adriangb: That test only works because `extra` in `ConfigDict` is ignored and `forbid` is used as the default. So we'd have to rewrite that test once we get `extra` (and `frozen`) working for dataclasses. So can we table it until then?

adriangb: In other words, I think the tests we want are:

```python
@pytest.mark.parametrize('config', [
    ConfigDict(validate_assignment=True),
    ConfigDict(validate_assignment=True, extra=Extra.ignore)
])
def test_validate_assignment_extra_ignore(config: ConfigDict):
    @pydantic.dataclasses.dataclass(config=config)
    class MyDataclass:
        a: int

    d = MyDataclass(1)
    assert d.a == 1

    d.extra_field = 1.23
    assert not hasattr(d, 'extra_field')


def test_validate_assignment_extra_forbid():
    @pydantic.dataclasses.dataclass(config=ConfigDict(validate_assignment=True, extra=Extra.forbid))
    class MyDataclass:
        a: int

    d = MyDataclass(1)
    assert d.a == 1

        d.extra_field = 1.23
        {
            'ctx': {'attribute': 'extra_field'},
            'input': 1.23,
            'loc': ('extra_field',),
            'msg': "Object has no attribute 'extra_field'",
            'type': 'no_such_attribute',
        }
    ]
```

dmontagu: > So can we table it until then?

I just was hesitant to delete a test as I worry it would be forgotten. If you don't think that's a concern I'm fine tabling.

adriangb: After discussion offline, I was wrong. I added a ~ similar test to the one you suggested 😄

PR #5308, opened 2023-03-30
Replace Validator with AnalyzedType

### LINKED ISSUE #5213: create free standing functions to match model methods
* a new module with all the methods - what the hell do we call this module?
* call from `BaseModel` methods
* call from `Validator` methods #4669
* add `BareModel`
* remove the `model_` prefix from the free standing functions
* extra functions we don't want to put on a model, like `dump_json_bytes`

Incomplete list of functions:
* `validate(type[T], data: any) -> T`
* `validate_json(type[T], data: str | bytes) -> T`
* `construct(type[T], data) -> T`
* `dump -> any`
* `dump_json -> str`
* `dump_json_bytes -> bytes`
* `json_schema -> dict`
* `json_schema_json -> bytes`

These should work on models, `BareModel`, dataclasses, and instrances of Validator (#4669).

Name ideas for this module: `standalone`, `methods` (technically they're the opposite), `free`???

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -985,18 +985,19 @@ except ValidationError as e:
     """
 ```
 
-## Model creation from `NamedTuple` or `TypedDict`
+## Using Pydantic without creating a BaseModel
 
-Sometimes you already use in your application classes that inherit from `NamedTuple` or `TypedDict`
-and you don't want to duplicate all your information to have a `BaseModel`.
-For this _pydantic_ provides `create_model_from_namedtuple` and `create_model_from_typeddict` methods.
-Those methods have the exact same keyword arguments as `create_model`.
+You may have types that are not `BaseModel`s that you want to validate data against.
+Or you may want to validate a `List[SomeModel]`, or dump it to JSON.
 
+To do this Pydantic provides `AnalyzedType`. An `AnalyzedType` instance behaves nearly the same as a `BaseModel` instance, with the difference that `AnalyzedType` is not an actual type so you cannot use it in type annotations and such.
 
 ```py
+from typing import List
+
 from typing_extensions import TypedDict
 
-from pydantic import ValidationError, Validator
+from pydantic import AnalyzedType, ValidationError
 
 
 class User(TypedDict):
@@ -1004,22 +1005,24 @@ class User(TypedDict):
     id: int
 
 
-UserValdiator = Validator(User)
-print(repr(UserValdiator(dict(name='Fred', id='3'))))
-#> {'name': 'Fred', 'id': 3}
+UserListValidator = AnalyzedType(List[User])
+print(repr(UserListValidator.validate_python([{'name': 'Fred', 'id': '3'}])))
+#> [{'name': 'Fred', 'id': 3}]
 
 try:
-    UserValdiator(dict(name='Fred', id='wrong', other='no'))
+    UserListValidator.validate_python([{'name': 'Fred', 'id': 'wrong', 'other': 'no'}])
 except ValidationError as e:
     print(e)
     """
-    2 validation errors for typed-dict
-    id
+    2 validation errors for list[typed-dict]
+    0 -> id
       Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='wrong', input_type=str]
-    other
+    0 -> other
       Extra inputs are not permitted [type=extra_forbidden, input_value='no', input_type=str]
     """
 ```
+
+For many use cases `AnalyzedType` can replace BaseModels with a `__root__` field in Pydantic V1.
 
 ## Custom Root Types
 
--- a/pydantic/analyzed_type.py
+++ b/pydantic/analyzed_type.py
@@ -1,0 +1,241 @@
+from __future__ import annotations as _annotations
+
+import sys
+from typing import TYPE_CHECKING, Any, Dict, Generic, Iterable, Set, TypeVar, Union, overload
+
+from pydantic_core import CoreConfig, CoreSchema, SchemaSerializer, SchemaValidator, core_schema
+from typing_extensions import Literal
+
+from pydantic.config import ConfigDict
+from pydantic.json_schema import DEFAULT_REF_TEMPLATE, GenerateJsonSchema
+
+from ._internal import _generate_schema, _typing_extra
+
+T = TypeVar('T')
+
+if TYPE_CHECKING:
+    # should be `set[int] | set[str] | dict[int, IncEx] | dict[str, IncEx] | None`, but mypy can't cope
+    IncEx = Union[Set[int], Set[str], Dict[int, Any], Dict[str, Any]]
+
+
+def _get_schema(type_: Any, config: CoreConfig | None, parent_depth: int) -> CoreSchema:
+    """
+      BaseModel uses it's own __module__ to find out where it was defined
+    and then look for symbols to resolve forward references in those globals
+    On the other hand this function can be called with arbitrary objects,
+    including type aliases where __module__ (always `typing.py`) is not useful
+    So instead we look at the globals in our parent stack frame
+    This works for the case where this function is called in a module that
+    has the target of forward references in its scope but
+    does not work for more complex cases
+    for example, take the following:
+
+    a.py
+    ```python
+    from typing import List, Dict
+    IntList = List[int]
+    OuterDict = Dict[str, 'IntList']
+    ```
+
+    b.py
+    ```python
+    from pydantic import AnalyzedType
+    from a import OuterDict
+    IntList = int  # replaces the symbol the forward reference is looking for
+    v = AnalyzedType(OuterDict)
+    v({"x": 1})  # should fail but doesn't
+    ```
+
+    If OuterDict were a BaseModel this would work because it would resolve
+    the forward reference within the `a.py` namespace.
+    But `AnalyzedType(OuterDict)`
+    can't know what module OuterDict came from.
+    In other words, the assumption that _all_ forward references exist in the
+    module we are being called from is not technically always true
+    Although most of the time it is and it works fine for recursive models and such/
+    BaseModel's behavior isn't perfect either and _can_ break in similar ways,
+    so there is no right or wrong between the two.
+    But at the very least this behavior is _subtly_ different from BaseModel's.
+    """
+    arbitrary_types = bool((config or {}).get('arbitrary_types_allowed', False))
+    local_ns = _typing_extra.parent_frame_namespace(parent_depth=parent_depth)
+    global_ns = sys._getframe(max(parent_depth - 1, 1)).f_globals.copy()
+    global_ns.update(local_ns or {})
+    gen = _generate_schema.GenerateSchema(arbitrary_types=arbitrary_types, types_namespace=global_ns, typevars_map={})
+    return gen.generate_schema(type_)
+
+
+# TODO: merge / replace this with _internal/_generate_schema.py::generate_config
+# once we change the config logic to make ConfigDict not be a partial
+def _translate_config(config: ConfigDict) -> core_schema.CoreConfig:
+    """
+    Create a pydantic-core config from a pydantic config.
+    """
+    unset: Any = object()
+    core_config: dict[str, Any] = dict(
+        title=config['title'] if 'title' in config and config['title'] is not None else unset,
+        typed_dict_extra_behavior=config['extra'].value if 'extra' in config and config['extra'] is not None else unset,
+        allow_inf_nan=config['allow_inf_nan'] if 'allow_inf_nan' in config else unset,
+        populate_by_name=config['populate_by_name'] if 'populate_by_name' in config else unset,
+        str_strip_whitespace=config['str_strip_whitespace'] if 'str_strip_whitespace' in config else unset,
+        str_to_lower=config['str_to_lower'] if 'str_to_lower' in config else unset,
+        str_to_upper=config['str_to_upper'] if 'str_to_upper' in config else unset,
+        strict=config['strict'] if 'strict' in config else unset,
+        ser_json_timedelta=config['ser_json_timedelta'] if 'ser_json_timedelta' in config else unset,
+        ser_json_bytes=config['ser_json_bytes'] if 'ser_json_bytes' in config else unset,
+        from_attributes=config['from_attributes'] if 'from_attributes' in config else unset,
+        validate_default=config['validate_default'] if 'validate_default' in config else unset,
+        str_max_length=(
+            config['str_max_length'] if 'str_max_length' in config and config['str_max_length'] is not None else unset
+        ),
+        str_min_length=config['str_min_length'] if 'str_min_length' in config else unset,
+    )
+    return CoreConfig(**core_config)  # type: ignore[misc]
+
+
+class AnalyzedType(Generic[T]):
+    if TYPE_CHECKING:
+
+        @overload
+        def __new__(cls, __type: type[T], *, config: ConfigDict | None = ...) -> AnalyzedType[T]:
+            ...
+
+        # this overload is for non-type things like Union[int, str]
+        # Pyright currently handles this "correctly", but MyPy understands this as AnalyzedType[object]
+        # so an explicit type cast is needed
+        @overload
+        def __new__(cls, __type: T, *, config: ConfigDict | None = ...) -> AnalyzedType[T]:
+            ...
+
+        def __new__(cls, __type: Any, *, config: ConfigDict | None = ...) -> AnalyzedType[T]:
+            raise NotImplementedError
+
+    def __init__(self, __type: Any, *, config: ConfigDict | None = None, _parent_depth: int = 2) -> None:
+        core_config: CoreConfig
+        if config is not None:
+            core_config = _translate_config(config)
+        else:
+            core_config = CoreConfig()
+        try:
+            core_config.update(__type.__pydantic_core_config__)
+        except AttributeError:
+            pass
+
+        core_schema: CoreSchema
+        try:
+            core_schema = __type.__pydantic_core_schema__
+        except AttributeError:
+            core_schema = _get_schema(__type, core_config, parent_depth=_parent_depth + 1)
+
+        validator: SchemaValidator
+        try:
+            validator = __type.__pydantic_validator__
+        except AttributeError:
+            validator = SchemaValidator(core_schema, core_config)
+
+        serializer: SchemaSerializer
+        try:
+            serializer = __type.__pydantic_serializer__
+        except AttributeError:
+            serializer = SchemaSerializer(core_schema, core_config)
+
+        self.core_schema = core_schema
+        self.validator = validator
+        self.serializer = serializer
+
+    def validate_python(self, __object: Any) -> T:
+        return self.validator.validate_python(__object)
+
+    def validate_json(self, __data: str | bytes) -> T:
+        return self.validator.validate_json(__data)
+
+    def dump_python(
+        self,
+        __instance: T,
+        *,
+        mode: Literal['json', 'python'] = 'python',
+        include: IncEx | None = None,
+        exclude: IncEx | None = None,
+        by_alias: bool = False,
+        exclude_unset: bool = False,
+        exclude_defaults: bool = False,
+        exclude_none: bool = False,
+        round_trip: bool = False,
+        warnings: bool = True,
+    ) -> Any:
+        return self.serializer.to_python(
+            __instance,
+            mode=mode,
+            by_alias=by_alias,
+            include=include,
+            exclude=exclude,
+            exclude_unset=exclude_unset,
+            exclude_defaults=exclude_defaults,
+            exclude_none=exclude_none,
+            round_trip=round_trip,
+            warnings=warnings,
+        )
+
+    def dump_json(
+        self,
+        __instance: T,
+        *,
+        indent: int | None = None,
+        include: IncEx | None = None,
+        exclude: IncEx | None = None,
+        by_alias: bool = False,
+        exclude_unset: bool = False,
+        exclude_defaults: bool = False,
+        exclude_none: bool = False,
+        round_trip: bool = False,
+        warnings: bool = True,
+    ) -> bytes:
+        return self.serializer.to_json(
+            __instance,
+            indent=indent,
+            include=include,
+            exclude=exclude,
+            by_alias=by_alias,
+            exclude_unset=exclude_unset,
+            exclude_defaults=exclude_defaults,
+            exclude_none=exclude_none,
+            round_trip=round_trip,
+            warnings=warnings,
+        )
+
+    def json_schema(
+        self,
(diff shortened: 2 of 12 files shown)


pydantic/main.py
```diff
@@ -937,54 +935,11 @@ def __init__(self, __type: Any, *, config: CoreConfig | None = None) -> None:
...
-        gen = _generate_schema.GenerateSchema(
-            arbitrary_types=arbitrary_types, types_namespace=global_ns, typevars_map={}
-        )
-        schema = gen.generate_schema(__type)
-        self._validator = SchemaValidator(schema, config=merged_config)
+        # TODO: clean this up somehow? should we just import this at the module level?
+        from pydantic.funcs import analyze_type
+
+        schema = analyze_type(__type, config=config, parent_depth=3)  # type: ignore
+        self._validator = SchemaValidator(schema.core_schema, config=schema.core_config)
```
adriangb, line 942: Not sure how we should deal with this and also what the relationship between `Validator` and the free-standing functions will be. Given that we have `analyze_types` I think there's something to be said for not introducing this thin abstraction over `SchemaValidator` and having users use `pydantic_core` directly? There's also now issues related to stack depth since I'm introducing another call here, it's making one of our tests fail...

pydantic/funcs.py
```diff
@@ -0,0 +1,301 @@
...
+def analyze_type(__type: type[T], *, config: CoreConfig | None = ...) -> TypeMetadata[T]:
+    ...
+
+
+@overload
+def analyze_type(__type: Any, *, config: CoreConfig | None = ...) -> TypeMetadata[Any]:
+    ...
+
+
+def analyze_type(
```
adriangb, line 34: I'm open to other naming or changes here, but I think in general it makes sense to have a single function to do the type analysis and be able to feed that into the other functions.

pydantic/funcs.py
```diff
@@ -0,0 +1,301 @@
...
+    # IntList = List[int]
+    # OuterDict = Dict[str, 'IntList']
+    # ```
+    #
+    # b.py
+    # ```python
+    # from pydantic import Validator
+    # from a import OuterDict
+    # IntList = int  # replaces the symbol the forward reference is looking for
+    # v = Validator(OuterDict)
```
adriangb, line 84: this comment would need updating since this is no longer within `Validator`

pydantic/funcs.py
```diff
@@ -0,0 +1,301 @@
...
+    __instance: Any,
+    __maybe_type: Any = None,
+    *,
+    schema: TypeMetadata[Any] | None = None,
+) -> bytes:
+    if schema is None and __maybe_type is None:
+        raise TypeError('You must provide an existing schema or a type to analyze')
+    elif schema is None:
+        schema = analyze_type(__maybe_type)
+    serializer = SchemaSerializer(schema.core_schema, schema.core_config)
```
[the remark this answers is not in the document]
adriangb: Yup done in bdf4320

pydantic/funcs.py
```diff
@@ -0,0 +1,541 @@
...
+def json_schemas(
+    __types_or_schemas: Iterable[AnalyzedType[Any] | type[Any]],
+    *,
+    by_alias: bool = True,
+    ref_template: str = DEFAULT_REF_TEMPLATE,
+    title: str | None = None,
+    description: str | None = None,
+    schema_generator: type[GenerateJsonSchema] = GenerateJsonSchema,
+) -> dict[str, Any]:
+    # TODO: can we use model.__schema_cache__?
```
dmontagu, line 523: I removed this in one of my PRs and replaced it with a WeakKeysDict that I think was an improvement. Considering there's no weird recursion issues during schema building I think better to not have the schemas live on the models (since this also means we get improved schema caching for dataclasses etc.)

adriangb:
please review

adriangb:
We should definitely not have this in `pydantic.funcs` since is' not really functions anymore. Should this be in the top level namespace? Or any better module names @samuelcolvin ?

docs/usage/models.md
```diff
@@ -996,20 +996,21 @@ Those methods have the exact same keyword arguments as `create_model`.
...
 
 
 class User(TypedDict):
     name: str
     id: int
 
 
-UserValdiator = Validator(User)
-print(repr(UserValdiator(dict(name='Fred', id='3'))))
+UserValdiator = analyze_type(User)
```
dmontagu, line 1008:
```suggestion
UserValidator = analyze_type(User)
```

pydantic/funcs.py
```diff
@@ -0,0 +1,226 @@
...
+    core_schema = (
+        __types_or_schema.core_schema
+        if isinstance(__types_or_schema, AnalyzedType)
+        else analyze_type(__types_or_schema).core_schema
+    )
+
+    return schema_generator_instance.generate(core_schema)
+
+
+def json_schemas(
```
dmontagu, line 199: this is a bit awkward now if everything else is living on AnalyzedType. Don't have a great suggestion

pydantic/tools.py
```diff
@@ -39,7 +22,7 @@ def parse_obj_as(type_: type[T], obj: Any, type_name: NameFactory | None = None)
         warnings.warn(
             'The type_name parameter is deprecated. parse_obj_as no longer creates temporary models', stacklevel=2
         )
-    return Validator(type_)(obj)
+    return funcs.analyze_type(type_).validate_python(obj)
```
dmontagu, line 25: We should probably explicitly deprecate `parse_obj_as` if this is the intended path now. Don't think we need to do that prior to alpha though. Maybe adding a TODO comment about it is a good idea though.

pydantic/funcs.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class AnalyzedType(Generic[T]):
+    def __init__(self, core_schema: core_schema.CoreSchema, core_config: CoreConfig) -> None:
+        self.core_schema = core_schema
+        self.core_config = core_config
+        self._validator: SchemaValidator | None = None
+        self._schema: SchemaSerializer | None = None
+
+    @property
```
dmontagu, line 30: any reason not to use `functools.cached_property` here? I think it cleans things up a bit as you can drop the `_valdiator` and `_schema` properties and don't need the if checks in the property. I guess in principle it's faster too as it sets an instance attribute when called.

pydantic/funcs.py
```diff
@@ -0,0 +1,226 @@
...
+    >>> assert validate(schema, ['123']) == [123]
+
+    Args:
+        __type: The type to generate a core schema for.
+        config: The pydantic-core config. Defaults to None.
+
+    Returns:
+        core_schema.CoreSchema: A pydantic-core CoreSchema.
+    """
+    # if this is a model with an existing schema short circuit
```
dmontagu, line 129: Have you thought about putting this logic in the `AnalyzedType.__init__` so you can just do `AnalyzedType(Dict[str, MyModel])`? I'm fine as is if you prefer it, I just imagine it could reduce the number of imports if you want to use `AnalyzedType` as an annotation for a custom function but can't instantiate it without the `analyze_type` function.

pydantic/funcs.py
adriangb: Isn’t it only available in 3.8?

dmontagu:
> We should definitely not have this in `pydantic.funcs` since is' not really functions anymore. Should this be in the top level namespace? Or any better module names @samuelcolvin ?

I like the idea of `AnalyzedType` (or whatever it ends up being called?) being in the top level namespace. I think other than the name, it is at least as worthy as `Validator` was (and I do think `Validator` deserved to be there).

Of the following two options:
* Moving `analyze_type` to `AnalyzedType.__init__` (I made a comment about this in the review above) and putting just `AnalyzedType` in the top level namespace
* Putting `analyze_type` and `AnalyzedType` in the top-level namespace 
I personally find the first option is a little more palatable than the second, but I'm good either way.

pydantic/funcs.py
adriangb: I have two reasons for this:
1. It’s easier to do overloads on functions because init doesn’t return anything
2. I want us to be able to construct an AnalyzedType with zero cost if we already have the core schema.

pydantic/tools.py
adriangb: I agree we should deprecate it.

pydantic/funcs.py
adriangb: I think I can move `json_schema` to be an instance method. Maybe `json_schemas` can be a static method?

pydantic/funcs.py
adriangb: But what I can do is make this a static method on AnalyzeType

dmontagu: both good points

pydantic/funcs.py
dmontagu: sounds good to me

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,217 @@
...
+    def analyze_type(__type: type[T], *, config: CoreConfig | None = None, parent_depth: int = 2) -> AnalyzedType[T]:
+        ...
+
+    @overload
+    @staticmethod
+    def analyze_type(__type: Any, *, config: CoreConfig | None = None, parent_depth: int = 2) -> AnalyzedType[Any]:
+        ...
+
+    @staticmethod
+    def analyze_type(__type: Any, *, config: CoreConfig | None = None, parent_depth: int = 2) -> AnalyzedType[Any]:
```
dmontagu, line 108: I wonder if we should change this argument name to `_parent_depth` to discourage users from trying to use it lol

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,217 @@
...
+        """Generate a pydantic-core schema and other metadata
```
dmontagu, line 109:
```suggestion
"""
Generate a pydantic-core-schema and other metadata
```
Is it normal to have text on same line as the start of the docstring? Not trying to be annoying here if you're doing this on purpose.

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,217 @@
...
+        that can be fed into other functions or Validator.
```
dmontagu, line 110: I think maybe this docstring needs updating now that it's not for `Validator`, not sure if it makes sense to just change it to `AnalyzedType` or the text should be changed a bit more deeply.

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,217 @@
...
+
+        >>> from typing import List
+        >>> from pydantic import AnalyzedType
+        >>> schema = AnalyzedType.analyze_type(List[int])
+        >>> assert validate(schema, ['123']) == [123]
+
+        Args:
+            __type: The type to generate a core schema for.
```
dmontagu, line 118: Probably makes sense to change this to `The type to analyze`

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,217 @@
...
+        # from pydantic import AnalyzedType
+        # from a import OuterDict
+        # IntList = int  # replaces the symbol the forward reference is looking for
+        # v = AnalyzedType.analyze_type(OuterDict)
+        # v({"x": 1})  # should fail but doesn't
+        # ```
+        #
+        # If OuterDict were a BaseModel this would work because it would resolve
+        # the forward reference within the `a.py` namespace.
+        # But `AnalyzedType.analyze_type(OuterDict)`
```
dmontagu, line 162: Just reflecting on this a bit, maybe we really should be accepting a namespace argument here _and_ in model_rebuild (rather than a parent namespace depth). I think the parent namespace thing is a good default behavior in most cases but accepting an arbitrary set of definitions does seem like it kind of resolves this in whatever edge case you can come up with.

Maybe a better API (here and elsewhere) would be for this argument to have type `int | dict[str, Any]`, and if you pass a dict it doesn't do any sys.getframe stuff and just uses the passed dict directly.

When I first started thinking about all this namespace stuff I understood it a lot less than I do now. (Not that I would say I understand it fully now..)

Doesn't need to be fixed in this PR, but I think we should maybe revisit this before the official v2 release.

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,212 @@
...
+        >>> assert validate(schema, ['123']) == [123]
+
+        Args:
+            __type: The type to generate a core schema for.
+            config: The pydantic-core config. Defaults to None.
+
+        Returns:
+            core_schema.CoreSchema: A pydantic-core CoreSchema.
+        """
+        # if this is a model with an existing schema short circuit
```
dmontagu, line 124: Seems like we should also have it "short circuit" if you call `analyze_type` on an `AnalyzedType` — and just return the same thing. Would make it a little easier to get analyzed types for a collection of things of which some might already be analyzed types.

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,212 @@
...
+        by_alias: bool = True,
+        ref_template: str = DEFAULT_REF_TEMPLATE,
+        schema_generator: type[GenerateJsonSchema] = GenerateJsonSchema,
+    ) -> dict[str, Any]:
+        schema_generator_instance = schema_generator(by_alias=by_alias, ref_template=ref_template)
+        return schema_generator_instance.generate(self.core_schema)
+
+    @staticmethod
+    def json_schemas(
+        __analyzed_types: Iterable[AnalyzedType[Any]],
```
dmontagu, line 189: as discussed elsewhere, I would suggest changing this to `__types: Iterable[Type[Any]]` and then in the first line of the actual function do something like
```python
__analyzed_types = (AnalyzedType.analyze_type(t) for t in __types)
```
(would need to make sure `AnalyzedType.analyze_type` is a no-op on `AnalyzedType` instances)

docs/usage/models.md
adriangb: done

pydantic/_analyzed_type.py
adriangb: done

pydantic/_analyzed_type.py
adriangb: lets implement that when there's a concrete use case

pydantic/_analyzed_type.py
samuelcolvin, line 1: Good, but the model should be public.

pydantic/_analyzed_type.py
```diff
@@ -0,0 +1,213 @@
...
+
+if TYPE_CHECKING:
+    # should be `set[int] | set[str] | dict[int, IncEx] | dict[str, IncEx] | None`, but mypy can't cope
+    IncEx = Union[Set[int], Set[str], Dict[int, Any], Dict[str, Any]]
+
+
+class AnalyzedType(Generic[T]):
+    def __init__(
+        self,
+        core_schema: core_schema.CoreSchema,
```
samuelcolvin, line 26: Maybe I'm missing something, but this should definitely take type as the first and only positional argument no?

samuelcolvin (commented):
Lots is good here, but I think we need to rethink the main interface.

I think simplicity of first experience is the most important consideration, and `__init__` taking a core schema makes it much harder to use.

pydantic/_analyzed_type.py
samuelcolvin: *module

pydantic/_analyzed_type.py
adriangb: That's what `AnalyzedType.analyze_type` (which I would rename to `.analyze`) is for, more in https://github.com/pydantic/pydantic/pull/5308#discussion_r1154709859

pydantic/_analyzed_type.py
adriangb: There's only one type here, and we want to export it to the `pydantic` namespace. Why don't we give ourselves the ability to rename or refactor this module in the future?

pydantic/_analyzed_type.py
adriangb: I'm open to accepting a namespace or a call depth to get it from instead of just the call depth. Then `BaseModel` could use the fact that it knows it's own module and whatnot.

pydantic/_analyzed_type.py
samuelcolvin: I don't agree, if we need a zero overhead type, make it private in the background.

The primary interface for processing (validating and serialising) data that doesn't fit into model, **must** be simple and obvious to use.

Otherwise we risk becoming attrs.

pydantic/_analyzed_type.py
samuelcolvin: We're in alpha, we can rename things easily.

We have private stuff in `_internal` and public stuff in public modules, let's not add a new form of private.

pydantic/_analyzed_type.py
adriangb: We might not be the only ones needing a zero-overhead type. The pattern of keeping `__init__` low overhead and doing extra work in static/class methods is widely used and well understood. But if you really feel that strongly I can try to move this into `__init__`...

pydantic/_analyzed_type.py
adriangb: Okay I just renamed it to `analyzed_type`

pydantic/_analyzed_type.py
samuelcolvin: I'm not opposed to having a public method with zero overhead, but the main interface we expose (along side BaseModel, dataclass, and validate_aeguments) needs to have the signature `Thing(TheType).validate(data)`.

Thanks for understanding.

adriangb:
please review

docs/usage/models.md
```diff
@@ -985,42 +985,47 @@ except ValidationError as e:
     """
 ```
 
-## Model creation from `NamedTuple` or `TypedDict`
+## Using Pydantic with creating a BaseModel
```
samuelcolvin, line 988:
```suggestion
## Using Pydantic without creating a BaseModel
```

docs/usage/models.md
```diff
@@ -985,42 +985,47 @@ except ValidationError as e:
...
     name: str
     id: int
 
 
-UserValdiator = Validator(User)
-print(repr(UserValdiator(dict(name='Fred', id='3'))))
-#> {'name': 'Fred', 'id': 3}
+UserList = List[User]
+
+UserListValidator = AnalyzedType(UserList)
```
samuelcolvin, line 1010:
```suggestion
UserListValidator = AnalyzedType(List[User])
```
otherwise inexperienced users might think you need to define `UserList` first.

pydantic/analyzed_type.py
```diff
@@ -0,0 +1,241 @@
+from __future__ import annotations
```
samuelcolvin, line 1:
```suggestion
from __future__ import annotations as _annotations
```
otherwise it's easy to accidentally use `annotations`.

pydantic/analyzed_type.py
```diff
@@ -0,0 +1,241 @@
...
+from pydantic_core import CoreConfig, CoreSchema, SchemaSerializer, SchemaValidator, core_schema
+from typing_extensions import Literal
+
+from pydantic.config import ConfigDict
+from pydantic.json_schema import DEFAULT_REF_TEMPLATE, GenerateJsonSchema
+
+from ._internal import (
+    _generate_schema,
+    _typing_extra,
+)
```
samuelcolvin, line 15:
```suggestion
from ._internal import _generate_schema, _typing_extra
```
ruff can be bad at correcting imports layout.

pydantic/analyzed_type.py
```diff
@@ -0,0 +1,241 @@
...
+
+T = TypeVar('T')
+
+if TYPE_CHECKING:
+    # should be `set[int] | set[str] | dict[int, IncEx] | dict[str, IncEx] | None`, but mypy can't cope
+    IncEx = Union[Set[int], Set[str], Dict[int, Any], Dict[str, Any]]
+
+
+def _get_schema(type_: Any, config: CoreConfig | None, parent_depth: int) -> CoreSchema:
+    arbitrary_types = bool((config or {}).get('arbitrary_types_allowed', False))
```
samuelcolvin, line 25: any particular reason to wrap with `bool` here?

pydantic/analyzed_type.py
```diff
@@ -0,0 +1,241 @@
...
+    local_ns = _typing_extra.parent_frame_namespace(parent_depth=parent_depth)
+    # BaseModel uses it's own __module__ to find out where it was defined
```
samuelcolvin, line 27: very good, might be easier to read as a docstring, even if we don't publish it in documentation.

pydantic/analyzed_type.py
```diff
@@ -0,0 +1,241 @@
...
+    # BaseModel's behavior isn't perfect either and _can_ break in similar ways,
+    # so there is no right or wrong between the two.
+    # But at the very least this behavior is _subtly_ different from BaseModel's.
+    global_ns = sys._getframe(max(parent_depth - 1, 1)).f_globals.copy()
+    global_ns.update(local_ns or {})
+    gen = _generate_schema.GenerateSchema(arbitrary_types=arbitrary_types, types_namespace=global_ns, typevars_map={})
+    return gen.generate_schema(type_)
+
+
+def _translate_config(config: ConfigDict) -> core_schema.CoreConfig:
```
samuelcolvin, line 69: Is this the same as `_internal/_generate_schema.py::generate_config`?

samuelcolvin: if it's not the same now because `config` really is partial, we should add a TODO to replace it when we update config logic.

pydantic/analyzed_type.py
```diff
@@ -0,0 +1,241 @@
...
+        def __new__(cls, __type: Any, *, config: ConfigDict | None = ...) -> AnalyzedType[T]:
+            raise NotImplementedError
+
+    def __init__(self, __type: Any, *, config: ConfigDict | None = None, _parent_depth: int = 2) -> None:
+        core_config: CoreConfig
+        if config is not None:
+            core_config = _translate_config(config)
+        else:
+            core_config = CoreConfig()
+        if hasattr(__type, '__pydantic_core_config__'):
```
samuelcolvin, line 118: for all these, might be cleaner to do

```py
try:
    core_config.update(__type.__pydantic_core_config__)
except Attribute_error:
    pass
```

pydantic/json_schema.py
```diff
@@ -15,8 +15,7 @@
 from .errors import PydanticInvalidForJsonSchema, PydanticUserError
 
 if TYPE_CHECKING:
-    from ._internal._dataclasses import PydanticDataclass
-    from .main import BaseModel
+    pass
```
samuelcolvin, line 18: remove?

samuelcolvin (commented):
All small things, otherwise looks great.

please update.

pydantic/analyzed_type.py
adriangb: after all of the .get() and {} it's untyped, wrapping it in a bool is an easy way to tell type checkers that, well, it's a bool

adriangb:
Please review

PR #5309, opened 2023-03-30
`__subclasscheck__` customisation

samuelcolvin:
To match `__isinstancecheck__`.


samuelcolvin:
doesn't seem to work. 🤷

PR #5310, opened 2023-03-30
Make model_dump_json return a str

dmontagu:
Selected Reviewer: @adriangb


tests/test_docs.py
```diff
@@ -12,7 +12,8 @@
 import pytest
 from pytest_examples import CodeExample, EvalExample, find_examples
 
-index_main = None
+INDEX_MAIN = None
+DOCS_ROOT = Path(__file__).parent.parent / 'docs'
```
dmontagu, line 16: Using this instead of just `'docs'` as the root in the tests makes it work when I run using the PyCharm pytest runner (which apparently doesn't run from the project root?).

I would very much appreciate if we could make this change as the pycharm pytest runner makes it _MUCH_ easier to browse the failures than just looking at console feedback.

tests/test_docs.py
```diff
@@ -73,14 +74,14 @@ def now(cls, *args, **kwargs):
 
 
 @pytest.mark.skipif(bool(skip_reason), reason=skip_reason or 'not skipping')
-@pytest.mark.parametrize('example', find_examples('docs', skip=sys.platform == 'win32'), ids=str)
+@pytest.mark.parametrize('example', find_examples(str(DOCS_ROOT), skip=sys.platform == 'win32'), ids=str)
```
samuelcolvin, line 77: we should update `find_examples` to take a `Path`, it's probably just an annotation change required.

samuelcolvin (approved):
LGTM

tests/test_docs.py
dmontagu: Created PR at https://github.com/pydantic/pytest-examples/pull/6, but I'll merge this with the str cast for now

PR #5312, opened 2023-03-30
Add deprecated v1 methods to BaseModel

dmontagu:
I'll note that I have not copied tests over from v1 for the newly-added methods. Let me know if you have opinions about if/how we do that.


pydantic/main.py
```diff
@@ -342,72 +338,6 @@ def model_construct(cls: type[Model], _fields_set: set[str] | None = None, **val
...
-            value = getattr(self, name, Undefined)
-            if value is not Undefined:
-                if deep:
-                    value = deepcopy(value)
-                _object_setattr(m, name, value)
-
-        return m
-
-    # @typing_extensions.deprecated('This method is now deprecated; use `model_copy` instead')
-    def copy(
```
dmontagu, line 364: I moved this and all other from-v1-deprecated methods to the bottom of the class in the same order they existed in v1.

I would also like to reorder the other (non-deprecated) methods of BaseModel, but will do that as a separate PR at an opportune time to reduce chances for merge conflicts etc.

pydantic/main.py
```diff
@@ -810,6 +564,383 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+                proto=proto,
+                content_type=content_type,
+                encoding=encoding,
+                allow_pickle=allow_pickle,
+                json_loads=cls.model_config['json_loads'],
+            )
+        except (ValueError, TypeError, UnicodeDecodeError) as e:
+            # TODO: this used to raise ValidationError([ErrorWrapper(e, loc=ROOT_KEY)], cls)
+            #   Do we want to make it possible to instantiate from python, just for deprecated functionality?
+            raise e
```
dmontagu, line 643: See TODO: comment above

pydantic/main.py
```diff
@@ -810,6 +564,383 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+            'and use `model_validate` instead.',
+            DeprecationWarning,
+        )
+        if not cls.model_config['from_attributes']:
+            raise PydanticUserError('You must set the config attribute `from_attributes=True` to use from_orm')
+        return cls.model_validate(obj)
+
+    @classmethod
+    def construct(cls: type[Model], _fields_set: set[str] | None = None, **values: Any) -> Model:
+        warnings.warn('The `construct` method is deprecated; use `model_construct` instead.', DeprecationWarning)
```
dmontagu, line 682: I have not copied over the v1 docstrings. Is that okay, or do we want to do that? I'm hoping the deprecation warning pointing the user at the other method is enough?

samuelcolvin: Agreed.

pydantic/_internal/_deprecated/_parse.py
```diff
@@ -0,0 +1,66 @@
...
+def load_str_bytes(
+    b: str | bytes,
+    *,
+    content_type: str = None,
+    encoding: str = 'utf8',
+    proto: Protocol = None,
+    allow_pickle: bool = False,
+    json_loads: Callable[[str], Any] = json.loads,
+) -> Any:
+    if proto is None and content_type:
```
samuelcolvin, line 24: These should have warning?

pydantic/main.py
```diff
@@ -810,6 +564,383 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+        *,
+        include: IncEx = None,
+        exclude: IncEx = None,
+        by_alias: bool = False,
+        exclude_unset: bool = False,
+        exclude_defaults: bool = False,
+        exclude_none: bool = False,
+        # TODO: What do we do about the following arguments?
+        #   Do they need to go on model_config now, and get used by the serializer?
+        encoder: typing.Callable[[Any], Any] | None = None,
```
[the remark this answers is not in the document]
samuelcolvin: I guess error if you use it. We need to point to docs on what to use instead.

pydantic/main.py
```diff
@@ -810,6 +564,383 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+        from .json import pydantic_encoder
+
+        return cls.model_config['json_dumps'](
+            cls.model_json_schema(by_alias=by_alias, ref_template=ref_template),
+            default=pydantic_encoder,
+            **dumps_kwargs,
+        )
+
+    @classmethod
+    def validate(cls: type[Model], value: Any) -> Model:
```
samuelcolvin, line 747: This was internal and wasn't documented, not sure we need it anymore.

pydantic/main.py
```diff
@@ -810,6 +564,383 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+
+    @classmethod
+    def update_forward_refs(cls, **localns: Any) -> None:
+        warnings.warn(
+            'The `update_forward_refs` method is deprecated; use `model_rebuild` instead.', DeprecationWarning
+        )
+        cls.model_rebuild(force=True, types_namespace=localns)
+
+    # Deprecated: This private method is only used for `BaseModel.copy`, which is deprecated
+    def _iter(
```
samuelcolvin, line 759: Can we move all this logic into the deprecated module as standalone functions?

dmontagu: good idea, I'll do that.

pydantic/_internal/_deprecated/_parse.py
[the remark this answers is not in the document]
dmontagu: good point, I'll just put it in a top-level `deprecated` module. Should it be `deprecated` or `_deprecated`?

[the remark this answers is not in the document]
samuelcolvin: I guess just `deprecated`, it's not private.

pydantic/main.py
samuelcolvin: My I think we just raise an error, there's no obvious replacement for it.

pydantic/_internal/_deprecated/_parse.py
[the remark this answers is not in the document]
dmontagu: I went with `deprecated` for now

pydantic/main.py
dmontagu: Moved these into deprecated/copy_internals.py, and retained the methods on the class with deprecation warnings. Let me know if you also want the methods themselves to have deprecation warnings. (If we do that, I think it may mean that calls to `Model.copy` trigger a bunch of warnings though)

pydantic/main.py
dmontagu: Looking at the definition from v1, it seems `model_validate` does the same thing for the most part: https://github.com/pydantic/pydantic/blob/11f7f9019016659e81c96481ffe5c42637489cf3/pydantic/main.py#L710-L719 

I realize there is the logic at the top related to what to do with subclass instances but that too seems handled by `model_validate` now with the recent changes you made in pydantic core. Is it really different enough to be worth removing instead of just adding a deprecation warning? I could add a note in the warning that the behavior may now be slightly different.

If you really think it's better to remove it, I'd suggest raising an error that says to replace it with model_validate, but I'm not sure what type of error to raise

samuelcolvin: Ok, let's go with a warning.

samuelcolvin:
conflicts and comments, please update.

pydantic/main.py
```diff
@@ -810,6 +564,223 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+        cls: type[Model],
+        b: str | bytes,
+        *,
+        content_type: str = None,
+        encoding: str = 'utf8',
+        proto: _deprecated_parse.Protocol = None,
+        allow_pickle: bool = False,
+    ) -> Model:
+        warnings.warn(
+            'The `parse_raw` method is deprecated; load the data and use `model_validate` instead.', DeprecationWarning
```
samuelcolvin, line 629: need to say

> use `model_validate_json` if your input is JSON, otherwise parse the data in your own code, and validate using `model_validate`

pydantic/main.py
```diff
@@ -810,6 +564,223 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+        cls: type[Model],
+        path: str | Path,
+        *,
+        content_type: str = None,
+        encoding: str = 'utf8',
+        proto: _deprecated_parse.Protocol = None,
+        allow_pickle: bool = False,
+    ) -> Model:
+        warnings.warn(
+            'The `parse_file` method is deprecated; load the data and use `model_validate` instead.', DeprecationWarning
```
samuelcolvin, line 657:
```suggestion
            'The `parse_file` method is deprecated; load the data and use `model_validate` instead.', DeprecationWarning
```
as above, recommend `model_validate_json`.

pydantic/main.py
```diff
@@ -810,6 +564,223 @@ def model_parametrized_name(cls, params: tuple[type[Any], ...]) -> str:
...
+            encoding=encoding,
+            allow_pickle=allow_pickle,
+            json_loads=cls.model_config['json_loads'],
+        )
+        return cls.parse_obj(obj)
+
+    @classmethod
+    def from_orm(cls: type[Model], obj: Any) -> Model:
+        warnings.warn(
+            'The `from_orm` method is deprecated; set model_config["from_attributes"]=True '
```
samuelcolvin, line 672:
```suggestion
            'The `from_orm` method is deprecated; set `model_config["from_attributes"]=True` '
```

samuelcolvin: I guess we can remove the note about `model_config["from_attributes"]=True` if they already have it set and we're not raising an error.

pydantic/deprecated/copy_internals.py
samuelcolvin, line 1: Might be worth removing `pydantic/deprecated` from coverage completely.

samuelcolvin (commented):
otherwise LGTM, either now or soon I think it's worth adding at least one test for each function to a new `tests/test_depreciated.py` module. Might move `test_orm_mode.py` tests into there too.

pydantic/main.py
samuelcolvin: https://github.com/pydantic/pydantic-core/pull/510

pydantic/main.py
samuelcolvin: on reflection, I think this is fine.

PR #5327, opened 2023-03-31
Config migration warning

dmontagu:
This adds migration warnings for configs that make use of config fields that have been removed or renamed in v2.

### DIFF
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -91,6 +91,39 @@ if TYPE_CHECKING:
 else:
 
     class ConfigDict(dict):
+        _V2_REMOVED_KEYS = {
+            'allow_mutation',
+            'error_msg_templates',
+            'fields',
+            'getter_dict',
+            'schema_extra',
+            'smart_union',
+            'underscore_attrs_are_private',
+        }
+        _V2_RENAMED_KEYS = {
+            'allow_population_by_field_name': 'populate_by_name',
+            'anystr_lower': 'str_to_lower',
+            'anystr_strip_whitespace': 'str_strip_whitespace',
+            'anystr_upper': 'str_to_upper',
+            'keep_untouched': 'ignored_types',
+            'max_anystr_length': 'str_max_length',
+            'min_anystr_length': 'str_min_length',
+            'orm_mode': 'from_attributes',
+            'validate_all': 'validate_default',
+        }
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+
+            deprecated_removed_keys = ConfigDict._V2_REMOVED_KEYS & self.keys()
+            deprecated_renamed_keys = ConfigDict._V2_RENAMED_KEYS.keys() & self.keys()
+            if deprecated_removed_keys or deprecated_renamed_keys:
+                renamings = {k: self._V2_RENAMED_KEYS[k] for k in sorted(deprecated_renamed_keys)}
+                renamed_bullets = [f'* {k!r} has been renamed to {v!r}' for k, v in renamings.items()]
+                removed_bullets = [f'* {k!r} has been removed' for k in sorted(deprecated_removed_keys)]
+                message = '\n'.join(['Valid config keys have changed in V2:'] + renamed_bullets + removed_bullets)
+                warnings.warn(message, UserWarning)
+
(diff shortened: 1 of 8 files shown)


pydantic/config.py
```diff
@@ -182,6 +215,7 @@ def get_config(config: ConfigDict | dict[str, Any] | type[Any] | None) -> Config
         )
         config_dict = {k: getattr(config, k) for k in dir(config) if not k.startswith('__')}
 
+    prepare_config(config_dict, error_label or 'ConfigDict')
```
dmontagu, line 218: I added this due to some inconsistency in when invalid values of `extra` would result in more useful error messages.

Before, you would get the more explicit error for things that created models, but not for dataclasses. In principle I think type hints should prevent issues, but I think this is an improvement because using `"ignore"` instead of `Extra.ignore` is a pretty easy mistake to make considering it was okay in v1, and if you do that, you get a confusing (in my opinion) error saying something like `type 'str' has no attribute 'value'` (since it tries to grab the .value off the enum).

pydantic/config.py
```diff
@@ -233,9 +267,10 @@ def build_config(
...
-def prepare_config(config: ConfigDict, cls_name: str) -> None:
-    if not isinstance(config['extra'], Extra):
+def prepare_config(config: ConfigDict | dict[str, Any], error_label: str) -> None:
+    extra = config.get('extra')
+    if extra is not None and not isinstance(extra, Extra):
         try:
             config['extra'] = Extra(config['extra'])
         except ValueError:
-            raise ValueError(f'"{cls_name}": {config["extra"]} is not a valid value for "extra"')
+            raise ValueError(f'{error_label!r}: {config["extra"]!r} is not a valid value for config[{"extra"!r}]')
```
dmontagu, line 276: I felt this name was more appropriate now that it can come from non-classname sources.

tests/test_config.py
```diff
@@ -387,3 +400,116 @@ class Config(BaseConfig):
...
+            pass
+
+    with pytest.warns(UserWarning, match=re.escape(warning_message)):
+
+        @validate_arguments(config=config_dict)
+        def my_function():
+            pass
+
+
+def test_invalid_extra():
```
dmontagu, line 466: I added this test to make sure you got the same error in every codepath where you can specify a ConfigDict

tests/test_config.py
```diff
@@ -387,3 +400,116 @@ class Config(BaseConfig):
...
+
+        my_partial_function = partial(my_wrapped_function)
+        my_partial_function.__annotations__ = my_wrapped_function.__annotations__
+        validate_arguments(config=config_dict)(my_partial_function)
+
+    with pytest.raises(ValueError, match=re.escape(error_message.format(label='ConfigDict'))):
+        get_config(config_dict)
+
+
+def test_invalid_config_keys():
```
dmontagu, line 504: There is a test for this in `test_decorator` but it felt appropriate to also include it here as I discovered it while working on the items above. I can drop it though if preferred.

tests/test_private_attributes.py
```diff
@@ -79,8 +79,6 @@ class Model(BaseModel):
 
         _foo: str
 
-        model_config = ConfigDict(underscore_attrs_are_private=True)
```
dmontagu, line 82: This causes a user warning now, so I removed it. However, currently, the message just says it was removed, not providing the extra detail that "v2 behaves as though it was set to True everywhere". If we want to provide this level of detail I could modify the `ConfigDict.__init__` to have more specific error messages for each individual field.

pydantic/decorator.py
```diff
@@ -210,11 +210,11 @@ def execute(self, m: BaseModel) -> Any:
     def create_model(self, fields: Dict[str, Any], takes_args: bool, takes_kwargs: bool, config: 'ConfigType') -> None:
         pos_args = len(self.arg_mapping)
 
-        config_dict = get_config(config)
+        config_dict = get_config(config, getattr(self.raw_function, '__name__', 'validate_arguments'))
```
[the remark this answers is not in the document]
samuelcolvin: fine by me.

samuelcolvin (commented):
otherwise LGTM, also conflicts. I might fix it soon, otherwise please update.

PR #5331, opened 2023-04-01
rename `serializer` to `field_serializer`

samuelcolvin:
renaming and a few other things.

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -575,8 +575,10 @@ def make_v1_generic_root_validator(
     return _wrapper2
 
 
-PlainSerializerWithoutInfo = Callable[[Any], Any]
-WrapSerializerWithoutInfo = Callable[[Any, SerializerFunctionWrapHandler], Any]
+GenericPlainSerializerFunctionWithoutInfo = Callable[[Any], Any]
+FieldPlainSerializerFunctionWithoutInfo = Callable[[Any, Any], Any]
+FieldWrapSerializerFunctionWithoutInfo = Callable[[Any, Any, SerializerFunctionWrapHandler], Any]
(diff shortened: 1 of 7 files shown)


pydantic/_internal/_decorators.py
```diff
@@ -649,6 +651,7 @@ def make_generic_field_serializer(
         if n_positional == 1:
             func1 = cast(PlainSerializerWithoutInfo, serializer)
 
+            @wraps(func1)
```
samuelcolvin, line 654: These function names come through in error messages, hence the `wraps`, we could also use clearer names for `_wrap1` etc.

tests/test_serialize.py
```diff
@@ -204,8 +210,31 @@ def test_serialize_ignore_info_wrap():
...
+
+        @field_serializer('x', json_return_type='str')
+        def customise_x_serialisation(self, v, info):
+            return f'{info.mode}:{v:,}'
+
+    assert MyModel(x=1234).model_dump() == {'x': 'python:1,234'}
+    assert MyModel(x=1234).model_dump(mode='foobar') == {'x': 'foobar:1,234'}
+
+
+def test_serialize_decorator_self_no_info():
```
samuelcolvin, line 232: This test is failing, @adriangb do you have a quick solution? I looked but couldn't solve it quickly.

Otherwise we can mark it as `xfail` and fix it later.

samuelcolvin (commented):
please review.

pydantic/_internal/_decorators.py
adriangb: FYI the reason they are called `_wrap1`  and not all just `wrap` is that MyPy really doesn't like it if you re-define the function, even if they are never defined in the same scope

samuelcolvin:
Auto merging.

PR #5335, opened 2023-04-01
Config cleanup

samuelcolvin:
changes:
* Remove more stuff from config that is no longer used
* move `json.py` in to `deprecated`
* use `to_python_jsonable` to generate json_schema

Causes the following changes:
* Decimal defaults in JSON Schema are now strings (lots of old issues with people (rightly) depanding that decimals are encoded to strings not floats)
* named tuples in JSON schema are now lists, ref #2711

selected reviewer: @dmontagu

PR #5338, opened 2023-04-02
Massively simplify JSON schema generation

dmontagu:
* Completely eliminates the JsonSchemaMetadata struct, relying solely on the CoreSchema metadata, and reducing the smattering of ways the JSON schema could be updated.
* Replaces `model_json_schema_metadata` with `model_modify_json_schema`
    * Perhaps this should be renamed to `__pydantic_modify_json_schema__`, perhaps not
* Re-adds non-`AnalyzedType` functions for computing JSON schema, called `model_json_schema` and `models_json_schema`. These are added into the `json_schema.py` but can be moved elsewhere, or reverted to being `AnalyzedType` methods, but I think having a free-standing function is simpler for docs, etc.

### DIFF
--- a/pydantic/_internal/_core_metadata.py
+++ b/pydantic/_internal/_core_metadata.py
@@ -2,21 +2,23 @@ from __future__ import annotations as _annotations
 
 import inspect
 import typing
+import warnings
 from typing import Any
 
 import typing_extensions
 from pydantic_core import CoreSchema, core_schema
 
-from ._typing_extra import EllipsisType
-
 if typing.TYPE_CHECKING:
-    from ..json_schema import JsonSchemaMetadata, JsonSchemaValue
+    from ..json_schema import JsonSchemaValue
 
 
 # CS is shorthand for Core Schema
 # JS is shorthand for JSON Schema
-_UPDATE_CS_FUNCTION_FIELD = 'pydantic_update_core_schema_function'
-_JS_METADATA_FIELD = 'pydantic_json_schema_metadata'
+_CS_UPDATE_FUNCTION_FIELD = 'pydantic_cs_update_function'
+_JS_OVERRIDE_FIELD = 'pydantic_js_override'
+_JS_CS_OVERRIDE_FIELD = 'pydantic_js_cs_override'
+_JS_MODIFY_FUNCTION_FIELD = 'pydantic_js_modify_function'
+_JS_PREFER_POSITIONAL_ARGUMENTS_FIELD = 'pydantic_js_prefer_positional_arguments'
 
 _CoreMetadata = typing.Dict[str, Any]
 
@@ -46,7 +48,7 @@ class CoreMetadataHandler:
             raise TypeError(f'CoreSchema metadata should be a dict; got {metadata!r}.')
 
     @property
-    def metadata(self) -> _CoreMetadata:
+    def _metadata(self) -> _CoreMetadata:
         """
         Retrieves the metadata dict off the schema, initializing it to a dict if it is None
         and raises an error if it is not a dict.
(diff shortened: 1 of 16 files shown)


dmontagu:
please review

I'll be happy to resolve any merge conflicts with #5335 but I don't think they will be significant.

pydantic/_internal/_core_metadata.py
```diff
@@ -2,21 +2,23 @@
...
 if typing.TYPE_CHECKING:
-    from ..json_schema import JsonSchemaMetadata, JsonSchemaValue
+    from ..json_schema import JsonSchemaValue
 
 
 # CS is shorthand for Core Schema
 # JS is shorthand for JSON Schema
-_UPDATE_CS_FUNCTION_FIELD = 'pydantic_update_core_schema_function'
-_JS_METADATA_FIELD = 'pydantic_json_schema_metadata'
+_CS_UPDATE_FUNCTION_FIELD = 'pydantic_cs_update_function'
```
samuelcolvin, line 17: is there no way for this to be a dataclass or typeddict?

samuelcolvin: a comment on each value/attribute and its meaning would be useful here.

pydantic/_internal/_std_types_schema.py
```diff
@@ -69,33 +70,28 @@ def to_enum(__input_value: Any, _: core_schema.ValidationInfo) -> Enum:
...
     literal_schema = core_schema.literal_schema(
         [m.value for m in enum_type.__members__.values()],
-        ref=enum_ref,
     )
-    js_metadata = JsonSchemaMetadata(
-        core_schema_override=literal_schema.copy(),
-        source_class=enum_type,
-        title=enum_type.__name__,
-        description=inspect.cleandoc(enum_type.__doc__ or 'An enumeration.'),
+    updates = {'title': enum_type.__name__, 'description': inspect.cleandoc(enum_type.__doc__ or 'An enumeration.')}
```
samuelcolvin, line 77: "An enumeration." seems like perhaps the most confusing description imaginable for a set of choices in JSON schema - should we just remove?

pydantic/json_schema.py
```diff
@@ -1210,6 +1124,80 @@ def render_warning_message(self, kind: JsonSchemaWarningKind, detail: str) -> st
...
+        json_schema['$defs'] = definitions
+    if title:
+        json_schema['title'] = title
+    if description:
+        json_schema['description'] = description
+
+    return json_schema
+
+
+# TODO: The JSON schema cache has been there from the beginning (https://github.com/pydantic/pydantic/pull/190 —
```
samuelcolvin, line 1155: let's move this into the PR body, I think it makes more sense there (at least in long form) than here.

samuelcolvin (commented):
otherwise, I think looking good.

samuelcolvin:
please update conflicts.

pydantic/_internal/_core_metadata.py
dmontagu: I'll convert it to a TypedDict for now

pydantic/_internal/_std_types_schema.py
dmontagu: I'll see if I can see any reason why it was added where it was added. I don't see much reason to default to having a description, we don't do that for models. If there's nothing obvious I think we can remove it.

pydantic/json_schema.py
dmontagu: I can move the comment to the PR body, but let me know if you have thoughts about what you want to do with the cache.

pydantic/_internal/_std_types_schema.py
dmontagu: Investigated — this is actually set as the docstring of the enum via `enum.EnumMeta.__new__`. I think the closest thing we could do is just explicitly look for this string and if it's exactly equal to this, remove it. I'm inclined not to change it for now though, since this has been the behavior all along.

PR #5340, opened 2023-04-03
[ALPHA] Update tests to show how to define root models in v2

dmontagu:
Demonstrate how to migrate root models

Selected Reviewer: @adriangb


tests/test_parse.py
```diff
@@ -53,79 +53,174 @@ class MyModel(BaseModel):
...
-    #   Probably also want to re-add some of the __root__ tests from v1 that test other behaviors,
-    #   such as serialization and JSON schema.
     class MyModel(BaseModel):
-        __root__: str
-
+        root: str
+
+        # Note that the following three definitions require no changes across all __root__ models
+        # I couldn't see a nice way to create a decorator that reduces the boilerplate,
+        # but if we want to discourage this pattern, perhaps that's okay?
```
dmontagu, line 62: Adding this comment to draw attention to this point

tests/test_parse.py
```diff
@@ -53,79 +53,174 @@ class MyModel(BaseModel):
...
     m = MyModel.model_validate('a')
-    assert m.model_dump() == {'__root__': 'a'}
-    assert m.__root__ == 'a'
+    assert m.root == 'a'
+
+    # Serialization
+    # TODO: Possible concern — `model_dump` is annotated as returning dict[str, Any] — is that okay, given
+    #   model_serializer could change that? Should we try to reflect it in the mypy plugin?
+    assert m.model_dump() == {'root': 'a'}
+    assert m.model_dump_json() == '"a"'
```
dmontagu, line 86: Adding this comment to draw attention to the TODO comment above

dmontagu:
please review

I have set the base branch of this to the branch from #5338 as I needed the changes there to handle the JSON schema stuff. We should merge that first, then can merge this.

samuelcolvin (commented):
otherwise LGTM.

Please update.

tests/test_deprecated.py
```diff
@@ -75,24 +74,55 @@ class Pokemon(BaseModel):
         jp_name: str
 
     class PokemonList(BaseModel):
-        __root__: List[Pokemon]
+        root: List[Pokemon]
```
adriangb, line 77: Just to clarify, there's technically no reason why this couldn't still be called `__root__`, it's just that there's no reason it has to be called that anymore and we are in fact explicitly prohibiting `__root__` because we want to be able to give users an error so that they don't think that `__root__` models will just work with no changes?

tests/test_deprecated.py
```diff
@@ -75,24 +74,55 @@ class Pokemon(BaseModel):
...
+
+        @model_serializer
+        def _serialize(self, info):
+            if info.mode == 'json':
+                return self.root
+            return self.__dict__
+
+        @classmethod
+        def model_modify_json_schema(cls, json_schema):
+            return json_schema['properties']['root']
```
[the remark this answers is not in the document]
dmontagu: They do, but they'll have to rename the attribute anyway as it will raise an error when they try to create the model if it has a field called `__root__`.

One thing we could do to ease the transition would be to actually allow that as a field name (the only allowed dunder field name) and emit a DeprecationWarning. (We could add the appropriate model_serializer, model_modify_json_schema, and root validator). However, it seems to me we probably shouldn't do this? (I'm open to being convinced though.)

Considering this is just a "reference" migration, I don't know that it makes sense to add `__root__` as a property since, as I said, they'd need to rename the field anyway; users can do this if for some reason they are okay changing the field name but don't want to change the access pattern. (In principle, in that case, they may also want/need to add a setter for that property too.)

tests/test_deprecated.py
```diff
@@ -75,24 +74,55 @@ class Pokemon(BaseModel):
...
+
+        @root_validator(pre=True)
+        @classmethod
+        def populate_root(cls, values):
+            return {'root': values}
+
+        @model_serializer
+        def _serialize(self, info):
+            if info.mode == 'json':
+                return self.root
```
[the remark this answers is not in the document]
dmontagu: From v1:
```from pydantic import BaseModel

class MyModel(BaseModel):
    __root__: str

m = MyModel.parse_obj('a')
print(dict(m))
# > {'__root__': 'a'}
print(m.dict())
# > {'__root__': 'a'}
print(m.json())
# > "a"
```
I'll note that in v1, we explicitly added these lines to the `.json()` method, lol:
```python
...
        if self.__custom_root_type__:
            data = data[ROOT_KEY]
...
```
Good catch on the include/exclude though, that does still get handled right in v1 .json(); I'll update it to use your definition which will do that right.

tests/test_parse.py
dmontagu: Do you want to do that in this PR? I feel like that's may end up being unfortunate for ergonomics, but I guess safer than having a type error. And definitely better than making it generic over the dump type.

tests/test_deprecated.py
samuelcolvin: I'm not sure, maybe dunder/private attributes can't be used as field names - have you tried?

tests/test_deprecated.py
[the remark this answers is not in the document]
samuelcolvin: okay, let's leave it as-is for now.

tests/test_deprecated.py
samuelcolvin: see above `__root__` doesn't currently work.

dmontagu: We are ignoring all dunder annotations when building fields out of the namespace; they essentially become class variables. (I'll note this is the same as the behavior in v1 for everything _except_ `__root__`.)

You are right that we are raising a special error for `__root__` to make sure users realize changes will be required.

We could make `__root__` work as a field name, and could even add these properties to get things as close to "just working" as possible. I'm not sure if that's the best idea given 1, Samuel has suggested we discourage this pattern going forward, and 2, if we try to make it "just work", people may expect us to fix any incompatibility bugs.

tests/test_deprecated.py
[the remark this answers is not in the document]
dmontagu: I've now updated it to use your wrap model_serializer everywhere.

dmontagu:
I've now responded to all the comments above. I think there are still some things to do but I'll say "please review" to get another round of feedback on the review comments

dmontagu:
the failing CI seems to be due to the enum descriptions in the JSON schema; this should get fixed once #5338 gets merged.

tests/test_deprecated.py
adriangb: That makes sense to me David and I agree. I do think the migration path should explain this reasoning, otherwise users might think that we arbitrarily are changing the field name in the guide when it's not needed and/or be surprised at it not working with `__root__`

adriangb (approved):

tests/test_parse.py
samuelcolvin: let's leave it, users can always change the signature.

PR #5345, opened 2023-04-03
Add strict and context arguments to validation functions

adriangb:
TODO: docstrings

### DIFF
--- a/pydantic/analyzed_type.py
+++ b/pydantic/analyzed_type.py
@@ -92,6 +92,8 @@ def _translate_config(config: ConfigDict) -> core_schema.CoreConfig:
         ),
         str_min_length=config['str_min_length'] if 'str_min_length' in config else unset,
     )
+    for k in [k for k in core_config if core_config[k] is unset]:
+        core_config.pop(k)
     return CoreConfig(**core_config)  # type: ignore[misc]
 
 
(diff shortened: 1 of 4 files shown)


tests/test_validator.py
```diff
@@ -92,3 +95,85 @@ def test_type_alias():
...
+    class Model(BaseModel):
+        x: int
+        y: str
+
+        model_config = ConfigDict(strict=True, title='FooModel')
+
+    analyzed = AnalyzedType(Model, config=ConfigDict(strict=False, str_max_length=20))
+
+    # strict=False gets applied to the outer Model but not to the inner typeddict validator
+    # so we're allowed to validate a dict but `x` still must be an int
```
adriangb, line 166: This is an interesting artifact of how we validate models by wrapping a typeddict schema. I would not expect this as a user. I would expect that `model_config` has the same meaning as the `config` argument to `AnalyzedType`.

pydantic/analyzed_type.py
```diff
@@ -130,26 +132,28 @@ def __init__(self, __type: Any, *, config: ConfigDict | None = None, _parent_dep
...
             serializer = SchemaSerializer(core_schema, core_config)
 
         self.core_schema = core_schema
         self.validator = validator
         self.serializer = serializer
 
-    def validate_python(self, __object: Any) -> T:
-        return self.validator.validate_python(__object)
+    def validate_python(self, __object: Any, *, strict: bool | None = None, context: dict[str, Any] | None = None) -> T:
+        return self.validator.validate_python(__object, strict=strict, context=context)
```
samuelcolvin, line 151: need to check whether `context=None` get's passed through as a value, e.g. is different to omitting context.

samuelcolvin (commented):
can you add a couple of tests for `BaseModel` with context, just to be sure it's actually working.

Otherwise LGTM.

adriangb:
Oh yeah I'll do that

pydantic/analyzed_type.py
adriangb: Should they be different? I thought it was None by default

adriangb: I confirmed that it's the same as far as the Python side is concerned

PR #5354, opened 2023-04-03
Get frozen fields working

dmontagu:
I don't love this approach, but it works.

I'd be happy to rework this with another approach, the main benefit to this approach is that it fully re-uses the existing annotation-unwrapping logic.

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -211,8 +211,6 @@ class WalkAndApply:
         return self._walk(schema)
 
     def _walk(self, schema: core_schema.CoreSchema) -> core_schema.CoreSchema:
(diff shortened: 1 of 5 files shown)


dmontagu:
please review

adriangb:
@samuelcolvin David and I discussed a bit offline. We both feel that this implementation works, but isn't ideal because the existing codebase isn't really designed to support this. We're not sure what the right path forward is but I am assigning this to you so that you can decide if you want to merge this as-is or think about a large scale rework in this area.

samuelcolvin:
let's think about it rather than doing anything in a hurry.

dmontagu:
Closing in favor of #5354

PR #5355, opened 2023-04-03
Convert schema_of to use AnalyzedType

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -42,14 +42,14 @@ def schema_of(
         ref_template=ref_template,
     )
     if title is not None:
-        warnings.warn(
-            'Passing a callable for the `title` parameter is deprecated and no longer supported',
-            DeprecationWarning,
-            stacklevel=2,
-        )
         if isinstance(title, str):
             res['title'] = title
         else:
+            warnings.warn(
+                'Passing a callable for the `title` parameter is deprecated and no longer supported',
+                DeprecationWarning,
+                stacklevel=2,
+            )
             res['title'] = title(type_)
     return res
 
--- a/pydantic/tools.py
+++ b/pydantic/tools.py
@@ -4,6 +4,8 @@ import json
 import warnings
 from typing import Any, Callable, Type, TypeVar, Union
 
+from pydantic.config import ConfigDict
+
 from . import AnalyzedType
 
 __all__ = 'parse_obj_as', 'schema_of', 'schema_json_of'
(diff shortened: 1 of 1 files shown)


adriangb:
@samuelcolvin do we plan to actually deprecate these functions? It seems sensible to have wrappers around `AnalyzedType` that build an `AnalyzedType` and call the corresponding method. So maybe these can become the solution to #5213 ?

dmontagu:
The main concern I have if we don't deprecate them is that, in v1, they would cache the parsing models (i.e., the thing we have replaced with AnalyzedType). If we don't port the caching for the `AnalyzedType`-wrapper-versions of these functions, it could result in significant performance regressions.

In principle this should be less of an issue for JSON schema, but I'll note that this is the one place where the `__schema_cache__` _is_ used in v1 — the `schema_of` function ultimately calls the `model.schema()` method, which does do JSON schema caching. So if someone was calling `schema_of` to respond to an API endpoint, they would be getting the caching behavior that is now gone.

(And to be clear, the caching behavior is not present on the `main` branch, so I think it's orthogonal to the use of AnalyzedType in this PR, but I do think it's relevant to the question of deprecation.)

adriangb:
Gotcha. Looks like we have to pick between definitely breaking a larger fraction of users in an understandable way or maybe breaking a small section of users in an unknown way.

adriangb:
The test failure seems real, I'm not sure how to immediately resolve it. It seems like the `title` in `ConfigDict` never makes it into schema generation

dmontagu:
easy way to resolve it: just set the title on the final result before returning it

adriangb:
please review

pydantic/tools.py
```diff
@@ -35,15 +36,20 @@ def schema_of(
...
-    json_schema = json_schema_generator.generate(core_schema)
-    if title is not None:
-        json_schema['title'] = title
-    return json_schema
+    if not isinstance(title, (str, type(None))):  # pragma: no cover
+        warnings.warn(
+            'The type_name parameter is deprecated. parse_obj_as no longer creates temporary models',
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
dmontagu, line 44: This logic doesn't seem right to me — it's checking `title` to say something about `parse_obj_as` and temporary models.

I will also add that I think the `title` value is being mishandled here — I think we need to add something like `if callable(title): title = title()`

adriangb, line 44:
```suggestion
        warnings.warn(
            'Passing a callable for the `title` parameter is deprecated and no longer supported',
            DeprecationWarning,
            stacklevel=2,
        )
```

adriangb: The signature is `Callable[[Type[Any]], str]`. So does it expect the type itself? If so, what's the point of a callback if the type is also a parameter?

adriangb:
```suggestion
        title = title(type_)
```

Is this really what we need? In abstract it seems like a pointless API, but if it's there for historic reasons it maybe makes sense...

dmontagu: > If so, what's the point of a callback if the type is also a parameter?

I think at some point we thought it was a cleaner API to do it this way (I think I was the person who added the NameFactory, for `parse_obj_as` (which was originally `parse_as_type`). Probably it never needed to exist.

dmontagu: Looks like it was my suggestion, and just came from analogy with the name generation stuff for generic models (which is done by overriding a method).  https://github.com/pydantic/pydantic/pull/934#discussion_r341298903 for reference.

I think you are right that we never needed to support name factories here, and it's a good idea to deprecate it.

adriangb: Thanks for digging. So the deprecation we have now should be okay?

dmontagu: Your suggestion above with `title(type_)` is correct, if we just emit a deprecation warning rather than raising an error, I think we should call.

dmontagu:
Looks good to me.

PR #5356, opened 2023-04-03
Error URLs

samuelcolvin:
Still need to do the redirects and write the docs, but this is almost ready to go.


samuelcolvin:
I can fix conflicts, but @dmontagu, @adriangb & @tpdorsey please can you review and let me know if you're happy with this.

The idea is that we have a `Literal` of codes so you can't use a code without adding it to the literal, then we check the `args` of the literal are all included in the help docs, and thereby guarantee that all pydantic errors called with a code are documented.

tpdorsey:
Conceptually this looks great. I didn't see any obvious problems. Would it be worth doing a pass on error messages for capitalization consistency?

samuelcolvin:
let's get it merged first, then work on the docs.

pydantic/main.py
```diff
@@ -920,9 +926,10 @@ def create_model(
                 f_annotation, f_value = f_def
             except ValueError as e:
                 raise PydanticUserError(
-                    'field definitions should either be a tuple of (<type>, <default>) or just a '
+                    'Field definitions should either be a tuple of (<type>, <default>) or just a '
                     'default value, unfortunately this means tuples as '
-                    'default values are not allowed'
+                    'default values are not allowed',
```
dmontagu, line 931: Random thing that maybe we can fix now — we never actually allow non-tuple field definitions now (because we don't allow non-annotated fields). So this error message can (and probably should) be updated. If you don't want to deal with changing that in this PR maybe we can still just insert a TODO comment into this PR to fix the message?

pydantic/errors.py
```diff
@@ -2,22 +2,64 @@
...
 __all__ = (
     'PydanticUserError',
-    'PydanticSchemaGenerationError',
     'PydanticUndefinedAnnotation',
+    'PydanticSchemaGenerationError',
     'PydanticInvalidForJsonSchema',
 )
 
+# TODO set up a cloudflare worker to redirect to the correct page
+DEV_ERROR_DOCS_URL = f'https://e.pydantic.dev/{VERSION}/d/'
```
dmontagu, line 17: I'm assuming in `e.pydantic.dev` the `e` is for errors? Is this better than `errors.pydantic.dev` or `pydantic.dev/errors` or similar? Also what's the `/d/` for? 

If you feel strongly about the `e.pydantic.dev` being right, it's not very important to me, just curious.

pydantic/_internal/_decorators.py
```diff
@@ -636,43 +637,6 @@ def _wrapper2(
...
-serializer('x')(ser_x)
-
-For all of these, you can also choose to omit the `info` argument, for example:
-
-@serializer('x')
-def ser_x(self, value: Any): ...
-
-@serializer('x', mode='wrap')
-def ser_x(self, value: Any, handler: pydantic.SerializerFunctionWrapHandler): ...
-"""
```
dmontagu, line 673: Probably a good idea to store this (or a reference to it anyway) somewhere for inclusion in the 'field-serializer-signature' errors page (I guess with some modifications).

dmontagu: Nevermind I see how it works, and you already did it.

pydantic/errors.py
samuelcolvin: The `d` for for developer errors - to differentiate from validation errors which I would guess should get the prefix `/v/`.

The `e.pydantic.dev` was to keep the url in the error message short, but we could use `errors`.

More generally, the rationale of using a custom subdomain and building a service to deal with these was three fold:
1. (most importantly) to give us flexibility about what we do with this error documentation in the future - once a version of pydantic is deployed with this feature, we can't update the URL so we want a way to direct people wherever we want without it constraining how we host the rest of the docs.
2. to keep the URL short and clear - I guess `errors.` improves clarity significantly with minimal increase in length.
3. to allow us to log analytics on this in future if we want to

pydantic/main.py
samuelcolvin: done.

PR #5343, opened 2023-04-03
[ALPHA] V2 Alpha Blog Post

samuelcolvin:
Replaces #5328

skip change file check

PR #5357, opened 2023-04-03
Update note in migration guide about `__root__` models

dmontagu:
skip change file check

Merges into https://github.com/pydantic/pydantic/pull/5343


samuelcolvin (approved):
LGTM

Issue #5351, opened 2023-04-03
V2 Generic Improvements


samuelcolvin:
* [ ] generic dataclasses
* [x] a `TypedDict` to gather [these](https://github.com/pydantic/pydantic/blob/06b8bdc6f07f12a15f5f0ba79e38f4a509803328/pydantic/main.py#L183-L188)
* [x] as @dmontagu mentioned, generic tests looks slow - is this indicative of generics being slow? Can we sort it by caching core-schema if there are no generic arguments?

dmontagu:
Generic tests benchmarking:

This script was written based on `tests.test_generics.test_generics_work_with_many_parametrized_base_models`:
```python
import time
from typing import TypeVar, Generic
from pydantic import BaseModel


def run(n_models: int):
    t0 = time.time()

    T = TypeVar('T')
    C = TypeVar('C')

    class A(BaseModel, Generic[T, C]):
        x: T
        y: C

    class B(A[int, C], BaseModel, Generic[C]):
        pass

    models = []
    for i in range(n_models):
        class M(BaseModel):
            pass

        M.__name__ = f'M{i}'
        models.append(M)

    generics = []
    for m in models:
        Working = B[m]
        generics.append(Working)

    t1 = time.time()
    print(f"{t1 - t0:.2f} seconds elapsed")


run(3000)
```
On my computer, on commit `af73124` (immediately before the schema building refactor), the runtime is ~1.38 seconds.

On the current main branch, I get ~2.40 seconds.

(I will do the same check for validation time shortly, though I suspect the impact is smaller.)

dmontagu:
Benchmarking script for validating with generic models:
```python
import time
from pathlib import Path
from typing import Generic, TypeVar, Union

from pydantic import BaseModel

T = TypeVar('T')

# Based on tests.test_generics.test_generic_recursive_models

class GenericModel1(BaseModel, Generic[T]):
    ref: 'GenericModel2[T]'

    model_config = dict(undefined_types_warning=False)


class GenericModel2(BaseModel, Generic[T]):
    ref: Union[T, GenericModel1[T]]

    model_config = dict(undefined_types_warning=False)


GenericModel1.model_rebuild()
GenericModelStr = GenericModel1[str]

class RecursiveModelStr1(BaseModel):
    ref: 'RecursiveModelStr2'

    model_config = dict(undefined_types_warning=False)


class RecursiveModelStr2(BaseModel):
    ref: Union[str, RecursiveModelStr1]

[...]

PR #5305, opened 2023-03-29
Model schema changes 3

samuelcolvin:
Following from #5302 and #5285.

With this "only" 15 tests are failing:

```
                                                          Summary of Failures                                                           
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  File                           ┃  Function                       ┃  Function Line  ┃  Error Line  ┃  Error                          ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│  tests/test_discriminated_uni…  │  test_nested_optional_unions    │  598            │  611         │  ValueError                     │
│  tests/test_generics.py         │  test_generic_recursive_models  │  1418           │  1442        │  Failed                         │
│  tests/test_generics.py         │  test_generic_recursive_model…  │  1471           │  1473        │  SchemaError                    │
│  tests/test_generics.py         │  test_generic_recursive_model…  │  1535           │  1537        │  SchemaError                    │
│  tests/test_generics.py         │  test_generic_recursive_model…  │  1593           │  1595        │  SchemaError                    │
│  tests/test_generics.py         │  test_generic_recursive_model…  │  1638           │  1640        │  SchemaError                    │
│  tests/test_generics.py         │  test_generic_recursive_model…  │  1668           │  1678        │  SchemaError                    │
│  tests/test_generics.py         │  test_generic_recursive_model…  │  1742           │  1756        │  SchemaError                    │
│  tests/test_generics.py         │  test_multi_inheritance_gener…  │  1917           │  1932        │  ValidationError                │
│  tests/test_generics.py         │  test_double_typevar_substitu…  │  2026           │  2032        │  ValidationError                │
│  tests/test_json.py             │  test_custom_encode_fallback_…  │  316            │  329         │  PydanticSchemaGenerationError  │
│  tests/test_json_schema.py      │  test_advanced_generic_schema   │  2371           │  2448        │  TypeError                      │
[...]

PR #5361, opened 2023-04-03
Move generic metadata into a typeddict

dmontagu:
Addresses one of the checkboxes in https://github.com/pydantic/pydantic/issues/5351

Selected Reviewer: @adriangb


pydantic/main.py
```diff
@@ -56,16 +56,35 @@
 _base_class_defined = False
 
 
+def build_generic_metadata(
```
dmontagu, line 59: Probably want to make this private at least, if not move into the `_generics` module. Since it might be convenient to see this logic while working on `ModelMetaclass` I wasn't sure if it was a good idea to pull it out of this module.

dmontagu:
please review

pydantic/_internal/_core_utils.py
```diff
@@ -2,13 +2,16 @@
...
-from typing import Any, Callable, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, Union, cast
 
 from pydantic_core import CoreSchema, CoreSchemaType, core_schema
 from typing_extensions import TypeGuard, get_args
 
 from . import _repr
 
+if TYPE_CHECKING:
+    from ._generics import PydanticGenericMetadata
```
adriangb, line 13: Are we doing this because of import cycle or performance?

pydantic/_internal/_generics.py
```diff
@@ -113,6 +114,14 @@ def __delitem__(self, key: KT) -> None:
 _GENERIC_TYPES_CACHE = GenericTypesCache()
 
 
+class PydanticGenericMetadata(typing_extensions.TypedDict, total=False):
+    args: tuple[Any, ...] | None
+    defaults: dict[str, Any] | None
```
adriangb, line 119: Some docstrings would be nice for these
```suggestion
    args: tuple[Any, ...] | None
    """Generic arguments to this model, e.g. `Model[int, str]` -> `(int, str)`"""
    defaults: dict[str, Any] | None
```

pydantic/main.py
adriangb: If this is a new function, any reason why it shouldn't be private? It seems to be used only in this module.

I like `update_generic_metadata_for_class`. I am curious as to why it's updating the metadata instead of "collecting it. I generally prefer to collect it and let the caller update it via `metadata = {**metadata, **collect_metadata_from_cls(cls)}` or something like that.

adriangb (approved):

pydantic/main.py
dmontagu: There are two "scenarios" where this `build_generic_metadata`/`update_generic_metadata_for_class` needs to do non-trivial work:

1. When you create a "new" generic class (i.e., when you create a new class that subclasses `BaseModel` _and_ `typing.Generic`). In this case, the `__pydantic_generic_metadata__` argument to `ModelMetaclass.__new__` will be `None`.
2. When you parametrize a generic class (i.e., call `BaseModel.__class_getitem__`), which in turn calls `create_generic_submodel`, this calls `ModelMetaclass.__new__` with a non-trivial `__pydantic_generic_metadata__`.

This implementation allows much/most of the logic to be shared between the two paths. I am not yet sure if this can be simplified; my intuition is it will require an annoying amount of logic duplication, but I'm looking into it.

pydantic/_internal/_core_utils.py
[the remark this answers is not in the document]
dmontagu: I think there was an import cycle, but I was able to eliminate the need for it. (Will push changes shortly.)

dmontagu:
@adriangb Sorry to drop a big refactor to this after you reviewed it, but after pulling the thread on the thing you asked about above, I realized I could now eliminate the `__pydantic_generic_typevars_map__` and `__pydantic_generic_defaults__` entirely (in addition to not really needing the `build_generic_metadata`).

(I think the changes in #5305 — the model creation refactor PR — opened up various simplifications to the generics building that hadn't been fully realized.)

Interestingly, we now only have `args`, `origin`, and `parameters`, which are all analogous to attributes used in the `typing` module. I think it still makes sense to pull them out special as I have done here, though perhaps the impact is a bit smaller, and also perhaps one day we will change it to `__args__`, `__origin__` etc., if python adds improved support for user-defined GenericAlias etc.

pydantic/_internal/_generics.py
dmontagu: since it's just args, origin, and parameters now, I just added comments saying `# analogous to typing._GenericAlias.__args__` etc. Let me know if you think that's insufficient.

dmontagu:
please review

pydantic/main.py
```diff
@@ -110,26 +108,26 @@ def hash_func(self_: Any) -> int:
...
+                cls.__pydantic_generic_metadata__ = __pydantic_generic_metadata__
             else:
-                new_typevars_map = dict(
-                    zip(_generics.iter_contained_typevars(__pydantic_generic_origin__), __pydantic_generic_args__ or ())
-                )
-                cls.__pydantic_generic_typevars_map__ = {**parent_typevars_map, **new_typevars_map}
+                parameters = getattr(cls, '__parameters__', ())
+                parent_parameters = getattr(cls, '__pydantic_generic_metadata__', {}).get('parameters', ())
+                if parameters and parent_parameters and not all(x in parameters for x in parent_parameters):
+                    # assert Generic in bases  # tests pass with this assert, but I think we shouldn't include it
```
adriangb, line 118: Why shouldn't we include it? It seems reasonable to me, if nothing else as documentation of what this codepath expects

adriangb (approved):
LGTM!

pydantic/main.py
dmontagu: Because if I am wrong and it is possible to hit this path _without_ `Generic in bases`, the AssertionError there will be a lot more confusing for a user, and have less useful context for us to debug a reported issue containing the traceback than we'd get if it hits the subsequent part of the code.

Perhaps including a listing of the base classes in the exception would be a best-of-both-worlds alternative. I'll make the change.

pydantic/main.py
```diff
@@ -110,26 +108,31 @@ def hash_func(self_: Any) -> int:
...
+                    parameters_str = ', '.join([str(x) for x in combined_parameters])
+                    error_message = (
+                        f'All parameters must be present on typing.Generic;'
+                        f' you should inherit from typing.Generic[{parameters_str}]'
+                    )
+                    if Generic not in bases:  # pragma: no cover
+                        # This branch will only be hit if I have misunderstood how `__parameters__` works.
+                        # If that is the case, and a user hits this, I could imagine it being very helpful
+                        # to have this extra detail in the reported traceback.
+                        error_message += f' (bases={bases})'
```
dmontagu, line 128: @adriangb are you cool with this tweak? If so, I think we can merge this.

PR #5362, opened 2023-04-04
Handle functools.partial/partialmethod in validators

### LINKED ISSUE #5352: Remove requirement that callables have `__qualname__` attribute

adriangb:
Fixes #5352

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -4,6 +4,7 @@ Logic related to validators applied to models etc. via the `@validator` and `@ro
 from __future__ import annotations as _annotations
 
 import warnings
+from functools import partial, partialmethod
 from inspect import Parameter, Signature, signature
 from typing import (
     TYPE_CHECKING,
@@ -293,7 +294,7 @@ class DecoratorInfos(Representation):
         self.model_serializer = {}
 
 
-def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
+def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:  # noqa: C901
     """
     We want to collect all DecFunc instances that exist as
     attributes in the namespace of the class (a BaseModel or dataclass)
@@ -321,8 +322,15 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
 
     for var_name, var_value in vars(cls).items():
         if isinstance(var_value, PydanticDecoratorMarker):
-            func = var_value.wrapped.__get__(None, cls)
             cls_ref = get_type_ref(cls)
+            try:
+                func = var_value.wrapped.__get__(None, cls)
+            except AttributeError:
+                if isinstance(var_value.wrapped, partial):
+                    # don't bind the class
+                    func = var_value.wrapped
+                else:
+                    raise
             shimmed_func = var_value.shim(func) if var_value.shim is not None else func
             info = var_value.decorator_info
             if isinstance(info, ValidatorDecoratorInfo):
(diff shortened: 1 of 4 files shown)


adriangb:
Missing serializers (WIP) but please review

pydantic/_internal/_decorators.py
```diff
@@ -341,66 +349,93 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
...
+def prepare_serializer_decorator(function: AnyDecoratorCallable, allow_reuse: bool) -> Any:
     """
     Warn about validators/serializers with duplicated names since without this, they can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
     """
-    if isinstance(function, staticmethod):
-        function = function.__func__
     if not allow_reuse and not in_ipython():
-        ref = f'{function.__module__}::{function.__qualname__}'
+        ref = get_function_ref(function)
```
dmontagu, line 398: In `check_for_duplicate_decorator_function ` below you do:
```python
        function = unwrap_wrapped_function(function)
        ref = get_function_ref(function)
```
Whereas here, you don't do the unwrapping. Is this intentional? Without digging into the full logic path it seems like it should be the same in both places but I could imagine it being possible you already know the function has been unwrapped here, or similar.

adriangb: I just deleted this function _and_ deleted the `function = unwrap_wrapped_function(function)` in `check_for_duplicate_decorator_function` because `get_function_ref` already does that.

adriangb:
please review

pydantic/_internal/_decorators.py
```diff
@@ -341,68 +349,83 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
...
-    """
-    if isinstance(function, staticmethod):
-        function = function.__func__
-    if not allow_reuse and not in_ipython():
-        ref = f'{function.__module__}::{function.__qualname__}'
-        if ref in _FUNCS:
-            warnings.warn(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
-        _FUNCS.add(ref)
-    return function
+AnyDecoratorCallable: TypeAlias = 'Union[ classmethod[Any], staticmethod[Any], partialmethod[Any], Callable[..., Any]]'
```
samuelcolvin, line 352:
```suggestion
AnyDecoratorCallable: TypeAlias = 'Union[classmethod[Any], staticmethod[Any], partialmethod[Any], Callable[..., Any]]'
```

pydantic/_internal/_decorators.py
```diff
@@ -341,68 +349,83 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
...
-    Unwrap unbound classmethods and staticmethods
-    """
-    if isinstance(function, (classmethod, staticmethod)):
-        return function.__func__
-    return function
+def unwrap_wrapped_function(
+    func: Any,
+    *,
+    unwrap_class_static_method: bool = True,
+) -> Any:
```
samuelcolvin, line 359: could we add a docstring explaining what this is doing.

pydantic/_internal/_decorators.py
```diff
@@ -341,68 +349,83 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
...
+            staticmethod,
+            classmethod,
+            partial,
+            partialmethod,
+        )
+    else:
+        all = (
+            partial,
+            partialmethod,
+        )
```
samuelcolvin, line 372:
```suggestion
        all = partial, partialmethod
```

pydantic/_internal/_decorators.py
```diff
@@ -341,68 +349,83 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
...
     Warn about validators with duplicated names since without this, they can be overwritten silently
     which generally isn't the intended behaviour, don't run in ipython (see #312) or if `allow_reuse` is True.
     """
     if not allow_reuse and not in_ipython():
-        function = unwrap_unbound_methods(function)
-        ref = f'{function.__module__}::{function.__qualname__}'
+        ref = get_function_ref(function)
         if ref in _FUNCS:
-            warnings.warn(f'duplicate validator function "{ref}"; if this is intended, set `allow_reuse=True`')
+            warnings.warn(f'duplicate {type} function "{ref}"; if this is intended, set `allow_reuse=True`')
```
samuelcolvin, line 428: Remember we want to replace this with #4436.

But happy to defer that to a new PR. @hramezani, probably don't start work on #4436 until this is merged.

samuelcolvin (commented):
otherwise I think looks good.

samuelcolvin:
also conflicts, please update.

pydantic/_internal/_decorators.py
adriangb: done

adriangb:
please review

hramezani:
Maybe not related to this PR, but I can't see any test for duplicate serializers function like [the test_duplicate for validator](https://github.com/pydantic/pydantic/blob/0cd57b3e0c75d9622012de542e5ddc110fe55743/tests/test_validators.py#L370). I think by the change of this PR, both serializers and validators are going to use `check_for_duplicate_decorator_function` but I think it worth to add some test for duplicate serializers. Also, the above-mentioned test is for `field_validator`, what about adding tests for other validators?

samuelcolvin:
Good point, but let's leave that until we replace #4436.

samuelcolvin:
great, thanks so much.

PR #5364, opened 2023-04-04
Implement Annotated validators

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -73,7 +73,6 @@ class FieldValidatorDecoratorInfo:
 
     fields: tuple[str, ...]
     mode: Literal['before', 'after', 'wrap', 'plain']
-    sub_path: tuple[str | int, ...] | None
     check_fields: bool | None
 
 
@@ -101,7 +100,6 @@ class FieldSerializerDecoratorInfo:
     type: Literal['general', 'field']
     json_return_type: JsonReturnTypes | None
     when_used: WhenUsed
-    sub_path: tuple[str | int, ...] | None
     check_fields: bool | None
 
 
@@ -455,23 +453,9 @@ def remove_params_with_defaults(sig: Signature) -> Signature:
     return Signature([p for p in sig.parameters.values() if p.default is Parameter.empty])
 
 
-@overload
-def make_generic_v2_field_validator(
-    validator: FieldWrapValidatorFunction, mode: Literal['wrap']
-) -> FieldWrapValidatorFunction:
-    ...
-
-
-@overload
-def make_generic_v2_field_validator(
-    validator: OnlyValueValidator | FieldValidatorFunction, mode: Literal['before', 'after', 'plain']
-) -> FieldValidatorFunction:
-    ...
-
-
-def make_generic_v2_field_validator(
+def make_generic_validator(
     validator: OnlyValueValidator | FieldValidatorFunction | FieldWrapValidatorFunction, mode: str
-) -> FieldValidatorFunction | FieldWrapValidatorFunction:
+) -> Any:
     """
     In order to support different signatures, including deprecated validator signatures from v1,
     we introspect the function signature and wrap it in a parent function that has a signature
@@ -556,9 +540,9 @@ AnySerializerFunction = Union[
 ]
 
 
-def make_generic_field_serializer(
+def make_generic_serializer(
     serializer: AnySerializerFunction, mode: Literal['plain', 'wrap'], type: Literal['field', 'general']
-) -> AnyCoreSerializer:
+) -> Any:
     """
     Wrap serializers to allow ignoring the `info` argument as a convenience.
     """
--- a/pydantic/decorators.py
+++ b/pydantic/decorators.py
@@ -188,7 +188,6 @@ def field_validator(
     *fields: str,
     mode: Literal['before', 'after', 'plain'] = ...,
     check_fields: bool | None = ...,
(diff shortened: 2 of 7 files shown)


pydantic/decorators.py
```diff
@@ -464,3 +464,8 @@ def dec(f: Callable[..., Any]) -> _decorators.PydanticDecoratorMarker[Any]:
         return dec
     else:
         return dec(__f)
+
+
+class AfterValidator:
+    def __init__(self, __func: _decorators.OnlyValueValidator | _core_schema.FieldValidatorFunction) -> None:
+        self.func = _decorators.make_generic_v2_field_or_general_validator(__func, mode='after')
```
adriangb: I wasn't sure where to put this, here is convenient because we have the `OnlyValueValidator` and other types, but I'm open to put it other places

adriangb:
I would like some initial feedback on the direction, please review

pydantic/_internal/_decorators.py
```diff
@@ -536,23 +536,7 @@ def wrapper3(value: Any, info: FieldValidationInfo) -> Any:
...
-def make_generic_v2_field_validator(
-    validator: OnlyValueValidator | FieldValidatorFunction, mode: Literal['before', 'after', 'plain']
-) -> FieldValidatorFunction:
-    ...
-
-
-def make_generic_v2_field_validator(
-    validator: OnlyValueValidator | FieldValidatorFunction | FieldWrapValidatorFunction, mode: str
-) -> FieldValidatorFunction | FieldWrapValidatorFunction:
+def make_generic_v2_field_or_general_validator(validator: Any, mode: str) -> Any:
```
adriangb, line 539: I gave up on the type annotations here...

pydantic/decorators.py
```diff
@@ -464,3 +464,10 @@ def dec(f: Callable[..., Any]) -> _decorators.PydanticDecoratorMarker[Any]:
         return dec
     else:
         return dec(__f)
+
+
+class AfterValidator:
+    def __init__(self, __func: _decorators.OnlyValueValidator | _core_schema.GeneralValidatorFunction) -> None:
+        self.func: _core_schema.GeneralValidatorFunction = _decorators.make_generic_v2_field_or_general_validator(
+            __func, mode='after'
+        )
```
adriangb, line 473: Is there a better place to put these?

pydantic/_internal/_decorators.py
samuelcolvin: fair, fine with me. This is the good side of python types.

samuelcolvin: side note: I do think we shouldn't use `v2` in names, that should be the default.

pydantic/decorators.py
samuelcolvin: I guess we need a new module, since they're not dectorators.

We combine this with some of the other types designed to be used as annotations, like `Strict` and put them in `annoated_types.py` - we can add a docstring say these are pydantic specific extensions to the ideas of `annoated-types`.

Should we make this a dataclass?

I know we haven't used dataclasses elsewhere here, but I think we have in `types.py`, I don't see why we shouldn't.

tests/test_validators.py
```diff
@@ -22,6 +23,19 @@
 from pydantic.decorators import field_validator, root_validator
 
 
+def test_annotated_validator() -> None:
+    MyInt = Annotated[int, AfterValidator(lambda x: x if x != -1 else 0)]
```
samuelcolvin, line 27: we should simplify tests to show using just one `AfterValidator`, then more complex usage.

tests/test_validators.py
```diff
@@ -22,6 +23,19 @@
...
+
+    def any_key_but_exclude(d: Dict[str, Any]) -> Dict[str, Any]:
+        return {k: v for k, v in d.items() if k != 'exclude'}
```
samuelcolvin, line 30: I thnk this could also be done in the item validator using [`PydanticOmit`](https://github.com/pydantic/pydantic-core/blob/f44a94298b4be4241c23f61bd528004c8a7e9777/pydantic_core/_pydantic_core.pyi#L208) without needing to iterate over the dict again.

We should probably check that behaviour rather than this route.

samuelcolvin (commented):
otherwise looking good.

I guess we need to add:
* `BeforeValidator`
* `WrapValidator`

tests/test_validators.py
adriangb: This was somewhat of an arbitrary example, I'll change it to something else

pydantic/decorators.py
adriangb: Sounds good. I can make this a dataclass. If we don't mind using dataclasses I think there's several other things that could be dataclasses, but I can start here

samuelcolvin: absolutely. I guess we can have something like

```py
if whatever version:
     dc_kwargs = {'slots': Ture}
else:
    dc_kwargs = {}
```

So we can use slots on these dataclasses in a backwards compatible way, we might already do that in `types.py`

adriangb: Yup we do that in Annotated-types, I can just copy it

tests/test_validators.py
adriangb: replaced

tests/test_validators.py
adriangb: split it up into two tests

adriangb:
please review

dmontagu (approved):

PR #5366, opened 2023-04-04
Add support for frozen and final fields

dmontagu:
Replaces #5354 

### DIFF
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -15,7 +15,7 @@ from pydantic_core import core_schema
 from ._forward_ref import PydanticForwardRef
 from ._generics import replace_types
 from ._repr import Representation
-from ._typing_extra import get_cls_type_hints_lenient, get_type_hints, is_classvar
+from ._typing_extra import get_cls_type_hints_lenient, get_type_hints, is_classvar, is_finalvar
 
 if TYPE_CHECKING:
     from ..fields import FieldInfo
@@ -152,6 +152,9 @@ def collect_fields(  # noqa: C901
         if is_classvar(ann_type):
             class_vars.add(ann_name)
             continue
+        if _is_finalvar_with_default_val(ann_type, getattr(cls, ann_name, Undefined)):
+            class_vars.add(ann_name)
+            continue
         if ann_name.startswith('_') or (omitted_fields and ann_name in omitted_fields):
             continue
 
@@ -264,3 +267,15 @@ def collect_fields(  # noqa: C901
             field.annotation = replace_types(field.annotation, typevars_map)
 
     return fields, class_vars
+
+
+def _is_finalvar_with_default_val(type_: type[Any], val: Any) -> bool:
+    from pydantic.fields import FieldInfo
+
+    if not is_finalvar(type_):
+        return False
+    if val is Undefined:
+        return False
+    if isinstance(val, FieldInfo) and (val.default is Undefined and val.default_factory is None):
+        return False
+    return True
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -274,8 +274,9 @@ def _is_finalvar_with_default_val(type_: type[Any], val: Any) -> bool:
 
     if not is_finalvar(type_):
         return False
-    if val is Undefined:
+    elif val is Undefined:
         return False
-    if isinstance(val, FieldInfo) and (val.default is Undefined and val.default_factory is None):
+    elif isinstance(val, FieldInfo) and (val.default is Undefined and val.default_factory is None):
         return False
-    return True
+    else:
+        return True
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -14,7 +14,7 @@ from typing import TYPE_CHECKING, Any, Callable, ForwardRef, Iterable, Mapping, 
 
 from annotated_types import BaseMetadata, GroupedMetadata
 from pydantic_core import SchemaError, SchemaValidator, core_schema
-from typing_extensions import Annotated, Literal, TypedDict, get_args, get_origin, is_typeddict
+from typing_extensions import Annotated, Final, Literal, TypedDict, get_args, get_origin, is_typeddict
 
 from ..errors import PydanticSchemaGenerationError, PydanticUndefinedAnnotation, PydanticUserError
 from ..fields import FieldInfo
@@ -40,6 +40,7 @@ from ._decorators import (
 from ._fields import PydanticGeneralMetadata, PydanticMetadata, Undefined, collect_fields, get_type_hints_infer_globalns
 from ._forward_ref import PydanticForwardRef, PydanticRecursiveRef
 from ._generics import recursively_defined_type_refs, replace_types
+from ._typing_extra import is_finalvar
 
 if TYPE_CHECKING:
     from ..config import ConfigDict
@@ -368,6 +369,10 @@ class GenerateSchema:
             # probably need to take care of other subclasses here
         elif isinstance(obj, typing.TypeVar):
             return self._unsubstituted_typevar_schema(obj)
+        elif is_finalvar(obj):
+            if obj is Final:
+                return core_schema.AnySchema(type='any')
+            return self.generate_schema(get_args(obj)[0])
 
         # TODO: _std_types_schema iterates over the __mro__ looking for an expected schema.
         #   This will catch subclasses of typing.Deque, preventing us from properly supporting user-defined
@@ -465,6 +470,7 @@ class GenerateSchema:
             serialization_exclude=common_field['serialization_exclude'],
             validation_alias=common_field['validation_alias'],
             serialization_alias=common_field['serialization_alias'],
+            frozen=common_field['frozen'],
             metadata=common_field['metadata'],
         )
 
@@ -486,6 +492,7 @@ class GenerateSchema:
             serialization_exclude=common_field['serialization_exclude'],
             validation_alias=common_field['validation_alias'],
             serialization_alias=common_field['serialization_alias'],
+            frozen=common_field['frozen'],
             metadata=common_field['metadata'],
         )
 
@@ -536,6 +543,7 @@ class GenerateSchema:
             serialization_exclude=True if field_info.exclude else None,
             validation_alias=field_info.alias,
             serialization_alias=field_info.alias,
+            frozen=field_info.frozen or field_info.final,
             metadata=metadata,
         )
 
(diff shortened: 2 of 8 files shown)


pydantic/_internal/_fields.py
```diff
@@ -264,3 +267,15 @@ def collect_fields(  # noqa: C901
             field.annotation = replace_types(field.annotation, typevars_map)
 
     return fields, class_vars
+
+
+def _is_finalvar_with_default_val(type_: type[Any], val: Any) -> bool:
+    from pydantic.fields import FieldInfo
```
dmontagu, line 273: don't like needing the local import but couldn't see anywhere it could go that would eliminate this, and putting this function in this file seemed like the best way to do it

pydantic/_internal/_generate_schema.py
```diff
@@ -368,6 +369,10 @@ def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
             # probably need to take care of other subclasses here
         elif isinstance(obj, typing.TypeVar):
             return self._unsubstituted_typevar_schema(obj)
+        elif is_finalvar(obj):
+            if obj is Final:
+                return core_schema.AnySchema(type='any')
+            return self.generate_schema(get_args(obj)[0])
```
dmontagu, line 375: This is essentially the same logic as was used in v1

pydantic/_internal/_typing_extra.py
```diff
@@ -200,7 +200,7 @@ def _check_finalvar(v: type[Any] | None) -> bool:
     return v.__class__ == Final.__class__ and (sys.version_info < (3, 8) or getattr(v, '_name', None) == 'Final')
 
 
-def is_finalvar(ann_type: type[Any]) -> bool:
+def is_finalvar(ann_type: Any) -> bool:
     return _check_finalvar(ann_type) or _check_finalvar(get_origin(ann_type))
```
dmontagu, line 204: I needed to change this to deal with `PydanticForwardRef`. As far as I can tell, the `_check_finalvar` wouldn't fail for those, though it is annotated as receiving a `type[Any]`

pydantic/fields.py
```diff
@@ -60,7 +62,6 @@ class FieldInfo(_repr.Representation):
         'allow_inf_nan': None,
         'min_items': None,
         'max_items': None,
-        'frozen': None,
```
dmontagu, line 63: I'm not sure why `frozen` was in the metadata lookup like this, or if it needed to be; moving it to being an attribute of `FieldInfo` didn't make any tests fail though and made the necessary changes easier.

pydantic/fields.py
```diff
@@ -279,6 +299,8 @@ def __repr_args__(self) -> ReprArgs:
                 continue
             elif s == 'repr' and self.repr is True:
                 continue
+            elif s == 'final':
```
dmontagu, line 302: This causes it to not show `final` in the repr for `FieldInfo`. I can undo this before merging if desired, it just requires some tests of repr to be updated which I figured wasn't helpful at this stage.

tests/test_edge_cases.py
```diff
@@ -2155,13 +2153,20 @@ class Foo(BaseModel):
     class Bar(BaseModel):
         model_config = ConfigDict(validate_assignment=True)
         a: str = Field(..., frozen=True)
+        c: Annotated[str, Field(frozen=True)]
```
dmontagu, line 2156: Just adding a check that it also works when applied through a pure Annotation

tests/test_main.py
```diff
@@ -1510,8 +1509,9 @@ class Entry(BaseModel):
     r.val = 101
     assert r.val == 101
     assert r.id == 1
-    with pytest.raises(TypeError, match='"id" has frozen set to True and cannot be assigned'):
+    with pytest.raises(ValidationError) as exc_info:
```
dmontagu, line 1512: It seems in v1, assigning to frozen fields raised a TypeError, rather than a ValidationError. Is this okay to change?

dmontagu: Err, I guess it looks like frozen fields weren't even a thing in v1. But the xfailing test was using a TypeError. I'm assuming it's okay to make this a ValidationError.

tests/test_main.py
```diff
@@ -1755,27 +1751,25 @@ def test_final_field_decl_with_default_val(ann):
...
-    Model.model_rebuild(ann=ann)
-
     assert 'a' in Model.__class_vars__
     assert 'a' not in Model.model_fields
 
 
-@pytest.mark.xfail(reason='waiting for https://github.com/pydantic/pydantic-core/pull/237')
 def test_final_field_reassignment():
     class Model(BaseModel):
+        model_config = ConfigDict(validate_assignment=True)
```
dmontagu, line 1760: This is a change from v1 — in v1:
* it would error for setting values to final fields even without having `validate_assignment=True` (or anything like it)
* it would raise a TypeError, not a ValidationError.

Do we need to change this? If so, I guess that would require a bit more work to support `Final` since it would presumably no longer be a pure pydantic_core thing. Is it possible to turn on `validate_assignment` on a per-field basis in pydantic core?

dmontagu: For reference:

https://github.com/pydantic/pydantic/blob/11f7f9019016659e81c96481ffe5c42637489cf3/tests/test_main.py#L2283-L2293

dmontagu:
please review

pydantic/_internal/_fields.py
```diff
@@ -264,3 +267,15 @@ def collect_fields(  # noqa: C901
...
+
+    if not is_finalvar(type_):
+        return False
+    if val is Undefined:
+        return False
+    if isinstance(val, FieldInfo) and (val.default is Undefined and val.default_factory is None):
+        return False
+    return True
```
samuelcolvin, line 281:
```suggestion
    if not is_finalvar(type_):
        return False
    elif val is Undefined:
        return False
    elif isinstance(val, FieldInfo) and (val.default is Undefined and val.default_factory is None):
        return False
    else:
        return True
```

pydantic/_internal/_fields.py
samuelcolvin: fine by me.

pydantic/_internal/_typing_extra.py
samuelcolvin: fine by me.

pydantic/fields.py
```diff
@@ -95,6 +96,8 @@ def __init__(self, **kwargs: Any) -> None:
         self.init_var = kwargs.get('init_var', None)
         self.kw_only = kwargs.get('kw_only', None)
         self.validate_default = kwargs.get('validate_default', None)
+        self.frozen = kwargs.get('frozen', None)
+        self.final = kwargs.get('final', None)
```
samuelcolvin, line 100: I thought we were combining these?

If there's a good reason not to, fine by me.

pydantic/fields.py
```diff
@@ -124,16 +127,25 @@ def from_annotation(cls, annotation: type[Any] | _forward_ref.PydanticForwardRef
...
         """
+        final = False
+        if _typing_extra.is_finalvar(annotation):
+            final = True
+            if annotation is not typing_extensions.Final:
+                annotation = typing_extensions.get_args(annotation)[0]
+
         if _typing_extra.is_annotated(annotation):
             first_arg, *extra_args = typing_extensions.get_args(annotation)
+            if _typing_extra.is_finalvar(annotation):
```
samuelcolvin, line 138: do we need this check too? looks like it duplicates the check above.

pydantic/fields.py
samuelcolvin: Up to you, is there a particular reason to omit "final"?

tests/test_main.py
samuelcolvin: fine by me.

tests/test_main.py
samuelcolvin: I think fine to keep the change.

Again, if they come with pitchforks, we can have special logic in core to raise a type error perhaps.

tests/test_main.py
```diff
@@ -1755,27 +1751,25 @@ def test_final_field_decl_with_default_val(ann):
...
+
         a: Final[int]
 
     obj = Model(a=10)
 
     with pytest.raises(
-        TypeError,
-        match=r'^"Model" object "a" field is final and does not support reassignment$',
-    ):
+        ValidationError,
```
samuelcolvin, line 1767: you can combine these three lines.

samuelcolvin: we should probably remove the magic comma thing from black.

samuelcolvin (commented):
otherwise LGTM. please update.

tests/test_main.py
dmontagu: I'm fine changing the type error, but I'm more concerned about needing to specify `validate_assignment=True` for it to be used. (Which might not be desirable.)

pydantic/fields.py
dmontagu: I think this is necessary to handle both `Final[Annotated[int, Field(...)]]` and `Annotated[Final[int], Field(...)]`

dmontagu: It looks like I have a typo and should be checking `if _typing_extra.is_finalvar(first_arg):` though

pydantic/fields.py
dmontagu: I think the reason I didn't combine them was two-fold: one, `ModelField` had an attribute called `final` in v1, but not `frozen`. So it felt a little weird to remove `final`, especially considering conceptually it's different from `frozen` (for starters, in theory we shouldn't allow overriding the default, though I suspect it's not worth the effort to enforce outside of type checking).

Second, I imagined we might eventually want to give different error messages depending on frozen vs. final. (I could imagine a newer user getting an error message saying 'frozen' when you just annotated as Final and possibly don't even know what frozen refers to could be confusing.) In which case perhaps it would be helpful to have the distinction.

However, in this PR, I just set the frozen attribute on the field based on `frozen=field_info.frozen or field_info.final`. So yeah maybe we should just combine the two and only store `frozen` on the field.

If you read the above and don't have an opinion/haven't changed your mind, I suppose I'll just combine them and drop the `final` attribute, and it can be a feature request in the future.

pydantic/fields.py
dmontagu: I think maybe we should remove it anyway, as noted above.

pydantic/fields.py
dmontagu: image
pycharm says we don't need to do `Annotated[Final[...` 👍

dmontagu: this is interesting though:
image
should we drop this functionality? haha. I guess no reason to given we've already ported it.

dmontagu:
please review

pydantic/fields.py
samuelcolvin: agreed.

samuelcolvin: let's keep it since we have it.

pydantic/fields.py
samuelcolvin: fine, let's keep it.

samuelcolvin:
great, thanks so much.

Issue #5367, opened 2023-04-04
Default values in JSON Schema don't use alias of nested type fields


jothepro:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

I am trying to define a model with a nested model that has default values for all its fields and aliases defined for some.
If the default values don't need to be changed, the whole nested model should be optional for the user. But the resulting model must contain an instance of the nested model with the default values.

I am defining a model like this:

```py
class InnerModel(BaseModel):
    foo: str = "bar"
    baz: str = Field(default="foobar", alias="my_alias")

class OuterModel(BaseModel):
    nested_field: InnerModel = InnerModel()
```

Using this model works as expected. I can create an instance of `OuterModel` without supplying a value for `nested_field`, and the field will be populated with the default value:

```py
outer_model = OuterModel()
print(outer_model.nested_field.baz) # "foobar"
```

Alternatively, I can also supply custom values for the `InnerModel`, using the alias in the constructor:

```py
outer_model = OuterModel(nested_field=InnerModel(my_alias="value"))
print(outer_model.nested_field.baz) # "value"
```

However, when rendering the JSON Schema, the default values of the nested model are reported incorrectly, referring to the field key instead of the alias:

```py
outer_model.model_json_schema() # 'default': {'foo': 'bar', 'baz': 'foobar'} <-- 'baz' should be named 'my_alias'!
```

 I'd expect the default values in the JSON Schema to use the alias, because this is the public interface that I am communicating to the user: `'default': {'foo': 'bar', 'my_alias': 'foobar'}`

I was able to fix the wrong schema by overriding the `default` field in the schema, but IMO this should be the default behavior anyways:

```py
class OuterModel(BaseModel):
    nested_field: InnerModel = Field(default=InnerModel(), json_schema_extra={
        "default": { value.alias or key: value.default for key, value in InnerModel.model_fields.items()}
    })
```

I wonder if I am doing something wrong of if this may be a bug? The same behavior can be observed vor both V1 and V2.

### Example Code

```Python
from pydantic import BaseModel, Field

class InnerModel(BaseModel):
    foo: str = "bar"
    baz: str = Field(default="foobar", alias="my_alias")

class OuterModel(BaseModel):
    nested_field: InnerModel = InnerModel()

outer_model = OuterModel()
print(outer_model.nested_field.baz) # "foobar"

[...]

PR #5368, opened 2023-04-04
Fix handling of default values of BaseModel subclasses in JSON schema

dmontagu:
Currently just adds a failing unit test, taken from #5367. Can be continued or closed/replaced once the necessary changes are in place in `pydantic_core`.


dmontagu:
I think this might already work without any changes once https://github.com/pydantic/pydantic-core/pull/533 is merged/released and pydantic is updated to use that release of pydantic core. (Since to_jsonable_python has `by_alias=True` by default)

dmontagu:
After the latest commit in pydantic/pydantic-core#533, I have confirmed that this test will pass.

PR #5372, opened 2023-04-04
Break out types into individual topics

tpdorsey:
WIP

To do:
* Still needs content work

## Change Summary

* Pulls Field Types subtopics into individual pages

## Checklist

Selected Reviewer: @hramezani


docs/usage/annotated_types.md
```diff
@@ -0,0 +1,102 @@
+---
+description: Support for annotated types including NamedTuple and TypedDict.
```
samuelcolvin, line 2: "annotated types" has a particular meaning - https://github.com/annotated-types/annotated-types so probably not a good name.

Perhaps "types with fields", and just start by saying there are four supported types with fields:
* base models
* dataclasses
* typed dicts
* named tuples

docs/usage/strict.md
```diff
@@ -0,0 +1,91 @@
+
+
+You can use the `StrictStr`, `StrictBytes`, `StrictInt`, `StrictFloat`, and `StrictBool` types
```
samuelcolvin, line 3: better to move these next to the types they go with, rather than being a section on their own.

docs/usage/bytes.md
samuelcolvin, line 1: probably shouldn't be called `bytes.md` if it doesn't include the `bytes` type, which should be next to `str` as they're often thought of together.

docs/usage/constrained.md
```diff
@@ -0,0 +1,151 @@
+
```
samuelcolvin, line 1: again these should be moved next to the types they relate to, rather than but in a bucket on their own.

docs/usage/pydantic_types.md
```diff
@@ -0,0 +1,169 @@
+---
+description: Useful types provided by Pydantic.
+---
+
+*pydantic* also provides a variety of other useful types:
+
+`FilePath`
```
samuelcolvin, line 7: file types can probably go in a file on their own.

docs/usage/pydantic_types.md
```diff
@@ -0,0 +1,169 @@
...
+: like `Path`, but the path must exist and be a file
+
+`DirectoryPath`
+: like `Path`, but the path must exist and be a directory
+
+`PastDate`
+: like `date`, but the date should be in the past
+
+`FutureDate`
```
samuelcolvin, line 16: date and datetime types can go in the datetime module.

docs/usage/pydantic_types.md
```diff
@@ -0,0 +1,169 @@
+---
+description: Useful types provided by Pydantic.
```
samuelcolvin, line 2: I guess you're aware but all of the types in here should be moved into the relevant file.

samuelcolvin (commented):
looks like we need files for the most common primative types:
* numbers - `int`, `float`, `Decimal`
* strings - `str`, `bytes`, `bytearray`
* lists & tuple
* dicts
* set & frozenset

samuelcolvin:
please update.

docs/usage/annotated_types.md
tpdorsey: Should this entire page be renamed then? The textual fix is simple

samuelcolvin: Yes I think so.

tpdorsey:
Please review

There is still content work to do here, but I think the pages are in place now

samuelcolvin:
thanks, this is a great start.

PR #5379, opened 2023-04-05
Add `@model_validator`

adriangb:
Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -38,17 +38,7 @@ from typing_extensions import Literal, Protocol, TypeAlias
 
 from ..errors import PydanticUserError
 from ._core_utils import get_type_ref
-<<<<<<< HEAD
 from ._dataclass import dataclass
-=======
-from ._repr import Representation
-
-if TYPE_CHECKING:
-    from typing_extensions import Literal
-
-    from pydantic.main import BaseModel
-
->>>>>>> 3d0b86c (Add @model_validator)
 
 FIELD_VALIDATOR_TAG = '_field_validator'
 ROOT_VALIDATOR_TAG = '_root_validator'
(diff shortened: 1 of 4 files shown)


pydantic/_internal/_decorators.py
```diff
@@ -786,3 +803,19 @@ def wrap_model_serializer_in_wrap_mode(
             raise TypeError(f'Unrecognized serializer signature for {serializer} with `mode={mode}`:{sig}')
         func = cast(AnyCoreSerializer, serializer)
         return func
+
+
+ModelType = TypeVar('ModelType', bound='BaseModel')
```
dmontagu, line 808: Should we loosen the bound here? E.g., I'm thinking this might apply to a dataclass or whatever.

dmontagu (commented):
Generally looking good to me

pydantic/_internal/_decorators.py
adriangb: Maybe, but then we end accepting `Callable[[Any, Any], Any]` in some situations lol. It's kind of a struggle to type these decorators, especially given how we let you include `info` or not and how we accept `classmethod[Any]` and such. I almost want to sit down with some of the typing folks at PyCon and see if they have any ideas. I do plan to do a bit more typing work on this PR, so I can play around with loosening the bound.

pydantic/_internal/_decorators.py
```diff
@@ -191,12 +193,32 @@ def __init__(
...
+class ModelValidatorDecoratorInfo(Representation):
+    """
+    A container for data from `@model_validator` so that we can access it
+    while building the pydantic-core schema.
+    """
+
+    __slots__ = ('mode',)
+
+    mode: Literal['wrap', 'before', 'after']
+    json_return_type: JsonReturnTypes | None
```
samuelcolvin, line 205: can be removed.

As mentioned on #5379, I think we should convert these to dataclasses.

pydantic/_internal/_decorators.py
samuelcolvin: I think we need to relax it to at least accept dataclass.

On this, I'd rather start loose and move to stricter types.

samuelcolvin (commented):
otherwise LGTM. please update.

pydantic/_internal/_decorators.py
adriangb: relaxed

pydantic/_internal/_decorators.py
adriangb: I'm confused: you linked to this pull request

adriangb: Nvm, makes sense now. See https://github.com/pydantic/pydantic/pull/5416

adriangb:
please review

dmontagu:
LGTM

PR #5380, opened 2023-04-05
fix "TypeError: init_private_attributes() got an unexpected keyword a…

commonism:
I expect this to meet the trivial criteria.

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -349,7 +349,7 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
             _fields_set = set(values.keys())
         _object_setattr(m, '__fields_set__', _fields_set)
         if hasattr(m, '__pydantic_post_init__'):
-            m.__pydantic_post_init__(context=None)
+            m.__pydantic_post_init__(_context=None)
         return m
 
     @classmethod


pydantic/main.py
```diff
@@ -349,7 +349,7 @@ def model_construct(cls: type[Model], _fields_set: set[str] | None = None, **val
             _fields_set = set(values.keys())
         _object_setattr(m, '__fields_set__', _fields_set)
         if hasattr(m, '__pydantic_post_init__'):
-            m.__pydantic_post_init__(context=None)
+            m.__pydantic_post_init__(_context=None)
```
dmontagu, line 352: I think this change would break the code here:

https://github.com/pydantic/pydantic/blob/d1b42d5ca2eb62936bd7c84622e9c45ebf494a9a/pydantic/main.py#L84-L93

I would suggest instead:
```suggestion
            m.__pydantic_post_init__(None)
```

Either way, we should add a test specifying precisely how keyword/positional arguments are passed. And I think we should _probably_ try to make sure we always call `__pydantic_post_init__` with `context` as a positional argument (as that is what `pydantic_core` seems to do).

dmontagu:
Closing in favor of #5382

PR #5382, opened 2023-04-05
Only call __pydantic_post_init__ with positional arguments

dmontagu:
Meant to replace #5380.

* Removes all references to `__pydantic_post_init__` and only refers to `model_post_init`.
* Ensures we only call the `model_post_init` using positional arguments.
* Changes free functions intended to behave as bound BaseModel methods to use `self: BaseModel` as the annotation, rather than `self_: Any`.
* Doesn't add `_init_private_attributes` into the model namespace (I think this would be _super_ confusing for a user if they happened to try to declare a method with the same name for any reason, and it got replaced due to the use of private attributes)
* Adds a (parametrized) test demonstrating the full spectrum of calls involving `model_post_init` during class instantiation.
* Takes `model_post_init` out of the `if TYPE_CHECKING:` block _without_ resulting in always calling it, by explicitly checking for overriding, rather than just the presence of the attribute. (This ensures calling `super().model_post_init` won't error even in direct subclasses of `BaseModel`.)

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -214,6 +214,8 @@ class GenerateSchema:
         """
         Generate schema for a pydantic model.
         """
+        from pydantic.main import BaseModel
+
         model_ref = get_type_ref(cls)
         cached_def = self.recursion_cache.get(model_ref)
         if cached_def is not None:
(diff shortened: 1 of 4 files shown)


dmontagu:
please review

hramezani (approved):
LGTM 👍 
Would be good to somebody else take a look as well

PR #5384, opened 2023-04-05
Remove contuple from types

hramezani:
It's not needed per discussion with @samuelcolvin 

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -16,7 +16,6 @@ from typing import (
     Hashable,
     List,
     Set,
-    Tuple,
     TypeVar,
     cast,
 )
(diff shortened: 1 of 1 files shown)


hramezani:
please review

dmontagu (approved):

dmontagu:
I see @samuelcolvin was assigned but this seems fine considering this function didn't exist in v1 and you can use Annotated  directly to achieve the same results. I'll just merge it 👍

hramezani:
> I see @samuelcolvin was assigned but this seems fine considering this function didn't exist in v1 and you can use Annotated directly to achieve the same results. I'll just merge it 👍

Yes, that's the reason. Thanks!

ModischFabrications:
Sorry to necro, but I was searching exactly for this type and this is the only mention. Might be sensible to add a hint to the docs why tuples are excluded, I was expecting one after finding the ones for lists and sets.

Issue #5383, opened 2023-04-05
`localns` arguments are not longer accepted.


commonism:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

I used localns to create nested models dynamically using pydanticv1.
Define all models, `….update_forward_refs({"ModelNameA":ModelA","ModelNameA":ModelA",})` for all of them in the proper order and it works.

How to without localns using pydanticv2?

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0a1
                 install path: ~/venv/openapi3/lib/python3.10/site-packages/pydantic
               python version: 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]
                     platform: Linux-5.19.0-35-generic-x86_64-with-glibc2.35
     optional deps. installed: ['email-validator', 'typing-extensions']
```

dmontagu:
Yeah, I realized we had eliminated this and was concerned someone might want the ability to specify the localns directly. I think we can re-add this into `model_rebuild`, I'll take a stab at it.

dmontagu:
@commonism the "fix" for this from #5385 is included now in the just-released `pydantic==2.0a2`. Please let us know if you have any further issues with this.

PR #5385, opened 2023-04-05
Add _types_namespace argument to model_rebuild

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5383

I called the argument `_types_namespace` instead of `_types_namespace` to try to discourage people from using it (like `_parent_namespace_depth`, you probably shouldn't change the value unless you know what you are doing).

But I would be open to changing this (and `_parent_namespace_depth`) to take the more standard approach of not including a leading `_` if anyone prefers.

@commonism If the unit test I added here isn't representative of your usage (e.g., if you use `create_model` to create the classes you mentioned in #5383) could you share a more representative unit test and/or some code that shows how you use this functionality? Want to make sure this change actually fixes things for you (and it doesn't get re-broken in the future).

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -389,6 +389,7 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         force: bool = False,
         raise_errors: bool = True,
         _parent_namespace_depth: int = 2,
+        _types_namespace: dict[str, Any] | None = None,
     ) -> bool | None:
         """
         Try to (Re)construct the model schema.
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

adriangb:
This looks good to me and I like the `_`. Let's see if this solves the original use case.

dmontagu:
@commonism just let us know if think you'll need any other changes, if it looks good to you we can merge it.

commonism:
Hi,

I think this will do - for now.
For the unit tests - I had to create with `types.new_class` - pydantic.create_model did not work for __root__ in v1.
https://github.com/commonism/aiopenapi3/blob/master/aiopenapi3/model.py#L146

I can not run my own unit tests yet, I'll keep you posted.

dmontagu:
Just an fyi — we removed `__root__` models in v2, though there is a migration path. Hopefully that resolves issues with `create_model`. Happy to help work through such issues if you have them though.

pydantic/main.py
```diff
@@ -403,6 +404,8 @@ def model_rebuild(
             types_namespace = cls.__pydantic_parent_namespace__
 
             types_namespace = _model_construction.get_model_types_namespace(cls, types_namespace)
+            if _localns:
+                types_namespace = {**types_namespace, **_localns}
```
commonism, line 408: I'd prefer this to be _localns only, types_namespace includes locals() or something and is not what I'd expect.
It would still be possible to have the `{**types_namespace, **_localns}` behaviour by passing 
`{**types_namespace, **_localns}` as _localns.

dmontagu: okay, that makes sense, so this would be more of a hard-override

adriangb (approved):
Looks good to me!

PR #5386, opened 2023-04-05
Add __pydantic_init_subclass__ classmethod

### LINKED ISSUE #5369: No way to access fields during __init_subclass__
### Initial Checks

### Description

There is no way to iterate through the `fields` from a class method anymore without instantiation

For example, if I want to iterate through all the fields at definition time to assert all of them have examples, I can't because there is no `cls.__fields__` anymore or no way to access the names of the fields from `cls`

I am more than likely missing something

### Example Code

```Python
from pydantic import BaseModel, Field, root_validator

ParentModelEnforcer(BaseModel):

   def __init_subclass__(cls, *args, **kwargs):
       # Here, now we have nothing to pass to `checker`
       _checker(cls, {})

   @root_validator()
   def _checker(cls, _values):
       # values is empty when called from __init_subclass__
       pass

ChildClass(ParentModelEnforcer):
   test_field_good: str = Field(..., example="pretty_test")
   test_field_bad: str
```

```Text
2.01a
```

dmontagu:
Closes #5369.

@samuelcolvin let me know if you think we should change this to be a non-dunder name (e.g., `model_init_subclass`). Should be trivial to change.

The only argument I see for keeping it as a dunder like this is to help draw users' attention to the fact that it really is meant to be similar to `__init_subclass__`. Otherwise I suspect it will be easier to skim over the definition without realizing that. But I see the argument either way.

(Presumably if/when we implement `BareModel` we'll use this name though.)

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -225,6 +225,9 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         This is necessary because `__init_subclass__` will always be called by `type.__new__`,
         and it would require a prohibitively large refactor to the `ModelMetaclass` to ensure that
         `type.__new__` was called in such a manner that the class would already be sufficiently initialized.
+
+        This will receive the same `kwargs` that would be passed to the standard `__init_subclass__`, namely,
+        any kwargs passed to the class definition that aren't existing config attributes.
         """
         pass
 
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -225,9 +225,6 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         This is necessary because `__init_subclass__` will always be called by `type.__new__`,
         and it would require a prohibitively large refactor to the `ModelMetaclass` to ensure that
         `type.__new__` was called in such a manner that the class would already be sufficiently initialized.
-
-        You can override this method in subclasses of `BaseModel` to perform any initialization logic,
-        in which case you should make it a `classmethod` and pass `cls` as the first argument.
         """
         pass
 
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -227,7 +227,7 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         `type.__new__` was called in such a manner that the class would already be sufficiently initialized.
 
         This will receive the same `kwargs` that would be passed to the standard `__init_subclass__`, namely,
-        any kwargs passed to the class definition that aren't existing config attributes.
+        any kwargs passed to the class definition that aren't used internally by pydantic.
         """
         pass
 
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -152,6 +152,10 @@ class ModelMetaclass(ABCMeta):
                 types_namespace,
                 raise_errors=False,
             )
+            # using super(cls, cls) on the next line ensures we only call the parent class's __pydantic_init_subclass__
+            # I believe the `type: ignore` is only necessary because mypy doesn't realize that this code branch is
+            # only hit for _proper_ subclasses of BaseModel
+            super(cls, cls).__pydantic_init_subclass__(**kwargs)  # type: ignore
             return cls
         else:
             # this is the BaseModel class itself being created, no logic required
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

pydantic/main.py
```diff
@@ -211,6 +215,19 @@ def __init__(__pydantic_self__, **data: Any) -> None:
     def __get_pydantic_core_schema__(cls, source: type[BaseModel], gen_schema: GenerateSchema) -> CoreSchema:
         return gen_schema.model_schema(cls)
 
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
```
samuelcolvin, line 219: I think it's a bit weird to include `kwargs` but omit the keyword arguments that `ModelMetaclass.__new__` takes by name.

I think this should be called with no arguments, but up for debate.

If we do pass `kwargs` please can we add something to the docstring.

samuelcolvin (commented):
otherwise LGTM.

please update.

pydantic/main.py
dmontagu: Well, for what it's worth, I just did it the same way that we handle the call to `super().__new__` (which will call `__init_subclass__` with the same kwargs) in line 109:

https://github.com/pydantic/pydantic/blob/0cd57b3e0c75d9622012de542e5ddc110fe55743/pydantic/main.py#L109

`__init_subclass__` will generally receives the keyword arguments that were passed in the class declaration:

```python
class MyClass:
    def __init_subclass__(cls, **kwargs):
        print(kwargs)

class MySubClass(MyClass, x=1):
    pass

#> {'x': 1}
```
which is why I wanted to pass them to `__pydantic_init_subclass__`.

I'm happy to add the other keyword arguments to the init_subclass call though; should we update the line 109 call too? (I think either way we probably should call with the same keyword arguments that the actual `__init_subclass__` receives.)

dmontagu: I'll note that there are tests that explicitly check the behavior of init_subclass with kwargs (and break if you remove the **kwargs from the call to `super().__new__` in the metaclass). So I think at most we may want to add the generic/namespace stuff into the `__init_subclass__` call.

I'm not sure what the right thing to do is (i.e., including the generic stuff and `__pydantic_reset_parent_namespace__`); happy to do it either way.

dmontagu: I confirmed we actually can't pass any of those keyword arguments to the standard `__init_subclass__`, so I'm not not inclined to pass them to `__pydantic_init_subclass__` either. Based on the way `__init_subclass__` works my impression is that the right thing to do here is to not pass those things to `__pydantic_init_subclass__` (especially considering you'll basically have access to them anyway since the `cls` will be initialized). If you still disagree I think maybe we should discuss over voice

dmontagu:
@samuelcolvin please review

* In this thread: https://github.com/pydantic/pydantic/pull/5386#discussion_r1160150907 I make the point that I think we should keep the kwargs as is (and try to describe why)
* I've added a comment about the kwargs to the `__pydantic_init_subclass__` docstring.

pydantic/main.py
samuelcolvin: okay, agreed.

pydantic/main.py
```diff
@@ -211,6 +215,22 @@ def __init__(__pydantic_self__, **data: Any) -> None:
...
+        This is intended to behave just like `__init_subclass__`, but is called by ModelMetaclass
+        only after the class is actually fully initialized. In particular, attributes like `model_fields` will
+        be present when this is called.
+
+        This is necessary because `__init_subclass__` will always be called by `type.__new__`,
+        and it would require a prohibitively large refactor to the `ModelMetaclass` to ensure that
+        `type.__new__` was called in such a manner that the class would already be sufficiently initialized.
+
+        This will receive the same `kwargs` that would be passed to the standard `__init_subclass__`, namely,
+        any kwargs passed to the class definition that aren't existing config attributes.
```
samuelcolvin, line 230:
```suggestion
        any kwargs passed to the class definition that aren't used internally by pydantic.
```

Since I think `config` (as in `ConfigDig`) kwargs are passed, but if I'm wrong, happy to revert.

samuelcolvin (approved):
but otherwise LGTM.

please update.

pydantic/main.py
dmontagu: What I meant was — "that aren't valid config fields" or similar, I realize what I wrote wasn't quite saying this. Which I think is more specific than your suggestion, _but_ I also think your suggestion is fine / potentially more "future-proof". So I committed your change.

PR #5390, opened 2023-04-05
Fix root validators & subclasses

### LINKED ISSUE #5388: `root_validator` methods are not called with subclasses as the first argument (always have the cls on which they were defined)
### Initial Checks

### Description

I believe functions decorated `root_validator` are supposed to act like class methods, receiving the class as the first argument, and I would expect that to work with subclasses.  For example:

```python
In [3]: class A:
   ...:     @classmethod
   ...:     def method(cls):
   ...:         print(cls)
   ...:
   ...: class B(A):
   ...:     ...
   ...:

In [4]: A.method()
<class '__main__.A'>

In [5]: B.method()
<class '__main__.B'>
```

However, `root_validator` methods *always* appear to receive the class on which they were defined, and are not inherited... see example below.

### Example Code

```Python

class Base(BaseModel):

    @root_validator(pre=True)
    def _some_validator(cls, values):
        print("cls", cls)
        return values

class Sub(Base):
    a: int = 1

Sub()
```

on pydantic v1, the above code snippet would print `'cls <class '__main__.Sub'>'`.  On v2 it prints `'cls <class '__main__.Base'>'`

```Text
╰─❯ python -c "import pydantic.version; print(pydantic.version.version_info())"
             pydantic version: 2.0a1
                 install path: /Users/talley/dev/forks/pydantic/pydantic
               python version: 3.10.6 | packaged by conda-forge | (main, Aug 22 2022, 20:41:54) [Clang 13.0.1 ]
                     platform: macOS-11.7.4-x86_64-i386-64bit
```

on `main` commit 0cd57b3e0c75d9622012de542e5ddc110fe55743
```

### PR DESCRIPTION
**Pull request by adriangb**
Fixes #5388.

The most important point is that this PR fundamentally changes what our decorators mean semantically. Currently they're a mix of "call this function by reference" and "call whatever is under this attribute name on this class". The second option is more in line with how regular python classes and inheritance work, I think that's the right path to go down and that's what this PR adopts. This has implications for duplicate validator detection / allow re-use which we can handle later.

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -237,9 +237,11 @@ class PydanticDecoratorMarker(Generic[ReturnType], Representation):
     def __get__(
         self, obj: object | None, objtype: type[object] | None = None
     ) -> Callable[..., ReturnType] | PydanticDecoratorMarker[ReturnType]:
-        if obj is None:
-            return self
-        return self.wrapped.__get__(obj, objtype)
+        try:
+            return self.wrapped.__get__(obj, objtype)
+        except AttributeError:
+            # not a descriptor, e.g. a partial object
+            return self.wrapped  # type: ignore[return-value]
 
 
 DecoratorInfoType = TypeVar('DecoratorInfoType', bound=DecoratorInfo)
@@ -259,13 +261,11 @@ class Decorator(Generic[DecoratorInfoType], Representation):
         cls_var_name: str,
         func: Callable[..., Any],
         shim: Callable[[Any], Any] | None,
-        unwrapped_func: Callable[..., Any],
         info: DecoratorInfoType,
     ) -> None:
         self.cls_ref = cls_ref
         self.cls_var_name = cls_var_name
         self.func = func
-        self.unwrapped_func = unwrapped_func
         self.info = info
         self.shim = shim
 
@@ -276,17 +276,7 @@ class Decorator(Generic[DecoratorInfoType], Representation):
         shim: Callable[[Any], Any] | None,
         info: DecoratorInfoType,
     ) -> Decorator[DecoratorInfoType]:
-        val = getattr(cls_, cls_var_name)
-        if isinstance(val, PydanticDecoratorMarker):
-            val = val.wrapped
-        try:
-            func = val.__get__(None, cls_)
-        except AttributeError:
-            if isinstance(val, partial):
-                # don't bind the class
-                func = val
-            else:
-                raise
+        func = getattr(cls_, cls_var_name)
         if shim is not None:
             func = shim(func)
         return Decorator(
@@ -294,7 +284,6 @@ class Decorator(Generic[DecoratorInfoType], Representation):
             cls_var_name=cls_var_name,
             func=func,
             shim=shim,
-            unwrapped_func=val,
             info=info,
         )
 
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -279,7 +279,14 @@ class Decorator(Generic[DecoratorInfoType], Representation):
         val = getattr(cls_, cls_var_name)
         if isinstance(val, PydanticDecoratorMarker):
             val = val.wrapped
-        func = val.__get__(None, cls_)
+        try:
+            func = val.__get__(None, cls_)
+        except AttributeError:
+            if isinstance(val, partial):
+                # don't bind the class
+                func = val
+            else:
+                raise
         if shim is not None:
             func = shim(func)
         return Decorator(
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: pydantic/_internal/_decorators.py [hunk 1]
```diff
@@ -264,15 +265,37 @@ def __init__(
...
+        self.shim = shim
 
-
-AnyDecorator = Union[
-    Decorator[ValidatorDecoratorInfo],
-    Decorator[FieldValidatorDecoratorInfo],
-    Decorator[RootValidatorDecoratorInfo],
-    Decorator[FieldSerializerDecoratorInfo],
-    Decorator[ModelSerializerDecoratorInfo],
-]
```
adriangb, line 275: Was not being used

pydantic/_internal/_decorators.py
```diff
@@ -310,48 +333,51 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
...
-            res.validator.update(existing.validator)
-            res.field_validator.update(existing.field_validator)
-            res.root_validator.update(existing.root_validator)
-            res.field_serializer.update(existing.field_serializer)
-            res.model_serializer.update(existing.model_serializer)
+            res.validator.update({k: v.bind_to_cls(cls) for k, v in existing.validator.items()})
+            res.field_validator.update({k: v.bind_to_cls(cls) for k, v in existing.field_validator.items()})
+            res.root_validator.update({k: v.bind_to_cls(cls) for k, v in existing.root_validator.items()})
+            res.field_serializer.update({k: v.bind_to_cls(cls) for k, v in existing.field_serializer.items()})
+            res.model_serializer.update({k: v.bind_to_cls(cls) for k, v in existing.model_serializer.items()})
```
adriangb, line 343: This is the key here. It ensures we get the value from the child instance if it gets overriden

pydantic/_internal/_decorators.py
```diff
@@ -264,15 +265,37 @@ def __init__(
...
+    @staticmethod
+    def build(
+        cls_: Any,
+        cls_var_name: str,
+        shim: Callable[[Any], Any] | None,
+        info: DecoratorInfoType,
+    ) -> Decorator[DecoratorInfoType]:
+        val = getattr(cls_, cls_var_name)
+        if isinstance(val, PydanticDecoratorMarker):
+            val = val.wrapped
```
adriangb, line 279: This allows the child to override without using the decorator again. But maybe we want to force users to explicitly apply the decorator?

tests/test_json.py
```diff
@@ -233,22 +233,6 @@ def serialize_timedt(self, _v: timedelta, _info):
...
-        @field_serializer('dt')
-        def serialize_dt(self, _v: datetime, _info):
-            return 'parent_encoder'
-
-    class Child(Parent):
-        @field_serializer('dt')
-        def serialize_dt(self, _v: datetime, _info):
-            return 'child_encoder'
-
-    assert Child().model_dump_json() == '{"dt":"child_encoder"}'
```
adriangb, line 249: Moved and modified this test to `test_serialize.py`

adriangb:
please review. I'll fix whatever is wrong with 3.11 tomorrow / before merging.

adriangb:
okay all fixed, please review

pydantic/_internal/_decorators.py
```diff
@@ -312,41 +331,35 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:  # noqa: C901
...
-            cls_ref = get_type_ref(cls)
-            try:
-                func = var_value.wrapped.__get__(None, cls)
-            except AttributeError:
-                if isinstance(var_value.wrapped, partial):
-                    # don't bind the class
-                    func = var_value.wrapped
-                else:
-                    raise
-            shimmed_func = var_value.shim(func) if var_value.shim is not None else func
```
adriangb, line 335: This logic got moved into `Decorator.build` so that it can be re-used

tests/test_serialize.py
```diff
@@ -522,3 +522,36 @@ class MyModel(BaseModel):
         ser = field_serializer('x', json_return_type='str')(partialmethod(func, expected=1234))
 
     assert MyModel(x=1234).model_dump() == {'x': '1,234'}
+
+
+def test_serializer_inheritance_override():
```
hramezani, line 527: There is a similar [test](https://github.com/pydantic/pydantic/blob/main/tests/test_serialize.py#L430:L444) from [my previous PR related to the multiple serializers function for a field](https://github.com/pydantic/pydantic/commit/0cd57b3e0c75d9622012de542e5ddc110fe55743). I think this one also should fail. Am I wrong?

adriangb: This is that test, reworked

hramezani (approved):

PR #5398, opened 2023-04-06
Rework allow_reuse in decorators

adriangb:
Closes #4436.

This implements pretty much the same behavior that pyright and ruff/flake8 (via `F811`) enforce.

### DIFF
--- a/pydantic/decorators.py
+++ b/pydantic/decorators.py
@@ -134,6 +134,8 @@ def validator(
     :param check_fields: whether to check that the fields actually exist on the model
     :param allow_reuse: whether to track and raise an error if another validator refers to the decorated function
     """
+    if allow_reuse is True:  # pragma: no cover
+        warn('`allow_reuse` is deprecated and will be ignored', DeprecationWarning)
     fields = tuple((__field, *fields))
     if isinstance(fields[0], FunctionType):
         raise PydanticUserError(
@@ -163,7 +165,6 @@ def validator(
             raise PydanticUserError(
                 '`@validator` cannot be applied to instance methods', code='validator-instance-method'
             )
-        _decorators.check_for_duplicate_decorator_function(f, allow_reuse=allow_reuse, type='validator')
         # auto apply the @classmethod decorator
         f = _decorators.ensure_classmethod_based_on_signature(f)
         wrap = _decorators.make_generic_v1_field_validator
@@ -186,7 +187,6 @@ def field_validator(
     mode: Literal['before', 'after', 'plain'] = ...,
     check_fields: bool | None = ...,
     sub_path: tuple[str | int, ...] | None = ...,
-    allow_reuse: bool = False,
 ) -> Callable[[_V2BeforeAfterOrPlainValidatorType], _V2BeforeAfterOrPlainValidatorType]:
     ...
 
@@ -198,7 +198,6 @@ def field_validator(
     mode: Literal['wrap'],
     check_fields: bool | None = ...,
     sub_path: tuple[str | int, ...] | None = ...,
-    allow_reuse: bool = False,
 ) -> Callable[[_V2WrapValidatorType], _V2WrapValidatorType]:
     ...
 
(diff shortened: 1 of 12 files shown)


tests/test_serialize.py
dmontagu, line 1: tests look good

pydantic/main.py
```diff
@@ -157,6 +166,10 @@ def hash_func(self_: Any) -> int:
             # this is the BaseModel class itself being created, no logic required
             return super().__new__(mcs, cls_name, bases, namespace, **kwargs)
 
+    @classmethod
+    def __prepare__(cls, *args: Any, **kwargs: Any) -> Mapping[str, object]:
+        return _ModelNamespaceDict()
```
dmontagu, line 171: this is very neat

pydantic/main.py
```diff
@@ -56,6 +56,15 @@
 _base_class_defined = False
 
 
+class _ModelNamespaceDict(dict):  # type: ignore[type-arg]
+    def __setitem__(self, k: str, v: object) -> None:
+        existing: Any = self.get(k, None)
+        if existing and v is not existing and isinstance(existing, _decorators.PydanticDecoratorMarker):
+            warnings.warn(f'`{k}` overrides an existing Pydantic `{existing.decorator_info.decorator_repr}` decorator')
```
dmontagu, line 63: Do we care about the case of `elif existing and v is not existing and isinstance(v, _decorators.PydanticDecoratorMarker)`? I guess seems okay to not cover it.

pydantic/decorators.py
```diff
@@ -298,6 +295,8 @@ def root_validator(
     Decorate methods on a model indicating that they should be used to validate (and perhaps modify) data either
     before or after standard model parsing/validation is performed.
     """
+    if allow_reuse is True:  # pragma: no cover
+        warn('`allow_reuse` is deprecated and will be ignored', DeprecationWarning)
```
dmontagu, line 299:
```suggestion
        warn('`allow_reuse` is deprecated and will be ignored; it should no longer be necessary', DeprecationWarning)
```
Can phrase this differently, but I think a good idea to tell people that their code will probably "just work" if they drop the `allow_reuse`, clarifying that it's not deprecated and will break their code.

dmontagu: (And same note above)

dmontagu (commented):
Looks like a big improvement. Couple notes but nothing major, I'd say merge when you feel ready

pydantic/main.py
adriangb: I don't think so. I don't want us to become a runtime linter lol, so I'm only covering the case of overriding a name with one of our decorators

dmontagu: lol sounds good

adriangb:
@dmontagu feedback addressed

pydantic/main.py
dmontagu: I guess the only thing I'll mention is that I think the original purpose of this was kind of runtime linting lol. So it might be a good idea to get Samuel's eyes on this and confirm he's in agreement. But I think we can merge with this decision and just change later.

adriangb: I mean it still does runtime linting, just in a limited fashion. cc @samuelcolvin

pydantic/main.py
```diff
@@ -56,6 +56,15 @@
 _base_class_defined = False
 
 
+class _ModelNamespaceDict(dict):  # type: ignore[type-arg]
```
samuelcolvin, line 59: please can we add a docstring explaining what this does?

samuelcolvin (commented):
looks great, thank you.

My only concern is whether we should keep the parameter and raise an error or warning, rather than remove it?

adriangb:
> My only concern is whether we should keep the parameter and raise an error or warning, rather than remove it?

I kept it for the V1 decorators and added a warning. But I removed it for the V2 decorators, I think a minor breaking change like that in an alpha is okay.

PR #5399, opened 2023-04-06
Use pyright for type checking; fix errors and add type ignores

adriangb:


pydantic/_internal/_dataclasses.py
```diff
@@ -76,9 +78,9 @@ def prepare_dataclass(
         warning_string = (
             f'`{name}` is not fully defined, you should define `{e}`, then call TODO! `methods.rebuild({name})`'
         )
-        if config['undefined_types_warning']:
+        if config['undefined_types_warning']:  # type: ignore
```
dmontagu, line 81: I don't love that we are now type-ignoring every access to `Config`. feels like we should either use `.get` or tweak the config implementation. We don't need to do that as part of this PR but we probably should do it sooner rather than later if we merge this.

pydantic/_internal/_decorators.py
```diff
@@ -383,7 +385,9 @@ def gather_decorator_functions(cls: type[Any]) -> DecoratorInfos:
 _FUNCS: set[str] = set()
 
 
-AnyDecoratorCallable: TypeAlias = 'Union[classmethod[Any], staticmethod[Any], partialmethod[Any], Callable[..., Any]]'
+AnyDecoratorCallable: TypeAlias = (
+    'Union[classmethod[Any, Any, Any], staticmethod[Any, Any], partialmethod[Any], Callable[..., Any]]'
```
dmontagu, line 389: we should confirm this doesn't break mypy for users, or at least be prepared to accept the fact that older mypy won't work. Maybe we can use the `if MYPY` trick, I think there's a way to do things differently based on mypy version info too (like sys.version_info)

pydantic/_internal/_fields.py
```diff
@@ -220,7 +221,7 @@ def collect_fields(  # noqa: C901
                 if not default.init:
                     # dataclasses.Field with init=False are not fields
                     continue
-                if DC_KW_ONLY and default.kw_only is True:
+                if DC_KW_ONLY and default.kw_only is True:  # type: ignore
```
dmontagu, line 224: what's this type error for?

pydantic/_internal/_fields.py
```diff
@@ -229,6 +230,7 @@ def collect_fields(  # noqa: C901
             # 2. To avoid false positives in the NameError check above
             try:
                 delattr(cls, ann_name)
+                cls = cast('type[BaseModel]', cls)
```
dmontagu, line 233: could you add a comment explaining how we know this cast is safe? I'm assuming it's related to the fact that `__pydantic_generic_x__` was present, but worth being clear

pydantic/_internal/_forward_ref.py
```diff
@@ -48,7 +49,7 @@ class PydanticForwardRef:
     """
 
     schema: core_schema.CoreSchema
-    model: type[BaseModel]
+    model: type[BaseModel] | type[PydanticDataclass]
```
dmontagu, line 52: any reason not to change this to a type[Any]? I can see these are basically the only types we use it on today, but I don't see any of the APIs depending on this being one of those two types. I might be missing something though.

pydantic/_internal/_generate_schema.py
```diff
@@ -151,23 +151,23 @@ def generate_config(config: ConfigDict, cls: type[Any]) -> core_schema.CoreConfi
...
+            str_strip_whitespace=config['str_strip_whitespace'],  # type: ignore
+            str_to_lower=config['str_to_lower'],  # type: ignore
+            str_to_upper=config['str_to_upper'],  # type: ignore
+            strict=config['strict'],  # type: ignore
+            ser_json_timedelta=config['ser_json_timedelta'],  # type: ignore
+            ser_json_bytes=config['ser_json_bytes'],  # type: ignore
+            from_attributes=config['from_attributes'],  # type: ignore
+            loc_by_alias=config['loc_by_alias'],  # type: ignore
+            revalidate_instances=config['revalidate_instances'],  # type: ignore
+            validate_default=config['validate_default'],  # type: ignore
```
dmontagu, line 170: all these type ignores seem pretty unfortunate, especially if you can't say that it's due to possibly-missing value, as opposed to invalid key

requirements/all.txt
```diff
@@ -185,12 +181,10 @@ tomli==2.0.1
     #   autoflake
     #   black
     #   coverage
-    #   mypy
     #   pytest
 typing-extensions==4.5.0
     # via
     #   dirty-equals
-    #   mypy
```
dmontagu, line 193: is mypy still in the testing dependencies?

pydantic/json_schema.py
```diff
@@ -461,6 +462,7 @@ def tagged_union_schema(self, schema: core_schema.TaggedUnionSchema) -> JsonSche
             if isinstance(v, (str, int)):
                 while isinstance(schema['choices'][v], (str, int)):
                     v = schema['choices'][v]
+                    assert isinstance(v, (int, str))
```
dmontagu, line 465: this is probably more efficient to do as a cast on the previous line, not that it really matters

pydantic/_internal/_validators.py
```diff
@@ -231,14 +231,14 @@ def __call__(  # noqa: C901 (ignore complexity)
                     {'multiple_of': self.multiple_of},
                 )
 
-        if self.gt is not None and not value > self.gt:
+        if self.gt is not None and not value > self.gt:  # type: ignore
```
dmontagu, line 234: maybe worth adding a note about why these type ignores are present (something about decimal IIRC?)

dmontagu (commented):
This might introduce some annoying conflicts with some of my open PRs but I'll deal with it.

pydantic/_internal/_repr.py
```diff
@@ -12,8 +12,10 @@
 from . import _typing_extra
 
 if typing.TYPE_CHECKING:
-    ReprArgs = typing.Iterable[tuple[str | None, Any]]
-    RichReprResult = typing.Iterable[Any | tuple[Any] | tuple[str, Any] | tuple[str, Any, Any]]
+    ReprArgs: typing_extensions.TypeAlias = 'typing.Iterable[tuple[str | None, Any]]'
+    RichReprResult: typing_extensions.TypeAlias = (
+        'typing.Iterable[Any | tuple[Any] | tuple[str, Any] | tuple[str, Any, Any]]'
+    )
```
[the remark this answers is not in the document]
dmontagu: Maybe the answer is just that typealias has to be a string

requirements/all.txt
dmontagu: Seems it should probably be in the all since it is necessary to run some of the tests

pydantic/_internal/_dataclasses.py
```diff
@@ -62,6 +62,8 @@ def prepare_dataclass(
             'Support for `__post_init_post_parse__` has been dropped, the method will not be called', DeprecationWarning
         )
 
+    cls = cast('type[PydanticDataclass]', cls)
```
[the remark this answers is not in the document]
adriangb: Yeah you get that because that attribute hasn['t been set. We're literally setting it a couple lines below. But without that cast pyright complains that you're assigning an unknown attribute.

pydantic/_internal/_dataclasses.py
adriangb: Totally agreed. This is just making explicit that what we're currently doing is questionable.

pydantic/_internal/_decorators.py
adriangb: We run mypy in tests, that kinda confirms it right?

pydantic/_internal/_fields.py
adriangb:
```
Cannot access member "kw_only" for type "Field[Unknown]
```

It's probably because they added it in Python 3.10 or something like that.

pydantic/_internal/_utils.py
```diff
@@ -14,13 +14,13 @@
...
-from typing_extensions import TypeGuard
+from typing_extensions import TypeAlias, TypeGuard
 
 from . import _repr, _typing_extra
 
 if typing.TYPE_CHECKING:
-    MappingIntStrAny = typing.Mapping[int | str, Any]
-    AbstractSetIntStr = typing.AbstractSet[int | str]
+    MappingIntStrAny: TypeAlias = 'typing.Mapping[int, Any] | typing.Mapping[str, Any]'
+    AbstractSetIntStr: TypeAlias = 'typing.AbstractSet[int] | typing.AbstractSet[str]'
```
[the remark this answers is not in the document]
adriangb: What pyright was complaining about was using `foo | bar` syntax on <3.10, even if it is the `if TYPE_CHECKING` block. So We need to use a `Union` or make it a string. And if you make it a string it needs to be a TypeAlias...

pydantic/json_schema.py
adriangb: asserts are nice in the sense that you can't lie about it, which is probably a good thing given that I'm kinda throwing these around willy nilly right now

pydantic/_internal/_validators.py
adriangb: This one was quite perplexing: 
> Operator ">" not supported for types "<subclass of int and Decimal> | <subclass of float and Decimal> | <subclass of str and Decimal> | Decimal" and "int | Decimal"

I can add that, but to be honest it doesn't even make much sense to me. This is probably a good candidate for exploring more later. We should come back in the future at some sprint and try to clean up type ignores

pydantic/_internal/_fields.py
dmontagu: This is a bit concerning to me, because it makes me think that pyright is treating our `FieldInfo` as a `dataclasses.Field`, due to the `@dataclass_transform` decorator and the `field_specifiers=Field`. I could be wrong about this though.

If I'm right about this, I think this might be one of the more potentially-risky sources of missed type checks. If we continue to run mypy in CI (even in non-recursive mode), this should be okay. (Or at least there's no regression)

pydantic/_internal/_decorators.py
dmontagu: Assuming one of the mypy tests covers this. I think it's probably fine, but I guess we can deal with it if/when people report issues. (And can tell them to upgrade mypy I guess lol.)

pydantic/json_schema.py
```diff
@@ -554,12 +558,13 @@ def _named_required_fields_schema(
...
             if self.by_alias:
-                alias = field.get('validation_alias', name)
+                alias: Any = field.get('validation_alias', name)
                 if isinstance(alias, str):
                     name = alias
                 elif isinstance(alias, list):
+                    alias = cast('list[str] | str', alias)
                     for path in alias:
-                        if isinstance(path, list) and len(path) == 1 and isinstance(path[0], str):
+                        if isinstance(path, list) and len(cast('list[Any]', path)) == 1 and isinstance(path[0], str):
```
dmontagu, line 567:
```suggestion
                        if isinstance(path, list) and len(path) == 1 and isinstance(path[0], str):
```
mypy gives a redundant cast error on this line, but I don't seem to get a pyright error if I remove the cast. So maybe better to remove it?

pydantic/_internal/_forward_ref.py
```diff
@@ -10,6 +10,7 @@
 
 if TYPE_CHECKING:
     from pydantic import BaseModel
+    from pydantic.dataclasses import PydanticDataclass
```
dmontagu, line 13:
```suggestion
    from pydantic._internal._dataclasses import PydanticDataclass
```
With the existing import, mypy gives the error:
```
pydantic/_internal/_forward_ref.py:13: error: Module "pydantic.dataclasses" does not explicitly export attribute "PydanticDataclass"  [attr-defined]
```
Minor thing, but I think it's right, and I don't see any reason not to do it this way.

dmontagu:
For what it's worth, I just checked, and with the old `fast.toml` mypy configuration, and mypy==1.2.0, there are only two type errors raised, and they are the two that I pointed out in comments (with suggestions to fix) above. And fixing them doesn't cause pyright errors.

pydantic/dataclasses.py
```diff
@@ -37,13 +37,13 @@ def dataclass(
         config: ConfigDict | type[object] | None = None,
         validate_on_init: bool | None = None,
         kw_only: bool = ...,
-    ) -> Callable[[type[_T]], type[PydanticDataclass]]:
```
dmontagu, line 40: This `Any` and the one below are the sources of the `Expression type contains "Any" (has type overloaded function)` mypy errors, and I suspect are why it weirdly adds the _T instead of Any.

Seems like a bit of an abuse of mypy, but considering it's a user-facing error, not just an us-facing error, I think it's worth just doing a # type: ignore for pyright

pydantic/_internal/_dataclasses.py
samuelcolvin: we absolutely need to change how config works, I'll work on that today.

.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
     pass_filenames: false
   - id: typecheck
     name: Typecheck
-    entry: make typecheck-fast
-    types: [python]
-    language: system
+    entry: pyright pydantic
```
samuelcolvin, line 21: surely we should just have one `make typecheck` which switches to use `pyright`?

samuelcolvin: okay, I see you've done that. I would prefer to stick with `language: system` and running the `make` command.

.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
     pass_filenames: false
   - id: typecheck
     name: Typecheck
-    entry: make typecheck-fast
-    types: [python]
```
samuelcolvin, line 22: `types` should remain.

pydantic/_internal/_fields.py
samuelcolvin: this is a `dataclasses.Field`, see the `if isinstance(default, dataclasses.Field):` line above.

I think this is fine.

pydantic/_internal/_forward_ref.py
samuelcolvin: Let's keep it stricter for now, easy to relax it - maybe add a comment to that effect.

pydantic/_internal/_validators.py
samuelcolvin: can we add this as a comment to help future travellers passing this way.

samuelcolvin (commented):
Shame about quite how many ignores there are, even accounting for the `ConfigDict` thing which was a mistake on my part.

Still on balance I think we go ahead with pyright.

pydantic/_internal/_model_construction.py
```diff
@@ -286,18 +288,18 @@ def __getattr__(self, item: str) -> None:
         __tracebackhide__ = True
         # raise an AttributeError if `item` doesn't exist
         getattr(SchemaValidator, item)
-        raise PydanticUserError(self._error_message, code=self._code)
+        raise PydanticUserError(self._error_message, code=cast(PydanticErrorCodes, self._code))
```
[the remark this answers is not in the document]
adriangb: great thanks!

pydantic/_internal/_model_construction.py
```diff
@@ -286,18 +288,18 @@ def __getattr__(self, item: str) -> None:
...
-    alias_generator = config['alias_generator']
+    alias_generator = config['alias_generator']  # type: ignore
     if alias_generator is None:
         return
 
     for name, field_info in fields.items():
         if field_info.alias_priority is None or field_info.alias_priority <= 1:
             alias = alias_generator(name)
-            if not isinstance(alias, str):
+            if not isinstance(alias, str):  # type: ignore
```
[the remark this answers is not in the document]
adriangb: Unnecessary `isinstance` call

pydantic/json_schema.py
```diff
@@ -256,13 +257,13 @@ def bool_schema(self, schema: core_schema.BoolSchema) -> JsonSchemaValue:
     def int_schema(self, schema: core_schema.IntSchema) -> JsonSchemaValue:
         json_schema = {'type': 'integer'}
         self.update_with_validations(json_schema, schema, self.ValidationsMapping.numeric)
-        json_schema = {k: v for k, v in json_schema.items() if v not in {math.inf, -math.inf}}
+        json_schema = {k: v for k, v in json_schema.items() if v not in {math.inf, -math.inf}}  # type: ignore
```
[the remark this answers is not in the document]
adriangb: `Expression will always evaluate to True since the types "str" and "float" have no overlap`, seems like `v` is a string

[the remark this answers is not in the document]
adriangb: `json_schema: dict[str, Any] = {'type': 'integer'}` fixed it

pydantic/_internal/_validators.py
adriangb: done

.pre-commit-config.yaml
adriangb: pyright is a node package. if we do that we'd have to somehow manage the node package ourselves. if we are already using pre-commit which can handle node packages, why not use that?

.pre-commit-config.yaml
samuelcolvin: well, we have to handle it anyway to support `pyright` call from `Makefile` or otherwise using pyright without pre-commit.

If it's really easier, happy to switch to just `pre-commit`, if so we need to call `pyright` via `pre-commit` in `Makefile` and anywhere else.

adriangb: That is exactly what I'm doing: I'm calling `pre-commit run typecheck` from `make typecheck`

.pre-commit-config.yaml
adriangb: Sorry, I was wrong: that is what I was intending to do but had not done it.

.pre-commit-config.yaml
samuelcolvin: okay, great.

is it easy to pass other arguments to pyright with `pre-commit` from the command line? Don't know what they are yet, but I could imagine we might want to.

adriangb: I’m not sure. But I propose we cross that bridge when we get there.

samuelcolvin: great, seems good to me.

samuelcolvin (approved):
please merge when you're happy.

PR #5410, opened 2023-04-07
Remove unused fixture and args in test_serializer.py

adriangb:
- Remove an unnecessary fixture
- Use `@wraps` on serializer shims to get the right func name in error messages

### DIFF
--- a/tests/test_serialize.py
+++ b/tests/test_serialize.py
@@ -22,16 +22,6 @@ from pydantic.annotated_arguments import PlainSerializer, WrapSerializer
 from pydantic.config import ConfigDict
 
 
-@pytest.fixture
-def reset_tracked_serializer():
-    from pydantic._internal._decorators import _FUNCS
-
-    original_tracked_serializers = set(_FUNCS)
-    yield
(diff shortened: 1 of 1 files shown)


adriangb:
please review

dmontagu:
> Use @wraps on serializer shims to get the right func name in error messages

doesn't look like there are any tests for this, worth adding one?

adriangb:
There’s a test that fails if that’s not there, but I can add a more explicit one

samuelcolvin:
Any idea what overhead is of wraps and another wrapper function? If possibly I'd prefer not to.

I think it would be less work and more performant to just support no `info` argument in core, and use the vanilla function where possible.

adriangb:
I don’t think it adds a function call. I think it just monkey patches attributes from the original function onto the new function (`__name__`, `__doc__`, etc)

adriangb:
It would be more performant to implement this in core but I seriously doubt it matters in anything but micro benchmarks. And the price to pay is more complexity in rust. Where is less debuggable and harder for contributors to follow and understand.

PR #5411, opened 2023-04-07
new `ConfigWrapper`

### LINKED ISSUE #5353: New Config machinery
The current "dict that's not a dict" thing is both ugly, not very typing save and results in some duplication.

We need a new `ConfigPlumbing` class that:
* holds the `config` dict
* holds the logic for handling deprecated `Config` classes
* holds the logic for merging config dicts
* holds the logic for converting it to a `CoreConfig`
* provides a way to access the default value even if the actual config is partial

We might also need to add a function to pydantic-core which returns a total `CoreConfig` with all the defaults set correctly as a single source of truth on what config defaults are.

samuelcolvin:
Fix #5353

Fix https://github.com/pydantic/pydantic-core/issues/425 - slightly amazingly, thanks for @dmontagu for leaving such a helpful comment.


Makefile
```diff
@@ -64,7 +64,7 @@ test-fastapi:
 	./tests/test_fastapi.sh
 
 .PHONY: all
-all: lint typecheck-fast testcov
+all: lint typecheck testcov
```
adriangb, line 67: Thanks sorry I missed this

docs/usage/models.md
```diff
@@ -853,7 +853,7 @@ try:
 except ValidationError as e:
     print(e)
     """
-    2 validation errors for InnerT[int]
+    2 validation errors for OuterT[int]
```
adriangb, line 856: Are these bugs that are getting fixed?

pydantic/_internal/_config.py
```diff
@@ -0,0 +1,244 @@
...
+                    return config_defaults[name]
+                except KeyError:
+                    raise AttributeError(f'Config has no attribute {name!r}') from None
+
+    def core_config(self, obj: Any = None) -> core_schema.CoreConfig:
+        """
+        Create a pydantic-core config, `obj` is just used to populate `title` if not set in config.
+
+        We don't use getattr here since we don't want to populate with defaults.
+        """
```
adriangb, line 116: Is there a better way to handle `obj`? Could this function accept a fallback instead?

adriangb (approved):
This looks great to me!

docs/usage/unions.md
```diff
@@ -205,7 +205,7 @@ try:
 except ValidationError as e:
     print(e)
     """
-    1 validation error for Dog
+    1 validation error for Model
```
adriangb, line 208: Same with these, does this have to do with the title in core schema changing?

docs/usage/models.md
samuelcolvin: yes https://github.com/pydantic/pydantic-core/issues/425

adriangb: nice!

pydantic/_internal/_config.py
samuelcolvin: I think this is fine for now.

samuelcolvin:
@dmontagu merge when/if you're happy.

dmontagu (approved):

PR #5415, opened 2023-04-07
fix type checking for python 3.7 and simplify ForwardRef wrapper

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_generics.py
+++ b/pydantic/_internal/_generics.py
@@ -75,7 +75,7 @@ else:
 
 if TYPE_CHECKING:
 
-    class DeepChainMap(ChainMap[KT, VT]):
+    class DeepChainMap(ChainMap[KT, VT]):  # type: ignore
         ...
 
 else:
--- a/pydantic/_internal/_typing_extra.py
+++ b/pydantic/_internal/_typing_extra.py
@@ -260,7 +260,7 @@ def get_cls_type_hints_lenient(obj: Any, globalns: dict[str, Any] | None = None)
                 if value is None:
                     value = NoneType
                 elif isinstance(value, str):
-                    value = make_forward_ref(value, is_argument=False, is_class=True)  # type: ignore
+                    value = _make_forward_ref(value, is_argument=False, is_class=True)  # type: ignore
 
                 try:
                     hints[name] = typing._eval_type(value, globalns, localns)  # type: ignore[attr-defined]
@@ -272,7 +272,7 @@ def get_cls_type_hints_lenient(obj: Any, globalns: dict[str, Any] | None = None)
 
 if sys.version_info < (3, 9):
 
-    def make_forward_ref(
+    def _make_forward_ref(
         arg: Any,
         is_argument: bool = True,
         *,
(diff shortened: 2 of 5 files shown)


pydantic/_internal/_generics.py
```diff
@@ -75,7 +75,7 @@ def __class_getitem__(cls, *args: Any) -> Any:
 
 if TYPE_CHECKING:
 
-    class DeepChainMap(ChainMap[KT, VT]):
+    class DeepChainMap(ChainMap[KT, VT]):  # type: ignore
```
adriangb, line 78: I'm not sure why Pyright complains about this since it's inside of a TYPE_CHECKING block, but the fix is easy enough...

pydantic/_internal/_typing_extra.py
```diff
@@ -281,18 +286,17 @@ def ForwardRefWrapper(arg: Any, is_argument: bool = True, *, is_class: bool = Fa
...
         try:
-            return typing.ForwardRef(arg, is_argument, is_class=is_class)
+            res = typing.ForwardRef(arg, is_argument, is_class=is_class)  # type: ignore
+            make_forward_ref = typing.ForwardRef  # type: ignore
+            return res
         except TypeError:
-            fr_has_is_class = False
             return typing.ForwardRef(arg, is_argument)
 
-    ForwardRef = ForwardRefWrapper  # noqa F811
```
adriangb, line 295: Pyright was not happy with this. It's quite strict as to what is a type and not, and assigning a function to a type is not okay with it. This change makes it explicit that we have a wrapper and not a type. It also simplifies the memory logic by just overriding itself on the first call if it can.

adriangb:
please review

pydantic/_internal/_typing_extra.py
dmontagu, line 1: The main issue I have — in principle — with changing this file like this is that (... if I recall correctly ...) many of the functions in this file are taken verbatim from python standard library. The upshot of that is that if these functions are changed for some reason in future versions of python, it should be more straightforward to update them, as the amount of changes necessary besides copy/pasting is minimal. That's why I made `ForwardRef = ForwardRefWrapper` below — this allows the functions here to be that much less modified from their original sources.

On the other hand, I will  admit that this diff isn't exactly large. The biggest issue to me is that it might be easy for whoever ends up working on it (which might not be you, me, or Samuel) to not realize "oh I need to replace `ForwardRef` with `_make_forward_ref` everywhere.

One solution, perhaps, is to add an explicit note about this at the top of this file. Or maybe it's just reasonable to not say anything and we just deal with it in the future if/when necessary. Anyway, if @samuelcolvin is good with this then I am.

dmontagu:
Other than the previous comment about the _make_forward_ref changes, I'm good with this, and happy for you to merge (whether or not any changes are made there).

pydantic/_internal/_typing_extra.py
adriangb: That makes sense. On the other hand tests would fail and looking at the history it would be pretty easy to spot what's going on. I'll add a comment here or there explaining and otherwise merge as is.

Issue #5406, opened 2023-04-07
Clean up `Field` parameters


samuelcolvin:
As per #5405:
* every parameter from V1 should be included
* all unused parameters should either:
  * raise an error, or
  * raise a warning and set the new parameter
* `**kwargs` should be included, but setting anything should raise an error

dmontagu:
To be clear — I'm assuming that `**kwargs` should NOT be part of the type-hinted signature used by IDEs/type-checkers, just, we should provide a better error than "missing keyword argument". Is that right?

PR #5425, opened 2023-04-10
Clean up Field parameters

hramezani:
## Change Summary
Clean up Field parameters

## Related issue number
Fixes https://github.com/pydantic/pydantic/issues/5406

## Checklist

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -11,6 +11,7 @@ import typing_extensions
 from . import types
 from ._internal import _fields, _forward_ref, _repr, _typing_extra, _utils
 from ._internal._fields import Undefined
+from .errors import PydanticUserError
 
 if typing.TYPE_CHECKING:
     from dataclasses import Field as DataclassField
@@ -300,6 +301,8 @@ class FieldInfo(_repr.Representation):
                 continue
             elif s == 'final':
                 continue
+            if s == 'frozen' and self.frozen is False:
+                continue
             if s == 'default_factory' and self.default_factory is not None:
                 yield 'default_factory', _repr.PlainRepr(_repr.display_as_type(self.default_factory))
             else:
@@ -349,10 +352,10 @@ def Field(
 ) -> Any:
     """
     Used to provide extra information about a field, either for the model schema or complex validation. Some arguments
-    apply only to number fields (``int``, ``float``, ``Decimal``) and some apply only to ``str``.
+    apply only to number fields (`int`, `float`, `Decimal`) and some apply only to `str`.
 
     :param default: since this is replacing the field's default, its first argument is used
-      to set the default, use ellipsis (``...``) to indicate the field is required
+      to set the default, use ellipsis (`...`) to indicate the field is required
     :param default_factory: callable that will be called when a default value is needed for this field
       If both `default` and `default_factory` are set, an error is raised.
     :param alias: the public name of the field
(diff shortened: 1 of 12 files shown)


hramezani:
please review

pydantic/fields.py
```diff
@@ -375,16 +379,18 @@ def Field(
       of digits within the decimal. It does not include a zero before the decimal point or trailing decimal zeroes.
     :param decimal_places: only applies to Decimals, requires the field to have at most a number of decimal places
       allowed. It does not include trailing decimal zeroes.
-    :param min_items: only applies to lists, requires the field to have a minimum number of
+    :param min_items: deprecated, use ``min_length`` instead.
```
samuelcolvin, line 382: we want to use single back-ticks since this will be rendered as markdown - might as well fix the whole docstring now.

pydantic/fields.py
```diff
@@ -393,7 +399,32 @@ def Field(
...
+    :param const: removed, use ``Literal`` instead'
+    :param unique_items: removed, use ``Set`` instead
+    :param allow_mutation: deprecated, use ``frozen`` instead
+    :param regex: removed, use `Pattern` instead
+    :param **extra: removed
     """
+    # Check deprecated & removed params of V1.
+    # This has to be removed deprecation period over.
+    if const:
+        raise ValueError('`const` is removed. use `Literal` instead')
```
samuelcolvin, line 411: Please use `PydanticUserError`, add a code and a section to docs if you think necessary, otherwise set `code=None`.

pydantic/fields.py
```diff
@@ -393,7 +399,32 @@ def Field(
...
+    if min_items:
+        warn('`min_items` is deprecated and will be removed. use `min_length` instead', DeprecationWarning)
+        min_length = min_items
```
samuelcolvin, line 414: `if min_length is None`.

pydantic/fields.py
```diff
@@ -393,7 +399,32 @@ def Field(
...
+    if max_items:
+        warn('`max_items` is deprecated and will be removed. use `max_length` instead', DeprecationWarning)
+        max_length = max_items
+    if unique_items:
+        raise ValueError('`unique_items` is removed. use `Set` instead')
```
samuelcolvin, line 419:
```suggestion
        raise ValueError('`unique_items` is removed, use `Set` instead')
```

Might be worth adding `(this feature is discussed in https://github.com/pydantic/pydantic-core/issues/296)`

https://github.com/pydantic/pydantic-core/issues/296

pydantic/fields.py
samuelcolvin: and below.

tests/test_annotated.py
```diff
@@ -22,22 +22,22 @@
         (
             lambda: Annotated[int, Field(gt=0)],
             5,
-            'FieldInfo(annotation=int, required=False, default=5, metadata=[Gt(gt=0)])',
+            'FieldInfo(annotation=int, required=False, default=5, metadata=[Gt(gt=0)], frozen=False)',
```
samuelcolvin, line 25: if we set the default of frozen to `False`, we should ignore `frozen=False` in repr.

samuelcolvin (commented):
otherwise LGTM.

Might be work asking @dmontagu about why he made the default of `frozen=None`.

please update.

pydantic/fields.py
hramezani: Fixed for all docstrings in code

pydantic/fields.py
hramezani: Done

pydantic/fields.py
hramezani: Good catch. done

tests/test_annotated.py
hramezani: ignored

hramezani:
please review

dmontagu:
> Might be work asking @dmontagu about why he made the default of frozen=None.

I don't think this was super intentional.

I think I may have been following the pattern used for `allow_inf_nan: bool | None = None,` as far as I can tell that is also a case where perhaps the default should be a boolean rather than `None`. Unless `None` has different behavior than one of the booleans, my inclination would be to change both cases to use `bool` only.

And given the following snippet from the docstring, I think `None` doesn't have special behavior there:
```python
:param allow_inf_nan: only applies to numbers, allows the field to be NaN or infinity (+inf or -inf),
        which is a valid Python float. Default True, set to False for compatibility with JSON.
```

pydantic/fields.py
```diff
@@ -335,65 +337,105 @@ def Field(
...
+            ),
+            code='field-removed-param',
+        )
+    if allow_mutation is False:
+        warn('`allow_mutation` is deprecated and will be removed. use `frozen` instead', DeprecationWarning)
+        frozen = True
+    if regex:
+        raise PydanticUserError('`regex` is removed. use `Pattern` instead', code='field-removed-param')
+    if extra:
+        raise PydanticUserError('Extra keyword arguments are not allowd on `Field`', code='field-removed-param')
```
samuelcolvin, line 437: * rename to `deprecated_kwargs`
* this has been replaced by `json_schema_extra`
* should therefore be a warning not an error.

samuelcolvin (commented):
otherwise LGTM.

please update.

hramezani:
please review

PR #5428, opened 2023-04-10
Clean up and add tests for extra and serialization

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/tests/test_deprecated.py
+++ b/tests/test_deprecated.py
@@ -1,7 +1,7 @@
 import platform
 import re
 from types import SimpleNamespace
-from typing import Dict, List
+from typing import Any, Dict, List, Type
 
 import pytest
 
@@ -9,7 +9,7 @@ from pydantic import BaseModel, ConfigDict, PydanticUserError, ValidationError, 
 from pydantic.config import Extra
 
 
-def deprecated_from_orm(model_type, obj):
+def deprecated_from_orm(model_type: Type[BaseModel], obj: Any) -> Any:
     with pytest.warns(
         DeprecationWarning,
         match=re.escape(
@@ -223,8 +223,7 @@ def test_properties():
     assert model.y == 5
 
 
-@pytest.mark.xfail(reason='working on V2')
-def test_extra_allow():
+def test_extra_allow_from_orm():
     class TestCls:
         x = 1
         y = 2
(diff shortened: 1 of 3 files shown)


adriangb:
please review

pydantic/_internal/_config.py
```diff
@@ -199,6 +199,7 @@ def prepare_config_extra(config: ConfigDict | dict[str, Any]) -> None:
     if extra is not None and not isinstance(extra, Extra):
         try:
             config['extra'] = Extra(extra)
+            warnings.warn('`extra` should be an instance of `pydantic.Extra`, not a string', DeprecationWarning)
```
dmontagu, line 202: Is there a reason we are using the `Extra` enum rather than `Literal['allow', 'forbid', 'ignore']`?

Personally, I find it pretty annoying that you have to add an import to specify this value. I recognize that currently there are various type hints indicating you need to use the `Extra` enum, and I imagine we need to keep it around for v1 compatibility, but I would prefer to reduce the use of that enum and replace it with a literal, rather than increase it.

This isn't an opinion I'd fight for though. And maybe I'm missing some other benefit to using the python Enum.

adriangb:
Hold until https://github.com/pydantic/pydantic-core/pull/537

pydantic/_internal/_config.py
adriangb: Removed the warning for now. I'm leaving in the use of the enum for now, partially because I'm too lazy to revert it, partially because that is what the type hint says now so it's more correct, at least for now. Let me know if you want me to revert that change.

dmontagu:
> Hold until [pydantic/pydantic-core#537](https://github.com/pydantic/pydantic-core/pull/537)

For what it's worth, I'm good with merging this now if you wanted, I recognize that will enable you to fix some more issues.

adriangb:
Let’s do that then

tests/test_edge_cases.py
```diff
@@ -1240,10 +1240,10 @@ class Model(BaseModel):
 
 def test_force_extra():
     class Model(BaseModel):
-        model_config = ConfigDict(extra='ignore')
+        model_config = ConfigDict(extra=Extra.ignore)
```
dmontagu, line 1243: Is there still a test that it coerces strings properly? I recognize other PRs are possibly changing this behavior, but I think that was the purpose of this test, and until we do change how `Extra` works, presumably we should still have a test that that works (for migration purposes or whatever; at least assuming it worked in v1, which I think it did). Even if it raises a deprecation warning.

adriangb: I’ll revert this change and rename the test

adriangb:
please review

dmontagu (approved):

PR #5431, opened 2023-04-10
Make Extra a literal

adriangb:
The only use case this breaks is if users were doing `Extra.allow.value` and such. I don't see any reason why they'd want to do that.

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -8,7 +8,7 @@ from typing_extensions import Literal, Self
 
 from pydantic.errors import PydanticUserError
 
-from ..config import ConfigDict, Extra
+from ..config import ConfigDict, ExtraValues
 
 DEPRECATION_MESSAGE = 'Support for class-based `config` is deprecated, use ConfigDict instead.'
 
@@ -30,7 +30,7 @@ class ConfigWrapper:
     str_strip_whitespace: bool
     str_min_length: int
     str_max_length: int | None
-    extra: Extra | None
+    extra: ExtraValues | None
     frozen: bool
     populate_by_name: bool
     use_enum_values: bool
@@ -118,7 +118,7 @@ class ConfigWrapper:
         core_config = core_schema.CoreConfig(
             **core_schema.dict_not_none(
                 title=self.config_dict.get('title') or (obj and obj.__name__),
-                extra_fields_behavior=extra and extra.value,
+                extra_fields_behavior=extra,
                 allow_inf_nan=self.config_dict.get('allow_inf_nan'),
                 populate_by_name=self.config_dict.get('populate_by_name'),
                 str_strip_whitespace=self.config_dict.get('str_strip_whitespace'),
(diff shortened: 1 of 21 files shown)


pydantic/config.py
```diff
@@ -1,17 +1,19 @@
...
 
 
-class Extra(str, Enum):
-    allow = 'allow'
-    ignore = 'ignore'
-    forbid = 'forbid'
+class Extra:
+    allow: Literal['allow'] = 'allow'
+    ignore: Literal['ignore'] = 'ignore'
+    forbid: Literal['forbid'] = 'forbid'
```
adriangb, line 13: We could easily add a deprecation warning on attribute access

adriangb:
TODO: fix mypy plugin if we decide to move forward with this

pydantic/config.py
dmontagu: Do we need to change this? I'm thinking we can just keep it as the enum for now and still add the deprecation warning on access. Just seems that much less likely to break things. I would give it a nonzero chance that someone is relying on the Enum-ness of it somewhere (e.g., using it as an argument to a FastAPI endpoint or something).

dmontagu: I guess if it were up to me, I would be more in favor of keeping it as an Enum or dropping it entirely than adding this, which I think as a user I would find a bit more surprising; admittedly it will keep more cases working without issue, but if it does cause errors, they will probably be harder/more surprising to debug than if the class was just missing.

If you can get it to say "this class is deprecated" when accessing the class attributes I think that's fine.

adriangb: Yup that’s what I was planning on doing.

samuelcolvin (commented):
I'm happy with this, but let's adding the deprecation warning.

samuelcolvin:
please update.

adriangb:
Added a deprecation warning. Please review.

hramezani:
LGTM!

hramezani:
LGTM!

pydantic/_internal/_config.py
```diff
@@ -8,7 +8,7 @@
 
 from pydantic.errors import PydanticUserError
 
-from ..config import ConfigDict, Extra
+from ..config import ConfigDict, ExtraValues
```
Kludex, line 11: Is there a reason for using both absolute and relative import here?

themattmorris:
I'm just curious what benefit was gained here that it merited this change?  Not a huge deal, but I personally prefer `extra = Extra.forbid` so I quickly know what members are available and I'm safe in the unlikely event that a member changes its value, and it's how I've been setting this configuration item.

I know `3.11` introduced some [enum breaking changes](https://github.com/python/cpython/issues/100458), but those can be resolved with `enum.StrEnum` when `>=3.11`.  I'm probably missing something?  Thanks!

adriangb:
Main reasons were:
- Unified API from a user’s perspective. We don’t have Enums for the other literals.
- Unified implementation from our perspective.

Your editor should give you autocomplete for the literal options just like you’d get for an enum.

We could also keep the Extra namespace, but I think we would have to do that for other literals as well then.

Issue #5426, opened 2023-04-10
Fix aliases


samuelcolvin:
See https://github.com/pydantic/pydantic/discussions/5393#discussioncomment-5569604 for context.

We need to:
* add `validation_alias` and `serialization_alias` to `Field`, if they're not set they generally default to `alias`
* not use `field.alias` for `serialization_alias`, [here](https://github.com/pydantic/pydantic/blob/c206df4e49a13937621420232629c9ef35483b29/pydantic/_internal/_generate_schema.py#L536) if it's not just a `str`
* not use `field.alias` to generate the signature [here](https://github.com/pydantic/pydantic/blob/c206df4e49a13937621420232629c9ef35483b29/pydantic/_internal/_model_construction.py#L242) if it's not a `str`

@hramezani might be good for you to work on this as you've just been working on `Field`.

dmontagu:
One thing I want to add possibly related to this:

Right now, it looks like `@dataclass_transform` uses the `alias` keyword argument to determine the name to use in the signature of the `__init__` method.

This is convenient in that it is the same behavior pydantic uses now unless you set `populate_by_name=True` in the config. However, the downside is that you can't change the behavior without using the plugin, and currently, if you do use the plugin, you don't have control over the fact that it will _always_ use the field name in the `__init__` signature.

I would propose that we provide some way to make it so the `validation_alias` is _not_ used as part of the signature. This would make it so that users could control what the generated signature is without being forced down some path as a function of their config settings. (To be clear, I'm more concerned with the mypy signature than the model_construction signature, but I guess it's potentially relevant to both.)

I could see doing this through another config flag or through a different keyword argument — the main thing to keep in mind is that `@dataclass_transform` will always set the `__init__` to use the alias if the plugin isn't in use, so I think it may be a good idea to provide a way of specifying keyword arguments besides `alias` that have the same effect.

samuelcolvin:
Well, with this change, the user could do

```py
class MyModel(BaseModel):
    foo: int = Field(alias='bar')
    # or, the same:
    foo: int = Field(validation_alias='bar', serialisation_alias='bar')
```

These are the same with pydantic, but different for dataclass transforms, does that help?

dmontagu:
Other than being a bit verbose, I think that covers most users' desired behavior.

If I'm thinking about it right, there are three things going on here:

* What alias is used for validation
* What alias is used for serialization
* What alias is used in the `__init__` signature

[...]

PR #5433, opened 2023-04-10
Fix aliases and add `validation_alias` and `serialization_alias` to `Field`

hramezani:
Fix aliases and add `validation_alias` and `serialization_alias` to `Field`
Fixes https://github.com/pydantic/pydantic/issues/5426

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -533,9 +533,7 @@ class GenerateSchema:
             schema,
             serialization_exclude=True if field_info.exclude else None,
             validation_alias=field_info.validation_alias,
-            serialization_alias=(
-                field_info.serialization_alias if isinstance(field_info.serialization_alias, str) else None
-            ),
+            serialization_alias=field_info.serialization_alias,
             frozen=field_info.frozen or field_info.final,
             metadata=metadata,
         )
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -532,8 +532,10 @@ class GenerateSchema:
         return _common_field(
             schema,
             serialization_exclude=True if field_info.exclude else None,
-            validation_alias=field_info.alias,
-            serialization_alias=field_info.alias,
+            validation_alias=field_info.validation_alias,
+            serialization_alias=(
+                field_info.serialization_alias if isinstance(field_info.serialization_alias, str) else None
+            ),
             frozen=field_info.frozen or field_info.final,
             metadata=metadata,
         )
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -237,7 +237,7 @@ def generate_model_signature(
     if var_kw:  # if custom init has no var_kw, fields which are not declared in it cannot be passed through
         allow_names = config_wrapper.populate_by_name
         for field_name, field in fields.items():
-            # alis have to be used for signature generation when it's a str
+            # when alias is a str it should be used for signature generation
             if isinstance(field.alias, str):
                 param_name = field.alias
             else:
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -237,6 +237,7 @@ def generate_model_signature(
     if var_kw:  # if custom init has no var_kw, fields which are not declared in it cannot be passed through
         allow_names = config_wrapper.populate_by_name
         for field_name, field in fields.items():
+            # alis have to be used for signature generation when it's a str
             if isinstance(field.alias, str):
                 param_name = field.alias
             else:
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -307,4 +307,6 @@ def apply_alias_generator(config: ConfigDict, fields: dict[str, FieldInfo]) -> N
             if not isinstance(alias, str):
                 raise TypeError(f'alias_generator {alias_generator} must return str, not {alias.__class__}')
             field_info.alias = alias
+            field_info.validation_alias = alias
+            field_info.serialization_alias = alias
             field_info.alias_priority = 1
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -237,7 +237,10 @@ def generate_model_signature(
     if var_kw:  # if custom init has no var_kw, fields which are not declared in it cannot be passed through
         allow_names = config_wrapper.populate_by_name
         for field_name, field in fields.items():
-            param_name = field.alias or field_name
+            if isinstance(field.alias, str):
+                param_name = field.alias
+            else:
+                param_name = field_name
             if field_name in merged_params or param_name in merged_params:
                 continue
             elif not is_valid_identifier(param_name):
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -319,11 +319,13 @@ def Field(
     default: Any = Undefined,
     *,
     default_factory: typing.Callable[[], Any] | None = None,
-    alias: str | None = None,
+    alias: str | list[str | int] | list[list[str | int]] | None = None,
     # TODO:
     #  Alternative 1: we could drop alias_priority and tell people to manually override aliases in child classes
     #  Alternative 2: we could add a new argument `override_with_alias_generator=True` equivalent to `alias_priority=1`
     alias_priority: int | None = None,
+    validation_alias: str | list[str | int] | list[list[str | int]] | None = None,
+    serialization_alias: str | None = None,
     title: str | None = None,
     description: str | None = None,
     examples: list[Any] | None = None,
@@ -348,8 +350,6 @@ def Field(
     strict: bool | None = None,
     json_schema_extra: dict[str, Any] | None = None,
     validate_default: bool | None = None,
-    validation_alias: str | list[str | int] | list[list[str | int]] | None = None,
-    serialization_alias: str | None = None,
     const: bool | None = None,
     unique_items: bool | None = None,
     allow_mutation: bool = True,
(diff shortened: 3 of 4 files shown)


hramezani:
please review

dmontagu:
Could you add a test checking what happens with the JSON schema when each of `alias`, `validation_alias`, and `serialization_alias` are set differently? I think we probably want to always use the `validation_alias`, until we add support for specifying whether you want the "input" or "output" to be generated for the JSON  schema (which we will definitely want to add at some point).

dmontagu:
Looks good to me.

The input/output thing is a bit more complicated so I think we should leave that for a separate PR, and probably need to have a brief-but-broader discussion about the API. The main issue is there needs to somehow be a way to signal that certain types in the schema should be input vs. output types, all within the same schema. (So that FastAPI can do one pass through the schema and get "output" versions for returns, and "input" versions for arguments.)

I have some ideas, but will need to discuss more and doesn't need to happen in this PR.

pydantic/_internal/_generate_schema.py
```diff
@@ -532,8 +532,10 @@ def _common_field_schema(self, name: str, field_info: FieldInfo, decorators: Dec
         return _common_field(
             schema,
             serialization_exclude=True if field_info.exclude else None,
-            validation_alias=field_info.alias,
-            serialization_alias=field_info.alias,
+            validation_alias=field_info.validation_alias,
+            serialization_alias=(
```
samuelcolvin, line 536: better that this is just `serialization_alias=field_info.serialization_alias`, then change field.

pydantic/fields.py
```diff
@@ -470,6 +478,8 @@ def Field(
         json_schema_extra=json_schema_extra,
         strict=strict,
         validate_default=validate_default,
+        validation_alias=validation_alias or alias,
+        serialization_alias=serialization_alias or alias,
```
samuelcolvin, line 482: i think we should only set `serialization_alias` to `alias` if `alias` is a string.

samuelcolvin (commented):
We still need tests for `alias=['foo', 'bar']` and `validation_alias=['foo', 'bar']`, I think this might case errors, although it should work.

Also do we need to make any changes to `alias_priority`?

samuelcolvin:
please update.

pydantic/fields.py
hramezani: [alias type is str|None](https://github.com/pydantic/pydantic/blob/1896500fb6bc371270b94f343390e2121dc0e193/pydantic/fields.py#L318), How can it be possible to have non string alias?

pydantic/_internal/_generate_schema.py
hramezani: done

pydantic/fields.py
```diff
@@ -344,6 +348,8 @@ def Field(
     strict: bool | None = None,
     json_schema_extra: dict[str, Any] | None = None,
     validate_default: bool | None = None,
+    validation_alias: str | list[str | int] | list[list[str | int]] | None = None,
```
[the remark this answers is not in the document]
hramezani: done

pydantic/fields.py
hramezani: done

hramezani:
@dmontagu @samuelcolvin I've changed `test_low_priority_alias` to demonstrate the behavior of `validation_alias` and `serialization_alias` with `alias_generator`.
Currently, `alias_generator` only applies on `alias` but by the change of this PR, the new generated alias will replace with old `validation_alias` and `serialization_alias`(which can be revert if you think it is not the right approach)

pydantic/_internal/_model_construction.py
```diff
@@ -307,4 +310,6 @@ def apply_alias_generator(config: ConfigDict, fields: dict[str, FieldInfo]) -> N
             if not isinstance(alias, str):
                 raise TypeError(f'alias_generator {alias_generator} must return str, not {alias.__class__}')
             field_info.alias = alias
+            field_info.validation_alias = alias
+            field_info.serialization_alias = alias
```
samuelcolvin, line 314: serialization alias should only be set here if `alias` is a string, we need a test for this.

pydantic/fields.py
```diff
@@ -47,6 +47,8 @@ class FieldInfo(_repr.Representation):
         'validate_default',
         'frozen',
         'final',
+        'validation_alias',
```
samuelcolvin, line 50: please move these up next to alias.

pydantic/fields.py
```diff
@@ -98,6 +100,8 @@ def __init__(self, **kwargs: Any) -> None:
         self.validate_default = kwargs.get('validate_default', None)
         self.frozen = kwargs.get('frozen', None)
         self.final = kwargs.get('final', None)
+        self.validation_alias = kwargs.get('validation_alias', None)
```
samuelcolvin, line 103: same.

pydantic/fields.py
```diff
@@ -402,6 +408,8 @@ def Field(
     :param json_schema_extra: extra dict to be merged with the JSON Schema for this field
     :param strict: enable or disable strict parsing mode
     :param validate_default: whether the default value should be validated for this field
+    :param validation_alias: the alias(es) to use to find the field in the validation data
```
samuelcolvin, line 411: same.

tests/test_aliases.py
```diff
@@ -44,7 +45,8 @@ class Model(BaseModel):
     assert Model().a == 'foobar'
     assert Model(_a='different').a == 'different'
     assert repr(Model.model_fields['a']) == (
-        "FieldInfo(annotation=str, required=False, default='foobar', alias='_a', alias_priority=2)"
+        "FieldInfo(annotation=str, required=False, default='foobar', alias='_a', "
```
samuelcolvin, line 48: i think we should only include `validation_alias` and `serialization_alias` in repr if they're different to `alias`.

pydantic/_internal/_model_construction.py
```diff
@@ -307,4 +310,6 @@ def apply_alias_generator(config: ConfigDict, fields: dict[str, FieldInfo]) -> N
             if not isinstance(alias, str):
```
samuelcolvin, line 310: we need to fix this, and update the signature of `f'alias_generator`

samuelcolvin: we also need a way for `alias_generator` to set only one of the aliases.

Perhaps we should change the signature to:

```py
class AliasGeneratorReturn(TypedDict, total=False):
    validation_alias: str | list[str | int] | list[list[str | int]]
    serialization_alias: str
```

Then the return type should be `AliasGeneratorReturn | str`.

samuelcolvin (dismissed):
still some more stuff to cleanup and fix.

pydantic/_internal/_model_construction.py
hramezani: based on the error above, `alias_generator` only returns string

samuelcolvin:
The only case where we don't (and can't) use `field.alias` (or `field.validation_alias`) for `generate_model_signature` is when that alias is not a `str`.

@dmontagu does that change your mind of remove `.alias`, and just leaving access to it via `@property`?

pydantic/_internal/_model_construction.py
```diff
@@ -237,7 +237,10 @@ def generate_model_signature(
     if var_kw:  # if custom init has no var_kw, fields which are not declared in it cannot be passed through
         allow_names = config_wrapper.populate_by_name
         for field_name, field in fields.items():
-            param_name = field.alias or field_name
+            if isinstance(field.alias, str):
```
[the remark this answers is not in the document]
hramezani: comment added

pydantic/_internal/_model_construction.py
hramezani: as `alias_generator` always returns a string this check is not needed.

tests/test_aliases.py
hramezani: done

pydantic/_internal/_model_construction.py
hramezani: Based on [@dmontagu comment](https://github.com/pydantic/pydantic/pull/5433#issuecomment-1505349710), `alias_generator` should still return `str`

pydantic/fields.py
```diff
@@ -315,11 +323,13 @@ def Field(
     default: Any = Undefined,
     *,
     default_factory: typing.Callable[[], Any] | None = None,
-    alias: str | None = None,
+    alias: str | list[str | int] | list[list[str | int]] | None = None,
```
dmontagu, line 326:
```suggestion
    alias: str | None = None,
```

dmontagu: For now, I think we should disallow the other kinds of alias to be passed into this function as the `alias` kwarg (as I think it will break type-checking anyway without a plugin or similar), and just allow the more complex variants in `validation_alias`.

pydantic/fields.py
```diff
@@ -315,11 +323,13 @@ def Field(
...
     # TODO:
     #  Alternative 1: we could drop alias_priority and tell people to manually override aliases in child classes
     #  Alternative 2: we could add a new argument `override_with_alias_generator=True` equivalent to `alias_priority=1`
```
dmontagu, line 329:
```suggestion
```
I think we can drop this comment now.

pydantic/fields.py
```diff
@@ -359,6 +369,8 @@ def Field(
     :param default_factory: callable that will be called when a default value is needed for this field
       If both `default` and `default_factory` are set, an error is raised.
     :param alias: the public name of the field
+    :param validation_alias: the alias(es) to use to find the field in the validation data
```
dmontagu, line 372:
```suggestion
    # TODO: Add documentation reference for non-str alias variants
    :param validation_alias: the alias(es) to use to find the field value during validation
```
We will need to document this better at some point but

pydantic/fields.py
```diff
@@ -441,11 +453,16 @@ def Field(
...
 
+    if serialization_alias is None and isinstance(alias, str):
+        serialization_alias = alias
+
     return FieldInfo.from_field(
         default,
         default_factory=default_factory,
         alias=alias,
         alias_priority=alias_priority,
+        validation_alias=validation_alias or alias,
```
dmontagu, line 464:
```suggestion
        validation_alias=validation_alias if validation_alias is not None else alias,
```
Could also just do this the same way you handle `serialization_alias` above, or else you can't do `validation_alias = ''`. I believe there are some tests for `alias = ''`, as weird a case as that is.

PR #5447, opened 2023-04-11
Add support for handling generic dataclasses

dmontagu:
Adds support for generic dataclasses and recursive (pydantic) dataclasses.

### DIFF
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -9,11 +9,13 @@ from functools import wraps
 from typing import Any, Callable, ClassVar, cast
 
 from pydantic_core import ArgsKwargs, SchemaSerializer, SchemaValidator, core_schema
+from typing_extensions import TypeGuard
 
 from ..errors import PydanticUndefinedAnnotation
 from ..fields import FieldInfo
-from . import _decorators
-from ._core_utils import get_type_ref
+from . import _decorators, _typing_extra
+from ._config import ConfigWrapper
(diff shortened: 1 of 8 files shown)


pydantic/_internal/_dataclasses.py
```diff
@@ -152,3 +156,22 @@ class B(A):
         and not hasattr(_cls, '__pydantic_validator__')
         and set(_cls.__dataclass_fields__).issuperset(set(getattr(_cls, '__annotations__', {})))
     )
+
+
+def rebuild_pydantic_dataclass(cls: type[PydanticDataclass], *, types_namespace: dict[str, Any] | None = None) -> None:
```
dmontagu, line 161: I think @adriangb noticed this, but rebuilding the dataclass causes you to lose information about what was on the `dataclasses.field(...)` default value.

One solution would be to store the "original" field defaults on the class so they can be re-considered if/when rebuilding is necessary. But perhaps there is a more efficient solution?

dmontagu:
@samuelcolvin I will probably need some guidance from you about how to get dataclasses with deferred annotations working.

We also may need to do some refactoring to make "rebuilding" possible without losing the info from `dataclasses.field(...)`.

I also have some hacks in the schema building in here currently that I think could be removed if we rework the dataclass building to be more similar to your "improved" BaseModel schema building. For me to do that, I think I would need some guidance on that as well.

dmontagu:
Closing in favor of #5473

PR #5450, opened 2023-04-11
Fix mypy issue

dmontagu:
Manual cherry pick of #5449 for v2.

This isn't necessary in v2 to prevent internal errors, but I'm not sure why the `ctx.api.defer()` was ever added in v1, but it's been there since the very first commit, and no tests fail when I remove it (perhaps mypy changed in a way that eliminated the need for this some time prior to the oldest version we currently test in v1, 0.910).

Considering I have no idea why it's there I think we should make this change and if it breaks anything we can add a test demonstrating the reported issue.

### DIFF
--- a/tests/mypy/outputs/v1.0.1/plugin_success.txt
+++ b/tests/mypy/outputs/v1.0.1/plugin_success.txt
@@ -3,3 +3,4 @@
 136: error: Unexpected keyword argument "description" for "AddProject"  [call-arg]
 278: error: Unexpected keyword argument "foo" for "MyDataClass"  [call-arg]
 278: error: Unexpected keyword argument "bar" for "MyDataClass"  [call-arg]
+292: error: Unused "type: ignore" comment
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ test.py
 pydantic/*.c
 pydantic/*.so
 /fastapi/
+.mypy-configs/
 
 # Other files and folders
 .python-version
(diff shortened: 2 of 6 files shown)


tests/mypy/outputs/v1.0.1/plugin_success.txt
```diff
@@ -3,3 +3,4 @@
 136: error: Unexpected keyword argument "description" for "AddProject"  [call-arg]
 278: error: Unexpected keyword argument "foo" for "MyDataClass"  [call-arg]
 278: error: Unexpected keyword argument "bar" for "MyDataClass"  [call-arg]
+292: error: Unused "type: ignore" comment
```
adriangb, line 6: Can we just remove the unused type ignore?

dmontagu: In principle it would be better if we could detect that this wasn't an untyped field and handle it properly, but I think that would amount to some complex type analysis to distinguish between `x = <some_function_that_returns_a_classmethod>()` and `x = <some_function_that_returns_an_int>()`. And, as noted in the original issue, I believe you can resolve this by giving it a proper annotation anyway like `classmethod` or whatever. (So I think it's okay to have the plugin raise that error.)

adriangb: That’s more than enough background explanation, thank you for digging and being thorough!

adriangb (approved):

Issue #5432, opened 2023-04-10
Object created by omitting a default value is not equal to object created with explicitly setting the default value


martina-oefelein:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

If a field has a default value, omitting the field when creating the object should have the same effect as explicitly setting the default value.

The objects look the same, but when compared for equality, this returns false.

The [announcement](https://docs.pydantic.dev/blog/pydantic-v2-alpha/#changes-to-basemodel) says:

> The `__eq__` method has changed for models; models are no longer considered equal to the dicts.

I wonder whether it's related to this change?

### Example Code

```Python
from pydantic import BaseModel


class SimpleModel(BaseModel):
    """Model with some simple fields."""

    field: str = "Test"


def test_default_value() -> None:
    """Omitting a value should be equivalent to setting it to its default.

    This works in Pydantic 1, but fails in Pydantic 2.
    """
    assert SimpleModel() == SimpleModel(field="Test")
    # E       AssertionError: assert SimpleModel(field='Test') == SimpleModel(field='Test')
    # E         Full diff:
    # E           SimpleModel(field='Test')
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0a1
                 install path: /Users/moefelei/Repos/cli-prototype/.venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.2 (main, Feb 16 2023, 03:07:35) [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.2.1-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']
```

dmontagu:
I definitely agree this would be a significant bug, but I can't seem to reproduce it:
```
from pydantic import BaseModel


class SimpleModel(BaseModel):
    """Model with some simple fields."""

    field: str = "Test"


def test_default_value() -> None:
    """Omitting a value should be equivalent to setting it to its default.

    This works in Pydantic 1, but fails in Pydantic 2.
    """
    assert SimpleModel() == SimpleModel(field="Test")
    print("success")


test_default_value()
#> success
```
Here's the relevant pieces I get from running `pip freeze`:
```
❯ pip freeze | grep pydantic         
pydantic==2.0a1
pydantic_core==0.23.0
```
I tested it on both python 3.10 and python 3.11.

martina-oefelein:
[...]

PR #5451, opened 2023-04-11
Pin the version of pydantic-core

dmontagu:


adriangb (approved):

dmontagu:
Thanks @martina-oefelein for reporting https://github.com/pydantic/pydantic/issues/5432 — seems to have been caused by pydantic_core releases not compatible with the alpha release. Changing to tight pinning here should prevent issues like this in the future. (Should be in effect as of the next alpha release.)

samuelcolvin:
Good idea, I was thinking we should do this.

PR #5462, opened 2023-04-12
Defer evaluation of dataclasses for test 2162

dmontagu:
Right now, the "issue 2162" dataclass tests cause both pydantic and non-pydantic dataclasses to get initialized any time you run any test from the test_dataclasses file.

This is super painful while trying to use a debugger to debug things like schema generation for one specific test, and I don't want to hit my breakpoints for the schema building for those dataclasses.

This PR changes it so that those don't get evaluated unless you run the specific test they are relevant to.

While I was at it, I gave it a more interpretable name than "issue 2162" as well.

Selected Reviewer: @kludex

### DIFF
--- a/tests/test_dataclasses.py
+++ b/tests/test_dataclasses.py
@@ -966,48 +966,69 @@ class BuiltInDataclassForPickle:
         restored_obj.value = 'value of a wrong type'
 
 
-def gen_2162_dataclasses():
-    @dataclasses.dataclass(frozen=True)
-    class StdLibFoo:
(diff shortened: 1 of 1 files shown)


dmontagu:
please review

dmontagu:
@Kludex I'm going to merge this for now as I am working on dataclasses in another PR and need this change but don't want it to show up in that diff.

BUT if you (or any else) has any objections to this change, I will be happy to revisit tomorrow / when you have a chance to review.

PR #5468, opened 2023-04-13
Get basic non-recursive generic dataclasses functionality working

dmontagu:
This PR includes various modifications to the dataclasses module that I think should be easier to review (and hopefully less controversial) than the changes to come for supporting recursive/recursive generic dataclasses, and other refactors to the dataclass schema building process.

Note this is currently a subset of #5447, but I think I will close #5447 as I think we need to make some bigger changes. This PR has the things from that PR that I would like to keep around.

I would like to get this PR (or at least a subset of it) merged before continuing on generic/recursive dataclasses for the sake of keeping the diffs as easy to review as possible.

Changes:
* Fix the core_utils `WalkAndApply` to handle dataclasses properly
* Update the handling of types_namespace in `prepare_dataclass` to be more compatible with recursive references
* Add a typeguarded `is_dataclass` function for better compatibility with our StandardDataclass protocol
* Add support for reading a typevars_map off of a dataclass and using it while building the core schema.
* Lightly rework `create_dataclass` to be compatible with generic dataclasses
* Add tests using AnalyzedType to validate data with generic dataclasses with each possible dataclass decorator: pydantic, stdlib, and pydantic + stdlib. 
* Add some failing tests for recursive dataclasses

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -340,6 +340,15 @@ class WalkAndApply:
         schema['fields'] = replaced_fields
         return schema
 
+    def handle_dataclass_args_schema(self, schema: core_schema.DataclassArgsSchema) -> CoreSchema:
+        replaced_fields: list[core_schema.DataclassField] = []
+        for field in schema['fields']:
+            replaced_field = field.copy()
+            replaced_field['schema'] = self._walk(field['schema'])
+            replaced_fields.append(replaced_field)
+        schema['fields'] = replaced_fields
+        return schema
+
     def handle_arguments_schema(self, schema: core_schema.ArgumentsSchema) -> CoreSchema:
         replaced_arguments_schema = []
         for param in schema['arguments_schema']:
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -9,6 +9,7 @@ from functools import wraps
 from typing import Any, Callable, ClassVar, cast
 
 from pydantic_core import ArgsKwargs, SchemaSerializer, SchemaValidator, core_schema
+from typing_extensions import TypeGuard
 
 from ..errors import PydanticUndefinedAnnotation
 from ..fields import FieldInfo
@@ -70,7 +71,7 @@ def prepare_dataclass(
 
     dataclass_ref = get_type_ref(cls)
     self_schema = core_schema.definition_reference_schema(dataclass_ref)
-    types_namespace = {**(types_namespace or {}), name: PydanticForwardRef(self_schema, cls)}
+    types_namespace = {name: PydanticForwardRef(self_schema, cls), **(types_namespace or {})}
     try:
         fields, _ = collect_fields(cls, bases, types_namespace, is_dataclass=True, dc_kw_only=kw_only)
     except PydanticUndefinedAnnotation as e:
@@ -123,7 +124,7 @@ def prepare_dataclass(
     return True
 
 
-def is_builtin_dataclass(_cls: type[Any]) -> bool:
+def is_builtin_dataclass(_cls: type[Any]) -> TypeGuard[type[StandardDataclass]]:
     """
     Whether a class is a stdlib dataclass
     (useful to discriminated a pydantic dataclass that is actually a wrapper around a stdlib dataclass)
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -373,6 +373,9 @@ class GenerateSchema:
         if std_schema is not None:
             return std_schema
 
+        if _typing_extra.is_dataclass(obj):
+            return self._dataclass_schema(obj, None)
+
         origin = get_origin(obj)
         if origin is None:
             if self.arbitrary_types:
@@ -382,6 +385,13 @@ class GenerateSchema:
                     f'Unable to generate pydantic-core schema for {obj!r}. '
                     f'Setting `arbitrary_types_allowed=True` in the model_config may prevent this error.'
                 )
+
+        # Need to handle generic dataclasses before looking for the schema properties because attribute accesses
+        # on _GenericAlias delegate to the origin type, so lose the information about the concrete parametrization
+        # As a result, currently, there is no way to cache the schema for generic dataclasses. This may be possible
+        # to resolve by modifying the value returned by `Generic.__class_getitem__`, but that is a dangerous game.
+        if _typing_extra.is_dataclass(origin):
+            return self._dataclass_schema(obj, origin)
 
         from_property = self._generate_schema_from_property(origin, obj)
         if from_property is not None:
@@ -898,8 +908,6 @@ class GenerateSchema:
             return None
 
         # Import here to avoid the extra import time earlier since _std_validators imports lots of things globally
-        import dataclasses
-
         from ._std_types_schema import SCHEMA_LOOKUP
 
         # instead of iterating over a list and calling is_instance, this should be somewhat faster,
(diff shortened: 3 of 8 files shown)


pydantic/_internal/_dataclasses.py
```diff
@@ -70,7 +71,7 @@ def prepare_dataclass(
 
     dataclass_ref = get_type_ref(cls)
     self_schema = core_schema.definition_reference_schema(dataclass_ref)
-    types_namespace = {**(types_namespace or {}), name: PydanticForwardRef(self_schema, cls)}
+    types_namespace = {name: PydanticForwardRef(self_schema, cls), **(types_namespace or {})}
```
adriangb, line 74: 👍🏻 yep this worked in my previous PR

pydantic/_internal/_dataclasses.py
```diff
@@ -123,7 +124,7 @@ def __init__(__dataclass_self__: PydanticDataclass, *args: Any, **kwargs: Any) -
     return True
 
 
-def is_builtin_dataclass(_cls: type[Any]) -> bool:
+def is_builtin_dataclass(_cls: type[Any]) -> TypeGuard[type[StandardDataclass]]:
```
adriangb, line 127: Nice! But does this tell us that it is a pydantic datacalss? E.g. `is_builtin_dataclass(bool)`. It's probably never called like that though.

pydantic/_internal/_generate_schema.py
```diff
@@ -911,17 +919,33 @@ def _std_types_schema(self, obj: Any) -> core_schema.CoreSchema | None:
...
         Generate schema for a dataclass.
         """
+
+        if origin is not None:
+            # In this case, we know that dataclass is a _GenericAlias, and origin is the generic type
+            # So it is safe to access dataclass.__args__ and origin.__parameters__
+            args: tuple[Any, ...] = dataclass.__args__  # type: ignore
+            parameters: tuple[_typing_extra.TypeVarType, ...] = origin.__parameters__  # type: ignore
+            typevars_map = dict(zip(parameters, args))
+            dataclass = origin
```
adriangb, line 937: Thank you for the comment :)

pydantic/_internal/_generate_schema.py
```diff
@@ -911,17 +919,33 @@ def _std_types_schema(self, obj: Any) -> core_schema.CoreSchema | None:
...
         # FIXME we need a way to make sure kw_only info is propagated through to fields
         fields, _ = collect_fields(
-            dataclass, dataclass.__bases__, self.types_namespace, dc_kw_only=True, is_dataclass=True
+            dataclass,
+            dataclass.__bases__,
+            self.types_namespace,
+            dc_kw_only=True,
+            is_dataclass=True,
+            typevars_map=typevars_map,
         )
```
adriangb, line 949: I agree with @samuelcolvin that if we have an `is_dataclass: bool` param we might just want two functions.

pydantic/_internal/_typing_extra.py
```diff
@@ -445,3 +448,11 @@ def evaluate_fwd_ref(
...
     ) -> Any:
         return ref._evaluate(globalns=globalns, localns=localns, recursive_guard=frozenset())
+
+
+def is_dataclass(_cls: type[Any]) -> TypeGuard[type[StandardDataclass]]:
+    # The dataclasses.is_dataclass function doesn't seem to provide TypeGuard functionality,
+    # so I created this convenience function
+    import dataclasses
+
+    return dataclasses.is_dataclass(_cls)
```
adriangb, line 458: How is this different from `is_builtin_dataclass` above?

tests/test_dataclasses.py
```diff
@@ -1395,17 +1395,54 @@ class Foo:
...
+        self_reference: Optional['MyDataclass'] = None
 
-    annotation = MyDataclass.__pydantic_fields__['self_reference'].annotation
-    # Currently, this is true:
-    # assert isinstance(MyDataclass, PydanticForwardRef)
-    # TODO: Probably need a way to "model_rebuild" a dataclass
-    assert annotation is MyDataclass
+    # rebuild_pydantic_dataclass(MyDataclass)
+
+    assert MyDataclass.__pydantic_fields__['self_reference'].annotation == Optional[MyDataclass]
```
adriangb, line 1406: ? I'm kinda confused by this. We don't need to do this sort of thing for models.

tests/test_dataclasses.py
```diff
@@ -1395,17 +1395,54 @@ class Foo:
...
+
+    @pydantic.dataclasses.dataclass
+    class D2:
+        d1: Optional[D1] = None
+
+    # rebuild_pydantic_dataclass(D1)
+
+    instance = D1(d2=D2(d1=D1(d2=D2(d1=D1()))))
+
+    assert AnalyzedType(D1).dump_python(instance) == {...}
```
adriangb, line 1441: Why are we using AnalyzedType in these tests? Are we actually testing it or does it make it easier to test the other stuff?

adriangb:
please update

pydantic/_internal/_dataclasses.py
dmontagu: It tells us specifically that it is _not_ a pydantic dataclass. But if it returns true you still know the thing passed is a StandardDataclass. So I think it's okay to annotate like this.

pydantic/_internal/_generate_schema.py
dmontagu: Yeah, he and I talked about it earlier, I'm redoing things now to move in that direction.

pydantic/_internal/_typing_extra.py
dmontagu: it returns True for pydantic dataclasses as well; is_builtin_dataclass is only true for _non_ pydantic dataclasses.

pydantic/_internal/_dataclasses.py
adriangb: Seems to me like it's the other way around: it tells us specifically that it _is_ a standard library dataclass, but it does _not_ tell us if it is a pydantic dataclass

pydantic/_internal/_typing_extra.py
adriangb: Gotcha. Maybe it should be:

```suggestion
def is_dataclass(_cls: type[Any]) -> TypeGuard[type[StandardDataclass] | type[PydanticDataclass]]:
    # The dataclasses.is_dataclass function doesn't seem to provide TypeGuard functionality,
    # so I created this convenience function
    import dataclasses

    return dataclasses.is_dataclass(_cls)
```

?

tests/test_dataclasses.py
adriangb: Ok that makes sense. Sorry I think I misread it as assignment instead of an assertion the first time around.

tests/test_dataclasses.py
dmontagu: Parametrized generic dataclasses don't do validation based on the parameters. (This is related to _GenericAlias calling methods on the origin class without the ability to reflect the arguments.) So the only way you can get validation for generic dataclasses that reflects the parameters is to have them be a field of something (a BaseModel, pydantic dataclass, or TypedDict), or to use AnalyzedType. AnalyzedType seemed to best-convey the intent here. Probably deserves a comment explaining this.

adriangb: Oh makes sense, thank you for explaining :)

pydantic/_internal/_core_utils.py
```diff
@@ -340,6 +340,15 @@ def handle_typed_dict_schema(self, schema: core_schema.TypedDictSchema) -> CoreS
...
         return schema
 
+    def handle_dataclass_args_schema(self, schema: core_schema.DataclassArgsSchema) -> CoreSchema:
+        replaced_fields: list[core_schema.DataclassField] = []
+        for field in schema['fields']:
+            replaced_field = field.copy()
+            replaced_field['schema'] = self._walk(field['schema'])
+            replaced_fields.append(replaced_field)
+        schema['fields'] = replaced_fields
+        return schema
```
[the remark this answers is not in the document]
dmontagu: This class recurses over a core schema applying a transform to nested item of type `CoreSchema`.

I can add a docstring here, but presumably I should add a docstring to the class and/or every method within the class. I wasn't sure whether this would be long lived when I first made it which is why I didn't doc anything, but at this point I think it has proven its utility enough that I would strongly argue for it to stay (though I think it may be better to put it in `pydantic_core`, as noted at the top of the module).

pydantic/_internal/_dataclasses.py
dmontagu: I'll also note that as of now the function is not actually used anywhere. So maybe we should just delete it anyway. (It's a hold-over from v1)

dmontagu: `        and not hasattr(_cls, '__pydantic_validator__')`
can this be the case for pydantic dataclasses?

dmontagu:
Closing in favor of #5473

pydantic/_internal/_typing_extra.py
dmontagu: a `PydanticDataclass` is already a `StandardDataclass` so I don't think the union is semantically meaningful, but if you think it conveys intent to humans better we could do that

PR #5472, opened 2023-04-13
Split `__get_pydantic_core_schema__` into `__modify_pydantic_core_schema__` and `__get_pydantic_core_schema__`

adriangb:
As discussed in meeting today. This should make it a lot clearer what these functions are expected to do.

Selected Reviewer: @kludex

### DIFF
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -377,7 +377,7 @@ from pydantic import BaseModel, ValidationError
 class RestrictCharacters:
     alphabet: Sequence[str]
 
-    def __get_pydantic_core_schema__(
+    def __modify_pydantic_core_schema__(
         self,
         schema: core_schema.CoreSchema,
     ) -> core_schema.CoreSchema:
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -366,9 +366,10 @@ included.
 
 ```py
 from dataclasses import dataclass
-from typing import Annotated, Sequence
+from typing import Sequence
 
 from pydantic_core import core_schema
+from typing_extensions import Annotated
 
 from pydantic import BaseModel, ValidationError
 
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -364,40 +364,61 @@ see [Custom Data Types](types.md#custom-data-types) for more details.
 *pydantic* will inspect the signature of `__modify_schema__` to determine whether the `field` argument should be
 included.
 
-```py output="json" test="xfail needs work" lint="skip"
-from typing import Any, Callable, Dict, Generator, Optional
+```py
+from dataclasses import dataclass
+from typing import Annotated, Sequence
 
-from pydantic import BaseModel, Field
-from pydantic_core.core_schema import ValidationInfo
+from pydantic_core import core_schema
+
+from pydantic import BaseModel, ValidationError
 
 
-class RestrictedAlphabetStr(str):
-    @classmethod
-    def __get_validators__(cls) -> Generator[Callable, None, None]:
-        yield cls.validate
+@dataclass
+class RestrictCharacters:
+    alphabet: Sequence[str]
 
-    @classmethod
-    def validate(cls, value: str, info: ValidationInfo):
-        alphabet = field.field_info.extra['alphabet']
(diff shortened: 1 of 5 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -1237,8 +1240,10 @@ def _common_field(
     serialization_alias: str | None = None,
     serialization_exclude: bool | None = None,
     frozen: bool | None = None,
+    field_info: FieldInfo,
     metadata: Any = None,
 ) -> _CommonField:
+    metadata[METADATA_FIELD_INFO_KEY] = field_info
```
dmontagu, line 1246: this seems dangerous since based on this annotation metadata may not be a dict, just pointing it out

pydantic/json_schema.py
```diff
@@ -35,11 +35,20 @@
...
 
 JsonSchemaValue = Dict[str, Any]
 
 
+def get_field_info(
+    schema: core_schema.CoreSchema | core_schema.TypedDictField | core_schema.DataclassField,
+) -> FieldInfo | None:
+    from pydantic._internal._generate_schema import METADATA_FIELD_INFO_KEY
+
+    return schema.get('metadata', {}).get(METADATA_FIELD_INFO_KEY, None)
```
dmontagu, line 49: I would look at the `pydantic._internal._core_metadata.CoreMetadataHandler` which handles some other edge cases this doesn't (e.g., metadata exists and isn't a dict)

pydantic/_internal/_generate_schema.py
dmontagu: (You may want to use `pydantic._internal._core_metadata.CoreMetadataHandler` and/or `pydantic._internal._core_metadata.build_metadata_dict`)

docs/usage/schema.md
```diff
@@ -355,49 +355,136 @@ Defaults can be set outside `Annotated` as the assigned value or with `Field.def
...
 
+from pydantic import BaseModel, ValidationError
 
-print(MyModel.schema_json(indent=2))
+
+@dataclass
+class RestrictCharacters:
+    alphabet: Sequence[str]
+
+    def __modify_pydantic_core_schema__(
```
dmontagu, line 446: this should be a classmethod right?

dmontagu: I think I misunderstood what was going on, I get it now.

adriangb: It doesn’t seem to need to be

pydantic/_internal/_generate_schema.py
```diff
@@ -1181,7 +1181,7 @@ def apply_single_annotation(  # noqa C901
     if metadata_schema is not None:
         return metadata_schema
 
-    metadata_get_schema = getattr(metadata, '__get_pydantic_core_schema__', None)
```
dmontagu, line 1184: If we keep the `__pydantic_core_schema__` thing maybe we should keep the `__get_pydantic_core_schema__` thing? I would think we either support both or neither.. Could be convinced otherwise if there's benefits for practicality

adriangb: Is `__pydantic_core_schema__` another function we support?

pydantic/types.py
```diff
@@ -389,8 +388,13 @@ def __class_getitem__(cls, item: AnyType) -> AnyType:
 
         @classmethod
         def __get_pydantic_core_schema__(
```
dmontagu, line 390: Would it make sense to use `__pydantic_core_schema__ = core_schema.json_schema(None)`? Or do we prefer it like this to defer the initialization cost.

dmontagu: Feels like ideally we'd use something like `@cached_property`, maybe later.

pydantic/_internal/_generate_schema.py
dmontagu: I see your trick below of having one handle the case where it is the type directly, and another where it is modifying an existing schema. Given that, I think we should probably drop the `__pydantic_core_schema__` check above, and just _only_ use `__modify_pydantic_core_schema__`. Unless there is some practical reason I'm missing.

PR #5480, opened 2023-04-14
Add migration error messages to V2

Kludex:
## Change Summary

This PR does two things:
1. Add deprecation warnings to the `deprecated` namespace.
2. Tries to improve the user experience on imports that were changed/removed from v1 to v2.

For 2, I've used the scripts on https://gist.github.com/Kludex/9087acb5af2a940bd3fcefce96799412 as a base line.

## Deprecation Warning on IDEs

The `deprecated` from [PEP702](https://peps.python.org/pep-0702/) is used for VSCode, and the `warnings.warn` is used for PyCharm. As a note, PyCharm doesn't "work" properly when using variables on warnings i.e. `warn(_MY_DESCRIPTION, DeprecationWarning)` doesn't work.

On VSCode:
image

On Pycharm:
image

## Related issue number

- Closes [#5016](https://github.com/pydantic/pydantic/issues/5016)

## Questions

1. Should I test that `__deprecated__` field is used on the objects that are using the deprecated decorator?

## Missing

- [ ] Tests for `DeprecationWarning`s introduced.

## Checklist

Selected Reviewer: @adriangb


samuelcolvin:
fix conflicts to run test.

docs/usage/errors.md
```diff
@@ -69,6 +69,10 @@ TODO
 
 TODO
 
+## Import error {#import-error}
+
+TODO
```
Kludex, line 74: Good enough?
```suggestion
This error is raised when you try to import an object that was available on V1, but it's not anymore on V2.
```

pydantic/deprecated/json.py
```diff
@@ -104,10 +113,16 @@ def custom_pydantic_encoder(type_encoders: Dict[Any, Callable[[Type[Any]], Any]]
         return pydantic_encoder(obj)
 
 
+@deprecated('timedelta_isoformat is deprecated.')
 def timedelta_isoformat(td: datetime.timedelta) -> str:
     """
     ISO 8601 encoding for Python timedelta object.
     """
+    warnings.warn(
+        'timedelta_isoformat is deprecated, use timedelta.total_seconds() instead.',
```
Kludex, line 122:
```suggestion
        'timedelta_isoformat is deprecated.',
```

pydantic/deprecated/tools.py
```diff
@@ -4,20 +4,26 @@
...
 
+@deprecated('parse_obj_as is deprecated. Use pydantic.AnalyzedType.validate_python instead.')
 def parse_obj_as(type_: type[T], obj: Any, type_name: NameFactory | None = None) -> T:
-    # TODO: add deprecation warning of some sort
+    warnings.warn(
+        'parse_obj_as is deprecated. Use pydantic.AnalyzedType.validate_python instead.',
+        DeprecationWarning,
+        stacklevel=2,
+    )
     if type_name is not None:  # pragma: no cover
```
Kludex, line 27: Should I remove this warning?

pydantic/deprecated/json.py
```diff
@@ -70,11 +73,13 @@ def decimal_encoder(dec_value: Decimal) -> Union[int, float]:
 }
 
 
+@deprecated('pydantic_encoder is deprecated, use BaseModel.model_dump instead.')
 def pydantic_encoder(obj: Any) -> Any:
```
adriangb, line 77: Are these really replacements? Wouldn't `AnalyzedType(obj).dump_json()` be a better replacement?

pydantic/deprecated/json.py
```diff
@@ -104,10 +113,16 @@ def custom_pydantic_encoder(type_encoders: Dict[Any, Callable[[Type[Any]], Any]]
...
+@deprecated('timedelta_isoformat is deprecated.')
 def timedelta_isoformat(td: datetime.timedelta) -> str:
     """
     ISO 8601 encoding for Python timedelta object.
     """
+    warnings.warn(
+        'timedelta_isoformat is deprecated.',
+        DeprecationWarning,
+        stacklevel=2,
+    )
```
adriangb, line 125: Note: I'm guessing we can't create a decorator that applies both because static type checkers rely on the `@deprecated` decorator being used directly, right?

pydantic/deprecated/tools.py
adriangb: I think it's okay to leave it. If you choose to ignore the first warning we still want the second one. We might even want it to be an error.

adriangb (approved):
Looks good, left a few small comments

pydantic/deprecated/json.py
Kludex: right :cry:

docs/usage/errors.md
```diff
@@ -69,6 +69,10 @@ TODO
 
 TODO
 
+## Import error {#import-error}
+
+This error is raised when you try to import an object that was available on V1, but it's not anymore on V2.
```
dmontagu, line 74:
```suggestion
This error is raised when you try to import an object that was available in V1 but has been removed in V2.
```

pydantic/deprecated/json.py
```diff
@@ -70,11 +73,13 @@ def decimal_encoder(dec_value: Decimal) -> Union[int, float]:
 }
 
 
+@deprecated('pydantic_encoder is deprecated, use BaseModel.model_dump instead.')
```
dmontagu, line 76:
```suggestion
@deprecated('pydantic_encoder is deprecated, use `pydantic_core.to_jsonable_python` instead.')
```
I think this is the better substitute. (And I think addresses @adriangb's point as well)

pydantic/deprecated/json.py
```diff
@@ -91,8 +96,12 @@ def pydantic_encoder(obj: Any) -> Any:
         raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
 
 
+@deprecated('custom_pydantic_encoder is deprecated, use BaseModel.model_dump instead.')
```
dmontagu, line 99:
```suggestion
@deprecated('custom_pydantic_encoder is deprecated')
```
I'm not sure what the replacement here is. @samuelcolvin is it possible to achieve something similar with `to_jsonable_python`?

pydantic/deprecated/json.py
dmontagu, line 125:
```suggestion
    warnings.warn('timedelta_isoformat is deprecated.', DeprecationWarning, stacklevel=2)
```
I think it makes it multiline because of the trailing comma on the final item, if you delete it it will make it one line. @samuelcolvin I think you brought this up before, if you're still open to it I'd be in favor of making it just always reformat to a single line when it can. (That's a separate thing to this PR of course.)

pydantic/errors.py
```diff
@@ -94,6 +95,13 @@ def from_name_error(cls, name_error: NameError) -> PydanticUndefinedAnnotation:
         return cls(name=name, message=str(name_error))
 
 
+class PydanticImportError(PydanticErrorMixin, ImportError):
+    """Error occurs when a module cannot be imported, given module changes between V1 and V2."""
```
dmontagu, line 99:
```suggestion
    """Error occurs when an import fails due to module changes between V1 and V2."""
```

dmontagu (commented):
I think this looks good with the suggested changes. Feel free to reject if you think the suggestions are wrong.

pydantic/deprecated/json.py
dmontagu: I see the `fallback` kwarg lets you handle the encoding of objects that would otherwise be non-serializable, but I don't see a way to override the encoding for specific types. @samuelcolvin it seems like it may be worth adding support for this in pydantic-core, which might also help resolve the xfailing `tests.test_forward_ref.test_json_encoder_str` (though would require support for something similar in `SchemaSerializer`).

In the meantime, my suggestion would be to also add a TODO note like this:
```suggestion
# TODO: Add a suggested migration path once there is a way to use custom encoders
@deprecated('custom_pydantic_encoder is deprecated.')
```
(happy to get Samuel's thoughts though.)

Issue #5488, opened 2023-04-14
Should setting `...` as the default make a dataclasses field required?


dmontagu:
In V1 dataclasses, setting a field default of `...` does not make the field required, but makes that the actual default value. This is the same behavior as standard dataclasses.

In V2 dataclasses, it makes the field required. Is this how it should work?

Standard + V1 behavior:
```python
from typing import Any
import dataclasses

import pydantic


@pydantic.dataclasses.dataclass
class PydanticDataclass:
    v: Any = ...


print(PydanticDataclass().v)
# > Ellipsis


@dataclasses.dataclass
class StandardDataclass:
    v: Any = ...


print(StandardDataclass().v)
# > Ellipsis
```
V2 behavior:
```python
from typing import Any

import pydantic


@pydantic.dataclasses.dataclass
class PydanticDataclass:
    v: Any = ...


print(PydanticDataclass().v)

#> pydantic_core._pydantic_core.ValidationError: 1 validation error for PydanticDataclass
#> v
#>   Field required [type=missing, input_value=ArgsKwargs(()), input_type=ArgsKwargs]
```

samuelcolvin:
I'm kind of neutral.

Would be nice for BaseModel and dataclass to have the same behaviour.

Would be nice for a pydantic dataclass to match a vanilla dataclass.

Would be nice for v1 and V2 to have the same behaviour.

Choose two.

I guess we choose 2 and 3?

adriangb:
Do we do the `...` doesn't count as a default value because of `foo: str = Field(...)`?

samuelcolvin:
When I first started pydantic I didn't know about annotation only fields, so I chose ellipsis as meaning required. It's stuck until now.

Happy to change.

adriangb:
Does that mean we can choose:

> Would be nice for BaseModel and dataclass to have the same behaviour.
> Would be nice for a pydantic dataclass to match a vanilla dataclass.

And change BaseModel to match vanilla dataclasses?

samuelcolvin:
Any way we could use sourcegraph or GitHub search to see how much code would break?

We might need to keep the behaviour but add a deprecation warning, then remove in V3.

r-richmond:
>I didn't know about annotation only fields, so I chose ellipsis as meaning required

as a personal antidote that was my initial understanding as well, I imagine that is not uncommon for new users of pydantic sit feels like the intended meaning.

[...]

Issue #5487, opened 2023-04-14
Should calls to pydantic dataclasses with invalid signature raise a ValidationError or TypeError?


dmontagu:
I noticed that now, instead of raising a `TypeError` when you call dataclasses with an invalid signature, we raise a `ValidationError`.

Are we happy with this? In v1 it raises a `TypeError`:

```python
import pydantic

@pydantic.dataclasses.dataclass
class X:
    x: int

X()

# V1 error:
# TypeError: X.__init__() missing 1 required positional argument: 'x'

# V2 error:
# pydantic_core._pydantic_core.ValidationError: 1 validation error for X
# x
#   Field required [type=missing, input_value=ArgsKwargs(()), input_type=ArgsKwargs]
```

samuelcolvin:
I think it's much better now.

dmontagu:
Main downside is the risk that someone had set up error handling for invalid calls. But I'm fine with this if you are. Feel free to just close the issue.

adriangb:
@samuelcolvin any reason we can't do something like:

```python
from typing import Any

from pydantic import ValidationError


class ValidationTypeError(ValidationError):
    pass


def func(*args: Any, **kwargs: Any) -> Any:
    # some validation is happening
    if len(args) < 2:
        raise ValidationTypeError(
            'not enough args',
            errors=[],
        )

try:
    func(1)
except TypeError:
    print('okay')


try:
    func(1)
except ValidationError:
    print('okay')
```

adriangb:
Since `ValidationError` currently inherits form `ValueError` it might be tricky (as in, require several macros) to make this work in Rust. It's doable but I'm not going to implement it unless we need / really want to.

PR #5489, opened 2023-04-14
Add support for Hashable

dmontagu:
Ran into two surprises while working on this:

https://github.com/pydantic/pydantic/issues/5487
https://github.com/pydantic/pydantic/issues/5488

But I think this we could merge this independently of addressing those issues.

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -358,6 +358,8 @@ class GenerateSchema:
             return self.generate_schema(obj.__supertype__)
         elif obj == re.Pattern:
             return self._pattern_schema(obj)
+        elif obj is collections.abc.Hashable or obj is typing.Hashable:
+            return self._hashable_schema()
         elif isinstance(obj, type):
             if obj is dict:
                 return self._dict_schema(obj)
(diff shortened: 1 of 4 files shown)


tests/test_validators_dataclass.py
```diff
@@ -180,7 +179,7 @@ def root_validator(cls, values):
     assert exc_info.value.errors() == [
         {
             'ctx': {'error': 'foobar'},
-            'input': HasRepr(IsStr(regex=re.escape("ArgsKwargs((1,), {'b': 'snap dragon'})"))),
+            'input': HasRepr("ArgsKwargs((1,), {'b': 'snap dragon'})"),
```
dmontagu, line 182: I added this test the other day, but while working on this I realized I didn't need the `IsStr(regex=` thing since I was just doing an exact equality check. So I removed here.

pydantic/_internal/_generate_schema.py
```diff
@@ -905,6 +907,13 @@ def _pattern_schema(self, pattern_type: Any) -> core_schema.CoreSchema:
         else:
             raise PydanticSchemaGenerationError(f'Unable to generate pydantic-core schema for {pattern_type!r}.')
 
+    def _hashable_schema(self) -> core_schema.CoreSchema:
+        return core_schema.custom_error_schema(
+            core_schema.is_instance_schema(collections.abc.Hashable),
+            custom_error_type='is_hashable',
+            custom_error_message='Input should be hashable',
+        )
```
dmontagu, line 915: This adds an error_type, but it's very similar to the `_type_schema` and I think it's reasonable to have a special message for this. I think it looks better than the error message we had for arbitrary types before. (Which you can see in the updated tests below.)

(Also, thanks github copilot, wouldn't have thought to use `custom_error_schema` without it.)

dmontagu: Also @tpdorsey for reference — not sure if we should do something to make sure the custom_error_type is tracked better/make sure it gets documented.

dmontagu:
please review

adriangb (approved):
What we have looks good to me. I must admit though I am a bit confused w.r.t how e.g. serialization works (any serializer?). But I think that can easily be resolved with some docs.

adriangb:
please update with tests discussed offline (or merge as is, they were minor). just don't want this to fall behind main.

PR #5490, opened 2023-04-15
Rework `__get_pydantic_core_schema__` APIs

adriangb:
Change the signature into (type, handler) -> CoreSchema; this operates similar to wrap validators (if you consider the input to be the type and the output the CoreSchema). The innermost handler is our internal schema generation.

All in all, this resolves several issues:
- Much more flexibility in schema creation, including handling of arbitrary types using Annotated without subclassing.
- Avoids exposing the internal `GenerateSchema` class
- Unifies the signature and usage of what is currently two mostly unrelated functions called `__get_pydantic_core_schema__`

Selected Reviewer: @dmontagu

### DIFF
--- a/docs/usage/schema.md
+++ b/docs/usage/schema.md
@@ -357,13 +357,12 @@ For versions of Python prior to 3.9, `typing_extensions.Annotated` can be used.
 
 ## Modifying schema in custom types and custom fields
 
-Custom types (used as `field_name: TheType` or `field_name: Annotated[TheType, ...]`) can *override* schema generation by implementing a `__get_pydantic_core_schema__` method.
+Custom types (used as `field_name: TheType` or `field_name: Annotated[TheType, ...]`) can *override* schema generation by implementing a `__modify_pydantic_core_schema__` method.
 This method receives a single positional argument with the type annotation that corresponds to this type (so in the case of `TheType[T][int]` it would be `TheType[int]`).
-All implementation of `__get_pydantic_core_schema__` *must* accept `**_kwargs` and ignore them; they are used for private implementations.
 
 ```py
 from dataclasses import dataclass
-from typing import Any, Dict, List
+from typing import Any, Callable, Dict, List, Type
 
 from pydantic_core import core_schema
 
@@ -379,8 +378,8 @@ class CompressedString:
         return ' '.join([self.dictionary[key] for key in self.text])
 
     @classmethod
-    def __get_pydantic_core_schema__(
-        cls, source: Any, **_kwargs: Any
+    def __modify_pydantic_core_schema__(
+        cls, source: Type[Any], handler: Callable[[Any], core_schema.CoreSchema]
     ) -> core_schema.CoreSchema:
         assert source is CompressedString
         return core_schema.no_info_after_validator_function(
@@ -431,7 +430,7 @@ This method receives a single positional argument `schema: pydantic_core.core_sc
 
 ```py
 from dataclasses import dataclass
-from typing import Sequence
+from typing import Any, Callable, Sequence, Type
 
 from pydantic_core import core_schema
 from typing_extensions import Annotated
@@ -444,9 +443,11 @@ class RestrictCharacters:
     alphabet: Sequence[str]
 
     def __modify_pydantic_core_schema__(
-        self,
-        schema: core_schema.CoreSchema,
+        self, source: Type[Any], handler: Callable[[Any], core_schema.CoreSchema]
     ) -> core_schema.CoreSchema:
+        if not self.alphabet:
+            raise ValueError('Alphabet may not be empty')
+        schema = handler(source)  # get the CoreSchema from the type / inner constraints
         if schema['type'] != 'str':
             raise TypeError('RestrictCharacters can only be applied to strings')
         return core_schema.no_info_after_validator_function(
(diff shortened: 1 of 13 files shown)


adriangb:
TODO: tests that put the new signature to use.

adriangb:
please review

docs/usage/schema.md
```diff
@@ -487,6 +489,142 @@ except ValidationError as e:
...
+from pydantic import BaseModel
+
+
+class AllowAnySubclass:
+    def __modify_pydantic_core_schema__(
+        self, source: Type[Any], handler: Callable[[], core_schema.CoreSchema]
+    ) -> core_schema.CoreSchema:
+        # we can't call handler since it will fail for abitrary types
+        def validate(value: Any) -> Any:
+            assert isinstance(value, source)
```
adriangb, line 603: This doesn't work yet, but @dmontagu and I were chatting and I'm positive we can refactor toward this if we want this API. It would make it very easy to manage what happens with arbitrary types (this example, the other option being `RequireExactSubclass` or something) and I think they'll be other use cases as well.

pydantic/_internal/_generate_schema.py
```diff
@@ -683,7 +683,7 @@ def _generate_parameter_schema(
             field = FieldInfo.from_annotated_attribute(annotation, default)
         assert field.annotation is not None, 'field.annotation should not be None when generating a schema'
         schema = self.generate_schema(field.annotation)
-        schema = apply_annotations(schema, field.metadata, self.definitions)
+        schema = apply_annotations(schema, field.metadata, self.definitions, annotation)
```
adriangb, line 686: One fair question when looking at this is "what if `annotation` is an `Annotated`?". It ends up not being a problem and the user always gets the `Annotated` unwrapped because:
1. `Annotated` self flatens: `assert Annotated[Annotated[str, "foo"], "bar"] == Annotated[str, 'foo', 'bar']`
2. We must have just unwrapped an `Annotated` to get here, and because of (1) they can't be immediately nested.

pydantic/_internal/_generate_schema.py
```diff
@@ -1176,36 +1176,51 @@ def apply_model_validators(
...
 ) -> core_schema.CoreSchema:
     """
     Apply arguments from `Annotated` or from `FieldInfo` to a schema.
     """
     handler = CoreMetadataHandler(schema)
-    for metadata in annotations:
-        schema = apply_single_annotation(schema, metadata, definitions)
+
+    def expand_grouped_metadata(annotation: Any) -> Iterator[Any]:
+        if isinstance(annotation, GroupedMetadata):
```
adriangb, line 1190: Is `FieldInfo` a `GroupedMetadata`? Should check and add a test if this is a bug.

docs/usage/schema.md
adriangb: See https://github.com/pydantic/pydantic/pull/5493

docs/usage/schema.md
```diff
@@ -444,9 +444,11 @@ class RestrictCharacters:
     alphabet: Sequence[str]
 
     def __modify_pydantic_core_schema__(
-        self,
-        schema: core_schema.CoreSchema,
+        self, source: Type[Any], handler: Callable[[], core_schema.CoreSchema]
     ) -> core_schema.CoreSchema:
```
dmontagu, line 448: I saw somewhere you mentioned the idea of having these behave like "wrap" validators. For that, I would think you'd want the signature to be:
```python
    def __modify_pydantic_core_schema__(
        self, source: Type[Any], schema: core_schema.CoreSchema, handler: Callable[[core_schema.CoreSchema], core_schema.CoreSchema]
    ) -> core_schema.CoreSchema:
```

That would let you receive the input schema (as `schema`), make changes before passing it to the other modifiers, pass the modified version into the handler, then make more changes before returning the final result.

More concretely, it might look like:
```python
    def __modify_pydantic_core_schema__(
        self, source: Type[Any], schema: core_schema.CoreSchema, handler: Callable[[core_schema.CoreSchema], core_schema.CoreSchema]
    ) -> core_schema.CoreSchema:
        modified_schema_before_applying_other_modifications = do_before_modifications(schema, source)
        modified_schema_after_applying_other_modifications = handler(modified_schema_before_applying_other_modifications)
        final_schema = do_after_modifications(schema, source)
        return final_schema
```

I would understand if we don't want to go whole-hog on supporting wrapping (though I think it lets you do things you couldn't otherwise do, which is nice), but if we don't do that, it's not clear to me what benefit we get from replacing `schema` with `handler: Callable[[], core_schema.CoreSchema]`.

If the only benefit is that you can choose not to call the handler if you don't need the input value, I am not sure it's worth changing the signature (other than adding `source`, which I think still makes sense), as it doesn't seem to enable much more utility for just a small potential performance improvement. But perhaps I'm misunderstanding the changes here.

If my assessment above is wrong, could you point me at an example of what this change to the signature enables?

adriangb: The big difference is that eventually we can invert the order and only generate the schema when the handler is called. Specifically this allows this function to handle cases where generating the schema would eventually fail (as is the case for arbitrary types), but also gives us more flexibility to make other changes down the road since we make less API promises.

I also think that the signature is not necessarily more complex (there's one argument less) and neither is usage (one function call).

What I am not sure about is what the signature for the handler should be. I think we could gave it require or optionally accept a source type (could be pass through) or CoreSchema (would have to be optional I think?). The latter would look a lot like your suggestion.

adriangb: Another way to view this is that the input is always a source type and the output is always a core schema. If you view it that way we should do `handler: Callable[[Any], CoreSchema]` for 1:1 parity with wrap validators (conceptually at least)

adriangb:
I merged #5493 into this and pushed a backup of the state right before that to https://github.com/pydantic/pydantic/pull/new/wrap-generate-schema-backup

docs/usage/schema.md
adriangb: I've not unified all of the signatures with the handler / callback

docs/usage/schema.md
adriangb: Works now

pydantic/_internal/_generate_schema.py
adriangb: That said, if you have something like `Annotated[List[Annotated[int, …], SomeMetadata]` then SomeMetadata should (need to add a test) get a `List[Annotated[…]]`

pydantic/main.py
```diff
@@ -231,7 +230,9 @@ def __init__(__pydantic_self__, **data: Any) -> None:  # type: ignore
         __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
 
     @classmethod
-    def __get_pydantic_core_schema__(cls, source: type[BaseModel], gen_schema: GenerateSchema) -> CoreSchema | None:
+    def __get_pydantic_core_schema__(
+        cls, __source: type[BaseModel], __handler: Callable[[Any], CoreSchema]
```
dmontagu, line 234: It looks like this is the only place with the two leading `_` in the signature — was that on purpose? I understand it's meant to indicate it will be called positionally, just since it didn't have those elsewhere just wanted to make sure it was on purpose here.

pydantic/_internal/_generate_schema.py
```diff
@@ -1173,41 +1188,67 @@ def apply_model_validators(
...
+) -> ModifyCoreSchemaWrapHandler:
     """
     Apply arguments from `Annotated` or from `FieldInfo` to a schema.
     """
-    handler = CoreMetadataHandler(schema)
     for metadata in annotations:
-        schema = apply_single_annotation(schema, metadata, definitions)
+        if metadata is None:
+            continue
+        # check if this supports
```
dmontagu, line 1201: Was this comment supposed to say more?

pydantic/main.py
adriangb: It is on purpose. Using `__` allows subclasses to name these whatever they want, including `_: Any` to tell their IDE they don't care about that parameter. If we did `source: ...` then they _have_ to call that parameter `source`.

pydantic/_internal/_generate_schema.py
adriangb: No I have no idea what where I was going with that; deleted it.

pydantic/_internal/_generate_schema.py
```diff
@@ -1173,41 +1188,66 @@ def apply_model_validators(
 
 
 def apply_annotations(
```
dmontagu, line 1190: I'll note that replacing `apply_annotations` with this keeps all tests passing, reduces the size of the "fork" coming from `if metadata_get_schema is not None`, and eliminates the need for the interior binding of the metadata/metadata_get_schema that is only necessary due to being in the for loop:
```
def apply_annotations(
    get_inner_schema: ModifyCoreSchemaWrapHandler,
    annotations: typing.Iterable[Any],
    definitions: dict[str, core_schema.CoreSchema],
) -> ModifyCoreSchemaWrapHandler:
    """
    Apply arguments from `Annotated` or from `FieldInfo` to a schema.
    """
    for metadata in annotations:
        if metadata is None:
            continue
        get_inner_schema = get_wrapped_inner_schema(get_inner_schema, metadata, definitions)

    return get_inner_schema


def get_wrapped_inner_schema(
    get_inner_schema: ModifyCoreSchemaWrapHandler, metadata: Any, definitions: dict[str, core_schema.CoreSchema]
) -> Callable[[Any, ModifyCoreSchemaWrapHandler], core_schema.CoreSchema]:
    metadata_get_schema = getattr(metadata, '__get_pydantic_core_schema__', None) or (
        lambda source, handler: handler(source)
    )

    def _new_inner_schema_handler(
        source: Any, handler: ModifyCoreSchemaWrapHandler = get_inner_schema
    ) -> core_schema.CoreSchema:
        schema = metadata_get_schema(source, handler)
        schema = apply_single_annotation(schema, metadata, definitions)
        metadata_js_modify_function = _get_pydantic_modify_json_schema(metadata)
        CoreMetadataHandler(schema).compose_js_modify_functions(metadata_js_modify_function)
        return schema

    return _new_inner_schema_handler
```
Feel free to leave as is if you prefer

dmontagu (approved):
Just had the one suggestion to simplify the `apply_annotations` function which you can take or leave. Otherwise, LGTM

pydantic/_internal/_generate_schema.py
adriangb: Awesome, it is much simpler, implemented!

PR #5492, opened 2023-04-15
rename functions in _decorators.py

adriangb:
I tried to figure out what these functions do based on their name and couldn't. I am suggesting these names as alternatives, only if it's clearer for others as well.

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -276,7 +276,7 @@ class DecoratorInfos:
         return res
 
 
-def inspect_validator(validator: Callable[..., Any], mode: FieldValidatorModes) -> bool:
+def does_any_validator_require_info_arg(validator: Callable[..., Any], mode: FieldValidatorModes) -> bool:
     """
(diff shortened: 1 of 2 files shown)


pydantic/_internal/_decorators.py
```diff
@@ -309,7 +309,9 @@ def inspect_validator(validator: Callable[..., Any], mode: FieldValidatorModes)
     )
 
 
-def inspect_field_serializer(serializer: Callable[..., Any], mode: Literal['plain', 'wrap']) -> tuple[bool, bool]:
+def does_field_serializer_require_info_arg(
```
samuelcolvin, line 312: This function returns two things.

samuelcolvin (commented):
Humm, I'm neutral. I agree "inspect" is generic, but these functions don't just decide one thing, they also raise errors if the signature is wrong.

pydantic/_internal/_decorators.py
adriangb: Good point. It’ll be nice when we can do `-> {“foo”: int}`.

adriangb:
Probably not worth it

PR #5493, opened 2023-04-15
Change signature of __get_pydantic_core_schema__ to hide GenerateSchema

adriangb:
Taking #5490 one step further, we can make the signatures of `__get_pydantic_core_schema__` and `__modify_pydantic_core_schema__` equivalent.

TODO: can we just collapse them into one thing again?
TODO: add tests for stuff like a try/except calling of handler

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -5,6 +5,7 @@ from __future__ import annotations as _annotations
 
 import collections.abc
 import dataclasses
+import inspect
 import re
 import sys
 import typing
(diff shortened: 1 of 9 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -248,8 +252,16 @@ def _generate_schema_from_property(self, obj: Any, source: Any) -> core_schema.C
         """
         get_schema = getattr(obj, '__get_pydantic_core_schema__', None)
         if get_schema is not None:
+            # (source) -> CoreSchema
+            if len(inspect.signature(get_schema).parameters) == 1:
+                return get_schema(source)
+
+            # (source, handler) -> CoreSchema
+            def handler(source: Any) -> core_schema.CoreSchema:
```
dmontagu, line 260: Here handler accepts a source: Any, but down below, the annotations are:
```python
    @classmethod
        cls, source: type[Any], handler: Callable[[], core_schema.CoreSchema]
    ) -> core_schema.CoreSchema:
```
Note — handler doesn't accept arguments. This seems wrong to me, it seems it `handler` should accept a type[Any] and return a `core_schema.CoreSchema`.

PR #5500, opened 2023-04-17
Add AliasPath and AliasPathChoices and use them in validation_alias type

hramezani:
## Change Summary

Add AliasPath and AliasPathChoices and use them in validation_alias type
## Checklist

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -348,7 +348,7 @@ class AliasPath:
         return self.paths
 
 
-class AliasPathChoices:
+class AliasChoices:
     def __init__(self, *args: str | AliasPath) -> None:
         self.choices = args
 
@@ -368,7 +368,7 @@ def Field(  # noqa C901
     default_factory: typing.Callable[[], Any] | None = None,
     alias: str | None = None,
     alias_priority: int | None = None,
-    validation_alias: str | AliasPathChoices | AliasPath | None = None,
+    validation_alias: str | AliasChoices | AliasPath | None = None,
     serialization_alias: str | None = None,
     title: str | None = None,
     description: str | None = None,
@@ -486,10 +486,10 @@ def Field(  # noqa C901
 
     converted_validation_alias: str | list[str | int] | list[list[str | int]] | None = None
     if validation_alias:
-        if not isinstance(validation_alias, (str, AliasPathChoices, AliasPath)):
-            raise TypeError('Invalid `validation_alias` type. it should be `str`, `AliasPathChoices`, or `AliasPath`')
+        if not isinstance(validation_alias, (str, AliasChoices, AliasPath)):
+            raise TypeError('Invalid `validation_alias` type. it should be `str`, `AliasChoices`, or `AliasPath`')
 
-        if isinstance(validation_alias, (AliasPathChoices, AliasPath)):
+        if isinstance(validation_alias, (AliasChoices, AliasPath)):
             converted_validation_alias = validation_alias.convert_to_aliases()
         else:
             converted_validation_alias = validation_alias
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -15,6 +15,7 @@ from . import types
 from ._internal import _fields, _forward_ref, _repr, _typing_extra, _utils
 from ._internal._fields import Undefined
 from ._internal._generics import replace_types
+from ._internal._internal_dataclass import slots_dataclass
 from .errors import PydanticUserError
 
 if typing.TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


hramezani:
please review

pydantic/fields.py
```diff
@@ -340,13 +340,35 @@ def __repr_args__(self) -> ReprArgs:
                     yield s, value
 
 
-def Field(
+class AliasPath:
+    def __init__(self, first_arg: str, *args: str | int) -> None:
+        self.paths = [first_arg] + list(args)
```
dmontagu, line 345: Minor point, but I would change this attribute's name from `paths` to `path` since the path is the sequence of fields, not each field on its own. I think it would be less of an issue if the name of the class as a whole didn't already have `Path` in the name.

pydantic/fields.py
```diff
@@ -340,13 +340,35 @@ def __repr_args__(self) -> ReprArgs:
...
+
+
+def Field(  # noqa C901
     default: Any = Undefined,
     *,
     default_factory: typing.Callable[[], Any] | None = None,
     alias: str | None = None,
     alias_priority: int | None = None,
-    validation_alias: str | list[str | int] | list[list[str | int]] | None = None,
+    validation_alias: str | AliasChoices | AliasPath | None = None,
```
dmontagu, line 371:
```suggestion
    validation_alias: str | AliasPath | AliasChoices | None = None,
```
this order seems more natural to me but I'm okay either way.

pydantic/fields.py
```diff
@@ -340,13 +340,35 @@ def __repr_args__(self) -> ReprArgs:
                     yield s, value
 
 
-def Field(
+class AliasPath:
+    def __init__(self, first_arg: str, *args: str | int) -> None:
+        self.paths = [first_arg] + list(args)
+
+    def convert_to_aliases(self) -> list[str | int]:
+        return self.paths
```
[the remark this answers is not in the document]
dmontagu: Looking farther I see why you did this, it lets you call the same method below. That's reasonable to me. Ignore the above (though I still think we should rename to `path` rather than `paths` and use `@slots_dataclass`).

tests/test_aliases.py
dmontagu, line 1: I would suggest we add some tests that check that `AliasPath` and `AliasChoices` work for parsing data. Right now it looks like we just test that they produce the right pydantic_core alias, but I also think we should have a couple tests like
```python
    class Model(BaseModel):
        x: str = Field(validation_alias=AliasChoices('a', AliasPath('b', 1)), [['a'], ['b', 1]])

    Model.model_validate({'a': 'hello'}).x == 'hello'
    Model.model_validate({'b': ['hello', 'world']}).x == 'world'
    with pytest.raises(...):
        Model.model_validate({'b': ['hello']})  # should error
    ...
```

dmontagu (approved):
Looks good, I definitely think this is an improvement.

I'll defer to you about whether to accept/reject the comments I've made.

dmontagu:
please update (just for hooky — no need to update if you reject the comments)

pydantic/fields.py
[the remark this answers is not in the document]
hramezani: Yes, exactly.
Renaming `paths` to `path` and using `@slots_dataclass` make sense. I will update

pydantic/fields.py
hramezani: Done

pydantic/fields.py
```diff
@@ -340,13 +340,35 @@ def __repr_args__(self) -> ReprArgs:
...
+
+
+class AliasPathChoices:
```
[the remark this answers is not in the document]
hramezani: Done

pydantic/fields.py
```diff
@@ -340,13 +340,35 @@ def __repr_args__(self) -> ReprArgs:
                     yield s, value
 
 
-def Field(
+class AliasPath:
```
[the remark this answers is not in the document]
hramezani: Done

pydantic/fields.py
hramezani: renamed to `path`

hramezani:
@dmontagu Thanks for the review. I've addressed all your comments.

pydantic/fields.py
```diff
@@ -340,13 +341,41 @@ def __repr_args__(self) -> ReprArgs:
                     yield s, value
 
 
-def Field(
+@slots_dataclass
+class AliasPath:
+    path: list[int | str]
+
+    def __init__(self, first_arg: str, *args: str | int) -> None:
+        self.path = [first_arg] + list(args)
```
Kludex, line 349: Should we have used positional parameter here or `__first_arg`? @hramezani

PR #5503, opened 2023-04-17
Fix / rewrite xfail tests in docs

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -693,10 +693,10 @@ In order to declare a generic model, you perform the following steps:
 
 Here is an example using `GenericModel` to create an easily-reused HTTP response payload wrapper:
 
-```py test="xfail looks like an error with generics!"
+```py
 from typing import Generic, List, Optional, TypeVar
 
-from pydantic import BaseModel, Field, ValidationError, field_validator
+from pydantic import BaseModel, ValidationError
 
 DataT = TypeVar('DataT')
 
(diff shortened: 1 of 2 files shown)


docs/usage/models.md
```diff
@@ -713,29 +713,28 @@ class DataModel(BaseModel):
...
     data: Optional[DataT] = None
-    error: Optional[Error] = Field(validate_default=True)
-
-    @field_validator('error')
-    def check_consistency(cls, v, info):
-        if v is not None and info.data['data'] is not None:
-            raise ValueError('must not provide both data and error')
-        if v is None and info.data.get('data') is None:
-            raise ValueError('must provide data or error')
-        return v
```
adriangb, line 724: This validate default stuff doesn't have anything to do with generic models (the docs section). I'm simplifying the example by removing it.

docs/usage/schema.md
```diff
@@ -695,52 +695,52 @@ class Model(BaseModel):
...
-
-print(Person.schema_json(indent=2))
-```
-
-
-For more fine-grained control, you can alternatively set `schema_extra` to a callable and post-process the generated schema.
-The callable can have one or two positional arguments.
-The first will be the schema dictionary.
-The second, if accepted, will be the model class.
-The callable is expected to mutate the schema dictionary *in-place*; the return value is not used.
```
adriangb, line 738: Since it now always has to be a callable having two sections doesn't make sense.

adriangb:
please review

docs/usage/models.md
dmontagu: I liked this example for the sake of showing how to build a type with a "common"/"universal" error field, and generic "data" field, as one approach to getting consistent error handling in a web API. (Essentially functioning as a field-name-tagged union.) (I vaguely recall having written this example myself, maybe I'm misremembering..)

I liked the example as I felt like it showed a bit more complex/"real world" usage. But perhaps the complexity distracts from the main point.

Either way, I think we should eventually make a point of having more "intermediate" level examples somewhere in the docs. But I guess that's something @tpdorsey will be tracking.

dmontagu (approved):

PR #5504, opened 2023-04-17
Merge model_modify_json_schema and __modify_pydantic_json_schema__

adriangb:


pydantic/main.py
```diff
@@ -424,7 +424,7 @@ def model_json_schema(
         return model_json_schema(cls, by_alias=by_alias, ref_template=ref_template, schema_generator=schema_generator)
 
     @classmethod
-    def model_modify_json_schema(cls, json_schema: JsonSchemaValue) -> JsonSchemaValue:
+    def __pydantic_modify_json_schema__(cls, schema: dict[str, Any]) -> dict[str, Any]:
```
dmontagu, line 427: I had added JsonSchemaValue so that we could eventually have a better type than just `dict[str, Any]` (even if it was just a recursive type equivalent to what a JsonValue should be).

Can drop that if you think it makes more sense to just use `dict[str, Any]`, but I would suggest if we drop it here that we drop it everywhere.

PR #5506, opened 2023-04-17
No case enum

### LINKED ISSUE #5495: V2 doesn't support bound enum generics
### Initial Checks

### Description

Provided code example raises the following error:
```
pydantic_core._pydantic_core.SchemaError: Error building "model" validator:
  SchemaError: Error building "typed-dict" validator:
  SchemaError: Field "bar":
  SchemaError: Error building "lax-or-strict" validator:
  SchemaError: Error building "function-after" validator:
  SchemaError: Error building "literal" validator:
  SchemaError: "expected" should have length > 0
```

### Example Code

```Python
from enum import Enum

T = TypeVar('T', bound=Enum)

class Foo(BaseModel, Generic[T]):
    bar: T
```

```Text
pydantic version: 2.0a2
        pydantic-core version: 0.24.0 release build profile
                 install path: C:\Program Files\Python\Lib\site-packages\pydantic
               python version: 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)]
                     platform: Windows-10-10.0.22621-SP0
```

dmontagu:
Closes #5495, and resolves some other bugs with validation of plain enum.Enum fields, and validation of all enums from JSON.

### DIFF
--- a/pydantic/_internal/_std_types_schema.py
+++ b/pydantic/_internal/_std_types_schema.py
@@ -17,6 +17,7 @@ from typing import Any, Callable
 from uuid import UUID
 
 from pydantic_core import CoreSchema, MultiHostUrl, PydanticCustomError, Url, core_schema
+from pydantic_core.core_schema import ValidatorFunctionWrapHandler
 from typing_extensions import Literal, get_args
 
 from ..json_schema import update_json_schema
@@ -63,6 +64,15 @@ def timedelta_schema(_schema_generator: GenerateSchema, _t: type[Any]) -> core_s
 
 @schema_function(Enum)
 def enum_schema(_schema_generator: GenerateSchema, enum_type: type[Enum]) -> core_schema.CoreSchema:
+    cases = [m.value for m in enum_type.__members__.values()]
+    if not cases:
+        # Use an isinstance check for enums with no cases.
+        # This won't work with serialization or JSON schema, but that's okay -- the most important
+        # use case for this is creating typevar bounds for generics that should be restricted to enums.
+        # This is more consistent than it might seem at first, since you can only subclass enum.Enum
+        # (or subclasses of enum.Enum) if all parent classes have no cases.
+        return core_schema.is_instance_schema(enum_type)
+
     def to_enum(__input_value: Any, _: core_schema.ValidationInfo) -> Enum:
         try:
             return enum_type(__input_value)
@@ -70,9 +80,7 @@ def enum_schema(_schema_generator: GenerateSchema, enum_type: type[Enum]) -> cor
             raise PydanticCustomError('enum', 'Input is not a valid enum member')
 
     enum_ref = get_type_ref(enum_type)
-    literal_schema = core_schema.literal_schema(
-        [m.value for m in enum_type.__members__.values()],
-    )
+    literal_schema = core_schema.literal_schema(cases)
     description = None if not enum_type.__doc__ else inspect.cleandoc(enum_type.__doc__)
     if description == 'An enumeration.':  # This is the default value provided by enum.EnumMeta.__new__; don't use it
         description = None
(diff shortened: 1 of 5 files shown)


dmontagu:
please review

pydantic/_internal/_std_types_schema.py
```diff
@@ -83,29 +91,37 @@ def to_enum(__input_value: Any, _: core_schema.ValidationInfo) -> Enum:
...
     else:
-        lax = core_schema.general_after_validator_function(to_enum, literal_schema, metadata=metadata)
-        json_type = 'str'
+
+        def wrap_to_enum(
+            __input_value: Any, __handler: ValidatorFunctionWrapHandler, _: core_schema.ValidationInfo
+        ) -> Enum:
+            if isinstance(__input_value, enum_type):
+                __input_value = __input_value.value
+            value = __handler(__input_value)
```
adriangb, line 116: What would `__handler` ever represent? Is there a test I can look at that demonstrates why this needs to be a wrap handler and not an `isinstance` handler or something?

My understanding is that we want `class MyEnum({int,str,float}, Enum)` to work w/ JSON but anything else would work w/ Python but not JSON.

Another thing is that I think that Enums should ~ always be strict:

```python
from enum import Enum, auto

class MyEnum2(Enum):
    x = auto()

AnalyzedType(MyEnum1).validate_python(1)  # fails

class MyEnum2(Enum):
    x = 1

AnalyzedType(MyEnum2).validate_python(1)  # fails
```

But that just be my opinion.

tests/test_types.py
```diff
@@ -1341,18 +1342,130 @@ def test_int_enum_successful_for_str_int(cooking_model):
...
 
+def test_plain_enum_validate():
+    class MyEnum(Enum):
+        a = 1
+
+    class Model(BaseModel):
+        x: MyEnum
+
+    m = Model(x=MyEnum.a)
+    assert m.x == MyEnum.a
```
adriangb, line 1353:
```suggestion
    m = Model(x=MyEnum.a)
    assert m.x is MyEnum.a
    
    with pytest.raises(ValidationError) as exc_info:
        Model.validate_python({'x': 1})
```

tests/test_types.py
```diff
@@ -1341,18 +1342,130 @@ def test_int_enum_successful_for_str_int(cooking_model):
...
+
+def test_plain_enum_validate_json():
+    class MyEnum(Enum):
+        a = 1
+
+    class Model(BaseModel):
+        x: MyEnum
+
+    m = Model.model_validate_json('{"x":1}')
+    assert m.x == MyEnum.a
```
adriangb, line 1364:
```suggestion
    m = Model.model_validate_json('{"x":1}')
    assert m.x is MyEnum.a
```

I'm suggesting using `is` because `MyEnum.a == 1` is True (I think).

pydantic/_internal/_std_types_schema.py
dmontagu: I was using `__handler` to get it to do the check that the value is in the allowed literal values. What's happening here is that it is unpacking the enum into its value, validating the value, then repackaging it into the Enum.

I think the reason this was necessary now was because the literal thing was using values and not Enums. Maybe that just doesn't need to be the case.

Also I realized you are right it should probably just always be strict. I'll try reworking this.

tests/test_types.py
dmontagu: == fails for plain enums like in this case, but I agree we should do the `is` since we can.

samuelcolvin (approved):
LGTM.

Feel free to merge if you're happy.

pydantic/_internal/_std_types_schema.py
dmontagu: Making enum parsing always strict is a significant breaking change so I'm not going to do that here, but I would be happy to make that change at some point if we establish broader consensus.

PR #5516, opened 2023-04-18
add codspell to pre-commit

samuelcolvin:
replace #5507.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,10 @@ lint:
 	ruff $(sources)
 	black $(sources) --check --diff
 
+.PHONY: codespell
+codespell:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -27,6 +27,10 @@ lint:
 	ruff $(sources)
 	black $(sources) --check --diff
 
+.PHONY: codespell
+codespell:
```
dmontagu, line 31: this is going to throw my muscle memory for a loop after typing codspeed enough times to get used to not including the 'e' 😄

samuelcolvin: https://twitter.com/samuel_colvin/status/1625468736277815299

PR #5533, opened 2023-04-19
Add an initial stab at recursive types

dmontagu:
This was inspired by some random investigations Adrian and I were doing, and I realized there should be a small-amount-of-code way to get this working.

There are a couple minor things that need to be addressed (i.e., what to do if you get a schema you can't handle, I'm not sure if that's even possible though), but I think the biggest issue is the horrific validation error messages for nontrivial cases (see the test related to `Json`).

Not sure if this is destined for inclusion but I though it was cool I could get it working with so little code and figured I'd at least share the implementation.


tests/test_recursive_type.py
```diff
@@ -0,0 +1,159 @@
...
+from typing_extensions import Annotated, TypeAlias
+
+from pydantic import BaseModel, RecursiveType
+
+Json: TypeAlias = list['Json'] | dict[str, 'Json'] | str | int | float | bool | None
+
+
+def test_recursive_type():
+    class Model(BaseModel):
+        x: Annotated[Json, RecursiveType('Json')]
```
dmontagu, line 12: The reason the `RecursiveType('Json')` thing is required is because you have to specify the name _somewhere_ that pydantic can introspect.

dmontagu:
Closing for now, but having investigated this some, I think we should require an annotation that represents a TypeAlias with similar semantics to how it will work with the generic syntax PEP (I think that's the one that adds the soft `type` keyword).

PR #5538, opened 2023-04-20
Change errors URL for next release

dmontagu:


tests/test_errors.py
```diff
@@ -1,14 +1,19 @@
...
 
 
 def test_user_error_url():
     with pytest.raises(PydanticUserError) as exc_info:
         BaseModel()
 
     # insert_assert(str(exc_info.value))
+    # assert str(exc_info.value) == (
+    #     'Pydantic models should inherit from BaseModel, BaseModel cannot be instantiated directly\n\n'
+    #     f'For further information visit https://errors.pydantic.dev/{VERSION}/u/base-model-instantiated'
```
dmontagu, line 13: I kept this here commented just to make sure when it gets fixed later, the version doesn't end up hard-coded.

pydantic/errors.py
```diff
@@ -22,7 +21,8 @@
 # We use this URL to allow for future flexibility about how we host the docs, while allowing for Pydantic
 # code in the while with "old" URLs to still work.
 # 'u' refers to "user errors" - e.g. errors caused by developers using pydantic, as opposed to validation errors.
-DEV_ERROR_DOCS_URL = f'https://errors.pydantic.dev/{VERSION}/u/'
+# DEV_ERROR_DOCS_URL = f'https://errors.pydantic.dev/{VERSION}/u/'
+DEV_ERROR_DOCS_URL = '<TODO: Set up the errors URLs>/'
```
dmontagu, line 25: @samuelcolvin if you want this to work differently it's totally fine with me, feel free to do a "suggestion" or whatever.

I would suggest we keep it functioning mostly the same way though (i.e., including the URL endings, even if the initial portion is not a valid URL), just so we don't have to change the implementation much.

PR #5553, opened 2023-04-22
Fix typing_extension version

Dreamsorcerer:
`@deprecated` doesn't exist in older versions:
https://github.com/pydantic/pydantic/blob/main/pydantic/deprecated/config.py#L6

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ classifiers = [
 ]
 requires-python = '>=3.7'
 dependencies = [
-    'typing-extensions>=4.2.0',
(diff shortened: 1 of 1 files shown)


yezz123 (approved):

yezz123:
duplicated https://github.com/pydantic/pydantic/pull/5552

samuelcolvin:
Same as #5552, but happy to accept whichever PR fixes this.

You need to update the requirements pinning files, you can either run `./requirements/refresh.sh`, or just use find and replace.

Please update.

Dreamsorcerer:
Looks like that's done in the other PR already.

PR #5570, opened 2023-04-25
Json schema modify function as middleware

### LINKED ISSUE #5520: Rename `__pydantic_modify_json_schema__` to `__get_pydantic_json_schema__`
Just creating the issue for tracking purposes. It can be done after refactoring `__pydantic_modify_json_schema__` to be more middleware-esque.

adriangb:
Closes #5520

Selected Reviewer: @dmontagu


tests/test_json_schema.py
```diff
@@ -2378,15 +2389,28 @@ def __get_pydantic_core_schema__(
...
 
         @classmethod
-        def __pydantic_modify_json_schema__(cls, field_schema: core_schema.CoreSchema) -> core_schema.CoreSchema:
+        def __get_pydantic_json_schema__(
+            cls, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+        ) -> JsonSchemaValue:
+            try:
+                field_schema = handler(core_schema)
+            except PydanticInvalidForJsonSchema:
+                field_schema = {}
```
adriangb, line 2413: This now fails for the case where `Gen` is not parametrized (as per above it returns an isinstance schema)

tests/test_json_schema.py
```diff
@@ -2378,15 +2389,28 @@ def __get_pydantic_core_schema__(
...
+
+                def js_func(s, h):
+                    # ignore the schema we were given and get a new CoreSchema
+                    s = handler(Optional[arg])
+                    return h(s)
+
                 return core_schema.general_plain_validator_function(
                     Gen,
-                    metadata=metadata,
+                    metadata={'pydantic_js_functions': [js_func]},
```
adriangb, line 2401: This is one area of the API that is still kludgy: `build_metadata_dict` used in this test is private; we shouldn't use private stuff in tests. Can we make `metadata` a `TypedDict` with `total=False` or something?

tests/test_json_schema.py
```diff
@@ -2378,15 +2389,28 @@ def __get_pydantic_core_schema__(
...
         ) -> core_schema.CoreSchema:
             if hasattr(source, '__args__'):
-                param = source.__args__[0]
-                metadata = build_metadata_dict(js_cs_override=handler(Optional[param]))
+                arg = source.__args__[0]
+
+                def js_func(s, h):
+                    # ignore the schema we were given and get a new CoreSchema
+                    s = handler(Optional[arg])
+                    return h(s)
```
adriangb, line 2397: Rewrote just for clarify of wtf is going on

tests/test_json_schema.py
```diff
@@ -2357,7 +2368,7 @@ class LocationBase(BaseModel):
     }
 
 
-def test_advanced_generic_schema():
+def test_advanced_generic_schema():  # noqa: C901
```
adriangb, line 2371: yeah this test is a beast

docs/usage/schema.md
```diff
@@ -190,10 +202,10 @@ print(schema_json_of(Pet, title='The Pet Schema', indent=2))
       "required": [
         "pet_type",
         "cat_name"
-      ]
+      ],
+      "title": "Cat"
```
adriangb, line 206: This seems like it just moved from above? Which makes sense since (1) we may be inserting things into the dict in different order and (2) the examples are in strings so if `print()` is different they'll be "updated"

docs/usage/schema.md
```diff
@@ -779,12 +792,7 @@ print(json.dumps(Person.model_json_schema(), indent=2))
     "name",
     "age"
   ],
-  "examples": [
-    {
-      "name": "John Doe",
-      "age": 25
-    }
```
adriangb, line 786: This I need to look into

adriangb:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -1316,15 +1357,15 @@ def get_first_arg(type_: Any) -> Any:
         return Any
 
 
-def _get_pydantic_modify_json_schema(obj: Any) -> typing.Callable[[JsonSchemaValue], JsonSchemaValue] | None:
-    js_modify_function = getattr(obj, '__pydantic_modify_json_schema__', None)
+def _get_pydantic_modify_json_schema(obj: Any) -> GetJsonSchemaFunction | None:
```
dmontagu, line 1360: Could we rename this function to `_extract_get_pydantic_json_schema` or similar, I think it will help make it clearer to remove the word `modify` here, as I think that's a vestige of `__pydantic_modify_json_schema__`

pydantic/_internal/_json_schema_shared.py
```diff
@@ -0,0 +1,92 @@
...
+
+from pydantic_core import core_schema
+
+JsonSchemaValue = Dict[str, Any]
+CoreSchemaOrField = Union[core_schema.CoreSchema, core_schema.DataclassField, core_schema.TypedDictField]
+
+
+class GetJsonSchemaHandler:
+    """
+    Handler to call into the next JSON schema generation function
```
dmontagu, line 16: I realize this is re-exported and intended to be public, but I think if I were to jump-to-definition and got brought to this file I might feel like I lost my bearings a bit and wonder why I'm in an internal file.

Because of that, I think it would be helpful to give this class a little bit more detailed of a docstring, basically describing how it is meant to be used (and possibly with some example usage?), and reinforcing that it is meant for "users" to use.

Happy to help draft something for this, I think we should update it a bit before merging so we don't forget, but it doesn't have to be "final draft" quality.

dmontagu: Maybe I'm misunderstanding how this should be used, I didn't realize it was an abstract class.

If it doesn't need to be abstract I think it might be better to not have it be this way as I think it increases the learning curve. (By that I mean, delete this class and change the name of `UnpackedRefJsonSchemaHandler` to the name this class currently has, then people can just subclass and override. I think that would help with discovering the implementation, etc.)

Or maybe I'm misunderstanding something about how this is meant to work?

pydantic/_internal/_json_schema_shared.py
```diff
@@ -0,0 +1,92 @@
...
+        original_schema = self.resolve_ref_schema(self.original_schema)
+        if original_schema is not self.original_schema and schema is not original_schema:
+            # a new schema was returned
+            original_schema.clear()
+            original_schema.update(schema)
+        # return self.original_schema, which may be a ref schema
+        return self.original_schema
+
+
+def wrap_json_schema_fn_for_model_or_custom_type_with_ref_unpacking(
```
dmontagu, line 83: nice name lol

tests/test_json_schema.py
dmontagu: I think metadata has the same problem as the ConfigDict — when I implemented this, I thought that using a typeddict directly meant that users/libraries/etc. wouldn't be able to add fields. But I think we realized we could solve that through having people create subclasses. So yeah, maybe that's a better way to do it, but I'll just note that I don't know how well it would work if trying to use two packages together that have custom metadata (maybe you can do diamond inheritance?)

At any rate, I'm in favor of eliminating the build_metadata_dict and using a typeddict, and just changing that API in the future if we need to.

tests/test_json_schema.py
dmontagu: I think this came from Sebastian originally lol

pydantic/_internal/_json_schema_shared.py
dmontagu: Okay I see you do use two different subclasses in internal code so maybe it's more confusing to give this a default implementation.

In that case, I think we should direct people to the two implementations (and possibly move the json schema one into this file? Not sure if that's possible.)

The main reason being I think it will be annoying for people to jump to the implementation while trying to debug code or whatever. (I personally hate when I am told that something matches an interface and then I can't jump to the definition, but if necessary, I'd hope we could at least make it easier to jump to the definition I care about.)

pydantic/_internal/_generate_schema.py
```diff
@@ -143,6 +154,22 @@ def apply_each_item_validators(
     return schema
 
 
+def modify_json_schema_model(
```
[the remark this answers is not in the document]
adriangb: I added a default implementation on `BaseModel`, but I kept the logic here.

pydantic/_internal/_json_schema_shared.py
adriangb: Moved them all into a single file.

Kludex:
It would be cool to mention an issue or add a description here. 👀

tests/test_json_schema.py
```diff
@@ -3459,3 +3492,124 @@ class OuterModel(BaseModel):
...
     }
+
+
+@pytest.mark.xfail(
+    reason=(
+        'We are calling __get_pydantic_json_schema__ too many times.'
+        ' The second time we analyze a model we get the CoreSchema from __pydantic_core_schema__.'
+        ' But then we proceed to append to the metadata json schema functions.'
+    )
+)
```
adriangb, line 3503: This equivalent test shows that this is a bug in `main` that I just can't fix in this PR:

```python
from typing import Any, Dict, List

from pydantic.json_schema import JsonSchemaValue

calls: List[str] = []

class Model(BaseModel):
    @classmethod
    def __modify_schema__(cls, schema: Dict[str, Any]) -> JsonSchemaValue:
        calls.append('x')
        schema['title'] += 'x'
        return schema


class OuterModel(BaseModel):
    x: Model


assert OuterModel.model_json_schema()['$defs']['Model']['title'] == 'Modelxx'
assert calls == ['x', 'x']
```

PR #5576, opened 2023-04-25
Use only relative imports

Kludex:
The idea here is mainly to make it easier to create `pydantic-next`, and have a single way to do _importing_.


Kludex:
please review

samuelcolvin:
thanks.

Issue #1465, opened 2020-04-30
`validate_arguments` skip arguments.


juan-carlos-calvo:
# Feature Request

`validate_arguments` validates all arguments by default. I would love to be able to choose which ones to validate, e.g.:
```
@validate_arguments(['arg2', 'arg3'])
def myfunc(arg1: int, arg2: float, arg3: MyType):

```

samuelcolvin:
happy to consider this, question is how complicated it would be to implement.

juan-carlos-calvo:
probably not that complicated. I'm writing this decorator myself right now. I'm basically applying `validate_arguments` on a partial version of the decorated function which only takes as input the arguments of interest. The only "challenging" part is to parse the arguments of the function to do so.

yosoycentli:
Hi there, I would like to know how this issue was going.
Maybe I can contribute.

smhavens:
Is this issue still open to work on? I would be interested if so!

ybressler:
I've taken a look at where I think changes would need to be introduced to make this happen, specifically: [pydantic/decorator.py](https://github.com/pydantic/pydantic/blob/2e17c85810fd7e335422dcbf2525d019cfeb5d8c/pydantic/decorator.py#L225)

Introducing this new behavior introduces some ambiguity, for example, say I have a function with required positional arguments:
```
def do_things(required_field: str, optional_field: Optional[str] = None):
    pass
```
Applying the following values: `@validate_arguments(['optional_field'])` basically means I'm turning my `required_field` into an optional one, which may validate an invalid function. (Meaning, the validator will pass but the function will encounter a `TypeError` Unless this is desired?)

ybressler:
AHA. I'm reading a bit further, perhaps the correct place to override this behavior is within the `config` argument of the validator! https://docs.pydantic.dev/usage/validation_decorator/#custom-config

I'll go deeper here. This is fun.

dmontagu:
In v2, the right way to do this would be to create an annotation that causes it to skip validation by using a pydantic core any schema. This would be done via:

```python
from typing import Any, Annotated

from pydantic_core import core_schema

from pydantic import validate_call


class SkipValidation:
    @classmethod
    def __get_pydantic_core_schema__(self, _source: Any, _handler: Any) -> core_schema.CoreSchema:
        return core_schema.any_schema()


@validate_call
def my_function(x: int, y: Annotated[int, SkipValidation]) -> None:
    print(repr(x), repr(y))


[...]

PR #5583, opened 2023-04-25
Add annotation that lets you skip validation

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/1465

Selected Reviewer: @adriangb


Kludex (approved):

pydantic/annotated_arguments.py
```diff
@@ -105,3 +105,16 @@ def __get_pydantic_core_schema__(
...
+    """
+    If this is applied as an annotation (e.g., via `x: Annotated[int, SkipValidation]`), validation will be skipped.
+
+    This can be useful if you want to use a type annotation for documentation/IDE/type-checking purposes,
+    and know that it is safe to skip validation for one or more of the fields.
+    """
+
+    @classmethod
+    def __get_pydantic_core_schema__(self, _source: Any, _handler: Any) -> core_schema.CoreSchema:
+        return core_schema.any_schema()
```
adriangb, line 120:
```suggestion
    @classmethod
    def __get_pydantic_core_schema__(cls, _source: Any, _handler: Any) -> core_schema.CoreSchema:
        return core_schema.any_schema()
```

adriangb (approved):

adriangb:
How does this work with JSON schema and serialization? Is there a world where we want:
- To preserve the JSON schema or serializer of the original type but skip validation?
- Let users specify a JSON schema or serializer to accompany this AnyValidator?

samuelcolvin:
JSON schema should be be preserved, e.g. generated from the inner type. 

I think we should use the serialiser from the inner type too. We could add a switch for this, but it could come later.

dmontagu:
@adriangb do you see how to make the JSON schema thing work in this way now that the "js_cs_override_schema" metadata thing is gone? I'm not sure what exactly the current signature is for json schema generation, or if there is a good way to handle this now.

@samuelcolvin is there a way to take a "regular" core schema and use it for serialization? Like, ideally I'd be able to do something like:
```python
    @classmethod
    def __get_pydantic_core_schema__(self, _source: Any, _handler: Any) -> core_schema.CoreSchema:
        schema = _handler(_source)
        return core_schema.any_schema(serialization=schema)
```
but I think serialization needs to specifically be a `SerSchema`. Maybe we need to do a rework to this.

Either way, I think this can wait until after v2, though @adriangb I think it would be good for us to make sure it's possible to handle the JSON schema for this without needing to make breaking changes. (I guess we can always go back to using new core schema metadata keys if we need to, so it doesn't need to be a breaking change, but.. I guess it would be better to avid that)

dmontagu:
please review

adriangb:
I think you’ll want `__prepare_pydantic_annotations__` so that you can still create a json schema.

dmontagu:
please update

dmontagu:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -742,7 +742,7 @@ def _generate_parameter_schema(
         else:
             field = FieldInfo.from_annotated_attribute(annotation, default)
         assert field.annotation is not None, 'field.annotation should not be None when generating a schema'
-        source_type, annotations = annotation, field.metadata
+        source_type, annotations = field.annotation, field.metadata
```
dmontagu, line 745: even if there are issues merging the rest of this, we should probably make this fix either way.

adriangb (approved):

Issue #5597, opened 2023-04-26
Unexpected behavior for `Sequence[int]` in Json Schema


lig:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

- [X] Using `Sequence[int]` for json is inconsistent.
- [ ] `BaseModel.validate_json` should raise if `BaseModel.model_json_schema` raises. See below

Generating json schema via `BaseModel.model_json_schema()` raises `PydanticInvalidForJsonSchema` error which seems as a correct behavior.

However, calling `BaseModel.validate_json(...)` results in an attempt to validate passed json value while the validation schema is invalid.

### Example Code

```Python
>>> from pydantic import BaseModel
>>> from typing import Sequence
>>> class Model(BaseModel):
...     int_sequence: Sequence[int]
...     int_list: list[int]
... 
>>> Model.model_json_schema()
pydantic.errors.PydanticInvalidForJsonSchema: Cannot generate a JsonSchema for core_schema.IsInstanceSchema (typing.Sequence)

>>> Model.model_validate_json('{"int_sequence": [1, 2, 3], "int_list": [1, 2, 3]}')
pydantic_core._pydantic_core.ValidationError: 1 validation error for Model
int_sequence
  Input should be an instance of Sequence [type=is_instance_of, input_value=[1, 2, 3], input_type=list]
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0a3
        pydantic-core version: 0.25.0 release build profile
               python version: 3.10.11 (main, Apr  5 2023, 00:00:00) [GCC 13.0.1 20230404 (Red Hat 13.0.1-0)]
                     platform: Linux-6.2.12-300.fc38.x86_64-x86_64-with-glibc2.37
     optional deps. installed: ['devtools', 'email-validator', 'typing-extensions']
```

lig:
Reopening as one test in #5598 is still failing

dmontagu:
@lig I'm not sure that test makes sense — see [this comment](https://github.com/pydantic/pydantic/pull/5598#issuecomment-1548507790) for more detail.

I think this might be coming from a misconception — the generated JSON Schema has nothing to do with actual validation, it is just a generated thing for use by web frameworks, etc.

So it's generally always possible that you have an invalid JSON schema and yet can successfully validate things, or vice versa, that you have a valid JSON schema but cannot actually validate anything. (Validation is done by the pydantic core schema.)

[...]

PR #5598, opened 2023-04-26
✅ Add tests illustrating #5597

lig:
## Change Summary

Add tests illustrating `Sequence[int]` behavior when using with json. 

## Related issue number

See #5597

## Checklist

### DIFF
--- a/tests/test_json_schema.py
+++ b/tests/test_json_schema.py
@@ -4039,3 +4039,15 @@ def test_sequences_int_json_schema(sequence_type):
         'required': ['int_seq'],
     }
     assert Model.model_validate_json('{"int_seq": [1, 2, 3]}')
+
+
+def test_validate_json_raises_if_json_schema_raises():
+    class Model(BaseModel):
+        int_seq: Sequence[int]
+
+    with pytest.raises(PydanticInvalidForJsonSchema) as e_json_schema:
+        Model.model_json_schema()
+    with pytest.raises(PydanticInvalidForJsonSchema) as e_validate_json:
+        Model.model_validate_json('{}')
+
+    assert e_json_schema.value.errors() == e_validate_json.value.errors()


samuelcolvin:
is this still needed? If so please update.

lig:
One test has been copied to #5706. Other two are still relevant after the rebase.

dmontagu:
@lig I only see one test left here, and as currently written, I don't think it makes sense:

```python
def test_validate_json_raises_if_json_schema_raises():
    class Model(BaseModel):
        int_seq: Sequence[int]

    with pytest.raises(PydanticInvalidForJsonSchema) as e_json_schema:
        Model.model_json_schema()
    with pytest.raises(PydanticInvalidForJsonSchema) as e_validate_json:
        Model.model_validate_json('{}')

    assert e_json_schema.value.errors() == e_validate_json.value.errors()
```

`Model.model_json_schema()` would be capable of raising a `PydanticInvalidForJsonSchema` (though it no longer does in this specific case). But `Model.model_validate_json('{}')` would raise a `ValidationError`. In particular, I don't think there's any world where the json schema errors would be equal to the errors arising from attempted validation. So I'm not sure what this final test is trying to demonstrate.

lig:
@dmontagu citing #5597 :

> Generating json schema via BaseModel.model_json_schema() raises PydanticInvalidForJsonSchema error which seems as a correct behavior.
> 
> However, calling BaseModel.validate_json(...) results in an attempt to validate passed json value while the validation schema is invalid.

So, It just doesn't make sense to me attempting to validate data using the schema that already has errors. What schema are we using for that?

lig:
> (Also, we do now generate a valid JSON schema for this model, which I think is reasonable, so you won't get a PydanticInvalidForJsonSchema anyway)

Oh, I didn't notice that. I'll update the test.

dmontagu:
> What schema are we using for that?

The pydantic core schema (available in `Model.__pydantic_core_schema__`). Technically, we use a slightly post-processed version of that schema during validation and serialization, as you can see here: 

https://github.com/pydantic/pydantic/blob/7df75059df9746a398a188553b88724e3df5c44f/pydantic/_internal/_model_construction.py#L204-L209

lig:
Ok. Let's close this and #5597 then. It just a naming confusion, I guess.

PR #5609, opened 2023-04-27
Make it possible to have a plain BaseModel field annotation

dmontagu:
Right now, if you try to use a plain BaseModel as a field, you get:

```python
  File "/Users/davidmontague/Programming/pydantic/pydantic/pydantic/_internal/_generate_schema.py", line 245, in model_schema
    fields = cls.model_fields
             ^^^^^^^^^^^^^^^^
AttributeError: type object 'BaseModel' has no attribute 'model_fields'. Did you mean: 'model_fields_set'?
```

Found an example of this here: https://github.com/gdsfactory/gdsfactory/blob/f3daacc74de6e59ad6b550d7daa5b34f142c0ba9/gdsfactory/technology/layer_views.py#L762

Doesn't seem unreasonable to use `BaseModel` as a field annotation in principle (though obviously the serialization may suffer..), and I just needed to set two of the class attributes to get this to stop erroring.

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -214,6 +214,8 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         __pydantic_generic_metadata__: typing.ClassVar[_generics.PydanticGenericMetadata]
         __pydantic_parent_namespace__: typing.ClassVar[dict[str, Any] | None]
     else:
+        # `model_fields` and `__pydantic_decorators__` must be set for
+        # pydantic._internal._generate_schema.GenerateSchema.model_schema to work for a plain BaseModel annotation
         model_fields = {}
         __pydantic_decorators__ = _decorators.DecoratorInfos()
         __pydantic_validator__ = _model_construction.MockValidator(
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

I'd like to get a release out ASAP tomorrow so that `gdsfactory` can see what the next problem they hit with pydantic v2 is; this is the first thing their CI is erroring on. So if this is acceptable, it would be great if it could be merged before I wake up so we can do the next alpha release (also with pydantic-core 0.27.0 hopefully) in the (US) morning.

pydantic/main.py
```diff
@@ -214,6 +214,8 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         __pydantic_generic_metadata__: typing.ClassVar[_generics.PydanticGenericMetadata]
         __pydantic_parent_namespace__: typing.ClassVar[dict[str, Any] | None]
     else:
+        model_fields = {}
+        __pydantic_decorators__ = _decorators.DecoratorInfos()
```
adriangb, line 218: A comment here explaining might be helpful? If I didn't know this was part of the PR I wouldn't think twice about this, but I cannot imagine how this fixes the problem (or why `model_fields` was being accessed in the first place).

adriangb (approved):

pydantic/main.py
dmontagu: We could use getattr there instead, but it seems to me better to have the type annotation be correct than to lie twice. But if it's possible for the class attributes to cause issues then we can keep them "missing" and just getattr in the `_generate_schema.model_schema`

PR #5610, opened 2023-04-27
Use PDM for lockfiles & env management

adriangb:
This replaces our scripted together requirements management with [PDM](https://pdm.fming.dev/latest/). Our current system has several limitations and can lead to upsies (see https://github.com/pydantic/pydantic/commit/fb2944808b2b216307fccd5babf11c44bfed55eb#diff-3362b72813f83c18d0f1be4e6039d1a1ddc4e16b1324c2a82700d74b990011f9).

I chose PDM because it can work with hatchling as the build backend, which minimizes the number of things that need to change in this PR (vs. something like Poetry which would require a lot more churn).


pydantic/networks.py
```diff
@@ -325,7 +325,10 @@ def validate_email(value: str) -> tuple[str, str]:
             'value_error', 'value is not a valid email address: {reason}', {'reason': str(e.args[0])}
         ) from e
 
-    return name or parts['local'], parts['email']
+    email = parts.normalized
+    assert email is not None
+    name = name or parts.local_part
+    return name, email
```
dmontagu, line 331: I'm curious, was this change necessary due to something in this PR? Seems unrelated lol

dmontagu: I guess maybe it was a dependency bump thing

Makefile
```diff
@@ -1,31 +1,36 @@
...
-rebuild-lockfiles:       ## Rebuild lockfiles from scratch, updating all dependencies
-	bash requirements/rebuild.sh
+rebuild-lockfiles: .pdm       ## Rebuild lockfiles from scratch, updating all dependencies
+	pdm lock --group :all --dev
 
 .PHONY: format
 format:
-	black $(sources)
-	ruff --fix $(sources)
+	pdm run black $(sources)
```
dmontagu, line 27: does this mean you have to do `pdm run <command>` any time you want to run something in your dev environment? I guess maybe you are supposed to activate the environment with pdm or whatever and then it's not necessary?

Does the `--no-isolation` flag used below cause it to not create an environment and just use whatever is active?

adriangb: You can still activate the environment and then just run the thing. But `pdm run` does that for you.

dmontagu:
Okay I looked at the docs page for PDM and it looks pretty well-thought-out. I do think it's a good idea to rework the dependency management stuff, given the "upsies" lol

Makefile
adriangb: > Does the --no-isolation flag used below cause it to not create an environment and just use whatever is active?

I believe so. I guess we'll see when that cron job CI file runs.

adriangb:
I'll say that while PDM seems well thought out, it is opinionated and seems to favor correctness over user friendliness. The docs are also not great in my opinion, they tell you an option exists but not how to use it or what it really means. Anyhow, it works.

pydantic/networks.py
adriangb: Yeah. I accidentally bumped the locked version (because I recreated all lockfiles) and figured I might as well fix this so we can bump it before 2.0 and not have to support multiple versions. But I'm also happy to move that to another PR.

adriangb:
please review

adriangb:
I also think there's something to be said for using a single tool (PDM?) for managing both the build backend, publishing and dependencies/lockfiles.

pydantic/networks.py
dmontagu: I'm good with fixing it here, easiest to just do it in one go.

samuelcolvin:
In principle I'm good with this - nice to have a solution to the requirements mess.

Can we update the contributions section to explain how to use this.

I'll try this out today.

Makefile
```diff
@@ -1,71 +1,76 @@
 .DEFAULT_GOAL := all
 sources = pydantic tests docs/plugins
 
+.PHONY: .pdm
+.pdm:  ## Check that PDM is isntallced
+	@pdm -V || echo 'Please install PDM: https://pdm.fming.dev/latest/\#installation'
+
+.PHONY: .pre-commit
+.pre-commit:  ## Check that PDM is isntallced
```
hramezani, line 9:
```suggestion
.pre-commit:  ## Check that pre-commit is installed
```

docs/contributing.md
```diff
@@ -40,29 +40,27 @@ You'll need to have a version between **Python 3.7 and 3.11**, **virtualenv**, *
...
-source env/bin/activate
-# Building docs requires 3.8. If you don't need to build docs you can use
-# whichever version; 3.7 will work too.
+# 2. Install PDM and pre-commit
+# We use pipx here, for other options see:
+# https://pdm.fming.dev/latest/#installation
+# https://pre-commit.com/#install
+# To get pipx itself:
+# https://pypa.github.io/pipx/
+pipx install pdm pre-commit
```
samuelcolvin, line 49: I think simpler to use `python -m venv` and `pip` here since that's the tool most people will understand and have installed.

Makefile
```diff
@@ -1,71 +1,76 @@
...
-rebuild-lockfiles:       ## Rebuild lockfiles from scratch, updating all dependencies
-	bash requirements/rebuild.sh
+rebuild-lockfiles: .pdm       ## Rebuild lockfiles from scratch, updating all dependencies
+	pdm lock --dev --group :all
 
 .PHONY: format
-format:
-	black $(sources)
-	ruff --fix $(sources)
+format: .pdm
```
samuelcolvin, line 26: I don't think we need ` .pdm` on this or later commands - we don't check everything else is installed.

If you do want this behaviour, maybe remove `PHONY` from the `.pdf` recipe and make will check it's been run without calling `pdm`.

docs/contributing.md
samuelcolvin: you could add a comment saying "pipx also works, if that's what you're used to".

samuelcolvin (commented):
otherwise I think this is excellent - I've just tried it and pdm looks great.

(Biggest problem is I keep mis-typing it "pdf")

docs/contributing.md
adriangb: The point is you want to install these tools _outside_ of the venv. We could say `pip install pdm pre-commit` but IMO using `pipx` is much nicer than global installs because of the isolation.

PR #5627, opened 2023-04-28
Backport update to email-validator>=2.0.0.post2

### LINKED ISSUE #3905: specifying default value in Field causes Pylance to expect an argument when calling the model constructor
# Bug

If you put the following snippet in VSCode and use Pylance (a common setup I assume), the below snippet will cause Pylance to complain `Argument missing for parameter value`.

Maybe it's a Pylance issue, maybe not.. I don't know enough, but it would be super awesome if there's some switch you can flip to support this :)

```python
from pydantic import BaseModel, Field, constr

class MyModel(BaseModel):
    value: str = Field('default_value', type=constr, min_length=1, max_length=200)

if __name__ == "__main__":
    a = MyModel()
    print(a)

```

adriangb:
Fixes #3905


adriangb:
@tiangolo looks like we're blocked by this: https://github.com/tiangolo/fastapi/blob/8ac8d70d52bb0dd9eb55ba4e22d3e383943da05c/pyproject.toml#L62

Any chance you could cut a FastAPI patch release that removes the upper version pin?

pydantic/networks.py
```diff
@@ -706,32 +706,34 @@ def validate(cls, value: NetworkType) -> Union[IPv4Network, IPv6Network]:
...
+        assert email is not None
+        name = name or parts.local_part
+        return name, email
+    else:
+        # email-validator >1, <2
+        at_index = email.index('@')
+        local_part = email[:at_index]  # RFC 5321, local part must be case-sensitive.
+        global_part = email[at_index:].lower()
+
+        return name or local_part, local_part + global_part
```
adriangb, line 739: This is the exact same code we had previously. To avoid adding a whole CI run just to test this version of email-validator, we are going to leave it uncovered. I ran locally and all tests pass with email-validator <2.

adriangb:
please review

setup.py
```diff
@@ -133,7 +133,7 @@ def extra(self):
         'typing-extensions>=4.2.0'
     ],
     extras_require={
-        'email': ['email-validator>=1.0.3'],
+        'email': ['email-validator>=2.0.0.post2'],
```
samuelcolvin, line 136: is this right? I thought we are supporting older `email-validator`?

samuelcolvin: I guess it's just `extra`, so we could do this, but I think it's more confusing that leaving it more open.

samuelcolvin (commented):
otherwise LGTM.

Please update, or merge if you're confident this is right.

setup.py
adriangb:
```suggestion
        'email': ['email-validator>=1.0.3'],
```

Issue #2264, opened 2021-01-15
Support for string fields that inherit from Generic


danni:
### Checks

* [-] I added a descriptive title to this issue
* [-] I have searched (google, github) for similar issues and couldn't find anything
* [-] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this feature/change is needed
* [-] After submitting this, I commit to one of:
    * Look through open issues and helped at least one other person
    * Hit the "watch" button on this repo to receive notifications and I commit to help at least 2 people that ask questions in the future
    * Implement a Pull Request for a confirmed bug

# Feature Request

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.7.3
            pydantic compiled: False
                 install path: /Users/danni/src/AP/apsys-vixen/.venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.1 (default, Jan  4 2021, 12:22:24)  [Clang 12.0.0 (clang-1200.0.32.28)]
                     platform: macOS-11.1-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']
```

We have a graph of entities that are linked by ref tuples of (typename, uuid). The `Ref` has a generic type to represent the return type of the Entity. We want to validate when parsing the model that typename refers to a valid reference for this field.

It's not currently possible to build a string field that inherits generic following the example in the docs, because the field value is suddenly misinterpreted as a sequence.

```py
from pydantic import BaseModel

class TypeName(str, Generic[T]):
    """This is a string, but that is validated against T."""

    ... # validation

class Ref(BaseModel, Generic[T]):
    typename: TypeName[T]
    uuid: UUID

    def resolve(self, graph: Graph) -> T: ...

class Entity(BaseModel, abc.ABC):
    typename: ClassVar[str]
    def resolve(self: T, document: Graph) -> T: return self

class Model(Entity):
    typename: ClassFinal = "model"
    parent: Ref["Model"]
```

The inclusion of `Generic` into `TypeName` results in.

```
E   pydantic.error_wrappers.ValidationError: 1 validation error for GazelleRef
E   typename
E     value is not a valid sequence (type=type_error.sequence)

pydantic/main.py:362: ValidationError
```

danni:
FYI I ultimately implemented this on `Ref`.

```py
    @classmethod
    def validate_typename(cls, value: Any, field: ModelField) -> Any:
        """Validate the typename implements the generic type."""

[...]

PR #5629, opened 2023-04-28
Error more eagerly for proper subclasses of generics that we don't know how to validate

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/2264. (As it fixes some confusing behavior.)

We should go further and make similar changes for other subclasses of generics which don't have a "standard" way to call `__init__`.

Users who want to support such classes should implement a `__get_pydantic_core_schema__` classmethod on the type in question, or if they don't control the type, on a pydantic-specific annotation used whenever that type occurs as a pydantic field annotation.

An example of how to do this is included in `tests.test_edge_cases.test_custom_generic_validators`

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -486,8 +486,13 @@ class GenerateSchema:
 
             return ordered_dict_schema(self, obj)
         elif issubclass(origin, typing.Sequence):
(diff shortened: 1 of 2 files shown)


dmontagu:
please review (also @adriangb for context, I'm not sure if you did something different/more general than this)

adriangb (approved):

PR #5641, opened 2023-04-29
Re-organize exports for new validators and serializers

adriangb:
Selected Reviewer: @hramezani

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -59,6 +59,7 @@ nav:
     - Color Types: usage/types/color_types.md
     - Datetimes: usage/types/datetime.md
     - Dicts and Mapping: usage/types/dicts_mapping.md
+    - Encoded Types: usage/types/encoded.md
     - Enums and Choices: usage/types/enums.md
     - File Types: usage/types/file_types.md
     - JSON: usage/types/json.md
@@ -83,10 +84,10 @@ nav:
   - usage/dataclasses.md
   - usage/validation_decorator.md
   - usage/postponed_annotations.md
-  - Errors:
-    - Usage Errors: usage/errors.md
-    - Validation Errors: usage/validation_errors.md
   - usage/conversion_table.md
+- Error Messages:
+  - Validation Errors: usage/validation_errors.md
+  - Usage Errors: usage/errors.md
 - Integrations:
   - 'Mypy': integrations/mypy.md
   - 'PyCharm': integrations/pycharm.md


samuelcolvin (changes requested):
I'm totally in agreement about moving methods to the new modules that make sense, but I really don't think we can remove the methods from `__init__.py`.

1. I think knowing that everything in the public API can be imported from `pydantic` is really helpful
2. One line imports for pydantic use is much cleaner and simpler
3. changing this now, especially when we're adding new recommended functions like `field_validator` will add a lot more churn to the migration

On balance I think this is one of those "correct" vs. "easy" decisions where I tend to lean towards easy and you tend to lean towards correct.

I think on this occasion we should revert that change.

The only exception to the "everything in `__init__.py`" rule right now (definitely the only important one) is `dataclass` - on this I think we should:
1. re-export our version of `dataclass`
2. rename the `dataclasses` module to `pydantic_dataclasses` to avoid confusion with std-lib dataclasses module.

adriangb:
If we're going to export something via `pydantic/__init__.py` I don't think there's much point in also exporting it from another module. Right?

samuelcolvin:
I think there is - some people might well want to work like

```py
from pydantic import validators

...
@validators.[tab completion]
```

More to the point, to "no export from another module" would require all other pydantic modules to be made private, which seems a little ugly.

pydantic/decorators.py
```diff
@@ -187,81 +153,6 @@ def dec(f: Any) -> _decorators.PydanticDescriptorProxy[Any]:
...
-        # auto apply the @classmethod decorator
-        f = _decorators.ensure_classmethod_based_on_signature(f)
-
-        dec_info = _decorators.FieldValidatorDecoratorInfo(fields=fields, mode=mode, check_fields=check_fields)
-        return _decorators.PydanticDescriptorProxy(f, dec_info)
-
-    return dec
-
-
 @overload
```
samuelcolvin, line 156: `root_validator` should surely move into `validators.py`?

More generally can we remove this module, or move it to `_internal`.

samuelcolvin (commented):
otherwise LGTM.

pydantic/decorators.py
adriangb: I didn't move it because:
1. Backwards compatibility with V1.
2. We want to deprecate `@root_validator` anyway

samuelcolvin: `root_validator` and `validator` where in `class_validators.py` in V1, the logic for supporting imports from the old path is already supported.

I think we move them.

samuelcolvin:
I have a fix for the above issues. Please don't push until I've pushed my fix - currently can't due to problems with GH.

samuelcolvin:
I have a fix for the above issues. Please don't push until I've pushed my fix - currently can't due to problems with GH.

samuelcolvin:
I have a fix for the above issues. Please don't push until I've pushed my fix - currently can't due to problems with GH.

samuelcolvin (approved):

samuelcolvin (approved):

PR #5646, opened 2023-04-30
Add `deprecated` on `BaseModel` methods

Kludex:
I'll add tests on Monday. This helps VSCode users.


dmontagu:
Unless this changes behavior in some meaningful way, I don't know that this needs new tests since it already has the deprecation warning emitted from the first line of the body. But I'll defer to you @Kludex if you had something in mind re: testing

dmontagu (approved):

PR #5648, opened 2023-04-30
demonstrate `custom_init` behaviour

samuelcolvin:
Show how https://github.com/pydantic/pydantic-core/pull/566 works

selected reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -283,6 +283,7 @@ class GenerateSchema:
             config=core_config,
             post_init=model_post_init,
             metadata=metadata,
+            custom_init='__init__' in cls.__dict__,
         )
         model_schema = consolidate_refs(model_schema)
         schema = apply_model_serializers(model_schema, decorators.model_serializer.values())
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -1313,7 +1313,6 @@ def test_self_recursive():
     assert m.model_dump() == {'sm': {'self': 123}}
 
 
-@pytest.mark.xfail(reason='need to detect and error if you override __init__; need to suggest a migration path')
 def test_nested_init():
     class NestedModel(BaseModel):
         self: str
@@ -1337,6 +1336,30 @@ def test_nested_init():
     assert m.self == 'Top Model'
     assert m.nest.self == 'Nested Model'
     assert m.nest.modified_number == 1
+
+
+def test_init_model_validate():
+    calls = []
+
+    class MyModel(BaseModel):
+        number: int = 1
+
+        def __init__(self, **kwargs):
+            super().__init__(**kwargs)
+            self.number += 1
+            calls.append(str(kwargs))
+
+    m = MyModel(number=1)
+    assert m.number == 2
+    assert calls == ["{'number': 1}"]
+
+    m = MyModel.model_validate(dict(number=1))
+    assert m.number == 2
+    # insert_assert(calls)
+    assert calls == [
+        "{'number': 1}",
+        "{'validated_data': ValidatedData(model_dict={'number': 2}, fields_set={'number'})}",
+    ]
 
 
 def test_init_inspection():
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -1353,12 +1353,21 @@ def test_init_model_validate():
     assert m.number == 2
     assert calls == ["{'number': 1}"]
 
-    m = MyModel.model_validate(dict(number=1))
-    assert m.number == 2
+    m = MyModel.model_validate(dict(number=2))
+    assert m.number == 3
     # insert_assert(calls)
     assert calls == [
         "{'number': 1}",
-        "{'validated_data': ValidatedData(model_dict={'number': 2}, fields_set={'number'})}",
+        # NOTE! model_dict has been modified in-place
+        "{'validated_data': ValidatedData(model_dict={'number': 3}, fields_set={'number'})}",
+    ]
+
+    m = MyModel.model_validate_json('{"number": 3}')
+    assert m.number == 4
+    assert calls == [
+        "{'number': 1}",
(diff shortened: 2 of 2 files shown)


samuelcolvin:
please review.

samuelcolvin:
Needs https://github.com/pydantic/pydantic-core/pull/566 to be merged before tests will pass.

tests/test_edge_cases.py
```diff
@@ -1313,7 +1313,6 @@ class Model(BaseModel):
     assert m.model_dump() == {'sm': {'self': 123}}
 
 
-@pytest.mark.xfail(reason='need to detect and error if you override __init__; need to suggest a migration path')
```
dmontagu, line 1316: There's a big fat TODO comment in this test that can be deleted, unfortunately I can't suggest it because it's a bit too far down

tests/test_edge_cases.py
```diff
@@ -1339,6 +1338,39 @@ class TopModel(BaseModel):
...
+    assert m.number == 2
+    assert calls == ["{'number': 1}"]
+
+    m = MyModel.model_validate(dict(number=2))
+    assert m.number == 3
+    # insert_assert(calls)
+    assert calls == [
+        "{'number': 1}",
+        # NOTE! model_dict has been modified in-place
+        "{'validated_data': ValidatedData(model_dict={'number': 3}, fields_set={'number'})}",
```
dmontagu, line 1362: Is there any potential for collisions if someone uses their own keyword argument with the same name? If it's checking that the type is `ValidatedData` then that's probably fine.

pydantic/_internal/_generate_schema.py
```diff
@@ -283,6 +283,7 @@ def model_schema(self, cls: type[BaseModel]) -> core_schema.CoreSchema:
             config=core_config,
             post_init=model_post_init,
             metadata=metadata,
+            custom_init='__init__' in cls.__dict__,
```
dmontagu, line 286: This way of setting the custom_init boolean doesn't preserve the setting for subclasses, which I think it should. Here's a failing test:
```python
def test_submodel_calls_custom_init():
    calls = []

    class InitModel(BaseModel):
        x: int
    
        def __init__(self, **kwargs):
            super().__init__(**kwargs)
            calls.append(kwargs)
    
    
    class SubInitModel(InitModel):
        pass
    
    
    SubInitModel.model_validate({'x': 1})
    assert calls
```
I think `custom_init=not (cls.__init__ is BaseModel.__init__)` might work better?

tests/test_edge_cases.py
samuelcolvin: There's a test for that on pydantic-core, it's fine.

pydantic/_internal/_generate_schema.py
samuelcolvin: Good catch, fixable I think.

dmontagu (commented):
This doesn't seem to behave quite as well with `model_validate` as in v1 when you have a different signature. Here's an example of something you can do in v1:
```python

calls = []


class InitModel(BaseModel):
    x: int

    def __init__(self, y: int, **kwargs):
        calls.append(('pre', kwargs.copy()))
        kwargs['x'] += y
        super().__init__(**kwargs)
        calls.append(('post', kwargs))



print(InitModel.parse_obj({'x': 1, 'y': 1}))
#> x=2
print(calls)
#> [('pre', {'x': 1}), ('post', {'x': 2})]
```
In v2, it errors when it tries to call the inner:
```python

calls = []


class InitModel(BaseModel):
    x: int

    def __init__(self, y: int, **kwargs):
        calls.append(('pre', kwargs.copy()))
        kwargs['x'] += y
        super().__init__(**kwargs)
        calls.append(('post', kwargs))



print(InitModel.model_validate({'x': 1, 'y': 1}))
#> TypeError: InitModel.__init__() missing 1 required positional argument: 'y'
```

I think it would be good if we could resolve this, I'm not quite sure what's going on, my immediate thought is it's related to preserving the "extra" values.

dmontagu:
Another thing that works in v1 but not in v2:
```python

calls = []


class InitModel(BaseModel):
    x: int

    def __init__(self, **kwargs):
        print(kwargs['x'])
        super().__init__(**kwargs)


InitModel.parse_obj({'x': 1})
#> 1
```
It seems hard to do anything before the call to `super().__init__`; given that, are there things that this enables that can't be done with `model_post_init`? (I will note that the custom inits I saw while investigating had keyword arguments added to the signature, so I think won't be compatible with this `validated_data` thing.

dmontagu:
I'm thinking if you call `BaseModel.__init__(instance, validated_data=ValidatedData(...))` on the inner call instead of calling `cls.__init__(validated_data=ValidatedData(...))` maybe it will help? Sorry if that's dumb, I haven't fully thought it through

dmontagu:
Okay, I looked into it a bit, and I realize I misunderstood what was happening; I will note that changing the definition of `model_validate` to:
```python
    @classmethod
    def model_validate(
        cls: type[Model], obj: Any, *, strict: bool | None = None, context: dict[str, Any] | None = None
    ) -> Model:
        # `__tracebackhide__` tells pytest and some other tools to omit this function from tracebacks
        __tracebackhide__ = True
        if cls.__init__ is BaseModel.__init__:
            return cls.__pydantic_validator__.validate_python(obj, strict=strict, context=context)
        return cls(**obj)
```
gets it to work the same as in v1 (unsurprisingly), though of course this will mishandle various things. But I'm thinking maybe it provides some insight into how to fix the behavior — if there is a custom init, maybe attempt to call it right away, rather than after getting the validated data?

dmontagu:
For context, here's the custom `__init__` example I found that did nontrivial stuff before and after the `__init__` call: https://github.com/gdsfactory/gdsfactory/blob/447d3c183aea2e157c3edb64691a32dee57136a9/gdsfactory/technology/layer_views.py#L426-L461

(There were two other custom inits in that repo but they started by calling `super().__init__(**kwargs)` so could just be replaced with `model_post_init`, I think.)

PR #5653, opened 2023-04-30
remove unused info arguments from validator functions

samuelcolvin:
We had a bunch of function validators which still took the `info` argument from before we made it optional.

Updating them to not take the argument, should be a little faster.

selected reviewer: @adriangb - I guess these could conflict with your work on how some of these validators work, happy to defer this if that's easier, hence assigning you.

Selected Reviewer: @adriangb


samuelcolvin:
please review.

adriangb:
This is great. I kept finding these in my related work but was too lazy to split it out. So I'll just have to deal with conflicts.

adriangb (approved):

PR #5657, opened 2023-05-01
Add support for `*` to `field_validator`

Kludex:
- Closes https://github.com/pydantic/pydantic/issues/5569

@dmontagu We talked about supporting `*` in a way that would be more "extensible", but... This was enough... 🤔

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -95,7 +95,6 @@ def check_validator_fields_against_field_name(
     field: str,
 ) -> bool:
     if isinstance(info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)):
-        # V1 compat: accept `'*'` as a wildcard that matches all fields
         if info.fields == ('*',):
             return True
     for v_field_name in info.fields:
@@ -108,7 +107,6 @@ def check_decorator_fields_exist(decorators: Iterable[AnyFieldDecorator], fields
     fields = set(fields)
     for dec in decorators:
         if isinstance(dec.info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)) and dec.info.fields == ('*',):
-            # V1 compat: accept `'*'` as a wildcard that matches all fields
             continue
         if dec.info.check_fields is False:
             continue
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -95,7 +95,7 @@ def check_validator_fields_against_field_name(
     field: str,
 ) -> bool:
     if isinstance(info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)):
-        if info.fields == ('*',):
+        if '*' in info.fields:
             return True
     for v_field_name in info.fields:
         if v_field_name == field:
@@ -106,7 +106,7 @@ def check_validator_fields_against_field_name(
 def check_decorator_fields_exist(decorators: Iterable[AnyFieldDecorator], fields: Iterable[str]) -> None:
     fields = set(fields)
     for dec in decorators:
-        if isinstance(dec.info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)) and dec.info.fields == ('*',):
+        if isinstance(dec.info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)) and '*' in dec.info.fields:
             continue
         if dec.info.check_fields is False:
             continue
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -94,7 +94,7 @@ def check_validator_fields_against_field_name(
     info: FieldDecoratorInfo,
     field: str,
 ) -> bool:
-    if isinstance(info, ValidatorDecoratorInfo):
+    if isinstance(info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)):
         # V1 compat: accept `'*'` as a wildcard that matches all fields
         if info.fields == ('*',):
             return True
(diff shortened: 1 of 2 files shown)


tests/test_validators.py
```diff
@@ -683,21 +686,28 @@ def check_a(cls, v: Any):
     assert check_calls == 1
 
 
-def test_wildcard_validators():
+@pytest.mark.parametrize(
```
Kludex, line 689: If you don't like this, I can create a separate test - but it will be the same logic.

Kludex:
please review

tests/test_validators.py
hramezani: fine for me

hramezani (approved):

hramezani:
LGTM 👍 

> but... This was enough... 🤔

I am not sure 😃 . @dmontagu may have better idea

pydantic/_internal/_generate_schema.py
```diff
@@ -94,7 +94,7 @@ def check_validator_fields_against_field_name(
     info: FieldDecoratorInfo,
     field: str,
 ) -> bool:
-    if isinstance(info, ValidatorDecoratorInfo):
+    if isinstance(info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)):
         # V1 compat: accept `'*'` as a wildcard that matches all fields
```
adriangb, line 98: Maybe remove this comment?

pydantic/_internal/_generate_schema.py
```diff
@@ -107,8 +106,7 @@ def check_validator_fields_against_field_name(
 def check_decorator_fields_exist(decorators: Iterable[AnyFieldDecorator], fields: Iterable[str]) -> None:
     fields = set(fields)
     for dec in decorators:
-        if isinstance(dec.info, ValidatorDecoratorInfo) and dec.info.fields == ('*',):
-            # V1 compat: accept `'*'` as a wildcard that matches all fields
+        if isinstance(dec.info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)) and dec.info.fields == ('*',):
```
dmontagu, line 109: I think better to do `'*' in dec.info.fields` rather than `dec.info.fields == ('*',)`. I could imagine scenarios where someone is deving and adds both `'*'` and a field name, and it ends up pretty confusing that it stops applying to all fields because another field becomes listed. I think this is less of an issue if it errors if you try to do this, I didn't check if it currently does.

dmontagu: actually maybe I misunderstood what this check is doing and it's good like this.

pydantic/_internal/_generate_schema.py
```diff
@@ -94,8 +94,7 @@ def check_validator_fields_against_field_name(
     info: FieldDecoratorInfo,
     field: str,
 ) -> bool:
-    if isinstance(info, ValidatorDecoratorInfo):
-        # V1 compat: accept `'*'` as a wildcard that matches all fields
+    if isinstance(info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)):
         if info.fields == ('*',):
```
dmontagu, line 98: I guess maybe this is where it should be changed to allow `'*'` to be present, not require exact equality on the fields (i.e., not allow others).

dmontagu: Yeah looks like on this branch, I get this error when I attempt to run the above:
```
pydantic.errors.PydanticUserError: Validators defined with incorrect fields: __main__.Model:5133041408.validate (use check_fields=False if you're inheriting from the model and intended this)
```
I think the code above should allow this. (Or if not, it should at least give a clearer error message. But I think it should be allowed.)

Kludex: This is a bit weird... And I actually didn't know people use it, but I've applied it. 🙏

pydantic/_internal/_generate_schema.py
```diff
@@ -107,8 +106,7 @@ def check_validator_fields_against_field_name(
 def check_decorator_fields_exist(decorators: Iterable[AnyFieldDecorator], fields: Iterable[str]) -> None:
     fields = set(fields)
     for dec in decorators:
-        if isinstance(dec.info, ValidatorDecoratorInfo) and dec.info.fields == ('*',):
-            # V1 compat: accept `'*'` as a wildcard that matches all fields
+        if isinstance(dec.info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)) and '*' in dec.info.fields:
```
dmontagu, line 109: Same here — any reason not to support `FieldSerializerDecoratorInfo`?

pydantic/_internal/_generate_schema.py
```diff
@@ -94,9 +94,8 @@ def check_validator_fields_against_field_name(
     info: FieldDecoratorInfo,
     field: str,
 ) -> bool:
-    if isinstance(info, ValidatorDecoratorInfo):
-        # V1 compat: accept `'*'` as a wildcard that matches all fields
-        if info.fields == ('*',):
+    if isinstance(info, (ValidatorDecoratorInfo, FieldValidatorDecoratorInfo)):
```
dmontagu, line 97: Any reason not to support `FieldSerializerDecoratorInfo`?

dmontagu (commented):
If you handle the serializer decorator, I think adding tests that look just like `test_wildcard_validators` and `test_wildcard_validator_error` will be good enough.

Depending on how much code it adds I think you can add the serializers into those same tests and just add checks when you call `model_dump`, or just create new tests.

Or we can just not add support for serializer, but it seems natural to me and like it won't take much effort.

Kludex:
For record keeping, we are not adding `FieldSerializerDecoratorInfo` because conceptually we have a single `field_serializer` per field.

PR #5662, opened 2023-05-01
Delegate equality checks for BaseModel

### LINKED ISSUE #5622: BaseModel only matches other BaseModel's (not even unittest.mock.ANY)
### Initial Checks

### Description

I've encountered some unexpected behavior while using Pydantic 2.0, specifically with the `__eq__` method in the `BaseModel` class. The current implementation of the method causes it to not match anything other than instances of `BaseModel`. This behavior might cause issues when using custom validation classes, e.g. `unittest.mock.ANY`.

Here's the code snippet of the `__eq__` method in Pydantic's `BaseModel`:
```python
def __eq__(self, other) -> bool:
    if not isinstance(other, BaseModel):
        return False
    ...
```

This implementation checks whether the `other` object is an instance of `BaseModel`, and immediately returns `False` if it is not. Due to this, when comparing a `BaseModel` instance with custom validation classes, like `unittest.mock.ANY`, the comparison will always return `False`.

While I understand that this check is in place to ensure that only instances of `BaseModel` can be compared (and potentially for performance), it would be helpful to have a more flexible implementation that allows compatibility with custom validation classes.

Also I do not want to simply reorder the comparison since then pytest will display the expected and actual results in the opposite order of what I want.

**Possible solution:**

One potential solution is to update the `__eq__` method in `BaseModel` to delegate the comparison to the `other` object if it is not an instance of `BaseModel`:

```python
def __eq__(self, other) -> bool:
    if not isinstance(other, BaseModel):
        return NotImplemented
```

By returning `NotImplemented`, we allow Python to delegate the comparison to the `other` object's `__eq__` method. If that method is capable of handling the comparison with a `BaseModel` instance, it will return the appropriate result. This makes the comparison more flexible and compatible with custom validation classes.

With this change, users can implement their own custom validation classes and handle comparisons with `BaseModel` instances in the custom class's `__eq__` method:
```python
class CustomValidationClass:
    def __eq__(self, other) -> bool:
        if isinstance(other, BaseModel):
            # Implement custom comparison logic here
            ...
```

Performance wise I don't think this will affect much. In the case of BaseModel's it has no change. For other types it may need to do a few more steps but nothing considerable (I have not performed any benchmarks in this regard).

If this solution or others turn out to have bad effects, performance wise or in some other way, is it possible to add a config option to enable this?

Please let me know if you have any questions or require further information on this issue. Thank you for creating this awesome project :pray:

**Side Note:**
I orginally discovered this kink with a Custom Validation class for Pydantic V1. The Custom Validation class would check for an instance of BaseModel. However the implementation in V1 first converts the model to a dict so it was no longer an instance of BaseModel and failed the test.

### Example Code

```Python
from unittest.mock import ANY

class MyModel(BaseModel):
    foo: str


if __name__ == "__main__":
    assert MyModel(foo="bar") == ANY
```

```Text
pydantic version: 2.0a3
pydantic-core version: 0.25.0 release build profile
install path: /Users/ulfur/projects/pydantic-test/.venv/lib/python3.10/site-packages/pydantic
python version: 3.10.1 (v3.10.1:2cd268a3a9, Dec  6 2021, 14:28:59) [Clang 13.0.0 (clang-1300.0.29.3)]
```

dmontagu:
Closes #5622 

Changes from returning `False` in `BaseModel` equality checks to telling python to look at the implementation from the other side of the `==`. I did some investigation in https://github.com/pydantic/pydantic/issues/5622 which I will copy below:

-----
_Quoting [this comment](https://github.com/pydantic/pydantic/issues/5622#issuecomment-1526721788)_

I did a little benchmarking of this:
```python
import time

class MyModel(BaseModel):
    a: int

@dataclass
class MyDataclass:
    a: int

m1 = MyModel(a=1)
m2 = MyDataclass(a=1)
t0 = time.time()
for _ in range(10_000_000):
    m1 == m2
t1 = time.time()

print(f"elapsed: {t1-t0:.3f}s")
# with return False:          elapsed: 1.793s
# with return NotImplemented: elapsed: 2.581s, 43% slower
```
As you can see, it runs _significantly_ slower with `return NotImplemented` vs. `return False`. However, I think there are several good arguments for us making this change:
* It does improve testability of various things and allows custom comparisons, as you've noted
* It's only slower in the case that you are doing an `==` comparison against something that isn't even a BaseModel; this seems pretty unlikely to be happening in "hot" code
* Even though it's noticeably slower, it's still pretty fast, at approximately 4M comparisons/second on my local machine.
* This is what `dataclasses` does: https://github.com/python/cpython/blob/d2e2e53f733f8c8098035bbbc452bd1892796cb3/Lib/dataclasses.py#L669-L671

Given all the above points, I'll open a PR making this change.

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -538,27 +538,27 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
         yield from self.__dict__.items()
 
     def __eq__(self, other: Any) -> bool:
-        if not isinstance(other, BaseModel):
-            return False
+        if isinstance(other, BaseModel):
+            # When comparing instances of generic types for equality, as long as all field values are equal,
+            # only require their generic origin types to be equal, rather than exact type equality.
+            # This prevents headaches like MyGeneric(x=1) != MyGeneric[Any](x=1).
+            self_type = self.__pydantic_generic_metadata__['origin'] or self.__class__
+            other_type = other.__pydantic_generic_metadata__['origin'] or other.__class__
 
-        # When comparing instances of generic types for equality, as long as all field values are equal,
-        # only require their generic origin types to be equal, rather than exact type equality.
-        # This prevents headaches like MyGeneric(x=1) != MyGeneric[Any](x=1).
-        self_type = self.__pydantic_generic_metadata__['origin'] or self.__class__
-        other_type = other.__pydantic_generic_metadata__['origin'] or other.__class__
-
-        if self_type != other_type:
-            return False
-
-        if self.__dict__ != other.__dict__:
-            return False
-
-        # If the types and field values match, check for equality of private attributes
-        for k in self.__private_attributes__:
-            if getattr(self, k, Undefined) != getattr(other, k, Undefined):
+            if self_type != other_type:
                 return False
 
-        return True
+            if self.__dict__ != other.__dict__:
+                return False
+
+            # If the types and field values match, check for equality of private attributes
+            for k in self.__private_attributes__:
+                if getattr(self, k, Undefined) != getattr(other, k, Undefined):
+                    return False
+
+            return True
+        else:
+            return NotImplemented  # delegate to the other item in the comparison
 
     def model_copy(self: Model, *, update: dict[str, Any] | None = None, deep: bool = False) -> Model:
         """
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -2203,3 +2203,14 @@ def test_model_validate_with_context():
     assert OuterModel.model_validate({'inner': {'x': 2}}, context={'multiplier': 1}).inner.x == 2
     assert OuterModel.model_validate({'inner': {'x': 2}}, context={'multiplier': 2}).inner.x == 4
     assert OuterModel.model_validate({'inner': {'x': 2}}, context={'multiplier': 3}).inner.x == 6
+
+
+def test_equality_delegation():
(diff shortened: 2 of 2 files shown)


pydantic/main.py
```diff
@@ -538,27 +538,27 @@ def __iter__(self) -> TupleGenerator:
...
-        if self_type != other_type:
-            return False
-
-        if self.__dict__ != other.__dict__:
-            return False
+        if isinstance(other, BaseModel):
+            # When comparing instances of generic types for equality, as long as all field values are equal,
+            # only require their generic origin types to be equal, rather than exact type equality.
+            # This prevents headaches like MyGeneric(x=1) != MyGeneric[Any](x=1).
+            self_type = self.__pydantic_generic_metadata__['origin'] or self.__class__
```
dmontagu, line 545: I don't know if the ordering here has any impact on performance, but it seemed to be it would at worst be equal speed for the common case of comparing two BaseModel instances, and it didn't seem to me to have much impact on readability.

dmontagu: (This looks like a big diff, but the only thing that changed was the order of the if statement, and the fact that if the types are different, it now returns NotImplemented rather than False)

dmontagu:
please review

samuelcolvin (approved):
LGTM.

samuelcolvin:
great, thank you.

Issue #4286, opened 2022-07-27
Alias generator with Pydantic dataclass not behaving as expected


CuriousHexagon:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug
Using Pydantic 1.9.0 and python 3.9.10

Using `alias_generator` in the `Config` for a `pydantic.dataclasses.dataclass`  throws `unexpected keyword` when using the alias, but also Pydantic is aware of the alias since `field required (type=value_error.missing)` is also thrown when the alias name is not used.

The same example works without issue with a `BaseModel` class. The documentation provides an example of using `Config` with a `pydantic.dataclasses.dataclass` but does not mention of any `Config` limitations when using such a dataclass.

```py

from pydantic.dataclasses import dataclass


def to_lower_camel(name: str) -> str:
    """
    Converts a snake_case string to lowerCamelCase
    """
    upper = "".join(word.capitalize() for word in name.split("_"))
    return upper[:1].lower() + upper[1:]


class MyConfig:
    alias_generator = to_lower_camel


@dataclass(config=MyConfig)
class User:
    name: str
    highest_score: int

    class Config:
        alias_generator = to_lower_camel
```
Exception thrown when trying to use the alias:
```py
user_a = User(name='abc', highestScore=100)

# OUTPUT 
# Traceback (most recent call last):
#     user = User(name='abc', highestScore=100)
# TypeError: __init__() got an unexpected keyword argument 'highestScore'
```
Exception thrown when not using the alias (expected - but demonstrates that neither naming convention now works):
```py
user_b = User(name='abc', highest_score=100)

# OUTPUT
# pydantic.error_wrappers.ValidationError: 1 validation error for User
# highestScore
#   field required (type=value_error.missing)
```

samuelcolvin:
Yes, we use the standard dataclass initialisation to resolve args and kwargs, then do validation, thus aliases can't be used because dataclasses don't know how to handle them.

This is will be fixed in pydantic V2, in fact I'm working on the functionality right now https://github.com/samuelcolvin/pydantic-core/pull/190.

We could definitely update the docs to mention which bits of config work, and which don't. Similar things have been reported before e.g. #986.

devraj:
Hey all,

[...]

PR #5673, opened 2023-05-03
Respect config.alias_generator in pydantic dataclass

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/4286

Selected Reviewer: @Kludex

### DIFF
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -263,7 +263,7 @@ print(Character.model_json_schema(by_alias=True))
 {
     'type': 'object',
     'properties': {
-        'ActorName': {'type': 'string', 'default': None, 'title': 'Actorname'},
+        'Name': {'type': 'string', 'default': None, 'title': 'Name'},
         'LanguageCode': {'type': 'string', 'default': None, 'title': 'Languagecode'},
         'Mood': {'type': 'string', 'default': None, 'title': 'Mood'},
         'Act': {'type': 'integer', 'default': 1, 'title': 'Act'},
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -263,7 +263,7 @@ print(Character.model_json_schema(by_alias=True))
 {
     'type': 'object',
     'properties': {
-        'Name': {'type': 'string', 'default': None, 'title': 'Name'},
+        'ActorName': {'type': 'string', 'default': None, 'title': 'Actorname'},
         'LanguageCode': {'type': 'string', 'default': None, 'title': 'Languagecode'},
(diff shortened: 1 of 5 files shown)


hramezani:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -534,14 +534,22 @@ def _generate_dc_field_schema(
...
         """
         common_field = self._common_field_schema(name, field_info, decorators)
+
+        validation_alias: str | list[str | int] | list[list[str | int]] | None = common_field['validation_alias']
+        serialization_alias: str | None = common_field['serialization_alias']
+        alias_generator = self.config_wrapper.alias_generator
+        if alias_generator:
+            validation_alias = alias_generator(name)
+            serialization_alias = alias_generator(name)
+
```
hramezani, line 544: We can move this change into [_common_field_schema](https://github.com/pydantic/pydantic/blob/0346ddb6a35770007f32815d8e4a179b778e0ef4/pydantic/_internal/_generate_schema.py#L549) but this function is used in `_generate_td_field_schema` as well. So, that's why I keep it in the `_generate_dc_field_schema`

dmontagu: I think it may be a good idea to move into `_common_field_schema` since soon there will be something equivalent for `_generate_model_field_schema` (change coming down the pipe in pydantic core). Does it behave horribly/not work for typeddict if someone did add an alias? I think it's reasonable if someone might want to use aliases with typeddict too.

hramezani: I moved it to `_common_field_schema` and also added a test for typeddict inspired by your example.

dmontagu (approved):

docs/usage/model_config.md
```diff
@@ -263,7 +263,7 @@ print(Character.model_json_schema(by_alias=True))
 {
     'type': 'object',
     'properties': {
-        'ActorName': {'type': 'string', 'default': None, 'title': 'Actorname'},
+        'Name': {'type': 'string', 'default': None, 'title': 'Name'},
```
dmontagu, line 266: I don't think this attribute name should be changed, because the alias generator isn't supposed to override explicitly-set aliases.

The relevant attribute is `FieldInfo.alias_priority`.

I think this shouldn't be too hard to fix, but we should look at how that is used now.

dmontagu: I think we probably want to make a version of `pydantic._internal._model_construction.apply_alias_generator` for dataclasses / typeddict

hramezani:
please review

dmontagu (approved):

dmontagu (approved):

PR #5674, opened 2023-05-03
Add `__prepare_pydantic_annotations__` for custom types

adriangb:
This covers two important use cases, both of which we use in `Decimal`:
- "Consuming" constraints when the core schema you return from `__get_pydantic_core_schema__` is not the one that the metadata should be applied to.
- Persisting state between `__get_pydantic_core_schema__` and `__get_pydantic_json_schema__` (e.g. if you want to generate the json schema on something other than the actual core schema but need info from within that core schema or the source type).

Both of these are _impossible_ to do as a user currently and even within our implementation require a lot of hacking around.

I'm doing this as standalone thing in this PR, a followup PR will refactor our current `Decimal` and everything else using `CustomValidator` stuff to be like this.

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -299,30 +299,14 @@ class GenerateSchema:
         Note: `__get_pydantic_core_schema__` takes priority so it can
         decide whether to use a `__pydantic_core_schema__` attribute, or generate a fresh schema.
         """
-        prepare = getattr(obj, '__prepare_pydantic_annotations__', None)
-        if prepare is None:
-            prepare = self._get_prepare_pydantic_annotations_for_known_type(obj)
-        if prepare is None:
-            return None
-
-        # make annotations a tuple to error if it gets mutated
-        # make the return type a list to support generators or returning a sequence
-        res = list(prepare(obj, []))
-        if not res:
-            raise PydanticSchemaGenerationError(
-                f'The type {obj} that implements `__prepare_pydantic_annotations__`'
-                ' returned no annotations when called.'
-                ' Custom types must return at least 1 item since the first item is the replacement source type.'
-            )
-        obj, *new_annotations = res
+        obj, new_annotations = self.prepare_annotations(obj, [])
         if new_annotations:
             return self.apply_annotations(
                 CallbackGetCoreSchemaHandler(partial(self.generate_schema, from_prepare_args=False)),
                 obj,
                 new_annotations,
             )
-        else:
-            return self.generate_schema(obj, from_prepare_args=False)
+        return None
 
     def _generate_schema_from_property(self, obj: Any, source: Any) -> core_schema.CoreSchema | None:
         """
@@ -1147,15 +1131,7 @@ class GenerateSchema:
 
         return None
 
-    def apply_annotations(
-        self,
-        get_inner_schema: GetCoreSchemaHandler,
-        source_type: Any,
-        annotations: typing.Iterable[Any],
-    ) -> CoreSchema:
-        """
-        Apply arguments from `Annotated` or from `FieldInfo` to a schema.
-        """
+    def prepare_annotations(self, source_type: Any, annotations: Iterable[Any]) -> tuple[Any, list[Any]]:
         prepare = getattr(source_type, '__prepare_pydantic_annotations__', None)
         annotations = list(annotations)
         if prepare is None:
@@ -1172,6 +1148,18 @@ class GenerateSchema:
                     ' Custom types must return at least 1 item since the first item is the replacement source type.'
                 )
             source_type, *annotations = res
+        return (source_type, annotations)
+
+    def apply_annotations(
+        self,
+        get_inner_schema: GetCoreSchemaHandler,
+        source_type: Any,
+        annotations: typing.Iterable[Any],
+    ) -> CoreSchema:
+        """
+        Apply arguments from `Annotated` or from `FieldInfo` to a schema.
+        """
+        source_type, annotations = self.prepare_annotations(source_type, annotations)
 
         idx = -1
         while True:
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1142,6 +1142,7 @@ class GenerateSchema:
         if prepare is None:
             # check if this is one of our "known" types
             prepare = self._get_prepare_pydantic_annotations_for_known_type(source_type)
+
         if prepare is not None:
             # make annotations a tuple to error if it gets mutated
             # make the return type a list to support generators or returning a sequence
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -306,9 +306,18 @@ class GenerateSchema:
         if prepare is not None:
             # make annotations a tuple to error if it gets mutated
             # make the return type a list to support generators or returning a sequence
-            new_annotations = list(prepare(obj, tuple(annotations)))
+            res = list(prepare(obj, tuple(annotations)))
+            if not res:
+                raise PydanticSchemaGenerationError(
+                    f'The type {obj} that implements `__prepare_pydantic_annotations__`'
+                    ' returned no new annotations when called.'
+                    ' You must return at least 1 item since the first item is the replacement source type.'
+                )
+            obj, *new_annotations = list(prepare(obj, tuple(annotations)))
             if new_annotations:
                 return self._annotated_args_schema(obj, new_annotations)
+            else:
+                return self.generate_schema(obj)
 
         return None
 
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1139,7 +1139,6 @@ class GenerateSchema:
         `Annotated[source_type, *annotations]` -> `Annotated[new_source_type, *new_annotations]`
         """
         prepare = getattr(source_type, '__prepare_pydantic_annotations__', None)
-        annotations = list(annotations)
         if prepare is None:
             # check if this is one of our "known" types
             prepare = self._get_prepare_pydantic_annotations_for_known_type(source_type)
@@ -1154,7 +1153,7 @@ class GenerateSchema:
                     ' Custom types must return at least 1 item since the first item is the replacement source type.'
                 )
             source_type, *annotations = res
-        return (source_type, annotations)
+        return (source_type, list(annotations))
 
     def _apply_annotations(
         self,
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -14,7 +14,7 @@ from functools import partial
 from inspect import Parameter, _ParameterKind, signature
 from itertools import chain
 from types import FunctionType, LambdaType, MethodType
-from typing import TYPE_CHECKING, Any, Callable, ForwardRef, Iterable, Mapping, Sequence, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Callable, ForwardRef, Iterable, Mapping, TypeVar, Union
 
 from annotated_types import BaseMetadata, GroupedMetadata
 from pydantic_core import CoreSchema, SchemaError, SchemaValidator, core_schema
@@ -195,12 +195,15 @@ class GenerateSchema:
     def arbitrary_types(self) -> bool:
         return self.config_wrapper.arbitrary_types_allowed
 
-    def generate_schema(self, obj: Any, from_dunder_get_core_schema: bool = True) -> core_schema.CoreSchema:
+    def generate_schema(
+        self, obj: Any, from_dunder_get_core_schema: bool = True, from_prepare_args: bool = True
+    ) -> core_schema.CoreSchema:
         schema: CoreSchema | None = None
 
-        schema = self._generate_schema_from_prepare_annotations(obj)
-        if schema:
-            return schema
+        if from_prepare_args:
+            schema = self._generate_schema_from_prepare_annotations(obj)
+            if schema:
+                return schema
 
         if from_dunder_get_core_schema:
             from_property = self._generate_schema_from_property(obj, obj)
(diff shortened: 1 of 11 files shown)


adriangb:
please review

tests/test_prepare_annotations.py
```diff
@@ -0,0 +1,116 @@
...
+
+def test_decimal_like_in_annotated() -> None:
+    def no_op_val(x: Any) -> Any:
+        return x
+
+    a = AnalyzedType(List[Annotated[MyDecimal, Gt(10), AfterValidator(no_op_val)]])
+
+    expected = de.IsPartialDict(
+        core_schema.list_schema(
+            de.IsPartialDict(
```
samuelcolvin, line 58: wow, I hadn't seen this before, pretty cool that it works, even if typing get's upset.

samuelcolvin (commented):
looks good in principle, I can review more fully tomorrow.

tests/test_prepare_annotations.py
adriangb: very convenient

adriangb:
Okay now I handled both `Decimal` and `SecretStr`/`SecretBytes`.

If you don't count the new tests, it's less LOC. And I was able to get rid of both `cs_update_function` in metadata and `CustomValidator`, neither of which were very nice IMO.

pydantic/_internal/_generate_schema.py
```diff
@@ -1106,6 +1116,72 @@ def _get_or_cache_recursive_ref(self, cls: type[Any]) -> tuple[str, core_schema.
...
+        from decimal import Decimal
+
+        from . import _std_types_schema as std_types
+
+        if obj is Decimal:
+            return std_types.decimal_prepare_pydantic_annotations
+
+        return None
+
+    def prepare_annotations(self, source_type: Any, annotations: Iterable[Any]) -> tuple[Any, list[Any]]:
```
samuelcolvin, line 1136: I think this can be private, also can you add a docstring.

pydantic/_internal/_generate_schema.py
```diff
@@ -1106,6 +1116,72 @@ def _get_or_cache_recursive_ref(self, cls: type[Any]) -> tuple[str, core_schema.
...
+            if not res:
+                raise PydanticSchemaGenerationError(
+                    f'The type {source_type} that implements `__prepare_pydantic_annotations__`'
+                    ' returned no annotations when called.'
+                    ' Custom types must return at least 1 item since the first item is the replacement source type.'
+                )
+            source_type, *annotations = res
+        return (source_type, annotations)
+
+    def apply_annotations(
```
samuelcolvin, line 1155: again, could be private?

pydantic/_internal/_generate_schema.py
```diff
@@ -1106,6 +1116,72 @@ def _get_or_cache_recursive_ref(self, cls: type[Any]) -> tuple[str, core_schema.
...
+        self,
+        get_inner_schema: GetCoreSchemaHandler,
+        source_type: Any,
+        annotations: typing.Iterable[Any],
+    ) -> CoreSchema:
+        """
+        Apply arguments from `Annotated` or from `FieldInfo` to a schema.
```
samuelcolvin, line 1162: probably worth explaining how this is different to `_annotated_schema`.

pydantic/_internal/_generate_schema.py
```diff
@@ -1106,6 +1116,72 @@ def _get_or_cache_recursive_ref(self, cls: type[Any]) -> tuple[str, core_schema.
...
             return obj_ref, None
 
+    def _annotated_schema(self, annotated_type: Any) -> core_schema.CoreSchema:
+        """
+        Generate schema for an Annotated type, e.g. `Annotated[int, Field(...)]` or `Annotated[int, Gt(0)]`.
+        """
+        first_arg, *other_args = get_args(annotated_type)
+        return self.apply_annotations(CallbackGetCoreSchemaHandler(self.generate_schema), first_arg, other_args)
+
+    def _get_prepare_pydantic_annotations_for_known_type(self, obj: Any) -> Callable[..., Any] | None:
```
samuelcolvin, line 1126: I'm a bit confused, is this temporary?

I'd rather we do this thing with decimals properly in this PR to demonstrate behaviour.

pydantic/_internal/_generate_schema.py
```diff
@@ -1106,6 +1116,72 @@ def _get_or_cache_recursive_ref(self, cls: type[Any]) -> tuple[str, core_schema.
...
+
+        return None
+
+    def prepare_annotations(self, source_type: Any, annotations: Iterable[Any]) -> tuple[Any, list[Any]]:
+        prepare = getattr(source_type, '__prepare_pydantic_annotations__', None)
+        annotations = list(annotations)
+        if prepare is None:
+            # check if this is one of our "known" types
+            prepare = self._get_prepare_pydantic_annotations_for_known_type(source_type)
+        if prepare is not None:
```
samuelcolvin, line 1142:
```suggestion
        else:
```

pydantic/_internal/_generate_schema.py
```diff
@@ -1106,6 +1116,72 @@ def _get_or_cache_recursive_ref(self, cls: type[Any]) -> tuple[str, core_schema.
...
+        from . import _std_types_schema as std_types
+
+        if obj is Decimal:
+            return std_types.decimal_prepare_pydantic_annotations
+
+        return None
+
+    def prepare_annotations(self, source_type: Any, annotations: Iterable[Any]) -> tuple[Any, list[Any]]:
+        prepare = getattr(source_type, '__prepare_pydantic_annotations__', None)
+        annotations = list(annotations)
```
samuelcolvin, line 1138: looks like this is either unecessary, or should be `tuple(annotations)`.

pydantic/_internal/_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+
+import dataclasses
+from typing import Any, Iterable
+
+import annotated_types as at
+
+from ._fields import PydanticGeneralMetadata, PydanticMetadata
+
+SEQUENCE_CONSTRAINTS = {'min_length', 'max_length'}
+COMPARISON_CONSTRAINTS = {'le', 'ge', 'lt', 'gt'}
```
samuelcolvin, line 11:
```suggestion
INEQUALITY_CONSTRAINTS = {'le', 'ge', 'lt', 'gt'}
```

pydantic/_internal/_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+NUMERIC_CONSTRAINTS = {'multiple_of', 'allow_inf_nan', *COMPARISON_CONSTRAINTS}
+
+CORE_SCHEMA_CONSTRAINTS = {
```
samuelcolvin, line 14: I'm a bit lost on how some of this is used, e.g. I don't see anywhere that `CORE_SCHEMA_CONSTRAINTS['list']` is used.

Also, I'm not sure this should be a dictionary, surely it would be fine if these were each separate valiables?

pydantic/_internal/_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+    'list': SEQUENCE_CONSTRAINTS,
+    'bytes': SEQUENCE_CONSTRAINTS,
+    'date': NUMERIC_CONSTRAINTS,
+    'time': NUMERIC_CONSTRAINTS,
+    'datetime': NUMERIC_CONSTRAINTS,
+    'timedelta': NUMERIC_CONSTRAINTS,
+}
+
+
+def collect_known_metadata(annotations: Iterable[Any]) -> tuple[dict[str, Any], list[Any]]:
```
samuelcolvin, line 27: please can you add a docstring here?

pydantic/_internal/_std_types_schema.py
```diff
@@ -131,35 +133,163 @@ def update_schema(_, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
...
+
+    def __get_pydantic_core_schema__(self, _source_type: Any, _handler: GetCoreSchemaHandler) -> CoreSchema:
+        return self.outer_core_schema
+
+    def __get_pydantic_json_schema__(self, _schema: CoreSchema, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
+        return handler(self.inner_core_schema)
+
+
+@slots_dataclass
+class DecimalValidator:
```
samuelcolvin, line 149: I'd prefer this doesn't move file in this PR, I agree having this and `_validators` separate is probably wrong, but it would be easier to review if we did that in a separate PR.

samuelcolvin (commented):
Some initial feedback, I need to review this more, and play with it.

I'm still not convinced that `__prepare_pydantic_annotations__` is the simplest way to achieve the functionality we need to add.

I'd also appreciate @dmontagu's opinion on this.

pydantic/_internal/_generate_schema.py
adriangb: It's already a private class. The only reason we have some public and some private things is that we used to at least partially expose this as part of a public interface. Now it's not, so we should just make everything public.

pydantic/_internal/_generate_schema.py
adriangb: see above

pydantic/_internal/_generate_schema.py
adriangb: We have a mix of places where we handle "known types". Some of it is in `GenerateSchema._generate_schema`, and some of it is in `_standard_types.py`. Decimal used to be in the latter, but the system in `_standard_types.py` doesn't really work with this new API. I could put thins is in `GenerateSchema._generate_schema` but that's already huge. So yes this is most likely temporary, although I think it may end up being the final resting place for Decimal and things like it.

pydantic/_internal/_metadata.py
adriangb: list is not being used but I added it just for completeness. happy to keep it to only the ones being used in this PR. also sure yeah separate variables works, it's just more imports if there ends up being a single module that uses them. And it's not like they're different types.

pydantic/_internal/_metadata.py
adriangb: yup

pydantic/_internal/_std_types_schema.py
samuelcolvin: I'm not sure, I need to play with this more and understand it better.

Leave for now.

pydantic/_internal/_metadata.py
samuelcolvin: let's have them all, but as variables.

pydantic/types.py
```diff
@@ -442,77 +443,20 @@ def __eq__(self, other: Any) -> bool:
...
-    ) -> JsonSchemaValue:
-        field_schema = handler(core_schema)
-        update_not_none(
-            field_schema,
-            type='string',
-            writeOnly=True,
-            format='password',
-        )
-        return field_schema
+    def __prepare_pydantic_annotations__(cls, source: type[Any], annotations: Iterator[Any]) -> Iterator[Any]:
```
dmontagu, line 454:
```suggestion
    def __prepare_pydantic_annotations__(cls, source: type[Any], annotations: Iterable[Any]) -> Iterable[Any]:
```
I saw `res = list(prepare(source_type, tuple(annotations)))` elsewhere so I don't think the thing getting passed into  this function is an Iterator in general

samuelcolvin:
please update.

dmontagu (approved):
This looks good to me in terms of API and implementation (though I agree with most of Samuel's suggestions about some of the details).

I think some clear documentation with:

* Some motivation for why this is necessary
* (Closely related) when you should define this method for a custom type/annotation
* _How_ to use it / how the mechanics of it work in some concrete examples (Decimal is okay, though a better example might not depend on any understanding of pydantic-core)

would go a long way toward making it clear why this is worth adding. I'm not _1000%_ convinced this is the best possible way to handle the "inversion of control" for types to decide how annotations apply to them. BUT I think it's carefully considered and sensible, and I haven't been able to come up with anything better. And I can't think of any scenario it wouldn't be able to handle.

(Honestly, I think structuring the documentation akin to a lightweight PEP would probably help with understanding of this functionality both internally (i.e., devs) and externally (i.e. (advanced) users).)

----

I think the `__get_pydantic_core_schema__`, `__prepare_pydantic_annotations__`, and `__get_pydantic_json_schema__` triumvirate should make it possible to extend pydantic to work ergonomically with just about any validation semantics for custom/third party types that anyone could desire — provided a willingness to get hands dirty with the annotation implementations.

pydantic/_internal/_generate_schema.py
adriangb: we do need to make sure it's a list because that's what we claim to return. But I moved it down to the return to clarify

pydantic/_internal/_generate_schema.py
samuelcolvin: makes sense.

pydantic/_internal/_metadata.py
adriangb: don thanks

pydantic/_internal/_metadata.py
adriangb: done

pydantic/_internal/_std_types_schema.py
adriangb: ok

pydantic/_internal/_generate_schema.py
samuelcolvin: nic-pick suggestion: newline before

adriangb: done

pydantic/types.py
adriangb: Yup thanks.

I think ideally we want to encourage the signature you are suggesting for users.
As part of writing the docs I plan on making some protocol classes that describe these APIs.
They will be helpful in terms of discoverability, reducing boilerplate because you can add it as a temporary base class and get autocomplete but also maybe leave it there permanently.

pydantic/_internal/_metadata.py
adriangb: done

pydantic/_internal/_generate_schema.py
adriangb: I agree with you David. I added a note to make a PR _after_ this one that does renaming.

samuelcolvin:
conflicts are preventing tests from running.

adriangb:
> conflicts are preventing tests from running

rebased

samuelcolvin (commented):
Okay, I've reviewed this and I'm happy with it.

My only suggestions are:
* would be amazing if we could get rid of `_get_prepare_pydantic_annotations_for_known_type` - from my reading I don't think it should be necessary
* Should we export a utility `MetadataApplier`, or at at least define it internally, it looks like it would be useful

adriangb:
> would be amazing if we could get rid of _get_prepare_pydantic_annotations_for_known_type - from my reading I don't think it should be necessary

That's internal, let's leave that for the next refactor.

> Should we export a utility MetadataApplier, or at at least define it internally, it looks like it would be useful

I agree that it might, but also can be done in a followup PR.

PR #5675, opened 2023-05-03
Pin urllib3 in CI

adriangb:
urllib3 just made a major release (2.0.0) and it seems that's broken cachecontrol, which PDM depends on. This pins urllib3 to <2.


samuelcolvin (approved):

samuelcolvin:
Nice fix.

PR #5676, opened 2023-05-04
Re-add schema_extra functionality removed from config

dmontagu:
Re-adds the `schema_extra` config key (though renames it to `json_schema_extra`), with the same types/function signatures accepted as in v1.

This was one of the bigger pain points for me updating the FastAPI tests, as the migration path wasn't super straightforward.

At PyCon I think we came to the conclusion that this was a reasonable thing to re-add.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -369,8 +369,8 @@ jobs:
   send-tweet:
     name: Send tweet
     runs-on: ubuntu-latest
-    needs: [deploy]
-    if: ${{ needs.deploy.status == 'OK' }}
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -369,8 +369,8 @@ jobs:
   send-tweet:
     name: Send tweet
     runs-on: ubuntu-latest
-    needs: [deploy]
```
dmontagu, line 372: looks like we had a botched merge, CI won't run without this change.

adriangb (approved):

PR #5680, opened 2023-05-04
Use pydantic core main

adriangb:


docs/usage/schema.md
```diff
@@ -718,37 +744,7 @@ top_level_schema = AnalyzedType.json_schemas(
...
-      },
-      "required": [
-        "a"
-      ],
-      "title": "Model"
-    }
-  }
-}
-"""
+#> {}
```
samuelcolvin, line 747: I'm concerned about what's going on here - this looks like a bug?

docs/usage/models.md
```diff
@@ -571,6 +577,7 @@ except ValidationError as e:
             'msg': 'value is not "bar", got "ber"',
             'input': 'ber',
             'ctx': {'wrong_value': 'ber'},
+            'url': 'https://errors.pydantic.dev/2/v/not_a_bar',
```
dmontagu, line 580: Reminder, we should eliminate these URLs

dmontagu (approved):

Issue #5072, opened 2023-02-18
v2 JSON schema — parsing vs. serialization


dmontagu:
In https://github.com/pydantic/pydantic/issues/1511 there is some discussion about what schema should be generated for Decimal. One of the points that has been brought up is the idea that you could produce a schema based on what is accepted/"validated", and you could produce a schema based on what is "serialized", and these could be different. (Most commonly, if multiple types pass validation, the serialization schema would be just one of those types.)

I want to dive into the implications of this a bit so we can come up with a plan for how we want to handle this, as it may be relevant for more types than just Decimal — basically any time the serialization and validation schemas are different for any reason.

--------

First, I'll note that when producing an OpenAPI spec with FastAPI, at least in the clients I've generated, the same definitions are used as "inputs" to API calls where appropriate, and as "outputs" from API calls where appropriate. This means that if a model is used as both an input and an output to (possibly different) API endpoints, there are significant practical benefits if the JSON schema produced for it works both for server-side parsing ("validation" from the perspective of pydantic) and client-side parsing ("serialization" from the perspective of pydantic). (Otherwise you could run into type errors when the generated client handles the server responses.)

When the serialization and validation schemas differ for a model, I see at least four choices that could make sense depending on how the model is used:
* "Validation": Produce a JSON schema based on the validation schema. I consider this the "naive" approach, and is how I've been proceeding so far with #5029
* "Serialization": Produce a JSON schema based on the serialization schema. This may be more appropriate for use with server responses, considering the server will know exactly how something will be serialized, even if different formats are allowed as inputs.
* "Intersection": Try to produce a JSON schema compatible with _both_ serialization and validation; in practice, this is likely to just be the serialization schema in most cases where it is sensible.
* "Union": Try to produce a JSON schema that represents all allowed validation inputs _and_ serialization outputs, even if they are different. Code written to handle the "union" as a response from the server will put some additional burden on the client to check the type of returned values even though in practice only one of the forms will be returned. But this would be the most convenient way to define a single model that was compatible with all possible representations.

Now, in principle, it should be possible to generate JSON schema in accordance with whichever of the above we want, and we could even add a flag for this. In particular, I think we _could_ do something where we pass a flag through to the JSON schema generation on a per-model (or even per-model-occurrence?) basis for whether to use the "validation" or "serialization" schema. And FastAPI/similar could handle this automatically depending on whether an annotation was for an input argument, or the return type of the endpoint.

[...]

PR #5707, opened 2023-05-06
Generate separate JSON schema for inputs vs. outputs

dmontagu:
Separates generation of JSON schema for inputs (i.e., validation) and outputs (i.e., serialization).

Closes https://github.com/pydantic/pydantic/issues/5072.

Selected Reviewer: @adriangb


dmontagu:
please review

pydantic/_internal/_schema_generation_shared.py
```diff
@@ -10,8 +10,7 @@
 from pydantic._internal._core_utils import CoreSchemaOrField
 
 if TYPE_CHECKING:
-    from pydantic.json_schema import GenerateJsonSchema
-
+    from pydantic.json_schema import GenerateJsonSchema, JsonSchemaMode
```
adriangb, line 13: Would it make sense to put `JsonSchemaMode` here to avoid circular imports?

pydantic/type_adapter.py
```diff
@@ -265,38 +273,41 @@ def json_schema(
...
-    ) -> dict[str, Any]:
-        """Generate JSON schemas for multiple models.
+    ) -> tuple[dict[tuple[JsonSchemaKeyT, JsonSchemaMode], DefsRef], JsonSchemaValue]:
+        """Generate JSON schemas for multiple type adapters.
 
         Args:
-            __types (Iterable[TypeAdapter[Any]]): The types to generate schemas for.
+            __inputs (Iterable[tuple[Hashable, JsonSchemaMode, TypeAdapter[Any]]]): The inputs to schema generation.
+                The first two items will form the keys of the first output mapping; the type adapters will provide
+                the core schemas converted into definitions in the output JSON schema.
```
adriangb, line 293: This seems pretty complex. But I think we already knew it needs to be reworked, it shouldn't hold up this PR.

adriangb (approved):
This looks great 🚀 ! My only high-level question is if we need the ability to generate a schema for `strict` validation. That can always come later.

pydantic/type_adapter.py
dmontagu: Yeah, I think the API will need to look _something_ like this (though possibly can make the key non-generic and just use the type), but I'm open to improvements.

Note to self though: either way, I need to update the annotation of `__inputs` since I changed `Hashable` to `JsonSchemaKeyT`

dmontagu:
please update (note to self)

PR #5717, opened 2023-05-09
Replace TypeError with ValidationError in forzen model

hramezani:
As discussed with @samuelcolvin the frozen model error is better to be a `ValidationError` instead of `TypeError`.
Error documentation will come later with the validation error docs PR

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -369,7 +369,12 @@ class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
             _object_setattr(self, name, value)
             return
         elif self.model_config.get('frozen', None):
-            raise TypeError(f'"{self.__class__.__name__}" is frozen and does not support item assignment')
+            error: pydantic_core.InitErrorDetails = {
+                'type': 'frozen_instance',
+                'loc': ('__root__',),
+                'input': value,
+            }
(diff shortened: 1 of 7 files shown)


hramezani:
please review

pydantic/main.py
```diff
@@ -369,7 +369,12 @@ def __setattr__(self, name: str, value: Any) -> None:
             _object_setattr(self, name, value)
             return
         elif self.model_config.get('frozen', None):
-            raise TypeError(f'"{self.__class__.__name__}" is frozen and does not support item assignment')
+            error: pydantic_core.InitErrorDetails = {
+                'type': 'frozen_instance',
+                'loc': ('__root__',),
```
dmontagu, line 374: Is this really the loc we want? I almost think an empty tuple would be better, was hoping to get rid of the awkward references to `__root__` if possible. @samuelcolvin do you have thoughts?

dmontagu: Also, given this happened in a call to `__setattr__` and you have the name, I think
```suggestion
                'loc': (name,),
```
might be better?

hramezani: I think `name` is better. Changed based on your suggestion.

PR #5720, opened 2023-05-09
Add PastDatetime and FutureDatetime types

hramezani:
Error documentation will come later with the validation error docs PR


hramezani:
Please review

pydantic/types.py
```diff
@@ -858,6 +863,40 @@ def __get_pydantic_core_schema__(
...
+        @classmethod
+        def __get_pydantic_core_schema__(
+            cls, source: type[Any], handler: GetCoreSchemaHandler
+        ) -> core_schema.CoreSchema:
+            if cls is source:
+                # used directly as a type
+                return core_schema.datetime_schema(now_op='past')
+            else:
+                schema = handler(source)
+                assert schema['type'] == 'datetime'
```
dmontagu, line 876: What is the purpose of this assertion? If there's a chance it could be hit, it may be worth giving a clearer message of what is going wrong if it _does_ get hit.

pydantic/types.py
```diff
@@ -858,6 +863,40 @@ def __get_pydantic_core_schema__(
         def __repr__(self) -> str:
             return 'NaiveDatetime'
 
+    class PastDatetime:
```
dmontagu, line 866: It's not a big deal, but if there's no reason not to, I would suggest making this (and `FutureDatetime`) a dataclass (see the items in `pydantic/annotated_arguments.py`) as it slightly improves the repr when displayed instantiated and included in an annotation.

hramezani: I've added a commit and make this and the other date and datime related types(`PastDate`, `FutureDate`, `AwareDatetime`, ..) a dataclass

pydantic/types.py
hramezani: Actually I don't know how to hit this line. I just keep it because it was in the other types(`AwareDatetime`, `FutureDate`, ...).
@samuelcolvin Do you have any idea here?

samuelcolvin: I guess the assert would fail if you use it as an annotation on something other than `datetime`.

If I'm wrong, we can remove, otherwise, yes a better error would be good.

samuelcolvin (commented):
otherwise LGTM.

please update.

pydantic/types.py
hramezani: dataclass change reverted

pydantic/types.py
hramezani: I've added a check that raises an error. added doc and test for error

docs/usage/errors.md
```diff
@@ -613,4 +613,18 @@ class MyModel(BaseModel):
     def serializer2(v):
         return v
 ```
+
+## Invalid annotated type {#invalid_annotated_type}
+
+This error is raised when annotation cannot annotated the type.
```
samuelcolvin, line 619:
```suggestion
This error is raised when an annotation cannot annotate a type.
```

samuelcolvin (approved):
otherwise lGTM.

PR #5708, opened 2023-05-07
Use return_schema for JSON schema of computed fields

dmontagu:
Requires https://github.com/pydantic/pydantic-core/pull/595 to be merged/released/used in pydantic before this has a chance of passing CI, but everything works for me locally against that branch of pydantic core.

I think probably some more tests should be added before merging this, but I'm opening the PR now for the sake of review.

Note this PR is against the branch from https://github.com/pydantic/pydantic/pull/5707 — this PR depends on separating the JSON  schema for serialization and validation, which is done in #5707. If/when that branch gets merged, this PR can be converted to being against the `main` branch.

I'll wait to tell hooky "please review" until at least one of https://github.com/pydantic/pydantic-core/pull/595 and https://github.com/pydantic/pydantic/pull/5707 get merged.

PR #5727, opened 2023-05-10
Uprev pydantic-core to 0.31.0

adriangb:
- Updates to computed_field: a return type is now required. It will be inferred from return type annotations
- Replaces `is_instance` validators w/ `json_or_python` validators

Requires https://github.com/pydantic/pydantic-core/pull/598 and https://github.com/pydantic/pydantic-core/pull/599

All tests passing locally, will mark as ready for review once we cut the core release

Selected Reviewer: @dmontagu


pydantic/_internal/_generate_schema.py
```diff
@@ -1476,11 +1476,13 @@ def _common_field(
     }
 
 
-def generate_computed_field(d: dict[str, Decorator[ComputedFieldInfo]]) -> list[core_schema.ComputedField] | None:
+def generate_computed_field(
```
dmontagu, line 1479: Does it make sense to just make this a method of `GenerateSchema`? That's what I did in https://github.com/pydantic/pydantic/pull/5708 but I'm happy to keep your approach of still returning a list. Not a big deal.

pydantic/_internal/_decorators.py
```diff
@@ -489,22 +489,36 @@ def unwrap_wrapped_function(
...
-        )
+            property,
+            *cached_property_list,
+        ]
     else:
-        all = partial, partialmethod
+        all = [partial, partialmethod, property, *cached_property_list]
 
-    while isinstance(func, all):
+    while isinstance(func, tuple(all)):
```
dmontagu, line 513: another super minor thing — any reason not to make `all` a tuple in the first place?

pydantic/_internal/_std_types_schema.py
```diff
@@ -103,28 +103,39 @@ def update_schema(_, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
...
         # Allow all types from JSON to get better error messages (numeric types will still fail validation in to_enum)
-        strict = core_schema.is_instance_schema(enum_type, json_types={'int', 'str', 'float'}, json_function=to_enum)
+        strict = core_schema.json_or_python_schema(
+            json_schema=core_schema.no_info_after_validator_function(to_enum, core_schema.str_schema()),
+            python_schema=core_schema.is_instance_schema(enum_type),
+        )
     elif issubclass(enum_type, float):
         updates['type'] = 'numeric'
         lax = core_schema.chain_schema([core_schema.float_schema(), to_enum_validator])
         # Allow str from JSON to get better error messages (str will still fail validation in to_enum)
```
dmontagu, line 129: Should these comments be removed now? I'm not sure they are still relevant given `json_types` is gone

pydantic/_internal/_generate_schema.py
adriangb: done

pydantic/_internal/_decorators.py
adriangb: Just that type checkers are less happy:
image
No idea why `isinstance` _requires_ a tuple, I suspect that is conceptually wrong.

pydantic/fields.py
```diff
@@ -748,7 +746,7 @@ class ComputedFieldInfo:
 
     decorator_repr: typing.ClassVar[str] = '@computed_field'
     wrapped_property: property
-    json_return_type: _core_schema.JsonReturnTypes | None
+    return_type: type[Any]
```
dmontagu, line 749: Is it okay that we use `Any` below but `type[Any]` here? I can understand why you might do this, but just want to make sure it was intentional.

adriangb: My reasoning was that `type[Any]` is technically incompatible if you try to assign `int | str`, but internally we pretty much want to treat it as a type, so `type[Any]` is more semantically meaningful

pyproject.toml
```diff
@@ -58,7 +61,7 @@ requires-python = '>=3.7'
 dependencies = [
     'typing-extensions>=4.5.0',
     'annotated-types>=0.4.0',
-    "pydantic-core==0.30.0",
+    "pydantic-core @ git+https://github.com/pydantic/pydantic-core.git@main",
```
dmontagu, line 64: I get that you are going to delete this, but if this works with CI, any chance we could keep it in this file, just commented out? Will make it easier for me to remember to set it up this way when I want to pre-emptively run CI against a not-yet-released pydantic core.

adriangb:
please update

adriangb:
please review

PR #5733, opened 2023-05-10
Improve validation errors doc

hramezani:

### DIFF
--- a/docs/usage/validation_errors.md
+++ b/docs/usage/validation_errors.md
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
 
 ## `arguments_type`
 
+This error is raised when arguments passed to a function are not
+a tuple, list or a dictionary.
 
 ## `assertion_error`
 
+This error is raised when there is a failing assertion during the validation.
+
+```py test="skip" lint="skip" upgrade="skip"
+from pydantic import BaseModel, field_validator
+
+
+class Model(BaseModel):
+    x: int
+
+    @field_validator('x')
+    @classmethod
+    def force_x_positive(cls, v):
+        assert v > 0
+        return v
+
+Model(x=-1)
+```
 
 ## `bool`
 
 
 ## `bool_parsing`
 
+This error is raised when the value is an invalid boolean string.
+
+```py test="skip" lint="skip" upgrade="skip"
+from pydantic import BaseModel
+
+
+class Model(BaseModel):
+    x: bool
+
+print(Model(x='test'))
+```
 
 ## `bool_type`
 
+This error is raised when the value type is not valid for a boolean field.
+
+```py test="skip" lint="skip" upgrade="skip"
+from pydantic import BaseModel
+
+
+class Model(BaseModel):
+    x: bool
+
+Model(x=None)
+```
 
 ## `bytes_too_long`
 
+This error is raised when the bytes value length is greater than `Field.max_length`.
+
+```py test="skip" lint="skip" upgrade="skip"
+from pydantic import BaseModel, Field
+
+
+class Model(BaseModel):
+    x: bytes = Field(max_length=3)
+
+Model(x=b'test')
+```
 
 ## `bytes_too_short`
 
+This error is raised when the bytes value length is less than `Field.min_length`.
+
+```py test="skip" lint="skip" upgrade="skip"
+from pydantic import BaseModel, Field
+
+
+class Model(BaseModel):
+    x: bytes = Field(min_length=3)
+
+Model(x=b't')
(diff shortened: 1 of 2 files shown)


docs/usage/validation_errors.md
```diff
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
 
 ## `arguments_type`
 
+This error is raised when arguments passed to a function are not
+a tuple, list or a dictionary.
```
dmontagu, line 7:
```suggestion
a tuple, list or dictionary.
```

docs/usage/validation_errors.md
```diff
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
...
 
 ## `assertion_error`
 
+This error is raised when there is a failing assertion during the validation.
```
dmontagu, line 11:
```suggestion
This error is raised when a failing assert statement is encountered during validation.
```

docs/usage/validation_errors.md
```diff
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
...
+
+Model(x=-1)
+```
 
 ## `bool`
 
 
 ## `bool_parsing`
 
+This error is raised when the value is an invalid boolean string.
```
dmontagu, line 34:
```suggestion
This error is raised when the value is a string that is not valid for coercion to a boolean.
```

docs/usage/validation_errors.md
```diff
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
...
+
+```py test="skip" lint="skip" upgrade="skip"
+from pydantic import BaseModel
+
+
+class Model(BaseModel):
+    x: bool
+
+print(Model(x='test'))
```
dmontagu, line 43:
```suggestion
print(Model(x='true'))  # OK
print(Model(x='test'))
```
I think it's maybe worth making clear via this kind of simple example here that some strings _are_ acceptable

docs/usage/validation_errors.md
```diff
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
...
+class Model:
+    y: Nested
+
+
+Model(y=Nested(x=1))
+```
 
 ## `date_from_datetime_inexact`
 
+This error is raised when the datetime value provided for a date field is not an exact date.
```
dmontagu, line 176:
```suggestion
This error is raised when the datetime value provided for a date field has a nonzero time component.
```

I'm not sure if my above comment is exactly right, but I think to a newcomer it may not necessarily be clear what is meant by "an exact date". In particular, if someone is opening this link, they were probably hoping for more clarity than that phrase provides, I think we can explain a bit more clearly here. (Something like "For a timestamp to parse into a field of type `date`, the time components must be all zero, and show an example that doesn't fail below.)

docs/usage/validation_errors.md
```diff
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
...
+    x: PastDate
+
+
+Model(x=date.today() + timedelta(1))
+```
+
 
 ## `date_type`
 
+This error is raised when the value type is not date for a strict date field.
```
dmontagu, line 263:
```suggestion
This error is raised when the value type is not of type `date` for a strict date field.
```

dmontagu (commented):
I kind of lost steam and didn't go through everything, but this is great. I think we can do another copy pass at some point, this is already pretty good though.

One thing I think may be fairly important — I think it would be good if we didn't skip these tests, as I think it's fairly important that if we change something that causes the example code to not behave correctly, that will be unfortunate for users.

Would it be possible to make these examples actually run (through use of `try: except:` or similar)? I recognize that might make things more verbose but otherwise I think the chance of one of the examples eventually quietly breaking is fairly high.

docs/usage/validation_errors.md
hramezani: only string I think

docs/usage/validation_errors.md
```diff
@@ -3,251 +3,1470 @@ with some suggestions on how to fix them.
...
+class Model(BaseModel):
+    x: date
+
+
+Model(x=datetime(2023, 1, 1, 12))
+```
 
 ## `date_from_datetime_parsing`
 
+This error is raised when the datetime value provided for a date field is invalid.
```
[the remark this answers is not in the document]
hramezani: Yes, I think only for `string`.
Changed it.

hramezani:
@dmontagu I've addressed all your comments and made all the tests runnable as you suggested.

PR #5736, opened 2023-05-10
Fixes discriminated unions not working on aliased literal fields. Issue #3849.

### LINKED ISSUE #3849: Discriminated union does not work with aliased literal field
### Checks

# Bug

```
             pydantic version: 1.9.0
            pydantic compiled: True
                 install path: /home/ljnsn/.virtualenvs/py39/lib/python3.9/site-packages/
pydantic
               python version: 3.9.7 (default, Sep 10 2021, 14:59:43)  [GCC 11.2.0]
                     platform: Linux-5.15.23-76051523-generic-x86_64-with-glibc2.34
```
It seems like the discriminator field for a discriminated union is only recognized when it is passed by its alias, even when `allow_population_by_field_name = True`. I would expect to be able to pass the discriminator field by field name, just like any other field, when that config option is set.

```py
from typing import Literal, Union

class Base(BaseModel):
    class Config:
        allow_population_by_field_name = True


class Cat(Base):
    pet_type: Literal["cat"] = Field(alias="petType")
    meows: int


class Dog(Base):
    pet_type: Literal["dog"] = Field(alias="petType")
    barks: float


class Lizard(Base):
    pet_type: Literal["reptile", "lizard"] = Field(alias="petType")
    scales: bool


class Model(Base):
    pet: Union[Cat, Dog, Lizard] = Field(..., discriminator="pet_type")
    n: int


print(Dog(**{"pet_type": "dog", "barks": 3.14}))

print(Model(pet={"petType": "dog", "barks": 3.14}, n=1))

print(Model(pet={"pet_type": "dog", "barks": 3.14}, n=1))
# > pydantic.error_wrappers.ValidationError: 1 validation error for Model pet
#   Discriminator 'pet_type' is missing in value (type=value_error.discriminated_union.miss
#   ing_discriminator; discriminator_key=pet_type)

```

benwah:
## Change Summary

The issue is described here: https://github.com/pydantic/pydantic/issues/3849

TL;DR:

When you have an aliased field, and it uses a discriminated union; you can no longer use the un-aliased field name to instantiate the model.

According to the discussion, this is not going to be a problem once Pydantic 2.x releases, but it's definitly problematic for 1.10.x which I'm currently using.

Here's the test failure without this change:

```
============================================================== FAILURES ==============================================================
_________________________________________ test_discriminated_union_validation_aliased_field __________________________________________

    def test_discriminated_union_validation_aliased_field():
        def to_upper(field_name: str) -> str:
            return field_name.upper()
    
    
        class PyBase(BaseModel):
            class Config:
                allow_population_by_field_name = True
                alias_generator = to_upper
    
    
        class Dog(PyBase):
            pet_type: Literal["dog"]
            first_name: str
    
    
        class Cat(PyBase):
            pet_type: Literal["cat"]
            first_name: str
    
    
        Pet = Annotated[Cat | Dog, Field(discriminator="pet_type")]
    
    
        class Model(PyBase):
            pet: Pet
    
    
>       m = Model.parse_obj({'pet': {'pet_type': 'cat', "first_name": "Benoit"}})

tests/test_discrimated_union.py:116: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pydantic/main.py:526: in parse_obj
    return cls(**obj)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

__pydantic_self__ = Model(), data = {'pet': {'first_name': 'Benoit', 'pet_type': 'cat'}}, values = {}, fields_set = {'pet'}
validation_error = ValidationError(model='Model', errors=[{'loc': ('PET',), 'msg': "Discriminator 'pet_type' is missing in value", 'type': 'value_error.discriminated_union.missing_discriminator', 'ctx': {'discriminator_key': 'pet_type'}}])

        """
        Create a new model by parsing and validating input data from keyword arguments.
    
        Raises ValidationError if the input data cannot be parsed to form a valid model.
        """
        if validation_error:
E           pydantic.error_wrappers.ValidationError: 1 validation error for Model
E           PET
E             Discriminator 'pet_type' is missing in value (type=value_error.discriminated_union.missing_discriminator; discriminator_key=pet_type)

pydantic/main.py:341: ValidationError
====================================================== short test summary info =======================================================
FAILED tests/test_discrimated_union.py::test_discriminated_union_validation_aliased_field - pydantic.error_wrappers.ValidationError...
========================================================= 1 failed in 0.17s ==========================================================
```

## Related issue number

Fix #3849

## Checklist

### DIFF
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -1107,7 +1107,10 @@ class ModelField(Representation):
         assert self.discriminator_alias is not None
 
         try:
-            discriminator_value = v[self.discriminator_alias]
+            try:
+                discriminator_value = v[self.discriminator_alias]
+            except KeyError:
+                discriminator_value = v[self.discriminator_key]
         except KeyError:
             return v, ErrorWrapper(MissingDiscriminator(discriminator_key=self.discriminator_key), loc)
         except TypeError:
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -1110,7 +1110,10 @@ class ModelField(Representation):
             try:
                 discriminator_value = v[self.discriminator_alias]
             except KeyError:
-                discriminator_value = v[self.discriminator_key]
+                if self.model_config.allow_population_by_field_name:
+                    discriminator_value = v[self.discriminator_key]
+                else:
+                    raise
         except KeyError:
             return v, ErrorWrapper(MissingDiscriminator(discriminator_key=self.discriminator_key), loc)
         except TypeError:
--- a/changes/5736-benwah.md
+++ b/changes/5736-benwah.md
@@ -1,0 +1,1 @@
+This solves the (closed) issue #3849 where aliased fields that use discriminated union fail to validate when the data contains the non-aliased field name.
--- a/tests/test_discrimated_union.py
+++ b/tests/test_discrimated_union.py
@@ -285,6 +285,64 @@ def test_discriminated_union_basemodel_instance_value_with_alias():
     assert Top(sub=B(literal='b')).sub.literal == 'b'
 
 
+def test_discriminated_union_model_with_alias():
+    class A(BaseModel):
+        literal: Literal['a'] = Field(alias='lit')
+
+    class B(BaseModel):
+        literal: Literal['b'] = Field(alias='lit')
+
+        class Config:
+            allow_population_by_field_name = True
+
+    class TopDisallow(BaseModel):
+        sub: Union[A, B] = Field(..., discriminator='literal', alias='s')
+
+    class TopAllow(BaseModel):
+        sub: Union[A, B] = Field(..., discriminator='literal', alias='s')
+
+        class Config:
+            allow_population_by_field_name = True
+
+    assert TopDisallow.parse_obj({'s': {'lit': 'a'}}).sub.literal == 'a'
+    assert TopDisallow.parse_obj({'s': {'literal': 'b'}}).sub.literal == 'b'
+
+    with pytest.raises(ValidationError) as exc_info:
+        TopDisallow.parse_obj({'s': {'literal': 'a'}})
+
+    assert exc_info.value.errors() == [
+        {'loc': ('s', 'A', 'lit'), 'msg': 'field required', 'type': 'value_error.missing'},
+    ]
+
+    with pytest.raises(ValidationError) as exc_info:
+        TopDisallow.parse_obj({'sub': {'lit': 'a'}})
+
+    assert exc_info.value.errors() == [
+        {'loc': ('s',), 'msg': 'field required', 'type': 'value_error.missing'},
+    ]
+
+    assert TopAllow.parse_obj({'s': {'lit': 'a'}}).sub.literal == 'a'
+    assert TopAllow.parse_obj({'s': {'lit': 'b'}}).sub.literal == 'b'
+    assert TopAllow.parse_obj({'s': {'literal': 'b'}}).sub.literal == 'b'
+    assert TopAllow.parse_obj({'sub': {'lit': 'a'}}).sub.literal == 'a'
+    assert TopAllow.parse_obj({'sub': {'lit': 'b'}}).sub.literal == 'b'
+    assert TopAllow.parse_obj({'sub': {'literal': 'b'}}).sub.literal == 'b'
+
+    with pytest.raises(ValidationError) as exc_info:
+        TopAllow.parse_obj({'s': {'literal': 'a'}})
+
+    assert exc_info.value.errors() == [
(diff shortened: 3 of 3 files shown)


benwah:
please review

benwah:
Hypothesis test suite is failing in my local tests, with and without these changes. So I don't think this is a regression.

pydantic/fields.py
```diff
@@ -1107,7 +1107,10 @@ def _validate_discriminated_union(
         assert self.discriminator_alias is not None
 
         try:
-            discriminator_value = v[self.discriminator_alias]
+            try:
+                discriminator_value = v[self.discriminator_alias]
+            except KeyError:
+                discriminator_value = v[self.discriminator_key]
```
dmontagu, line 1113: Could you check the value of `cls.Config. allow_population_by_field_name` or similar before attempting to use this attribute?

In this case, you may need to rework the way the try-excepts work to make sure it still hits the KeyError even when population by field name is not allowed.

benwah: Makes sense, will do!

benwah: @dmontagu I simplified my test models (no alias generator) and made more exhaustive tests.

It looks like there's no need to check `cls.Config.allow_population_by_field`. 

Looks like that's already taken care of by how the values are populated: https://github.com/pydantic/pydantic/blob/1.10.X-fixes/pydantic/main.py#L1057-L1059

dmontagu: @benwah this is an exotic example, but it demonstrates why the config check is still necessary:
```python
from typing import Annotated, Literal

from pydantic import Field


def to_upper(field_name: str) -> str:
    return field_name.upper()


class PyBase(BaseModel):
    class Config:
        pass


class Dog(PyBase):
    x: Literal["dog"] = Field("dog", alias='y')
    y: None = Field(None, alias='x')


class Cat(PyBase):
    x: Literal["cat"] = Field("cat", alias='y')
    y: None = Field(None, alias='x')


Pet = Annotated[Cat | Dog, Field(discriminator="x")]


class Model(PyBase):
    pet: Pet

m = Model.parse_obj({'pet': {'x': 'abc'}})  # note: due to aliases, x is NOT the discriminator
# On main:
"""
pet
  Discriminator 'x' is missing in value
"""

# On your branch:
"""
pet
  No match for discriminator 'x' and value 'abc' (allowed values: 'cat', 'dog')
"""
```
While the error is admittedly confusing either way (always a possibility when you have aliases that are actually names of other fields..), I would argue the error in this scenario is more correct on main than not.

Considering `model` is present in this function call I think it should be reasonable to check the value of the config setting and eliminate this (or other possible issues I haven't considered) as a potential regression.

benwah: I've implemented the changes you've suggested, but I'm wondering it now makes sense that this call should fail:

https://github.com/pydantic/pydantic/pull/5736/commits/b17a4c95d49ec2e878f6a298b1e4db448c09ec1a#diff-e2a21a51ea103c68ad6398fa0acb2c702bd756957159ab9498354e8227cdc5bbL308-R319

The discriminator is owned by the sub-model `B`, which should accept populating by field name.

However now we consider that the discriminator is missing.

Also shouldn't the `loc` be `("s", "B", "lit")`?

benwah: .. also please consider this change in error message:

https://github.com/pydantic/pydantic/commit/b17a4c95d49ec2e878f6a298b1e4db448c09ec1a#diff-e2a21a51ea103c68ad6398fa0acb2c702bd756957159ab9498354e8227cdc5bbL313-R330

dmontagu: I see your point; in v2 we go by the property name rather than the alias when specifying the discriminator so, although it is perhaps even a bit _more_ restrictive, this particular inconsistency has been eliminated.

I see arguments in favor of each approach here, but my gut reaction is that I'd be more comfortable being more restrictive rather than less (so would prefer to require the config setting be set). If that is enough to solve your problem, let's go with that. (I.e., with the branch as it currently stands).

benwah: Sounds good! This is indeed enough to fix my problem.

PR #5737, opened 2023-05-10
Don't try to automatically handle subclasses of known types

adriangb:
Handles a couple of them, leaving Mappings (OrderedDict specifically) out to avoid conflicting with @lig 's work.
I'm also trying to do a better job of handling constraints on these types, e.g. `MaxLen` on `Path`.

Selected Reviewer: @hramezani


adriangb:
please review
(test failures are due to GHA being flakey)

tests/test_types.py
```diff
@@ -1536,7 +1537,10 @@ class Model(BaseModel):
     ],
 )
 def test_invalid_schema_constraints(kwargs, type_):
-    with pytest.raises(SchemaError, match='Invalid Schema:\n.*\n  Extra inputs are not permitted'):
+    match = (
+        r'(:?Invalid Schema:\n.*\n  Extra inputs are not permitted)|(:?The following constraints cannot be applied to)'
+    )
+    with pytest.raises((SchemaError, TypeError), match=match):
```
adriangb, line 1543: It's now a mix, depending on where the error is coming from. It will be just the latter since my long term plan is for constraints to not automatically insert themselves into the core schema blindly, instead they will check if the core schema they see accepts them as constraints and if not insert a Python function validator version of the consteraint.

tests/test_types.py
```diff
@@ -2472,23 +2476,73 @@ class Model(BaseModel):
...
+            'msg': 'Input should be a valid string',
+            'input': not_a_valid_input_type,
+        },
+        {
+            'type': 'bytes_type',
+            'loc': ('function-after[uuid_validator(), union[str,bytes]]', 'bytes'),
+            'msg': 'Input should be a valid bytes',
+            'input': not_a_valid_input_type,
+        },
+    ]
```
adriangb, line 2530: These error's aren't perfect, but I think they're better than what we had before. If you read between the lines and noise it's saying that the input can be one of a UUID instance, a string or bytes, checked in that order, which is exactly what should be happening.

tests/test_types.py
```diff
@@ -2926,6 +2980,15 @@ class Model(BaseModel):
     assert Model.model_validate_json(json.dumps({'foo': str(value)})).foo == result
 
 
+def test_path_validation_constrained():
+    ta = TypeAdapter(Annotated[Path, Field(min_length=9, max_length=20)])
```
adriangb, line 2984: I think we should be using `TypeAdapter` more internally, it makes the test simpler and removes the model from the core schema, which makes debugging easier.

tests/test_types.py
```diff
@@ -2969,7 +3032,19 @@ class Model(BaseModel):
...
+            'msg': 'Input should be an instance of Path',
+            'input': 123,
+            'ctx': {'class': 'Path'},
+        },
+        {
+            'type': 'string_type',
+            'loc': ('foo', 'function-after[path_validator(), str]'),
+            'msg': 'Input should be a valid string',
+            'input': 123,
+        },
```
adriangb, line 3047: Same here: the input can be either a string or a Path instance, the error is spot on.

tests/test_types.py
```diff
@@ -4127,50 +4202,6 @@ class Model(BaseModel):
     assert Model(v=deque((1, 2, 3))).model_dump_json() == '{"v":[1,2,3]}'
 
 
-def test_deque_any_maxlen():
```
adriangb, line 4130: I removed support for "preserving" maxlen on a dequeue.
The issue is that we can only do that in some cases.
In particular, we can't do that in JSON mode.
I think we should stick to our promise of "we only guarantee that we give you what is reflected in the type annotation".
If you want the output dequeue to always have a given maxlen then you need to add a validator that enforces that; we can't know you're intended maxlen just based on `deque[int]`.

tests/test_types.py
```diff
@@ -4401,7 +4432,9 @@ def test_custom_generic_containers():
     T = TypeVar('T')
 
     class GenericList(List[T]):
-        pass
+        @classmethod
+        def __get_pydantic_core_schema__(cls, source_type: Any, handler: GetCoreSchemaHandler) -> CoreSchema:
+            return core_schema.no_info_after_validator_function(GenericList, handler(List[get_args(source_type)[0]]))
```
adriangb, line 4437: These all still work, we just require you to write these two lines of code now to confirm how you want them to work. I imagine we could provide some sort of decorator or mixin that does this for you, something like:

```python
class PydanticCustomSequence:
    @classmethod
    def __get_pydantic_core_schema__(cls, source_type: Any, handler: GetCoreSchemaHandler) -> CoreSchema:
        return core_schema.no_info_after_validator_function(cls, handler(cls[get_args(source_type)[0]]))
```

It would need to be a bit more complex to handle `list` on Python < 3.9, handle bare types (`x: GenericList`), etc.

tests/test_json_schema.py
```diff
@@ -2186,6 +2186,10 @@ class Example(BaseModel):
 
 def test_path_modify_schema():
     class MyPath(Path):
+        @classmethod
+        def __get_pydantic_core_schema__(cls, _source_type: Any, handler: GetCoreSchemaHandler) -> CoreSchema:
+            return handler(Path)
```
adriangb, line 2191: Some of these are _trivial_, I'm happy to ask users to write this to avoid doing the wrong thing.
But also as per below, it would be relatively easy to provide some sort of opt-in functionality for these "simple" cases.

tests/test_annotated.py
```diff
@@ -233,6 +233,7 @@ class _(BaseModel):
         x: Annotated[CustomType, GroupedMetadataMarker(), PydanticMetadata()]
 
     assert calls == [
+        'GroupedMetadataMarker:iter',
```
adriangb, line 236: It seems like we end up calling `GroupedMetadataMarker` once more than we did before, I'm not sure why. But `GroupedMetadata` should _definitely_ be idempotent, so it shouldn't really be a problem.

pyproject.toml
```diff
@@ -65,8 +65,6 @@ dependencies = [
 ]
 dynamic = ['version', 'readme']
 
-entry-points.hypothesis = {_ = 'pydantic._hypothesis_plugin'}
```
adriangb, line 68: Unrelated to this PR but caused issues for me locally

pydantic/_internal/_validators.py
```diff
@@ -105,21 +104,6 @@ def _import_string_logic(dotted_path: str) -> Any:
         raise ImportError(f'Module "{module_path}" does not define a "{class_name}" attribute') from e
 
 
-def uuid_validator(__input_value: str | bytes) -> UUID:
```
adriangb, line 108: Moved inline to where it gets used. We should probably get rid of this file eventually.

pydantic/_internal/_generate_schema.py
```diff
@@ -1153,15 +1129,28 @@ def _annotated_schema(self, annotated_type: Any) -> core_schema.CoreSchema:
...
 
-        if obj is Decimal:
-            return std_types.decimal_prepare_pydantic_annotations
+        for gen in (
+            std_types.decimal_prepare_pydantic_annotations,
+            std_types.sequence_like_prepare_pydantic_annotations,
+            std_types.datetime_prepare_pydantic_annotations,
+            std_types.uuid_prepare_pydantic_annotations,
+            std_types.path_schema_prepare_pydantic_annotations,
+        ):
```
adriangb, line 1150: I think we should collapse these into a single `std_types.match_to_known_type` or something that does this for loop internally. But open to other ideas for how to organize this.

tests/test_generics.py
```diff
@@ -1174,9 +1185,9 @@ class Model(BaseModel, Generic[T, KT, VT]):
 
     # The following assertions are just to document the current behavior, and should
     # be updated if/when we do a better job of respecting the exact annotated type
-    assert type(m.counter_field) is Counter.__origin__
+    assert type(m.counter_field) is CustomCounter
     assert type(m.default_dict_field) is dict
-    assert type(m.deque_field) is deque
+    assert type(m.deque_field) is CustomDeque
```
adriangb, line 1190: I think we have several of these inconsistencies where we handled subclasses in some cases but not others. The main impetus for this PR is to resolve these things by not doing the wrong thing silently and instead requiring some input from users on how they want us to handle their custom types.
Given how much easier it is now to handle both first-party and third-party custom types in V2 compared to V1, this is a very reasonable thing to do.

(this was all discussed offline, I'm just leaving it here for the record)

pydantic/_internal/_std_types_schema.py
```diff
@@ -307,141 +294,296 @@ def validate(self, __input_value: int | float | str) -> Decimal:  # noqa: C901 (
...
                 instance_schema,
-                core_schema.no_info_after_validator_function(path_validator, core_schema.str_schema()),
+                core_schema.no_info_after_validator_function(path_validator, constrained_str_schema),
             ],
-            custom_error_type='path_type',
-            custom_error_message='Input is not a valid path',
-            strict=True,
+            # custom_error_type='path_type',
+            # custom_error_message='Input is not a valid path',
+            # strict=True,
```
dmontagu, line 439: should these commented lines be removed?

adriangb: Oh yes good catch

pydantic/_internal/_std_types_schema.py
```diff
@@ -307,141 +294,296 @@ def validate(self, __input_value: int | float | str) -> Decimal:  # noqa: C901 (
...
-    if obj == deque:
-        # bare `deque` type used as annotation
-        return _deque_any_schema()
+        if self.mapped_origin in (list, set, frozenset):
+            if self.mapped_origin is list:
+                constrained_schema = core_schema.list_schema(items_schema, **metadata)
+            elif self.mapped_origin is set:
+                constrained_schema = core_schema.set_schema(items_schema, **metadata)
+            else:
+                assert self.mapped_origin is frozenset  # safety check in case we forget to add a case
```
dmontagu, line 496: might be more typesafe to do an `else: assert_never(self.mapped_origin)`, if that works anyway. Not a big deal

adriangb: I don’t think this part of the codebase is well typed enough for that to work

pyproject.toml
dmontagu: Maybe a better idea to comment this out rather than remove it entirely? So we at least remember to look into it? Maybe a non-issue if it will be necessary again when we work on the hypothesis plugin again..

adriangb: We’ll need to add it back to get hypothesis working, so I don’t think there’s much of a point in keeping it around

tests/test_generics.py
dmontagu: I think the most important thing given this change is making sure users have an easy time finding the docs for custom types and/or examples like this (even if just in tests)

adriangb: We actually point you to `__get_pydantic_core_schema__` in the error you get, but I agree an explicit test that shows that would be good in this PR

tests/test_types.py
dmontagu: I think this is a regression — if we receive a value that is of type `deque` and it already has a `maxlen` set, why shouldn't we preserve the value? Obviously if we receive a json object (or are otherwise in JSON mode) it makes sense that it can't be preserved, but if we get an object already of type `deque` during validation I don't see why we shouldn't try to preserve the `maxlen`. I would expect most users using `deque` as an annotation are more concerned with the python-side API anyway.

Is it hard to preserve the maxlen for some reason, or are there cases where it would have to be lost that don't involve conversion to JSON? Considering this was an explicitly raised issue that I recently addressed and closed, I would like to preserve the fix unless it's really problematic.

> I think we should stick to our promise of "we only guarantee that we give you what is reflected in the type annotation".

As far as I'm aware, the value of `maxlen` is an attribute of things of type `deque`, not tied to the type hint. I recognize it's not a perfect analogy, but I think not preserving the value of `maxlen` is more akin to not preserving the value of a field of a BaseModel than it is to returning a value of a less specific type (when a field is annotated as a less specific type).

I recognize it's not preserved under a round-trip to JSON but I think if you are using a deque with a maxlen, you are doing things from the python API anyway.

tests/test_json_schema.py
dmontagu: Might be a good idea to add a link to custom types docs in the error message that gets raised if something is not a recognized type

tests/test_types.py
adriangb: I’m happy to re evaluate, it’s not like it can’t be implemented.

Take the case of passing in a list: you end up with a Deque with no maxlen. Maybe you could argue there is no issue if you’re in Python in strict mode, but I fear that users might assume that the output has a maxlen when it doesn’t.

dmontagu (approved):
I worry that this could break a significant number of custom types, hard to know until we get the feedback.

As I've noted, I think the main mitigation will be making it easy to discover how to define custom types, and/or perhaps re-adding special handling for cases where there is a reasonable init (e.g., a custom subclass of `Path`, which does have an `__init__`) if we get multiple people requesting it be changed back. (I think it's unlikely we'll get a lot of feedback about it either way, but at the same time we've gotten a lot of issues over the years that I would have considered unlikely.)

Anyway, I'll approve this for now to keep you moving, but please do consider my comment about deque/maxlen below before merging.

tests/test_types.py
dmontagu: I think that's reasonable since that's what happens for e.g. `deque([1, 2, 3])`. Perhaps a better example would be — what happens with private attributes of a BaseModel during validation? I'm assuming they get preserved (though not 100% sure lol)

adriangb:
I’ll revert the deque behavior and add a link to custom types doc or better info in the error messages (along with tests for those).

Agreed about the feedback and use cases. It would be good to get the feature requests and understand how that first into our new tooling for custom types then implement subclass handling and make decisions from then. Certainly better than the inconsistent / buggy behavior we have now.

adriangb:
Both of those tasks done, sending this to merge

PR #5740, opened 2023-05-11
stop BaseModel inheriting from `_repr.Representation`

samuelcolvin:
adds complexity to the MRO, and can be confusing when reading the code - hence better to remove the inheritance.

Selected Reviewer: @adriangb


adriangb:
Is this duplicating the methods? Or are they different? If it's the former maybe just do `__repr__ = _repr.Representation.__repr__`?

samuelcolvin:
please review.

samuelcolvin:
> Is this duplicating the methods? Or are they different? If it's the former maybe just do `__repr__ = _repr.Representation.__repr__`?

good idea, will do.

adriangb (approved):

Issue #4687, opened 2022-10-29
Nested defaultdict loses its default factory


Thoufak:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

Pydantic fails to properly construct the defaultdict when it is nested into another defaultdict.
After the model construction, the outer dict will have the correct default_factory, but the inner one will have its factory set to None.

### Example Code

```Python
from collections import defaultdict
from typing import DefaultDict
from pydantic import BaseModel


class TestModel(BaseModel):
    d: DefaultDict[int, DefaultDict[int, int]]


model = TestModel(d=defaultdict(lambda: defaultdict(int)))
parsed = TestModel.parse_raw(model.json())
print(parsed.d[1])     # defaultdict(None, {})
print(parsed.d[1][2])  # KeyError: 2
```

### Python, Pydantic & OS Version

```Text
pydantic version: 1.9.1
            pydantic compiled: True
                 install path: /Users/abc/PycharmProjects/Pybugtic/resources/venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.5 (default, Jul  9 2022, 23:20:38)  [Clang 13.1.6 (clang-1316.0.21.2.5)]
                     platform: macOS-12.3-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
- [X] [Data validation/parsing](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://pydantic-docs.helpmanual.io/usage/exporting_models/) - `.dict()` and `.json()`
- [ ] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)
- [ ] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)
- [ ] [Model Config](https://pydantic-docs.helpmanual.io/usage/model_config/)
[...]

PR #5744, opened 2023-05-11
Do not try to automatically handle subclasses of mappings

### LINKED ISSUE #3417: DefaultDict default function gets replaced with type creator in object creation
### Checks

# Bug

```
             pydantic version: 1.8.2
            pydantic compiled: True
                 install path: /usr/local/lib/python3.9/site-packages/pydantic
               python version: 3.9.4 (default, Apr  5 2021, 01:50:46)  [Clang 12.0.0 (clang-1200.0.32.29)]
                     platform: macOS-11.6-x86_64-i386-64bit
```

```py
from collections import defaultdict
from typing import DefaultDict

import pydantic

    values: DefaultDict[str, int]

example = Model(values=defaultdict(lambda: 6, {"low": 3, "high": 9}))
assert example.values["low"] == 3  # passes
assert example.values["high"] == 9  # passes
assert example.values["medium"] == 6, "defaultdict default function not retained"  # fails
```

This breaks the `constant_factory` example as provided on https://docs.python.org/3/library/collections.html#defaultdict-examples as otherwise legitimate and useful use of `defaultdict`. I was hoping to use this in a situation where there are different default values for different instances of the model, so source-level constants do not work.

adriangb:
- Better handling of constraints for mappings
- Better handling of DefaultDict
    - only infer a default factory for obvious cases (int, float, bool, list, set and dict basically) and fail for other cases
    - allow specifying a default factory via `OrderedDict[int, Annotated[list, Field(default_factory=...)]]`
- Better handling of OrderedDict

Fixes #3417, #4687

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -413,15 +413,6 @@ class GenerateSchema:
             return self._pattern_schema(obj)
         elif obj is collections.abc.Hashable or obj is typing.Hashable:
             return self._hashable_schema()
-        elif isinstance(obj, type):
-            if obj is dict:
-                return self._dict_schema(obj)
-            if issubclass(obj, dict):
-                # TODO: We would need to handle generic subclasses of certain typing dict subclasses here
-                #   This includes subclasses of typing.Counter, typing.DefaultDict, and typing.OrderedDict
-                #   Note also that we may do a better job of handling typing.DefaultDict by inspecting its arguments.
-                return self._dict_subclass_schema(obj)
-            # probably need to take care of other subclasses here
         elif isinstance(obj, typing.TypeVar):
             return self._unsubstituted_typevar_schema(obj)
         elif is_finalvar(obj):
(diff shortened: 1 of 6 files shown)


adriangb:
please review

lig:
I would try copying here this test https://github.com/pydantic/pydantic/pull/5598/files#diff-cbcfeba88238560f75e880d530f4cd82ab6bf90247012e8f664281941f8da1f7R3782 from #5598 as well.

The test were failing before.

pydantic/_internal/_std_types_schema.py
```diff
@@ -614,56 +600,185 @@ def sequence_like_prepare_pydantic_annotations(source_type: Any, annotations: It
...
+                f' Only {allowed_msg} are supported, other types require an explicit default factory'
+                ' ' + instructions
+            )
+        return allowed_default_types[values_type_origin]
 
-    if keys_arg == typing.Any and values_arg == typing.Any:
-        # `OrderedDict[Any, Any]`
-        return _ordered_dict_any_schema()
+    # Assume Annotated[..., Field(...)]
+    field_info = next((v for v in get_args(values_source_type) if isinstance(v, FieldInfo)), None)
```
adriangb, line 676: This works for now, but I feel like it’s a bit hacky. I think we should try to better integrate this later, eg to give a meaningful error if `Field(default_factory=…)` is used in places where it isn’t allowed (we currently ignore it)

pydantic/_internal/_generate_schema.py
```diff
@@ -473,26 +464,10 @@ def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
         elif issubclass(origin, typing.Tuple):  # type: ignore[arg-type]
             # TODO: To support generic subclasses of typing.Tuple, we need to better-introspect the args to origin
```
dmontagu, line 465: Should this comment be removed? Not sure how you feel about generic subclasses of Tuple in the context of this PR

dmontagu (approved):
If this will cause custom types from v1 to break in v2, we should probably include a warning note in the migration guide and/or next release's release notes.

pydantic/_internal/_std_types_schema.py
dmontagu: It worries me a bit to not actually check that `values_source_type` is actually annotated before iterating over the `get_args` call's results. I'm thinking we could do something like this:

https://github.com/pydantic/pydantic/blob/6100242ff4fbd97e51dd625bb90c6e7683371295/pydantic/_internal/_generate_schema.py#L217

to ensure this and make it safer here in the event that `get_args` starts handling more things in the future or whatever.

(Might make sense to make that a utility function or similar anyway.)

If you disagree, I don't feel strongly.

tests/test_main.py
```diff
@@ -1589,49 +1586,25 @@ class Item(BaseModel):
...
-        x: Mapping[str, int]
-
-    d = defaultdict(int)
-    d['foo'] = '2'
-    d['bar']
-
-    m = Model(x=d)
-    assert isinstance(m.x, defaultdict)
-    assert m.x['foo'] == 2
-    assert m.x['bar'] == 0
```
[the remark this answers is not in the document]
dmontagu: Probably a good idea to document this in the migration notes, I'll add it to the list

dmontagu (approved):
Maybe worth addressing https://github.com/pydantic/pydantic/pull/5744/files#r1193010982, otherwise lgtm

PR #5754, opened 2023-05-12
Eliminate the undefined types warning

dmontagu:
I believe we discussed this some time around the end of pycon.

I think another viable alternative might be just changing the default config value, but I think if there's any doubt we should just remove this config setting and handle things the way we did in v1 where we only error for undefined types when you attempt to instantiate them.

It's pretty annoying to have to change config everywhere there is a forward ref, and _also_ need to do the model rebuild.

(I did this because I started down the path of rebuilding dataclasses and I felt like if we were going to make this change, I wanted to do so before porting the logic to dataclasses.)


dmontagu:
On Adrian's suggestion I made it auto-rebuild as well, happy to remove that though if preferred.

adriangb (approved):

dmontagu:
please review

adriangb:
Looks good to me since I had suggested auto rebuilding and that’s all you did since last review 😀

PR #5758, opened 2023-05-15
Add pydantic-settings test to CI

hramezani:
Needs [this PR in pydantic-settings](https://github.com/pydantic/pydantic-settings/pull/58) to be merged and a new release on `pydantic-settings`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -261,6 +261,28 @@ jobs:
     - name: test
       run: make test-fastapi
 
+  test-pydantic-settings:
+    name: test Pydantic Settings
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: pdm-project/setup-pdm@v3
+      with:
+        python-version: '3.10'
+        cache: true
+
+    - name: install
+      run: |
+        pdm install -G :all
+
+    - run: pdm info && pdm list
+
+    - run: ' pdm run python -c "import pydantic.version; print(pydantic.version.version_info())"'
+
+    - name: test
+      run: make test-pydantic-settings
+
   # https://github.com/marketplace/actions/alls-green
   check:  # This job does nothing and is only used for the branch protection
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,7 @@ jobs:
 
     - run: pdm info && pdm list
 
-    - run: ' pdm run python -c "import pydantic.version; print(pydantic.version.version_info())"'
+    - run: 'pdm run python -c "import pydantic.version; print(pydantic.version.version_info())"'
 
     - run: mkdir coverage
 
@@ -262,7 +262,7 @@ jobs:
       run: make test-fastapi
 
   test-pydantic-settings:
-    name: test Pydantic Settings
+    name: test pydantic settings
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -261,6 +261,28 @@ jobs:
...
+        python-version: '3.10'
+        cache: true
+
+    - name: install
+      run: |
+        pdm install -G :all
+
+    - run: pdm info && pdm list
+
+    - run: ' pdm run python -c "import pydantic.version; print(pydantic.version.version_info())"'
```
samuelcolvin, line 281:
```suggestion
    - run: 'pdm run python -c "import pydantic.version; print(pydantic.version.version_info())"'
```

.github/workflows/ci.yml
```diff
@@ -261,6 +261,28 @@ jobs:
     - name: test
       run: make test-fastapi
 
+  test-pydantic-settings:
+    name: test Pydantic Settings
```
samuelcolvin, line 265:
```suggestion
    name: test pydantic settings
```

.github/workflows/ci.yml
```diff
@@ -261,6 +261,28 @@ jobs:
     - name: test
       run: make test-fastapi
 
+  test-pydantic-settings:
```
samuelcolvin, line 264: please add to `check`.

tests/test_pydantic_settings.sh
```diff
@@ -0,0 +1,16 @@
+#! /usr/bin/env bash
+
+set -x
+set -e
+
+cd pydantic-settings
+git fetch --tags
+latest_tag_commit=$(git rev-list --tags --max-count=1)
```
samuelcolvin, line 8: why don't we test against `main`?

samuelcolvin (commented):
otherwise LGTM.

hramezani:
All done, I think the CI will be green after we merged https://github.com/pydantic/pydantic-settings/pull/58

hramezani:
please review

tests/test_pydantic_settings.sh
```diff
@@ -0,0 +1,12 @@
+#! /usr/bin/env bash
```
lig, line 1: Would it be better to put this file into `pydantic-settings` repo. It won't work without it anyway.

tests/test_pydantic_settings.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -x
+set -e
+
+cd pydantic-settings
```
lig, line 6: This will work only when invoking from the root project dir.

`cd "$(dirname $0)/../pydantic-settings"` will allow to call this script from anywhere and always get the same result.

Also, using `pushd <dir>` at the start of the script and `popd` at the end will allow returning to the original director after script has finished running.

lig (changes requested):

tests/test_pydantic_settings.sh
hramezani: I think this file shouldn't be in `pydantic-settings` repo because it belongs to this repo.
Also, we have same file for `fast-api`

lig: 👌🏻

tests/test_pydantic_settings.sh
lig: You just do `pushd "$(dirname $0)/../pydantic-settings"` at the start and `popd` at the end, no need in any other `cd` invocations. This way it will always work no matter what the working dir is and it will return it to the original one.

lig (changes requested):

hramezani:
please review

lig (approved):
Thank you! :)

Issue #5721, opened 2023-05-09
ForwarRef of dataclasses


shenek:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

I'm not able to make forward references to work together with dataclasses.

Based on https://docs.pydantic.dev/dev-v2/usage/errors/#undefined-annotation I created a code that should work, but it doesn't.

### Example Code

```Python
import typing
from pydantic.dataclasses import dataclass


class Config:
    undefined_types_warning = False


@dataclass(config=Config)
class AA:
    a1: typing.ForwardRef("A")
    a2: typing.ForwardRef("A")


@dataclass(config=Config)
class A:
    x: int
    double: typing.Optional[AA] = None


AA.__pydantic_model__.model_rebuild()


aa = AA(a1=A(5), a2=A(6))
```

### Exception
```
Traceback (most recent call last):
  File "/home/stepan/.virtualenvs/NIBBLER/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py", line 349, in _generate_schema
    obj = _typing_extra.evaluate_fwd_ref(obj, globalns=self.types_namespace)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stepan/.virtualenvs/NIBBLER/lib/python3.11/site-packages/pydantic/_internal/_typing_extra.py", line 465, in evaluate_fwd_ref
    return ref._evaluate(globalns=globalns, localns=localns, recursive_guard=frozenset())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/typing.py", line 857, in _evaluate
    eval(self.__forward_code__, globalns, localns),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'A' is not defined. Did you mean: 'AA'?

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/tmp/tree.py", line 9, in <module>
    @dataclass(config=Config)
     ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #5760, opened 2023-05-15
Support recursive dataclasses

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5721. This is also a precursor to work on generic dataclasses.

I'll open the PR now, but I think there's a variety of things to establish some consensus on before merging:

* Do we want to unify the `model-not-fully-defined` and (analogous) `dataclass-not-fully-defined` usage errors?
  * Actually, currently in this PR, I have removed `model-not-fully-defined` and replaced it with a common `class-not-fully-defined` error, but would be happy to revert this.
  * If we do revert this change, is `dataclass-not-fully-defined` an acceptable name?
    * Would we essentially fully duplicate the documentation of `model-not-fully-defined` considering it's largely similar and we have a docs section for each usage error code?
* I have renamed `BaseModel.__pydantic_model_complete__` to `BaseModel.__pydantic_complete__` and used the same property on pydantic dataclasses. Any objections?
* With BaseModels, to do an explicit rebuild you call the `BaseModel.model_rebuild` method. I have exported an analogous function called `pydantic.dataclasses.rebuild_dataclass`.
  * Is this name (and import path) acceptable?
  * Also, the function is defined in `pydantic._internal._dataclasses` and re-exported in `pydantic.dataclasses`. Is this acceptable?
  * (Actually, I think Adrian wanted to add a method called `__pydantic_dataclass_rebuild__` or similar to pydantic dataclasses — is that still the case? If so, what should the name of that method be?)

I'll be happy to make any requested changes.


pydantic/_internal/_dataclasses.py
```diff
@@ -143,3 +173,41 @@ class B(A):
...
+    *,
+    force: bool = False,
+    raise_errors: bool = True,
+    _parent_namespace_depth: int = 2,
+    _types_namespace: dict[str, Any] | None = None,
+) -> bool | None:
+    """
+    Compare with BaseModel.model_rebuild
+
+    TODO: Update this docstring
```
dmontagu, line 189: Need to update this, but wanted to get some review before doing so as I suspect this may be affected by some requested refactors..

pydantic/main.py
```diff
@@ -699,10 +699,10 @@ def model_rebuild(
 
         Returns:
             bool or None: Returns `None` if model schema is complete and no rebuilding is required, or `True` if
-                        rebuilding is successful, otherwise returns `None`.
+                        rebuilding is successful, otherwise returns `False`.
```
dmontagu, line 702: Looks like this was just a typo in the old docs

tests/test_dataclasses.py
```diff
@@ -1439,25 +1446,100 @@ class MyDataclass:
...
+    # Confirm D1 and D2 have been rebuilt
+    assert isinstance(D1.__pydantic_validator__, SchemaValidator)
+    assert isinstance(D2.__pydantic_validator__, SchemaValidator)
+
+    assert TypeAdapter(D1).dump_python(instance) == {'d2': {'d1': {'d2': {'d1': {'d2': None}}}}}
 
     with pytest.raises(ValidationError) as exc_info:
-        D1(d2=D2(d1=D1(d2=D2(d1=D2()))))
-    assert exc_info.value.errors(include_url=False) == [...]
+        D2(d1=D2())
```
dmontagu, line 1482: I removed the parts of the original expression here that don't get hit because the error gets raised while evaluating the inner components of the expression.

Note to self: Before merging, should add some checks that are similar in intent to the original expression, but with a deeply nested dict, to make sure the full recursion is working properly.

pydantic/dataclasses.py
```diff
@@ -197,8 +201,10 @@ def create_dataclass(cls: type[Any]) -> type[PydanticDataclass]:
         cls.__doc__ = original_doc
         cls.__module__ = original_cls.__module__
         cls.__qualname__ = original_cls.__qualname__
-        _pydantic_dataclasses.set_dataclass_fields(cls)
```
dmontagu, line 200: I moved this into the `complete_dataclass` method, so that the fields would also be updated when doing a rebuild.

pydantic/_internal/_dataclasses.py
```diff
@@ -143,3 +173,41 @@ class B(A):
...
+    if not force and cls.__pydantic_complete__:
+        return None
+    else:
+        if _types_namespace is not None:
+            types_namespace: dict[str, Any] | None = _types_namespace.copy()
+        else:
+            if _parent_namespace_depth > 0:
+                frame_parent_ns = _typing_extra.parent_frame_namespace(parent_depth=_parent_namespace_depth) or {}
+                # cls_parent_ns = cls.__pydantic_parent_namespace__ or {}
+                # cls.__pydantic_parent_namespace__ = {**cls_parent_ns, **frame_parent_ns}
```
dmontagu, line 200: Not sure if we want to store the original namespace on the dataclasses like we do for BaseModel.

This helps in the scenario where rebuilds are necessary when importing from a separate module and there was a ForwardRef to a type in the module in which the class was _originally_ defined.

This may also be necessary when parametrizing generic cyclically-recursive dataclasses outside the module in which they were defined.

I've left these lines present and commented as a reminder to decide about this. I'm okay either way.

docs/usage/errors.md
```diff
@@ -2,6 +2,74 @@ Below are details on common errors developers can see when working with pydantic
 with some suggestions on how to fix them.
 
 {% raw %}
+## Class not fully defined {#class-not-fully-defined}
+
+This error is raised when a type referenced in an annotation of a pydantic-validated type
+(such as a subclass of BaseModel, or a pydantic dataclass) is not defined:
```
samuelcolvin, line 8:
```suggestion
(such as a subclass of `BaseModel`, or a pydantic `dataclass`) is not defined:
```

docs/usage/errors.md
```diff
@@ -2,6 +2,74 @@ Below are details on common errors developers can see when working with pydantic
...
+class Bar(BaseModel):
+    b: 'Foo'
+
+
+Foo.model_rebuild()
+
+foo = Foo(a={'b': {'a': None}})
+```
+
+In other cases, the error message should indicate how to rebuild the model with the appropriate type definition.
```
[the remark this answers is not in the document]
samuelcolvin: `#class-not-fully-defined` LGTM.

pydantic/_internal/_dataclasses.py
```diff
@@ -65,19 +70,55 @@ def complete_dataclass(
...
+    try:
+        if get_core_schema:
+            schema = get_core_schema(cls, partial(gen_schema.generate_schema, from_dunder_get_core_schema=False))
+        else:
+            schema = gen_schema.generate_schema(cls, from_dunder_get_core_schema=False)
+    except PydanticUndefinedAnnotation as e:
+        cls_name = cls.__name__
+        if raise_errors:
+            raise
+        usage_warning_string = (
```
samuelcolvin, line 106:
```suggestion
        usage_warning_message = (
```

pydantic/_internal/_dataclasses.py
samuelcolvin: I'm pretty neutral / would defer to you on it.

My preference is "simple wins", so I think I'd prefer not to store the namespace, but don't really mind.

pydantic/dataclasses.py
```diff
@@ -13,14 +13,18 @@
...
 from .fields import Field
 
 if TYPE_CHECKING:
     from ._internal._dataclasses import PydanticDataclass
 
-__all__ = ('dataclass',)
+__all__ = (
+    'dataclass',
+    'rebuild_dataclass',
+)
```
samuelcolvin, line 27:
```suggestion
__all__ = 'dataclass', 'rebuild_dataclass'
```

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin (approved):
conflicts, otherwise LGTM.

Merge when passing.

samuelcolvin:
please update.

pydantic/_internal/_dataclasses.py
dmontagu: Okay, given there is a lot of complexity related to generic dataclasses and how exactly we should handle them, and not handling it properly is not a regression from v1 (as it wasn't handled properly there), my new plan is to not fiddle with the parent namespace, and also not support generic dataclasses in v2 (but perhaps in v2.1, or _maybe_ sooner if we happen to get it all working in time.. but the pressure is off.)

PR #5761, opened 2023-05-15
🔥 Remove `_base_class_defined` hack in favor of an empty bases check

lig:
## Change Summary

Remove `_base_class_defined` hack and check for empty `bases` as an indicator for detecting `BaseModel` in `ModelMetaclass`. These classes are no longer tied to their places in the module.

## Related issue number

None 

## Checklist

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -102,7 +102,7 @@ class ModelMetaclass(ABCMeta):
             type: the new class created by the metaclass
         """
         # Note `ModelMetaclass` refers to `BaseModel`, but is also used to *create* `BaseModel`, so we rely on the fact
-        # that `BaseModel` itself won't have any bases, but any subclass of it will, to determine that the `__new__`
+        # that `BaseModel` itself won't have any bases, but any subclass of it will, to determine whether the `__new__`
         # call we're in the middle of is for the `BaseModel` class.
         if bases:
             base_field_names, class_vars, base_private_attributes = _collect_bases_data(bases)


lig:
please review

pydantic/main.py
```diff
@@ -1322,11 +1316,13 @@ def create_model(
 
 
 def _collect_bases_data(bases: tuple[type[Any], ...]) -> tuple[set[str], set[str], dict[str, ModelPrivateAttr]]:
+    if bases == (_BaseModelBase,):
+        bases = ()
```
dmontagu, line 1320:
```suggestion
```
It seems to me this is unnecessary since the only place where this is called already checks for this conditional and bails out. (And all tests pass for me locally without these two lines.)

dmontagu:
I'll just note this may be in tension with some of the goals of https://github.com/pydantic/pydantic/pull/5740 — simplifying the MRO etc.

I like getting rid of the hacky check for subclass of BaseModel, but I don't have a strong opinion about what's best here (or whether there might be other unintended consequences, especially for performance). Happy to defer to @adriangb or @samuelcolvin

pydantic/main.py
lig: Yes. I was just a bit paranoid. And there was an old check there. So, I'm not oppose to removing this but would like to have a some kind of a check here. Maybe an `assert` like `assert(bases != (_BaseModelBase,))`?

lig:
> I'll just note this may be in tension with some of the goals of #5740 — simplifying the MRO etc.

I think, it would be even better. I'm happy to remove `_BaseModelBase` if `ModelBase` doesn't have any other bases.

pydantic/main.py
lig: BTW, this could be safely removed if `BaseModel` has no bases.

pydantic/main.py
```diff
@@ -106,106 +101,102 @@ def __new__(
         Returns:
             type: the new class created by the metaclass
         """
-        if _base_class_defined:
-            base_field_names, class_vars, base_private_attributes = _collect_bases_data(bases)
+        # Note `ModelMetaclass` refers to `BaseModel`, but is also used to *create* `BaseModel`, so we need to keep
+        # track of whether we've created the `BaseModel` class yet, and therefore whether it's safe to refer to it. If
+        # it *hasn't* been created, we assume that the `__new__` call we're in the middle of is for the `BaseModel`
+        # class, since that's defined immediately after the metaclass.
+        if bases == (_BaseModelBase,):
```
samuelcolvin, line 108: can we change this to `if bases != ...` and then revert most of these changes here?

pydantic/main.py
```diff
@@ -220,7 +211,13 @@ def __instancecheck__(self, instance: Any) -> bool:
         return hasattr(instance, '__pydantic_validator__') and super().__instancecheck__(instance)
 
 
-class BaseModel(_repr.Representation, metaclass=ModelMetaclass):
+class _BaseModelBase(_repr.Representation):
```
samuelcolvin, line 214: I assume this is here to be used by `RootModel`? Otherwise I'd rather not have it.

samuelcolvin (commented):
I'm okay with this in principle, but better if we can reduce the size of the change.

samuelcolvin (approved):
LGTM. @adriangb and @dmontagu are you happy with this?

pydantic/main.py
```diff
@@ -1337,7 +1333,7 @@ def _collect_bases_data(bases: tuple[type[Any], ...]) -> tuple[set[str], set[str
     class_vars: set[str] = set()
     private_attributes: dict[str, ModelPrivateAttr] = {}
     for base in bases:
-        if _base_class_defined and issubclass(base, BaseModel) and base != BaseModel:
+        if issubclass(base, BaseModel) and base != BaseModel:
```
dmontagu, line 1336:
```suggestion
        if base == BaseModel:
            continue
```
(then dedent the remaining bits)

pydantic/main.py
```diff
@@ -106,7 +101,11 @@ def __new__(
         Returns:
             type: the new class created by the metaclass
         """
-        if _base_class_defined:
+        # Note `ModelMetaclass` refers to `BaseModel`, but is also used to *create* `BaseModel`, so we need to keep
+        # track of whether we've created the `BaseModel` class yet, and therefore whether it's safe to refer to it. If
+        # it *hasn't* been created, we assume that the `__new__` call we're in the middle of is for the `BaseModel`
+        # class, since that's defined immediately after the metaclass.
+        if bases:
```
dmontagu, line 108: It seems to me the comment above this line should be updated a bit. At least to say something like "`BaseModel` itself won't have any bases, but any subclass of it will"

dmontagu (approved):
I'm good with the change, I just think the big comment that seems to have been moved verbatim probably deserves an update before merging.

pydantic/main.py
lig: Updated the comment

pydantic/main.py
```diff
@@ -106,7 +101,10 @@ def __new__(
         Returns:
             type: the new class created by the metaclass
         """
-        if _base_class_defined:
+        # Note `ModelMetaclass` refers to `BaseModel`, but is also used to *create* `BaseModel`, so we rely on the fact
+        # that `BaseModel` itself won't have any bases, but any subclass of it will, to determine that the `__new__`
```
dmontagu, line 105:
```suggestion
        # that `BaseModel` itself won't have any bases, but any subclass of it will, to determine whether the `__new__`
```
I think this reads a bit better since the next line checks that it's _not_ the BaseModel class, as phrased it reads to me like it's confirming we are doing BaseModel, which confused me for a sec.

PR #5765, opened 2023-05-15
✨ Add `RootModel` support

### LINKED ISSUE #5498: Add `RootModel` to pydantic?
**update by @samuelcolvin:** yes we should add this, but it needs to significantly rework `BaseModel` to use a core schema which is just the inner type. This might require some changes to pydantic-core.

---

### Initial Checks

### Description

`__root__` Has been removed, but it is easier to use.

So, should we consider providing a convenience class (named `RootModel`) to pydantic to make it easier for users?

```python
class RootModel(BaseModel):
    @root_validator(pre=True)
    def populate_root(cls, _values):
        return {"root": _values}

    @model_serializer(mode="wrap")
    def _serialize(self, _handler, _info):
        data = _handler(self)
        if _info.mode == "json":
            return data["root"]
        else:
            return data

    @classmethod
    def model_modify_json_schema(cls, json_schema):
        metadata = {"title": cls.model_config.get("title", None) or cls.__name__, "description": getdoc(cls) or None}
        metadata = {k: v for k, v in metadata.items() if v is not None}
        return {**json_schema["properties"]["root"], **metadata}
```

```python

class Sellout(BaseModel):
    a: str
    b: int


class SelloutList(RootModel):
    root: List[Sellout]


class SelloutDict(RootModel):
    root: Dict[str, Sellout]
```

### Affected Components

lig:
## Change Summary

Add `RootModel` support

## Related issue number

fix #5498 

## Checklist

Selected Reviewer: @hramezani


pydantic/_internal/_fields.py
```diff
@@ -110,6 +111,8 @@ def collect_model_fields(  # noqa: C901
             continue
         if ann_name.startswith('_'):
             continue
+        if cls.__pydantic_root_model__ and ann_name != 'root':
```
lig, line 114: I'd like to create a constant with the root field name or shall we put it in the model config maybe?

pydantic/_internal/_generate_schema.py
```diff
@@ -260,32 +258,45 @@ def model_schema(self, cls: type[BaseModel]) -> core_schema.CoreSchema:
...
+            root_field = self._common_field_schema('root', fields['root'], decorators)
+            model_schema = core_schema.model_schema(
+                cls,
+                root_field['schema'],
+                custom_init=getattr(cls, '__pydantic_custom_init__', None),
+                root_model=True,
+                post_init=getattr(cls, '__pydantic_post_init__', None),
+                config=core_config,
+                ref=model_ref,
+                metadata={**metadata, **root_field['metadata']},
```
lig, line 274: This needs more use cases to be able to test it properly. Works for now.

pydantic/main.py
```diff
@@ -217,6 +220,19 @@ def __instancecheck__(self, instance: Any) -> bool:
         """
         return hasattr(instance, '__pydantic_validator__') and super().__instancecheck__(instance)
 
+    @staticmethod
+    def _collect_bases_data(bases: tuple[type[Any], ...]) -> tuple[set[str], set[str], dict[str, ModelPrivateAttr]]:
```
lig, line 224: Moved from a module level function to avoid "is not defined" errors.

pydantic/main.py
```diff
@@ -258,6 +274,8 @@ class BaseModel(metaclass=ModelMetaclass):
         __pydantic_extra__: dict[str, Any] | None = None
         __pydantic_generic_metadata__: typing.ClassVar[_generics.PydanticGenericMetadata]
         __pydantic_parent_namespace__: typing.ClassVar[dict[str, Any] | None]
+        __pydantic_custom_init__: typing.ClassVar[bool]
+        __pydantic_post_init__: typing.ClassVar[None | Literal['model_post_init']]
```
lig, line 278: Custom init and post init markers act as cache for corresponding check results.

pydantic/main.py
```diff
@@ -271,6 +289,7 @@ class BaseModel(metaclass=ModelMetaclass):
     model_config = ConfigDict()
     __slots__ = '__dict__', '__pydantic_fields_set__', '__pydantic_extra__'
     __pydantic_model_complete__ = False
+    __pydantic_root_model__: typing.ClassVar[bool] = False
```
lig, line 292: Root model indicator. This could be transformed into enum in the future if we will need another exotic `BaseModel` subclass.

pydantic/main.py
```diff
@@ -284,6 +303,8 @@ def __init__(__pydantic_self__, **data: Any) -> None:  # type: ignore
         __tracebackhide__ = True
         __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
 
+    __init__.__pydantic_base_init__ = True  # type: ignore
```
lig, line 306: Custom init indicator. This way we do not need comparing an init method to `BaseModel.__init__`. This allows to avoid marking `RootModel.__init__` as custom init.

pydantic/main.py
```diff
@@ -620,7 +641,7 @@ def model_construct(cls: type[Model], _fields_set: set[str] | None = None, **val
             _fields_set = set(values.keys())
         _object_setattr(m, '__pydantic_fields_set__', _fields_set)
         _object_setattr(m, '__pydantic_extra__', _extra)
-        if type(m).model_post_init is not BaseModel.model_post_init:
+        if cls.__pydantic_post_init__:
```
lig, line 644: The check is cached in `cls`.

pydantic/main.py
```diff
@@ -1216,6 +1237,49 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
...
+            mode=mode,
+            include=include,
+            exclude=exclude,
+            by_alias=by_alias,
+            exclude_unset=exclude_unset,
+            exclude_defaults=exclude_defaults,
+            exclude_none=exclude_none,
+            round_trip=round_trip,
+            warnings=warnings,
+        )['root']
```
lig, line 1277: @samuelcolvin Shall we return an actual value from the pydantic-core part instead of a dict with a single attribute? It seems logical as we require a bare value for initializing the schema.

tests/test_main.py
```diff
@@ -2220,3 +2221,65 @@ class Model(BaseModel):
...
+
+
+@parametrize_root_model()
+def test_root_model_specialized(root_type, root_value, dump_value):
+    Model = RootModel[root_type]
+
+    m = Model(root_value)
+
+    with pytest.warns(UserWarning, match='but got `dict` - serialized value may not be as expected'):
+        assert m.model_dump() == dump_value
```
lig, line 2255: Model dump always complains about the root value data type stating that it should be `dict`. It looks like something isn't right on the `pydantic-core` side.

samuelcolvin: I guess you'll need a custom serialised, but again it might be fixed with `root_model=True` in the core-schema.

pydantic/main.py
lig: It has been set. I'll add a check for that in the test.

lig: https://github.com/pydantic/pydantic/pull/5765/files#diff-41d180f6f7233d175c2dfe36c45c4ea80eed4754fa7ed4eed46ccde1b2a778c7R2270

`root_model is True`

pydantic/main.py
dmontagu: Could we move this annotation to live alongside the others (like `__pydantic_post_init__: typing.ClassVar[...` above), and change this to just `__pydantic_root_model__ = False`? I guess ideally we'd do the same with `__pydantic_model_complete__` (though I thought that got renamed to `__pydantic_complete__`.. maybe not yet? or maybe it will be once merged with main)

lig: Well "others" are there under the comment "populated by the metaclass, defined here to help IDEs only". This on the other hand is defined here. First, I did exactly what you are suggesting but then I've stumbled upon that comment and noticed `__pydantic_model_complete__` here. As far as I've understood the code, this is the place for things that are defined here and "others" are there just for type checking purposes and are being effectively defined elsewhere (actually in the metaclass).

pydantic/main.py
dmontagu: @samuelcolvin I don't see anything in `pydantic-core/src/serializers` that references `root_model` so I think there probably needs to be some tweaks for it to extract that field automatically

pydantic/main.py
```diff
@@ -1216,6 +1237,49 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
...
+
+    def __iter__(self) -> TupleGenerator:
+        yield ('__root__', self.model_dump())
```
dmontagu, line 1280:
```suggestion
        yield ('root', self.model_dump())
```
Thinking this might be preferable since the field name is no longer `__root__`

pydantic/main.py
dmontagu: @lig I do think we should get that fixed in pydantic-core before merging this because I think without a fix there, we won't get proper serialization when using a RootModel as a field

pydantic/main.py
lig: This was requested by @samuelcolvin for v1 compatibility https://github.com/pydantic/pydantic/issues/5498#issuecomment-1547625102

I will add a comment here.

pydantic/main.py
lig: Totally agree

dmontagu:
Other things to add tests for:
* field validators / serializers for the root field
* subclasses of `RootModel[T]`

dmontagu:
Wow, looks like recursive stuff is actually working.

However, I ran into some issues computing the repr of these things, which I fixed with the following hack:
```diff
diff --git a/pydantic/main.py b/pydantic/main.py
index f662c736..cfd22aea 100644
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -836,7 +836,7 @@ class BaseModel(metaclass=ModelMetaclass):
             for k, v in self.__dict__.items()
             if not k.startswith('_') and (k not in self.model_fields or self.model_fields[k].repr)
         )
-        pydantic_extra = self.__pydantic_extra__
+        pydantic_extra = getattr(self, '__pydantic_extra__', None)
         if pydantic_extra is not None:
             yield from ((k, v) for k, v in pydantic_extra.items())
         yield from ((k, getattr(self, k)) for k, v in self.model_computed_fields.items() if v.repr)
```

However, I think this is not the right solution — the problem is that `__pydantic_extra__` is not getting set as it should. I think this is normally done by pydantic_core for regular models, but obviously won't be done for the non-model schema; @lig I think it might be a good idea to just set `__pydantic_extra__ = None` on the `RootModel` class to eliminate this issue. (Should also add a test that `repr(RootModel[int](1))` doesn't error or whatever).

The following code may serve as a test of both recursive RootModel _and_ RootModel repr:
```python
from pydantic import RootModel

class A(RootModel[list['B']]):
    def my_a_method(self):
        pass

class B(RootModel[dict[str, A | None]]):
    def my_b_method(self):
        pass

assert repr(A.model_validate([{}])) == 'A(root=[B(root={})])'
```

dmontagu:
Also Closes https://github.com/pydantic/pydantic/pull/5765

pydantic/main.py
lig: Fixed in https://github.com/pydantic/pydantic-core/pull/613

lig:
Recursive code doesn't work for me.

lig:
please review :)

pydantic/_internal/_fields.py
```diff
@@ -110,6 +111,8 @@ def collect_model_fields(  # noqa: C901
...
+            continue
```
dmontagu, line 115: I think we should raise an error here

pydantic/_internal/_model_construction.py
```diff
@@ -68,10 +68,8 @@ def inspect_namespace(  # noqa C901
     raw_annotations = namespace.get('__annotations__', {})
 
     if '__root__' in raw_annotations or '__root__' in namespace:
-        # TODO: Update error message with migration description and/or link to documentation
-        #   Issue: https://github.com/pydantic/pydantic/issues/5498
         raise TypeError(
-            '__root__ models are no longer supported in v2; a migration guide will be added in the near future'
+            'Use special RootModel class instead of __root__ BaseModel attribute in Pydantic V2 to define root models'
```
dmontagu, line 72:
```suggestion
            'To define root models, use `pydantic.RootModel` rather than a field called '__root__''
```

pydantic/_internal/_fields.py
lig: A test case that hits this line would definitely help. Also, I agree that the use case doesn't make sense. One can use a private attribute if desired.

pydantic/main.py
dmontagu: I think Samuel was more trying to say that it should be a dict, not that it should have that specific value for its key. But I might be wrong.. @samuelcolvin ?

tests/test_types_typeddict.py
```diff
@@ -605,31 +604,6 @@ class RecursiveGenTypedDict(TypedDict, Generic[T]):
...
-        rec: 'RecursiveGenTypedDict[T]'
-
-    IntModel = RecursiveGenTypedDictModel[int]
-
-    class RecursiveGenTypedDict(TypedDict, Generic[T]):
-        foo: Optional['RecursiveGenTypedDict[T]']
-        ls: List[T]
-
-    int_data: RecursiveGenTypedDict[int] = {'foo': {'foo': None, 'ls': [1]}, 'ls': [1]}
-    IntModel.model_rebuild()
```
dmontagu, line 632: This was removed because it was misbehavior documented in this test, and was fixed by another tweak.

pydantic/_internal/_fields.py
dmontagu: 🤷 I don't think there's much risk of confusing the root field name. Please don't put it in the model config, we should try to reduce/eliminate that as much as possible imo. We could theoretically add if people need it for some reason but I think it's okay for now to just require the use of `root` as the name.

pydantic/_internal/_fields.py
dmontagu: Test case:
```python
class MyModel(RootModel[int]):
    x: str
```
does this hit it? I'll try, but that's what I was thinking would

dmontagu: Yep I confirmed it does:
```
from pydantic import RootModel


class MyModel(RootModel[int]):
    x: str

print(MyModel.model_fields)
#> {'root': FieldInfo(annotation=int, required=True)}
```
and adding `assert False` on this line makes the above error.

I would suggest the following:
```suggestion
            raise TypeError('Cannot add additional fields to subclasses of RootModel`)
```

tests/test_parse.py
```diff
@@ -49,7 +49,7 @@ class Foo(BaseModel):
 def test_root_model_error():
     with pytest.raises(
         TypeError,
-        match='__root__ models are no longer supported in v2',
+        match='Use special RootModel class instead of __root__ BaseModel attribute',
```
dmontagu, line 52:
```suggestion
        match="To define root models, use `pydantic.RootModel` rather than a field called '__root__'",
```

pydantic/_internal/_fields.py
lig: Done

pydantic/main.py
dmontagu: @lig looking back at this, I think that when Samuel says `dict` he means `model.dict()` not `dict(model)`. I think we shouldn't override the BaseModel implementation of `__iter__` here, but _should_ override `def dict()`

dmontagu: Suggestion:
```python
@deprecated('The `dict` method is deprecated; use `model_dump` instead.')
    def dict(
        self,
        *,
        include: IncEx = None,
        exclude: IncEx = None,
        by_alias: bool = False,
        exclude_unset: bool = False,
        exclude_none: bool = False,
    ) -> typing.Dict[str, Any]:  # noqa UP006
        warnings.warn('The `dict` method is deprecated; use `model_dump` instead.', DeprecationWarning)
        value = self.model_dump(
            include=include,
            exclude=exclude,
            by_alias=by_alias,
            exclude_unset=exclude_unset,
            exclude_none=exclude_none,
        )
        return {'__root__': value}
```

dmontagu: this is definitely going to have breaking incompatibilities with v1 because you won't get RootModel's serialized as `{'__root__': ...}` as fields due to the use of `self.model_dump`. It would need to be an argument to `model_dump` or something for that to work. But.. I think that's a bad idea and people should just update their usage of `RootModel`.

dmontagu: That's not to say that we shouldn't make this change though, if someone was calling `.dict()` on the root model I think this is reasonable.

dmontagu (approved):
This looks good to me, but @samuelcolvin could you take a look at this before we merge it? 

Specifically, I'd like to get your take on this thread: https://github.com/pydantic/pydantic/pull/5765/files#r1199494554

But also just overall if you have any concerns.

tests/test_main.py
```diff
@@ -2218,3 +2219,96 @@ class Model(BaseModel):
             also_protect_field: str
 
             model_config = ConfigDict(protected_namespaces=('protect_me_', 'also_protect_'))
+
+
+def parametrize_root_model():
```
samuelcolvin, line 2224: I guess these should go in a new file.

samuelcolvin (commented):
in general these look great.

pydantic/main.py
dmontagu: I discussed this with Samuel, and we agreed it makes sense to just not attempt to be backwards compatible with v1 like this because of the inconsistency that it wont' do the right thing when the rootmodel is a field.

I added a test showing how you can re-achieve this functionality by adding a mode='wrap' field serializer. I think we can add that to the documentation.

tests/test_root_model.py
```diff
@@ -0,0 +1,154 @@
...
+
+def test_v1_compatibility_serializer():
+    class MyInnerModel(BaseModel):
+        x: int
+
+    class MyRootModel(RootModel[MyInnerModel]):
+        # The following field_serializer can be added to achieve the same behavior as v1 had for .dict()
+        @field_serializer('root', mode='wrap')
+        def embed_in_dict(self, v: Any, handler: SerializerFunctionWrapHandler):
+            return {'__root__': handler(v)}
```
dmontagu, line 145: I think we'll want to add this field_serializer into the documentation of `RootModel`.

tests/test_parse.py
```diff
@@ -48,8 +48,7 @@ class Foo(BaseModel):
 
 def test_root_model_error():
     with pytest.raises(
-        TypeError,
-        match='__root__ models are no longer supported in v2',
+        TypeError, match="To define root models, use `pydantic.RootModel` rather than a field called '__root__"
```
dmontagu, line 51:
```suggestion
        TypeError, match="To define root models, use `pydantic.RootModel` rather than a field called '__root__'"
```

dmontagu:
I talked with Samuel about this morning, and we agreed that it was best to revert some of the changes related to producing a dict when calling the `dict()` method, because:
1. It won't behave consistently recursively without a lot of additional effort
2. It's a 3-line field serializer definition to get the v1 behavior.

The other changes also ensures it behaves more consistently with `BaseModel` in all the other ways.

PR #5768, opened 2023-05-16
Add protected namespaces

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/4915

Based on the discussion with @samuelcolvin , we decided to [postpone handling of extra attrs](https://github.com/pydantic/pydantic/issues/5769)

Selected Reviewer: @lig

### DIFF
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -162,7 +162,7 @@ and has a type that is not in this tuple (or otherwise recognized by pydantic), 
   see [#3994](https://github.com/pydantic/pydantic/pull/3994) for more details, added in **V1.10**
 
 **`protected_namespaces`**
-: a `list` of strings that prevent model to have field which conflict with them (default: `['model_']`).
+: a `tuple` of strings that prevent model to have field which conflict with them (default: `('model_', )`).
 see [the dedicated section](#protected-namespaces)
 
 ## Change behaviour globally
@@ -355,7 +355,7 @@ print(Model(x=[1, 2]))
 ## Protected Namespaces
 
 Protected namespaces prevent model from having field that starts with them.
-It can be a list of strings and the default value is `['model_']`.
+It can be a list of strings and the default value is `('model_', )`.
 
 ```py
 from pydantic import BaseModel
@@ -363,11 +363,11 @@ from pydantic import BaseModel
 try:
 
     class Model(BaseModel):
-        model_x: str
+        model_prefixed_field: str
 
 except NameError as e:
     print(e)
-    #> Field "model_x" has conflict with protected namespace "model_"
+    #> Field "model_prefixed_field" has conflict with protected namespace "model_"
 ```
 
 You can change it or define multiple value for it:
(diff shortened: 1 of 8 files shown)


hramezani:
please review

docs/usage/model_config.md
```diff
@@ -346,3 +350,40 @@ print(Model(x=[1, '2']))
...
+Protected namespaces prevent model from having field that starts with them.
+It can be a list of strings and the default value is `['model_']`.
+
+```py
+from pydantic import BaseModel
+
+try:
+
+    class Model(BaseModel):
+        model_x: str
```
lig, line 366: IMO, it's easier to follow the logic of examples when field and vars have meaningful names.

Something like `model_prefixed_field` would be better here as well as `protect_me_`/`also_protect_` instead of `p1_`/`p2_` below.

pydantic/_internal/_config.py
```diff
@@ -53,6 +53,7 @@ class ConfigWrapper:
     # whether to validate default values during validation, default False
     validate_default: bool
     validate_return: bool
+    protected_namespaces: list[str]
```
lig, line 56: imo, `tuple` is better here as we are not going to update those frequently

pydantic/_internal/_fields.py
```diff
@@ -97,6 +99,10 @@ def collect_model_fields(  # noqa: C901
 
     class_vars: set[str] = set()
     for ann_name, ann_type in type_hints.items():
+        # check protected namespaces
+        for p in config_wrapper.protected_namespaces:
```
lig, line 103: For me the code looks self explanatory. Other checks below don't have similar comments. I'd rather renamed `p` to `protected_namespace` for clarity and removed the comment.

tests/test_main.py
```diff
@@ -2190,3 +2190,28 @@ class Model(BaseModel):
...
+
+
+def test_protected_namespace_default():
+    with pytest.raises(NameError, match='Field "model_x" has conflict with protected namespace "model_"'):
+
+        class Model(BaseModel):
+            model_x: str
+
+
+def test_custom_protected_namespace():
```
lig, line 2202: I would also tested that providing a custom namespace clears the default setting and `model_` would be allowed.

tests/test_main.py
```diff
@@ -2190,3 +2190,28 @@ class Model(BaseModel):
...
+    with pytest.raises(NameError, match='Field "test_x" has conflict with protected namespace "test_"'):
+
+        class Model(BaseModel):
+            test_x: str
+
+            model_config = ConfigDict(protected_namespaces=['test_'])
+
+
+def test_multiple_protected_namespace():
```
lig, line 2211: I would also test that something like `p3_` is allowed while `p1_` and `p2_` are protected.

lig (changes requested):

tests/test_main.py
```diff
@@ -2193,25 +2193,28 @@ class Model(BaseModel):
...
+            model_prefixed_field: str
 
 
 def test_custom_protected_namespace():
-    with pytest.raises(NameError, match='Field "test_x" has conflict with protected namespace "test_"'):
+    with pytest.raises(NameError, match='Field "test_field" has conflict with protected namespace "test_"'):
 
         class Model(BaseModel):
-            test_x: str
+            model_prefixed_field: str
```
lig, line 2206: I think it worth adding a comment here about why this field is present here and that it should be defined before the protected one for test to work correctly.

lig (approved):
LGTM 👍🏻

docs/usage/model_config.md
```diff
@@ -346,3 +350,40 @@ print(Model(x=[1, '2']))
 print(Model(x=[1, 2]))
 #> x=[1, 2]
 ```
+
+
+## Protected Namespaces
+
+Protected namespaces prevent model from having field that starts with them.
+It can be a list of strings and the default value is `('model_', )`.
```
adriangb, line 358:
```suggestion
Pydantic prevents collisions between model attributes and `BaseModel`'s own methods by namespacing them with the prefix `model_` and preventing setting fields or extra fields that start with that prefix.
This is customizable using the `protected_namespaces` option in the model's config so that you can allow overriding `model_` or add your own protected namespaces.
```

adriangb (approved):

PR #5776, opened 2023-05-16
Use uvicorn-style strings for ImportString

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5718

Note: I modified some of the error messages to more exactly match the error you get if you attempt the import the normal way. I can revert those changes if desireable.

Selected Reviewer: @hramezani


dmontagu:
please review

pydantic/_internal/_validators.py
```diff
@@ -45,34 +45,70 @@ def sequence_validator(
...
-        module_path, class_name = dotted_path.strip(' ').rsplit('.', 1)
-    except ValueError as e:
-        raise ImportError(f'"{dotted_path}" doesn\'t look like a module path') from e
+    components = dotted_path.strip().split(':')
+    if len(components) > 2:
+        raise ImportError(f"Import strings should have at most one ':'; received {dotted_path!r}")
+
+    module_path = components[0]
+    if not module_path:
+        raise ImportError(f'Import strings should have a nonempty module name; received {dotted_path!r}')
```
hramezani, line 83: Do we have a test for this block?

pydantic/_internal/_validators.py
```diff
@@ -45,34 +45,70 @@ def sequence_validator(
...
+            return _import_string_logic(value, require_colon_for_attributes)
         except ImportError as e:
             raise PydanticCustomError('import_error', 'Invalid python path: {error}', {'error': str(e)})
     else:
         # otherwise we just return the value and let the next validator do the rest of the work
         return value
 
 
-def _import_string_logic(dotted_path: str) -> Any:
+def _import_string_logic(dotted_path: str, require_colon_for_attributes: bool) -> Any:
```
hramezani, line 59: Question: What happens if we call the function like `_import_string_logic('collections.abc:Mapping', False)`?
It's returning `<class 'collections.abc.Mapping'>`. Is it correct according to `require_colon_for_attributes=False`

hramezani:
I've left some comments. not sure if they need an update or not.

please update

pydantic/_internal/_validators.py
dmontagu: good catch, I meant to add it but probably just forgot. Should be easy enough to add to the parametrized test

dmontagu:
I noted on slack, but I want to tweak this a bit to allow `'math.cos'` and just fall back to `'math:cos'` if that module doesn't exist (while still preserving the ability to specify `my.package.app` and `my.package:app` in the event that these are both present).

I can keep it working where `math.cos` raises an error and requires you to use `math:cos`, but I tried that and it meant:
* I had to change the way the migration import thing Marcelo wrote works (to use the `:`s in the lookup strings and/or to have an escape hatch for this erroring behavior)
* I would still have to resolve some issue with testing of `pydantic-settings`, since this causes those tests to fail (it attempts to import `'math.cos'` or similar), seems like an annoying chicken-and-egg thing to get them both updated; allowing `'math.cos'` would solve that issue (or at least kick it down the road).
* I think many users will find it more intuitive to use 'math.cos' than 'math:cos', and by automatically attempting the fallback I see minimal chance for mistakes. And I've already written the handling since I set it up to _suggest_ the change if the attribute exists, but we can easily change to just quietly load it automatically.

Any objections?

pydantic/_internal/_validators.py
dmontagu: I've removed this flag and just automatically fallback to `a.b:c` if you try to use `a.b.c` and that module doesn't exist.

dmontagu:
please review

dmontagu:
@hramezani I think this is good to go now, if you don't have any issues feel free to merge.

hramezani (approved):
LGTM 👍

Issue #5773, opened 2023-05-16
`SimpleIsInstance` type annotation


dmontagu:

dmontagu:
Use cases: int (handling larger-than-`i64`), preserve subclasses, replacement for 'allow_arbitrary_types'.

samuelcolvin:
Usage:
* `SimpleIsInstance[str]`
* or `Annotated[str, SimpleIsInstance(json_function)]`

Do we want to support support `json_types` as provided by pydantic-core? maybe not.

dmontagu:
Related: https://github.com/pydantic/pydantic/issues/5631

samuelcolvin:
See https://github.com/duo-labs/py_webauthn/issues/113 for related issue with subclasses of `bytes`.

PR #5778, opened 2023-05-16
Add SimpleIsInstance type

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5773

Looking for feedback on the implementation here before doing documentation/etc.

I've set it up so it has the standard behavior for JSON and JSON schema by default, just uses `isinstance` for python. We could make this configurable but I wanted to hash out the desired configuration choices first.

---

Unfortunately, due to the way JSON gets parsed on the Rust side, it seems that we don't get perfect preservation of big-ints, which was one of the things I was hoping to address with this change:

```python
import json

from pydantic.types import SimpleIsInstance


class Model(BaseModel):
    x: SimpleIsInstance[int]


value = dict(x=10000000000000000000000000000000000000000)
print(Model.model_validate(value))
#> x=10000000000000000000000000000000000000000


json_value = json.dumps(value)
print(json_value)
#> {"x": 10000000000000000000000000000000000000000}

print(Model.model_validate_json(json_value))
#> x=9223372036854775807
```

Not sure if it makes sense to worry about this, but I at least wanted to point it out.

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -28,6 +28,7 @@ from pydantic_core import CoreSchema, PydanticCustomError, PydanticKnownError, c
 from typing_extensions import Annotated, Literal, Protocol
 
 from ._internal import _fields, _internal_dataclass, _known_annotated_metadata, _validators
+from ._internal._generics import get_origin
 from ._internal._internal_dataclass import slots_dataclass
 from ._migration import getattr_migration
 from .annotated import GetCoreSchemaHandler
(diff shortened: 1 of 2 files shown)


dmontagu:
The only reason I didn't already do that was to avoid conflict with dirty_equals. But it's probably worth it given use of both together should be pretty rare.

samuelcolvin:
`IsInstanceOf`?

dmontagu:
I think it probably makes sense to just keep it as `IsInstance` unless you prefer that name @samuelcolvin. (I don't think the conflict is worth being concerned about)

dmontagu:
@adriangb I'm interested in your opinion on this note:

> I've set it up so it has the standard behavior for JSON and JSON schema by default, just uses isinstance for python. We could make this configurable but I wanted to hash out the desired configuration choices first.

In particular, we could add arguments to the dataclass and just use "default values" when applied via `IsInstance[<type>]` (so `IsInstance[<type>]` is like `Annotated[<type>, IsInstance()]`, but you can manually do `Annotated[<type>, IsInstance(disable_json_loading=True)]` or similar). But I'm not sure what configuration knobs it makes sense to add, was hoping for insight from you.

adriangb:
> I'm potentially happy to add none and just call this done, but if you have any concerns or good ideas I think we can still augment this a bit.

It would be backwards compatible to add knobs right? If so I'd say lets not add them for now.

dmontagu:
I could rename to `InstanceOf` instead of `IsInstance`? I prefer `InstanceOf` to `IsInstanceOf`, but between the two I don't care. I'd like to merge this though when possible.

dmontagu:
please update (need to call it InstanceOf instead of IsInstance and then we'll merge)

dmontagu:
please review

adriangb (approved):

Issue #5771, opened 2023-05-16
V1 inside V2 for migration purposes


samuelcolvin:

samuelcolvin:
most recent tag inside V2, then update on future releases.

tlambert03:
thanks @samuelcolvin, much appreciated

caniko:
@tlambert03 I don't think this can solve the issue you raised in discussion. You don't have control over upstream packages when you install them, this solution is for those that are migrating.

I would just stick to v1 if I was you, it is much safer and easier for those in your situation.

tlambert03:
thanks. perhaps i don't understand what's happening here then?  (I assumed this meant there would be a v1 namespace inside of v2 ... but maybe I'm wrong)

> I would just stick to v1 if I was you, it is much safer and easier for those in your situation. 

again, @caniko, I'm afraid this misses the point.  don't want to go through this again, but it doesn't matter whether a library (not an application) "sticks to v1" or not.  as soon as the end user pip installs another package with an incompatible version of pydantic into that same environment (i.e. if it was technically an "unsolvable" environment but pip installs it anyway), then the package using v1 is broken unless there is a "safe" v1 namespace that they can use.

@samuelcolvin, is that the goal here?

samuelcolvin:
you have it right, you'll be able to import pydantic v1 from `pydantic.v1`.

tlambert03:
Excellent.  Then this is exactly what I was asking for.  Much appreciated

caniko:
You want to ensure that the enduser doesn't accidentally use the wrong version by installing the latest version of Pydantic. That is enabled by this.

This is not what I am referring to.

I am talking about packages that use v1 now that won't be migrated to v2 because they work fine for their intended use (paraphrasing you here). These packages also don't have upper bound semver. These packages would break in v2. That is what I am saying.

caniko:
You still won't be able to bring this upstream. So the initial issue is unresolved.

paddymul:
Are there any guides for library writers thinking of using pydantic?  I want to use it to do input validation, but don't expect any deep dependencies on PyDantic functionality.  Has there been a backport of the `v1` namespace?

would something like this be appropriate
```python

import pydantic
if pydantic.version.split('.')[0] < 2:
    pydanticv1 = pydantic
else:
    pydanticv1 = pydantic.v1
```

tlambert03:
[...]

PR #5780, opened 2023-05-17
Add v1 namespace inside pydantic

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/5771

Selected Reviewer: @lig

### DIFF
--- a/update_v1.sh
+++ b/update_v1.sh
@@ -3,9 +3,6 @@
 set -x
 set -e
 
-# Remove current V1
-rm -rf pydantic/v1
-
 echo "cloning pydantic V1"
 git clone -b 1.10.X-fixes https://github.com/pydantic/pydantic.git pydantic-v1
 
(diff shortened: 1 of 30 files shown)


hramezani:
please review

update_v1.sh
```diff
@@ -0,0 +1,25 @@
+#! /usr/bin/env bash
+
+set -x
+set -e
+
+# Remove current V1
+rm -rf pydantic/v1
```
lig, line 7: IMO, it would be better to move this right before the `cp` below. This way if something goes wrong with the clone and/or checkout we will still have the files.

It doesn't matter that much as we still can do `git reset --hard` anyway. So, up to you.

lig (approved):

update_v1.sh
hramezani: Done!

Issue #4178, opened 2022-06-21
Pickling error with `create_model`


sunnyosun:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
 pydantic version: 1.9.0
            pydantic compiled: True
                 install path: /opt/miniconda3/envs/py39/lib/python3.9/site-packages/pydantic
               python version: 3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:52)  [Clang 13.0.1 ]
                     platform: macOS-12.4-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']
```

Minimum Reproducible Example

```py
from pydantic import create_model, BaseModel
import pickle

class Model1(BaseModel):
    x: str = 'a'
    y: str = 'b'

Model2 = create_model('Model1', **{'x': 'a', 'y': 'b'})

print(pickle.dumps(Model1))
print(pickle.dumps(Model2))

```

Pickling error with Model2 but not Model1
<img width="648" alt="Screen Shot 2022-06-22 at 01 44 50" src="https://user-images.githubusercontent.com/38218185/174864913-fa447da7-ce40-4b03-8d73-3ca0ad3ee010.png">

raceychan:
I know this is irrelevant, but may I ask what IDE do you use for jupyter? that tracebook looks great.

sunnyosun:
I use the integrated Jupiter notebook in vs-code! :)

~Sunny

PR #5781, opened 2023-05-17
Add note about dynamically created model pickling

hramezani:
Closes https://github.com/pydantic/pydantic/issues/4178

Selected Reviewer: @tpdorsey

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -918,6 +918,12 @@ except ValidationError as e:
     """
 ```
 
+!!! note
+    To pickle a dynamically created model:
+
+    - the model must be defined globally
+    - it must provide `__module__`
+
 ## Using Pydantic without creating a `BaseModel`
 
 You may have types that are not `BaseModel`s that you want to validate data against.


hramezani:
please review

docs/usage/models.md
```diff
@@ -914,6 +914,9 @@ except ValidationError as e:
       Assertion failed, must be alphanumeric [type=assertion_error, input_value='scolvi%n', input_type=str]
     """
 ```
+!!! note
+    Pickle will work for dynamically created model only if it was defined globally
+    with its class name and module where it's defined was specified.
```
tpdorsey, line 919: I may be missing something about this. Would the following capture the same meaning?

"Pickle will work for a dynamically created model only if the model was defined globally, specifying both its class name and the module in which the class was defined."

hramezani: @tpdorsey I just copied the description from https://github.com/pydantic/pydantic/blob/793b484da1eb8141ce5ed0d4ec9413fb706950e7/tests/test_create_model.py#L46-L56


It may help you to understand it better.

I think the dynamic model has to have these two conditions to be pickable:

- Has to provide `__module__`
- Has to be defined globally. So, defining a dynamic model in function doesn't work

docs/usage/models.md
```diff
@@ -914,6 +914,9 @@ except ValidationError as e:
       Assertion failed, must be alphanumeric [type=assertion_error, input_value='scolvi%n', input_type=str]
     """
 ```
+!!! note
+    Pickle will work for dynamically created model only if it was defined globally
```
tpdorsey, line 918:
```suggestion
    To pickle a dynamically created model: 
    
    - the model must be defined globally
```

docs/usage/models.md
tpdorsey, line 919:
```suggestion
    - it must provide `__module__`
```

tpdorsey: This works, I'll edit the above to use this language. Thx

tpdorsey (approved):
Minor suggested edit

Issue #5096, opened 2023-02-22
JSON schema: Callable


dmontagu:
(This was left over from https://github.com/pydantic/pydantic/pull/5029)

We should resolve the xfailure of the following test:

https://github.com/pydantic/pydantic/blob/73373c3e08fe5fe23e4b05f549ea34e0da6a16b7/tests/test_json_schema.py#L1145-L1173

I don't have an opinion about what we _should_ do; I would be happy to keep backwards compatibility or break it, I would guess this is a pretty rare edge case.

I could also see it making sense to do things differently depending on whether there was a default value.

dmontagu:
I've got something in the works that I think will solve this and other issues — an annotation you can add that just conveniently lets you override the "base" JSON schema to use for the field. That way, you don't have to subclass `GenerateJsonSchema` just to say "ignore this field".

The main thing I'm still trying to figure out is a clean API for omitting the field/etc. from the JSON schema, which I think we should also be able to do.

adriangb:
Can we make `__get_pydantic_json_schema__` return `None | JsonSchemaValue`?

dmontagu:
The main nuisance to that is the `__modify_schema__` function used to be supposed to return None (and modify the value in place). So I think it will be a common mistake to keep the logic doing that, which might more-silently misbehave than it currently would (since it will warn you now, and do what you intended).

I know we could have the behavior depend on the name of the function, but I think some people will still do the migration wrong at first. I could be convinced to make the change anyway, but another option is to introduce a new exception that should be raised, which I think wouldn't have any potential for conflict with existing APIs.

I have written some code that does that, but I could be convinced to just have `None` means omit.

dmontagu:
@adriangb I added a comment on #5788 which I just opened about returning `None | JsonSchemaValue` — it's also easy to accidentally not return anything and then get slightly different behavior. Doesn't mean we can't make the change, but I view it as another potential minor source of confusion.

I think that omitting fields is a pretty rare thing to want to do, whereas forgetting to return something from your function (while quickly developing/prototyping) is probably more common, so it might be unfortunate if we make that behave differently. That said, I don't know what we do now and maybe it's not any better. (Of course we could fix that, but 🤷 adding an exception just for omitting fields also feels .. meh..)

adriangb:
OK yeah that makes sense, I’m convinced. Can we re-use the existing Omit
error?

dmontagu:
[...]

Issue #4079, opened 2022-05-15
model.schema() fails on arbitrary types


ringohoffman:
### Checks

* [X] I added a descriptive title to this issue
* [X] I have searched (google, github) for similar issues and couldn't find anything
* [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```bash
$ python -c "import pydantic.utils; print(pydantic.utils.version_info())"
             pydantic version: 1.9.0
            pydantic compiled: True
                 install path: /home/matthew/.conda/envs/glass/lib/python3.9/site-packages/pydantic
               python version: 3.9.12 (main, Apr  5 2022, 06:56:58)  [GCC 7.5.0]
                     platform: Linux-5.4.0-107-generic-x86_64-with-glibc2.31
     optional deps. installed: ['typing-extensions']
```

I want to include arbitrary types in my `BaseModel` subclass, so I use `arbitrary_types_allowed=True`, but this ends up breaking `Model.schema()`:

```python
import pydantic

class MyClass: 
    x: int
    y: int

class Model(pydantic.BaseModel):
    v: MyClass

    class Config:
        arbitrary_types_allowed = True

Model.schema()
```

```
ValueError: Value not declarable with JSON Schema, field: name='v' type=MyClass required=True
```

What about a pydantic type, like `PyObject`?

```python
import pydantic

class Model(pydantic.BaseModel):
    v: pydantic.PyObject

    class Config:
        arbitrary_types_allowed = True

Model.schema()
```

```
ValueError: Value not declarable with JSON Schema, field: name='v' type=PyObject required=True
```

There's no way `typing.Callable` should work then, right?

```python
import pydantic
import typing

class Model(pydantic.BaseModel):
    v: typing.Callable

    class Config:
        arbitrary_types_allowed = True

Model.schema()
```

```
UserWarning: Callable v was excluded from schema since JSON schema has no equivalent type.
  Model.schema()

{'title': 'Model', 'type': 'object', 'properties': {}}
```

[...]

PR #5788, opened 2023-05-17
Add a convenience type to override JSON schema

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5096
Closes https://github.com/pydantic/pydantic/issues/4079

Introduces an annotation `WithJsonSchema` that you can use to override the JSON schema for a type.

Also makes it so that if you raise `pydantic_core.PydanticOmit` while generating the JSON schema for a _field_ of any type that has fields (i.e., model, dataclass, typeddict), that field is omitted from the generated JSON schema.

The tests added to this PR demonstrate how to use this functionality to address:
* Callable, related to https://github.com/pydantic/pydantic/issues/5096
* Arbitrary types, related to https://github.com/pydantic/pydantic/issues/4079

### DIFF
--- a/pydantic/errors.py
+++ b/pydantic/errors.py
@@ -150,14 +150,4 @@ class PydanticInvalidForJsonSchema(PydanticUserError):
         super().__init__(message, code='invalid-for-json-schema')
 
 
-class OmitJsonSchemaField(Exception):
-    """
-    An exception raised during generation of the JSON schema that is intended
-    specifically to indicate to the schema generator that the field should be excluded from
-    the JSON schema.
-
-    This exception should generally not be propagated outside of method calls on the GenerateJsonSchema class.
-    """
-
-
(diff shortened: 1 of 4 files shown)


dmontagu:
@adriangb has suggested using `pydantic_core.PydanticOmit` rather than a new exception class. As long as we document it somewhere (which we need to do anyway), I think that's fine, I'll make the change.

dmontagu:
Also Adrian I know I suggested we should re-export `pydantic_core.PydanticOmit` but I take it back — considering this will be pretty niche anyway I think it's fine to just leave as an import from `pydantic_core` unless we get feedback to the contrary. (Happy to change that though if you or Samuel prefer.)

dmontagu:
please review

adriangb (approved):
Looks good to me!

PR #5789, opened 2023-05-17
Use Unpack on an empty typeddict to disallow **kwargs in pydantic.Field

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5099

Uses some new functionality from `typing_extensions` to make sure that the `**kwargs` is treated as disallowed even though there isn't an overload missing the `**kwargs`.

This works correctly with pyright, but unfortunately I don't think it works properly with mypy. I'm not sure if this is a mypy bug, perhaps worth reporting.

-----

An alternative that _would_ work would be to use a single-item overload and a `# type: ignore`. That would be an easy change to make but I remember getting pushback the last time I tried using a single-item overload to lie to mypy about the function signature. 

I don't think `if TYPE_CHECKING:` is a great solution here because we do want the body of the function to get type checked, just want it to be treated as having a different signature.


dmontagu:
So far the best-behaving alternative in terms of type-checker error display that I've found is:
```diff
diff --git a/pydantic/fields.py b/pydantic/fields.py
index 3a76960c..feb8c92c 100644
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -533,11 +533,8 @@ class AliasChoices:
         return aliases
 
 
-class _EmptyKwargs(typing_extensions.TypedDict):
-    pass
 
-
-def Field(  # noqa C901
+def _Field(  # noqa C901
     default: Any = Undefined,
     *,
     default_factory: typing.Callable[[], Any] | None = None,
@@ -712,6 +709,48 @@ def Field(  # noqa C901
     )
 
 
+if typing.TYPE_CHECKING:
+    def Field(  # noqa C901
+        default: Any = Undefined,
+        *,
+        default_factory: typing.Callable[[], Any] | None = None,
+        alias: str | None = None,
+        alias_priority: int | None = None,
+        validation_alias: str | AliasPath | AliasChoices | None = None,
+        serialization_alias: str | None = None,
+        title: str | None = None,
+        description: str | None = None,
+        examples: list[Any] | None = None,
+        exclude: typing.AbstractSet[int | str] | typing.Mapping[int | str, Any] | Any = None,
+        include: typing.AbstractSet[int | str] | typing.Mapping[int | str, Any] | Any = None,
+        gt: float | None = None,
+        ge: float | None = None,
+        lt: float | None = None,
+        le: float | None = None,
+        multiple_of: float | None = None,
+        allow_inf_nan: bool | None = None,
+        max_digits: int | None = None,
+        decimal_places: int | None = None,
+        min_items: int | None = None,
+        max_items: int | None = None,
+        min_length: int | None = None,
+        max_length: int | None = None,
+        frozen: bool = False,
+        pattern: str | None = None,
+        discriminator: str | None = None,
+        repr: bool = True,
+        strict: bool | None = None,
+        json_schema_extra: dict[str, Any] | None = None,
+        validate_default: bool | None = None,
+        const: bool | None = None,
+        unique_items: bool | None = None,
+        allow_mutation: bool = True,
+        regex: str | None = None,
+    ) -> Any:
+        ...
+else:
+    Field = _Field
+
 class ModelPrivateAttr(_repr.Representation):
     """A descriptor for private attributes in class models.
```
but this is also kind of gross

Issue #5784, opened 2023-05-17
Incorrect result when dumping nested model when allowing extra.


Jerry-Ma:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

Hi,
It appears that `model_dump` of nested model gives incorrect result when `extra="allow"`. Here is a minimum working example that demonstrates the issue:

### Example Code

```Python
from pydantic import BaseModel, ConfigDict, Field
from pydantic.version import version_info


class Inner(BaseModel):
    a: str = "a"


class Outer(BaseModel):
    # this cause the inner model incorrectly dumpped:
    model_config = ConfigDict(extra="allow")
    inner: Inner = Field(default_factory=Inner)


def test_dump():
    m = Outer.model_validate({})

    print(m.model_dump())


print(version_info())
test_dump()
```

Output:

```
             pydantic version: 2.0a4
        pydantic-core version: 0.31.0 release build profile
                 install path: /home/ma/.pyenv/versions/3.10.6_sys/envs/tolteca_v2/lib/python3.10/site-packages/pydantic
               python version: 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]
                     platform: Linux-5.15.85-penguins-rule-x86_64-with-glibc2.35
     optional deps. installed: ['typing-extensions']
/home/ma/.pyenv/versions/tolteca_v2/lib/python3.10/site-packages/pydantic/main.py:525: UserWarning: Pydantic serializer warnings:
  Expected `general-fields` but got `tuple` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_python(
{'inner': ({'a': 'a'}, None)}
```

There is, for some reason, an extra tuple around the dumped inner model. The expected output should be `{'inner': {'a': 'a'}}`.

Let me know if you need more details. Thanks a lot!

Jerry-Ma:
Update:
It appears that pydantic v2.0a3 and pydantic_core v0.30.0 does not have this issue.

dmontagu:
@Jerry-Ma sorry — when you said 2.0a3 does not have this issue I thought you meant it was fixed in 2.0a3, I missed that it was reported in 2.0a4. We'll work on fixing it!

Jerry-Ma:
Ah, I see. Sorry I didn't make it very clear. Much appreciated for all the work!

dmontagu:
[...]

PR #5793, opened 2023-05-18
Do not inherit config in model fields that are themselves of type model

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5784

This is the pydantic-side companion to https://github.com/pydantic/pydantic-core/pull/612. Currently, pydantic doesn't work with the `main` branch of `pydantic-core`, but if I rebase the change from that PR onto the `v0.31.0` commit (which pydantic is currently using), the xfailing tests I've added to this PR start passing.

The PR body from https://github.com/pydantic/pydantic-core/pull/612 goes into great detail about the problems this solves, but I'll also add a bit of detail here demonstrating that this actually reverts to the v1 behavior. (The following code is basically the same as one of the tests I added in here, just has been modified to be simultaneously compatible with v1 and v2, with deprecation warnings in v2):

```python
from typing import List

class Inner(BaseModel):
    a: str


class Outer(BaseModel):
    # this cause the inner model incorrectly dumpped:
    class Config:
        str_to_lower = True
        anystr_lower = True

    x: List[str]  # should be converted to lower
    inner: Inner  # should not have fields converted to lower



m = Outer.parse_obj(dict(x=['Abc'], inner=dict(a='Def')))
print(m.dict())
# Current v2:
#> {'x': ['abc'], 'inner': {'a': 'def'}}

# Preferred in v2 (and current in v1):
#> {'x': ['abc'], 'inner': {'a': 'Def'}}
```

### DIFF
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -109,17 +109,18 @@ class ConfigWrapper:
                 except KeyError:
                     raise AttributeError(f'Config has no attribute {name!r}') from None
 
-    def core_config(self, obj: Any = None) -> core_schema.CoreConfig:
+    def core_config(self, obj: Any) -> core_schema.CoreConfig:
         """
         Create a pydantic-core config, `obj` is just used to populate `title` if not set in config.
 
+        Pass `obj=None` if you do not want to attempt to infer the `title`.
+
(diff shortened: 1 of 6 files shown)


adriangb (approved):

adriangb:
Discussed in the pydantic-core PR

PR #5800, opened 2023-05-19
Improve typechecking, docs, and minor bugs with Field/FieldInfo

dmontagu:
Selected Reviewer: @lig


tests/test_aliases.py
```diff
@@ -424,14 +424,13 @@ class Model(BaseModel):
...
         (AliasChoices('a', AliasPath('b', 1)), [['a'], ['b', 1]]),
-        (AliasChoices(), None),
     ],
 )
 def test_validation_alias_path(input, expected):
     class Model(BaseModel):
         x: str = Field(validation_alias=input)
 
-    assert Model.model_fields['x'].validation_alias == expected
+    assert Model.model_fields['x'].validation_alias == input
```
hramezani, line 433: if we don't need `expected`, we can remove it from `parametrize` parameters

dmontagu:
please review

dmontagu:
This adds a change that breaks pydantic-settings. I have opened a PR that will fix that in https://github.com/pydantic/pydantic-settings/pull/59. I will disable CI for pydantic settings here for now then we can reenable once everything is back in sync.

dmontagu:
Not sure why cloudflare pages is failing but that's the only problem now. (Because I disabled the run of pydantic-settings. We could re-enable once it is updated to use this version of pydantic.)

dmontagu:
Going to try reopening to see if that resolves the cloudflare build issue..

dmontagu:
Hmm nope couldn't get it to work. @samuelcolvin I guess I need some info about the cloudflare thing

hramezani:
@dmontagu Why do we need to add`pydantic-settings` as submodule here?

hramezani (approved):

PR #5801, opened 2023-05-19
Rework `__prepare_pydantic_annotations__` to make it easier to wrap "known" types

adriangb:
Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -557,7 +557,6 @@ class GenerateSchema:
             return schema
 
         source_type, annotations = field_info.annotation, field_info.metadata
-        source_type, annotations = self._prepare_annotations(source_type, annotations)
         schema = self._apply_annotations(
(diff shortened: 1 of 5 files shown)


adriangb:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -1126,9 +1122,16 @@ def _apply_annotations(
...
+                remaining = annotations[idx + 1 :]
+                new_source_type, remaining = prepare(source_type, tuple(remaining))
                 annotations = previous + remaining
+                if new_source_type is not source_type:
+                    return self._apply_annotations(
+                        get_inner_schema,
+                        new_source_type,
+                        annotations,
+                        check_prepare_on_source=check_prepare_on_source,
+                    )
```
adriangb, line 1134: This is the only part that's a bit wonky about this PR: if something replaces the source type we "start over" going from right to left again. My main fear is that it might be a bit difficult to explain, but this is a _very_ advanced use case so maybe that's okay.

dmontagu (approved):

PR #5806, opened 2023-05-19
Remove various TODO comments

dmontagu:
I think they have all been addressed or are reasonable to remove. I think I've done about 30-50% of them, want to do more, but figure smaller might be easier to review..

Selected Reviewer: @lig

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -124,7 +124,8 @@ def filter_field_decorator_info_by_field(
 def apply_each_item_validators(
     schema: core_schema.CoreSchema, each_item_validators: list[Decorator[ValidatorDecoratorInfo]]
 ) -> core_schema.CoreSchema:
-    # TODO: remove this V1 compatibility shim once it's deprecated
+    # This V1 compatibility shim should eventually be removed
+
     # push down any `each_item=True` validators
     # note that this won't work for any Annotated types that get wrapped by a function validator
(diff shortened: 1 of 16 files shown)


dmontagu:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -425,10 +424,6 @@ def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
         if _typing_extra.is_dataclass(obj):
             return self._dataclass_schema(obj, None)
 
-        # TODO: _std_types_schema iterates over the __mro__ looking for an expected schema.
```
adriangb, line 428: I still need to remove this for a couple of types (just network types I think)

pydantic/_internal/_generate_schema.py
```diff
@@ -462,7 +457,6 @@ def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
         elif issubclass(origin, Annotated):  # type: ignore[arg-type]
             return self._annotated_schema(obj)
         elif issubclass(origin, typing.Tuple):  # type: ignore[arg-type]
-            # TODO: To support generic subclasses of typing.Tuple, we need to better-introspect the args to origin
```
adriangb, line 465: I wonder if we can just remove the tuple handling from here in general.

adriangb (approved):

PR #5819, opened 2023-05-22
Add a convenience type for transforming a core schema via annotation

dmontagu:
The added tests show how this can be used to transform both first- and third-party defined custom types that are basically minor extensions of non-custom types, and make them compatible with `FieldInfo`- and `annotated_types`-style constraints.

Note: these tests were specifically inspired by:
* https://github.com/pydantic/pydantic/discussions/5798 in the case of first-party types
* https://github.com/pydantic/pydantic/issues/5053 in the case of third-party types


samuelcolvin (approved):
:shrug: okay, this is risking getting beyond me.

But I don't think that's because of this.

Issue #5821, opened 2023-05-23
Cannot use Literal when I use typing-extension==4.6.0


choiking10:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

I update my typing extention 4.6.0, I face this error

```
.venv/lib/python3.9/site-packages/ddtrace/internal/module.py:216: in _exec_module
    self.loader.exec_module(module)
src/entry.py:3: in <module>
    from apis import router as api_router
.venv/lib/python3.9/site-packages/ddtrace/internal/module.py:216: in _exec_module
    self.loader.exec_module(module)
src/apis/__init__.py:5: in <module>
    from .recsys import router as recsys_router
.venv/lib/python3.9/site-packages/ddtrace/internal/module.py:216: in _exec_module
    self.loader.exec_module(module)
src/apis/recsys/__init__.py:3: in <module>
    from .course.endpoints import router as course_router
.venv/lib/python3.9/site-packages/ddtrace/internal/module.py:216: in _exec_module
    self.loader.exec_module(module)
src/apis/recsys/course/__init__.py:1: in <module>
    from . import cell
.venv/lib/python3.9/site-packages/ddtrace/internal/module.py:216: in _exec_module
    self.loader.exec_module(module)
src/apis/recsys/course/cell/__init__.py:1: in <module>
    from . import container
.venv/lib/python3.9/site-packages/ddtrace/internal/module.py:216: in _exec_module
    self.loader.exec_module(module)
src/apis/recsys/course/cell/container.py:4: in <module>
    from core.models.feature_store.cell_dtos import CourseCellDTO
.venv/lib/python3.9/site-packages/ddtrace/internal/module.py:216: in _exec_module
    self.loader.exec_module(module)
src/core/models/feature_store/cell_dtos.py:14: in <module>
    class VocaCellDTO(BaseModel):
pydantic/main.py:197: in pydantic.main.ModelMetaclass.__new__
    ???
pydantic/fields.py:506: in pydantic.fields.ModelField.infer
    ???
[...]

PR #5826, opened 2023-05-23
Fix Literal bug with typing-extension==4.6.0

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/5821


pydantic/typing.py
```diff
@@ -91,6 +94,11 @@ def get_all_type_hints(obj: Any, globalns: Any = None, localns: Any = None) -> A
 AnnotatedTypeNames = {'AnnotatedMeta', '_AnnotatedAlias'}
 
 
+LITERAL_TYPES: Set[Union[TypingSpecialForm, TypingExtensionSpecialForm]] = {Literal}
```
AlexWaygood, line 97: My advice would be to go for a slightly less precise annotation here -- we definitely regard `_SpecialForm` as an undocumented implementation detail, both at CPython and at `typing_extensions` :)

```suggestion
LITERAL_TYPES: Set[object] = {Literal}
```

hramezani: Agree, Changed!

tests/test_typing.py
```diff
@@ -124,3 +124,18 @@ def test_convert_generics_pep604():
     assert (
         convert_generics(dict['Hero', list['Team']] | int) == dict[ForwardRef('Hero'), list[ForwardRef('Team')]] | int
     )
+
+
+def test_is_literal__with_typing_extension_literal():
```
AlexWaygood, line 129:
```suggestion
def test_is_literal_with_typing_extension_literal():
```

AlexWaygood (commented):
I'm not familiar with `pydantic` internals, but this looks like a reasonable fix to me!

smartYSC:
Will this fix be delivered as Pydantic 1.10.8 immediately? If yes, I will wait and not implement any workarounds.

AlexWaygood:
Honestly, I feel like this situation illustrates the inadequacies of SemVer more than anything else

samuelcolvin:
> than

agreed.

> I wonder if it might not be a good idea to introduce an upper bound (`<5`) on the `typing-extensions` package as well.

No. There are many articles out there about why including upper bounds in package dependencies is almost always a bad idea.

> Will this fix be delivered as Pydantic 1.10.8 immediately? If yes, I will wait and not implement any workarounds.

We'll do our best to get it out asap.

samuelcolvin:
No problem @AlexWaygood, these things happen. It's the responsibility of app developers to pin dependencies.

CF build failures is a random download issue, see [my angry tweet](https://twitter.com/samuel_colvin/status/1660973330793472000).

pydantic/typing.py
```diff
@@ -354,10 +360,7 @@ def is_none_type(type_: Any) -> bool:
 else:
 
     def is_none_type(type_: Any) -> bool:
-        for none_type in NONE_TYPES:
-            if type_ is none_type:
-                return True
-        return False
+        return type_ in NONE_TYPES
```
samuelcolvin, line 363: This looks weird enough that we might have done it for some strange reason, can we confirm this was just a code smell?

hramezani: The existing approach(checking with `is`) does not work in  `Python 3.9` and `typing-extension==4.6.0`. That's why I've changed it.

Note: the new approach is also Ok with `typing-extension==4.5.0`.

samuelcolvin:
> This should be fixed now in the just-released 1.10.8, which should be available through PyPI shortly.

Done

PR #5844, opened 2023-05-24
⬆ Fix some tests with Cython 3.0 beta

lig:


samuelcolvin:
lets not both, let's stick to <3.

PR #5849, opened 2023-05-24
Ignore pytest warnings on 3.12

JelleZijlstra:
## Change Summary

I decided to try to run the pydantic test suite against current Python 3.12 to look for issues in typing. I had to ignore these warnings (already tracked in pytest-dev/pytest#10977) first to get the tests to run at all.

Now a lot of tests fail for me locally with errors like

```
E               pytest.PytestUnraisableExceptionWarning: Exception ignored in: None
E               
E               Traceback (most recent call last):
E                 File "/Users/jelle/py/pydantic/pydantic/_internal/_core_utils.py", line 247, in walk
E                   return f(schema.copy(), self._walk)
E                            ^^^^^^^^^^^^^
E               ResourceWarning: Object of type pydantic_core._pydantic_core.SchemaValidator is not untracked before destruction
```

That might be a bug in `pydantic_core` but I haven't investigated further. Still, this PR at least enables running the test suite.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,6 +145,8 @@ xfail_strict = true
 filterwarnings = [
     'error',
     'ignore:path is deprecated.*:DeprecationWarning:',
+    # Work around https://github.com/pytest-dev/pytest/issues/10977 for Python 3.12
+    'ignore:(ast\.Str|ast\.NameConstant|ast\.Num|Attribute s) is deprecated and will be removed.*:DeprecationWarning:',
 ]
 
 # configuring https://github.com/pydantic/hooky


samuelcolvin:
Thanks so much, see also https://github.com/pydantic/pydantic-core/pull/629 which I just created.

samuelcolvin:
Lets merge this at least. Probably need to get pydantic-core passing before returning to this issue.

JelleZijlstra:
Thanks, sounds good! I realized the ResourceWarning I posted above might just be because I was using a debug build; I don't think there's any change around that area in 3.12. Still would be good to fix, of course.

samuelcolvin:
very surprisingly tests are passing with 3.12 on pydantic-core, I'm just running a full build to see what happens with that.

dmontagu:
Is this a thing where we need to do a better job of handling GC stuff with PyO3 for certain structs? I know that was causing memory leaks before for certain validators, but I think we now have tests showing that memory usage isn't growing out of control. Do we _know_ whether this warning is spurious or not at this stage? (Maybe of interest to @davidhewitt as well?)

JelleZijlstra:
My guess would be that the warning is real and indicates a bug in PyO3, but it's not new in 3.12. The warning is raised only if CPython is compiled in debug mode: https://github.com/python/cpython/blob/22618a86f4c42aaecf5057055b41552346ddd3aa/Modules/gcmodule.c#L2350

messense:
That `ResourceWarning` was reported to PyO3 in the past: https://github.com/PyO3/pyo3/issues/3064

Issue #5271, opened 2023-03-24
V2 dataclass improvements


samuelcolvin:
* make sure valididators are working on vanilla dataclasses as fields
* make sure `kw_only` is correctly inherited
* make sure serialization schema is right

Fix

```py
from datetime import datetime

from pydantic import field_validator
from pydantic.dataclasses import dataclass


@dataclass
class DemoDataclass:
    ts: datetime = None

    @field_validator('ts', mode='before')
    def set_ts_now(cls, v):
        return v or datetime.now()


print(DemoDataclass())
print(DemoDataclass(ts='2017-11-08T14:00'))
```

Assigned myself, but happy for someone else to work on this.

adriangb:
> make sure valididators are working on vanilla dataclasses as fields

They are, not sure what you meant by this.

> make sure kw_only is correctly inherited

I believe this works, here's a quick test. Is there something specific that wasn't working?

```python
from pydantic.dataclasses import dataclass


@dataclass(kw_only=True)
class Parent:
    x: int


@dataclass
class Child(Parent):
    y: int


print(Child(x=1, y=1))
print(Child(1, x=1))
print(Child(1, 1))
"""
y
  Field required [type=missing, input_value=ArgsKwargs((1,), {'x': 1}), input_type=ArgsKwargs]
"""
```

> make sure the serialization schema is right

This is quite vague, dataclass serialization works in general.

---

@samuelcolvin since it seems like all of this works I'd like to close this as completed, unless you have any more specifics on the issues listed.

adriangb:
Does that work in V1? I don’t think it should work or we should support it. The type hint there is also wrong (the default is not a valid value).

dmontagu:
if you look at the fix, it seems pretty trivial to support. I think this is a legitimate use case because it lets you use vanilla dataclasses in your python code but still get validators applied when validating with pydantic

adriangb:
I saw the fix, it does look simple but I’d like to see it pass all of the inheritance tests and whatnot before we declare it as fixed. We should also at least document for the record what the V1 behavior is.

dmontagu:
Also, one interesting thing I found with regard to kw_only:

```python
from pydantic.dataclasses import dataclass


@dataclass(kw_only=True)
class Parent:
    x: int


@dataclass
class Child(Parent):
    y: int


print(Child(1, x=1))
```
[...]

PR #5857, opened 2023-05-24
Handle validators on vanilla dataclass fields

dmontagu:
Partially addresses #5271.

Still need to add a test or two, but opening the PR now so work doesn't end up duplicated.

See https://github.com/pydantic/pydantic/issues/5271#issuecomment-1561832416 for at least one test we can add.

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -241,18 +241,15 @@ class DecoratorInfos:
         res = DecoratorInfos()
         for base in model_dc.__bases__[::-1]:
             existing = cast(Union[DecoratorInfos, None], getattr(base, '__pydantic_decorators__', None))
-            if existing is not None:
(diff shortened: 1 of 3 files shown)


dmontagu:
@adriangb If you are totally against this change we can leave it, but if you'd just want to see other tests pass first, I'm open to any suggestions.

adriangb (approved):

adriangb:
If you feel confident that this doesn't break under edge cases then let's go for it

dmontagu:
I wouldn't say confident, but it _seems_ sound to me. I'll give a bit more thorough review of where we construct schemas related to dataclasses before merging.

@adriangb there's nothing similar with TypedDict now is there? I think we can't do serializers/validators as methods on those (in not-small-part related to inability to inspect class hierarchy), just making sure though.

adriangb:
It should work for TypedDict on the latest typing-extensions version and Python 3.12.

Issue #5496, opened 2023-04-16
V2 pydantic dataclass: InitVar and inheritance don't work together


andriilahuta:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

When using `from __future__ import annotations` the code example raises the following error:
``pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for dataclasses.InitVar[int]. Setting `arbitrary_types_allowed=True` in the model_config may prevent this error.``

### Example Code

```Python
from __future__ import annotations
from dataclasses import InitVar
from pydantic.dataclasses import dataclass

@dataclass
class Foo:
    init: InitVar[int]


@dataclass
class Bar(Foo):
    arg: int
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0a2
        pydantic-core version: 0.24.0 release build profile
                 install path: C:\Program Files\Python\Lib\site-packages\pydantic
               python version: 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)]
                     platform: Windows-10-10.0.22621-SP0
     optional deps. installed: ['typing-extensions']
```

samuelcolvin:
Thanks for reporting, confirmed.

I get a slightly different error `Forward references must evaluate to types. Got dataclasses.InitVar[int].`, but that might be cause I'm on 3.10.

Still definitely needs to be fixed.

dmontagu:
I did some investigation into this, and part of the problem is that dataclasses doesn't expose via "normal" public APIs whether it has determined if a field is an InitVar or not. (In this case, if you look at what it stores on `.__dataclass_fields__`, it has a ForwardRef as the type annotation, even though internally it did realize it should be an `InitVar` by doing some sort of namespace evaluation.)

However, there _is_ a way to determine if `dataclasses` realized the type should be an `InitVar` using only public APIs — the "field" will not show up in `dataclasses.fields(MyDataclass)`, but _will_ be present in `inspect.signature(MyDataclass)`. The main challenge remaining is that recognizing the field is an `InitVar` doesn't tell you what type it is, so you still need to make sure to evaluate the ForwardRef properly. The most likely way for this to work would be to add `InitVar` to the namespace, but there are various scenarios where this might not work perfectly (e.g., `ForwardRef('dataclasses.InitVar[int]')`, importing `InitVar` under a different name, etc.).

[...]

PR #5859, opened 2023-05-24
Add a test demonstrating initvar works with inheritance

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/5496.

~It seems that at some point since the 2.0a4 release, this got fixed, but I've added a test written from the code in the issue. (This test fails in 2.0a4.)~

This got fixed for python 3.11 after #5760, but not for other python versions..

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -211,3 +211,8 @@ def dataclass(
 
 
 __getattr__ = getattr_migration(__name__)
+
+if sys.version_info <= (3, 10):
+    # Monkeypatch dataclasses.InitVar so that typing doesn't error if it occurs as a type when evaluating type hints
+    # This is not necessary in 3.11
+    dataclasses.InitVar.__call__ = lambda: None
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -6,7 +6,7 @@ from __future__ import annotations as _annotations
 import dataclasses
 import sys
 import types
-from typing import TYPE_CHECKING, Any, Callable, Generic, TypeVar, overload
+from typing import TYPE_CHECKING, Any, Callable, Generic, NoReturn, TypeVar, overload
 
 from typing_extensions import Literal, dataclass_transform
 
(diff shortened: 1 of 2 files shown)


dmontagu:
Okay nevermind, it's just resolved in 3.11 :(

For what it's worth, it started working in 3.11 after #5760.  I'll keep working through this.

pydantic/dataclasses.py
```diff
@@ -211,3 +211,8 @@ def create_dataclass(cls: type[Any]) -> type[PydanticDataclass]:
 
 
 __getattr__ = getattr_migration(__name__)
+
+if (3, 8) <= sys.version_info < (3, 11):
```
dmontagu, line 215: Without this, `dataclasses.InitVar[<type>]` raises errors when returned by `typing.get_type_hints`:

```
>           raise TypeError(f"{msg} Got {arg!r:.100}.")
E           TypeError: Forward references must evaluate to types. Got dataclasses.InitVar[int].
```
The issue is, even if you were willing to rely on dataclasses to tell you what is or isn't an InitVar, you really need this function (or something similar) to allow you to resolve forward references to get what the generic parameter resolves to.

I think monkeypatching `InitVar` like this is preferable to copying the whole logic of `get_type_hints` with a tiny tweak to not error if the result isn't callable. But I suppose I could be convinced away from this if we think it's really problematic.

dmontagu:
please review

pydantic/dataclasses.py
```diff
@@ -211,3 +211,8 @@ def create_dataclass(cls: type[Any]) -> type[PydanticDataclass]:
...
+    # Monkeypatch dataclasses.InitVar so that typing doesn't error if it occurs as a type when evaluating type hints
+    # This is not necessary in 3.11, and we don't support InitVar in python < 3.8.
+    dataclasses.InitVar.__call__ = lambda: None  # type: ignore
```
adriangb, line 218: What happens if this gets used before pydantic gets imported?

adriangb (approved):

pydantic/dataclasses.py
dmontagu: It would error if you tried to use `typing.get_type_hints` first, but currently I don't think there is a way to get to the point of building a dataclass schema without importing something from `pydantic.dataclasses`, at least the following works:
```import dataclasses

initvars = []
@dataclasses.dataclass
class A:
    x: 'dataclasses.InitVar[int]'
    def __post_init__(self, x):
        initvars.append(x)

class B(BaseModel):
    a: A

B.model_validate({'a': {'x': 2}})

assert initvars == [2]

B.model_validate({'a': {'x': 'abc'}})  # errors
```

Do you think we need to be more careful about ensuring the monkeypatch happens no matter what import path is followed?

dmontagu: `InitVar.__call__` doesn't actually get called — the `typing.get_type_hints` function errors if it doesn't think the annotations are actually types. It doesn't think `InitVar[int]` is a type pre-3.11 I guess, but one thing you can do to get it to make it think something is a type is to make that thing callable (it has a boolean check `if callable(maybe_type):` in which case it lets you use that annotation and not error). That's why this monkeypatch works — as long as `get_type_hints` doesn't get called on the type until after the monkeypatch is applied (which I think is the case because that doesn't happen until you are building the schema).

adriangb:
Please go ahead and merge

PR #5861, opened 2023-05-25
Support TypeAliasType for generic and recursive type aliases

### LINKED ISSUE #5171: V2: Support forward references in recursive types
Example:

```python
JSON = Union[str, int, float, bool, None, Mapping[str, 'JSON'], List['JSON']]
```

Given how popular this feature request was for MyPy ([https://github.com/python/typing/issues/182](https://github.com/python/typing/issues/182)), I think it would be nice if it "just worked" for Pydantic. Since `pydantic-core` already supports recursive types, it's just a matter of properly handling the recursive forward references.

@dmontagu I wonder if we could use similar work to what you did with generics for these sorts of recursive references? Ultimately I think we need to convert this into a `DefinitionReference` core schema, but I suspect we'll have to do a bit of smart stuff like your generics but with the forward references to detect the recursive forward reference and not just try to "resolve" it.

adriangb:
Fixes #5171
Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 from typing import Any, Callable, Iterable, TypeVar, Union, cast
 
 from pydantic_core import CoreSchema, core_schema
-from typing_extensions import TypeGuard, get_args
+from typing_extensions import TypeAliasType, TypeGuard, get_args
 
 from . import _repr
 
@@ -81,8 +81,11 @@ def get_type_ref(type_: type[Any], args_override: tuple[type[Any], ...] | None =
         args = generic_metadata['args'] or args
 
     module_name = getattr(origin, '__module__', '<No __module__>')
-    qualname = getattr(origin, '__qualname__', f'<No __qualname__: {origin}>')
-    type_ref = f'{module_name}.{qualname}:{id(origin)}'
+    if isinstance(origin, TypeAliasType):
+        type_ref = f'{module_name}.{origin.__name__}'
+    else:
+        qualname = getattr(origin, '__qualname__', f'<No __qualname__: {origin}>')
+        type_ref = f'{module_name}.{qualname}:{id(origin)}'
 
     arg_refs: list[str] = []
     for arg in args:
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1063,7 +1063,7 @@ class GenerateSchema:
         if obj_ref in self.recursion_cache:
             return obj_ref, self.recursion_cache[obj_ref]
         else:
-            self.recursion_cache[obj_ref] = self.definitions[obj_ref] = core_schema.definition_reference_schema(obj_ref)
+            self.recursion_cache[obj_ref] = core_schema.definition_reference_schema(obj_ref)
             return obj_ref, None
 
     def _computed_field_schema(self, d: Decorator[ComputedFieldInfo]) -> core_schema.ComputedField:
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -199,12 +199,10 @@ class GenerateSchema:
         from_prepare_args: bool = True,
     ) -> core_schema.CoreSchema:
         if isinstance(obj, type(Annotated[int, 123])):
-            schema = self._annotated_schema(obj)
-        else:
-            schema = self._generate_schema_for_type(
-                obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
-            )
-        return schema
+            return self._annotated_schema(obj)
+        return self._generate_schema_for_type(
+            obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
+        )
 
     def _generate_schema_for_type(
         self,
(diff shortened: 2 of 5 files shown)


adriangb:
please review

pydantic/_internal/_core_utils.py
```diff
@@ -81,8 +81,11 @@ def get_type_ref(type_: type[Any], args_override: tuple[type[Any], ...] | None =
         args = generic_metadata['args'] or args
 
     module_name = getattr(origin, '__module__', '<No __module__>')
-    qualname = getattr(origin, '__qualname__', f'<No __qualname__: {origin}>')
-    type_ref = f'{module_name}.{qualname}:{id(origin)}'
+    if isinstance(origin, TypeAliasType):
+        type_ref = f'{module_name}.{origin.__name__}'
```
adriangb, line 85: No point in having the `id` there for `TypeAliasType` and since this ends up in the JsonSchema that gives us a cleaner name.

pydantic/_internal/_generate_schema.py
```diff
@@ -199,10 +199,12 @@ def generate_schema(
...
-            return self._annotated_schema(obj)
-        return self._generate_schema_for_type(
-            obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
-        )
+            schema = self._annotated_schema(obj)
+        else:
+            schema = self._generate_schema_for_type(
+                obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
+            )
+        return schema
```
adriangb, line 207:
```suggestion
            return self._annotated_schema(obj)
        return self._generate_schema_for_type(
            obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
        )
```

pydantic/_internal/_generate_schema.py
```diff
@@ -300,7 +302,8 @@ def model_schema(self, cls: type[BaseModel]) -> core_schema.CoreSchema:
 
         model_schema = consolidate_refs(model_schema)
         schema = apply_model_serializers(model_schema, decorators.model_serializers.values())
-        return apply_model_validators(schema, decorators.model_validators.values())
+        schema = apply_model_validators(schema, decorators.model_validators.values())
+        return schema
```
adriangb, line 306:
```suggestion
        return apply_model_validators(schema, decorators.model_validators.values())
```

pydantic/_internal/_generate_schema.py
```diff
@@ -1028,7 +1066,7 @@ def _get_or_cache_recursive_ref(self, cls: type[Any]) -> tuple[str, core_schema.
         if obj_ref in self.recursion_cache:
             return obj_ref, self.recursion_cache[obj_ref]
         else:
-            self.recursion_cache[obj_ref] = core_schema.definition_reference_schema(obj_ref)
+            self.recursion_cache[obj_ref] = self.definitions[obj_ref] = core_schema.definition_reference_schema(obj_ref)
```
adriangb, line 1069:
```suggestion
            self.recursion_cache[obj_ref] = core_schema.definition_reference_schema(obj_ref)
```

pydantic/_internal/_generate_schema.py
```diff
@@ -1051,10 +1089,7 @@ def _annotated_schema(self, annotated_type: Any) -> core_schema.CoreSchema:
         """
         Generate schema for an Annotated type, e.g. `Annotated[int, Field(...)]` or `Annotated[int, Gt(0)]`.
         """
-        first_arg, *other_args = get_args(annotated_type)
-
-        source_type, annotations = self._prepare_annotations(first_arg, other_args)
```
adriangb, line 1056: This is redundant, it gets called in `_apply_annotations` anyway

tests/test_type_alias_type.py
```diff
@@ -0,0 +1,303 @@
...
+            'msg': 'List should have at most 1 item after validation, not 2',
+            'input': [1, 2],
+            'ctx': {'field_type': 'List', 'max_length': 1, 'actual_length': 2},
+        }
+    ]
+
+    assert t.json_schema() == {'type': 'array', 'items': {'type': 'integer'}, 'maxItems': 1}
+
+
+@pytest.mark.xfail(reason='Not working yet, we cannot apply constraints to named/ref types')
```
adriangb, line 156: This is a tricky scenario:

```python
from pydantic import BaseModel, PlainSerializer

def serialize_user(user: 'User'):
    return f'User(<{id(user)}>)'

class User(BaseModel):
    friends: Annotated['User', PlainSerializer(serialize_user)]
```

The problem is that at the point in time when we want to apply `PlainSerializer` to `User` the `User` schema is a `definition-ref` schema that is not yet defined. It so happens that you can add a `serialization` onto a `defintion-ref` schema, and in fact I can't think of any constraints you can add via `Annotated` that would need to modify the model schema itself (e.g. a `AfterValidator` would wrap it, only things like `Gt` modify the actual schema).

But that doesn't work in general: if the "named schema" is a `list[int]` (via `TypeAliasType`) then it should be possible to apply constraints like `MinLen`:

```T = TypeVar('T')
MyList = TypeAliasType('MyList', Annotated['MyList[T]', MinLen(1)], type_params=(T,))```

This should produce something along the lines of  `{'type': 'list', 'items_schema': {'type': 'definition-ref', 'schema_ref': 'MyList_int'}, 'min_length': 1, 'ref': 'MyList_int'}}`.

I tried quite hard to get this all working and could not, so I'm marking these as xfails for now and we can loop back and try to resolve them later.

adriangb:
please review

hramezani (approved):

adriangb:
No, it seems complicated to get right.

PR #5869, opened 2023-05-25
Disallow non-nested types as non-annotated values

dmontagu:
@Kludex as discussed on slack

### DIFF
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -31,7 +31,6 @@ if typing.TYPE_CHECKING:
 IGNORED_TYPES: tuple[Any, ...] = (
     FunctionType,
     property,
-    type,
     classmethod,
     staticmethod,
     PydanticDescriptorProxy,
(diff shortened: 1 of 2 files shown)


Kludex:
Just for the record, the behavior before this PR was to ignore the field.

The simplest example that was brought to me was:

```python


class Foo(BaseModel):
    a = int  # This didn't raise an error before this PR
```

pydantic/_internal/_model_construction.py
```diff
@@ -75,6 +74,13 @@ def inspect_namespace(  # noqa C901
     for var_name, value in list(namespace.items()):
         if var_name == 'model_config':
             continue
+        elif (
+            isinstance(value, type)
+            and value.__module__ == namespace['__module__']
+            and value.__qualname__.startswith(namespace['__qualname__'])
+        ):
+            # `value` is a nested type defined in this namespace; don't error
```
Kludex, line 82: I'm missing something. Conceptually, why is "skipping" nested types defined in this namespace a solution for this issue?

dmontagu: Without this skipping, if you do:
```python
class A(BaseModel):
    class B:
        x = 1
```
it will error. But if you are defining `B` inside of `A`, I'm sort of assuming that was intentional. (This is also necessary to allow `class Config` to not error, though we could handle that based on the name `Config`. I just think we should also allow nested types to be defined without erroring.)

Kludex: got it, thanks 🙏

Kludex (approved):

PR #5873, opened 2023-05-25
Fix kw_only inheritance

dmontagu:
Partially addresses #5271 — I don't think there are any other kw_only issues, at least not that I'm aware of.

(All that's left for #5271 is to add tests of serialization decorators, I think.)

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -546,7 +546,7 @@ class GenerateSchema:
             name,
             common_field['schema'],
             init_only=field_info.init_var or None,
-            kw_only=field_info.kw_only,
+            kw_only=None if field_info.kw_only else False,
             serialization_exclude=common_field['serialization_exclude'],
             validation_alias=common_field['validation_alias'],
             serialization_alias=common_field['serialization_alias'],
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

hramezani (approved):

pydantic/_internal/_generate_schema.py
```diff
@@ -948,6 +948,9 @@ def _dataclass_schema(
             )
         decorators = getattr(dataclass, '__pydantic_decorators__', None) or DecoratorInfos.build(dataclass)
         args = [self._generate_dc_field_schema(k, v, decorators) for k, v in fields.items()]
+        # Move kw_only=False args to the start of the list, as this is how vanilla dataclasses work.
+        # Note that when kw_only is missing or None, it is treated as equivalent to kw_only=True
+        args = [a for a in args if a.get('kw_only') is False] + [a for a in args if a.get('kw_only') is not False]
```
lig, line 953:
```suggestion
        args.sort(key=lambda a: a.get('kw_only') is not False)
```

lig: It's faster and should have less memory footprint.

```
>>> timeit("args = [a for a in args if a is False] + [a for a in args if a is not False]", setup="args = [False, True, None, True, None, False]")
0.46712355499039404
>>> timeit("args.sort(key=lambda a: a is not False)", setup="args = [False, True, None, True, None, False]")
0.3070593159936834
```

lig (changes requested):

pydantic/_internal/_generate_schema.py
lig: I'll open a new PR

PR #5874, opened 2023-05-25
Fix dataclass stuff

### LINKED ISSUE #5271: V2 dataclass improvements

dmontagu:
Closes #5271 

Without this change, we get some misbehavior due to looking up the schema from a parent class.

In particular, on the current main branch:
```python
import dataclasses

import pydantic


class MyDataclass:
    x: int


@dataclasses.dataclass
class MySubDataclass(MyDataclass):
    y: int


class Model(BaseModel):
    dc: MySubDataclass


print(Model.parse_obj({'dc': {'x': 1, 'y': 2}}))
"""
  File "<string>", line 3, in __repr__
AttributeError: 'MySubDataclass' object has no attribute 'y'
"""
```
This error happens because the `__pydantic_validator__` (or maybe `__pydantic_core_schema__`..) from `MyDataclass` is finding its way into the schema used by `Model` thanks to `MySubDataclass.__pydantic_validator__ is MyDataclass.__pydantic_validator__` (since `MySubDataclass` isn't a pydantic dataclass, it doesn't get an updated validator). And therefore the validator isn't setting the `y` attribute on the validated instance the way it should.

I've written this using v1-style method names, since I confirmed that v1 does not have this bug.

(I discovered this while trying to test serialization worked for arbitrary combinations of pydantic/non-pydantic dataclasses.)

-----

The solution is to be careful not to read any of `__pydantic_core_schema__`, `__pydantic_validator__`, `__pydantic_serializer__` from proper parent classes when building schemas.

----

I think whether we merge this change or not, we are in the slightly precarious situation that we need to remember to be careful not to use attribute access on unknown types to retrieve the `__pydantic_serializer__` or similar when there is any chance that that could be associated to a different parent type (the most obvious way this can happen today is if you subclass a pydantic dataclass with a non-pydantic dataclass, you'll run into this problem, but in general you can imagine this happening for other types).

I think our current approach of storing these things as class attributes might be kind of unfortunate, since really I think, conceptually, it would be better if we had a cache whose keys were the model/dataclass/etc. types, and whose values were the schemas/validators/serializers, rather than retrieving them from (possibly strict parent) class attributes.

-----

This same issue may also currently be causing misbehavior in pydantic-core with `to_jsonable_python` for instances of a non-pydantic-dataclass that is a subclass of a pydantic-dataclass, since we do look for the presence of a `__pydantic_serializer__` when serializing, and that may not be for the right class.

I'm not sure what the best way to resolve that is, and I suspect it will eventually make sense to resolve it, but I believe that's a smaller issue than what is addressed here for `TypeAdapter` etc. (In particular, it's generally safe to use a parent class serializer to serialize instances of a subclass, though you might not get the representation you'd expect; however, it's generally _unsafe_ to do the same for validation, which is why this PR is more important.)

### DIFF
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -174,7 +174,7 @@ def is_builtin_dataclass(_cls: type[Any]) -> TypeGuard[type[StandardDataclass]]:
 
 
 def is_pydantic_dataclass(_cls: type[Any]) -> TypeGuard[type[PydanticDataclass]]:
-    return dataclasses.is_dataclass(_cls) and hasattr(_cls, '__pydantic_validator__')
+    return dataclasses.is_dataclass(_cls) and '__pydantic_validator__' in _cls.__dict__
 
 
 def rebuild_dataclass(
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -240,7 +240,7 @@ class DecoratorInfos:
         # reminder: dicts are ordered and replacement does not alter the order
         res = DecoratorInfos()
         for base in model_dc.__bases__[::-1]:
-            existing = cast(Union[DecoratorInfos, None], getattr(base, '__pydantic_decorators__', None))
+            existing = cast(Union[DecoratorInfos, None], base.__dict__.get('__pydantic_decorators__'))
             if existing is None:
                 existing = DecoratorInfos.build(base)
             res.validators.update({k: v.bind_to_cls(model_dc) for k, v in existing.validators.items()})
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -946,7 +946,7 @@ class GenerateSchema:
                 self.types_namespace,
                 typevars_map=typevars_map,
             )
-        decorators = getattr(dataclass, '__pydantic_decorators__', None) or DecoratorInfos.build(dataclass)
+        decorators = dataclass.__dict__.get('__pydantic_decorators__') or DecoratorInfos.build(dataclass)
         args = [self._generate_dc_field_schema(k, v, decorators) for k, v in fields.items()]
         has_post_init = hasattr(dataclass, '__post_init__')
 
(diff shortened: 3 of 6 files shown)


adriangb (approved):

dmontagu:
Closes #5271

Without this change, we get misbehavior due to looking up the schema from a parent class.

On current main branch:
```python
import dataclasses

import pydantic


class MyDataclass:
    x: int


@dataclasses.dataclass
class MySubDataclass(MyDataclass):
    y: int


class Model(BaseModel):
    dc: MySubDataclass


print(Model.parse_obj({'dc': {'x': 1, 'y': 2}}))
# > AttributeError: 'MySubDataclass' object has no attribute 'y'
```

(I discovered this while trying to test serialization worked for arbitrary combinations of pydantic/non-pydantic dataclasses.)

The solution is to be careful not to read any of `__pydantic_core_schema__`, `__pydantic_validator__`, `__pydantic_serializer__` from proper parent classes when building schemas.

----

I think we are in the slightly precarious situation that we need to remember to be careful not to use attribute access on unknown types to retrieve the `__pydantic_serializer__` or similar when there is any chance that that could be associated to a different parent type (the most obvious way this can happen today is if you subclass a pydantic dataclass with a non-pydantic dataclass, you'll run into this problem, but in general you can imagine this happening for other types).

I think our current approach of storing these things as class attributes might be kind of unfortunate, since really I think, conceptually, it would be better if we had a cache whose keys were the model/dataclass/etc. types, and whose values were the schemas/validators/serializers, rather than retrieving them from (possibly strict parent) class attributes.

-----

This same issue may also currently be causing misbehavior in pydantic-core with `to_jsonable_python` for instances of a non-pydantic-dataclass that is a subclass of a pydantic-dataclass, since we do look for the presence of a `__pydantic_serializer__` when serializing, and that may not be for the right class.

I'm not sure what the best way to resolve that is, and I suspect it will eventually make sense to resolve it, but I believe that's a smaller issue than what is addressed here for `TypeAdapter` etc.

PR #5876, opened 2023-05-26
Ensure `__get_pydantic_json_schema__` always gets the schema passed down from a wrapping handler

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -102,12 +102,12 @@ def complete_dataclass(
             schema = get_core_schema(
                 cls,
                 CallbackGetCoreSchemaHandler(
-                    partial(gen_schema.generate_schema_for_type, from_dunder_get_core_schema=False),
+                    partial(gen_schema.generate_schema, from_dunder_get_core_schema=False),
                     gen_schema.generate_schema,
                 ),
             )
         else:
-            schema = gen_schema.generate_schema_for_type(cls, from_dunder_get_core_schema=False)
+            schema = gen_schema.generate_schema(cls, from_dunder_get_core_schema=False)
     except PydanticUndefinedAnnotation as e:
         cls_name = cls.__name__
         if raise_errors:
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -20,6 +20,7 @@ from ._fields import collect_dataclass_fields
 from ._generate_schema import GenerateSchema
 from ._generics import get_standard_typevars_map
 from ._model_construction import MockValidator
+from ._schema_generation_shared import CallbackGetCoreSchemaHandler
 
 if typing.TYPE_CHECKING:
     from ..config import ConfigDict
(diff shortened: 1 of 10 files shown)


tests/test_json_schema.py
```diff
@@ -490,27 +490,17 @@ class Model(BaseModel):
         a: bytes = b'foobar'
         b: Decimal = Decimal('12.34')
 
-    with pytest.warns(
-        DeprecationWarning,
-        match=re.escape('The `schema_json` method is deprecated; use `model_json_schema` and json.dumps instead.'),
-    ):
-        schema_json = Model.schema_json(indent=2)
-        loaded_schema_json = json.loads(schema_json)
```
adriangb, line 498: This just made no sense to have here

adriangb:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -486,8 +489,12 @@ def _arbitrary_type_schema(self, obj: Any, type_: Any) -> CoreSchema:
         else:
             raise PydanticSchemaGenerationError(
                 f'Unable to generate pydantic-core schema for {obj!r}. '
-                f'Set `arbitrary_types_allowed=True` in the model_config to ignore this error'
-                f' or implement `__get_pydantic_core_schema__` on your type to fully support it.'
+                'Set `arbitrary_types_allowed=True` in the model_config to ignore this error'
+                ' or implement `__get_pydantic_core_schema__` on your type to fully support it.'
+                'If you got this error by calling handler(<some type>) within'
```
dmontagu, line 494:
```suggestion
                ' If you got this error by calling handler(<some type>) within'
```

dmontagu: Also, I think it might be nice to add two newlines before the extra info, since I think needing to add `arbitrary_types_allowed=True` will be far more common than calling `handler(<some type>)`.

dmontagu (approved):
I think the `arbitrary_types_allowed=True` error should be a little cleaner but overall this seems good to me.

Curious -- is this enough to get recursive type alias types JSON schema over the line, or does that stil need more?

dmontagu (approved):
I think the `arbitrary_types_allowed=True` error should be a little cleaner but overall this seems good to me.

Curious -- is this enough to get recursive type alias types JSON schema over the line, or does that stil need more?

adriangb:
> Curious -- is this enough to get recursive type alias types JSON schema over the line, or does that stil need more?

Unfortunately, no. This still needs the changes in #5861 but since we should fix this anyway I didn't bother trying that branch on top of these changes, etc.

pydantic/_internal/_dataclasses.py
```diff
@@ -98,9 +99,15 @@ def __init__(__dataclass_self__: PydanticDataclass, *args: Any, **kwargs: Any) -
...
     try:
         if get_core_schema:
-            schema = get_core_schema(cls, partial(gen_schema.generate_schema, from_dunder_get_core_schema=False))
+            schema = get_core_schema(
+                cls,
+                CallbackGetCoreSchemaHandler(
+                    partial(gen_schema.generate_schema_for_type, from_dunder_get_core_schema=False),
+                    gen_schema.generate_schema,
+                ),
+            )
```
adriangb, line 108: We fixed this bug for BaseModel recently, just neglected to do so for dataclasses

pydantic/_internal/_generate_schema.py
```diff
@@ -193,8 +193,25 @@ def arbitrary_types(self) -> bool:
...
+        obj: Any,
     ) -> core_schema.CoreSchema:
+        return self._annotated_schema(obj)
+
+    def generate_schema_for_type(
+        self,
+        obj: Any,
+        from_dunder_get_core_schema: bool = True,
+        from_prepare_args: bool = True,
+    ) -> CoreSchema:
```
adriangb, line 214: I think I can revert/minimize some of these changes

adriangb:
please review

Issue #5582, opened 2023-04-25
Support configs, field validators and such on TypedDict


adriangb:
Since https://github.com/python/typing_extensions/pull/150 is merged we can now store data on TypedDict classes and should be able to treat them much more like models.

adriangb:
We still need configs. But that can be it's own feature request.

PR #5877, opened 2023-05-26
Add validators for TypedDict

adriangb:
Partially resolves https://github.com/pydantic/pydantic/issues/5582

Selected Reviewer: @lig

### DIFF
--- a/pydantic/_internal/_decorators.py
+++ b/pydantic/_internal/_decorators.py
@@ -3,13 +3,15 @@ Logic related to validators applied to models etc. via the `@field_validator` an
 """
 from __future__ import annotations as _annotations
 
+from collections import deque
 from dataclasses import dataclass, field
 from functools import partial, partialmethod
 from inspect import Parameter, Signature, isdatadescriptor, ismethoddescriptor, signature
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Generic, TypeVar, Union, cast
+from itertools import islice
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Generic, Iterable, TypeVar, Union
 
 from pydantic_core import core_schema
-from typing_extensions import Literal, TypeAlias
+from typing_extensions import Literal, TypeAlias, is_typeddict
 
 from ..errors import PydanticUserError
 from ..fields import ComputedFieldInfo
@@ -188,7 +190,7 @@ class Decorator(Generic[DecoratorInfoType]):
         shim: Callable[[Any], Any] | None,
         info: DecoratorInfoType,
     ) -> Decorator[DecoratorInfoType]:
-        func = getattr(cls_, cls_var_name)
+        func = get_attribute_from_bases(cls_, cls_var_name)
         if shim is not None:
             func = shim(func)
         return Decorator(
@@ -206,6 +208,90 @@ class Decorator(Generic[DecoratorInfoType]):
             shim=self.shim,
             info=self.info,
         )
+
+
+def get_bases(tp: type[Any]) -> tuple[type[Any], ...]:
+    if is_typeddict(tp):
+        return tp.__orig_bases__  # type: ignore
+    try:
+        return tp.__bases__
+    except AttributeError:
+        return ()
+
+
+def mro(tp: type[Any]) -> tuple[type[Any], ...]:
+    """
+    Calculate the Method Resolution Order of bases using the C3 algorithm.
+
+    See https://www.python.org/download/releases/2.3/mro/
+    """
+
+    # try to use the existing mro, for performance mainly
+    # but also because it helps verify the implementation below
+    if not is_typeddict(tp):
+        try:
+            return tp.__mro__
+        except AttributeError:
+            # GenericAlias and some other cases
+            pass
+
+    def merge_seqs(seqs: list[deque[type[Any]]]) -> Iterable[type[Any]]:
+        while True:
+            non_empty = [seq for seq in seqs if seq]
+            if not non_empty:
+                # Nothing left to process, we're done.
+                return
+            candidate: type[Any] | None = None
+            for seq in non_empty:  # Find merge candidates among seq heads.
+                candidate = seq[0]
+                not_head = [s for s in non_empty if candidate in islice(s, 1, None)]
+                if not_head:
+                    # Reject the candidate.
+                    candidate = None
+                else:
+                    break
+            if not candidate:
+                raise TypeError('Inconsistent hierarchy, no C3 MRO is possible')
+            yield candidate
+            for seq in non_empty:
+                # Remove candidate.
+                if seq[0] == candidate:
+                    seq.popleft()
+
+    bases = get_bases(tp)
+    seqs = [deque(mro(base)) for base in bases] + [deque(bases)]
+    res = tuple(merge_seqs(seqs))
+
+    return (tp,) + res
+
+
+def get_attribute_from_bases(tp: type[Any], name: str) -> Any:
+    """
+    Get the attribute from the next class in the MRO that has it,
+    aiming to simulate calling the method on the actual class.
+
+    The reason for iterating over the mro instead of just getting
+    the attribute (which would do that for us) is to support TypedDict,
+    which lacks a real __mro__, but can have a virtual one constructed
+    from its bases (as done here).
+
+    Args:
+        tp (type[Any]): The type or class to search for the attribute.
+        name (str): The name of the attribute to retrieve.
+
+    Returns:
+        Any: The attribute value, if found.
+
+    Raises:
+        AttributeError: If the attribute is not found in any class in the MRO.
+    """
+    try:
+        return getattr(tp, name)
+    except Exception as e:
+        for base in reversed(mro(tp)):
+            if hasattr(base, name):
+                return getattr(base, name)
+        raise e
 
 
 @slots_dataclass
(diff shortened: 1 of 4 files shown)


adriangb:
please review

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +210,90 @@ def bind_to_cls(self, cls: Any) -> Decorator[DecoratorInfoType]:
         )
 
 
+def get_bases(tp: type[Any]) -> tuple[type[Any], ...]:
+    if is_typeddict(tp):
+        return tp.__orig_bases__  # type: ignore
```
lig, line 215: I see here https://github.com/python/cpython/pull/103698/files it should be a tuple as well. Could we do something like `typing.cast` here instead of `type: ignore`?

The type error here is likely caused by the fact that `type[Any]` doesn't necessarily  have `__orig_bases__` attr. Shall we add an assert here or cast `tp` then?

A `TypeGuard` on `is_typeddict` would help. Shall we add a todo here and contribute the `TypeGuard` to `typing_extensions`?

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +210,90 @@ def bind_to_cls(self, cls: Any) -> Decorator[DecoratorInfoType]:
...
+                    candidate = None
+                else:
+                    break
+            if not candidate:
+                raise TypeError('Inconsistent hierarchy, no C3 MRO is possible')
+            yield candidate
+            for seq in non_empty:
+                # Remove candidate.
+                if seq[0] == candidate:
+                    seq.popleft()
```
lig, line 258: This doesn't have to create а closure as far as I can see. Could be a private module level function, imo.

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +210,90 @@ def bind_to_cls(self, cls: Any) -> Decorator[DecoratorInfoType]:
...
+    Returns:
+        Any: The attribute value, if found.
+
+    Raises:
+        AttributeError: If the attribute is not found in any class in the MRO.
+    """
+    e: Exception | None = None
+    for base in [tp, *reversed(mro(tp))]:
+        try:
+            return getattr(base, name)
```
lig, line 290: `hasattr` and `if` is faster than catching exception on `getattr`. Could we use `hasattr` here?

pydantic/_internal/_decorators.py
```diff
@@ -239,8 +325,9 @@ def build(model_dc: type[Any]) -> DecoratorInfos:  # noqa: C901 (ignore complexi
 
         # reminder: dicts are ordered and replacement does not alter the order
         res = DecoratorInfos()
-        for base in model_dc.__bases__[::-1]:
-            existing = cast(Union[DecoratorInfos, None], getattr(base, '__pydantic_decorators__', None))
+        # use __orig_bases__ to support TypedDict but fall back to __bases__ for models and dataclasses
```
lig, line 328: Are you sure this comment belongs here? I don't see how it applies to the code below.

pydantic/_internal/_decorators.py
```diff
@@ -298,6 +385,10 @@ def build(model_dc: type[Any]) -> DecoratorInfos:  # noqa: C901 (ignore complexi
                         model_dc, cls_var_name=var_name, shim=None, info=info
                     )
                 setattr(model_dc, var_name, var_value.wrapped)
+        try:
+            setattr(model_dc, '__pydantic_decorators__', res)
+        except (TypeError, AttributeError):
```
lig, line 390: Can we check explicitly for cases where we cannot set `__pydantic_decorators__` here and avoid relying on exceptions? At least a comment explaining the motivation behind this would be helpful here, imo.

lig (changes requested):

lig:
please update

pydantic/_internal/_decorators.py
adriangb: Not that simple: https://github.com/microsoft/pyright/issues/3442.
A type ignore here is by far the easiest solution.

pydantic/_internal/_decorators.py
adriangb: Closures don't have to exist only to capture variables. Here it is serving as a namespace to keep things tidy since the function is only called from one place.

pydantic/_internal/_decorators.py
adriangb: I agree it can be removed now, thanks.

pydantic/_internal/_decorators.py
adriangb: Reworked a bit. There's a comment now and no try/except

pydantic/_internal/_decorators.py
adriangb: Changed to

```python
    try:
        return getattr(tp, name)
    except Exception as e:
        for base in reversed(mro(tp)):
            if hasattr(base, name):
                return getattr(base, name)
        raise e
```

pydantic/_internal/_decorators.py
```diff
@@ -208,6 +210,89 @@ def bind_to_cls(self, cls: Any) -> Decorator[DecoratorInfoType]:
...
+    if is_typeddict(tp):
+        return tp.__orig_bases__  # type: ignore
+    try:
+        return tp.__bases__
+    except AttributeError:
+        return ()
+
+
+def mro(tp: type[Any]) -> tuple[type[Any], ...]:
+    """Calculate the Method Resolution Order of bases using the C3 algorithm.
```
dmontagu, line 223:
```suggestion
    """
    Calculate the Method Resolution Order of bases using the C3 algorithm.
```

pydantic/dataclasses.py
```diff
@@ -162,7 +162,7 @@ def create_dataclass(cls: type[Any]) -> type[PydanticDataclass]:
         original_cls = cls
 
         config_wrapper = _config.ConfigWrapper(config)
-        decorators = _decorators.DecoratorInfos.build(cls)
+        cls.__pydantic_decorators__ = _decorators.DecoratorInfos.build(cls)  # type: ignore
```
dmontagu, line 165: Note that this PR https://github.com/pydantic/pydantic/pull/5874 will move to re-computing the decorators whenever they aren't present on the exact class. Not sure if that will cause issues with this change.

dmontagu: (I just merged that PR)

adriangb: I was just cleaning up here: we used `decorators`  just as a temp variable and then assigned to `cls. __pydantic_decorators__`.

dmontagu: the line immediately before the old version of this was:
```python
        cls = dataclasses.dataclass(  # type: ignore[call-overload]
            cls,
            init=init,
            repr=repr,
            eq=eq,
            order=order,
            unsafe_hash=unsafe_hash,
            frozen=frozen,
            **kwargs,
        )
```
does this not change the value of `cls`? (Just modify it?) That's why I thought there was a change to the semantics not just cleanup.

adriangb: let me just revert this then

tests/test_types_typeddict.py
```diff
@@ -646,3 +647,87 @@ class Model(BaseModel):
...
+        x: int
+
+    class Child(Parent):
+        y: float
+
+    ta = TypeAdapter(Child)
+    assert ta.validate_python({'x': '1', 'y': '1.0'}) == {'x': 1, 'y': 1.0}
+
+
+def test_typeddict_field_validator(TypedDict: Any) -> None:
```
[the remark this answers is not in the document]
adriangb: done

adriangb:
We'll need to disregard the pydantic-settings dependency conflict and update that separately.

adriangb:
please review

adriangb:
I guess we also want to support `model_config` on TypedDict?

dmontagu:
might be a good idea to use a dunder for it (like we do on dataclasses)?

adriangb:
Sounds good. Not sure that needs to block this PR though.

pydantic/_internal/_decorators.py
```diff
@@ -239,8 +325,8 @@ def build(model_dc: type[Any]) -> DecoratorInfos:  # noqa: C901 (ignore complexi
 
         # reminder: dicts are ordered and replacement does not alter the order
         res = DecoratorInfos()
-        for base in model_dc.__bases__[::-1]:
-            existing = cast(Union[DecoratorInfos, None], base.__dict__.get('__pydantic_decorators__'))
+        for base in reversed(mro(model_dc)[1:]):
+            existing: DecoratorInfos | None = base.__dict__.get('__pydantic_decorators__', None)
```
dmontagu, line 329:
```suggestion
            existing: DecoratorInfos | None = base.__dict__.get('__pydantic_decorators__')
```
`.get` returns `None` if not present, so I think this is equivalent. Not a big deal

dmontagu (approved):
LGTM; I'm not sure what config stuff remains; if you have any concrete ideas of "shortcomings" it's maybe worth creating an issue describing it when we merge, if you aren't sure, 🤷 maybe we just wait for the feature request.

pydantic/_internal/_decorators.py
adriangb: I remember that get does and pop doesn’t or the other way around, so I always try to be explicit because I get them confused, and I feel like it helps people who don’t know what the default return value is

adriangb: That is, I will gladly add four characters and some spots to alleviate the cognitive load of remembering which works, and which doesn’t. Unless they both work the same in which case I’m just dumb.

adriangb:
Looks like adding a `ConfigDict` would require changes in `pydantic-core`. I propose we merge this as-is and work on that later.

PR #5883, opened 2023-05-26
Enforce constraints on arbitrary types via functional validators

adriangb:
This makes constraints work on arbitrary types that don't "know" how to consume them by applying a functional validator that enforces the constraint in Python. Of course, if the type wants to it can "consume" the constraint via `__prepare_pydantic_annotations__` to e.g. use a more efficient implementation, **or** if the schema is a schema that accepts the constraint directly (e.g. `gt` w/ a `float` schema) then it gets applied that way instead of via Python.

Selected Reviewer: @dmontagu


adriangb:
please review

pydantic/_internal/_known_annotated_metadata.py
```diff
@@ -28,31 +33,130 @@
...
+                schema,
+            )
+    elif isinstance(annotation, at.MultipleOf):
+        if schema['type'] in NUMERIC_SCHEMA_TYPES:
+            schema.update(schema_update)  # type: ignore
+        else:
+            return cs.no_info_after_validator_function(
+                partial(_validators.multiple_of_validator, multiple_of=annotation.multiple_of),
+                schema,
+            )
```
samuelcolvin, line 112: could we turns these into a dict and a lookup rather than duplication, no issue if that's not possible or would be uglier.

pydantic/_internal/_known_annotated_metadata.py
```diff
@@ -28,31 +33,130 @@
...
+    elif isinstance(annotation, at.MaxLen):
+        if schema['type'] in SEQUENCE_SCHEMA_TYPES:
+            schema.update(schema_update)  # type: ignore
+        else:
+            return cs.no_info_after_validator_function(
+                partial(_validators.max_length_validator, max_length=annotation.max_length),
+                schema,
+            )
```
samuelcolvin, line 128: same.

pydantic/_internal/_known_annotated_metadata.py
```diff
@@ -28,31 +33,130 @@
...
+    elif isinstance(annotation, at.Predicate):
+        predicate_name = f'{annotation.func.__qualname__} ' if hasattr(annotation.func, '__qualname__') else ''
+
+        def val_func(v: Any) -> Any:
+            # annotation.func may also raise an exception, let it pass through
+            assert annotation.func(v), f'Predicate {predicate_name}failed'
```
samuelcolvin, line 134: might be cleaner to raise a `PydanticCustomError`.

pydantic/_internal/_validators.py
```diff
@@ -226,3 +226,55 @@ def ip_v6_interface_validator(__input_value: Any) -> IPv6Interface:
         return IPv6Interface(__input_value)
     except ValueError:
         raise PydanticCustomError('ip_v6_interface', 'Input is not a valid IPv6 interface')
+
+
+def greater_than_validator(x: Any, gt: Any) -> Any:
+    if not (x > gt):
+        raise PydanticCustomError('greater_than', 'Input should be greater than {gt}', {'gt': gt})
```
samuelcolvin, line 233:
```suggestion
        raise PydanticKnownError('greater_than', {'gt': gt})
```

pydantic/_internal/_validators.py
```diff
@@ -226,3 +226,55 @@ def ip_v6_interface_validator(__input_value: Any) -> IPv6Interface:
...
+    return x
+
+
+def greater_than_or_equal_validator(x: Any, ge: Any) -> Any:
+    if not (x >= ge):
+        raise PydanticCustomError('greater_than_equal', 'Input should be greater than or equal to {ge}', {'le': ge})
```
samuelcolvin, line 239:
```suggestion
        raise PydanticKnownError('greater_than_equal', {'le': ge})
```

pydantic/_internal/_validators.py
```diff
@@ -226,3 +226,55 @@ def ip_v6_interface_validator(__input_value: Any) -> IPv6Interface:
...
+    return x
+
+
+def less_than_validator(x: Any, lt: Any) -> Any:
+    if not (x < lt):
+        raise PydanticCustomError('less_than', 'Input should be less than {lt}', {'lt': lt})
```
samuelcolvin, line 245:
```suggestion
        raise PydanticKnownError('less_than', {'lt': lt})
```

pydantic/_internal/_validators.py
```diff
@@ -226,3 +226,55 @@ def ip_v6_interface_validator(__input_value: Any) -> IPv6Interface:
...
+    return x
+
+
+def less_than_or_equal_validator(x: Any, le: Any) -> Any:
+    if not (x <= le):
+        raise PydanticCustomError('less_than_equal', 'Input should be less than or equal to {le}', {'le': le})
```
samuelcolvin, line 251:
```suggestion
        raise PydanticKnownError('less_than_equal', {'le': le})
```

pydantic/_internal/_validators.py
```diff
@@ -226,3 +226,55 @@ def ip_v6_interface_validator(__input_value: Any) -> IPv6Interface:
...
+    return x
+
+
+def multiple_of_validator(x: Any, multiple_of: Any) -> Any:
+    if not (x % multiple_of == 0):
+        raise PydanticCustomError(
+            'multiple_of', 'Input should be a multiple of {multiple_of}', {'multiple_of': multiple_of}
+        )
```
samuelcolvin, line 259:
```suggestion
        raise PydanticKnownError('multiple_of', {'multiple_of': multiple_of})
```

pydantic/_internal/_validators.py
```diff
@@ -226,3 +226,55 @@ def ip_v6_interface_validator(__input_value: Any) -> IPv6Interface:
...
+    return x
+
+
+def min_length_validator(x: Any, min_length: Any) -> Any:
+    if not (len(x) >= min_length):
+        raise PydanticCustomError(
+            'too_short',
+            'Should have at least {min_length} item after validation, not {actual_length}',
+            {'min_length': min_length, 'actual_length': len(x)},
+        )
```
samuelcolvin, line 269:
```suggestion
        raise PydanticKnownError('too_short', {'min_length': min_length, 'actual_length': len(x)})
```

pydantic/_internal/_validators.py
```diff
@@ -226,3 +226,55 @@ def ip_v6_interface_validator(__input_value: Any) -> IPv6Interface:
...
+    return x
+
+
+def max_length_validator(x: Any, max_length: Any) -> Any:
+    if not (len(x) < max_length):
+        raise PydanticCustomError(
+            'too_long',
+            'Should have at most {min_length} item after validation, not {actual_length}',
+            {'max_length': max_length, 'actual_length': len(x)},
+        )
```
samuelcolvin, line 279:
```suggestion
        raise PydanticKnownError('too_long', {'max_length': max_length, 'actual_length': len(x)})
```

samuelcolvin (commented):
otherwise LGTM.

pydantic/_internal/_known_annotated_metadata.py
adriangb: Any suggestions for the right error to raise? I was hoping to make it "similar" to what would happen if you called the predicate by hand inside of your own functional validator, and since the predicate returns a boolean I felt like an `assert` would be the natural thing to do.

pydantic/_internal/_known_annotated_metadata.py
adriangb: going to leave this as is for now. I think I'll be fighting the type checker to access `annotation.multiple_of` and such if not.

pydantic/_internal/_validators.py
adriangb: Won't work:

```
    def greater_than_validator(x: Any, gt: Any) -> Any:
        if not (x > gt):
>           raise PydanticKnownError('greater_than', {'gt': gt})
E           TypeError: GreaterThan: 'gt' context value must be a Number
```

pydantic/_internal/_known_annotated_metadata.py
adriangb: done

pydantic/_internal/_validators.py
samuelcolvin: Should be fixed in core, bit this is fine for now.

samuelcolvin: I hadn't realised it was every single one that was using `PydanticCustomError`, we should adopt `PydanticKnownError` where possible.

adriangb: Unless I’m missing something, you already pointed that out and as per above, it won’t work unless we make extensive changes in pydantic-core. It also makes no difference for users as far as I can tell.

samuelcolvin (approved):

PR #5889, opened 2023-05-26
Fix `@model_validator` on nested models

### LINKED ISSUE #5790: model_validator not called for nested Models
### Initial Checks

### Description

Using the model_validator as proposed in #4617 I fail to validate nested self-referencing models.
I was using a root_validator previously, which worked fine.

Using a root_validator:

```python
from typing import Dict, Any

class Schema(BaseModel):
    properties: Dict[str, "Schema"] = None
    extensions: Dict[str, Any] = None
    model_config = {'arbitrary_types_allowed': False, 'extra': 'allow'} # use with forbid

    @root_validator(pre=True)
    def validate_extensions(cls, values):
        print(f"root_validate {values}")
        ext = dict()
        for k,v in values.items():
            if k.startswith("x-"):
                ext[k] = v
        if ext:
            for k in ext.keys():
                del values[k]
            values["extensions"] = ext
        return values


v = Schema.model_validate(
    {
        "x-Schema":"Schema"
    }
)
assert v.__pydantic_extra__ == {}
assert v.extensions["x-Schema"] == "Schema"

v = Schema.model_validate(
    {
        "x-Schema":"Schema",
        "properties": {
            "A": {
                "x-A":"A"
            }
        }
    }
)


assert v.properties["A"].extensions is not None and "x-A" in v.properties["A"].extensions
```

the validator is called for the nested property:
```
root_validate {'x-Schema': 'Schema'}
root_validate {'x-Schema': 'Schema', 'properties': {'A': {'x-A': 'A'}}}
root_validate {'x-A': 'A'}
```

Using a model_validator I miss the validation of the nested property:
```
    assert v.properties["A"].extensions is not None and "x-A" in v.properties["A"].extensions
AssertionError
model_validate {'x-Schema': 'Schema'}
model_validate {'x-Schema': 'Schema', 'properties': {'A': {'x-A': 'A'}}}
```

Am I wrong assuming this should work?

### Example Code

```Python
from typing import Dict, Any

class Schema(BaseModel):
    properties: Dict[str, "Schema"] = None
    extensions: Dict[str, Any] = None
    model_config = {'arbitrary_types_allowed': False, 'extra': 'allow'} # or use with forbid

    @model_validator(mode="before")
    def validate_extensions(cls, values):
        print(f"model_validate {values}")
        ext = dict()
        for k,v in values.items():
            if k.startswith("x-"):
                ext[k] = v
        if ext:
            for k in ext.keys():
                del values[k]
            values["extensions"] = ext
        return values


v = Schema.model_validate(
    {
        "x-Schema":"Schema"
    }
)
assert v.__pydantic_extra__ == {}
assert v.extensions["x-Schema"] == "Schema"

v = Schema.model_validate(
    {
        "x-Schema":"Schema",
        "properties": {
            "A": {
                "x-A":"A"
            }
        }
    }
)


assert v.properties["A"].extensions is not None and "x-A" in v.properties["A"].extensions
```

```Text
pydantic version: 2.0a4
```

adriangb:
Fixes #5790

Selected Reviewer: @Kludex


adriangb:
TODO: need test for `@model_serializer`

adriangb:
please review

pydantic/functional_serializers.py
```diff
@@ -181,7 +181,7 @@ def model_serializer(
     *,
     mode: Literal['plain', 'wrap'] = 'plain',
     json_return_type: _core_schema.JsonReturnTypes | None = None,
-) -> Callable[[Any], _decorators.PydanticDescriptorProxy[Any]] | _decorators.PydanticDescriptorProxy[Any]:
+) -> Callable[[Any], Any]:
```
dmontagu, line 184: Is it a problem that the non-callable return type was removed here? I mean, I'm guessing it's not a problem, but it would it be more careful in some way? Happy to leave like this if preferable just want to make sure it wasn't an oversight.

adriangb: I just pushed a slightly better version. We can get even more granular (e.g. defining a couple of `Protocol`s like we do form some of the other validators/serializers) but I don't want to do that in this PR, I'm only changing it because it was causing issues with the test I'm adding here.

samuelcolvin: pls can we add (if we don't already have) an example of both `@model_validator` and `@model_validator(...)` in mypy tests please.

samuelcolvin (changes requested):
otherwise LGTM, tests seem to be failing

dmontagu (approved):

Issue #3523, opened 2021-12-14
Layout conflict in ModelMetaclass when mixing in two models with private variables.


igoras1993:
# Bug
Disclaimer: I have checked out stackoverflow, issue tracker, docs, google etc. for resources on this, but didnt found any solution. 

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.8.2
            pydantic compiled: True
                 install path: /home/igor/code/issues-reports/pydantic-private-mixins/venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.9 (main, Nov 16 2021, 03:08:02)  [GCC 9.3.0]
                     platform: Linux-5.11.0-34-generic-x86_64-with-glibc2.31
     optional deps. installed: ['typing-extensions']

```

Following code looks innocent, It just uses mixin models to merge some private attributes in one bigger model:

```py
import pydantic
import typing as t


class ModelMixin(pydantic.BaseModel):
    _mixin_private: t.Optional[str] = pydantic.PrivateAttr(None)


class Model(pydantic.BaseModel):
    public: str = "default"
    _private: t.Optional[str] = pydantic.PrivateAttr(None)


class NewModel(ModelMixin, Model):
    pass

```
But quite surprisingly it raises a `TypeError` on a class creation time:

```
Traceback (most recent call last):
  File "/code/issues-reports/pydantic-private-mixins/report.py", line 14, in <module>
    class NewModel(ModelMixin, Model):
  File "pydantic/main.py", line 367, in pydantic.main.ModelMetaclass.__new__
  File "/usr/lib/python3.9/abc.py", line 106, in __new__
    cls = super().__new__(mcls, name, bases, namespace, **kwargs)
TypeError: multiple bases have instance lay-out conflict
```
It seems that **at least two** models with private attributes will trigger this exception. When private attributes are defined in only one of the models, everything works great. 

I can dig into this, but probably not this week. But please confirm that this is an issue, not the intended behavior. Also, if You have any ideas/hints where to start, it would be appreciated!

igoras1993:
Okay, some update. 
It seems like problem with `__slots__` and multiple inheritance. Pydantic's documentation states directly, that creating a private attribute will also create a `__slots__` entry:
> [Upon class creation pydantic constructs __slots__ filled with private attributes.](https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes)

[...]

PR #5891, opened 2023-05-26
Use `__pydantic_private__` for storing private attributes

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/3523

Requires https://github.com/pydantic/pydantic-core/pull/640.

The only weird thing remaining is that the private attributes are no longer member descriptors. I don't know how much that matters.. But the test of `__set_name__` functionality didn't require modifications.. 🤷 

Selected Reviewer: @hramezani


tests/test_private_attributes.py
```diff
@@ -245,17 +241,21 @@ class ParentBModel(GrandParentModel):
...
-        assert repr(Model._bar) == "<member '_bar' of 'ParentBModel' objects>"
-        assert repr(Model._baz) == "<member '_baz' of 'Model' objects>"
+    # if platform.python_implementation() == 'PyPy':
+    #     assert repr(Model._foo).startswith('<member_descriptor object at')
+    #     assert repr(Model._bar).startswith('<member_descriptor object at')
+    #     assert repr(Model._baz).startswith('<member_descriptor object at')
+    # else:
+    #     assert repr(Model._foo) == "<member '_foo' of 'GrandParentModel' objects>"
+    #     assert repr(Model._bar) == "<member '_bar' of 'ParentBModel' objects>"
+    #     assert repr(Model._baz) == "<member '_baz' of 'Model' objects>"
```
dmontagu, line 251: With this change, instead of `"<member '_foo' of 'GrandParentModel' objects>"` you get `"ModelPrivateAttr(default={'a': {}})"`.

This seems to me to be the biggest potential "issue" remaining with this change.

As far as I can tell, these checks of `repr` aren't really checking any deeper behavior, but I don't know if this is likely reflective of deeper issues caused by the fact that `Model._foo` would no longer be a member object.

I'm thinking _maybe_ there's something we could do in the `ModelMetaclass.__getattr__` or similar but I'm in a bit over my head and this stuff is not super well documented.

But `tests.test_create_model.test_private_attr_set_name` and `tests.test_create_model.test_private_attr_set_name` are still passing which makes me hopeful that this is not having significant impacts on usage.

dmontagu:
please review

samuelcolvin (approved):

PR #5898, opened 2023-05-27
Move model before validators

samuelcolvin:
As discussed yesterday with @dmontagu.

I know this changes the behaviour of `test_model_validator_before`, but I think it's actually more correct.

Previously `before` model validators were called when the input is a model, even when validation wasn't being run. Now before validators are called in the same scenarios as the rest of model validation is run:
* on both instances of the model type and instances of subclasses if `revalidate_instances='always'`
* only on instances of subclasses if `revalidate_instances='subclass-instances'`
* on neither instances of the model type nor instances of subclasses if `revalidate_instances='never'`

As previously, `before` model validators are always called when the input does not pass `isinstance(input, Model)`.

This change does mean the logic for when before validators are called differs from `wrap` and `after` model validators - but honestly I think `wrap` and `after` model validators are wrong, although I'm not sure that's worth changing.

We should probably add tests for:
* `model_validator(mode='before')` on `RootModel`
* `model_validator(mode='before')` on `dataclass`

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -267,7 +267,7 @@ class GenerateSchema:
         if cls.__pydantic_root_model__:
             root_field = self._common_field_schema('root', fields['root'], decorators)
             inner_schema = root_field['schema']
-            inner_schema = apply_before_model_validators(inner_schema, model_validators)
+            inner_schema = apply_model_validators(inner_schema, model_validators, 'inner')
             model_schema = core_schema.model_schema(
                 cls,
                 inner_schema,
@@ -290,7 +290,7 @@ class GenerateSchema:
 
             inner_schema = apply_validators(fields_schema, decorators.root_validators.values())
             inner_schema = define_expected_missing_refs(inner_schema, recursively_defined_type_refs())
-            inner_schema = apply_before_model_validators(inner_schema, model_validators)
+            inner_schema = apply_model_validators(inner_schema, model_validators, 'inner')
 
             model_schema = core_schema.model_schema(
                 cls,
@@ -305,7 +305,7 @@ class GenerateSchema:
 
         model_schema = consolidate_refs(model_schema)
         schema = apply_model_serializers(model_schema, decorators.model_serializers.values())
-        return apply_other_model_validators(schema, model_validators)
+        return apply_model_validators(schema, model_validators, 'outer')
 
     def _generate_schema_from_prepare_annotations(self, obj: Any) -> core_schema.CoreSchema | None:
         """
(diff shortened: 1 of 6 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -1299,6 +1308,24 @@ def apply_model_serializers(
...
+    return apply_model_validators(inner_schema, [x for x in validators if x.info.mode == 'before'])
+
+
+def apply_other_model_validators(
+    model_schema: core_schema.CoreSchema, validators: Iterable[Decorator[ModelValidatorDecoratorInfo]]
+) -> core_schema.CoreSchema:
+    """
+    model validators other than `before` are applied to the outer model_schema
+    """
+    return apply_model_validators(model_schema, [x for x in validators if x.info.mode != 'before'])
```
adriangb, line 1326:
```suggestion
def apply_model_validators(
    model_schema: core_schema.CoreSchema, validators: Iterable[Decorator[ModelValidatorDecoratorInfo]], mode: str
) -> core_schema.CoreSchema:
    """
    model validators other than `before` are applied to the outer model_schema
    """
    return apply_model_validators(model_schema, [x for x in validators if x.info.mode == mode])
```

Maybe a single one? Or accept `mode: set[str]`?

tests/test_model_validator.py
```diff
@@ -43,6 +43,32 @@ def val_model(cls, values: Any, info: ValidationInfo) -> dict[str, Any] | Model:
                 values.y += 1
             return values
 
+    m = Model(x=1, y=2)
+    assert m.model_dump() == {'x': 2, 'y': 3}
+    # model not changed because we don't revalidate m
+    assert Model.model_validate(m).model_dump() == {'x': 2, 'y': 3}
```
adriangb, line 49: Is this the change of behavior? Does this mean that we don't run before model validators if it's already a model?

adriangb (approved):
What's the change of behavior for `test_model_validator_before`? I don't see that test being changed.

tests/test_model_validator.py
samuelcolvin: yes, will add. This is `test_model_validator_before` changing.

pydantic/_internal/_generate_schema.py
samuelcolvin: fixed.

tests/test_model_validator.py
samuelcolvin: fixed.

samuelcolvin:
please review.

samuelcolvin:
Merging this, happy to change/revert/add more if needed.

PR #5901, opened 2023-05-27
Update typing-extensions to 4.6.1

adriangb:


.github/workflows/ci.yml
```diff
@@ -268,7 +268,7 @@ jobs:
       fail-fast: false
       matrix:
         # test the oldest supported version and main
-        typing-extensions-version: ['4.5.0', 'main']
+        typing-extensions-version: ['4.6.1', 'main']
```
hramezani, line 271: 4.6.2 is out. Should we use that one?

adriangb: No this should be the minimum supported version. We don't need 4.6.2 as far as I can tell so let's stick to 4.6.1 for now.

samuelcolvin (approved):

fergusdixon:
A consequence of this is that Tensorflow is not compatible with Pydantic V2 due to the restraints on typing_extensions
What was the reasoning behind restricting this?

adriangb:
We wanted to support TypedDict and TypeAliasType.

I'm not sure why TensorFlow puts an upper bound on te, it's a relatively stable package.

fergusdixon:
Seems to be a lot of back and forth between TF, wrapt, and typing-extensions over who exactly needs to fix what
They're tracking it here https://github.com/tensorflow/tensorflow/issues/60687

adriangb:
Thanks for the link. Not much we can do at this point, hopefully TF / CPython / TE solve it eventually.

PR #5905, opened 2023-05-28
Uprev pydantic-core to 0.37.0

adriangb:
Selected Reviewer: @hramezani

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -99,6 +99,7 @@ from pydantic import (
 )
 from pydantic.annotated import GetCoreSchemaHandler
 from pydantic.errors import PydanticSchemaGenerationError
+from pydantic.functional_validators import AfterValidator
 from pydantic.json_schema import GetJsonSchemaHandler, JsonSchemaValue
 from pydantic.types import AllowInfNan, ImportString, InstanceOf, SecretField, SkipValidation, Strict, TransformSchema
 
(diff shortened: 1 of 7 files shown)


tests/test_validator.py
```diff
@@ -69,7 +69,7 @@ def test_global_namespace_variables():
 
 
 def test_local_namespace_variables():
-    IntList = List[int]
+    IntList = List[int]  # noqa: F841
```
adriangb, line 72: I had to rebuild the lockfile instead of refreshing it for some reason. Not a big deal but that updated Ruff, which triggered this false positive.

adriangb:
please review

tests/test_types.py
```diff
@@ -5370,11 +5371,39 @@ class Model(BaseModel):
...
+
+
+def test_annotated_default_value_validate_default() -> None:
+    t = TypeAdapter(Annotated[List[int], Field(default=['1', '2'])], config=ConfigDict(validate_default=True))
+
+    r = t.get_default_value()
+    assert r is not None
+    assert r.value == [1, 2]
+
+    # insert_assert(t.json_schema())
```
hramezani, line 5391: Do we need this assertion? As it's commented

tests/test_types.py
```diff
@@ -5370,11 +5371,39 @@ class Model(BaseModel):
...
+
+    # the order of the args should not matter, we always put the default value on the outside
+    for tp in (WithDefaultValue[WithAfterValidator[List[int]]], WithAfterValidator[WithDefaultValue[List[int]]]):
+        t = TypeAdapter(tp, config=ConfigDict(validate_default=True))
+
+        r = t.get_default_value()
+        assert r is not None
+        assert r.value == [2, 4]
+
+        # insert_assert(t.json_schema())
```
hramezani, line 5408: Same here

hramezani (approved):

tests/test_types.py
adriangb: yeah we leave those in there, makes it easy to know how to regenerate em

PR #5920, opened 2023-05-29
support for `return_schema` on functional serializers

samuelcolvin:
Companion to pydantic/pydantic-core#646.


pydantic/functional_serializers.py
```diff
@@ -18,13 +18,13 @@ class PlainSerializer:
 
     Attributes:
         func (core_schema.SerializerFunction): The serializer function.
-        json_return_type (core_schema.JsonReturnTypes | None, optional): The JSON return type. Defaults to `None`.
+        return_schema: Schema to build serializer for return type. Defaults to `None`.
```
dmontagu, line 21: Would it make sense to change this to be more similar to `@computed_field`? There, we have
`return_type: Any = _UNSET` and docstring (which probably needs some updates):
```
    return_type: optional return for serialization logic to expect when serialising to JSON, if included
            this must be correct, otherwise a `TypeError` is raised.
            If you don't include a return type Any is used, which does runtime introspection to handle arbitrary
            objects.
```

dmontagu: and logic:
```python
        if return_type is _UNSET:
            # try to get it from the type annotation
            res = _typing_extra.get_type_hints(_decorators.unwrap_wrapped_function(f), include_extras=True)
            return_type = res.get('return', _UNSET)
            if return_type is _UNSET:
                raise PydanticUserError(
                    'Computed field is missing return type annotation or specifying `return_type`'
                    ' to the `@computed_field` decorator (e.g. `@computed_field(return_type=int|str)`)',
                    code='model-field-missing-annotation',
                )
```

samuelcolvin: ye, probably good idea.

dmontagu: And during schema generation we just do 
```
return_type_schema = self.generate_schema(d.info.return_type)
```
in `pydantic._internal._generate_schema.GenerateSchema._computed_field_schema`

PR #5923, opened 2023-05-29
Hide __pydantic_*__ from @dataclass_transform

dmontagu:
I believe @tiangolo can confirm this behaves correctly.


tiangolo (approved):
Thank you! I confirm that this works. 🚀 

I pulled this branch, re-compiled/installed, confirmed it works, then checked out `main` again to confirm it is needed.

I also ran a model script to make sure it was running (although tests would catch that as well).

Issue #5919, opened 2023-05-29
V2 beta 1 outstanding issues


samuelcolvin:
Getting the beta out has taken longer than expected, but it's worth it - we're looking really good.

Still to do:
* [x] decide on https://github.com/pydantic/pydantic-core/pull/645, and test in pydantic
* [x] https://github.com/pydantic/pydantic-core/pull/646
* [x] #5920
* [x] https://github.com/pydantic/pydantic/issues/5914 - fixed in https://github.com/pydantic/pydantic-core/pull/651
* [x] https://github.com/pydantic/pydantic/issues/5918
* [x] Issue with `Config` reporting by @tiangolo while working on FastAPI #5921
* [x] I'd be more comfortable if we improved the coverage on `json_schema.py`, see [here](https://smokeshow.helpmanual.io/5f4g1p5r3y474i3o296d/d_2705a9c05af28909_json_schema_py.html) - currently it's bad, but probably requires @dmontagu to improve it as well as make sure the logic is correct
* [ ] What about the [discriminated union tests](https://github.com/pydantic/pydantic/blob/74539c9c0ff06016c1c058b658b57d855f66d7aa/tests/test_discriminated_union.py#L351-L354)  hat are `xfail`ing with enums - do we need to many any changes to enums/literals?
* [x] https://github.com/pydantic/pydantic-core/issues/649

samuelcolvin:
To think about after release:
* [ ] https://github.com/pydantic/speedate/issues/34

dmontagu:
Can we do [this](https://github.com/pydantic/pydantic/pull/5935/files#r1210388231) before the release?

Specifically, changing `from pydantic.annotated import GetCoreSchemaHandler` to `from pydantic.handlers import GetCoreSchemaHandler`. It seems @adriangb didn't mind the suggested change but didn't want to do in that PR. Making a note here before opening a PR for it.

hramezani:
Closing this one. I moved the remaining issues to https://github.com/pydantic/pydantic/issues/6096

PR #5934, opened 2023-05-30
Increase coverage for `json_schema.py`

Kludex:
- Related to https://github.com/pydantic/pydantic/issues/5919


pydantic/json_schema.py
```diff
@@ -222,16 +222,15 @@ def generate_definitions(
         returns the generated definitions paired with a mapping from the input keys to the definition references.
 
         Args:
-            inputs (Sequence[tuple[JsonSchemaKeyT, JsonSchemaMode, core_schema.CoreSchema]]): A sequence of tuples,
-                where:
+            inputs: A sequence of tuples, where:
```
Kludex, line 225: I took the opportunity to remove the types on the docstrings. I can add them back if you think they should be done separately.

pydantic/json_schema.py
```diff
@@ -741,7 +740,7 @@ def default_schema(self, schema: core_schema.WithDefaultSchema) -> JsonSchemaVal
             default = schema['default']
         elif 'default_factory' in schema:
             default = schema['default_factory']()
-        else:
+        else:  # pragma: no cover
```
Kludex, line 743: Should we use `assert_never` instead of the `ValueError`? Can this be reached?

lig: It could be also done like this
```py
if condition1:
elif condition2:
else:
    assert condition3
```
This gets the code to the full coverage if we have tests for all cases

tests/test_aliases.py
```diff
@@ -405,10 +405,16 @@ class Model(BaseModel):
 def test_aliases_json_schema():
     class Model(BaseModel):
         x: str = Field(alias='x_alias', validation_alias='x_val_alias', serialization_alias='x_ser_alias')
+        y: str = Field(validation_alias=AliasChoices('y_alias', 'another_alias'))
+        z: str = Field(validation_alias=AliasChoices(AliasPath('z_alias', 'even_another_alias'), 'and_another'))
```
lig, line 409: I's rather separated cases using `pytest.mark.parametrize`. This allows easier debugging and shows which case is failing exactly.

tests/test_types_typeddict.py
```diff
@@ -192,16 +194,73 @@ class Data(BaseModel):
...
+            schema = handler(source_type)
+            schema['computed_fields'] = [
+                core_schema.computed_field(property_name='another_b', return_schema=core_schema.int_schema())
+            ]
+            return schema
+
     class Model(BaseModel):
         data: Data
         data_td: DataTD
+        custom_td: CustomTD
```
lig, line 213: Here I would also suggest separating cases. Consider this example of using classes as `parametrize` arguments https://github.com/pydantic/pydantic/blob/bb79d99ad3e93e33997e0b0eda23bc3c52919ace/tests/test_root_model.py#L9

tests/test_validate_call.py
```diff
@@ -12,6 +12,7 @@
 from pydantic import Field, TypeAdapter, ValidationError, validate_call
 
 skip_pre_38 = pytest.mark.skipif(sys.version_info < (3, 8), reason='testing >= 3.8 behaviour only')
+skip_pre_39 = pytest.mark.skipif(sys.version_info < (3, 9), reason='testing >= 3.9 behaviour only')
```
lig, line 15: We have similar checks in other tests modules as well. I think, we could extract this useful syntax sugar into `conftest.py`.

lig (changes requested):
please update

tests/test_validate_call.py
lig: I'm not sure how to do it nicely though:)

pydantic/json_schema.py
Kludex:
```suggestion
        else:
```

tests/test_types_typeddict.py
Kludex: I'd have to call the fixture `TypedDict` directly if I want to do that - which I'm fine with the current one 👀 - but if you have suggestions...

Kludex:
please review

tests/test_types_typeddict.py
lig: It's up to you. If it looks too difficult improving right now let's keep it as is then.

lig (approved):
👍🏻 LGTM as it is

pyproject.toml
```diff
@@ -172,6 +172,7 @@ source = ['pydantic']
 omit = ['pydantic/deprecated/*', 'pydantic/v1/*']
 branch = true
 context = '${CONTEXT}'
+dynamic_context = "test_function"
```
Kludex, line 175:
```suggestion
```

pydantic/json_schema.py
```diff
@@ -628,14 +627,15 @@ def tuple_variable_schema(self, schema: core_schema.TupleVariableSchema) -> Json
...
         Returns:
             JsonSchemaValue: The generated JSON schema.
         """
-        json_schema: JsonSchemaValue = {'type': 'array', 'items': {}}
-        if 'items_schema' in schema:
-            json_schema['items'] = self.generate_inner(schema['items_schema'])
+        # NOTE: The `items_schema` is always added, even when we explicitly create a
+        # tuple variable schema without an `items_schema`.
+        items = self.generate_inner(schema['items_schema'])  # type: ignore
+        json_schema: JsonSchemaValue = {'type': 'array', 'items': items}
```
Kludex, line 638: We should revert this.

PR #5935, opened 2023-05-30
Update generic custom type example in `docs/usage/types/custom.md`

adriangb:
Selected Reviewer: @dmontagu

### DIFF
--- a/docs/usage/types/custom.md
+++ b/docs/usage/types/custom.md
@@ -192,53 +192,64 @@ If the Generic class that you are using as a sub-type has a classmethod
 Because you can declare validators that receive the current `field`, you can extract
 the `sub_fields` (from the generic class type parameters) and validate data with them.
 
-```py test="xfail - what do we do with generic custom types"
-from typing import Generic, TypeVar
+```py
+from dataclasses import dataclass
+from typing import Any, Generic, TypeVar
 
-from pydantic import BaseModel, ValidationError
-from pydantic.fields import ModelField
+from pydantic_core import CoreSchema, core_schema
+from typing_extensions import get_args, get_origin
+
+from pydantic import BaseModel, GetCoreSchemaHandler, ValidationError
 
 AgedType = TypeVar('AgedType')
 QualityType = TypeVar('QualityType')
 
 
 # This is not a pydantic model, it's an arbitrary generic class
+@dataclass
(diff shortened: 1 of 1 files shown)


docs/usage/types/custom.md
```diff
@@ -188,11 +188,13 @@ If the Generic class that you are using as a sub-type has a classmethod
...
-```py test="xfail - what do we do with generic custom types"
-from typing import Generic, TypeVar
+```py
+from typing import Any, Generic, TypeVar, get_args, get_origin
+
+from pydantic_core import CoreSchema, core_schema
 
 from pydantic import BaseModel, ValidationError
-from pydantic.fields import ModelField
+from pydantic.annotated import GetCoreSchemaHandler
```
dmontagu, line 197: Any chance we could change this import path before the beta release? If you prefer it I'll shut up but it just feels like `annotated` is a very confusing module name for it.

My preference would be just in `from pydantic import GetCoreSchemaHandler` but could do `pydantic.handlers` or something. Since the module only has `__all__ = 'GetJsonSchemaHandler', 'GetCoreSchemaHandler'` it seems `handlers` is a better name than `annotated`.

docs/usage/types/custom.md
```diff
@@ -255,11 +261,19 @@ model = Model(
     thing=TastingModel(name='Python', aged='Not much', quality='Awesome'),
 )
 print(model)
+"""
+wine=<__main__.TastingModel object at 0x0123456789ab> cheese=<__main__.TastingModel object at 0x0123456789ab> thing=<__main__.TastingModel object at 0x0123456789ab>
```
dmontagu, line 265: whoa how does it know to use that memory address? I was going to suggest changing the repr but looking at it and seeing there is a pattern makes me think something smart is happening here and it's useful for reinforcing that it's a custom class. 🤯

dmontagu (approved):

docs/usage/types/custom.md
adriangb: I don't know how it would work haha, I just changed it to a dataclass so the repr is nice now

docs/usage/types/custom.md
adriangb: Also let's leave this for another change

dmontagu:
please update

adriangb:
please review

PR #5936, opened 2023-05-30
Improve test coverage of pydantic._internal._core_utils

hramezani:
Not sure these tests are relevant. It might be better to move this function to `pydantic-core`

Selected Reviewer: @lig

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -7,11 +7,12 @@ from typing import Callable, Dict, Generic, List, NewType, Tuple, TypeVar, Union
 
 import pytest
 from dirty_equals import IsList
-from pydantic_core import PydanticCustomError
+from pydantic_core import PydanticCustomError, core_schema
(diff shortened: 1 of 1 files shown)


hramezani:
please review

lig:
please update

hramezani:
please review

hramezani:
> Should we be testing private modules like this, tho?
> 
> I'd think that we want to reach that part of the code via public API, which would also allow us more freedom to change underneath implementation (if needed). 🤔

Yeah, Best way is to reach that part via public API. but it's hard or maybe not possible. That's why we don't have coverage for some lines of these functions.

As I mentioned in PR description, we might need to move them to `pydantic-core` because these test make much more sense there

Kludex:
Based on your previous comment... Should we then add those tests here? Alternatively, we can add pragmas, and create an issue on `pydantic-core`s side.

lig (approved):

PR #5944, opened 2023-05-30
Pushdown config

adriangb:
This changes the config `model_config = ConfigDict()` on BaseModel and `__pydantic_config__ = ConfigDict()` on dataclasses and TypedDict) to "push down". This means that configs set on an outer model in the case of nested models get recursively pushed onto their children models as long as the child does not explicitly set the config param.

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -138,6 +138,11 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
     __init__.__pydantic_base_init__ = True  # type: ignore
 
     @classmethod
+    @property
+    def __pydantic_config__(cls) -> ConfigDict:
+        return cls.model_config
+
+    @classmethod
     def __get_pydantic_core_schema__(cls, __source: type[BaseModel], __handler: GetCoreSchemaHandler) -> CoreSchema:
         """Hook into generating the model's CoreSchema.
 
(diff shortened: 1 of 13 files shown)


pydantic/main.py
```diff
@@ -137,6 +137,11 @@ def __init__(__pydantic_self__, **data: Any) -> None:  # type: ignore
 
     __init__.__pydantic_base_init__ = True  # type: ignore
 
+    @classmethod
+    @property
```
adriangb, line 141: This was deprecated in 3.11+, need to figure out what to do.

adriangb:
@samuelcolvin this is as far as I can get. The main issue I've run into is that adding the config to the ref causes generics to break because [generics also use refs outside of `GenerateSchema`](https://github.com/pydantic/pydantic/blob/main/pydantic/_internal/_generics.py#L339-L368) so there's no way for them to track the "current ref".

adriangb:
If we remove this branch: https://github.com/pydantic/pydantic/blob/ad45332a3ecb5e35fe459ba0a5013a5b4cc6e4a8/pydantic/_internal/_core_utils.py#L121

All tests pass but we will not have resolved the issue of having a single ref with different configs. That would simply be a bug. To make it less likely, we could not push down configs for BaseModel, which IIRC is also the V1 behavior. Still, we should add a test to demonstrate the bug so we can try to fix it later.

PR #5949, opened 2023-05-31
🐛 Fix `RootModel` default values

### LINKED ISSUE #5903: Issues with `RootModel`
@lig in #5898 specifically 652ed996e64bff7fd4d22caf58dca14b68ef971c I found an issue with `RootModel` where it didn't support assignment.

I've fixed that issue in that PR, but I suspect there might be more issues with other slots attributes of `RootModel`, e.g. `__pydantic_private__` and `__pydantic_extra__`.

Please could you add some tests to check these attributes are all working correctly.

ideally we'd change `__slots__` to `__slots__ = 'root', '__pydantic_fields_set__', '__pydantic_extra__', '__pydantic_private__'` - maybe you tried that and found it didn't work.

You might also be able to override `__setattr__` on `RootModel` to be a lot simpler?

### LINKED ISSUE #5922: `default_factory` doesn't work in `RootModel`
### Initial Checks

### Description

Given the following `RootModel` with the root field having a default factory:
```python
class MyRootModel(pydantic.RootModel):
    root: UUID = Field(default_factory=uuid4)
```

Attempting to initialize the model without providing a value to the root field results in an exception:

```shell
In [1]: MyRootModel()
TypeError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 MyRootModel()

TypeError: __init__() missing 1 required positional argument: 'root'
```

### Example Code

_No response_

```Text
pydantic version: 2.0a4
        pydantic-core version: 0.37.0 release build profile
                 install path: /Users/thatguysimon/Library/Caches/pypoetry/virtualenvs/pydantic-v2-test-O79Hdn-W-py3.9/lib/python3.9/site-packages/pydantic
               python version: 3.9.16 (main, Dec  7 2022, 10:15:43)  [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.4-x86_64-i386-64bit
```

lig:
## Change Summary

Fix handling default values in RootModel

## Related issue number

Fix #5903
Fix #5922 

## Checklist

Selected Reviewer: @hramezani


lig:
Chaining this one after https://github.com/pydantic/pydantic/pull/5948

lig:
please review

tests/test_root_model.py
```diff
@@ -325,17 +326,44 @@ class Model(RootModel[int]):
...
-    assert m.model_dump == 42
+    assert m.model_dump() == 42
+
+
+def test_root_model_wrong_default_value_without_validate_default():
+    class Model(RootModel):
+        root: int = '42'
+
+    m = Model()
+    assert m.root == '42'
```
hramezani, line 343: As it is only one assertion and we don't need `m` after that.

```suggestion
    assert Model().root == '42'
```

up to you to change it or not

pydantic/main.py
```diff
@@ -1102,8 +1110,16 @@ class RootModel(BaseModel, typing.Generic[RootModelRootType]):
 
     root: RootModelRootType
 
-    def __init__(__pydantic_self__, root: RootModelRootType) -> None:  # type: ignore
+    def __init__(
+        __pydantic_self__, root: RootModelRootType | typing_extensions.Literal[_RootModelNoValue] = _RootModelNoValue
```
hramezani, line 1114:
```suggestion
        __pydantic_self__, root: RootModelRootType | Literal[_RootModelNoValue] = _RootModelNoValue
```

lig:
@hramezani please review

pydantic/main.py
```diff
@@ -1102,8 +1110,16 @@ class RootModel(BaseModel, typing.Generic[RootModelRootType]):
...
+    def __init__(
+        __pydantic_self__, root: RootModelRootType | Literal[_RootModelNoValue] = _RootModelNoValue
+    ) -> None:  # type: ignore
         __tracebackhide__ = True
+        if root is _RootModelNoValue:
+            root_field = __pydantic_self__.model_fields['root']
+            root = root_field.get_default(call_default_factory=True)
+            if not (__pydantic_self__.model_config.get('validate_default', False) or root_field.validate_default):
+                _object_setattr(__pydantic_self__, '__dict__', {'root': root})
+                return
```
hramezani, line 1122: I would return `None` here. up to you to accept:
```suggestion
                return None
```

hramezani (approved):
LGTM 👍

pydantic/main.py
lig: I think that in this case it's better to use the bare `return` as an indication that return value has no particular meaning as opposite to the example above:

```py
def odd_or_none(value: int) -> int | None:
    if value % 2 == 0:
        return None
    return value
```

In this example we explicitly return `None` as the result of the function is going to be used somewhere and `None` would mean something to that logic.

On the other hand, in `__init__` methods and any method that always returns `None` just by convention the bare `return` emphasis the fact that the result is not going to be used.

hramezani: Agree!

pydantic/main.py
```diff
@@ -1094,16 +1095,31 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
...
+        __pydantic_self__, root: RootModelRootType | Literal[_RootModelNoValue] = _RootModelNoValue
+    ) -> None:  # type: ignore
         __tracebackhide__ = True
-        # this is required set setattr works on RootModels, should be moved to rust
-        _object_setattr(__pydantic_self__, '__pydantic_fields_set__', {'root'})
+        if root is _RootModelNoValue:
+            root_field = __pydantic_self__.model_fields['root']
+            root = root_field.get_default(call_default_factory=True)
+            if not (__pydantic_self__.model_config.get('validate_default', False) or root_field.validate_default):
+                _object_setattr(__pydantic_self__, '__dict__', {'root': root})
```
samuelcolvin, line 1121: I generally prefer having an `else:` statement here.

pydantic/main.py
```diff
@@ -1094,16 +1095,31 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
...
 
-    def __init__(__pydantic_self__, root: RootModelRootType) -> None:  # type: ignore
+    def __init__(
+        __pydantic_self__, root: RootModelRootType | Literal[_RootModelNoValue] = _RootModelNoValue
+    ) -> None:  # type: ignore
         __tracebackhide__ = True
-        # this is required set setattr works on RootModels, should be moved to rust
-        _object_setattr(__pydantic_self__, '__pydantic_fields_set__', {'root'})
+        if root is _RootModelNoValue:
+            root_field = __pydantic_self__.model_fields['root']
```
samuelcolvin, line 1118: technically in this case we should have

```py
_object_setattr(__pydantic_self__, '__pydantic_fields_set__', set())
```

To match `BaseModel` logic.

samuelcolvin (changes requested):
otherwise LGTM.

When we have time, we should fix this in core - maybe create an issue on core now.

pydantic/main.py
lig: Right. Thank you. This is the place to fix it. I had a discussion with @dmontagu regarding this.

pydantic/main.py
lig: It's going to change with `__pydantic_fields_set__` fixed in this PR, I believe.

lig:
> When we have time, we should fix this in core - maybe create an issue on core now.

https://github.com/pydantic/pydantic-core/issues/654

lig:
please review

pydantic/main.py
```diff
@@ -1094,18 +1095,33 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
 RootModelRootType = typing.TypeVar('RootModelRootType')
 
 
+class _RootModelSpecialValues(enum.Enum):
+    no_value = enum.auto()
+
+
+_RootModelNoValue = _RootModelSpecialValues.no_value
```
dmontagu, line 1102: Could we just do `_RootModelNoValue = object()` and not introduce an enum for this? Or at least do the same thing we do with `_Undefined`:

https://github.com/pydantic/pydantic/blob/cf48b1a5a494604ec161ed85df6285d683084baa/pydantic/_internal/_fields.py#L39C7-L57

(We could also just use `pydantic._internal._fields.Undefined` for this directly?)

pydantic/main.py
```diff
@@ -1094,18 +1095,33 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
...
-    # was used
-    __pydantic_fields_set__ = {'root'}  # It's fine having a set here as it will never change
     __pydantic_private__ = None
     __pydantic_extra__ = None
 
     root: RootModelRootType
 
-    def __init__(__pydantic_self__, root: RootModelRootType) -> None:  # type: ignore
+    def __init__(
+        __pydantic_self__, root: RootModelRootType | Literal[_RootModelNoValue] = _RootModelNoValue
```
dmontagu, line 1113: I don't love that this exposes a private type in the signature, it's annoying if you want to subclass and override the init and keep the signature the same. But maybe this is the best alternative. @samuelcolvin I do think it might be nice to make `pydantic._internal._fields.Undefined` public for this and other use cases (such as introducing a "blessed" way to cause fields to not get serialized). I guess we can do that later though.

tests/test_root_model.py
```diff
@@ -325,17 +328,54 @@ class Model(RootModel[int]):
...
+
+def test_root_model_default_value_with_validate_default():
+    class Model(RootModel):
+        model_config = ConfigDict(validate_default=True)
+
+        root: int = '42'
+
+    m = Model()
+    assert m.root == 42
+    assert m.model_dump() == 42
```
dmontagu, line 366: Can you add an assertion for the `__pydantic_fields_set__` here and below?

pydantic/main.py
```diff
@@ -1094,18 +1095,33 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
...
+    ) -> None:  # type: ignore
         __tracebackhide__ = True
+        if root is not _RootModelNoValue:
+            _object_setattr(__pydantic_self__, '__pydantic_fields_set__', {'root'})
+        else:
+            _object_setattr(__pydantic_self__, '__pydantic_fields_set__', set())
+            root_field = __pydantic_self__.model_fields['root']
+            root = root_field.get_default(call_default_factory=True)
+            if not (__pydantic_self__.model_config.get('validate_default', False) or root_field.validate_default):
```
dmontagu, line 1122: Under what circumstances does this branch get hit? And does this match `BaseModel`?

Also, it seems to me we could do the `_object_setattr(__pydantic_self__, '__pydantic_fields_set__', set())` line in the else branch of this conditional, is that wrong? Maybe that's what @samuelcolvin was referring to below?

pydantic/main.py
lig: I like the idea to just use the same `Undefined` here. Thank you!

pydantic/main.py
lig: This gets hit by default. This is the branch for "do not validate default" which is the default behavior.

We want `__pydantic_fields_set__ == set()` for all `Undefined` regardless of `validate_default`. Thus, it looks logical to just set it right away for the entire `Undefined` branch.

tests/test_root_model.py
lig: Added and found a bug:)

lig:
@samuelcolvin @dmontagu please review

pydantic/main.py
lig: The `__pydantic_fields_set__` is being set in `__new__` since the last PR update.

samuelcolvin (changes requested):
This still isn't working correctly:

```py
from pydantic import RootModel, Field


class MyRootModel(RootModel):
    root: int = Field(42)


m = MyRootModel.model_validate_json('1')
print(m.model_fields_set)  # wrong

m = MyRootModel.model_validate()  # fails
```


Honestly, I think you need to fix thee issues with `__pydantic_fields_set__` in pydantic-core

pydantic/main.py
```diff
@@ -1096,16 +1096,33 @@ def _calculate_keys(self, *args: Any, **kwargs: Any) -> Any:
...
-    # TODO: Make `__pydantic_fields_set__` logic consistent with `BaseModel`, i.e. it should be `set()` if default value
-    # was used
-    __pydantic_fields_set__ = {'root'}  # It's fine having a set here as it will never change
     __pydantic_private__ = None
     __pydantic_extra__ = None
 
     root: RootModelRootType
 
-    def __init__(__pydantic_self__, root: RootModelRootType) -> None:  # type: ignore
+    def __new__(cls, root: RootModelRootType | _fields._UndefinedType = _fields.Undefined) -> RootModel:
```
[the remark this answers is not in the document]
lig: Well, it would but it takes time. I'll look into fixing it in `pydantic-core`, have no issue with that.

lig:
the logic has been moved to pydantic-core

please review

dmontagu (approved):
LGTM

PR #5950, opened 2023-05-31
Improve test coverage for pydantic.types

hramezani:
Selected Reviewer: @lig

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -425,7 +425,7 @@ class PathType:
     @staticmethod
     def validate_new(path: Path, _: core_schema.ValidationInfo) -> Path:
         if path.exists():
-            raise PydanticCustomError('path_exists', 'path already exists')
+            raise PydanticCustomError('path_exists', 'Path already exists')
         elif not path.parent.exists():
             raise PydanticCustomError('parent_does_not_exist', 'Parent directory does not exist')
         else:
@@ -498,9 +498,6 @@ class SecretField(Generic[SecretType]):
     def __eq__(self, other: Any) -> bool:
         return isinstance(other, self.__class__) and self.get_secret_value() == other.get_secret_value()
 
-    def __hash__(self) -> int:
-        return hash(self.get_secret_value())
-
     def __len__(self) -> int:
         return len(self._secret_value)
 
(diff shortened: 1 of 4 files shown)


pydantic/types.py
```diff
@@ -498,9 +498,6 @@ def __prepare_pydantic_annotations__(
     def __eq__(self, other: Any) -> bool:
         return isinstance(other, self.__class__) and self.get_secret_value() == other.get_secret_value()
 
-    def __hash__(self) -> int:
-        return hash(self.get_secret_value())
```
Kludex, line 502: Why is the hash removed?

pydantic/types.py
```diff
@@ -549,7 +546,7 @@ def serialize(
             # hence we just use `secret_display`
             return _secret_display(value.get_secret_value())
         else:
-            return value
+            return value  # pragma: no cover
```
hramezani, line 549: Seems this line is not reachable because the function is used only for `json`

https://github.com/pydantic/pydantic/blob/c18947586b237c66113ebd1bef7465210ab45772/pydantic/types.py#L581-L585

Kludex, line 549: Why this is not tested?

Kludex: This should be reachable somehow, otherwise the conditional doesn't make sense, right?

hramezani: I think it should. but I don't know how TBH.
I was about to remove the condition but as I am not fully aware I've decided to keep the condition and ignore the coverage.

pydantic/types.py
hramezani: I thought ‍‍‍`__hash__` is defined in `SecretStr` and `SecretBytes` which was not correct.
change reverted and tests added for `SecretStr` and `SecretBytes`

pydantic/types.py
Kludex: I think a way to test this would be to create a new type: `class NewSecret(SecretStr)`, and override the `__get_pydantic_core_schema__`, and replace the value you want.

Kludex:
please update

hramezani:
please review

lig (approved):
👍🏻

PR #5951, opened 2023-05-31
📦 Import `RootModel` dynamically when accessed directly only

lig:
## Change Summary

Implement mechanics for dynamic imports of public api exports and use it for `RootModel`.

## Related issue number

None

## Checklist

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/__init__.py
+++ b/pydantic/__init__.py
@@ -176,7 +176,7 @@ __all__ = [
 # A mapping of {<member name>: <module name>} defining dynamic imports
 _dynamic_imports = {'RootModel': '.root_model'}
 if typing.TYPE_CHECKING:
-    from .root_model import *
+    from .root_model import RootModel
 
 _getattr_migration = getattr_migration(__name__)
 
--- a/pydantic/__init__.py
+++ b/pydantic/__init__.py
@@ -1,3 +1,5 @@
+import typing
+
 import pydantic_core
 from pydantic_core.core_schema import (
     FieldSerializationInfo,
@@ -80,7 +82,6 @@ __all__ = [
     # main
     'BaseModel',
     'create_model',
-    'RootModel',
     # network
     'AnyUrl',
     'AnyHttpUrl',
@@ -101,6 +102,8 @@ __all__ = [
     'MySQLDsn',
     'MariaDBDsn',
     'validate_email',
+    # root_model
+    'RootModel',
     # tools
     'parse_obj_as',
     'schema_of',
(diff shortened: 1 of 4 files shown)


lig:
please review

pydantic/__init__.py
```diff
@@ -159,5 +162,18 @@
...
+# A mapping of {<member name>: <module name>} defining dynamic imports
+__all_dynamic__ = {'RootModel': '.root_model'}
+if typing.TYPE_CHECKING:
+    from .root_model import *
+
+__getattr_migration__ = getattr_migration(__name__)
+
+
+def __getattr__(attr_name: str) -> object:
+    if attr_name in __all_dynamic__:
```
samuelcolvin, line 174:
```suggestion
    attr = _dynamic_imports.get(attr_name
    if attr is None:
        return _getattr_migration(attr_name)
    else:
        ...
```
Better to look up `_dynamic_imports` once not twice.

samuelcolvin (commented):
otherwise LGTM.

Please update.

adriangb (approved):

lig:
please review

adriangb:
It's already approved, please update

lig:
I would like @samuelcolvin's approve as he has commented on this one

please review

pydantic/__init__.py
```diff
@@ -161,5 +164,20 @@
     'VERSION',
 ]
 
+# A mapping of {<member name>: <module name>} defining dynamic imports
+_dynamic_imports = {'RootModel': '.root_model'}
+if typing.TYPE_CHECKING:
+    from .root_model import *
```
dmontagu, line 170:
```suggestion
    from .root_model import RootModel
```
Any reason that we shouldn't do this? I'm okay with the `import *` if necessary just seems better since you can cmd+f etc.

lig: This is consistent with the approach we use for other non-dynamic imports. However, for dynamics imports we still need to define each one explicitly in `_dynamic_imports`. Thus, we can use explicit import here as using the wildcard import doesn't simplify the workflow of adding a new dynamic import much.

Go to definition is working as expected and I don't see a profit in allowing this line popping up in the search results for "RootModel" as the class is defined elsewhere.

IMO, the wildcard here follows the way static imports are being defined and makes an explicit reference to them by making it visually similar.

samuelcolvin: I think an explicit `from .root_model import RootModel` might help IDEs and static type checkers more, though I can't prove that for sure.

samuelcolvin (approved):
otherwise LGTM.

Please update

adriangb (approved):

PR #5963, opened 2023-06-01
Increase coverage for `json_schema.py`

Kludex:
- Related to #5919

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -864,11 +864,8 @@ class GenerateJsonSchema:
         return openapi_discriminator
 
     def chain_schema(self, schema: core_schema.ChainSchema) -> JsonSchemaValue:
-        try:
-            # Note: If we wanted to generate a schema for the _serialization_, would want to use the _last_ step:
-            return self.generate_inner(schema['steps'][0])
-        except IndexError as e:
-            raise ValueError('Cannot generate a JsonSchema for a zero-step ChainSchema') from e
+        # Note: If we wanted to generate a schema for the _serialization_, would want to use the _last_ step:
+        return self.generate_inner(schema['steps'][0])
 
     def lax_or_strict_schema(self, schema: core_schema.LaxOrStrictSchema) -> JsonSchemaValue:
         """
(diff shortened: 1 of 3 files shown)


pydantic/json_schema.py
```diff
@@ -864,11 +864,8 @@ def _extract_discriminator(
         return openapi_discriminator
 
     def chain_schema(self, schema: core_schema.ChainSchema) -> JsonSchemaValue:
-        try:
-            # Note: If we wanted to generate a schema for the _serialization_, would want to use the _last_ step:
-            return self.generate_inner(schema['steps'][0])
-        except IndexError as e:
-            raise ValueError('Cannot generate a JsonSchema for a zero-step ChainSchema') from e
```
Kludex, line 871: It's impossible to reach this, because a chain schema will always have more than zero steps. It's being validated on the rust side.

lig: an assert instead of try then?

Kludex: assert what?

Kludex: `assert len(schema['steps']) > 0`?

lig: Something like that, yes.

lig: Or just `assert schema['steps']` it's effectively the same.

Kludex: what's the goal of adding that assertion?

Kludex:
please review

pydantic/json_schema.py
lig: 1. We eliminating a separate logic branch that is introduced by `try..except`
2. We explicitly express our knowledge that there couldn't be an empty collection here
3. We ensure that in case something changed we will get into this assert.

In this particular case it looks like [3] doesn't help much as we will get IndexError anyway.

In terms of [2], the performance degradation isn't significant but it is there.

Maybe just removing `try..except` and leaving a comment instead of the assert is enough here.

lig (approved):
👍🏻

Kludex:
Gracias 🙏

pydantic/json_schema.py
Kludex: I think I didn't push the code with the comment, I'm going to add it on the next PR. 🤔

PR #5966, opened 2023-06-01
Increase coverage for `_generate_schema.py`

Kludex:
This PR adds more coverage to the `_generate_schema.py` file.

It also adds a new error code to notify the user when they are using the old `__modify_schema__`.

Selected Reviewer: @dmontagu

### DIFF
--- a/docs/usage/errors.md
+++ b/docs/usage/errors.md
@@ -71,6 +71,47 @@ foo = Foo(a={'b': {'a': None}})
 
 In other cases, the error message should indicate how to rebuild the class with the appropriate type defined.
 
+## Custom JSON Schema {#custom-json-schema}
+
+The `__modify_schema__` method is not supported in V2. You should use the `__get_pydantic_json_schema__` method instead.
+
+The `__modify_schema__` used to receive a single argument representing the JSON schema. See the example below:
+
+```py title="Old way"
+from pydantic import BaseModel, PydanticUserError
+
+try:
+
+    class Model(BaseModel):
+        @classmethod
+        def __modify_schema__(cls, field_schema):
+            field_schema.update(examples='examples')
+
+except PydanticUserError as exc_info:
+    assert exc_info.code == 'custom-json-schema'
+```
+
+The new method `__get_pydantic_json_schema__` receives two arguments: the first is a dictionary denoted as `CoreSchema`,
+and the second a callable `handler` that receives a `CoreSchema` as parameter, and returns a JSON schema. See the example
+below:
+
+```py title="New way"
+from pydantic_core import CoreSchema
+
+from pydantic import BaseModel
+from pydantic.json_schema import GetJsonSchemaHandler, JsonSchemaValue
+
+
+class Model(BaseModel):
+    @classmethod
+    def __get_pydantic_json_schema__(
+        cls, schema: CoreSchema, handler: GetJsonSchemaHandler
+    ) -> JsonSchemaValue:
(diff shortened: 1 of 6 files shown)


docs/usage/errors.md
```diff
@@ -71,6 +71,47 @@ foo = Foo(a={'b': {'a': None}})
 
 In other cases, the error message should indicate how to rebuild the class with the appropriate type defined.
 
+## Custom JSON Schema {#custom-json-schema}
```
Kludex, line 74: Can you check if this is fine, please? 🙏

pydantic/_internal/_generate_schema.py
```diff
@@ -1461,12 +1460,12 @@ def _extract_get_pydantic_json_schema(tp: Any, schema: CoreSchema) -> GetJsonSch
     """Extract `__get_pydantic_json_schema__` from a type, handling the deprecated `__modify_schema__`"""
     js_modify_function = getattr(tp, '__get_pydantic_json_schema__', None)
 
-    if js_modify_function is None and hasattr(tp, '__modify_schema__'):
-        warnings.warn(
-            'The __modify_schema__ method is deprecated, use __get_pydantic_json_schema__ instead',
-            DeprecationWarning,
+    if hasattr(tp, '__modify_schema__'):
```
Kludex, line 1463: @adriangb I've removed, as we talked. jfyk

docs/usage/errors.md
```diff
@@ -71,6 +71,48 @@ foo = Foo(a={'b': {'a': None}})
...
+
+The `__modify_schema__` method is not supported in V2. You should use the `__get_pydantic_json_schema__` method instead.
```
dmontagu, line 76:
```suggestion
The `__modify_schema__` method is no longer supported in V2. You should use the `__get_pydantic_json_schema__` method instead.
```

pydantic/_internal/_generate_schema.py
dmontagu: @Kludex I'm wondering if maybe it's a good idea to allow both being defined and if so ignore `__modify_schema__` — that way a type can maybe try to be compatible with both pydantic v1 and pydantic v2 at the same time. If that doesn't make sense then maybe this change is fine.

Also, originally we wanted to keep support for `__modify_schema__` and just deprecate it to make it easier to migrate. Did we change our mind on that?

dmontagu (approved):

pydantic/json_schema.py
```diff
@@ -863,6 +861,7 @@ def _extract_discriminator(
 
     def chain_schema(self, schema: core_schema.ChainSchema) -> JsonSchemaValue:
         # Note: If we wanted to generate a schema for the _serialization_, would want to use the _last_ step:
+        # There are always more than zero steps, since the ChainSchema is validated on the pydantic-core side.
```
adriangb, line 864: @dmontagu should we use the last item in the chain if we’re producing the serialization schema

adriangb (approved):

pydantic/_internal/_generate_schema.py
adriangb: @dmontagu I told Marcelo to just remove it since it seems like our attempt at backwards compatibility wasn’t being used. My intention was to just ignore the problem for now and circle back later to determine if we should add a backwards compatibility shim or not based on feedback in the beta. I’m open to adding that now provided we can make it work in most cases (it will definitely not work in cases like generics, which is the main argument to not try at all)

PR #5978, opened 2023-06-02
Improve test coverage of pydantic._internal._typing_extra

hramezani:

### DIFF
--- a/pydantic/_internal/_typing_extra.py
+++ b/pydantic/_internal/_typing_extra.py
@@ -280,13 +280,7 @@ if sys.version_info < (3, 9):
 
         Implemented as EAFP with memory.
         """
-        global _make_forward_ref
-        try:
-            res = typing.ForwardRef(arg, is_argument, is_class=is_class)  # type: ignore
-            _make_forward_ref = typing.ForwardRef  # type: ignore
-            return res
-        except TypeError:
-            return typing.ForwardRef(arg, is_argument)
+        return typing.ForwardRef(arg, is_argument)
 
 else:
     _make_forward_ref = typing.ForwardRef
--- a/tests/test_typing.py
+++ b/tests/test_typing.py
@@ -1,13 +1,22 @@
 import sys
 import typing
 from collections import namedtuple
-from typing import Callable, NamedTuple
+from typing import Callable, ClassVar, ForwardRef, NamedTuple
 
 import pytest
 from typing_extensions import Literal, get_origin
 
 from pydantic import Field  # noqa: F401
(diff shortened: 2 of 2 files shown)


hramezani:
please review

pydantic/_internal/_typing_extra.py
```diff
@@ -283,8 +283,10 @@ def _make_forward_ref(
         global _make_forward_ref
         try:
             res = typing.ForwardRef(arg, is_argument, is_class=is_class)  # type: ignore
-            _make_forward_ref = typing.ForwardRef  # type: ignore
-            return res
+            # we have to ignore the following two lines from coverage because
+            # `is_class` does not exist in 3.8 and it always throws `TypeError`
+            _make_forward_ref = typing.ForwardRef  # type: ignore # pragma: no cover
+            return res  # pragma: no cover
```
Kludex, line 289: Then why do we have this logic? This is only for <=3.8 anyway 🤔

hramezani: Good catch 🎯 
After talking with @adriangb, I've changed it.

pydantic/_internal/_typing_extra.py
```diff
@@ -280,13 +280,7 @@ def _make_forward_ref(
...
-        try:
-            res = typing.ForwardRef(arg, is_argument, is_class=is_class)  # type: ignore
-            _make_forward_ref = typing.ForwardRef  # type: ignore
-            return res
-        except TypeError:
-            return typing.ForwardRef(arg, is_argument)
+        return typing.ForwardRef(arg, is_argument)
 
 else:
     _make_forward_ref = typing.ForwardRef
```
adriangb, line 286: Do we even need this if/else clause then?

pydantic/_internal/_typing_extra.py
adriangb: For the record: we tracked down the origin of this logic to https://github.com/pydantic/pydantic/commit/594effa279668bd955e98f1cd5c036b37d3bbd40, so it's been there from the start. That was huge commit so it's highly likely this was never needed and we've just carried it forward.

pydantic/_internal/_typing_extra.py
hramezani: I think we need because we are using `_make_forward_ref` with `is_class` in some places and `is_class` is not available in `python < 3.9`

Did I miss something?

adriangb: Gotcha yes makes sense 👍🏻

Kludex (approved):

PR #5983, opened 2023-06-02
Add from_attributes to TypeAdapter.validate_python

adriangb:
cc @tiangolo

### DIFF
--- a/pydantic/type_adapter.py
+++ b/pydantic/type_adapter.py
@@ -166,20 +166,28 @@ class TypeAdapter(Generic[T]):
         self.validator = validator
         self.serializer = serializer
 
-    def validate_python(self, __object: Any, *, strict: bool | None = None, context: dict[str, Any] | None = None) -> T:
+    def validate_python(
+        self,
+        __object: Any,
+        *,
+        strict: bool | None = None,
+        from_attributes: bool | None = None,
+        context: dict[str, Any] | None = None,
+    ) -> T:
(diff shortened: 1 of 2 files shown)


pydantic/type_adapter.py
```diff
@@ -166,20 +166,28 @@ def __init__(self, type: Any, *, config: ConfigDict | None = None, _parent_depth
...
+        from_attributes: bool | None = None,
+        context: dict[str, Any] | None = None,
+    ) -> T:
         """
         Validate a Python object against the model.
 
         Args:
             __object (Any): The Python object to validate against the model.
             strict (bool | None, optional): Whether to strictly check types. Defaults to None.
+            from_attributes (bool | None, optional): Weather to extract data from object attributes. Defaults to None.
```
tiangolo, line 183:
```suggestion
            from_attributes (bool | None, optional): Whether to extract data from object attributes. Defaults to None.
```

tiangolo (approved):
Nice, thanks! :rocket:

samuelcolvin:
Let's add to `BaseModel` methods too while we're here.

adriangb:
This is already merged. Happy to add that tomorrow if you think it’s really needed, otherwise I was planning on holding off until someone requests it.

samuelcolvin:
🤦 not the first time I've missed the pr is merged.

I think worth adding to base model before we make another release.

Issue #5980, opened 2023-06-02
SchemaError when multiple computed fields return same model class instance


hateyouinfinity:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

When one defines a model that has multiple computed fields that have the same type, that type being another pydantic model, a `SchemaError` is raised:
```python
File ~\AppData\Local\Programs\Python\Python310\lib\site-packages\pydantic\_internal\_model_construction.py:176, in ModelMetaclass.__new__(mcs, cls_n
ame, bases, namespace, __pydantic_generic_metadata__, __pydantic_reset_parent_namespace__, **kwargs)
    174 types_namespace = get_cls_types_namespace(cls, parent_namespace)
    175 set_model_fields(cls, bases, config_wrapper, types_namespace)
--> 176 complete_model_class(
    177     cls,
    178     cls_name,
    179     config_wrapper,
    180     raise_errors=False,
    181     types_namespace=types_namespace,
    182 )
    183 # using super(cls, cls) on the next line ensures we only call the parent class's __pydantic_init_subclass__
    184 # I believe the `type: ignore` is only necessary because mypy doesn't realize that this code branch is
    185 # only hit for _proper_ subclasses of BaseModel
    186 super(cls, cls).__pydantic_init_subclass__(**kwargs)  # type: ignore[misc]

File ~\AppData\Local\Programs\Python\Python310\lib\site-packages\pydantic\_internal\_model_construction.py:408, in complete_model_class(cls, cls_nam
e, config_wrapper, raise_errors, types_namespace)
    406 simplified_core_schema = inline_schema_defs(schema)
    407 cls.__pydantic_validator__ = SchemaValidator(simplified_core_schema, core_config)
--> 408 cls.__pydantic_serializer__ = SchemaSerializer(simplified_core_schema, core_config)
    409 cls.__pydantic_complete__ = True
    411 # set __signature__ attr only for model class, but not for its instances

SchemaError: Error building `model` serializer:
  SchemaError: Error building `model-fields` serializer:
  SchemaError: Computed field `comp2`:
  SchemaError: Duplicate ref: `__main__.CompModel:2066165399728`
```

No errors on 2.0a4, only on 2.0b1.

### Example Code

```Python
from pydantic import BaseModel, computed_field


class CompModel(BaseModel):
    val: int


class MyModel(BaseModel):
    val: int = 1

    @computed_field
    @property
[...]

PR #5987, opened 2023-06-03
Fix SchemaError when multiple computed fields point to the same type

adriangb:
Fixes https://github.com/pydantic/pydantic/issues/5980

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -250,8 +250,8 @@ class _WalkCoreSchema:
         return f(schema.copy(), self._walk)
 
     def _walk(self, schema: core_schema.CoreSchema, f: Walk) -> core_schema.CoreSchema:
-        ser_schema: core_schema.SerSchema | None = schema.get('serialization', None)  # type: ignore
         schema = self._schema_type_to_method[schema['type']](schema, f)
+        ser_schema: core_schema.SerSchema | None = schema.get('serialization', None)  # type: ignore
         if ser_schema:
             schema['serialization'] = self._handle_ser_schemas(ser_schema.copy(), f)
         return schema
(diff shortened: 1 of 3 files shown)


adriangb:
please review

pydantic/_internal/_core_utils.py
```diff
@@ -250,8 +250,8 @@ def walk(self, schema: core_schema.CoreSchema, f: Walk) -> core_schema.CoreSchem
         return f(schema.copy(), self._walk)
 
     def _walk(self, schema: core_schema.CoreSchema, f: Walk) -> core_schema.CoreSchema:
-        ser_schema: core_schema.SerSchema | None = schema.get('serialization', None)  # type: ignore
         schema = self._schema_type_to_method[schema['type']](schema, f)
+        ser_schema: core_schema.SerSchema | None = schema.get('serialization', None)  # type: ignore
```
adriangb, line 254: Ups not technically necessary in the end but this ordering is better anyway.

pydantic/_internal/_core_utils.py
```diff
@@ -366,6 +366,13 @@ def handle_model_fields_schema(self, schema: core_schema.ModelFieldsSchema, f: W
         if 'extra_validator' in schema:
             schema['extra_validator'] = self.walk(schema['extra_validator'], f)
         replaced_fields: dict[str, core_schema.ModelField] = {}
+        replaced_computed_fields: list[core_schema.ComputedField] = []
+        for computed_field in schema.get('computed_fields', None) or ():
+            replaced_field = computed_field.copy()
+            replaced_field['return_schema'] = self.walk(computed_field['return_schema'], f)
+            replaced_computed_fields.append(replaced_field)
+        if replaced_computed_fields:
+            schema['computed_fields'] = replaced_computed_fields
```
Kludex, line 375: Is `BaseModel` the only class that takes the `@computed_field`s in consideration?

Kludex (approved):

pydantic/_internal/_core_utils.py
adriangb: Good point, added tests.

Issue #5981, opened 2023-06-02
KeyError: 'items_schema' when using pydantic.TypeAdapter(tuple).json_schema()


commonism:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

I still rely on the glue provided in #5381, I made it use TypeAdapter.
`now` it fails for "tuple".

### Example Code

```Python
import pydantic

pydantic.TypeAdapter(tuple).json_schema()
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0b1
        pydantic-core version: 0.38.0 release build profile
                 install path: ~/workspace/pydantic/pydantic
               python version: 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]
                     platform: Linux-5.19.0-42-generic-x86_64-with-glibc2.35
     optional deps. installed: ['email-validator', 'typing-extensions']
```

adriangb:
Talk about MRE, thank you!

PR #5988, opened 2023-06-03
Fix json schema for single empty tuple

adriangb:
Fixes https://github.com/pydantic/pydantic/issues/5981

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -630,8 +630,6 @@ class GenerateJsonSchema:
         Returns:
             JsonSchemaValue: The generated JSON schema.
         """
-        # NOTE: The `items_schema` is always added, even when we explicitly create a
-        # tuple variable schema without an `items_schema`.
         items_schema = {} if 'items_schema' not in schema else self.generate_inner(schema['items_schema'])
         json_schema = {'type': 'array', 'items': items_schema}
         self.update_with_validations(json_schema, schema, self.ValidationsMapping.array)
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -632,8 +632,11 @@ class GenerateJsonSchema:
         """
         # NOTE: The `items_schema` is always added, even when we explicitly create a
         # tuple variable schema without an `items_schema`.
-        items = self.generate_inner(schema['items_schema'])  # type: ignore
-        json_schema: JsonSchemaValue = {'type': 'array', 'items': items}
+        if 'items_schema' in schema:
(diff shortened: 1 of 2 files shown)


adriangb:
please review

pydantic/json_schema.py
```diff
@@ -632,8 +632,8 @@ def tuple_variable_schema(self, schema: core_schema.TupleVariableSchema) -> Json
         """
         # NOTE: The `items_schema` is always added, even when we explicitly create a
         # tuple variable schema without an `items_schema`.
```
Kludex, line 634:
```suggestion
```

pydantic/json_schema.py
```diff
@@ -632,8 +632,8 @@ def tuple_variable_schema(self, schema: core_schema.TupleVariableSchema) -> Json
...
-        items = self.generate_inner(schema['items_schema'])  # type: ignore
-        json_schema: JsonSchemaValue = {'type': 'array', 'items': items}
+        items_schema = {} if 'items_schema' not in schema else self.generate_inner(schema['items_schema'])
+        json_schema = {'type': 'array', 'items': items_schema}
```
Kludex, line 636: This is my bad. I broke it just before the release here: https://github.com/pydantic/pydantic/pull/5934/files#r1215343647

We can just revert it
```suggestion
        json_schema: JsonSchemaValue = {'type': 'array', 'items': {}}
        if 'items_schema' in schema:
            json_schema['items'] = self.generate_inner(schema['items_schema'])
```

Kludex (changes requested):

pydantic/json_schema.py
Kludex: oh, I just noticed it's the same code 👀 

Anyway... suggestion still applies as it looks more readable

adriangb: The new version is the same as the list version. Let’s try to avoid delaying work for minor things like this.

adriangb:
please review

pydantic/json_schema.py
Kludex: I don't see how a suggestion would make that effect...

Kludex (approved):

pydantic/json_schema.py
adriangb: This could have been merged yesterday. In these situations I usually like to approve *and* leave nits/suggestions so the author can decide if they want to just merge, fix a couple and merge, or fix all them and merge. But we avoid a 24 hour delay of another round of review.

Issue #5998, opened 2023-06-04
Field serializers clobber serialization for the field type's schema


adriangb:
```python
from pydantic import BaseModel, field_serializer


class User(BaseModel):
    y: str


class Model(BaseModel):
    user1: User
    user2: User

    @field_serializer('user1')
    def serialize_user1(self, v: User):
        return f'User1({v.y})'

    @field_serializer('user2')
    def serialize_user2(self, v: User):
        return f'User2({v.y})'

m = Model(user1=User(y='1'), user2=User(y='2'))
# fails, actual: {'user1': 'User2(1)', 'user2': 'User2(2)'}
assert m.model_dump() == {'user1': 'User1(1)', 'user2': 'User2(2)'} 
```

PR #6007, opened 2023-06-05
Allow applying serializers and constraints to named/ref schemas

adriangb:
Fixes https://github.com/pydantic/pydantic/issues/5998

This makes some fundamental changes to core schema and json schema building.
In particular:
- Any schemas with a `"ref"` or `"$ref"` we immediately stick into `"definitions"` or `"$defs"` respectively. I added a `resolve_ref_schema` method to `GetCoreSchemaHandler` that is analogous to the one we already had in `GetJsonSchemaHandler`. This means that implementing `__get_pydantic_core_schema__` can be more complex because you're forced to deal with refs explicitly, but it resolves several inconsistencies and bugs.
- I was able to make automatic unpacking of `"ref"` schemas work for the source type itself, which means that if you implement `__get_pydantic_core_schema__` on `Model(BaseModel)` and call `handler(source_type)` you don't get back a `"ref"` schema, we unpack it for you because a source type should edit its own schema in-place.
- We no longer "automatically unpack" `"$ref"` JSON schemas. To avoid making JSON schema generation more difficult we were previously automatically unpacking `"$ref"` for every handler call, but this was complex and led to inconsistencies the same as mentioned above. This leads to regression for the case of "source types" because as per above those should have the `"ref"` unpacked for them. Fixing this will require some more work that I would like to defer from this PR. In particular, when we generate the JSON schema now we just have a list of functions to call, we have no way of knowing if they came from metadata in `Annotated` or `__get_pydantic_json_schema__` on the source type itself.

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_annotated_handlers.py
+++ b/pydantic/_internal/_annotated_handlers.py
@@ -98,7 +98,7 @@ class GetCoreSchemaHandler:
             __maybe_ref_schema: A `CoreSchema`, `ref`-based or not.
 
         Raises:
-            LookupError: if the ref is not found.
+            LookupError: If the `ref` is not found.
 
         Returns:
             A concrete `CoreSchema`.
--- a/pydantic/_internal/_annotated_handlers.py
+++ b/pydantic/_internal/_annotated_handlers.py
@@ -101,6 +101,6 @@ class GetCoreSchemaHandler:
             LookupError: if the ref is not found.
 
         Returns:
-            CoreSchema: A concrete CoreSchema.
+            A concrete `CoreSchema`.
         """
         raise NotImplementedError
--- a/pydantic/_internal/_annotated_handlers.py
+++ b/pydantic/_internal/_annotated_handlers.py
@@ -95,7 +95,7 @@ class GetCoreSchemaHandler:
         This means you don't have to check before calling this function.
 
         Args:
-            __maybe_ref_schema: A CoreSchema, ref based or not.
+            __maybe_ref_schema: A `CoreSchema`, `ref`-based or not.
 
         Raises:
             LookupError: if the ref is not found.
--- a/pydantic/_internal/_annotated_handlers.py
+++ b/pydantic/_internal/_annotated_handlers.py
@@ -88,3 +88,19 @@ class GetCoreSchemaHandler:
             CoreSchema: the `pydantic-core` CoreSchema generated.
         """
         raise NotImplementedError
+
+    def resolve_ref_schema(self, __maybe_ref_schema: core_schema.CoreSchema) -> core_schema.CoreSchema:
+        """Get the real schema for a `definition-ref` schema.
+        If the schema given is not a `definition-ref` schema, it will be returned as is.
+        This means you don't have to check before calling this function.
+
+        Args:
+            __maybe_ref_schema: A CoreSchema, ref based or not.
+
+        Raises:
+            LookupError: if the ref is not found.
+
+        Returns:
+            CoreSchema: A concrete CoreSchema.
+        """
+        raise NotImplementedError
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1499,7 +1499,8 @@ class _Definitions:
 
         This should be called for any type that can be identified by reference.
         This includes any recursive types.
-        At present the following types can be named / recursive:
+        At present the following types can be named/recursive:
+        
         - BaseModel
         - Dataclasses
         - TypedDict
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1499,6 +1499,7 @@ class _Definitions:
 
         This should be called for any type that can be identified by reference.
         This includes any recursive types.
+        
         At present the following types can be named/recursive:
         
         - BaseModel
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1499,9 +1499,9 @@ class _Definitions:
 
         This should be called for any type that can be identified by reference.
         This includes any recursive types.
-        
+
         At present the following types can be named/recursive:
-        
+
         - BaseModel
         - Dataclasses
         - TypedDict
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1492,7 +1492,7 @@ class _Definitions:
         Get a definition for `tp` if one exists.
 
         If a definition exists, a tuple of `(ref_string, CoreSchema)` is returned.
-        If no definition exists yet a tuple of (ref_string, None) is returned.
+        If no definition exists yet, a tuple of `(ref_string, None)` is returned.
 
         Note that the returned CoreSchema will always be a DefinitionReferenceSchema,
         not the actual definition itself.
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1491,7 +1491,7 @@ class _Definitions:
         """
         Get a definition for `tp` if one exists.
 
-        If a definition exists a tuple of (ref_string, CoreSchema) is returned.
+        If a definition exists, a tuple of `(ref_string, CoreSchema)` is returned.
         If no definition exists yet a tuple of (ref_string, None) is returned.
 
         Note that the returned CoreSchema will always be a DefinitionReferenceSchema,
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -164,33 +164,6 @@ def modify_model_json_schema(
     return json_schema
 
 
-class Definitions:
-    def __init__(self) -> None:
-        self.seen: set[str] = set()
-        self.definitions: dict[str, core_schema.CoreSchema] = {}
-
-    @contextmanager
-    def get_schema_or_ref(self, cls: Any) -> Iterator[tuple[str, None] | tuple[str, CoreSchema]]:
-        ref = get_type_ref(cls)
-        if ref in self.seen or ref in self.definitions:
-            yield (ref, core_schema.definition_reference_schema(ref))
-        else:
-            self.seen.add(ref)
-            try:
-                yield (ref, None)
-            finally:
-                self.seen.discard(ref)
-
-
-def resolve_original_schema(schema: CoreSchema, definitions: dict[str, CoreSchema]) -> CoreSchema | None:
-    if schema['type'] == 'definition-ref':
-        return definitions.get(schema['schema_ref'], None)
-    elif schema['type'] == 'definitions':
-        return schema['schema']
-    else:
-        return schema
-
-
 class GenerateSchema:
     __slots__ = '_config_wrapper_stack', 'types_namespace', 'typevars_map', 'recursion_cache', 'definitions', 'defs'
 
@@ -205,7 +178,7 @@ class GenerateSchema:
         self.types_namespace = types_namespace
         self.typevars_map = typevars_map
 
-        self.defs = Definitions()
+        self.defs = _Definitions()
 
     @property
     def config_wrapper(self) -> ConfigWrapper:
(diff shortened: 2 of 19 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -1200,20 +1214,51 @@ def inner_handler(obj: Any) -> CoreSchema:
...
+            return schema
+
+        if schema['type'] == 'definition-ref':
+            ref = schema['schema_ref']
+            if ref in self.defs.definitions:
+                schema = self.defs.definitions[ref].copy()
+                new_ref = ref + f'_{repr(metadata)}'
+                if new_ref in self.defs.definitions:
+                    return self.defs.definitions[new_ref]
+                schema['ref'] = new_ref  # type: ignore
```
adriangb, line 1254: This is the meat and potatoes, and also the maybe most jank part

pydantic/_internal/_generate_schema.py
```diff
@@ -245,68 +268,70 @@ def model_schema(self, cls: type[BaseModel]) -> core_schema.CoreSchema:
...
-        fields = cls.model_fields
-        decorators = cls.__pydantic_decorators__
-        check_decorator_fields_exist(
-            chain(
-                decorators.field_validators.values(),
-                decorators.field_serializers.values(),
-                decorators.validators.values(),
-            ),
-            fields.keys(),
-        )
```
adriangb, line 261: This diff isn't actually as bad as it seems, it's coming form indenting the body of all of these functions 1 more tab/4 spaces because I added a context manager at the top of every one.

docs/usage/schema.md
```diff
@@ -669,6 +658,17 @@ print(json.dumps(top_level_schema, indent=2))
...
+      "type": "object",
+      "properties": {
+        "a": {
+          "type": "string",
+          "default": null,
+          "title": "A"
+        }
+      },
+      "title": "Foo"
+    },
```
adriangb, line 671: See https://github.com/pydantic/pydantic/issues/6013

tests/test_json_schema.py
```diff
@@ -3497,41 +3497,35 @@ class Model(BaseModel):
...
-            'number': {'title': 'Number', 'type': 'integer'},
-            'pet': {
-                'discriminator': {
-                    'mapping': {'cat': '#/$defs/Cat', 'dog': '#/$defs/Dog'},
-                    'propertyName': 'typeOfPet',
-                },
-                'oneOf': [{'$ref': '#/$defs/Cat'}, {'$ref': '#/$defs/Dog'}],
-                'title': 'Pet',
-            },
+            'pet': {'oneOf': [{'$ref': '#/$defs/Cat'}, {'$ref': '#/$defs/Dog'}], 'title': 'Pet'},
```
adriangb, line 3504: Seems like we lost the discriminator here, that's wrong

tests/test_root_model.py
```diff
@@ -43,10 +43,6 @@ class InnerModel(BaseModel):
 def test_root_model_specialized(root_type, root_value, dump_value):
     Model = RootModel[root_type]
 
-    assert Model.__pydantic_core_schema__['type'] == 'model'
-    assert Model.__pydantic_core_schema__['root_model'] is True
-    assert Model.__pydantic_core_schema__['custom_init'] is False
```
adriangb, line 48: Removing these since they're making assumptions that we don't necessarily promise and we shouldn't be testing implementation details like `root_model=True`.

tests/test_types_typeddict.py
```diff
@@ -211,33 +211,99 @@ class Model(BaseModel):
...
+                'properties': {'a': {'type': 'integer', 'title': 'A'}},
+                'required': ['a'],
                 'title': 'Data',
-                'properties': {'a': {'title': 'A', 'type': 'integer'}},
+            },
+        },
+    }
+
+    # insert_assert(Model.model_json_schema(mode='serialization'))
+    assert Model.model_json_schema(mode='serialization') == {
```
adriangb, line 247: This needs to be reviewed carefully.

pydantic/_internal/_generate_schema.py
```diff
@@ -1200,20 +1211,63 @@ def inner_handler(obj: Any) -> CoreSchema:
...
+        if isinstance(metadata, FieldInfo):
+            for field_metadata in metadata.metadata:
+                schema = self.apply_single_annotation(schema, field_metadata)
+            if metadata.discriminator is not None:
+                schema = _discriminated_union.apply_discriminator(
+                    schema,
+                    metadata.discriminator,
+                    {
+                        **{s: core_schema.definition_reference_schema(s) for s in self.defs.seen},
+                        **self.defs.definitions,
```
adriangb, line 1231: This feels kinda hacky

tests/test_root_model.py
dmontagu: I think it's maybe worth keeping the `custom_init` one as there's a little trick to get this to be False, the others I agree can be removed.

pydantic/_internal/_known_annotated_metadata.py
```diff
@@ -140,10 +140,12 @@ def val_func(v: Any) -> Any:
...
-
-    # for all other annotations just update the schema
-    # this includes things like `strict` which apply to pretty much every schema
-    schema.update(schema_update)  # type: ignore
+    elif schema_update:
+        # for all other annotations just update the schema
+        # this includes things like `strict` which apply to pretty much every schema
+        schema.update(schema_update)  # type: ignore
+    else:
+        return None
```
adriangb, line 148:
```suggestion
    else:
        # no update was made
        return None
```

pydantic/_internal/_schema_generation_shared.py
```diff
@@ -12,64 +12,12 @@
...
-    """
-    A GetJsonSchemaHandler implementation that automatically unpacks `$ref`
-    schemas so that the caller doesn't have to worry about that.
-
-    This is used for custom types and models that implement `__get_pydantic_core_schema__`
-    so they always get a `non-$ref` schema.
-
-    Used internally by Pydantic, please do not rely on this implementation.
-    See `GetJsonSchemaHandler` for the handler API.
-    """
```
adriangb, line 30: This functionality was half broken before and is fully broken now. We're just going to have to face the fact that if you are messing with schemas you'll have to deal with refs.

tests/test_json.py
```diff
@@ -80,6 +80,7 @@ def test_json_serialization(ser_type, gen_value, json_output):
     config_wrapper = ConfigWrapper({'arbitrary_types_allowed': False})
     gen = GenerateSchema(config_wrapper, None)
     schema = gen.generate_schema(ser_type)
+    schema = gen.collect_definitions(schema)
     serializer = SchemaSerializer(schema)
     assert serializer.to_json(gen_value()) == json_output
```
adriangb, line 85: Seems like this could be replaced by TypeAdapter

adriangb (commented):
Gave the diff a look over. Generally, this looks right to me. I would suggest waiting for #6043 since I think that will minimize the diff in several places.

adriangb:
please review

pydantic/_internal/_annotated_handlers.py
```diff
@@ -88,3 +88,19 @@ def generate_schema(self, __source_type: Any) -> core_schema.CoreSchema:
...
+        This means you don't have to check before calling this function.
+
+        Args:
+            __maybe_ref_schema: A CoreSchema, ref based or not.
+
+        Raises:
+            LookupError: if the ref is not found.
+
+        Returns:
+            CoreSchema: A concrete CoreSchema.
```
tpdorsey, line 104:
```suggestion
            A concrete `CoreSchema`.
```

tpdorsey: Specifying the return type is not required. mkdocstrings infers it from the signature.

pydantic/_internal/_annotated_handlers.py
```diff
@@ -88,3 +88,19 @@ def generate_schema(self, __source_type: Any) -> core_schema.CoreSchema:
...
         """
         raise NotImplementedError
+
+    def resolve_ref_schema(self, __maybe_ref_schema: core_schema.CoreSchema) -> core_schema.CoreSchema:
+        """Get the real schema for a `definition-ref` schema.
+        If the schema given is not a `definition-ref` schema, it will be returned as is.
+        This means you don't have to check before calling this function.
+
+        Args:
+            __maybe_ref_schema: A CoreSchema, ref based or not.
```
tpdorsey, line 98:
```suggestion
            __maybe_ref_schema: A `CoreSchema`, `ref`-based or not.
```

pydantic/_internal/_annotated_handlers.py
```diff
@@ -88,3 +88,19 @@ def generate_schema(self, __source_type: Any) -> core_schema.CoreSchema:
...
+
+        Raises:
+            LookupError: if the ref is not found.
```
tpdorsey, line 101:
```suggestion
            LookupError: If the `ref` is not found.
```

tpdorsey (commented):
Minor suggested docstring edits.

pydantic/_internal/_generate_schema.py
```diff
@@ -166,8 +164,35 @@ def modify_model_json_schema(
     return json_schema
 
 
+class Definitions:
```
samuelcolvin, line 167: Please can we stick to a "newpaper style" in terms of order of types added to modules - public objects should come first and supporting object should be defined later in the module.

I know we're already broken this rule in this file a log, but I'd rather add this near the end.

Could we add a docstring explaining what this does?

pydantic/_internal/_generate_schema.py
```diff
@@ -166,8 +164,35 @@ def modify_model_json_schema(
...
+    def __init__(self) -> None:
+        self.seen: set[str] = set()
+        self.definitions: dict[str, core_schema.CoreSchema] = {}
+
+    @contextmanager
+    def get_schema_or_ref(self, cls: Any) -> Iterator[tuple[str, None] | tuple[str, CoreSchema]]:
```
samuelcolvin, line 173: docstring  please.

pydantic/_internal/_generate_schema.py
```diff
@@ -224,86 +257,78 @@ def _generate_schema_for_type(
...
+
+        return remove_unnecessary_invalid_definitions(schema)
+
+    def _model_schema(self, cls: type[BaseModel]) -> core_schema.CoreSchema:
+        """Generate schema for a Pydantic model."""
+        with self.defs.get_schema_or_ref(cls) as (model_ref, maybe_schema):
+            if maybe_schema is not None:
+                return maybe_schema
+
+            fields = cls.model_fields
```
samuelcolvin, line 273: I'm assume most of this is just an indent change, so I'm not reviewing it.

samuelcolvin (approved):
just a few small comments here.

Overall I think this looks good, but I do think it's worth waiting for @dmontagu to review this one more time - as you say, this stuff is pretty complicated and David wrote much of it.

pydantic/_internal/_generate_schema.py
```diff
@@ -1504,3 +1477,50 @@ def _common_field(
...
+    def __init__(self) -> None:
+        self.seen: set[str] = set()
+        self.definitions: dict[str, core_schema.CoreSchema] = {}
+
+    @contextmanager
+    def get_schema_or_ref(self, tp: Any) -> Iterator[tuple[str, None] | tuple[str, CoreSchema]]:
+        """
+        Get a definition for `tp` if one exists.
+
+        If a definition exists a tuple of (ref_string, CoreSchema) is returned.
```
tpdorsey, line 1494:
```suggestion
        If a definition exists, a tuple of `(ref_string, CoreSchema)` is returned.
```

pydantic/_internal/_generate_schema.py
```diff
@@ -1504,3 +1477,50 @@ def _common_field(
...
+        If no definition exists yet a tuple of (ref_string, None) is returned.
```
tpdorsey, line 1495:
```suggestion
        If no definition exists yet, a tuple of `(ref_string, None)` is returned.
```

pydantic/_internal/_generate_schema.py
```diff
@@ -1504,3 +1477,50 @@ def _common_field(
...
+
+        Note that the returned CoreSchema will always be a DefinitionReferenceSchema,
```
tpdorsey, line 1497:
```suggestion
        Note that the returned `CoreSchema` will always be a `DefinitionReferenceSchema`,
```

pydantic/_internal/_generate_schema.py
```diff
@@ -1504,3 +1477,50 @@ def _common_field(
...
+        not the actual definition itself.
+
+        This should be called for any type that can be identified by reference.
+        This includes any recursive types.
```
tpdorsey, line 1501:
```suggestion
        This includes any recursive types.
        
```

pydantic/_internal/_generate_schema.py
```diff
@@ -1504,3 +1477,50 @@ def _common_field(
...
+        At present the following types can be named / recursive:
```
tpdorsey, line 1502:
```suggestion
        At present the following types can be named/recursive:
        
```

tpdorsey: List needs to start with a blank line.

tpdorsey (commented):
Minor suggested docstring edits

adriangb:
@tpdorsey thanks for the suggestions. Please feel free to just push / commit these sorts of things in the future. I have no problem with anyone making changes to a branch I’m working on unless it’s likely to cause merge conflicts or is fundamentally changing logic (or meaning in the case of docstrings).

I will also point out that all of this is a private class in a private module, thus it’s really for Pydantic’s developers to look at more so than being rendered in documentation.

dmontagu (approved):

PR #6008, opened 2023-06-05
Allow `__pydantic_config__`  to be used on vanilla dataclasses

adriangb:
cc @dmontagu 

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -178,7 +178,12 @@ def dataclass(
 
         original_cls = cls
 
-        config_wrapper = _config.ConfigWrapper(config or getattr(cls, '__pydantic_config__', None) or None)
+        cls_config = getattr(cls, '__pydantic_config__', None)
+        config_dict = config
+        if config_dict is None:
+            if cls_config is not None:
+                config_dict = cls_config
+        config_wrapper = _config.ConfigWrapper(config_dict)
         decorators = _decorators.DecoratorInfos.build(cls)
 
         # Keep track of the original __doc__ so that we can restore it after applying the dataclasses decorator
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -178,10 +178,12 @@ def dataclass(
 
         original_cls = cls
 
-        cls_config = getattr(cls, '__pydantic_config__', None)
         config_dict = config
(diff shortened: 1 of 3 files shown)


pydantic/dataclasses.py
```diff
@@ -178,7 +178,7 @@ def create_dataclass(cls: type[Any]) -> type[PydanticDataclass]:
 
         original_cls = cls
 
-        config_wrapper = _config.ConfigWrapper(config)
+        config_wrapper = _config.ConfigWrapper(config or getattr(cls, '__pydantic_config__', None) or None)
```
dmontagu, line 181:
```suggestion
        config_wrapper = _config.ConfigWrapper(getattr(cls, '__pydantic_config__', None) if config is None else config)
```
Otherwise the empty config would be ignored. I think if an empty dict is passed though that should signify something different than passing just `None`.

If that seems wrong for some reason feel free to reject this.

dmontagu: Also, I'm not sure what the final `or None` is doing there, which is making we nervous that I'm misunderstanding something and there really is a good reason to convert anything falsy to `None`

adriangb: Yeah makes sense. I implemented it differently since this was going to trigger reformatting anyway. And added a test.

pydantic/dataclasses.py
```diff
@@ -178,7 +178,11 @@ def create_dataclass(cls: type[Any]) -> type[PydanticDataclass]:
 
         original_cls = cls
 
-        config_wrapper = _config.ConfigWrapper(config)
+        cls_config = getattr(cls, '__pydantic_config__', None)
+        config_dict = config
+        if config_dict is None and cls_config is not None:
+            config_dict = cls_config
+        config_wrapper = _config.ConfigWrapper(config_dict)
```
dmontagu, line 185:
```suggestion
        config_dict = config
        if config_dict is None:
            # if not explicitly provided, read from the type
            cls_config = getattr(cls, '__pydantic_config__', None)
            if cls_config is not None:
                config_dict = cls_config
        config_wrapper = _config.ConfigWrapper(config_dict)
```

Feel free to take or leave this change if you prefer the current implementation. I'm approving either way.

dmontagu (approved):

PR #6009, opened 2023-06-05
Give nicer error when using a bare root_validator

dmontagu:
Without this change, the following code snippet
```python
class Model(BaseModel):
    @root_validator
    @classmethod
    def validate_values(cls, values):
        return values
```
raises:
```
TypeError: root_validator() takes 0 positional arguments but 1 was given
```
Because this was valid code in v1, I think this is kind of problematic as the error message gives minimal insight into what needs to be changed to resolve the issue.

----

*With* the change in this PR, you instead get:
```
PydanticUserError: If you use `@root_validator` with pre=False (the default) you MUST
specify `skip_on_failure=True`. Note that `@root_validator` is deprecated and should
be replaced with `@model_validator`.
```
(The error message is not actually multiline, I just formatted it that way here for readability.)

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/deprecated/class_validators.py
+++ b/pydantic/deprecated/class_validators.py
@@ -188,7 +188,7 @@ def root_validator(
 
 
 def root_validator(
-    *,
+    *__args,
     pre: bool = False,
     skip_on_failure: bool = False,
     allow_reuse: bool = False,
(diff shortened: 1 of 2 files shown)


dmontagu:
I ran into this while working on the migration guide, as I had a note that you can't use bare `@root_validator` any more. But I think the change here is a better user experience than needing to read the migration guide up front to understand why bare `@root_validator` is an error.

please review

adriangb (approved):

Issue #6011, opened 2023-06-05
extra=allow breaks Base64Str


commonism:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

A model_config with extra:allow changes the behavior of a Base64Str attribute - the value is stored in `model_extra["data"]` instead of `data` and not encoded therefore.

### Example Code

```Python
from pydantic import BaseModel, Base64Str

# works
class GoodProperty(BaseModel):
    data: Base64Str

v = GoodProperty.model_construct(data="test")
assert v.model_dump() == {"data":'dGVzdA==\n'}

# fails
class BadProperty(BaseModel):
    model_config = dict(extra="allow")
    data: Base64Str

v = BadProperty.model_construct(data="test")
assert v.model_dump() == {"data":'dGVzdA==\n'}
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0b2
        pydantic-core version: 0.38.0 release build profile
                 install path: ~/workspace/pydantic/pydantic
               python version: 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]
                     platform: Linux-5.19.0-42-generic-x86_64-with-glibc2.35
     optional deps. installed: ['email-validator', 'typing-extensions']
```

dmontagu:
This seems to be related to a bigger issue that field serializers aren't getting called when extra='allow' — will look into this...

dmontagu:
Turns out it was just a bug in `model_construct`. PR incoming.

PR #6012, opened 2023-06-05
Fix bug with model_construct when extra=allow

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6011

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -484,13 +484,18 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
         """
         m = cls.__new__(cls)
         fields_values: dict[str, Any] = {}
+        defaults: dict[str, Any] = {}  # keeping this separate from `fields_values` helps us compute `_fields_set`
         for name, field in cls.model_fields.items():
             if field.alias and field.alias in values:
-                fields_values[name] = values[field.alias]
+                fields_values[name] = values.pop(field.alias)
             elif name in values:
-                fields_values[name] = values[name]
+                fields_values[name] = values.pop(name)
             elif not field.is_required():
-                fields_values[name] = field.get_default(call_default_factory=True)
+                defaults[name] = field.get_default(call_default_factory=True)
+        if _fields_set is None:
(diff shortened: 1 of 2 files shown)


tests/test_construction.py
```diff
@@ -500,3 +500,12 @@ def test_model_copy(ModelTwo):
...
     assert m.a != m2.a
+
+
+def test_pydantic_extra():
+    class Model(BaseModel):
+        model_config = dict(extra='allow')
+        x: int
+
+    m = Model.model_construct(x=1, y=2)
+    assert m.__pydantic_extra__ == {'y': 2}
```
dmontagu, line 511: without this change, you get `{'x': 1, 'y': 2}`

dmontagu:
please review

pydantic/main.py
```diff
@@ -484,13 +484,18 @@ def model_construct(cls: type[Model], _fields_set: set[str] | None = None, **val
...
         for name, field in cls.model_fields.items():
             if field.alias and field.alias in values:
-                fields_values[name] = values[field.alias]
+                fields_values[name] = values.pop(field.alias)
             elif name in values:
-                fields_values[name] = values[name]
+                fields_values[name] = values.pop(name)
             elif not field.is_required():
-                fields_values[name] = field.get_default(call_default_factory=True)
+                defaults[name] = field.get_default(call_default_factory=True)
```
adriangb, line 494: Do we care about the actual value here or just that there is a default? Calling the default factory seems suspect

dmontagu: We need the actual value as we are using it to populate the dict for the model. `call_default_factory=True` just means that it will be called if present. We would only do this if you don't pass a value for the field, and given that if that were the case and we didn't do this it would result in a malformed model I think this is preferable. Either way, there was an existing and old test for this: `tests.test_construction.test_construct_default_factory`

adriangb: Oh I see this is only when calling model_construct

adriangb (approved):

PR #6016, opened 2023-06-06
uprev coverage, try to fix ci

samuelcolvin:
related to https://github.com/pdm-project/pdm/issues/1981 - try to fix by updating `pdm.lock` and thereby prompting `pdm-setup` to not use the cache files.


samuelcolvin:
fixed by #6014

PR #6023, opened 2023-06-06
🐛 Fix aliases priority

Kludex:
## Change Summary

Fix aliases priority.

## Checklist

### DIFF
--- a/tests/test_aliases.py
+++ b/tests/test_aliases.py
@@ -238,6 +238,62 @@ def test_low_priority_alias():
     assert [f.serialization_alias for f in Child.model_fields.values()] == ['w_ser_alias', 'X', 'Y', 'Z']
 
 
+def test_aliases_priority():
+    class Model(BaseModel):
+        x: int = Field(..., alias='x1', validation_alias='x2')
+
+    # alias should take priority over validation_alias
+    model = Model(x2=1)
+    assert model.x == 1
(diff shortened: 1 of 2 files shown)


tests/test_aliases.py
```diff
@@ -238,6 +238,62 @@ class Child(Parent):
     assert [f.serialization_alias for f in Child.model_fields.values()] == ['w_ser_alias', 'X', 'Y', 'Z']
 
 
+def test_aliases_priority():
```
dmontagu, line 241: Can we break these into separate tests? I don't like redefining the type multiple times in the same test, I got super confused for a minute because I looked halfway down after seeing the definition at the top before I realized the type was getting redefined in each block

dmontagu (approved):
Other than wanting the mega test split up, this looks good to me

tests/test_aliases.py
dmontagu: (Also would be annoying to debug if something breaks some of these cases and not others, as it will fail out on the first one rather than letting you see precisely which ones fail.)

dmontagu: Could also use `pytest.mark.parametrize` but no need if it's hard to structure the code in a parametric way — I think it's fine to just have a handful of similar-but-different tests if that's easier.

Kludex: I didn't want to parametrize, but this case it didn't look that bad... 👀

PR #6031, opened 2023-06-07
📝 Add documentation for `PhoneNumber` and `ABARoutingNumber`

Kludex:
- Complementary PR for https://github.com/pydantic/pydantic-extra-types/pull/60

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -79,6 +79,8 @@ nav:
     - Extra Types:
       - Color Types: usage/types/extra_types/color_types.md
       - Payment Card Numbers: usage/types/extra_types/payment_cards.md
+      - Phone Numbers: usage/types/extra_types/phone_numbers.md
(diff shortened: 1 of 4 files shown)


mkdocs.yml
```diff
@@ -79,6 +79,8 @@ nav:
     - Extra Types:
       - Color Types: usage/types/extra_types/color_types.md
       - Payment Card Numbers: usage/types/extra_types/payment_cards.md
+      - Phone Numbers: usage/types/extra_types/phone_numbers.md
+      - Routing Number: usage/types/extra_types/routing_number.md
```
hramezani, line 83: `Routing Number` or `Routing Numbers`? I can see `Routing Numbers` in `types.md`.

Also, the file name has to be fixed I think. because we have `phone_numbers.md` for `Phone Numbers` but the file name for `Routing Number` is singular.

I would suggest to make all of them plural

Kludex: done

PR #6043, opened 2023-06-08
Sort produced json schemas

### LINKED ISSUE #6013: Produce best-effort deterministically sorted json schemas
A lot of our tests, and probably user tests as well, at least via FastAPI, make assertions about the JSON schemas. It is very convenient for the order of keys and items in these schemas to be as deterministic as possible so that tests don't have to change if the schema gets shuffled because of internal refactors. This is especially important when comparing a text representation (like we do in our examples). When comparing as Python objects assertions won't fail (order of dicts does not matter when comparing them for equality) but the diff will change, which is annoying.

At the very least we could insert a function at the end of the json schema production that goes through it and recursively sorts keys and items.

adriangb:
closes #6013 

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -45,10 +45,11 @@ A type alias for defined schema types that represents a union of `core_schema.Co
 `core_schema.CoreSchemaFieldType`.
 """
 
-JsonSchemaValue = Dict[str, Any]
+JsonSchemaValue = Dict[str, Any]  # type: ignore
 """
 A type alias for a JSON schema value. This is a dictionary of string keys to arbitrary values.
 """
+
 
 JsonSchemaMode = Literal['validation', 'serialization']
 """
@@ -261,7 +262,7 @@ class GenerateJsonSchema:
                     refs_map[(key, mode)] = defs_ref
 
         self._used = True
-        return refs_map, self.definitions
+        return refs_map, _sort_json_schema(self.definitions)  # type: ignore
 
     def generate(self, schema: CoreSchema, mode: JsonSchemaMode = 'validation') -> JsonSchemaValue:
         """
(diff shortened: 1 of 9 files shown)


pydantic/json_schema.py
```diff
@@ -45,11 +45,12 @@
 `core_schema.CoreSchemaFieldType`.
 """
 
-JsonSchemaValue = Dict[str, Any]
+JsonSchemaValue = Dict[str, Any]  # type: ignore
 """
 A type alias for a JSON schema value. This is a dictionary of string keys to arbitrary values.
 """
 
+
```
adriangb, line 53:
```suggestion
```

pydantic/json_schema.py
```diff
@@ -45,11 +45,12 @@
 `core_schema.CoreSchemaFieldType`.
 """
 
-JsonSchemaValue = Dict[str, Any]
+JsonSchemaValue = Dict[str, Any]  # type: ignore
```
adriangb, line 48:
```suggestion
JsonSchemaValue = Dict[str, Any]
```

adriangb:
please review

Kludex (approved):

pydantic/json_schema.py
```diff
@@ -1621,3 +1621,18 @@ def _make_json_hashable(value: _Json) -> _HashableJson:
...
+        for key in sorted(value.keys()):
+            sorted_dict[key] = _sort_json_schema(value[key])
+        return sorted_dict  # type: ignore
+    elif isinstance(value, list):  # type: ignore
+        sorted_list: list[JsonSchemaValue] = []
+        for item in value:  # type: ignore
+            sorted_list.append(_sort_json_schema(item))
+        return sorted_list  # type: ignore
+    else:
+        return value
```
Kludex, line 1638: For the record, this is the only thing added here.

PR #6052, opened 2023-06-08
Docs updates for type adapter

tpdorsey:
## Checklist

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -761,12 +761,15 @@ except ValidationError as e:
     - the model must be defined globally
     - it must provide `__module__`
 
-## Using Pydantic without creating a `BaseModel`
+## `TypeAdapter`
+
+`TypeAdapter` enables using Pydantic for type validation and coercion without creating a `BaseModel`
 
 You may have types that are not `BaseModel`s that you want to validate data against.
 Or you may want to validate a `List[SomeModel]`, or dump it to JSON.
 
-To do this Pydantic provides `TypeAdapter`. A `TypeAdapter` instance behaves nearly the same as a `BaseModel` instance, with the difference that `TypeAdapter` is not an actual type so you cannot use it in type annotations and such.
+To do this, Pydantic provides `TypeAdapter`. A `TypeAdapter` instance behaves nearly the same as a `BaseModel` instance,
+with the difference that `TypeAdapter` is not an actual type, so you cannot use it in type annotations and such.
 
 ```py
 from typing import List
@@ -798,12 +801,12 @@ except ValidationError as e:
     """
 ```
 
-For many use cases `TypeAdapter` can replace BaseModels with a `__root__` field in Pydantic V1.
+For many use cases, `TypeAdapter` can replace `BaseModel`s with a `__root__` field in Pydantic V1.
 
 ### Parsing data into a specified type
 
 `TypeAdapter` can be used to apply the parsing logic to populate pydantic models in a more ad-hoc way.
-This function behaves similarly to `BaseModel.model_validate`, but works with arbitrary pydantic-compatible types.
+This function behaves similarly to `BaseModel.model_validate`, but works with arbitrary Pydantic-compatible types.
 
 This is especially useful when you want to parse results into a type that is not a direct subclass of `BaseModel`.
 For example:
@@ -830,7 +833,7 @@ print(items)
 
 `TypeAdapter` is capable of parsing data into any of the types pydantic can handle as fields of a `BaseModel`.
 
-## Custom root types
+## `RootModel` and custom root types
 
 Pydantic models can be defined with a custom root type by declaring the `RootModel`.
 
--- a/pydantic/type_adapter.py
+++ b/pydantic/type_adapter.py
@@ -97,7 +97,11 @@ def _getattr_no_parents(obj: Any, attribute: str) -> Any:
 
 
 class TypeAdapter(Generic[T]):
-    """A class representing the type adapter.
+    """
+    A class representing the type adapter.
+
+    A `TypeAdapter` instance behaves nearly the same as a `BaseModel` instance,
+    with the difference that `TypeAdapter` is not an actual type, so you cannot use it in type annotations.
 
     Attributes:
         core_schema (CoreSchema): The core schema for the type.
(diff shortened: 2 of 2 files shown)


tpdorsey:
please review

pydantic/type_adapter.py
```diff
@@ -338,7 +345,8 @@ def json_schemas(
         ref_template: str = DEFAULT_REF_TEMPLATE,
         schema_generator: type[GenerateJsonSchema] = GenerateJsonSchema,
     ) -> tuple[dict[tuple[JsonSchemaKeyT, JsonSchemaMode], DefsRef], JsonSchemaValue]:
-        """Generate JSON schemas for multiple type adapters.
+        """
+        Generate JSON schemas for multiple type adapters.
```
dmontagu, line 349:
```suggestion
        Generate a JSON schema including definitions from multiple type adapters.
```

pydantic/type_adapter.py
```diff
@@ -179,41 +183,41 @@ def validate_python(
...
-            T: The validated object.
-
+            The validated object.
         """
         return self.validator.validate_json(__data, strict=strict, context=context)
 
     def get_default_value(self, *, strict: bool | None = None, context: dict[str, Any] | None = None) -> Some[T] | None:
-        """Get the default value for this model / type.
+        """
+        Get the default value for the model.
```
dmontagu, line 213:
```suggestion
        Get the default value for the wrapped type.
```

pydantic/type_adapter.py
```diff
@@ -97,7 +97,11 @@ def _getattr_no_parents(obj: Any, attribute: str) -> Any:
 
 
 class TypeAdapter(Generic[T]):
-    """A class representing the type adapter.
+    """
+    A class representing the type adapter.
+
+    A `TypeAdapter` instance behaves nearly the same as a `BaseModel` instance,
+    with the difference that `TypeAdapter` is not an actual type, so you cannot use it in type annotations.
```
dmontagu, line 104:
```suggestion
    A `TypeAdapter` instance exposes some of the functionality from `BaseModel` instance methods
    for types that do not have such methods (such as dataclasses, primitive types, and more).
    
    Note that `TypeAdapter` is not an actual type, so you cannot use it in type annotations.
```

docs/usage/models.md
```diff
@@ -830,7 +833,7 @@ print(items)
 
 `TypeAdapter` is capable of parsing data into any of the types pydantic can handle as fields of a `BaseModel`.
 
-## Custom root types
+## `RootModel` and custom root types
```
hramezani, line 836: Same here: https://github.com/pydantic/pydantic/blob/31a4ac3f3b802fadbd90b2be2fc575b20a79c356/docs/usage/dataclasses.md?plain=1#L389
https://github.com/pydantic/pydantic/blob/31a4ac3f3b802fadbd90b2be2fc575b20a79c356/docs/usage/exporting_models.md?plain=1#L135

docs/usage/models.md
```diff
@@ -798,12 +801,12 @@ except ValidationError as e:
     """
 ```
 
-For many use cases `TypeAdapter` can replace BaseModels with a `__root__` field in Pydantic V1.
+For many use cases, `TypeAdapter` can replace `BaseModel`s with a `__root__` field in Pydantic V1.
```
dmontagu, line 804:
```suggestion
For some use cases, `TypeAdapter` can replace `BaseModel`s with a `__root__` field in Pydantic V1.
```
I think saying *many* here is a bit too strong; most of the usage I saw of `__root__` when searching github involved the use of methods, which is what would require `RootModel`. I think saying "For some use cases" is fine though, regardless of the actual distribution.

It may make sense to point people at `RootModel` here since we are referencing `__root__` (as this is the way to handle the other usages of `__root__` from V1).

pydantic/type_adapter.py
```diff
@@ -179,41 +183,41 @@ def validate_python(
...
     ) -> T:
-        """Validate a JSON string or bytes against the model.
+        """
+        Validate a JSON string or bytes against the model.
 
         Args:
             __data: The JSON data to validate against the model.
-            strict: Whether to strictly check types. Defaults to None.
-            context: Additional context to use during validation. Defaults to None.
+            strict: Whether to strictly check types. Defaults to `None`.
```
dmontagu, line 203:
```suggestion
            strict: Whether to strictly check types. Defaults to `None`.
```
I think it's a bit confusing to describe this as a "whether", then give three choices (`True` / `False` / `None`). I don't know the best way to communicate the meaning to the users, but for context, the way it works (if I understand correctly, anyway) is that `True` means definitely do strict mode, `False` means definitely don't do strict mode, and `None` means use whatever the default is based on the wrapped type / config.

I think it's worth keeping this description short, but it would be nice if we could provide a bit more clarity here. Maybe something like
```suggestion
            strict: Pass a boolean to override the type checking strictness for the wrapped type. Defaults to `None`.
```

pydantic/type_adapter.py
```diff
@@ -276,7 +281,8 @@ def dump_json(
         round_trip: bool = False,
         warnings: bool = True,
     ) -> bytes:
-        """Serialize the given instance to JSON.
+        """
+        Serialize the model to JSON.
```
dmontagu, line 285:
```suggestion
        Serialize an instance of the adapted type to JSON.
```

pydantic/type_adapter.py
```diff
@@ -315,7 +321,8 @@ def json_schema(
         schema_generator: type[GenerateJsonSchema] = GenerateJsonSchema,
         mode: JsonSchemaMode = 'validation',
     ) -> dict[str, Any]:
-        """Generate a JSON schema for the model.
+        """
+        Generate a JSON schema for the model.
```
dmontagu, line 325:
```suggestion
        Generate a JSON schema for the adapted type.
```

tpdorsey:
please review

tpdorsey:
Closes DOC-88

docs/usage/dataclasses.md
```diff
@@ -388,7 +388,7 @@ methods decorated with `model_validator`.
 ## JSON dumping
 
 _Pydantic_ dataclasses do not feature a `.model_dump_json()` function. To dump them as JSON, you will need to
-make use of the [RootModel](models.md#custom-root-types) as follows:
+make use of the [RootModel](models.md#rootmodel-and-custom-root-types) as follows:
```
adriangb, line 391: Can't we also use a TypeAdapter? Maybe I'm getting confused by the diff.

adriangb:
@tpdorsey i think you just need to run `make format`

PR #6064, opened 2023-06-08
Minor docs improvements

dmontagu:
Closes DOC-6
Closes DOC-9
Closes DOC-19

### DIFF
--- a/pydantic/validate_call.py
+++ b/pydantic/validate_call.py
@@ -1,4 +1,6 @@
-"""Decorators for validating function calls."""
+"""
+Decorators for validating function calls.
+"""
 from __future__ import annotations as _annotations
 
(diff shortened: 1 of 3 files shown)


adriangb (approved):

pydantic/validate_call.py
```diff
@@ -32,7 +34,7 @@ def validate_call(
     validate_return: bool = False,
 ) -> AnyCallableT | Callable[[AnyCallableT], AnyCallableT]:
     """
-    Decorator to validate the arguments passed to a function and, optionally, the return value.
+    Returns a decorated version of the function that validates the arguments and, optionally, the return value.
```
Kludex, line 37: To me, if a function description starts with "returns", then it shouldn't be here, but on the "Return" part of the docstrings. 👀

PR #6072, opened 2023-06-09
Create `fields.md` on the documentation

Kludex:
Closes DOC-103

## Change Summary

- Rename `schema.md` to `json_schema.md`.
- Create initial `fields.md` documentation.

Selected Reviewer: @adriangb


Kludex:
please review

docs/index.md
```diff
@@ -31,7 +31,7 @@ Some of the main features of Pydantic include:
 - [**Custom and nested models**](/usage/models.md): You can define models (similar to classes) that contain other models, allowing for complex data structures to be neatly and efficiently represented.
 - [**Generic models**](/usage/models.md/#generic-models): Pydantic supports generic models, which allow the declaration of models that are "parameterized" on one or more fields.
 - [**Dataclasses**](/usage/dataclasses.md): Pydantic supports `dataclasses.dataclass`, offering same data validation as using `BaseModel`.
-- [**Model schema generation**](/usage/schema.md): Pydantic models can be converted to and from a JSON Schema, which can be useful for documentation, code generation, or other purposes.
+- [**JSON schema generation**](/usage/json_schema.md): Pydantic models can be converted to and from a JSON Schema, which can be useful for documentation, code generation, or other purposes.
```
dmontagu, line 34:
```suggestion
- [**JSON schema generation**](/usage/json_schema.md): Pydantic models can be converted to a JSON Schema, which can be useful for documentation, code generation, or other purposes.
```

docs/usage/fields.md
```diff
@@ -0,0 +1,85 @@
+
+The `Field` class is used to define fields in models.
+
+TODO: Refactor the `Field customization` section in the `json_schema.md`.
```
dmontagu, line 4: Assuming you want this PR merged before finishing this TODO, could you:
* add a note specifically saying to move documentation of `pydantic.fields.Field` here
* also mention that we should document all arguments to `Field` here, obviously that's closely related to the previous bullet — but not equivalent imo
* create a linear task for `usage/fields.md` and add it as a subtask of the `Review usage documentation` task

(The above bullets may or may not apply depending on how much work you want to put into this file before merging.)

Kludex: Hmm... Can't I just mention the 3 points that you said on the linear task? 🤔

docs/usage/fields.md
```diff
@@ -0,0 +1,85 @@
...
+For validation and serialization, you can define an alias for a field.
+
+There are three ways to define an alias:
+
+* `Field(..., alias='foo')`
+* `Field(..., validation_alias='foo')`
+* `Field(..., serialization_alias='foo')`
+
+The `alias` parameter is used for both _validation_ and _serialization_. The `validation_alias` and `serialization_alias`
+parameters are used for _validation_ and _serialization_ respectively.
```
dmontagu, line 17:
```suggestion
The `alias` parameter is used for both validation _and_ serialization. If you want to use 
_different_ aliases for validation and serialization respectively, you can use the`validation_alias`
and `serialization_alias` parameters, which will apply only in their respective use cases.
```

docs/usage/fields.md
```diff
@@ -0,0 +1,85 @@
...
+
```
dmontagu, line 18: I think around this point I would add a little info callout talking about Intellisense in vscode and how it will use the value provided in `alias` for the `__init__` signature. Then I would say something to the effect of "if you set `populate_by_name=True` in the model config, and want to use the field name in the `__init__`, use `Field(alias=str(<alias value>))`.

I might also mention that at this time it is not possible for Intellisense to reflect the value of `validation_alias` in the `__init__`, but if desired you could get the modified init signature by replacing `Field(validation_alias='myValidationAlias')` with `Field(alias='myValidationAlias', serialization_alias='my_field_name')`.

dmontagu: For purposes of "flow", I can see how it might make sense to put such a callout at the end of this file rather than here.

But I feel this will be a fairly frequent point of frustration for users, so if possible I'd like to include it near the top to make it more likely for people to stumble upon it.

docs/usage/fields.md
dmontagu: yes, as long as it is tracked somewhere that people will be forced to acknowledge

dmontagu (approved):

docs/usage/fields.md
```diff
@@ -0,0 +1,179 @@
...
+The `Field` class is used to define fields in models.
```
adriangb, line 2:
```suggestion
The `Field` class is used to customize and add metadata to fields of models.
```

adriangb (commented):
Looks like I'm too late to review :)

docs/usage/fields.md
Kludex: I'll work on this file. I'll include this. 🙏

dmontagu: Not too late I'll add it to my next docs PR. Do you plan to add more suggestions?

adriangb: No all good, I didn't have any more queued up I was just starting to review

PR #6076, opened 2023-06-09
Update booleans.md and bytesize.md

dmontagu:
Closes DOC-62, DOC-63

Selected Reviewer: @lig

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -1133,7 +1133,8 @@ print(Model(a=3.000, b='2.72', c=b'binary data').model_dump())
 This is a deliberate decision of *pydantic*, and in general it's the most useful approach. See
 [here](https://github.com/pydantic/pydantic/issues/578) for a longer discussion on the subject.
 
-Nevertheless, [strict type checking](types/types.md#strict-types) is partially supported.
+Nevertheless, [strict type checking](types/standard_types.md#strict-types) is partially supported.
+[TODO: Discuss strict mode, but also retain the previous link to types/standard_types.md#strict-types]
 
 ## Model signature
 
--- a/docs/usage/types/booleans.md
+++ b/docs/usage/types/booleans.md
@@ -5,23 +5,17 @@ description: Support for Boolean types.
 `bool`
 : see below for details on how bools are validated and what values are permitted
 
-!!! warning
-    The logic for parsing `bool` fields has changed as of version **v1.0**.
-
-    Prior to **v1.0**, `bool` parsing never failed, leading to some unexpected results.
-    The new logic is described below.
-
 A standard `bool` field will raise a `ValidationError` if the value is not one of the following:
 
 * A valid boolean (i.e. `True` or `False`),
 * The integers `0` or `1`,
 * a `str` which when converted to lower case is one of
   `'0', 'off', 'f', 'false', 'n', 'no', '1', 'on', 't', 'true', 'y', 'yes'`
-* a `bytes` which is valid (per the previous rule) when decoded to `str`
+* a `bytes` which is valid per the previous rule when decoded to `str`
 
 !!! note
     If you want stricter boolean logic (e.g. a field which only permits `True` and `False`) you can
-    use [`StrictBool`](#strict-types).
+    use [`StrictBool`](standard_types.md#strict-types).
 
 Here is a script demonstrating some of these behaviors:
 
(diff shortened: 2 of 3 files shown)


dmontagu:
please review

docs/usage/models.md
```diff
@@ -1133,7 +1133,8 @@ print(Model(a=3.000, b='2.72', c=b'binary data').model_dump())
 This is a deliberate decision of *pydantic*, and in general it's the most useful approach. See
 [here](https://github.com/pydantic/pydantic/issues/578) for a longer discussion on the subject.
 
-Nevertheless, [strict type checking](types/types.md#strict-types) is partially supported.
+Nevertheless, [strict type checking](types/standard_types.md#strict-types) is partially supported.
+[TODO: Discuss strict mode, but also retain the previous link to types/standard_types.md#strict-types]
```
dmontagu, line 1137: I noticed this needs to be addressed here, so wanted to create a note, but there is a separate issue for working on this file and didn't want to get distracted by fixing that now..

docs/usage/types/booleans.md
```diff
@@ -5,23 +5,17 @@ description: Support for Boolean types.
 `bool`
 : see below for details on how bools are validated and what values are permitted
 
-!!! warning
```
dmontagu, line 8: This warning was for v1.0; it seems to me safe to remove this now, especially since older docs versions will still be accessible.

tpdorsey (approved):

hramezani (approved):

docs/usage/types/booleans.md
lig: I'd agree. @tpdorsey ?

docs/usage/types/bytesize.md
```diff
@@ -1,32 +1,49 @@
...
 The following arguments are available when using the `conbytes` type function
 
-- `strip_whitespace: bool = False`: removes leading and trailing whitespace
-- `to_upper: bool = False`: turns all characters to uppercase
-- `to_lower: bool = False`: turns all characters to lowercase
 - `min_length: int = None`: minimum length of the byte string
 - `max_length: int = None`: maximum length of the byte string
 - `strict: bool = False`: controls type coercion
 
+However, for the sake of improved integration with type checkers, we now discourage the use of `conbytes` and other
```
lig, line 22: "discouraging" seems ambiguous here. Are those deprecated or not? If they aren't deprecated then we just should support them. Are we going to deprecate those in one of the following releases?

docs/usage/types/bytesize.md
```diff
@@ -1,32 +1,49 @@
...
+
+
 ## Using ByteSize
 
-You can use the `ByteSize` data type to convert byte string representation to
-raw bytes and print out human readable versions of the bytes as well.
+You can use the `ByteSize` data type to (case-insensitively) convert a string representation of a number of bytes into
+an integer, and also to print out human-readable strings representing a number of bytes.
+
+We interpret `'1KB'` to mean 1000 bytes, and `'1KiB'` to mean 1024 bytes. In general, including a middle `'i'`
```
lig, line 44: > including a middle `'i'`

Well, yes but it sounds hacky to me as I know that those are just from an ISO standard and I would like to see a guarantee that this follows the `IEC 80000-13` standard.

lig (changes requested):

lig:
please update

docs/usage/types/booleans.md
tpdorsey: Agree

docs/usage/types/bytesize.md
dmontagu: @lig do you have a link we can use for reference about this? I couldn't find/come up with a good way to describe it, would much rather defer to someone else's description. Open to suggestions here if you want to take a stab.

lig: I think the closest link that doesn't point to a paywall directly is to the Wikipedia page https://en.wikipedia.org/wiki/ISO/IEC_80000

lig:
```suggestion
In conformance with [IEC 80000-13 Standard](https://en.wikipedia.org/wiki/ISO/IEC_80000) we interpret `'1KB'` to mean 1000 bytes, and `'1KiB'` to mean 1024 bytes. In general, including a middle `'i'`
```

dmontagu:
@lig I merged your suggestion, did you have any other feedback? (hooky: please review)

lig (approved):
👍🏻

PR #6078, opened 2023-06-10
Update documentation about exporting models

dmontagu:
* I changed 'serialise' to 'serialize' (and similar for -ing, etc.) everywhere. The method names included the 'z' so I think that was the way to go, and ultimately I believe there were far fewer occurrences with an 's' than with a 'z'.
* I reordered some of the sections in the exporting models docs in a way that I felt was more cohesive.
* There was one regression relative to V1 that I noticed in the docs, and which I'm not sure was intentional. I'll point it out in a comment on the code. I'm not sure what we should do about it.


docs/usage/exporting_models.md
```diff
@@ -302,9 +235,47 @@ print(m.model_dump_json())
...
+#> {'inner': {'x': 1}}
+```
+!!! warning "Migration Warning"
+    This behavior is different from how things worked in Pydantic V1, where we would always include
+    all (subclass) fields when recursively dumping models to dicts. The motivation behind this change in
+    behavior is that it helps ensure that you know precisely which fields could be included when serializing,
+    even if subclasses get passed when instantiating the object. In particular, this can help prevent surprises
+    when adding sensitive information like secrets as fields of subclasses.
+
+TODO: Add and document `SerializeAsAny` type.
```
dmontagu, line 274: I have docs ready for this. @adriangb you didn't do any work on this (or run into issues trying) did you? If not, I'm happy to quickly throw together a PR adding this along with docs after this gets merged.

docs/usage/exporting_models.md
```diff
@@ -494,13 +468,15 @@ t = Transaction(
 
 print(t.model_dump())
 #> {'id': '1234567890'}
-# TODO this is wrong! not all of "user" should be excluded
+# TODO: this is wrong! not all of "user" should be excluded
+# TODO: do we need to fix the type of the argument to Field? Or do we want to just remove that functionality?
```
dmontagu, line 472: The line `    user: User = Field(exclude={'username'})` above is actually a type error for the `exclude` argument. It seems we changed how that argument works. I noticed this before, but wasn't quite sure how we want this to work. @samuelcolvin @adriangb thoughts?

This section may need a bigger rework if we aren't going to fix that.

docs/usage/exporting_models.md
adriangb: I may have fixed it in #6007 but I didn’t add a type and export it.

tpdorsey (approved):
Doc portion of this looks good. Approved pending resolution of comments.

docs/blog/pydantic-v2.md
```diff
@@ -210,7 +210,7 @@ In future direct validation of JSON will also allow:
     While in theory other formats could be specifically supported, the overheads and development time are
     significant and I don't think there's another format that's
     used widely enough to be worth specific logic. Other formats can be parsed to python then validated, similarly
-    when serialising, data can be exported to a python object, then serialised,
+    when serializing, data can be exported to a python object, then serialized,
```
adriangb, line 213: Just for the record, `serialising` is also correct, it's just a UK thing.

docs/usage/exporting_models.md
```diff
@@ -64,15 +69,22 @@ print(Model(x=['{"a": 1}', '[1, 2]']).model_dump(round_trip=True))
...
+## `model.model_dump_json(...)`
 
-*pydantic* models can also be converted to dictionaries using `dict(model)`, and you can also
-iterate over a model's field using `for field_name, value in model:`. With this approach the raw field values are
-returned, so sub-models will not be converted to dictionaries.
+The `.model_dump_json()` method will serialize a model to JSON.
 
-Example:
+The behavior of this method is similar to `.model_dump` (including how it handles
+[`RootModel`](models.md#custom-root-types) and subclasses), just that the returned result will be a JSON-encoded string.
```
adriangb, line 77: I thought for RootModel in JSON there is no `__root__`/`root` key?

adriangb (approved):

docs/usage/exporting_models.md
lig: There is no `root` key in `json`. It's the same as for `model_dump`. So, `RootModel.model_dump()` doesn't have the `root` field and neither `RootModel.model_dump_json()` does.

IMO, this documentation doesn't contradict this.

dmontagu: How would you guys feel about me replacing lines 74-77 with:

```
The `.model_dump_json()` method serializes a model directly to a JSON-encoded string that is equivalent to the result produced by `.model_dump`.
```

I feel like this addresses the confusion, and is more terse to boot. If I get a 👍 I'll push the change (couldn't do a suggestion since deleted lines are in the span)

docs/blog/pydantic-v2.md
dmontagu: yeah, I just wanted to standardize it, and since we use `serialize` more frequently than `serialise` throughout the codebase I just find-and-replaced everywhere

docs/usage/exporting_models.md
dmontagu: I made the change since I was working on this to resolve conflicts anyway. Happy to reconsider if appropriate

PR #6089, opened 2023-06-12
Improve `pydantic/_internal/_core_metadata.py` docstring

hramezani:
Closes DOC-23
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_core_metadata.py
+++ b/pydantic/_internal/_core_metadata.py
@@ -18,7 +18,7 @@ class CoreMetadata(typing_extensions.TypedDict, total=False):
     """A `TypedDict` for holding the metadata dict of the schema.
 
     Attributes:
-        pydantic_js_functions: List of Json schema functions.
+        pydantic_js_functions: List of JSON schema functions.
         pydantic_js_prefer_positional_arguments: Whether JSON schema generator will
             prefer positional over keyword arguments for an 'arguments' schema.
     """
--- a/pydantic/_internal/_core_metadata.py
+++ b/pydantic/_internal/_core_metadata.py
@@ -15,7 +15,7 @@ if typing.TYPE_CHECKING:
 
 
 class CoreMetadata(typing_extensions.TypedDict, total=False):
-    """A TypedDict for holding the metadata dict of the schema
+    """A `TypedDict` for holding the metadata dict of the schema.
 
     Attributes:
         pydantic_js_functions: List of Json schema functions.
--- a/pydantic/_internal/_core_metadata.py
+++ b/pydantic/_internal/_core_metadata.py
@@ -4,30 +4,30 @@ import typing
 from typing import Any
 
 import typing_extensions
-from pydantic_core import CoreSchema
 
 if typing.TYPE_CHECKING:
-    from ..json_schema import JsonSchemaValue
     from ._schema_generation_shared import (
         CoreSchemaOrField as CoreSchemaOrField,
     )
     from ._schema_generation_shared import (
         GetJsonSchemaFunction,
-        GetJsonSchemaHandler,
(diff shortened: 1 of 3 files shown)


pydantic/_internal/_core_metadata.py
```diff
@@ -60,16 +60,6 @@ def metadata(self) -> CoreMetadata:
             raise TypeError(f'CoreSchema metadata should be a dict; got {metadata!r}.')
         return metadata  # type: ignore[return-value]
 
-    def get_json_schema(
```
hramezani, line 63: same here. there is just one test that was added to cover this function. I've removed both the function and test

pydantic/_internal/_core_metadata.py
```diff
@@ -4,32 +4,32 @@
...
+    """
+
     pydantic_js_functions: list[GetJsonSchemaFunction]
 
     # If `pydantic_js_prefer_positional_arguments` is True, the JSON schema generator will
     # prefer positional over keyword arguments for an 'arguments' schema.
     pydantic_js_prefer_positional_arguments: bool | None
 
 
-class UpdateCoreSchemaCallable(typing_extensions.Protocol):
```
hramezani, line 28: Unused class. at least I couldn't find any usage of this class in the code.

hramezani:
please review

pydantic/_internal/_core_metadata.py
```diff
@@ -4,32 +4,32 @@
...
         CoreSchemaOrField as CoreSchemaOrField,
     )
     from ._schema_generation_shared import (
         GetJsonSchemaFunction,
-        GetJsonSchemaHandler,
     )
 
 
 class CoreMetadata(typing_extensions.TypedDict, total=False):
+    """A TypedDict for holding the metadata dict of the schema
```
tpdorsey, line 18:
```suggestion
    """A `TypedDict` for holding the metadata dict of the schema.
```

pydantic/_internal/_core_metadata.py
```diff
@@ -4,32 +4,32 @@
...
+
+    Attributes:
+        pydantic_js_functions: List of Json schema functions.
```
tpdorsey, line 21:
```suggestion
        pydantic_js_functions: List of JSON schema functions.
```

dmontagu: @hramezani I see you marked this resolved but didn't change it; I also support this change

pydantic/_internal/_core_metadata.py
dmontagu: (Same with this — I support the change)

dmontagu (commented):
handful of minor suggestions then I think it will be good to merge, please update

pydantic/json_schema.py
```diff
@@ -392,7 +393,7 @@ def handler_func(schema_or_field: CoreSchemaOrField) -> JsonSchemaValue:
 
             def new_handler_func(
                 schema_or_field: CoreSchemaOrField,
-                current_handler: _core_metadata.GetJsonSchemaHandler = current_handler,
+                current_handler: GetJsonSchemaHandler = current_handler,
```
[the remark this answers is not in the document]
hramezani: We used `_core_metadata.GetJsonSchemaHandler` which was wrong because `GetJsonSchemaHandler` is defined in `_schema_generation_shared` and imported in `_core_metadata`.

As there are some removed codes in `_core_metadata`, we don't need to import `GetJsonSchemaHandler` in `_core_metadata`. So, we have to fix the import here as well.

I've changed the next line as well.

hramezani:
please review

dmontagu (approved):

PR #6100, opened 2023-06-12
Fix `__get_pydantic_core_schema__` for Enum

### LINKED ISSUE #5999: `to_string_ser_schema` doesn't work properly anymore
### Initial Checks

### Description

I have a test to check how some schemas work, and it is broken after an update. 

`model.model_dump(mode='json') == {'flag': str(flag)}` returns False, but

```py
>           assert model.model_dump(mode='json') == {'flag': str(flag)}
E           AssertionError: assert {'flag': 3} == {'flag': '3'}
E             Differing items:
E             {'flag': 3} != {'flag': '3'}
E             Use -v to get more diff
```

I suppose the problem is near from `to_string_ser_schema`

### Example Code

```Python
import enum
from typing import Any, Callable
from pydantic_core import CoreSchema, core_schema

def test_permission_flag():
    @enum.unique
    class PermissionFlag(enum.IntFlag):
        TEST1 = 1 << 0
        TEST2 = 1 << 1

        @classmethod
            cls, _source: type[BaseModel], _handler: Callable[[Any], CoreSchema],
        ) -> CoreSchema:
            schema = core_schema.union_schema([
                core_schema.int_schema(),
                core_schema.str_schema(),
                core_schema.is_instance_schema(cls),
            ])

            return core_schema.no_info_after_validator_function(
                function=cls.validate,
                schema=schema,
                serialization=core_schema.to_string_ser_schema(),
            )

    class TestModel(BaseModel):
        flag: PermissionFlag

    for permission_type in {PermissionFlag, int, str}:
        flag = permission_type(PermissionFlag.TEST1 | PermissionFlag.TEST2)
        model = TestModel(flag=flag)

        assert model.flag == flag
        assert model.model_dump() == {'flag': flag}
        assert model.model_dump(mode='json') == {'flag': str(flag)}
```

```Text
pydantic version: 2.0b2
                 install path: /workspaces/project/.venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.3 (main, Apr 12 2023, 14:31:14) [GCC 10.2.1 20210110]
                     platform: Linux-5.15.0-1038-azure-x86_64-with-glibc2.31
```

adriangb:
Fixes #5999

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -430,6 +430,11 @@ class GenerateSchema:
         if _typing_extra.is_dataclass(obj):
             return self._dataclass_schema(obj, None)
 
+        res = self._get_prepare_pydantic_annotations_for_known_type(obj, (), skip_if_has_get_core_schema=False)
+        if res is not None:
+            source_type, annotations = res
+            return self._apply_annotations(lambda x: x, source_type, annotations)
+
         origin = get_origin(obj)
 
         if isinstance(obj, TypeAliasType) or isinstance(origin, TypeAliasType):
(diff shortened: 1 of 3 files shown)


adriangb:
please review

adriangb:
Clearly an in depth refactor is needed here to organize the flow in `GenerateSchema`. We've got several boolean arguments like the one I'm adding here and it's not immediately clear what calls what. But this is a small change for now that fixes this particular bug. I think we should target this refactor for after the v2 release when we will undoubtedly get more bugs reported.

dmontagu (approved):

suharnikov:
@adriangb This place is a can of horrors. I tried to make a check for `__get_pydantic_core_schema__` in a slightly different place (in `enum_prepare_pydantic_annotations`), but then I noticed that other standard types may also be prone to this problem. Indeed, this place requires deep refactoring.

adriangb:
This only applies to enums. We don’t produce schemas for _subclasess_ of other standard types.

PR #6104, opened 2023-06-13
🐛 Fix RootModel json schema meta info

### LINKED ISSUE #6000: RootModel's `model_json_schema` miss field `title` and `description`
### Initial Checks

### Description

`SelloutList` expects a title named `SelloutList`, but gets a `None`

By default, we should use the name of the class as the title and the document of the class as the description.

### Example Code

```Python
from typing import List

class Sellout(BaseModel):
    """Sellout doc"""
    a: str
    b: int


class SelloutList(RootModel):
    """SelloutList doc"""
    root: List[Sellout]


print(Sellout.model_json_schema().get("title"))
# Sellout
print(Sellout.model_json_schema().get("description"))
# Sellout doc
print(SelloutList.model_json_schema().get("title"))
# None
print(SelloutList.model_json_schema().get("description"))
# None
```

```Text
pydantic version: 2.0b2
                 install path: D:\workspace\test\venv\Lib\site-packages\pydantic
               python version: 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)]
                     platform: Windows-10-10.0.22621-SP0
```

lig:
## Change Summary

Fix RootModel metadata handling

## Related issue number

Fixes #6000 

## Checklist

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -157,9 +157,19 @@ def modify_model_json_schema(
     json_schema = handler(schema_or_field)
     original_schema = handler.resolve_ref_schema(json_schema)
     if 'title' not in original_schema:
+        # Preserve the fact that definitions schemas should never have sibling keys:
+        if '$ref' in original_schema:
+            ref = original_schema['$ref']
+            original_schema.clear()
+            original_schema['allOf'] = [{'$ref': ref}]
         original_schema['title'] = cls.__name__
     docstring = cls.__doc__
     if docstring and 'description' not in original_schema:
+        # Preserve the fact that definitions schemas should never have sibling keys:
+        if '$ref' in original_schema:
+            ref = original_schema['$ref']
+            original_schema.clear()
+            original_schema['allOf'] = [{'$ref': ref}]
         original_schema['description'] = docstring
     return json_schema
 
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -271,7 +271,7 @@ class GenerateSchema:
                     post_init=getattr(cls, '__pydantic_post_init__', None),
                     config=core_config,
                     ref=model_ref,
-                    metadata={**metadata, **root_field['metadata']},
+                    metadata=metadata,
                 )
             else:
                 self._config_wrapper_stack.append(config_wrapper)
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -156,11 +156,6 @@ def modify_model_json_schema(
     """Add title and description for model-like classes' JSON schema."""
     json_schema = handler(schema_or_field)
     original_schema = handler.resolve_ref_schema(json_schema)
-    # Preserve the fact that definitions schemas should never have sibling keys:
-    if '$ref' in original_schema:
-        ref = original_schema['$ref']
-        original_schema.clear()
-        original_schema['allOf'] = [{'$ref': ref}]
     if 'title' not in original_schema:
         original_schema['title'] = cls.__name__
     docstring = cls.__doc__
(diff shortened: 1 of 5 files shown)


tests/test_root_model.py
```diff
@@ -401,3 +401,20 @@ class Model(BaseModel):
...
+
+    class SubclassedModel(RootModel):
+        """Subclassed Model docstring"""
+
+        root: int
+
+    parametrized_json_schema = ParametrizedModel.model_json_schema()
+    subclassed_json_schema = SubclassedModel.model_json_schema()
+
+    assert parametrized_json_schema.get('title') == 'RootModel[int]'
```
luolingchun, line 417: Why is the title not `ParametrizedModel`， but `RootModel[int]`?

```python
ParametrizedModel = RootModel[int]

ParametrizedModel2 = RootModel[int]
```

It would be strange if `ParameterizedModel` and `ParameterizedModel2` had the same title.

lig: This is basically an equivalent to

```py
class FirstModel(RootModel):
    root: int

SecondModel = FirstModel

assert SecondModel.model_json_schema()['title'] == 'FirstModel'
```

An assignment `ParametrizedModel = RootModel[int]` does not create a new object it just assigns the value of `RootModel[int]` to the variable `ParametrizedModel`.

Indeed, this test passes

```py
    ParametrizedModel = RootModel[int]
    ParametrizedModel2 = RootModel[int]

    assert ParametrizedModel is ParametrizedModel2
```

dmontagu: I'll note we may be able to support this better with `typing_extensions.TypeAliasType`, but @lig is right that without something like that there's no great way to know what the variable name is in python.

lig:
please review

tests/test_root_model.py
```diff
@@ -401,3 +401,20 @@ class Model(BaseModel):
 
     m = Model()
     assert m.value.root == 42
+
+
+def test_root_model_json_schema_meta():
+    ParametrizedModel = RootModel[int]
```
Kludex, line 407: Can this be done?
```suggestion
    ParametrizedModel = Annotated[RootModel[int], Field(description="potato")]
```

Kludex: Actually, can we add a test for this? 🤔

lig: Good idea! Wilco.

lig: This doesn't work actually. And I don't see how it could work. As far as I understand, we don't have control over what is outside RootModel here.

Kludex: I mean, does it work inside a model?

```py
class Model(BaseModel):
   a: Annotated[RootModel[int], Field(description="potato")]
```

Kludex: This is not a blocker for this PR btw, just wondering.

lig: Well, it works for RootModel when it acts as a field like it would for any field and it doesn't work when used on a RootModel when it's a final class like it wouldn't on any other model class.

pydantic/_internal/_generate_schema.py
```diff
@@ -157,9 +157,19 @@ def modify_model_json_schema(
...
+            original_schema.clear()
+            original_schema['allOf'] = [{'$ref': ref}]
         original_schema['title'] = cls.__name__
     docstring = cls.__doc__
     if docstring and 'description' not in original_schema:
+        # Preserve the fact that definitions schemas should never have sibling keys:
+        if '$ref' in original_schema:
+            ref = original_schema['$ref']
+            original_schema.clear()
+            original_schema['allOf'] = [{'$ref': ref}]
```
Kludex, line 172: Can we add a test for this?

https://smokeshow.helpmanual.io/2i1c2e311g3d3i5a3w6c/d_412a780b5b51bae3__generate_schema_py.html

Kludex (changes requested):

pydantic/_internal/_generate_schema.py
lig: That was added because of a failing test. Adding this has made test to pass.

Kludex: That is not being reached

image

lig: I've rewritten the code to be more DRY and added a test case. This should do the trick.

Kludex (approved):

PR #6107, opened 2023-06-13
Replace `deprecated-kwargs` with `removed-kwargs`

hramezani:
W‍‍e use `deprecated-kwargs` error code for removed kwargs. So, it's better to replace it with `removed-kwargs`.
Right now, there is no usecase for `deprecated-kwargs` error code but we can add it again in the future if we need it.

Selected Reviewer: @samuelcolvin`

### DIFF
--- a/docs/usage/errors.md
+++ b/docs/usage/errors.md
@@ -381,7 +381,7 @@ except PydanticUserError as exc_info:
     assert exc_info.code == 'config-both'
 ```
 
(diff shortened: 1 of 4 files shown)


hramezani:
please review

Kludex (approved):
For the record, can you link the issue/describe why this is needed?

hramezani:
> For the record, can you link the issue/describe why this is needed?

There is no issue for this. So, I've updated the PR description.

PR #6111, opened 2023-06-13
Make unannotated private attrs into private attrs

### LINKED ISSUE #5972: [PYD-129] _private attrs
Could you provide some info on the private model attributes, why it behaves differently when annotated or not:
```Python

class Foo(BaseModel):
    number: int
    _private_attr1: int = 42
    _private_attr2 = 42


x1 = Foo._private_attr1
x2 = Foo._private_attr2
print(x1, type(x1))
print(x2, type(x2))

# x1   ->  default=42 <class 'pydantic.fields.ModelPrivateAttr'>
# x2  ->  42 <class 'int'>

```

```Text
pydantic version: 2.0b1
                 install path: .venv/lib/python3.9/site-packages/pydantic
               python version: 3.9.15 (main, Dec  6 2022, 21:16:57)  [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.3.1-arm64-arm-64bit
```

<sub>[PYD-129](https://linear.app/pydantic/issue/PYD-129/private-attrs)</sub>

dmontagu:
Closes #5972 

I see three reasonable alternative choices for the behavior for leading-`_` attributes on BaseModel classes:

* `ClassVar`
* `PrivateAttr`
* raise an error, same as we would for public attributes

As this PR stands, I've made it so that they become a `PrivateAttr`. Mostly because it was super easy to do. But that can be changed easily.

However, in v1, these leading-underscore fields were _always_ classvars, even when they had an annotation, which perhaps should be the preferred implementation to reduce backwards compatibility issues? I'm not sure why we changed to defaulting to setting as a PrivateAttr in v2; this has led to some confusion as noted in this comment https://github.com/pydantic/pydantic/issues/5972#issuecomment-1573705778

I'll also note there has been a feature request to be able to use leading-`_` attributes as fields (e.g., for `_id` for use with mongodb), which is another angle to the problem not yet resolved.

@samuelcolvin do you remember the motivation behind making `_`-attributes private automatically?

### DIFF
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -101,7 +101,7 @@ def collect_model_fields(  # noqa: C901
         if _is_finalvar_with_default_val(ann_type, getattr(cls, ann_name, Undefined)):
             class_vars.add(ann_name)
             continue
-        if ann_name.startswith('_'):
+        if not is_valid_field_name(ann_name):
             continue
         if cls.__pydantic_root_model__ and ann_name != 'root':
             raise NameError(
@@ -209,3 +209,11 @@ def collect_dataclass_fields(
             field.apply_typevars_map(typevars_map, types_namespace)
 
     return fields
+
+
+def is_valid_field_name(name: str) -> bool:
+    return not name.startswith('_')
+
+
+def is_valid_privateattr_name(name: str) -> bool:
+    return name.startswith('_') and not name.startswith('__')
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -288,7 +288,7 @@ def inspect_namespace(  # noqa C901
         elif var_name.startswith('__'):
             continue
         elif var_name.startswith('_'):
-            if var_name in raw_annotations and not is_classvar(raw_annotations[var_name]):
+            if var_name not in raw_annotations or not is_classvar(raw_annotations[var_name]):
                 private_attributes[var_name] = PrivateAttr(default=value)
                 del namespace[var_name]
         elif var_name in base_class_vars:
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -16,7 +16,7 @@ from ..fields import Field, FieldInfo, ModelPrivateAttr, PrivateAttr
 from ._config import ConfigWrapper
 from ._core_utils import flatten_schema_defs, inline_schema_defs
 from ._decorators import ComputedFieldInfo, DecoratorInfos, PydanticDescriptorProxy
-from ._fields import Undefined, collect_model_fields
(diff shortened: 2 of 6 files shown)


samuelcolvin (approved):
Personally I'm happy with this. As explained in https://github.com/pydantic/pydantic/issues/5972#issuecomment-1593271034, the change was made to remove the need for `underscore_attrs_are_private`.

@dmontagu or @vitalik or anyone else, if you don't like this, happy to discuss alternatives.

dmontagu:
I'm fine with this, or with adding a generic annotation and requiring it (something like `x: PrivateAttribute[int] = ...`).

Either way, we end up committed to not changing it without introducing a config setting, I just know there has been some interest in being able to have fields with leading-`_`, which might be possible to support without introducing config if we required a `PrivateAttribute[...]` annotation, since otherwise it would be a breaking change.

153957:
> I'll also note there has been a feature request to be able to use leading-_ attributes as fields (e.g., for _id for use with mongodb), which is another angle to the problem not yet resolved.

We have the same request but for ArangoDB, in addition to `_id` (and `_key`) the edges in ArangoDB also use `_from` and `_to` fields to indicate which vertices are linked.

I think what is confusing to me is trying to figure out from docs and examples how exactly `PrivateAttr` and/or `_`-prefixed fields are handled differently from other attributes. Are they included in the dict, schema, etc? From my testing it appeared they were not.

For Pydantic v1 we only needed to override the `is_valid_field` to allow names starting with `_`, then it behaved as desired for us, I think right now we can not easily do the same for Pydantic v2, please correct me if I am wrong.

lig (approved):

dmontagu:
@153957 How were you overriding the `is_valid_field` function? Was it by monkey patching?

What I'm thinking we can do is to implement `is_valid_field_name` and `is_valid_privateattr_name` functions with similar usage to how `is_valid_field` was used in v1. I took a stab at this and felt it actually significantly improved the readability of the code anyway, and intend to add the changes into this PR. Then I would think you'd be able to use the same trick you used in v1.

That said, if you really are monkey patching the `is_valid_field` function, well, I don't intend to document this approach, encourage it, or promise it won't break in the future. But I don't see much harm in keeping something similar working for now.

I think longer term, a better solution would be that you can make non-leading-underscore attributes into private attributes by doing `x: int = PrivateAttr(...)`, and leading-underscore attributes into fields by doing `_x: int = Field(...)`. And I wouldn't consider that a breaking change. But probably not going to get that into the initial release of pydantic v2.

153957:
> @153957 How were you overriding the `is_valid_field` function? Was it by monkey patching?

Yes, like this:
```python
def is_valid_field(name: str):
    """[…]"""
    if not name.startswith('__'):
        return True
    return name == '__root__'


pydantic.main.is_valid_field = is_valid_field
```
Being able to continue to do this for v2 would be good enough for now 👍🏽.
Something in the future which has slightly clearer behaviour and documentation might be nice.

PR #6116, opened 2023-06-13
Add notes about getattr and delattr

### LINKED ISSUE #5946: BaseModel.__getattr__ ceased working with private attributes
### Initial Checks

### Description

BaseModel.\_\_getattr\_\_ is not usable any longer with private attributes - I guess related to #5891 .

https://github.com/pydantic/pydantic/blob/d91c1ee37a10ac4f968b100ec1821786f509e6b4/pydantic/_internal/_model_construction.py#L116-L117

### Example Code

```Python


class Target(BaseModel):
    data: str


class Reference(BaseModel):
    model_config = dict(
        extra="ignore"
    )
    _target: object = PrivateAttr()

    def __getattr__(self, item):
        print(f"getattr {item}")
        if item != "_target":
            return getattr(self._target, item)
        else:
            return getattr(self, item)

    def __setattr__(self, item, value):
        print(f"setattr {item}")
        if item != "_target":
            setattr(self._target, item, value)
        else:
            super().__setattr__(item, value)


r = Reference()
r._target = Target(data='test')
assert r._target.data == "test"
assert hasattr(r, "data")
```

```Text
pydantic version: 2.0a4
                 install path: ~/workspace/pydantic/pydantic
                     platform: Linux-5.19.0-42-generic-x86_64-with-glibc2.35
```

dmontagu:
Closes #5946

Also open to other solutions, such as not conditionally-overriding these methods, and just always paying the price of extra. But in #5946 @commonism said he was okay with this.

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -107,6 +107,21 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
         __pydantic_parent_namespace__: typing.ClassVar[dict[str, Any] | None]
         __pydantic_custom_init__: typing.ClassVar[bool]
         __pydantic_post_init__: typing.ClassVar[None | Literal['model_post_init']]
+
+        def __getattr__(self, item: str) -> Any:
+            """
+            If you override this in a class with private attributes or `model_config['extra'] == 'allow'`,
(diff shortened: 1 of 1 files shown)


dmontagu:
Closing in favor of #6165

PR #6117, opened 2023-06-13
✅ Update FastAPI test script

Kludex:
- Same as https://github.com/pydantic/pydantic/pull/6084 for 1.10.x.


samuelcolvin:
thanks :pray:.

Issue #2955, opened 2021-07-02
Deserialization of email with a dot in the name in NameEmail


ozeranskiy:
### Checks

* [x] I added a descriptive title to this issue
* [x] I have searched (google, github) for similar issues and couldn't find anything
* [x] I have read and followed [the docs](https://pydantic-docs.helpmanual.io/) and still think this is a bug

# Bug
If you try to deserialize a string like `firstname.lastname <firstname.lastname@example.com>` into a `NameEmail` object (as a pydantic model argument), you get an exception `pydantic.error_wrappers.ValidationError`.

Output of `python -c "import pydantic.utils; print(pydantic.utils.version_info())"`:
```
             pydantic version: 1.8.2
            pydantic compiled: False
                 install path: /<some_path>/pypoetry/virtualenvs/<some_project_name>/lib/python3.7/site-packages/pydantic
               python version: 3.7.10 (default, Apr 27 2021, 08:49:44)  [Clang 12.0.0 (clang-1200.0.32.29)]
                     platform: Darwin-20.5.0-x86_64-i386-64bit
     optional deps. installed: ['dotenv', 'email-validator', 'typing-extensions']
```

```py
from pydantic import BaseModel, NameEmail


class Message(BaseModel):
    to: NameEmail


Message(to='firstname.lastname <firstname.lastname@example.com>')
```
Traceback
```py
Traceback (most recent call last):
  File "/<some_path>/scratch_1.py", line 8, in <module>
    Message(to='firstname.lastname <firstname.lastname@example.com>')
  File "/<some_path>/pypoetry/virtualenvs/<some_project_name>/lib/python3.7/site-packages/pydantic/main.py", line 406, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for Message
to
  value is not a valid email address (type=value_error.email)
```

If the name is specified without a dot (for example, with a space), everything works as it should.

```py
from pydantic import BaseModel, NameEmail


class Message(BaseModel):
    to: NameEmail


msg = Message(to='firstname lastname <firstname.lastname@example.com>')

print(msg)
# output 
# to=NameEmail(name='firstname lastname', email='firstname.lastname@example.com')
```

String like `firstname.lastname <firstname.lastname@example.com>` can in real life be received if an email like `firstname.lastname@example.com` has been serialized from `NameEmail` to a string (as json value, for example).

```py
class Message(BaseModel):
    to: NameEmail

    class Config:
[...]

PR #6125, opened 2023-06-13
Improve the regex used for NameEmail validation

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/2955

While I agree it would be better to move the logic to rust as suggested in https://github.com/pydantic/pydantic/issues/2955#issuecomment-1414339014, in the interest of fixing this for users, I would suggest we just make this change now.

Happy to add more test cases, special callout to @henrybetts for doing the research and putting together an improved regex ages ago. I took an approach inspired by https://github.com/pydantic/pydantic/issues/2955#issuecomment-1147734780, and just cleaned up the regex a bit (and used a function to build it to get a bit better implicit documentation), and the tests all seem to be passing now as desired.

If anyone wants more tests cases let me know, but I think this covers the ones called out explicitly in the issue.

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -376,7 +376,15 @@ class IPvAnyNetwork:
         return cls(__input_value)  # type: ignore[return-value]
 
 
-pretty_email_regex = re.compile(r' *([\w ]*?) *<(.+?)> *')
+def _build_pretty_email_regex() -> re.Pattern:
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

Kludex (approved):

PR #6134, opened 2023-06-14
Improve `pydantic/_internal/_discriminated_union.py` docstring

hramezani:


hramezani:
please review

pydantic/_internal/_discriminated_union.py
```diff
@@ -12,6 +12,21 @@
...
+        discriminator: The name of the field which will serve as the discriminator.
+        definitions: A mapping of schema ref to schema.
+
+    Returns:
+        The new core schema.
+
+    Raises:
+        TypeError
+        ValueError
+        PydanticUserError
```
Kludex, line 28: Info missing here. 👀

Kludex (changes requested):

pydantic/_internal/_discriminated_union.py
```diff
@@ -12,6 +12,28 @@
...
+        definitions: A mapping of schema ref to schema.
+
+    Returns:
+        The new core schema.
+
+    Raises:
+        TypeError:
+            * If `discriminator` is used with invalid union variant.
+            * If `discriminator` is used with `Union` type with one variant.
+            * If `discriminator` value mapped to multiple choices.
```
Kludex, line 29: Do this look fine in the API documentation?

hramezani: not sure :). I can check. Do you have any suggestions?

pydantic/_internal/_discriminated_union.py
```diff
@@ -12,6 +12,28 @@
...
+        ValueError: If the definition for ref is missing.
```
dmontagu, line 30:
```suggestion
        ValueError:
            If the definition for ref is missing.
```
this may help it render better in the API docs, responding to Marcelo's point above. I haven't checked — maybe it's fine now, or maybe this doesn't actually help. But I think it at least makes it easier to read in the source code 😄

pydantic/_internal/_discriminated_union.py
hramezani: Ok, I've changed it and now it looks like:

pic

PR #6144, opened 2023-06-15
Fix definitions unpacking and registering when building a schema that uses a model

adriangb:
Selected Reviewer: @lig

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -360,9 +360,10 @@ class GenerateSchema:
 
         schema = self._unpack_refs_defs(schema)
 
-        if 'ref' in schema:
-            self.defs.definitions[schema['ref']] = schema
-            return core_schema.definition_reference_schema(schema['ref'])
+        ref: str | None = schema.get('ref', None)
+        if ref:
+            self.defs.definitions[ref] = schema
+            return core_schema.definition_reference_schema(ref)
         return schema
 
     def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -321,6 +321,10 @@ class GenerateSchema:
         return None
 
     def _unpack_refs_defs(self, schema: CoreSchema) -> CoreSchema:
+        """Unpack all 'definitions' schemas into `GenerateSchema.defs.definitions`
+        and return the inner schema.
+        """
+
         def get_ref(s: CoreSchema) -> str:
             return s['ref']  # type: ignore
 
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -34,6 +34,7 @@ from ._core_metadata import (
 from ._core_utils import (
     CoreSchemaOrField,
     define_expected_missing_refs,
+    flatten_schema_defs,
     get_type_ref,
     is_list_like_schema_with_items_schema,
     remove_unnecessary_invalid_definitions,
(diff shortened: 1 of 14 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -427,13 +444,9 @@ def _generate_schema(self, obj: Any) -> core_schema.CoreSchema:  # noqa: C901
...
             return self._callable_schema(obj)
         elif inspect.isclass(obj) and issubclass(obj, Enum):
-            from ._std_types_schema import enum_prepare_pydantic_annotations
+            from ._std_types_schema import get_enum_core_schema
 
-            # TODO: consider moving `enum_prepare_pydantic_annotations` out of `_std_types_schema`
-            res = enum_prepare_pydantic_annotations(obj, (), getattr(obj, '__pydantic_config__', ConfigDict()))
-            assert res is not None
-            source_type, annotations = res
-            return self._apply_annotations(lambda x: x, source_type, annotations)
```
adriangb, line 436: Cleaning up this TODO

pydantic/_internal/_generate_schema.py
```diff
@@ -1180,9 +1193,8 @@ def inner_handler(obj: Any) -> CoreSchema:
             get_inner_schema = self._get_wrapped_inner_schema(get_inner_schema, annotation, pydantic_js_functions)
 
         schema = get_inner_schema(source_type)
-        metadata_schema = resolve_original_schema(schema, self.defs.definitions)
```
adriangb, line 1183: Cleaning up the enum todo let us remove this line which really never should have been here in the first place. If you think about, we expressly _do not_ want to apply the metadata to the original schema.

pydantic/_internal/_schema_generation_shared.py
```diff
@@ -80,12 +80,14 @@ def generate_schema(self, __source_type: Any) -> core_schema.CoreSchema:
...
+            ref = maybe_ref_schema['schema_ref']
+            if ref not in self._generate_schema.defs.definitions:
+                raise LookupError(
+                    f'Could not find a ref for {ref}.'
+                    ' Maybe you tried to call resolve_ref_schema from within a recursive model?'
+                )
+            return self._generate_schema.defs.definitions[ref]
+        elif maybe_ref_schema['type'] == 'definitions':
+            return self.resolve_ref_schema(maybe_ref_schema['schema'])
+        return maybe_ref_schema
```
adriangb, line 93: This fixes the unpacking of `definitions` which happens in the test I added

pydantic/json_schema.py
```diff
@@ -404,7 +404,8 @@ def new_handler_func(
 
             current_handler = _schema_generation_shared.GenerateJsonSchemaHandler(self, new_handler_func)
 
-        return current_handler(schema)
+        json_schema = current_handler(schema)
+        return json_schema
```
adriangb, line 408:
```suggestion
        return current_handler(schema)
```

tests/test_json.py
```diff
@@ -77,12 +78,8 @@ class MyModel(BaseModel):
...
 )
 def test_json_serialization(ser_type, gen_value, json_output):
-    config_wrapper = ConfigWrapper({'arbitrary_types_allowed': False})
-    gen = GenerateSchema(config_wrapper, None)
-    schema = gen.generate_schema(ser_type)
-    schema = gen.collect_definitions(schema)
-    serializer = SchemaSerializer(schema)
-    assert serializer.to_json(gen_value()) == json_output
+    ta: TypeAdapter[Any] = TypeAdapter(ser_type)
+    assert ta.dump_json(gen_value()) == json_output
```
adriangb, line 82: This needed to be cleaned up anyway and kept a test from failing with the new changes. I didn't dig into why but I can if anyone cares enough.

tests/test_main.py
```diff
@@ -2444,3 +2444,26 @@ def set_x(m: 'OuterModel') -> 'OuterModel':
...
+
+
+def test_resolve_def_schema_from_core_schema() -> None:
+    class Inner(BaseModel):
+        x: int
+
+    class Marker:
+        def __get_pydantic_core_schema__(self, source_type: Any, handler: GetCoreSchemaHandler) -> CoreSchema:
+            schema = handler(source_type)
+            resolved = handler.resolve_ref_schema(schema)
```
adriangb, line 2456: This used to fail because at the point when it was called `GenerateSchema` had no knowledge of the `Inner ` definition.

PR #6157, opened 2023-06-15
Bug fix for forward refs in generics

### LINKED ISSUE #6130: list is not equivalent to typing.List under combination of generics and futures
### Initial Checks

### Description

This is a strange one - under the below conditions, exchanging `typing.List` for the more modern `list` produces the below error. I can't reproduce in a simpler example than this - it looks to me like a bug in pydantic:

```
  File "pydantic/validators.py", line 751, in pydantic.validators.find_validators

  File "{path_to_test}/test.py", line 11, in <module>
    outer = Outer["Inner"]
  File "{path_to_pydantic}/pydantic/generics.py", line 163, in __class_getitem__
    _prepare_model_fields(created_model, fields, instance_type_hints, typevars_map)
  File "{path_to_pydantic}/pydantic/generics.py", line 389, in _prepare_model_fields
    field.prepare()
  File "pydantic/fields.py", line 552, in pydantic.fields.ModelField.prepare
  File "pydantic/fields.py", line 758, in pydantic.fields.ModelField._type_analysis
  File "pydantic/fields.py", line 808, in pydantic.fields.ModelField._create_sub_type
  File "pydantic/fields.py", line 436, in pydantic.fields.ModelField.__init__
  File "pydantic/fields.py", line 557, in pydantic.fields.ModelField.prepare
  File "pydantic/fields.py", line 831, in pydantic.fields.ModelField.populate_validators
  File "pydantic/validators.py", line 760, in find_validators
RuntimeError: error checking inheritance of 'Inner' (type: str)
```

I've also displayed the minimal example below - thanks!

### Example Code

```Python
from typing import Generic, List, TypeVar

ChangingType = TypeVar("ChangingType")

class Outer(GenericModel, Generic[ChangingType]):
    fields_: list[ChangingType] # Swapping list -> List on this line fixes the bug

outer = Outer["Inner"]
class Inner(BaseModel):
    new: list[outer]

outer.update_forward_refs()
Inner.update_forward_refs()

out = Inner.parse_obj({"new": [{"fields_": []}]})

print(out)
```

```Text
pydantic version: 1.10.9
pydantic compiled: True
install path: /{package location}/.venv/lib64/python3.10/site-packages/pydantic
python version: 3.10.10 (main, Feb  8 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
platform: Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.35
```

### Affected Components

- [x] [Field Types](https://docs.pydantic.dev/usage/types/) - adding or changing a particular data type
- [X] [Generic Models](https://docs.pydantic.dev/usage/models/#generic-models)

mark-todd:
## Change Summary

Converts string to Forward Refs to allow for generic case
Also to fix discriminated unions allows Literals to pass through replace types unchanged.

## Related issue number

fix #6130

Selected Reviewer: @dmontagu

### DIFF
--- a/changes/6130-mark-todd.md
+++ b/changes/6130-mark-todd.md
@@ -1,0 +1,1 @@
+Fixed bug with generics receiving forward refs
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -267,7 +267,8 @@ def replace_types(type_: Any, type_map: Mapping[Any, Any]) -> Any:
     if origin_type is Annotated:
         annotated_type, *annotations = type_args
         return Annotated[replace_types(annotated_type, type_map), tuple(annotations)]
-
+    if origin_type is typing.Literal:
+        return type_map.get(type_, type_)
     # Having type args is a good indicator that this is a typing module
     # class instantiation or a generic alias of some sort.
     if type_args:
--- a/pydantic/generics.py
+++ b/pydantic/generics.py
@@ -31,6 +31,8 @@ from .utils import all_identical, lenient_issubclass
 
 if sys.version_info >= (3, 10):
     from typing import _UnionGenericAlias
+if sys.version_info >= (3, 8):
+    from typing import Literal
 
 GenericModelT = TypeVar('GenericModelT', bound='GenericModel')
 TypeVarType = Any  # since mypy doesn't allow the use of TypeVar as a type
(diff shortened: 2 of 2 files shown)


dmontagu:
I see this is breaking one of the discriminated union tests, do you have an idea of how to resolve that? If we can get that fixed I'd be happy to merge this.

mark-todd:
Hi David - so I've been trying to figure this out, this much I know:

The test used to fail for me in python3.10, then I added the check for is Literal (it was failing because the string in the literal was being converted to a ForwardRef)

I fixed it, but then when I uploaded I found it was producing the same bug in python3.9 and below, but not 3.10 or 3.11. I don't understand this because it should use the same function. It must be something to do with the conditional imports based on version that are changing the behaviour. I'll look into it some more and see what I can find but any thoughts on it would be appreciated!

mark-todd:
I've had a follow up thought on this - 3.9 or less does not have the list dict features, so it's possible there is a different part of the code that executes instead of replace args but that has a similar role.

mark-todd:
@dmontagu All tests fixed - should be good to merge

PR #6173, opened 2023-06-16
Improve `pydantic/main.py` docstrings

hramezani:


pydantic/main.py
```diff
@@ -84,6 +84,9 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
         __pydantic_parent_namespace__: Parent namespace of the model.
         __pydantic_custom_init__: Custom init of the model.
         __pydantic_post_init__: Post init of the model.
+        model_config: Configurations of model.
```
dmontagu, line 87:
```suggestion
        model_config: Configuration settings for the model.
```
Any chance we could move this to the top or bottom of the list of `__pydantic_*__` things? Since the name is a different format, I feel it would be nice if it weren't just in the middle like this

dmontagu (approved):
Please consider moving the location of `model_config` in the type hints at the top of BaseModel, but otherwise LGTM. (Feel free to merge when ready either way.)

hramezani:
> Please consider moving the location of `model_config` in the type hints at the top of BaseModel, but otherwise LGTM. (Feel free to merge when ready either way.)

I move both `model_fields` and `model_config` to the top

PR #6174, opened 2023-06-16
Add Examples annotation and re-organize existing annotations

adriangb:


adriangb:
please review

pydantic/json_schema.py
```diff
@@ -1658,3 +1660,59 @@ def _sort_json_schema(value: JsonSchemaValue) -> JsonSchemaValue:
...
+@_internal_dataclass.slots_dataclass
+class Examples:
+    """Add examples to a JSON schema.
+
+    Examples should be a map of example names (strings)
+    to example values (any valid JSON).
+
+    If `mode` is set this will only apply to that schema generation mode,
+    allowing you to add different examples for validation and serialization.
+    """
```
adriangb, line 1703: @tpdorsey any suggestions as to how to add docs/examples for these things? I see that they should be in the API docs but I'm not sure if I can put examples in there or if I need to make a new markdown block/section, etc.

pydantic/functional_serializers.py
```diff
@@ -227,3 +227,34 @@ def dec(f: Callable[..., Any]) -> _decorators.PydanticDescriptorProxy[Any]:
...
+    SerializeAsAny = Annotated[AnyType, ...]  # SerializeAsAny[list[str]] will be treated by type checkers as list[str]
+    """Force serialization to ignore whatever is defined in the schema and instead ask the object
+    itself how it should be serialized.
+    In particular, this means that model subclasses will include fields present in the subclass but not in the
+    original schema.
+    """
+else:
+
+    @_internal_dataclass.slots_dataclass
+    class SerializeAsAny:  # noqa: D101
```
adriangb, line 245: @lig moving this wasn't necessarily related to adding `Examples`, I'm happy to split up the PR if you'd like

pydantic/json_schema.py
tpdorsey: @adriangb check field.py and functional_validators.py for examples of examples in docstrings.

You don't have to use the "Example:" heading. That just puts the example in a pop-out admonition when built in the docs. You can also include the example code inline with the description part of the docstring.

Note that examples are tested even in docstrings.

pydantic/functional_serializers.py
```diff
@@ -227,3 +227,34 @@ def dec(f: Callable[..., Any]) -> _decorators.PydanticDescriptorProxy[Any]:
...
+
+AnyType = TypeVar('AnyType')
+
+
+if TYPE_CHECKING:
+    SerializeAsAny = Annotated[AnyType, ...]  # SerializeAsAny[list[str]] will be treated by type checkers as list[str]
+    """Force serialization to ignore whatever is defined in the schema and instead ask the object
+    itself how it should be serialized.
+    In particular, this means that model subclasses will include fields present in the subclass but not in the
+    original schema.
```
dmontagu, line 240:
```suggestion
    In particular, this means that when model subclasses are serialized, fields present in the subclass
    but not in the original schema will be included.
```

pydantic/json_schema.py
tpdorsey: I'm pretty sure about that tested part because I recall getting hung up on it recently.

adriangb: Okay I see: https://github.com/pydantic/pydantic/blob/beff6468d22c38adb950fed430db4066045e4272/pydantic/functional_validators.py#L31-L63

dmontagu (approved):
LGTM

pydantic/json_schema.py
adriangb: Bad news: these examples are not run. I put an `assert False` in one and it didn't work.
They also don't get formatted, type-checked, etc.

adriangb:
I'm going to merge this as is for now and we can figure out the docs later

PR #6180, opened 2023-06-19
💥 Use custom `PydanticDeprecationWarning` warning instead of the generic one

### LINKED ISSUE #5571: Add migration documentation for filterwarnings/DeprecationWarning
### Initial Checks

### Description

Having a list of all DeprecationWarning to use in pyproject/[tool.pytest.ini_options]/filterwarnings would allow taking care of things which actually break tests when moving to v2 and take care of the warnings once the real problems are resolved.

### Example Code

```Python
[tool.pytest.ini_options]
filterwarnings = [
# …
    # pydantic2
    'ignore:Support for "config" as "type" is deprecated and will be removed in a future version:DeprecationWarning',
    'ignore:Valid config keys have changed in V2:UserWarning',
    'ignore:The `update_forward_refs` method is deprecated; use `model_rebuild` instead.:DeprecationWarning',
    'ignore:The `parse_obj` method is deprecated; use `model_validate` instead.:DeprecationWarning',
    'ignore:The `construct` method is deprecated; use `model_construct` instead.:DeprecationWarning',
    'ignore:The `dict` method is deprecated; use `model_dump` instead.:DeprecationWarning',
    'ignore:`pydantic.config.Extra` is deprecated, use literal values instead .*$:DeprecationWarning',
    'ignore:Support for class-based `config` is deprecated, use ConfigDict instead.:DeprecationWarning',
    'ignore:The `__fields_set__` attribute is deprecated, use `model_fields_set` instead.:DeprecationWarning',
]
```

```Text
pydantic version: 2.0a2
        pydantic-core version: 0.23.1 release build profile
```

lig:
## Change Summary

Implement custom `PydanticDeprecationWarning`.
- Provides a single class to ignore or test for.
- Provides info on Pydantic version the deprecation was introduced in.
- Provides a link to the V2 migration guide for things deprecated in v2.0. 

## Related issue number

Fixes #5571

## Checklist

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -8,6 +8,7 @@ from typing_extensions import Literal, Self
 
 from ..config import ConfigDict, ExtraValues, JsonSchemaExtraCallable
 from ..errors import PydanticUserError
+from ..warnings import PydanticDeprecationWarning
 
 DEPRECATION_MESSAGE = 'Support for class-based `config` is deprecated, use ConfigDict instead.'
 
@@ -197,7 +198,7 @@ def prepare_config(config: ConfigDict | dict[str, Any] | type[Any] | None) -> Co
         return ConfigDict()
 
     if not isinstance(config, dict):
-        warnings.warn(DEPRECATION_MESSAGE, DeprecationWarning)
+        warnings.warn(PydanticDeprecationWarning(DEPRECATION_MESSAGE, since=(2, 0)))
         config = {k: getattr(config, k) for k in dir(config) if not k.startswith('__')}
 
     config_dict = cast(ConfigDict, config)
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -8,7 +8,7 @@ from typing_extensions import Literal, Self
 
 from ..config import ConfigDict, ExtraValues, JsonSchemaExtraCallable
 from ..errors import PydanticUserError
-from ..warnings import PydanticDeprecationWarning
+from ..warnings import PydanticDeprecatedSince20
 
 DEPRECATION_MESSAGE = 'Support for class-based `config` is deprecated, use ConfigDict instead.'
 
@@ -198,7 +198,7 @@ def prepare_config(config: ConfigDict | dict[str, Any] | type[Any] | None) -> Co
         return ConfigDict()
 
     if not isinstance(config, dict):
-        warnings.warn(PydanticDeprecationWarning(DEPRECATION_MESSAGE, since=(2, 0)))
+        warnings.warn(DEPRECATION_MESSAGE, PydanticDeprecatedSince20)
         config = {k: getattr(config, k) for k in dir(config) if not k.startswith('__')}
 
     config_dict = cast(ConfigDict, config)
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -10,6 +10,10 @@ from ..config import ConfigDict, ExtraValues, JsonSchemaExtraCallable
 from ..errors import PydanticUserError
 from ..warnings import PydanticDeprecatedSince20
 
+if not TYPE_CHECKING:
+    # See PyCharm issues PY-21915 and PY-51428
+    DeprecationWarning = PydanticDeprecatedSince20
+
 DEPRECATION_MESSAGE = 'Support for class-based `config` is deprecated, use ConfigDict instead.'
 
 
@@ -198,7 +202,7 @@ def prepare_config(config: ConfigDict | dict[str, Any] | type[Any] | None) -> Co
         return ConfigDict()
 
     if not isinstance(config, dict):
-        warnings.warn(DEPRECATION_MESSAGE, PydanticDeprecatedSince20)
+        warnings.warn(DEPRECATION_MESSAGE, DeprecationWarning)
         config = {k: getattr(config, k) for k in dir(config) if not k.startswith('__')}
 
     config_dict = cast(ConfigDict, config)
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -11,7 +11,8 @@ from ..errors import PydanticUserError
 from ..warnings import PydanticDeprecatedSince20
 
 if not TYPE_CHECKING:
-    # See PyCharm issues PY-21915 and PY-51428
+    # See PyCharm issues https://youtrack.jetbrains.com/issue/PY-21915
+    # and https://youtrack.jetbrains.com/issue/PY-51428
     DeprecationWarning = PydanticDeprecatedSince20
 
 DEPRECATION_MESSAGE = 'Support for class-based `config` is deprecated, use ConfigDict instead.'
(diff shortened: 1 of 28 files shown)


lig:
This needs an url for Migration Guide on `errors.pydantic.dev`

lig:
please review

commonism:
In case you do not want to keep the v2 migration warnings forever, maybe there any future benefit of Pydanticv2MigrationDeprecationWarning as subclass of PydanticDeprecationWarning?

lig:
@commonism what is a use case for that? When all `since=(2, 0)` will go we will remove the piece of code that adds the Migration Guide link.

commonism:
Based on expected_removal=(2,x) - Pydanticv2xRemovalWarning
Given Pydanticv20RemovalWarning, Pydanticv21RemovalWarning, Pydanticv22RemovalWarning.
Easy to search for when removing the `x-1` warnings from the code for release v2x.
Easy to ignore the v22 warnings by class name while keeping the v20/v21 warnings - without matching them by message.

lig:
hey me, please update

pydantic/_internal/_config.py
```diff
@@ -8,6 +8,11 @@
 
 from ..config import ConfigDict, ExtraValues, JsonSchemaExtraCallable
 from ..errors import PydanticUserError
+from ..warnings import PydanticDeprecatedSince20
+
+if not TYPE_CHECKING:
+    # See PyCharm issues PY-21915 and PY-51428
+    DeprecationWarning = PydanticDeprecatedSince20
```
lig, line 15: This makes it work in PyCharm. My concern is that we won't be able using more than one `DeprecationWarning` subclass in a single module file.

lig (commented):
please review

pydantic/warnings.py
```diff
@@ -0,0 +1,45 @@
...
+        self.expected_removal = expected_removal if expected_removal is not None else (since[0] + 1, 0)
+
+    def __str__(self) -> str:
+        message = (
+            f'{self.message}. Deprecated in Pydantic V{self.since[0]}.{self.since[1]}'
+            f' to be removed in V{self.expected_removal[0]}.{self.expected_removal[1]}.'
+        )
+        if self.since == (2, 0):
+            message += ' See Pydantic V2 Migration Guide at https://errors.pydantic.dev/v2.0/migration/'
+        return message
```
Kludex, line 38: The integration of this warning with the IDEs is also not ideal, since the `__str__` can't be used, and I'd expect us to have that in there. This includes both VSCode, and PyCharm.

I also think that each warning should link to the specific part of the documentation instead of a generic link to the guide (I don't think we have everything covered there), and it's likely that we are going to have multiple links on the same warning.

I suggest a simple implementation instead:
```suggestion
class PydanticDeprecationWarning(DeprecationWarning):
    ...
```

pydantic/warnings.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class PydanticDeprecatedSince20(PydanticDeprecationWarning):
+    """A specific `PydanticDeprecationWarning` subclass defining functionality deprecated since Pydantic 2.0."""
+
+    def __init__(self, message: str, *args: object) -> None:
+        super().__init__(message, *args, since=(2, 0))
```
Kludex, line 45:
```suggestion
class PydanticDeprecatedV2(PydanticDeprecationWarning):
    ...
```

pydantic/main.py
```diff
@@ -47,6 +48,9 @@
     Model = typing.TypeVar('Model', bound='BaseModel')
     # should be `set[int] | set[str] | dict[int, IncEx] | dict[str, IncEx] | None`, but mypy can't cope
     IncEx: typing_extensions.TypeAlias = 'set[int] | set[str] | dict[int, Any] | dict[str, Any] | None'
+else:
+    # See PyCharm issues PY-21915 and PY-51428
```
Kludex, line 52: The actual links would be more helpful than copy paste on google... 👀

pydantic/warnings.py
lig: > The integration of this warning with the IDEs is also not ideal, since the __str__ can't be used, and I'd expect us to have that in there. This includes both VSCode, and PyCharm.

This is true for any custom `DeprecationWarning`. The current `PydanticDeprecationWarning` implementation is compatible with the `DeprecationWarning` in terms of the first argument being a deprecation reason provided explicitly in the code. It is true that everything else is being added in runtime.

lig: > I suggest a simple implementation instead

Basically, you're suggesting throwing away the entire PR. Any particular reason for that?

Kludex: > Basically, you're suggesting throwing away the entire PR. Any particular reason for that?

Well, if you don't want the IDE support of "showing the entire message", then no need to change. 🤷‍♂️

samuelcolvin: I think it's fine for the IDE to not show the entire message.

pydantic/_internal/_config.py
samuelcolvin: 🤷, shame but no other way. Technically we could use multiple subclasses, we'd just need to assign different classes to `DeprecationWarning` through the file.

Ugly I know, but not the end of the world

pydantic/warnings.py
```diff
@@ -0,0 +1,45 @@
...
+        self.since = since
+        self.expected_removal = expected_removal if expected_removal is not None else (since[0] + 1, 0)
+
+    def __str__(self) -> str:
+        message = (
+            f'{self.message}. Deprecated in Pydantic V{self.since[0]}.{self.since[1]}'
+            f' to be removed in V{self.expected_removal[0]}.{self.expected_removal[1]}.'
+        )
+        if self.since == (2, 0):
+            message += ' See Pydantic V2 Migration Guide at https://errors.pydantic.dev/v2.0/migration/'
```
samuelcolvin, line 37: We should use the actual version from `version.py` here, not `v2.0`.

pydantic/warnings.py
```diff
@@ -0,0 +1,45 @@
...
+        return message
+
+
+class PydanticDeprecatedSince20(PydanticDeprecationWarning):
+    """A specific `PydanticDeprecationWarning` subclass defining functionality deprecated since Pydantic 2.0."""
+
+    def __init__(self, message: str, *args: object) -> None:
```
samuelcolvin, line 44: best to specifically include `expected_removal` here to make reading the signature easier.

samuelcolvin (dismissed):
overall LGTM, couple of things to change.

lig:
> You need to add the category for `@deprecated(<msg>, category=DeprecationWarning)`.

Actually, it's possible to use `PydanticDeprecatedSince20` explicitly in the decorator as VS Code supports that correctly.

pydantic/warnings.py
lig: Added it to the super call below.

lig:
please review

Kludex (approved):

PR #6185, opened 2023-06-19
Ignore deprecated methods of `pydantic.main` in pydocstyle check

hramezani:
Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -194,7 +194,7 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
 
         def __init_subclass__(cls, **kwargs: Unpack[ConfigDict]):
             """This signature is included purely to help type-checkers check arguments to class declaration, which
-            provides a way to conveniently set model_config key/value pairs:
+            provides a way to conveniently set model_config key/value pairs.
 
             ```py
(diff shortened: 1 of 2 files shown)


hramezani:
Please review

pydantic/main.py
```diff
@@ -929,7 +930,7 @@ def parse_file(
     #     "The `from_orm` method is deprecated; set "
```
Kludex, line 930: Since you are here, can you uncomment this? 🤔

Kludex: I think it was commented by mistake? 🤔

hramezani: probably it's not commented by mistake. it could be because of mypy error

hramezani: Take a look at your commit message here https://github.com/pydantic/pydantic/pull/5646/commits/32c468c4de12d1d85926987f0ab591370a397023

Kludex: Why only on `orm_mode`? 🤔

Kludex (approved):

pydantic/main.py
hramezani: I will take a look later. Now, I've reverted the change.

PR #6193, opened 2023-06-20
Complete `pydantic/root_model.py` docstring

hramezani:
Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/root_model.py
+++ b/pydantic/root_model.py
@@ -1,4 +1,4 @@
-from __future__ import annotations as _annotations
+from __future__ import annotations as _annotations  # noqa: D100
 
 import typing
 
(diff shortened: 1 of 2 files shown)


hramezani:
pleaes review

pydantic/root_model.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import annotations as _annotations
+from __future__ import annotations as _annotations  # noqa: D100
```
Kludex, line 1: Why?

Kludex (approved):

pydantic/root_model.py
hramezani: Actually, it was because we didn't have the doc at the first line of the file and my IDE fixed it by ignoring it.

Fixed by adding doc for the first line.

PR #6194, opened 2023-06-20
Fixed literal validator errors for unhashable values

### LINKED ISSUE #6188: Validating a dict value agains a Literal type gives the wrong ValidationError
### Initial Checks

### Description

Validating a dict value agains a Literal type gives the wrong ValidationError. Instead of `unexpected value; permitted: '*'` the validation error just says `unhashable type: 'dict' (type=type_error)`.  Which does not describe the problem and is confusing, e.g. if handing this to API users in fastAPI. I would also expect an `unexpected value; permitted: '*'` for all non string values.

The respective validator in `pydantic/validators.py` (line 480ff) only catches a KeyError:

```py
def make_literal_validator(type_: Any) -> Callable[[Any], Any]:

    # To have a O(1) complexity and still return one of the values set inside the `Literal`,
    # we create a dict with the set values (a set causes some problems with the way intersection works).
    # In some cases the set value and checked value can indeed be different (see `test_literal_validator_str_enum`)
    allowed_choices = {v: v for v in permitted_choices}

        try:
            return allowed_choices[v]
        except KeyError:

    return literal_validator
```

IMHO, it should also catch a  `TypeError` (i.e. the "unhashable type") and produce a corresponding error. Or, it could check for `isinstance(v, str)` first and raise respective error if it is not a string.

### Example Code

```Python
import pydantic
from typing import Literal

pydantic.parse_obj_as(Literal['*'], {})
```

```Text
pydantic version: 1.10.9
            pydantic compiled: True
                 install path: ********.pyenv/lib/python3.9/site-packages/pydantic
               python version: 3.9.16 (main, Dec  7 2022, 10:16:11)  [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.2-x86_64-i386-64bit
```

### Affected Components

markus1978:
## Change Summary

If a value is not hashable and used on the dict of expected values a `TypeError` is raised. The code now also catches the `TypeError` along side the `KeyError` to cause the validator to fail with a `WrongConstantError`. 

## Related issue number

fix #6188

## Checklist

### DIFF
--- a/pydantic/validators.py
+++ b/pydantic/validators.py
@@ -488,7 +488,7 @@ def make_literal_validator(type_: Any) -> Callable[[Any], Any]:
     def literal_validator(v: Any) -> Any:
         try:
             return allowed_choices[v]
-        except KeyError:
+        except (KeyError, TypeError):
             raise errors.WrongConstantError(given=v, permitted=permitted_choices)
 
     return literal_validator
--- a/changes/6188-markus1978.md
+++ b/changes/6188-markus1978.md
@@ -1,0 +1,1 @@
+Fixed literal validator errors for unhashable values.
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -1189,6 +1189,24 @@ def test_literal_validator():
     ]
 
 
+def test_literal_validator_non_str_value():
+    class Model(BaseModel):
+        a: Literal['foo']
+
+    Model(a='foo')
+
+    with pytest.raises(ValidationError) as exc_info:
+        Model(a={'bar': 'foo'})
+    assert exc_info.value.errors() == [
+        {
+            'loc': ('a',),
+            'msg': "unexpected value; permitted: 'foo'",
+            'type': 'value_error.const',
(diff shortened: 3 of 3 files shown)


pydantic/validators.py
```diff
@@ -486,6 +486,9 @@ def make_literal_validator(type_: Any) -> Callable[[Any], Any]:
     allowed_choices = {v: v for v in permitted_choices}
 
     def literal_validator(v: Any) -> Any:
+        if not isinstance(v, Hashable):
```
markus1978, line 489: Wasn't really sure what you prefer here. The alternative to this guard if, would be catching `TypeError`. Less readable, but would only cause overhead in the error case.

markus1978: Catching `TypeError` as well, would also remove the more or less duplicated core for raising the error. I changed it to catch a `TypeError` now.

hramezani (approved):

hramezani:
Thanks @markus1978

Issue #6192, opened 2023-06-20
`KeyError: 'required'` when generating schema using `typed_dict_field`


antonagestam:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

It seems like there's a mismatch between expectations of `typed_dict_field` when generating a JSON schema for a model, and its defaults. `SchemaValidator` seemingly interprets omitting the `required` argument as defaulting to `required=True` (see how validation does not pass in the example when omitting the field). But when trying to generate a JSON schema for a model that uses this core schema, it fails with a `KeyError`.

Seemingly related, the failing line ends with `# type: ignore  # required is always populated`.

### Example Code

```python
from pydantic import BaseModel, ValidationError
from pydantic_core import core_schema, SchemaValidator


schema = core_schema.typed_dict_schema(
    {
        "subunits": core_schema.typed_dict_field(
            core_schema.int_schema(),
            # Uncommenting this line makes the example pass.
            # required=True,
        ),
    }
)
validator = SchemaValidator(schema)
result = validator.validate_python({"subunits": 100})
print(f"{result=}")

try:
    result = validator.validate_python({})
except ValidationError:
    pass
else:
    assert False, "did not raise"


class CustomType:
    def __get_pydantic_core_schema__(self):
        return schema


class Model(BaseModel):
    t: CustomType


print(Model.model_json_schema())
```

Running the example outputs:

```
result={'subunits': 100}
Traceback (most recent call last):
  File "/Users/annevoab/projects/immoney/repro.py", line 34, in <module>
    print(Model.model_json_schema())
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/annevoab/.pyenv/versions/immoney/lib/python3.11/site-packages/pydantic/main.py", line 570, in model_json_schema
    return model_json_schema(
           ^^^^^^^^^^^^^^^^^^
  File "/Users/annevoab/.pyenv/versions/immoney/lib/python3.11/site-packages/pydantic/json_schema.py", line 1586, in model_json_schema
    return schema_generator_instance.generate(cls.__pydantic_core_schema__, mode=mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6200, opened 2023-06-20
Fix access of 'required' key for TypedDictField in GenerateJsonSchema

adriangb:
Fixes https://github.com/pydantic/pydantic/issues/6192

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -889,8 +889,9 @@ class GenerateJsonSchema:
         return self.generate_inner(schema['json_schema'])
 
     def typed_dict_schema(self, schema: core_schema.TypedDictSchema) -> JsonSchemaValue:
+        total = schema.get('total', True)
         named_required_fields: list[tuple[str, bool, CoreSchemaField]] = [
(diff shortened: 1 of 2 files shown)


pydantic/json_schema.py
```diff
@@ -1064,24 +1065,32 @@ def field_is_present(self, field: CoreSchemaField) -> bool:
...
+        total: bool,
     ) -> bool:
         """Whether the field should be marked as required in the generated JSON schema.
         (Note that this is irrelevant if the field is not present in the JSON schema.).
+
+        Args:
+            field: The schema for the field itself.
+            total: Only applies to `TypedDictField`s.
+                Indicates if the `TypedDict` this field belongs to is total, in which case any fields that don't
+                explicitly specify `required=False` are required.
```
hramezani, line 1079: You may can add `Returns` and `Raises` section to docstring.

hramezani:
Left a small suggestion. other than that LGTM 👍 
Thanks!

pydantic/json_schema.py
adriangb: Seems overkill to me

PR #6207, opened 2023-06-21
Improve `pydantic/mypy.py` docstring

hramezani:
Selected Reviewer: @lig

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -1,3 +1,5 @@
+"""This module contains related classes and functions for mypy plugin."""
+
 from __future__ import annotations
 
 import sys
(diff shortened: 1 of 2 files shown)


pydantic/mypy.py
```diff
@@ -760,15 +819,21 @@ def __init__(
         self.has_alias_generator = has_alias_generator
 
     def set_values_dict(self) -> dict[str, Any]:
```
hramezani, line 821: I think `get_values_dict` is a better name for this method.

What do you think?

hramezani:
please review

pydantic/mypy.py
lig: I'd rather agree. The current name looks like an accident to me.

pydantic/mypy.py
```diff
@@ -1,3 +1,5 @@
+"""This module contains related classes and functions for mypy plugin."""
```
lig, line 1:
```suggestion
"""This module includes classes and functions designed specifically for use with the mypy plugin."""
```

lig (approved):
👍🏻

suggestions are up to you

PR #6222, opened 2023-06-22
Replace if instead of elif in _internal/_fields.py

mo1ein:
## Change Summary
I think that when we have a `return` statement, we don't need `elif` or `else`.

## Checklist

### DIFF
--- a/changes/6222-mo1ein.md
+++ b/changes/6222-mo1ein.md
@@ -1,0 +1,1 @@
+Replace if instead of elif in _internal/_fields.py.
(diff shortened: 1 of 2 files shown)


samuelcolvin:
I agree with @PrettyWood, but thanks for the suggestion.

lig:
I'm totally agree that this PR reduces code complexity and I would prefer the code from PR instead of the current one.

However, there are a lot of similar places in the current code base. Some even use a mixture of techniques. I would argue for the Pydantic team to agree on the preferred way and all the cases would be refactored in a single PR.

At the moment I'd say that the change doesn't introduce any new functionality and with no official code style agreed on there is no benefit from changing the existent code in both directions.

PR #6236, opened 2023-06-23
Add warning about datamodel-code-generator to docs

hramezani:


docs/integrations/datamodel_code_generator.md
```diff
@@ -1,5 +1,11 @@
 # Code Generation with datamodel-code-generator
 
+!!! warning
+    [datamodel-code-generator still does not support Pydantic V2](https://github.com/koxudaxi/datamodel-code-generator/issues/803).
```
Kludex, line 4: Are we sure it doesn't work? I asked without confirming.

hramezani: Not sure. but as there is an open issue for that and also its [dependency on Pydantic V1](https://github.com/koxudaxi/datamodel-code-generator/blob/82e45f4f5d1801ae95cbc87bf355561bdfc088de/pyproject.toml#L74C12-L74C19), I think it does not work.

Kludex: It says "Likely requires" on the issue... 👀 🤷‍♂️

hramezani: Yeah, but probably it requires a new release.
Do I need to change the warning 🤔 ?

Kludex (approved):

PR #6243, opened 2023-06-24
Add type key to enums

adriangb:
Fixes https://github.com/tiangolo/fastapi/discussions/9709#discussioncomment-6247094

While the `type` is not strictly required, and in fact cannot be included for mixed type enums, it does seem that Swagger does not render schemas correctly without it.

Refs:
- https://json-schema.org/understanding-json-schema/reference/type.html
- https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -557,12 +557,18 @@ class GenerateJsonSchema:
             return {'const': expected[0]}
         else:
             types = {type(e) for e in expected}
-            if types == {
-                str,
-            }:
+            if types == {str}:
                 return {'enum': expected, 'type': 'string'}
-            elif types.issubset({int, float}):
+            elif types == {int}:
+                return {'enum': expected, 'type': 'integer'}
+            elif types == {float}:
                 return {'enum': expected, 'type': 'number'}
+            elif types == {bool}:
+                return {'enum': expected, 'type': 'boolean'}
+            elif types == {list}:
+                return {'enum': expected, 'type': 'array'}
+            elif types == {None}:
+                return {'enum': expected, 'type': 'null'}
             else:
                 return {'enum': expected}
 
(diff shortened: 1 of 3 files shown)


adriangb:
please review

pydantic/json_schema.py
```diff
@@ -556,7 +556,21 @@ def literal_schema(self, schema: core_schema.LiteralSchema) -> JsonSchemaValue:
...
+            elif types == {int}:
+                return {'enum': expected, 'type': 'integer'}
+            elif types == {float}:
+                return {'enum': expected, 'type': 'number'}
+            elif types == {bool}:
+                return {'enum': expected, 'type': 'boolean'}
+            elif types == {list}:
+                return {'enum': expected, 'type': 'array'}
+            elif types == {None}:
+                return {'enum': expected, 'type': 'null'}
```
Kludex, line 571: Is this a possible scenario?

Kludex (approved):
I don't think we should be trying to make OpenAPI 3.0 (OpenAPI version FastAPI uses) happy. I think FastAPI should migrate to the new Swagger 5.x, which supports OpenAPI 3.1 (which complies with JSON Schema Draft 2020-12 - version Pydantic v2 supports).

That said... This PR makes sense for the new Swagger as well. 😅 

You can see the new Swagger [here](https://editor-next.swagger.io/).
Screenshot 2023-06-24 at 18 39 39

And... Should we add the `pragma: no cover` on those?
Screenshot 2023-06-24 at 18 33 42

adriangb:
What we should do is add tests :). I am away from keyboard, but feel free to push if you want to.

pydantic/json_schema.py
adriangb: nope, I removed it

adriangb:
All covered now: 
image

PR #6247, opened 2023-06-25
💚 Add timeout on `publish_docs` job

Kludex:
The `publish_job` job got stuck yesterday. This will at least prevent that from happening again on that job.

There's no global timeout for GitHub Actions.

See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes.

Selected Reviewer: @dmontagu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,6 +326,7 @@ jobs:
 
   publish_docs:
     needs: [check]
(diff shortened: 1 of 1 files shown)


Kludex:
please review

hramezani (approved):

PR #6249, opened 2023-06-26
👷 Enable FastAPI tests

Kludex:


Kludex:
This PR is already working. The problem is that FastAPI is already broken based on what we have on `main`. 😅

Kludex:
This is a bit funny... 🤔 

I've created https://github.com/pydantic/pydantic-core/pull/693/ on `pydantic-core`, and https://github.com/tiangolo/fastapi/pull/9747 on `fastapi`.

The `test-fastapi` job will only pass when both `pydantic-core` PR, and `fastapi` PR are merged.

PR #6251, opened 2023-06-26
Run `pdm add -G settings "pydantic-settings>=2.0b1"`

m9810223:
## Change Summary
Make this command to install `pydantic-settings` with pydantic:
```shell
pip install pydantic[settings]
```

## Checklist

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -462,11 +462,11 @@ dependencies = [
 
 [[package]]
 name = "pydantic-settings"
-version = "2.0b1"
+version = "2.0b2"
 requires_python = ">=3.7"
 summary = "Settings management using Pydantic"
 dependencies = [
-    "pydantic>=2.0b1",
+    "pydantic>=2.0b3",
     "python-dotenv>=0.21.0",
 ]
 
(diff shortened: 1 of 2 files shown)


adriangb:
@samuelcolvin I personally like this idea. I actually think we should make extras the de-facto way to get `pydantic-settings` or `pydantic-extra-types`, primarily because of the tight coupling between pydantic and these other packages, and because it would give us more flexibility down the road to monorepo them or something like that. We could still keep `pydantic-settings` as it's own PyPi package but declare it as an implementation detail.

What do you think of adding this for now, and if you have any thoughts on making it the recommended approach as described above?

lig:
Usually, extras not just add extra packages but enable additional functionality in the same package. If `pydantic[settings]` would make `pydantic.settings` (notice the dot) working then it seems appropriate. If it just installing additional packages it looks counterintuitive to me.

adriangb:
That's a good point. I would be in favor of `pydantic.settings` for what it's worth, but then it's not as simple of a change as I was thinking, and maybe this PR is not worth it.

m9810223:
I am trying to find some specifications ...
- https://peps.python.org/pep-0508/
- https://peps.python.org/pep-0685/

What I think is, in v1, `BaseSettings` can be imported under `pydantic`.
With installing `pydantic[settings]`, it's easier to import it.

Maybe modifing [_dynamic_imports in pydantic/__init__](https://github.com/pydantic/pydantic/blob/9cfad5c3cee6a31d2bcb410feffff2afc4d0c489/pydantic/__init__.py#L192) is another achievement?

PR #6258, opened 2023-06-26
Document "wrap" validators

tpdorsey:
Add documentation for before, after, and wrap validators

## Related issue number

Closes PYD-98

## Checklist

### DIFF
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -175,11 +175,15 @@ except ValidationError as e:
     """
 ```
 
-Note that the type of validator dictates the order of validation, not the order in which validators are applied in an `Annotated` type or class definition. In the following example, `func2` will be called before `func1`.
+!!! note "Validation order"
+    Validation order matters. Within a given type, validation goes from right to left and back. That is, it goes
+    from right to left running all "before" validators (or calling into "wrap" validators), then left to right back out calling all "after" validators.
 
-```py test="skip" lint="skip" upgrade="skip"
-MyVal = Annotated[int, AfterValidator(func1), BeforeValidator(func2)]
-```
+    In the following example, `func2` will be called before `func1`.
+
+    ```py test="skip" lint="skip" upgrade="skip"
+    MyVal = Annotated[int, AfterValidator(func1), BeforeValidator(func2)]
+    ```
 
 A few more things to note:
 
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -1,3 +1,5 @@
+## Field validators
+
 Custom validation and complex relationships between objects can be achieved using the `@field_validator` decorator.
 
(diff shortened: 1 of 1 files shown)


docs/usage/validators.md
```diff
@@ -368,3 +368,52 @@ print(DemoDataclass())
...
+
+You can also use the  `mode` argument for `@field_validator` and `@model_validator` to specify when validation occurs.
+
+
+```py
+from typing import Annotated
+
+from pydantic import AfterValidator, BaseModel, ValidationError
+
+MyInt = Annotated[int, AfterValidator(lambda v: v + 1)]
```
adriangb, line 392: I think an example of `WrapValidator` would also be nice.

adriangb (approved):
I've found an area of confusion is the relationship between the order in `Annotated` or in the class definition (in this case inheritance is also relevant) and how the validators are run. E.g. how `Annotated[..., AfterValidator(func1), BeforeValidator(func2)]` `func2` will be called before `func1`.

Also worth explaining that you can use multiple before/after/wrap validators but only 1 plain validator.

docs/usage/validators.md
tpdorsey: @adriangb good callouts. I added these examples, and also consolidated my new section with "Pre and per-item validators" to reduce redundancy.

docs/usage/validators.md
```diff
@@ -159,12 +175,18 @@ except ValidationError as e:
     """
 ```
 
+Note that the type of validator dictates the order of validation, not the order in which validators are applied in an `Annotated` type or class definition. In the following example, `func2` will be called before `func1`.
```
adriangb, line 178: Need to clarify that within a given type they go from right to left and back. That is it goes from right to left running all before validators (or calling into wrap validators) and then left to right back out calling all after validators. Doesn’t need to be in this paragraph, it’s just that as it stands right now this says “the order in Annotated doesn’t matter” but it does.

docs/usage/validators.md
```diff
@@ -195,7 +217,43 @@ print(DemoModel(int_list=[3, 2, 1], name_list=['adrian g', 'David']))
 #> int_list=[1, 2, 3] name_list=['Adrian G', 'David']
 ```
 
-## Validate Always
+### Wrap validators
+
+Wrap validators are useful for cases where you want to validate a value, but also want to return a default value if validation fails.
+
+```py
```
adriangb, line 224: Did this come from a test? If so we can delete the test

adriangb (approved):
Just some small notes. Please merge whenever you feel ready.

docs/usage/validators.md
tpdorsey: No this is adapted from the docstring

PR #6262, opened 2023-06-27
Update docs for validation errors

dmontagu:
* Review and update `validation_errors.md`
* Fix a couple type errors in the conversion_table.py (noticed while trying to add examples to the `validation_errors.md`)
* Replace the direct use of the schema classes in GenerateSchema with schema functions for consistency with other schema creation (also noticed while trying to add examples to the `validation_errors.md`)
* Add a test that ensures that all error codes in `pydantic_core.core_schema.ErrorType` are either documented or manually removed from the expected values (this is done when the error cannot be produced by code paths currently exposed in pydantic, rather than pydantic-core)

This was brutal to go through, I think I got everything but wouldn't be surprised if there was still a couple minor inconsistencies or typos.

Selected Reviewer: @hramezani

### DIFF
--- a/docs/usage/validation_errors.md
+++ b/docs/usage/validation_errors.md
@@ -156,7 +156,7 @@ This error is also raised for strict fields when the input value is not an insta
 
 ## `callable_type`
 
-This error is raised when the input value is valid as a `Callable`:
+This error is raised when the input value is not valid as a `Callable`:
 
 ```py
 from typing import Any, Callable
--- a/docs/usage/validation_errors.md
+++ b/docs/usage/validation_errors.md
@@ -787,10 +787,10 @@ except ValidationError as exc:
 # from JSON
 i64_max = 9_223_372_036_854_775_807
 try:
-    Model.model_validate_json(json.dumps(-i64_max * 2))
+    Model.model_validate_json(json.dumps({'x': -i64_max * 2}))
 except ValidationError as exc:
     print(repr(exc.errors()[0]['type']))
-    #> 'dict_type'
+    #> 'int_parsing_size'
 ```
 
 ## `int_type`
@@ -1081,10 +1081,10 @@ class Model(BaseModel):
 
 
 try:
-    Model(x=None)
+    Model(x=BadMapping())
 except ValidationError as exc:
     print(repr(exc.errors()[0]['type']))
-    #> 'dict_type'
+    #> 'mapping_type'
 ```
 
 ## `missing`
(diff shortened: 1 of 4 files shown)


dmontagu:
please review

docs/usage/validation_errors.md
```diff
@@ -123,13 +147,16 @@ class Model(BaseModel):
...
+    print(repr(exc.errors()[0]['type']))
+    #> 'bytes_type'
 ```
 
+This error is also raised for strict fields when the input value is not an instance of `bytes`.
+
 ## `callable_type`
 
-This error is raised when the value is not a `Callable`.
+This error is raised when the input value is valid as a `Callable`:
```
hramezani, line 159:
```suggestion
This error is raised when the input value is not valid as a `Callable`:
```

docs/usage/validation_errors.md
```diff
@@ -380,34 +353,14 @@ class Model(BaseModel):
 
 try:
     Model(x=datetime(2000, 1, 1))
-except ValidationError as exc_info:
-    assert exc_info.errors()[0]['type'] == 'datetime_future'
-```
-
-## `datetime_naive`
```
hramezani, line 387: Why did you remove this error?

docs/usage/validation_errors.md
```diff
@@ -792,13 +754,48 @@ class Model(BaseModel):
...
+    print(repr(exc.errors()[0]['type']))
+    #> 'int_parsing_size'
+
+# from JSON
+i64_max = 9_223_372_036_854_775_807
+try:
+    Model.model_validate_json(json.dumps(-i64_max * 2))
+except ValidationError as exc:
+    print(repr(exc.errors()[0]['type']))
+    #> 'dict_type'
```
hramezani, line 793: we are in the `int_parsing_size` section, but the error here is `dict_type`

docs/usage/validation_errors.md
```diff
@@ -1069,14 +1081,15 @@ class Model(BaseModel):
 
 
 try:
-    Model(x=BadMapping())
-except ValidationError as exc_info:
-    assert exc_info.errors()[0]['type'] == 'mapping_type'
+    Model(x=None)
+except ValidationError as exc:
+    print(repr(exc.errors()[0]['type']))
+    #> 'dict_type'
```
hramezani, line 1087: Shouldn't it be `mapping_type`?

docs/usage/validation_errors.md
```diff
@@ -1165,29 +1180,18 @@ class Model(BaseModel):
...
-from pydantic import BaseModel
-
-
-class Model(BaseModel):
-    x: str
-
-
-Model.model_validate(Model(x='test'), strict=True)
+except ValidationError as exc:
+    print(repr(exc.errors()[0]['type']))
```
hramezani, line 1188: I think you forgot to add `#> model_class_type ` as `print` statement output.
I am curious how the test passes without this.

docs/usage/validation_errors.md
```diff
@@ -1494,13 +1516,59 @@ class Model(BaseModel):
...
+
+
+class Model(BaseModel):
+    x: NaiveDatetime
+
+
+try:
+    Model(x=datetime.now(tz=timezone.utc))
+except ValidationError as exc:
+    print(repr(exc.errors()[0]['type']))
```
hramezani, line 1566: same here. print statement output is missed

hramezani:
Great work @dmontagu 

I left some comments.

Please update

docs/usage/validation_errors.md
dmontagu: good catch, I'll fix it

docs/usage/validation_errors.md
dmontagu: yep need to fix that too

docs/usage/validation_errors.md
dmontagu: it means it's not hitting the error, which is why I changed to `print` from the way it was before — the old way, since it did an assert, you couldn't tell if it was hitting the error. Good catch, I'll fix this.

docs/usage/validation_errors.md
dmontagu: thanks

dmontagu:
@hramezani I think I addressed everything you pointed out. Please review

PR #6265, opened 2023-06-27
Improve types dict mapping doc

hramezani:

### DIFF
--- a/docs/usage/types/dicts_mapping.md
+++ b/docs/usage/types/dicts_mapping.md
@@ -63,6 +63,8 @@ except ValidationError as e:
     We therefore recommend using [typing-extensions](https://pypi.org/project/typing-extensions/) with Python 3.8 as well.
 
 Same as `dict` but Pydantic will validate the dictionary since keys are annotated.
+You can define `__pydantic_config__` to change the model inherited from `TypedDict`.
+See [Model Config](../model_config.md) for more details.
 
 ```py
 from typing import Optional
@@ -79,7 +81,7 @@ class UserIdentity(TypedDict, total=False):
 
 
 class User(TypedDict):
-    __pydantic_config__ = ConfigDict(extra='forbid')  # type: ignore
+    __pydantic_config__ = ConfigDict(extra='forbid')
 
     identity: UserIdentity
     age: int
--- a/docs/usage/types/dicts_mapping.md
+++ b/docs/usage/types/dicts_mapping.md
@@ -4,12 +4,55 @@
 : `dict(v)` is used to attempt to convert a dictionary;
   see `typing.Dict` below for sub-type constraints
 
-`typing.Dict`
-: see [Typing Iterables](#typing-iterables) below for more detail on parsing and validation
+```py
+from pydantic import BaseModel, ValidationError
 
-`subclass of typing.TypedDict`
-: Same as `dict` but _pydantic_ will validate the dictionary since keys are annotated.
-  See [Annotated Types](#annotated-types) below for more detail on parsing and validation
+
+class Model(BaseModel):
+    x: dict
+
+
+m = Model(x={'foo': 1})
+print(m.model_dump())
+#> {'x': {'foo': 1}}
+
+try:
+    Model(x='test')
+except ValidationError as e:
+    print(e)
+    """
+    1 validation error for Model
+    x
+      Input should be a valid dictionary [type=dict_type, input_value='test', input_type=str]
+    """
+```
+
+`typing.Dict`
+
+```py
+from typing import Dict
+
+from pydantic import BaseModel, ValidationError
+
+
+class Model(BaseModel):
+    x: Dict[str, int]
+
+
+m = Model(x={'foo': 1})
+print(m.model_dump())
+#> {'x': {'foo': 1}}
+
+try:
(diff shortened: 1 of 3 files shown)


hramezani:
please review

docs/usage/types/dicts_mapping.md
```diff
@@ -18,3 +61,67 @@
...
+
+
+# `total=False` means keys are non-required
+class UserIdentity(TypedDict, total=False):
+    name: Optional[str]
+    surname: str
+
+
+class User(TypedDict):
+    __pydantic_config__ = ConfigDict(extra='forbid')  # type: ignore
```
Kludex, line 82: Can we link the docs where we talk about the `__pydantic_config__`? I think on `config_models.md`.

Also, how does the `type: ignore` looks in the documentation?

docs/usage/types/dicts_mapping.md
```diff
@@ -18,3 +61,67 @@
...
+```py
+from typing import Optional
+
+from typing_extensions import TypedDict
+
+from pydantic import BaseModel, ConfigDict, ValidationError
+
+
+# `total=False` means keys are non-required
+class UserIdentity(TypedDict, total=False):
```
Kludex, line 76: You can use the `# (1)!` feature of mkdocs-material 👀

Kludex (changes requested):

docs/usage/types/dicts_mapping.md
Kludex: Apply if you want.

docs/usage/types/dicts_mapping.md
hramezani: > Can we link the docs where we talk about the __pydantic_config__? I think on config_models.md.

I've added a short description and a link to `config_models.md`

> Also, how does the type: ignore looks in the documentation?

It was ok but I've deleted it as it will pass without this as well.

Kludex (approved):

docs/usage/types/dicts_mapping.md
samuelcolvin: please create an issue about this, we should fix it.

docs/usage/types/dicts_mapping.md
```diff
@@ -18,3 +61,69 @@
...
+
+
+class User(TypedDict):
+    __pydantic_config__ = ConfigDict(extra='forbid')
+
+    identity: UserIdentity
+    age: int
+
+
+class Model(BaseModel):
```
samuelcolvin, line 90: better to use a `TypeAdapter` than an outer model here.

docs/usage/types/dicts_mapping.md
```diff
@@ -18,3 +61,69 @@
     Prior to Python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
     But required and optional fields are properly differentiated only since Python 3.9.
     We therefore recommend using [typing-extensions](https://pypi.org/project/typing-extensions/) with Python 3.8 as well.
+
+Same as `dict` but Pydantic will validate the dictionary since keys are annotated.
```
samuelcolvin, line 65: We need a better description of what a typeddict is here.

docs/usage/types/dicts_mapping.md
```diff
@@ -18,3 +61,69 @@
...
+You can define `__pydantic_config__` to change the model inherited from `TypedDict`.
+See [Model Config](../model_config.md) for more details.
+
+```py
+from typing import Optional
+
+from typing_extensions import TypedDict
+
+from pydantic import BaseModel, ConfigDict, ValidationError
```
samuelcolvin, line 74: I think this should be split into multiple code examples - the simplest showing the very basic use of `TypedDict`, then showing more advanced usage like Config` and `total=False`

docs/usage/types/types_fields.md
```diff
@@ -62,12 +62,13 @@ class UserIdentity(TypedDict, total=False):
 
 
 class User(TypedDict):
+    __pydantic_config__ = ConfigDict(extra='forbid')
+
     identity: UserIdentity
     age: int
 
 
 class Model(BaseModel):
```
samuelcolvin, line 71: this change would mean slightly different behaviour, are you sure this is right?

samuelcolvin (changes requested):
otherwise LGTM

docs/usage/types/dicts_mapping.md
hramezani: Added some descriptions inspired by `typing.TypedDcit` doc

docs/usage/types/dicts_mapping.md
hramezani: Done

docs/usage/types/types_fields.md
hramezani: we can keep this line here. but it does not relate to the `ValidationError` below(second validation error regarding extra `email` key in input values).

The goal is to show extra `email` key is not permitted. but the doc test is passing because it does not fall into the `except` block. but with  this change and adding `__pydantic_config__` to `User` class, it will work

hramezani:
@samuelcolvin I've addressed all your comments.
Please review

samuelcolvin (approved):

PR #6279, opened 2023-06-28
🏷️ Comply with all possible variations of `model_validator`

### LINKED ISSUE #6277: [PYD-136] ModelAfterValidator causing warnings from Pylance and mypy
### Initial Checks

### Description

I want to use an ModelAvterValidator via the `model_validator` decorator, however, both Pylance and mypy are issuing warnings when doing so. To make sure I was using it correctly, I tried using the example code from the documentation and I am experiencing the same issues there.
The code below results in the following two warnings:
```
Argument 1 has incompatible type "Callable[[UserModel, UserModel], Any]"; expected "Union[ModelAfterValidator, ModelAfterValidatorWithoutInfo]"  [arg-type] (mypy)

Argument of type "(cls: Self@UserModel, m: UserModel) -> UserModel" cannot be assigned to parameter of type "_AnyModeAfterValidator"
  Type "(cls: Self@UserModel, m: UserModel) -> UserModel" cannot be assigned to type "_AnyModeAfterValidator"
    Type "(cls: Self@UserModel, m: UserModel) -> UserModel" cannot be assigned to type "(self: _ModelType@__call__, __info: ValidationInfo) -> _ModelType@__call__"
      Parameter name mismatch: "self" versus "cls"
      Parameter 2: type "ValidationInfo" cannot be assigned to type "UserModel"
        "ValidationInfo" is incompatible with "UserModel"
    Type "(cls: Self@UserModel, m: UserModel) -> UserModel" cannot be assigned to type "(self: _ModelType@__call__) -> _ModelType@__call__"
      Parameter name mismatch: "self" versus "cls" (pylance)
```

When changing the parameter name `cls` to `self` as suggested by Pylance, the corresponding lines in the latter warning disappear, but nothing else happens and I am not sure whether that is what I am supposed to do there.

As my IDE I use VS Code OSS with the python, pylance, and mypy extensions, among others.

### Example Code

```Python


class UserModel(BaseModel):
    username: str
    password1: str
    password2: str

    def check_passwords_match(cls, m: "UserModel"):
        pw1 = m.password1
        pw2 = m.password2
        if pw1 is not None and pw2 is not None and pw1 != pw2:
            raise ValueError("passwords do not match")
        return m
```

```Text
pydantic version: 2.0b3
pydantic-core version: 0.39.0 release build profile
install path: <path-to-workspace>/env/lib/python3.11/site-packages/pydantic
platform: Linux-6.3.9-arch1-1-x86_64-with-glibc2.37
```

Selected Assignee: @hramezani

Selected Assignee: @Kludex

<sub>[PYD-136](https://linear.app/pydantic/issue/PYD-136/modelaftervalidator-causing-warnings-from-pylance-and-mypy)</sub>

Selected Assignee: @lig

Kludex:
- Closes #6277

I've used the following code to make sure neither `mypy` or `pyright` raise errors:

```py
from pydantic import BaseModel, model_validator, ValidationInfo


class UserModel(BaseModel):
    username: str
    password1: str
    password2: str

    def check_passwords_match(m: "UserModel") -> "UserModel":
        return m

    @classmethod
    def check_password_strength(cls, m: "UserModel") -> "UserModel":
        return m

    @staticmethod
    def check_password_strength2(m: "UserModel") -> "UserModel":
        return m

    def check_username(m: "UserModel", info: ValidationInfo) -> "UserModel":
        return m

    @classmethod
    def check_username2(cls, m: "UserModel", info: ValidationInfo) -> "UserModel":
        return m

    @staticmethod
    def check_username3(m: "UserModel", info: ValidationInfo) -> "UserModel":
        return m


user = UserModel(username="samuelcolvin", password1="zxcvbn", password2="zxcvbn")
print(user)
```


adriangb:
I think we should encourage:

```python
    def check_passwords_match(self: Self) -> Self:
        return self
```

over

```python
    def check_passwords_match(m: "UserModel") -> "UserModel":
        return m
```

Kludex:
> I think we should encourage:
> 
> ```python
>     @model_validator(mode="after")
>     def check_passwords_match(self: Self) -> Self:
>         return self
> ```
> 
> over
> 
> ```python
>     @model_validator(mode="after")
>     def check_passwords_match(m: "UserModel") -> "UserModel":
>         return m
> ```

Ok... 🤔

This PR doesn't say otherwise, does it?

adriangb:
Are you arguing that those are the same thing?

adriangb:
> If I understand correctly, in this case check_passwords_match is a static method?

There are some funky implementation details, but functionally it's the same as if it were an instance method. Why do you think it's a static method?

> but here we still have the argument named self

I don't see any `self` in your example, I don't know what you're talking about.

Kludex:
- Closed by https://github.com/pydantic/pydantic/pull/6514

Issue #6286, opened 2023-06-28
`typing.ClassVar` does not work with `from __future__ import annotations`


adriangb:
```python
from __future__ import annotations

from typing import ClassVar, Dict

from pydantic import BaseModel


class Parent(BaseModel):
    _FIELD_UPDATE_STRATEGY: ClassVar[Dict[str, int]] = {}

print(type(Parent._FIELD_UPDATE_STRATEGY))
#> <class 'pydantic.fields.ModelPrivateAttr'>
print(Parent._FIELD_UPDATE_STRATEGY)
#> default={}
```

Selected Assignee: @adriangb

PR #6287, opened 2023-06-28
Fix privateattr clobbering classvars with future annotations

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6286

### DIFF
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -355,6 +355,18 @@ def set_model_fields(
     cls.model_fields = fields
     cls.__class_vars__.update(class_vars)
 
+    for k in class_vars:
+        # Class vars should not be private attributes
+        #     We remove them _here_ and not earlier because we rely on inspecting the class to determine its classvars,
+        #     but private attributes are determined by inspecting the namespace _prior_ to class creation.
+        #     In the case that a classvar with a leading-'_' is defined via a ForwardRef (e.g., when using
+        #     `__future__.annotations`), we want to remove the private attribute which was detected _before_ we knew it
+        #     evaluated to a classvar
+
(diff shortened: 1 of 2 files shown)


pydantic/_internal/_model_construction.py
```diff
@@ -355,6 +355,18 @@ def set_model_fields(
...
+        # Class vars should not be private attributes
+        #     We remove them _here_ and not earlier because we rely on inspecting the class to determine its classvars,
+        #     but private attributes are determined by inspecting the namespace _prior_ to class creation.
+        #     In the case that a classvar with a leading-'_' is defined via a ForwardRef (e.g., when using
+        #     `__future__.annotations`), we want to remove the private attribute which was detected _before_ we knew it
+        #     evaluated to a classvar
+
+        value = cls.__private_attributes__.pop(k, None)
+        if value is not None and value.default is not PydanticUndefined:
+            setattr(cls, k, value.default)
```
adriangb, line 368: This doesn't look too bad, but just to play devil's advocate could we defer the private attribute stuff (or even the whole `set_model_fields` thing) until after class creation so that everything is happening at the same time?

adriangb (approved):

pydantic/_internal/_model_construction.py
dmontagu: I tried, it would have required a lot more rework of how the private attribute collection works. Which we could do, but I didn't want to waste time trying to unwind that now. Don't have to merge this now if we want to go back and try harder. I just wanted to finish figuring out the v2 release plan (/ work through anything else that's higher priority) first

adriangb: Well it's approved so feel free to merge or defer, whatever you feel is best

PR #6305, opened 2023-06-30
Change mypy tests to put comments in-line in a python module

dmontagu:
Doing this in preparation for the update to the plugin, to make it easier to review the outputs.

I've also dropped testing against pre-1.0 versions of mypy. Seems to me like now is a reasonable time to do that. (I'll add more versions, probably including 1.2.0, 1.3.0, 1.4.1, after this is merged.)

I'll try to open the PR for the plugin against this branch even before this gets merged so you can review.


dmontagu:
Also, I'll note that while I really wanted to put the comments in-line in the same file, and I actually implemented it that way first, but I couldn't come up with a reasonable way to do that across all plugin configurations, so I just implemented it like this where it uses a separate folder for each plugin config.

samuelcolvin (approved):
looks great (can't pretend I've reviewed every line)

PR #6307, opened 2023-06-30
PYD-131 Links between API and usage docs

samuelcolvin:
Currently not working for docstrings on functions, @pawamoy any suggestions on how to fix this?

### DIFF
--- a/docs/extra/tweaks.css
+++ b/docs/extra/tweaks.css
@@ -93,3 +93,19 @@ aside.blog img {
 .md-typeset__table > table tr:first-child td {
   border-top: none;
 }
+
+/* API documentation link admonition */
+:root {
+  --md-admonition-icon--api: url('data:image/svg+xml;charset=utf-8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path d="M7 7H5a2 2 0 0 0-2 2v8h2v-4h2v4h2V9a2 2 0 0 0-2-2m0 4H5V9h2m7-2h-4v10h2v-4h2a2 2 0 0 0 2-2V9a2 2 0 0 0-2-2m0 4h-2V9h2m6 0v6h1v2h-4v-2h1V9h-1V7h4v2Z"/></svg>')
+}
+.md-typeset .admonition.api, .md-typeset details.api {
+  border-color: #448aff;
+}
+.md-typeset .api > .admonition-title, .md-typeset .api > summary {
+  background-color: #448aff1a;
+}
+.md-typeset .api > .admonition-title::before, .md-typeset .api > summary::before {
+  background-color: #448aff;
+  -webkit-mask-image: var(--md-admonition-icon--api);
+          mask-image: var(--md-admonition-icon--api);
+}
--- a/docs/plugins/griffe_doclinks.py
+++ b/docs/plugins/griffe_doclinks.py
@@ -1,0 +1,49 @@
+import ast
+import re
+from functools import partial
+from pathlib import Path
+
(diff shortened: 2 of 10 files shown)


pawamoy:
If by "not working" you mean that you get AliasResolutionErrors or similar, you can try to skip aliases:

```python
def update_docstrings_recursively(obj: GriffeObject, python_file_name: str) -> None:
    if obj.docstring:
        obj.docstring.value = update_docstring(obj, python_file_name)
    for member in obj.members.values():
        if not member.is_alias:
            update_docstrings_recursively(member, python_file_name)
```

The pre-condition is that the first object you pass to `update_docstrings_recursively` is not itself an alias (which shouldn't be the case if you call it from `visit_module`).

pawamoy:
OK for some reason the condition mentioned above was not sufficient.

Tested this and it builds:

```python
def update_docstrings_recursively(obj: GriffeObject, python_file_name: str) -> None:
    if obj.is_alias:
        return
    if obj.docstring:
        obj.docstring.value = update_docstring(obj, python_file_name)
    for member in obj.members.values():
        update_docstrings_recursively(member, python_file_name)
```

samuelcolvin:
more to do, but I think we should merge this.

pawamoy:
Just a heads up letting you know that `VisitorExtension` and `InspectorExtension` will be deprecated in the next Griffe release. They were actually not flexible enough, and I reworked the extension system to make it more dev-friendly. Here is the updated code for the new, upcoming Extension:

```python
import ast
import re
from functools import partial
from pathlib import Path
from typing import Tuple

from griffe.dataclasses import Object as GriffeObject
from griffe.extensions import Extension
from griffe.agents.nodes import ObjectNode
from griffe.logger import get_logger
from pymdownx.slugs import slugify

DOCS_PATH = Path(__file__).parent.parent
slugifier = slugify(case='lower')
logger = get_logger(__name__)   # better than print

def find_heading(content: str, slug: str, file_path: Path) -> Tuple[str, int]:
    for m in re.finditer('^#+ (.+)', content, flags=re.M):
        heading = m.group(1)
        h_slug = slugifier(heading, '-')
        if h_slug == slug:
            return heading, m.end()
    raise ValueError(f'heading with slug {slug!r} not found in {file_path}')


def replace_links(m: re.Match, *, api_link: str) -> str:
    usage_path, slug = m.groups()
    rel_file = f'{usage_path}.md'
    file_path = DOCS_PATH / rel_file
    content = file_path.read_text()
    heading, heading_end = find_heading(content, slug, file_path)

    next_heading = re.search('^#+ ', content[heading_end:], flags=re.M)
    if next_heading:
        next_section = content[heading_end : heading_end + next_heading.start()]
    else:
        next_section = content[heading_end:]

    if f'[{api_link}]' not in next_section:
        logger.info(f'inserting API link "{api_link}" into {file_path.relative_to(DOCS_PATH)}')
        file_path.write_text(
            f'{content[:heading_end]}\n\n'
            '??? api "API Documentation"\n'
            f'    [`{api_link}`][{api_link}]<br>'
            f'{content[heading_end:]}'
        )

    return f'!!! abstract "Usage Documentation"\n    [{heading}](../{rel_file}#{slug})\n'


class DocLinksExtension(Extension):
    def on_instance(self, node: ast.AST | ObjectNode, obj: GriffeObject) -> None:
        if obj.docstring:
            obj.docstring.value = re.sub(
                r'usage[\- ]docs: ?https://docs\.pydantic\.dev/.+?/(.+?)/?#(\S+)',
                partial(replace_links, api_link=obj.path),
                obj.docstring.value,
                flags=re.I,
            ) 
```

PR #6316, opened 2023-06-30
Use exclude in pyproject.toml with black

dmontagu:
Putting it against the mypy plugin branch rather in that PR directly so we don't have to re-run that PR through CI.


samuelcolvin (approved):
LGTM

hramezani (approved):

PR #6324, opened 2023-06-30
Update the error message shown for protected namespaces

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6322

Is there a good way to link to a docs page in a way that will cause a test failure if the link breaks? Not sure if we can/should change the error type here to achieve that. I'm also okay just hardcoding a link to the docs about `ConfigDict` if it feels it would be useful. Or not — I think this at least helps people figure out there's a way to address this problem.

### DIFF
--- a/docs/usage/model_config.md
+++ b/docs/usage/model_config.md
@@ -588,7 +588,9 @@ try:
 except NameError as e:
     print(e)
     """
-    Field "model_prefixed_field" has conflict with protected namespace "model_"
+    Field "model_prefixed_field" has conflict with protected namespace "model_".
+
+    You may be able to resolve this error by setting `model_config['protected_namespaces'] = ()`.
     """
 ```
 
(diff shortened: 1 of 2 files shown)


docs/usage/model_config.md
```diff
@@ -610,7 +612,9 @@ try:
 except NameError as e:
     print(e)
     """
-    Field "also_protect_field" has conflict with protected namespace "also_protect_"
+    Field "also_protect_field" has conflict with protected namespace "also_protect_".
+
+    You may be able to resolve this error by setting `model_config['protected_namespaces'] = ('protect_me_',)`.
```
dmontagu, line 617: We could also only show this extra bit of the error message when `model_config['protected_namespaces'] == ('model_',)` but I don't think it hurts anything to show the full info, I could see it being useful if the protected namespaces were set on some parent class

hramezani (approved):

dmontagu:
Note that #6324 may be preferable; I think we should get @samuelcolvin's opinions before merging anything.

dmontagu:
Closing in favor of #6326

PR #6326, opened 2023-06-30
Warn instead of error protected namespace

dmontagu:
This is an alternative to #6324 also meant to address #6322 which I'm not sure would be preferable, but I wanted to at least share the work for discussion.

The biggest issue I see with this currently is that it won't error for `model_config` which feels like a not-unlikely source of collisions. But I don't currently see a way around that, as it may be desirable to give that attribute an annotation. (Open to suggestions there.)


dmontagu:
please review

tests/test_main.py
```diff
@@ -2251,33 +2252,50 @@ class Model(BaseModel):
 
 
 def test_protected_namespace_default():
-    with pytest.raises(NameError, match='Field "model_prefixed_field" has conflict with protected namespace "model_"'):
+    warnings.filterwarnings('error')
+    try:
+        with pytest.raises(
```
lig, line 2257: Maybe just `pytest.warns`? (also below)

lig (changes requested):

lig:
please update

tests/test_main.py
dmontagu: yep, no idea why I did  this here. I did need to do it in the docs because I wanted to print / not use pytest in the doc tests, but here, pytest.warns is definitely better.

dmontagu:
please review

lig (approved):
👍🏻

pydantic/_internal/_fields.py
```diff
@@ -109,7 +110,23 @@ def collect_model_fields(  # noqa: C901
             continue
         for protected_namespace in config_wrapper.protected_namespaces:
             if ann_name.startswith(protected_namespace):
-                raise NameError(f'Field "{ann_name}" has conflict with protected namespace "{protected_namespace}"')
+                valid_namespaces = tuple(x for x in config_wrapper.protected_namespaces if not ann_name.startswith(x))
```
hramezani, line 113: as `valid_namespaces` is used in `else` block, it might be better to move it there to prevent extra loop

PR #6327, opened 2023-06-30
Touch up Decimal validator

adriangb:
This improves performance slightly by not always going through Python, but error messages actually get less intelligible.

skip change file check

Selected Reviewer: @dmontagu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -303,56 +303,8 @@ jobs:
         with:
           jobs: ${{ toJSON(needs) }}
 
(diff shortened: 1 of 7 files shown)


adriangb:
please review

dmontagu (approved):
Seems fine to me if there's a good performance improvement but we really need to fix these error messages. In particular, I think there should be a way to have a union schema where you don't get an error for every possible case, among other things, as is it seems painfully verbose. But that's not a problem introduced by this PR

adriangb:
Agreed!

PR #6331, opened 2023-06-30
Update models.md

dmontagu:
Update the contents of the models.md file. 

This file definitely needs to be broken up, and the docs need to be reorganized a bit, but let's not do that now.

### DIFF
--- a/docs/usage/models.md
+++ b/docs/usage/models.md
@@ -1,10 +1,10 @@
 The primary means of defining objects in Pydantic is via models. Models are simply classes which inherit from
-`pydantic.BaseModel`.
+[`pydantic.BaseModel`][pydantic.BaseModel].
 
 You can think of models as similar to structs in languages like C, or as the requirements of a single endpoint
 in an API.
 
-Models share many similarities with Python's dataclasses, but have been designed with some subtle but important
+Models share many similarities with Python's dataclasses, but have been designed with some subtle-yet-important
 differences that streamline certain workflows related to validation, serialization, and JSON schema generation.
 You can find more discussion of this in the [Dataclasses](dataclasses.md) section of the docs.
 
@@ -262,7 +262,7 @@ print(pydantic_model.model_dump(by_alias=True))
 
 ### Nested attributes
 
-When using attributes to parse models, Model instances will be created from both top level attributes and
+When using attributes to parse models, model instances will be created from both top-level attributes and
 deeper-nested attributes as appropriate.
 
 Here is an example demonstrating the principle:
@@ -316,7 +316,7 @@ name='Anna' age=20.0 pets=[Pet(name='Bones', species='dog'), Pet(name='Orion', s
 Pydantic will raise `ValidationError` whenever it finds an error in the data it's validating.
 
 A single exception of type `ValidationError` will be raised regardless of the number of errors found,
-and that `ValidationError` will contain information about all the errors and how they happened.
+and that `ValidationError` will contain information about all of the errors and how they happened.
 
 See [Error Handling](../errors/errors.md) for details on standard and custom errors.
 
@@ -409,7 +409,7 @@ except ValidationError as e:
     """
 ```
 
-If you want to validate serialized data in a format besides JSON, you should load the data into a dict yourself and
+If you want to validate serialized data in a format other than JSON, you should load the data into a dict yourself and
 then pass it to `model_validate`.
 
 !!! note
@@ -555,7 +555,7 @@ except ValidationError as e:
     """
 ```
 
-If you set the `model_config` or make use of `@field_validator` or other pydantic decorators in your generic model
+If you set the `model_config` or make use of `@field_validator` or other Pydantic decorators in your generic model
 definition, they will be applied to parametrized subclasses in the same way as when inheriting from a `BaseModel`
 subclass. Any methods defined on your generic class will also be inherited.
 
@@ -563,7 +563,7 @@ Pydantic's generics also integrate properly with type checkers, so you get all t
 you would expect if you were to declare a distinct type for each parametrization.
 
 !!! note
-    Internally, pydantic creates subclasses of `BaseModel` at runtime when generic models are parametrized.
+    Internally, Pydantic creates subclasses of `BaseModel` at runtime when generic models are parametrized.
     These classes are cached, so there should be minimal overhead introduced by the use of generics models.
 
 To inherit from a generic model and preserve the fact that it is generic, the subclass must also inherit from
@@ -681,8 +681,8 @@ except ValidationError as e:
 When using bound type parameters, and when leaving type parameters unspecified, Pydantic treats generic models
 similarly to how it treats built-in generic types like `List` and `Dict`:
 
-* If you don't specify parameters before instantiating the generic model, they are treated as the bound of the `TypeVar`
-* If the `TypeVar`s involved have no bounds, they are treated as `Any`
+* If you don't specify parameters before instantiating the generic model, they are treated as the bound of the `TypeVar`.
+* If the `TypeVar`s involved have no bounds, they are treated as `Any`.
 
 Also, like `List` and `Dict`, any parameters specified using a `TypeVar` can later be substituted with concrete types.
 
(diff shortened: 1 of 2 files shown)


docs/usage/models.md
```diff
@@ -1,12 +1,13 @@
-!!! warning
-    This page still needs to be updated for v2.0.
-
 The primary means of defining objects in Pydantic is via models. Models are simply classes which inherit from
-`BaseModel`.
+`pydantic.BaseModel`.
```
tpdorsey, line 2:
```suggestion
[`pydantic.BaseModel`][pydantic.BaseModel].
```

docs/usage/models.md
```diff
@@ -1,12 +1,13 @@
...
 
-You can think of models as similar to types in strictly typed languages, or as the requirements of a single endpoint
+You can think of models as similar to structs in languages like C, or as the requirements of a single endpoint
 in an API.
 
+Models share many similarities with Python's dataclasses, but have been designed with some subtle but important
```
tpdorsey, line 7:
```suggestion
Models share many similarities with Python's dataclasses, but have been designed with some subtle-yet-important
```

docs/usage/models.md
```diff
@@ -256,11 +260,12 @@ print(pydantic_model.model_dump(by_alias=True))
     The example above works because aliases have priority over field names for
     field population. Accessing `SQLModel`'s `metadata` attribute would lead to a `ValidationError`.
 
-### Recursive models
+### Nested attributes
 
-Model instances will be parsed recursively as well as at the top level.
+When using attributes to parse models, Model instances will be created from both top level attributes and
```
tpdorsey, line 265:
```suggestion
When using attributes to parse models, model instances will be created from both top-level attributes and
```

docs/usage/models.md
```diff
@@ -310,8 +315,8 @@ name='Anna' age=20.0 pets=[Pet(name='Bones', species='dog'), Pet(name='Orion', s
 
 Pydantic will raise `ValidationError` whenever it finds an error in the data it's validating.
 
-One exception will be raised regardless of the number of errors found, that `ValidationError` will
-contain information about all the errors and how they happened.
+A single exception of type `ValidationError` will be raised regardless of the number of errors found,
+and that `ValidationError` will contain information about all the errors and how they happened.
```
tpdorsey, line 319:
```suggestion
and that `ValidationError` will contain information about all of the errors and how they happened.
```

docs/usage/models.md
```diff
@@ -404,6 +409,15 @@ except ValidationError as e:
     """
 ```
 
+If you want to validate serialized data in a format besides JSON, you should load the data into a dict yourself and
```
tpdorsey, line 412:
```suggestion
If you want to validate serialized data in a format other than JSON, you should load the data into a dict yourself and
```

docs/usage/models.md
```diff
@@ -541,18 +555,18 @@ except ValidationError as e:
     """
 ```
 
-If you set `Config` or make use of `validator` in your generic model definition, it is applied
-to concrete subclasses in the same way as when inheriting from `BaseModel`. Any methods defined on
-your generic class will also be inherited.
+If you set the `model_config` or make use of `@field_validator` or other pydantic decorators in your generic model
```
tpdorsey, line 558:
```suggestion
If you set the `model_config` or make use of `@field_validator` or other Pydantic decorators in your generic model
```

docs/usage/models.md
```diff
@@ -541,18 +555,18 @@ except ValidationError as e:
...
 
-Pydantic's generics also integrate properly with mypy, so you get all the type checking
-you would expect mypy to provide if you were to declare the type without using `GenericModel`.
+Pydantic's generics also integrate properly with type checkers, so you get all the type checking
+you would expect if you were to declare a distinct type for each parametrization.
 
 !!! note
-    Internally, pydantic uses `create_model` to generate a (cached) concrete `BaseModel` at runtime,
-    so there is essentially zero overhead introduced by making use of `GenericModel`.
+    Internally, pydantic creates subclasses of `BaseModel` at runtime when generic models are parametrized.
```
tpdorsey, line 566:
```suggestion
    Internally, Pydantic creates subclasses of `BaseModel` at runtime when generic models are parametrized.
```

docs/usage/models.md
```diff
@@ -664,11 +678,11 @@ except ValidationError as e:
...
 ```
 
-Pydantic also treats `GenericModel` similarly to how it treats built-in generic types like `List` and `Dict` when it
-comes to leaving them unparameterized, or using bounded `TypeVar` instances:
+When using bound type parameters, and when leaving type parameters unspecified, Pydantic treats generic models
+similarly to how it treats built-in generic types like `List` and `Dict`:
 
-* If you don't specify parameters before instantiating the generic model, they will be treated as `Any`
-* You can parametrize models with one or more *bounded* parameters to add subclass checks
+* If you don't specify parameters before instantiating the generic model, they are treated as the bound of the `TypeVar`
```
tpdorsey, line 684:
```suggestion
* If you don't specify parameters before instantiating the generic model, they are treated as the bound of the `TypeVar`.
```

docs/usage/models.md
```diff
@@ -664,11 +678,11 @@ except ValidationError as e:
...
+* If the `TypeVar`s involved have no bounds, they are treated as `Any`
```
tpdorsey, line 685:
```suggestion
* If the `TypeVar`s involved have no bounds, they are treated as `Any`.
```

docs/usage/models.md
```diff
@@ -878,11 +886,13 @@ print(items)
 
 ## `RootModel` and custom root types
 
-Pydantic models can be defined with a custom root type by declaring the `RootModel`.
+Pydantic models can be defined with a "custom root type" by subclassing `pydantic.RootModel`.
```
tpdorsey, line 889:
```suggestion
Pydantic models can be defined with a "custom root type" by subclassing [`pydantic.RootModel`][pydantic.RootModel].
```

docs/usage/models.md
```diff
@@ -878,11 +886,13 @@ print(items)
...
 
-The root type can be any type supported by pydantic, and is specified by the type hint of `RootModel`.
+The root type can be any type supported by pydantic, and is specified by the generic parameter to `RootModel`.
```
tpdorsey, line 891:
```suggestion
The root type can be any type supported by Pydantic, and is specified by the generic parameter to `RootModel`.
```

docs/usage/models.md
```diff
@@ -936,13 +947,36 @@ print([pet for pet in pets])
...
+print(my_pets.describe())
+#> Pets: dog, cat
+```
+
+
 ## Faux immutability
 
-Models can be configured to be immutable via `frozen = True`. When this is set, attempting to change the
+Models can be configured to be immutable via `model_config['frozen'] = True`. When this is set, attempting to change the
 values of instance attributes will raise errors. See [model config](model_config.md) for more details on `Config`.
```
tpdorsey, line 975:
```suggestion
values of instance attributes will raise errors. See [Model Config](model_config.md) for more details.
```

docs/usage/models.md
```diff
@@ -1148,20 +1153,39 @@ assert m1.uid != m2.uid
...
-!!! warning
-    The `default_factory` expects the field type to be set.
+You can find more information in the documentation of the [`Field` function](fields.md).
 
 ## Automatically excluded attributes
 
-Class variables which begin with an underscore and attributes annotated with `typing.ClassVar` will be
-automatically excluded from the model.
+### Class vars
+Attributes annotated with `typing.ClassVar` will are properly treated by Pydantic as class variables, and will not
```
tpdorsey, line 1161:
```suggestion
Attributes annotated with `typing.ClassVar` are properly treated by Pydantic as class variables, and will not
```

docs/usage/models.md
```diff
@@ -1260,10 +1284,12 @@ print(inspect.signature(MyModel))
 ```
 
 To be included in the signature, a field's alias or name must be a valid Python identifier.
-Pydantic prefers aliases over names, but may use field names if the alias is not a valid Python identifier.
+Pydantic will prioritize a field's alias over its name when generating the signature, but may use the field name if the
+alias is not a valid Python identifier.
 
-If a field's alias and name are both invalid identifiers, a `**data` argument will be added.
-In addition, the `**data` argument will always be present in the signature if `Config.extra` is `'allow'`.
+If a field's alias and name are _both_ invalid identifiers (which may be possible through exotic use of `create_model`),
```
tpdorsey, line 1290:
```suggestion
If a field's alias and name are _both_ not valid identifiers (which may be possible through exotic use of `create_model`),
```

tpdorsey (approved):
Some minor suggested edits. Looks good.

PR #6336, opened 2023-07-01
Expose ability to generate a CoreSchemaOrField for json schema building from a FieldInfo

dmontagu:
cc @tiangolo @adriangb 

@tiangolo would be good if you could confirm whether this would do the trick for you. I think @adriangb might want it addressed in a different way, I also understand that perspective. I was surprised by how little code it took to do this though (and how it affected basically nothing else).

Also, to be clear, I don't think this is fully ready for primetime, probably could use some minor tweaks/improved tests/a bit of documentation, but I also think there's a good chance it needs substantial revisions before merging and/or will be rejected, so I don't want to do that until it's confirmed.

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -265,6 +265,9 @@ class GenerateSchema:
             obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
         )
 
+    def generate_field_schema(self, name: str, field_info: FieldInfo) -> core_schema.ModelField:
+        return self._generate_md_field_schema(name, field_info, DecoratorInfos())
+
     def _generate_schema_for_type(
         self,
         obj: Any,
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -223,7 +223,7 @@ class GenerateJsonSchema:
         return mapping
 
     def generate_definitions(
-        self, inputs: Sequence[tuple[JsonSchemaKeyT, JsonSchemaMode, core_schema.CoreSchema]]
+        self, inputs: Sequence[tuple[JsonSchemaKeyT, JsonSchemaMode, CoreSchemaOrField]]
     ) -> tuple[dict[tuple[JsonSchemaKeyT, JsonSchemaMode], DefsRef], dict[DefsRef, JsonSchemaValue]]:
         """Generates JSON schema definitions from a list of core schemas, pairing the generated definitions with a
         mapping that links the input keys to the definition references.
@@ -233,7 +233,7 @@ class GenerateJsonSchema:
 
                 - The first element is a JSON schema key type.
                 - The second element is the JSON mode: either 'validation' or 'serialization'.
-                - The third element is a core schema.
+                - The third element is a core schema or core schema field.
 
         Returns:
             A sequence of tuples, where:
@@ -272,7 +272,7 @@ class GenerateJsonSchema:
         self._used = True
         return refs_map, _sort_json_schema(self.definitions)  # type: ignore
 
-    def generate(self, schema: CoreSchema, mode: JsonSchemaMode = 'validation') -> JsonSchemaValue:
+    def generate(self, schema: CoreSchemaOrField, mode: JsonSchemaMode = 'validation') -> JsonSchemaValue:
         """Generates a JSON schema for a specified schema in a specified mode.
 
         Args:
--- a/pydantic/type_adapter.py
+++ b/pydantic/type_adapter.py
@@ -6,6 +6,7 @@ from dataclasses import is_dataclass
 from typing import TYPE_CHECKING, Any, Dict, Generic, Iterable, Set, TypeVar, Union, overload
 
 from pydantic_core import CoreSchema, SchemaSerializer, SchemaValidator, Some
+from pydantic_core.core_schema import ModelField
 from typing_extensions import Literal, is_typeddict
 
 from pydantic.errors import PydanticUserError
@@ -13,6 +14,7 @@ from pydantic.main import BaseModel
 
 from ._internal import _config, _core_utils, _discriminated_union, _generate_schema, _typing_extra
 from .config import ConfigDict
+from .fields import FieldInfo
 from .json_schema import (
     DEFAULT_REF_TEMPLATE,
     DefsRef,
@@ -77,6 +79,18 @@ def _get_schema(type_: Any, config_wrapper: _config.ConfigWrapper, parent_depth:
     gen = _generate_schema.GenerateSchema(config_wrapper, types_namespace=global_ns, typevars_map={})
     schema = gen.generate_schema(type_)
     schema = gen.collect_definitions(schema)
+    return schema
+
+
+def _get_field_schema(
+    name: str, field_info: FieldInfo, config_wrapper: _config.ConfigWrapper, parent_depth: int
+) -> ModelField:
+    local_ns = _typing_extra.parent_frame_namespace(parent_depth=parent_depth)
+    global_ns = sys._getframe(max(parent_depth - 1, 1)).f_globals.copy()
+    global_ns.update(local_ns or {})
+    gen = _generate_schema.GenerateSchema(config_wrapper, types_namespace=global_ns, typevars_map={})
+    schema = gen.generate_field_schema(name, field_info)
+    schema['schema'] = gen.collect_definitions(schema['schema'])
     return schema
 
 
@@ -170,6 +184,14 @@ class TypeAdapter(Generic[T]):
         self.core_schema = core_schema
         self.validator = validator
         self.serializer = serializer
+
+    @classmethod
+    def model_field_schema(
+        cls, name: str, field: FieldInfo, config: ConfigDict | None = None, _parent_depth: int = 2
+    ) -> ModelField:
+        """Intended for use building JSON schemas from FieldInfos"""
+        config_wrapper = _config.ConfigWrapper(config)
+        return _get_field_schema(name, field, config_wrapper, parent_depth=_parent_depth + 1)
 
     def validate_python(
         self,
--- a/tests/test_json_schema.py
+++ b/tests/test_json_schema.py
@@ -4886,3 +4886,59 @@ def test_build_schema_for_fieldinfo():
         'examples': ['yes please', 'oh, no'],
         'type': 'string',
     }
+
+
+def test_build_schemas_from_field_info_with_name_collisions(create_module):
+    module1 = create_module(
+        # language=Python
+        """
+from pydantic import BaseModel
+
+
+class Model(BaseModel):
+    x: int
+    """
+    )
+
+    module2 = create_module(
+        # language=Python
+        """
+from pydantic import BaseModel
+
+
+class Model(BaseModel):
+    y: float
+    """
+    )
+
+    field1 = FieldInfo(annotation=module1.Model)
+    field2 = FieldInfo(annotation=module2.Model)
+
+    model_field_schema1 = TypeAdapter.model_field_schema('model1', field1, config=None)
+    model_field_schema2 = TypeAdapter.model_field_schema('model2', field2, config=None)
+
+    assert GenerateJsonSchema().generate_definitions(
+        [
+            (module1.Model, 'validation', model_field_schema1),
+            (module2.Model, 'validation', model_field_schema2),
+        ]
+    ) == (
+        {
+            (module1.Model, 'validation'): f'{module1.__name__}__Model',
+            (module2.Model, 'validation'): f'{module2.__name__}__Model',
+        },
+        {
+            f'{module1.__name__}__Model': {
+                'properties': {'x': {'title': 'X', 'type': 'integer'}},
+                'required': ['x'],
+                'title': 'Model',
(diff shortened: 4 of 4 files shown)


tiangolo (approved):
Yes! This works for me! 🎉 

I was playing with this branch locally, there's one little "inconsistency", but I just discovered that it's actually fine.

---

A test is worth a thousand words:

```Python
from pydantic import BaseModel, Field, TypeAdapter
from pydantic.fields import FieldInfo
from pydantic.json_schema import GenerateJsonSchema


class SubItem(BaseModel):
    name: str


class Item(BaseModel):
    sub: Annotated[SubItem, Field(description="sub item", examples=[{"name": "foo"}])]


def test_model_json_schema():
    # insert_assert(Item.model_json_schema())
    assert Item.model_json_schema() == {
        "$defs": {
            "SubItem": {
                "properties": {"name": {"title": "Name", "type": "string"}},
                "required": ["name"],
                "title": "SubItem",
                "type": "object",
            }
        },
        "properties": {
            "sub": {
                "allOf": [{"$ref": "#/$defs/SubItem"}],
                "description": "sub item",
                "examples": [{"name": "foo"}],
            }
        },
        "required": ["sub"],
        "title": "Item",
        "type": "object",
    }


def test_build_schema_for_fieldinfo():
    field_info = FieldInfo.from_annotation(
        Annotated[SubItem, Field(description="sub item", examples=[{"name": "foo"}])]
    )

    model_field_schema = TypeAdapter.model_field_schema("sub", field_info)
    # insert_assert(GenerateJsonSchema().generate_inner(model_field_schema))
    assert GenerateJsonSchema().generate_inner(model_field_schema) == {
        "$ref": "#/$defs/SubItem",
        "description": "sub item",
        "examples": [{"name": "foo"}],
    }
```

(this test passes, but has a tiny inconsistency in the layout of the data)

I'm using `generator.generate_inner()` becasue that's what I actually use (and need) in FastAPI, to be able to get schemas that use `$ref` and that are part of a bigger schema (OpenAPI, including all the JSON Schemas, to be referenced with `$ref`).

The little inconsistency is that in an equivalent Pydantic model, the field ends up with:

```
"allOf": [{"$ref": "#/$defs/SubItem"}]
```

I think this is because a previous version of JSON Schema didn't allow `$ref` to have sibling keys.

but in the `TyperAdapter` version it ends up with:

```
"$ref": "#/$defs/SubItem"
```

Nevertheless, the latest JSON Schema supports having `$ref` with siblings, so this is actually correct: https://json-schema.org/draft/2020-12/json-schema-core.html#appendix-G-2.6.1.10

And it renders fine in Swagger UI with OpenAPI. So this would definitely work for me!

Just wanted to mention the inconsistency I noticed between generating JSON Schema from a model and from a type adapter, but it's not something that affects me, just in case it points to something incorrect internally in some way.

dmontagu:
whoa we can have siblings keys for $ref? That probably means we can remove a lot of otherwise-error-prone code and also will probably simplify things for users in some cases where there may-or-may-not be an allOf

adriangb:
Agreed. I still think we should provide a better way of dealing with the need to generate multiple json schemas while resolving conflicts and such.

tiangolo:
I found the right/better link saying `$ref` can now have siblings: https://json-schema.org/draft/2019-09/release-notes.html#core-vocabulary

tiangolo:
Just FYI, I updated the tests in `main-pv2` in FastAPI with this branch, and now I made this draft PR that just installs this branch to confirm it makes tests pass: https://github.com/tiangolo/fastapi/pull/9777 🎉

adriangb:
@tiangolo could you clarify a bit what the use case is within FastAPI for this? I'm assuming it's to take a collection of types and generate JSON schemas for all of them (which will go into path operations, query parameters, etc. in the OpenAPI def) with a single `"$def"` section, handling name collisions and such. Is this right?

tiangolo:
@adriangb yep! That's right. In FastAPI, the first-class citizens are fields (in v1 there was a `ModelField` object) not the models. And this is indeed the pure types with their metadata, including lists, ints, and anything else that is not strictly a Pydantic model (and also including Pydantic models).

I collect all those, generate the JSON Schemas for them, and then put any relevant definitions in OpenAPI's components (e.g. for models), and the rest that don't produce a reusable definition but a pure JSON Schema I put them directly in the relevant parts of OpenAPI (e.g. a path operation parameter).

The specific thing I'm not being able to do is this, with FastAPI:

```Python
from fastapi import FastAPI, Query

app = FastAPI()


@app.get("/items/")
async def create_item(
    batch: str = Query(
        default="asdf",
        description="some query param",
        examples=["yes please", "oh, no"],
    ),
) -> str:
    return ""
```

The JSON Schema for that includes `batch`, includes the default value, but doesn't include the `description` and `examples`.


<details>
<summary>Here's a test for that code with the long OpenAPI</summary>

There's a comment inside the OpenAPI showing where the extra metadata should be:

```Python
def test_model_field():
    # insert_assert(app.openapi())
    assert app.openapi() == {
        "openapi": "3.1.0",
        "info": {"title": "FastAPI", "version": "0.1.0"},
        "paths": {
            "/items/": {
                "get": {
                    "summary": "Create Item",
                    "operationId": "create_item_items__get",
                    "parameters": [
                        {
                            "name": "batch",
                            "in": "query",
                            "required": False,
                            "schema": {
                                "type": "string",
                                "default": "asdf",
                                "title": "Batch",
                                # Here's the problem, this doesn't show up without this PR
                                "description": "some query param",
                                "examples": ["yes please", "oh, no"],
                                # End of the problem
                            },
                            "description": "some query param",
                        }
                    ],
                    "responses": {
                        "200": {
                            "description": "Successful Response",
                            "content": {
                                "application/json": {
                                    "schema": {
                                        "type": "string",
                                        "title": "Response Create Item Items  Get",
                                    }
                                }
                            },
                        },
                        "422": {
                            "description": "Validation Error",
                            "content": {
                                "application/json": {
                                    "schema": {
                                        "$ref": "#/components/schemas/HTTPValidationError"
                                    }
                                }
                            },
                        },
                    },
                }
            }
        },
        "components": {
            "schemas": {
                "HTTPValidationError": {
                    "properties": {
                        "detail": {
                            "items": {"$ref": "#/components/schemas/ValidationError"},
                            "type": "array",
                            "title": "Detail",
                        }
                    },
                    "type": "object",
                    "title": "HTTPValidationError",
                },
                "ValidationError": {
                    "properties": {
                        "loc": {
                            "items": {
                                "anyOf": [{"type": "string"}, {"type": "integer"}]
                            },
                            "type": "array",
                            "title": "Location",
                        },
                        "msg": {"type": "string", "title": "Message"},
                        "type": {"type": "string", "title": "Error Type"},
                    },
                    "type": "object",
                    "required": ["loc", "msg", "type"],
                    "title": "ValidationError",
                },
            }
        },
    }
```

</details>

adriangb:
Right but instead of you calling `GenerateSchema.generate_inner` and cobbling together a solution from disparate pieces of functionality in Pydantic I feel like it'd be better if Pydantic had first-class support for this, essentially I'd like something like `TypeAdapter.json_schemas` that actually works well, and in particular I'd like an easy hook to customize name clash resolution.

adriangb:
To be clear, I'm not against this change, I think making `Field` and `TypeAdapter` play well together makes a ton of sense.

tiangolo:
Agreed with both @adriangb!

---

More context on my side:

The main reason why I used `GenerateSchema.generate_inner` and not something else is because if I'm asking for a model schema, it will give me the `{"$ref": "blahblah/Foo"}` instead of rendering the model schema at the top level and including internal defs, which is what other methods did (I don't remember exactly which ones I tried).

This way I can re-use that ref schema in OpenAPI.

For FastAPI, the current (and I think sort of ideal? at least up to now) process would be:

* Collect all Pydantic *fields* (including Pydantic models, but also other things), each field is a combination of a type and a `FieldInfo` (e.g. `Annotated[str, Field(description="The name of the user")]`)
* Give a list of all that (some of them reference each other) to `GenerateSchema`
* Get a bunch of definitions that can be put in `$defs` (or in `components` in OpenAPI)

Up to this point, the process is just for `GenerateSchema` to know all the models, resolve name conflicts, and generate the definitions and usable refs.

* Later, give `GenerateSchema` (currently with `GenerateSchema.generate_inner`) one specific *field* (not model, a combination of type and `FieldInfo`)
* It gives back something I can inject in OpenAPI, so a `{"$ref": "blahblah/Foo"}` or a pure JSON Schema for things that don't live in the pre-created definitions.

Now, for this current process, the main thing I'm missing is being able to get all that, but making sure `FieldInfo` metadata attached to a type is preserved (what this PR does).

But if there's any (feasible) better way to do that process with different steps, that would also be very welcome, of course. 😅

adriangb:
> But if there's any (feasible) better way to do that process with different steps, that would also be very welcome, of course

That's what I think we should design :)

> each field is a combination of a type and a FieldInfo (e.g. Annotated[str, Field(description="The name of the user")])

I think this is an interesting bit. I get the feeling that `FieldInfo` is being abused to represent a different concept just because it can hold a type, a default value, and a name (the name matters, e.g. `authorization: FromHeader[str]` the `authorization` part matters. But maybe that's fine.

At the very least, I think this PR would be better if the function had the signature `generate_schemas(fields: list[FieldInfo], mode: Literal["input", "output"]) -> tuple[list[JsonSchemaValue], dict[str, JsonSchemaValue]]` where the returned tuple is (1:1 mapping to input fields, definitions).

And I don't think that should go in TypeAdapter, it seems like a better fit for a function in `json_schema.py`.

adriangb:
Looking through what we have, we already have a `GenerateJsonSchema.generate_definitions` that does pretty much exactly what you want Sebastián. It seems like we just need a version that accepts a `FieldInfo`. Given that you can always convert that to `Annotated[field_info.annotation, field_info]` (assuming we fix #6353) is there any reason that wouldn't work? I'm also not sure that using a type as a key (which is what `GenerateJsonSchema.generate_definitions` does) is always going to work, e.g. you could have two `List[int]` and I'd assume want to treat them separately? Not sure about that last bit.

tiangolo:
> I think this is an interesting bit. I get the feeling that FieldInfo is being abused to represent a different concept just because it can hold a type, a default value, and a name (the name matters, e.g. authorization: FromHeader[str] the authorization part matters. But maybe that's fine.

Just to clarify, in FastAPI the main class I move around is a custom `ModelField` that contains the `name` and the `FieldInfo` (as the name is not stored anywhere in v2, not even in `FieldInfo`, only in the parent model if there's one). 

Up to now, from what I've seen, the `FieldInfo` is the lowest common denominator that can store the information and can have information related to more than Pydantic models (it can hold information about lists, ints, etc). I think the lower layer would be pure Pydantic core schemas, but those don't contain the metadata (e.g. `description` and `examples`), that metadata is inside closures in callables in the Pydantic core `"metadata"` dict key.

> As it stands right now, I think fixing https://github.com/pydantic/pydantic/issues/6353 and related issues may be a better path forward so that at least TypeAdapter(Annotated[TheType, Field(...)]) works correctly.

Probably, it seems using `TypeAdapter` currently doesn't generate the extra metadata from `FieldInfo` for JSON Schema, if solving multiple `Field()` in annotated solves it, then that would certainly solve it.

> Looking through what we have, we already have a GenerateJsonSchema.generate_definitions that does pretty much exactly what you want Sebastián. It seems like we just need a version that accepts a FieldInfo. Given that you can always convert that to Annotated[field_info.annotation, field_info] (assuming we fix https://github.com/pydantic/pydantic/issues/6353) is there any reason that wouldn't work? I'm also not sure that using a type as a key (which is what GenerateJsonSchema.generate_definitions does) is always going to work, e.g. you could have two List[int] and I'd assume want to treat them separately? Not sure about that last bit.

Yep, `GenerateJsonSchema.generate_definitions` is exactly what I use right now. The only problem is that currently, it doesn't take types, it takes Pydantic core schemas, so `Annotated[field_info.annotation, field_info]` (which is what I was using up to this PR) still needed to go in a `TypeAdapter` to extract the Pydantic core schema from that.

I was doing:

```Python
TypeAdapter(Annotated[self.field_info.annotation, self.field_info]).core_schema
```

...but that didn't include the JSON Schema metadata (`description`, `examples`, etc).

I was collection all those, and then passing them to `GenerateJsonSchema.generate_definitions()`.

And then the next thing I need to do is "query" the `GenerateJsonSchema` for the JSON Schema for a specific type annotation (a specific `ModelField`).

If `GenerateJsonSchema.generate_definitions` could take `FieldInfo`s directly, that would definitely work. If it still took core schemas but `TypeAdapter` had a way to generate core schemas that end up including the metadata, that would work as well.

If there are two `List[int]` without more metadata I think it would be fine to use the same JSON Schema 🤔 ...it would probably be different if one is `Annotated[List[int], Query(description="the tokens")]` and the other is `Annotated[List[int], Query(description="the search terms")]`, but that's already different, I guess different objects with different id/hash, so that should be fine. I'm not sure if I'm missing something... but I think that's fine how it has worked up to now. The key thing now is being able to pass that, in some way, to `GenerateJsonSchema.generate_definitions()`, and then being able to query it again.

---

It also doesn't have to allow "querying" the `GenerateJsonSchema.generate_definitions()`, if the initial return had a way to map from input to output definitions, I could (probably?) handle the rest outside.

PR #6359, opened 2023-07-02
✏️ Invert order of `typing` <-> `typing_extensions`

Kludex:
The description of the error message is not accurate.

Selected Reviewer: @adriangb

### DIFF
--- a/docs/usage/errors.md
+++ b/docs/usage/errors.md
@@ -290,8 +290,7 @@ except PydanticUserError as exc_info:
 
 ## `TypedDict` version {#typed-dict-version}
 
-This error is raised when you use `typing_extensions.TypedDict`
-instead of `typing.TypedDict` on Python < 3.11.
+This error is raised when you use `typing.TypedDict` instead of `typing_extensions.TypedDict` on Python < 3.12.
 
 ## Model parent field overridden {#model-field-overridden}
 
--- a/docs/usage/errors.md
+++ b/docs/usage/errors.md
@@ -290,7 +290,7 @@ except PydanticUserError as exc_info:
 
 ## `TypedDict` version {#typed-dict-version}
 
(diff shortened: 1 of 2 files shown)


Kludex:
please review

docs/usage/errors.md
```diff
@@ -290,8 +290,7 @@ except PydanticUserError as exc_info:
 
 ## `TypedDict` version {#typed-dict-version}
 
-This error is raised when you use `typing_extensions.TypedDict`
-instead of `typing.TypedDict` on Python < 3.11.
+This error is raised when you use `typing.TypedDict` instead of `typing_extensions.TypedDict` on Python < 3.12.
```
hramezani, line 293: Shouldn't it be `Python < 3.11`?

There is a [note in docstring of the method](https://github.com/pydantic/pydantic/blob/35e9fd00cc436e4898dd41a065479618753eead6/pydantic/_internal/_generate_schema.py#L784-L785) that says  `we only support typing.TypedDict on Python >= 3.11`.

But in the [error message](https://github.com/pydantic/pydantic/blob/35e9fd00cc436e4898dd41a065479618753eead6/pydantic/_internal/_generate_schema.py#L797) of the same function, we have 3.12

I think we need to change the docstring or error message

Kludex: The docstring is wrong.

hramezani: Will you fix it in this commit?

Kludex: Yes

Kludex: Ok, it looks like it's supposed to be < 3.11 instead of < 3.12.

Kludex:
please review

docs/usage/errors.md
adriangb: It needs to be 3.12 for TypedDic

adriangb:
Superseded by #6370

PR #6382, opened 2023-07-03
🐛 Don't call `repr()` on qualname retrieval

### LINKED ISSUE #6368: Super class variables no longer available to subclasses
### Initial Checks

### Description

Hi, I have an issue with accessing variables in a superclass from a base class. I added an example code in the section below.

Whenever I run the code, I get the following error below. I clipped most of the error message to reduce verbosity.
```
...
  File "random.py", line 22, in __repr__
    return f"{__class__.__name__}(number={self.number})"
  File "random.py", line 12, in number
    return self._number
AttributeError: 'DerivedClass' object has no attribute '_number'
```
Was this error intended to happen?

### Example Code

```Python

class BaseClass:

    @validate_call
    def __init__(self, number: int) -> None:
        self._number = number

    @property
    def number(self) -> int:
        return self._number


class DerivedClass(BaseClass):

    @validate_call
    def __init__(self, number: int) -> None:
        super().__init__(number)

    def __repr__(self) -> str:
        return f"{__class__.__name__}(number={self.number})"
    
d = DerivedClass(50)
print(d)
```

```Text
pydantic version: 2.0
                 install path: /Users/troiw/opt/miniconda3/envs/torch_mdn/lib/python3.8/site-packages/pydantic
               python version: 3.8.16 (default, Jun 12 2023, 12:55:15)  [Clang 14.0.6 ]
```

Selected Assignee: @Kludex

Kludex:
- Closes #6368

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -83,7 +83,7 @@ def get_type_ref(type_: type[Any], args_override: tuple[type[Any], ...] | None =
     if isinstance(origin, TypeAliasType):
         type_ref = f'{module_name}.{origin.__name__}'
     else:
-        qualname = getattr(origin, '__qualname__', f'<No __qualname__: {origin}>')
+        qualname = getattr(origin, '__qualname__', '<No __qualname__>')
         type_ref = f'{module_name}.{qualname}:{id(origin)}'
 
     arg_refs: list[str] = []
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -83,7 +83,10 @@ def get_type_ref(type_: type[Any], args_override: tuple[type[Any], ...] | None =
     if isinstance(origin, TypeAliasType):
         type_ref = f'{module_name}.{origin.__name__}'
     else:
-        qualname = getattr(origin, '__qualname__', '<No __qualname__>')
+        try:
+            qualname = getattr(origin, '__qualname__', f'<No __qualname__: {origin}>')
+        except Exception:
+            qualname = getattr(origin, '__qualname__', '<No __qualname__>')
         type_ref = f'{module_name}.{qualname}:{id(origin)}'
 
     arg_refs: list[str] = []
--- a/tests/test_validate_call.py
+++ b/tests/test_validate_call.py
@@ -637,3 +637,15 @@ def test_model_as_arg() -> None:
 
     res = f1({'x': '1'}, {'y': '2'})  # type: ignore
     assert res == ({'x': 1}, Model2(y=2))
+
+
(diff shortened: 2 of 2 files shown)


pydantic/_internal/_core_utils.py
```diff
@@ -83,7 +83,7 @@ def get_type_ref(type_: type[Any], args_override: tuple[type[Any], ...] | None =
     if isinstance(origin, TypeAliasType):
         type_ref = f'{module_name}.{origin.__name__}'
     else:
-        qualname = getattr(origin, '__qualname__', f'<No __qualname__: {origin}>')
+        qualname = getattr(origin, '__qualname__', '<No __qualname__>')
```
Kludex, line 86: We can have a fallback to `__str__` if wanted... 🤷‍♂️

Kludex, line 86: We can have a fallback to `__str__` if wanted... 🤷‍♂️

Kludex: The thing is that `repr()` is always called, even if `origin.__qualname__` exists... 🤔

Anyway... This PR solves the issue.

Kludex:
please review

dmontagu:
While this may solve the issue, I'd like to better understand why it solves it. Something seems very weird about the reported error.. I'm looking into this more deeply now.

dmontagu:
Okay I see what's happening — it's using `__get__` to bind the `__init__` method, which is the only method where this issue happens, because it's the only instance method that can be called prior to the instance being initialized (duh..).

There are a lot of places where the get_type_ref happens where I still feel it would be useful to have the repr of the object (e.g., with a functools.partial), and it seems unfortunate to throw that all out so that it can't error for a `validated_call` on an `__init__`. Would you object to trying to get the repr and handling _any_ exception by replacing it with what has happened here? I'll add a suggestion on the file to make it more clear what I mean.

pydantic/_internal/_core_utils.py
dmontagu, line 86:
```suggestion
        try:
            qualname = getattr(origin, '__qualname__', f'<No __qualname__: {origin}>')
        except Exception:
            qualname = getattr(origin, '__qualname__', f'<No __qualname__>')
```

Kludex: Yeah, this is what I mentioned on https://github.com/pydantic/pydantic/pull/6382#discussion_r1250725725

Kludex: (I've edited your msg because it had a f-string, fyi)

Kludex:
please review

dmontagu (approved):

PR #6398, opened 2023-07-03
Allow combining `Field` annotations by merging them

### LINKED ISSUE #6353: Using multiple `Field`s in a single `Annotated` is broken
For BaseModel metadata gets dropped. For dataclasses we prohibit it, but I suspect we don't do a good job of detecting it in all cases.

```python
from typing import Annotated

class Model(BaseModel):
    foo: Annotated[int, Field(description='xyz'), Field(title='abc')]


print(Model.model_json_schema())  # title gets dropped
```

Similarly with TypeAdapter:

```python
from typing import Annotated

from pydantic import Field, TypeAdapter


# all metadata ignored
print(
    TypeAdapter(
        Annotated[int, Field(description='xyz'), Field(title='abc')]
    ).json_schema()
)
```

(I think this last case is being fixed in https://github.com/pydantic/pydantic/pull/6336)

```python
from typing import Annotated

from pydantic import TypeAdapter, Field, dataclasses


@dataclasses.dataclass
class Model:
    foo: Annotated[int, Field(description='xyz'), Field(title='abc')]


print(TypeAdapter(Model).json_schema())  # TypeError: Field may not be used twice on the same field
```

I think we should figure out if we can merge `Fields` w/ some sort of semantics for overriding and support that everywhere.

Selected Assignee: @lig

adriangb:
Fixes #6353, closes #6336

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -677,8 +677,6 @@ class GenerateSchema:
         # the default validator needs to go outside of any other validators
         # so that it is the topmost validator for the field validator
         # which uses it to check if the field has a default value or not
-        if field_info.default is ...:
-            print(1)
         if not field_info.is_required():
             schema = wrap_default(field_info, schema)
 
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -677,6 +677,8 @@ class GenerateSchema:
         # the default validator needs to go outside of any other validators
         # so that it is the topmost validator for the field validator
         # which uses it to check if the field has a default value or not
+        if field_info.default is ...:
+            print(1)
         if not field_info.is_required():
             schema = wrap_default(field_info, schema)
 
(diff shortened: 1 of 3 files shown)


adriangb:
@tiangolo  I think this solves the problem you raised in https://github.com/pydantic/pydantic/pull/6336#issuecomment-1616930587

pydantic/_internal/_generate_schema.py
```diff
@@ -1228,6 +1230,11 @@ def _apply_annotations(  # noqa: C901
         # expand annotations before we start processing them so that `__prepare_pydantic_annotations` can consume
         # individual items from GroupedMetadata
         annotations = list(_known_annotated_metadata.expand_grouped_metadata(annotations))
+        non_field_infos, field_infos = [a for a in annotations if not isinstance(a, FieldInfo)], [
+            a for a in annotations if isinstance(a, FieldInfo)
+        ]
+        if field_infos:
+            annotations = [*non_field_infos, FieldInfo.merge_field_infos(*field_infos)]
```
adriangb, line 1235: This re-ordering does have some possible consequences. In particular, each `FieldInfo` currently hooks into `__get_pydantic_core_schema__` on its own so the order matters. But this re-ordering says "even if everything else runs in the order it is put into Annotated, we do special handling for Field() so that it runs at the end".

adriangb:
please review

tiangolo:
Yesss! This solves it for me, and simplifies the code on FastAPI as I can now just use the `TypeAdapter.core_schema` directly. 🎉 

If this goes in (and is released), I'll be able to make the final FastAPI beta, and then the final release in one or two days. 🚀

hramezani (approved):

PR #6400, opened 2023-07-03
Add note about changes to regex engine used

### LINKED ISSUE #6374: Missing breaking change note for look-around regexes/patterns
### Initial Checks

### Description

Look-around regexes/patterns (e.g. `^(?!_)(?!.*__)[a-z_]{1,64}(?<!_)$`) have been supported in `v1`, but fail with the following error in `v2`:

```
pydantic_core._pydantic_core.SchemaError: Error building "model" validator:
  SchemaError: Error building "model-fields" validator:
  SchemaError: Field "value":
  SchemaError: Error building "dict" validator:
  SchemaError: Error building "str" validator:
  SchemaError: regex parse error:
    ^(?!_)(?!.*__)[a-z_]{1,64}(?<!_)$
     ^^^
error: look-around, including look-ahead and look-behind, is not supported
```

I didn't see this noted as a breaking change (or I overlooked it!).

Is this something that's planned to be supported again?

(pydantic `v2` is looking very cool so far! 🎉)

### Example Code

```Python
import pydantic

    x: pydantic.constr(pattern=r"^(?!_)(?!.*__)[a-z_]{1,64}(?<!_)$")
```

```Text
             pydantic version: 2.0
                 install path: /Users/empicano/Library/Caches/pypoetry/virtualenvs/server-W6QUvoSE-py3.11/lib/python3.11/site-packages/pydantic
               python version: 3.11.0 (main, Dec 15 2022, 21:05:29) [Clang 14.0.0 (clang-1400.0.29.102)]
                     platform: macOS-12.6-x86_64-i386-64bit
```

Selected Assignee: @adriangb

adriangb:
Fixes #6374

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -506,6 +506,15 @@ except ValidationError as e:
     """
 ```
 
+#### Patterns / regex on strings
+
+Pydantic V1 used Python's regex library. Pydantic V2 uses the Rust [regex crate].
+This crate is not just a "Rust version of regular expressions", it's a completely different approach to regular expressions.
+In particular, it promises linear time searching of strings in exchange for dropping a couple of features (namely look arounds and backreferences).
+We believe this is a tradeoff worth making, in particular because Pydantic is used to validate untrusted input where ensuring things don't accidentally run in exponential time depending on the untrusted input is important.
+
+[regex crate]: https://github.com/rust-lang/regex
+
 ### Introduction of `TypeAdapter`
 
 Pydantic V1 had weak support for validating or serializing non-`BaseModel` types.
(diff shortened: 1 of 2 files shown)


adriangb:
please review

lig (approved):
👍🏻

docs/migration.md
```diff
@@ -506,6 +506,15 @@ except ValidationError as e:
     """
 ```
 
+#### Patterns / regex on strings
+
+Pydantic V1 used Python's regex library. Pydantic V2 uses the Rust [regex crate].
```
Kludex, line 511: Hmm... Should we add between parenthesis what "crate" means?

Kludex (approved):

docs/migration.md
adriangb: I don't think it matters, if someone cares they can click the link.

adriangb:
cc @empicano

empicano:
Very neat! Thanks for the quick response and the possible workaround 👍

adriangb:
Note for future: it’d be interesting to have an API to let users override the default behavior eg if you want to interpret Field(pattern=…) to use Python’s regex module by default.

PR #6402, opened 2023-07-03
Rework how JSON schema refs get built/"simplified"

### LINKED ISSUE #6369: Nested models created with `create_model` with same name results in key error
### Initial Checks

### Description

When using the same model name in `create_model` as an inner model (dynamic or static) this results in a key error when creating the JSON schema:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "<REDACTED>/app/__main__.py", line 83, in <module>
    print(model_b.model_json_schema())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<REDACTED>/.venv/lib/python3.11/site-packages/pydantic/main.py", line 372, in model_json_schema
    return model_json_schema(
           ^^^^^^^^^^^^^^^^^^
  File "<REDACTED>/.venv/lib/python3.11/site-packages/pydantic/json_schema.py", line 1972, in model_json_schema
    return schema_generator_instance.generate(cls.__pydantic_core_schema__, mode=mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<REDACTED>/.venv/lib/python3.11/site-packages/pydantic/json_schema.py", line 320, in generate
    json_schema = self.resolve_collisions(json_schema)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<REDACTED>/.venv/lib/python3.11/site-packages/pydantic/json_schema.py", line 1684, in resolve_collisions
    json_schema = self.change_defs_ref(defs_ref, choice, json_schema)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<REDACTED>/.venv/lib/python3.11/site-packages/pydantic/json_schema.py", line 1712, in change_defs_ref
    self.definitions[new] = self.definitions.pop(old)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'aInput'
```

When using the same model next to each other this will work:
```python
from pydantic import create_model, BaseModel

class A(BaseModel):
    inner: str

model_a1 = create_model(
    "A",
    inner_a1=(str, ...),
)
model_a2 = create_model(
    "A",
    inner_a2=(str, ...),
)

model_c = create_model(
    "B",
    outer_a1=(model_a1, ...),
    outer_a2=(model_a2, ...),
)

print(model_c.model_json_schema()) # shows two distinct schemas
```

### Example Code

```Python

model_a = create_model(
    "a",
    inner=(str, ...),
)
model_b = create_model(
    "a",
    outer=(model_a, ...),
)

print(model_b.model_json_schema()) # Key Error
```

```Text
pydantic version: 2.0
                 install path: <REDACTED>
               python version: 3.11.4 (main, Jun  7 2023, 00:34:59) [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.4.1-arm64-arm-64bit
```

dmontagu:
This closes #6369, and also addresses various other issues:

* Fixes a bug where names weren't being simplified fully in the presence of nested models (i.e., you'd get separate Input and Output schemas for an outer model even if the inner model had the same Input and Output schema, now you don't)
* Fixes a bug where for certain $ref collisions, one model might be remapped and the other not (it was intended that both get renamed to include more detail in the $ref)
* Makes the names of generated JSON schemas stable across runs even when multiple models have the exact same qualname. Previously we used `id(model)` to ensure the names were unique, but this would result in different schemas on every run. In this PR I have fixed this so that when there are qualname conflicts, we add `__1` to the first encounter with that qualname, `__2` to the second, etc., instead of using the `id`.

----

However, this PR currently involves some changes to the "ostensibly-public" APIs of `GenerateJsonSchema`.

 @samuelcolvin: @adriangb and I discussed and I think we should make some of the "implementation detail" methods from `GenerateJsonSchema` private (which I at least partially do in this PR..), and publish something describing in more detail what we intend to treat as public and what we don't from the semver perspective.

In particular, after discovering the issues with the current approach from investigating #6369, I really think we should change the implementation a bit, which is done in the PR I'm about to open, but technically it involves replacing some public methods of `GenerateJsonSchema`. I think given how advanced this use case is we should probably make this change now.

Happy to discuss more before merging anything, and happy to put in more effort to rework this to not have any theoretically-breaking changes, but I think there's an argument to be made that we never intended to make these methods public and that we are now renaming them to make them private.

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -5,6 +5,8 @@ import inspect
 import math
 import re
 import warnings
+from collections import defaultdict
+from copy import deepcopy
 from dataclasses import is_dataclass
 from enum import Enum
 from typing import (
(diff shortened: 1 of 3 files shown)


pydantic/json_schema.py
```diff
@@ -120,6 +122,91 @@ class PydanticJsonSchemaWarning(UserWarning):
...
+            remapping = _DefinitionsRemapping(defs_remapping, json_remapping)
+            new_definitions_schema = remapping.remap_json_schema({'$defs': copied_definitions})
+            if definitions_schema == new_definitions_schema:
+                # We've reached the fixed point
+                return remapping
+            definitions_schema = new_definitions_schema
+
+        warnings.warn(
+            'Failed to simplify the JSON schema definitions, this may be a bug in Pydantic', PydanticJsonSchemaWarning
+        )
```
adriangb, line 177: I feel like we should just error here. I'd rather have someone report the error so we can fix it with a reproducible example than just warn.

dmontagu: Converted to an error

adriangb (approved):

Issue #5163, opened 2023-03-09
Improve performance of union validation by inferring a discriminator


dmontagu:
In cases where a field is comprised of a union of types that each have a literal field that is incompatible with the other types in the union, it may be (significantly) more performant to use the TaggedUnionSchema validation logic to perform validation, as it will ensure that the item is validated against the correct schema within the union choices immediately, and also ensures that no time is wasted validating against other schemas when checking the discriminator is enough to know validation is going to fail.

Today, this can be achieved by explicitly assigning/annotating the attribute with `Field(discriminator='kind')` (or similar). However, it may be straightforward to automatically infer a discriminator for a union type, reducing the need for explicit description by users, and improving performance even for users who are unaware of this functionality.

(We could go even farther than looking for Literal attributes, and produce more involved discriminators that would look for specific attributes only defined on one of the cases, but I think this may quickly get into the realm of unnecessary implementation complexity or producing validation error messages that are more-confusing/difficult-to-debug.)

Creating this issue for tracking purposes. @adriangb @samuelcolvin please comment if there's anything you wanted to add/emphasize/alter relative to what I've said above.

dmontagu:
@samuelcolvin I believe we looked into what typescript does at PyCon, and when multiple interfaces are present in a union with a shared literal attribute, it does seem to infer that a discriminator should be used.

I think it would be straightforward to handle many common cases automatically (namely, where every member of the union has a field of literal type with non-overlapping values). While one could argue that _in theory_ this might result in some confusing error messages in the case where you meant to pass data for one case but used the literal for another, I think _in practice_ this pretty universally improves the understandability of the error messages involved.

If there's consensus that inferring a discriminator in such cases is a good thing, I would be happy to attempt to implement this (I think it should be a fairly small and self-contained modification to `pydantic._internal._generate_schema.GenerateSchema._union_schema`).

derbasti381:
Hello David,
I'm not sure if this is the right place for my report, but hope you can have a look at it.

I adapted my code regarding Unions from v1 validators to v2 discriminated Unions, which worked well in v2.0b3. With v2.0 I get `TypeError: `discriminator` can only be used with `Union` type with more than one variant`. This was temporarily also thrown in v2.0b3, but after I removed all v1 validators and added the Literals to my `type` attributes, everything worked well.

This is my Dict with Annotated objects as values which all have `type: Literal['nameXY']` attributes.
```
objects: Dict[
    str,
    Annotated[
[...]

PR #6405, opened 2023-07-03
Allow single-item discriminated unions

dmontagu:
I don't have proof yet as I can't reproduce the issue, but I suspect this may resolve the issue brought up in https://github.com/pydantic/pydantic/issues/5163#issuecomment-1619203179.

Even if not, I think this is an improvement worth making, as I don't see any reason we should error if you produce a single-item discriminated union. In particular, it may be desirable to do this as a way to immediately fail validation if the tag is wrong rather than producing all the possible associated errors.

### DIFF
--- a/pydantic/_internal/_discriminated_union.py
+++ b/pydantic/_internal/_discriminated_union.py
@@ -181,10 +181,11 @@ class _ApplyInferredDiscriminator:
             return definitions_wrapper
 
         if schema['type'] != 'union':
-            raise TypeError('`discriminator` can only be used with `Union` type with more than one variant')
-
-        if len(schema['choices']) < 2:
-            raise TypeError('`discriminator` can only be used with `Union` type with more than one variant')
+            # If the schema is not a union, it probably means it just had a single member and
+            # was flattened by pydantic_core.
+            # However, it still may make sense to apply the discriminator to this schema,
+            # as a way to get discriminated-union-style error messages, so we allow this here.
+            schema = core_schema.union_schema([schema])
 
         # Reverse the choices list before extending the stack so that they get handled in the order they occur
         self._choices_to_handle.extend(schema['choices'][::-1])
--- a/tests/test_discriminated_union.py
+++ b/tests/test_discriminated_union.py
@@ -13,22 +13,48 @@ from pydantic._internal._discriminated_union import apply_discriminator
 from pydantic.errors import PydanticUserError
 
 
-def test_discriminated_union_only_union():
+def test_discriminated_union_type():
     with pytest.raises(
-        TypeError, match='`discriminator` can only be used with `Union` type with more than one variant'
+        TypeError, match="'str' is not a valid discriminated union variant; should be a `BaseModel` or `dataclass`"
     ):
 
         class Model(BaseModel):
             x: str = Field(..., discriminator='qwe')
 
 
-def test_discriminated_union_single_variant():
-    with pytest.raises(
-        TypeError, match='`discriminator` can only be used with `Union` type with more than one variant'
-    ):
+@pytest.mark.parametrize('union', [True, False])
+def test_discriminated_single_variant(union):
+    class InnerModel(BaseModel):
+        qwe: Literal['qwe']
 
-        class Model(BaseModel):
-            x: Union[str] = Field(..., discriminator='qwe')
+    class Model(BaseModel):
+        if union:
+            x: Union[InnerModel] = Field(..., discriminator='qwe')
+        else:
+            x: InnerModel = Field(..., discriminator='qwe')
+
+    assert Model(x={'qwe': 'qwe'}).x.qwe == 'qwe'
+    with pytest.raises(ValidationError) as exc_info:
+        Model(x={'qwe': 'asd'})
+    assert exc_info.value.errors(include_url=False) == [
+        {
+            'ctx': {'discriminator': "'qwe'", 'expected_tags': "'qwe'", 'tag': 'asd'},
+            'input': {'qwe': 'asd'},
+            'loc': ('x',),
+            'msg': "Input tag 'asd' found using 'qwe' does not match any of the expected " "tags: 'qwe'",
+            'type': 'union_tag_invalid',
+        }
+    ]
+
+
+def test_discriminated_union_single_variant():
+    class InnerModel(BaseModel):
+        qwe: Literal['qwe']
+
+    class Model(BaseModel):
+        x: Union[InnerModel] = Field(..., discriminator='qwe')
+
+    assert Model(x={'qwe': 'qwe'}).x.qwe == 'qwe'
 
 
 def test_discriminated_union_invalid_type():
@@ -812,15 +838,6 @@ def test_invalid_discriminated_union_type() -> None:
 
         class Model(BaseModel):
             pet: Union[Cat, Dog, str] = Field(discriminator='pet_type')
-
-
-def test_single_item_union_error() -> None:
-    fields = {'kind': core_schema.typed_dict_field(core_schema.literal_schema(['only_choice']))}
-    schema = core_schema.union_schema([core_schema.typed_dict_schema(fields=fields)])
-    with pytest.raises(
-        TypeError, match='`discriminator` can only be used with `Union` type with more than one variant'
-    ):
-        apply_discriminator(schema, 'kind')
 
 
 def test_invalid_alias() -> None:


dmontagu:
Update: the issue described in the comment linked above is resolved on the current `main` branch even without this PR.

However, I would still be in favor of merging this for the reasons described above. Obviously it's less urgent though given it's not necessary to resolve a reported issue.

dmontagu:
please review

adriangb:
wouldn't the right thing to do be to just return the inner schema? why have a union at all if there's just 1 thing?

dmontagu:
I can imagine scenarios where it's useful to get the discriminated union behavior. In particular, rather than reporting all the possible errors if the discriminator is missing, it just reports one.

More concretely, here is an example:
```python

from pydantic import BaseModel, TypeAdapter, Field, ValidationError

class Model1(BaseModel):
    kind: Literal['one']
    x: str
    y: int
    z: float
    w: bool


class Model2(BaseModel):
    kind: Literal['two']
    x: int
    y: str
    z: bool
    w: float


data = {'kind': 'two', 'x': 1, 'y': 'a', 'z': False, 'w': 1.5}

adapter = TypeAdapter(Model1)
try:
    adapter.validate_python(data)
except ValidationError as exc:
    print(exc)
    """
    4 validation errors for Model1
    kind
      Input should be 'one' [type=literal_error, input_value='two', input_type=str]
        For further information visit https://errors.pydantic.dev/2.0.2/v/literal_error
    x
      Input should be a valid string [type=string_type, input_value=1, input_type=int]
        For further information visit https://errors.pydantic.dev/2.0.2/v/string_type
    y
      Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='a', input_type=str]
        For further information visit https://errors.pydantic.dev/2.0.2/v/int_parsing
    w
      Input should be a valid boolean [type=bool_type, input_value=1.5, input_type=float]
        For further information visit https://errors.pydantic.dev/2.0.2/v/bool_type
    """

adapter_with_discriminator = TypeAdapter(Annotated[Model1, Field(..., discriminator='kind')])
try:
    adapter_with_discriminator.validate_python(data)
except ValidationError as exc:
    print(exc)
"""
1 validation error for tagged-union[Model1]
  Input tag 'two' found using 'kind' does not match any of the expected tags: 'one' [type=union_tag_invalid, input_value={'kind': 'two', 'x': 1, '...', 'z': False, 'w': 1.5}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.0.2/v/union_tag_invalid
"""
```

In a world where you know you'll get one of two inputs, and you can distinguish by the kind field, but you only want one of the two to pass validation, I can see the error from `adapter_with_discriminator.validate_python(data)` being much preferable.

-------

Another case that seems like it could be useful would be if you have logic for determining what items should be in a tagged union where having a single item is still sensible:
```python
from typing import Literal, Annotated, Union

from pydantic import BaseModel, TypeAdapter, Field, ValidationError


class Cat(BaseModel):
    kind: Literal['cat'] = 'cat'


class Dog(BaseModel):
    kind: Literal['dog'] = 'dog'


class Fish(BaseModel):
    kind: Literal['fish'] = 'fish'


def get_pet_adapter(cats_allowed: bool, dogs_allowed: bool, fish_allowed: bool) -> TypeAdapter:
    allowed_kinds = []
    if cats_allowed:
        allowed_kinds.append(Cat)
    if dogs_allowed:
        allowed_kinds.append(Dog)
    if fish_allowed:
        allowed_kinds.append(Fish)
    return TypeAdapter(Annotated[Union[tuple(allowed_kinds)], Field(discriminator='kind')])


adapter = get_pet_adapter(cats_allowed=False, dogs_allowed=True, fish_allowed=False)
adapter.validate_python({'kind': 'dog'})
```

While this could potentially be worked around by simplifying the union, I still feel there is utility in being able to say "if the discriminator field fails validation, don't bother validating anything else".

The main reason why I think we should do this is that, as far as I can tell:
* it adds no complexity to the implementation (arguably simplifies it)
* I don't see any reason why it would be actively undesirable to be able to do tagged-union-style validation for a single case when a user went out of their way to specify a discriminator
* I _can_ imagine scenarios where it would be significantly beneficial

tests/test_discriminated_union.py
```diff
@@ -13,22 +13,49 @@
...
+def test_discriminated_single_variant(union):
+    class InnerModel(BaseModel):
+        qwe: Literal['qwe']
+        y: int
+
+    class Model(BaseModel):
+        if union:
+            x: Union[InnerModel] = Field(..., discriminator='qwe')
+        else:
+            x: InnerModel = Field(..., discriminator='qwe')
```
davidhewitt, line 35: `Union[InnerModel]` just evaluates to `InnerModel` and I think this is unlikely to change, so there may be no need to test the two branches?

ref https://docs.python.org/3/library/typing.html#typing.Union

pydantic/_internal/_discriminated_union.py
```diff
@@ -181,10 +181,11 @@ def _apply_to_root(self, schema: core_schema.CoreSchema) -> core_schema.CoreSche
             return definitions_wrapper
 
         if schema['type'] != 'union':
-            raise TypeError('`discriminator` can only be used with `Union` type with more than one variant')
-
-        if len(schema['choices']) < 2:
-            raise TypeError('`discriminator` can only be used with `Union` type with more than one variant')
+            # If the schema is not a union, it probably means it just had a single member and
+            # was flattened by pydantic_core.
```
davidhewitt, line 185: As per other comment I think it's the `Union` implementation, not `pydantic_core`, responsible for the flattening.

davidhewitt (commented):
Arguably this removes the distinction between treatment of a single type and a 1-element union when using a discriminator.

Presumably this would (very slightly) negatively impact performance for users who use this, because `pydantic_core` would first validate the tag and then prodeed to validate the whole type (which would include the tag again) - i.e. it's not an optimization in the one-member case, unlike for multiple members.

An argument to not do this would be that by forbidding it we don't give users the opportunity to deoptimize themselves. Users can already handle the case of single members differently in the examples @dmontagu illustrated above.

Overall this seems like a tradeoff of usability versus performance. I think because the error (of trying to use `discriminant` for a non-union type, i.e. single-element union) fails at model build time, I'm currently falling just on the side of not allowing this.

We could benchmark the performance impact; maybe the overhead added is irrelevant, in which case I'd be👍

pydantic/_internal/_discriminated_union.py
ehiggs-deliverect: When I get here in the debugger, I actually see my type is `tagged-union`

dmontagu: @davidhewitt there is also a thing in pydantic-core that flattens unions. I think in principle you could get a `core_schema.union_schema([<one_choice>])`, that this would handle. I agree it's not likely to be hit currently.

dmontagu:
> Overall this seems like a tradeoff of usability versus performance. I think because the error (of trying to use discriminant for a non-union type, i.e. single-element union) fails at model build time, I'm currently falling just on the side of not allowing this.

@davidhewitt note that this is opt-in — you still have to say you want to use a discriminator to get this behavior. Given that, I feel like it's not really a tradeoff in performance, because you can just choose not to include the discriminator field in the `Field` call and you will get the current behavior.

Am I misunderstanding something, or do you think even with that it should still be an error?

dmontagu:
@davidhewitt also to be clear:

> I think because the error (of trying to use discriminant for a non-union type, i.e. single-element union) fails at model build time

This PR would change that:
```python
from typing import Literal

class Model1(BaseModel):
    kind: Literal['one']
    x: str
    y: int
    z: float
    w: bool


class M(BaseModel):
    x: Model1 = Field(discriminator='kind')


data_one = {'kind': 'one', 'x': 'a', 'y': 1, 'z': 1.5, 'w': True}
data_two = {'kind': 'two', 'x': 1, 'y': 'a', 'z': False, 'w': 1.5}

print(M(x=data_one))
#> x=Model1(kind='one', x='a', y=1, z=1.5, w=True)

M(x=data_two)
"""
pydantic_core._pydantic_core.ValidationError: 1 validation error for M
x
  Input tag 'two' found using 'kind' does not match any of the expected tags: 'one' [type=union_tag_invalid, input_value={'kind': 'two', 'x': 1, '...', 'z': False, 'w': 1.5}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.0.2/v/union_tag_invalid
"""
```

davidhewitt:
> Am I misunderstanding something, or do you think even with that it should still be an error?

My perception was that the error at build time is good because it stops users adding a performance overhead.

But on reflection I retract my viewpoint; I'm in favour of the usability improvement proposed here. If it's a performance hit, that's something we can solve in the implementation.

davidhewitt (approved):

Issue #6390, opened 2023-07-03
validate_call makes bound methods no longer equal


jaraco:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

When wrapping a method in `validate_call` it alters the assumption that a bound method is equal to itself.

Running the example produces the following error on Pydantic 2 but not Pydantic 1:

```
 draft @ pip-run 'pydantic<2@git+https://github.com/pydantic/pydantic' -- pyd2.py
 draft @ pip-run 'pydantic>=2@git+https://github.com/pydantic/pydantic' -- -c 'import pydantic.version; print(pydantic.version.version_info())'
             pydantic version: 2.0
        pydantic-core version: 2.0.1 release build profile
                 install path: /private/var/folders/sx/n5gkrgfx6zd91ymxr2sr9wvw00n8zm/T/pip-run-d58gw1wg/pydantic
               python version: 3.11.4 (main, Jun 15 2023, 07:55:38) [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.4.1-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
 draft @ pip-run 'pydantic>=2@git+https://github.com/pydantic/pydantic' -- pyd2.py
/Users/jaraco/draft/pyd2.py:14: PydanticDeprecatedSince20: The `validate_arguments` method is deprecated; use `validate_call` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0/migration/
  b = validate_arguments(a)
Traceback (most recent call last):
  File "/Users/jaraco/draft/pyd2.py", line 21, in <module>
    assert thing.c == thing.c, f'{thing.c} != {thing.c}'
           ^^^^^^^^^^^^^^^^^^
AssertionError: ValidateCallWrapper(<bound method Thing.a of <__main__.Thing object at 0x1057b3590>>) != ValidateCallWrapper(<bound method Thing.a of <__main__.Thing object at 0x1057b3590>>)
```

### Example Code

```Python
from pydantic import validate_arguments

try:
    from pydantic import validate_call
except ImportError:
    # Pydantic 1
    validate_call = validate_arguments


class Thing:
    def a(self):
        pass

    b = validate_arguments(a)
    c = validate_call(a)


thing = Thing()
assert thing.a == thing.a
[...]

PR #6406, opened 2023-07-03
Bind `validate_call` to instance when decorating methods

dmontagu:
Resolves https://github.com/pydantic/pydantic/issues/6390

I'll also note that this results in a **_100x_** speedup of the following (admittedly trivial) benchmark:
```python
import time

class Thing:
    def a(self):
        pass

    c = validate_call(a)


thing = Thing()

t0 = time.time()
for _ in range(100_000):
    thing.c()
t1 = time.time()
print(f'{t1 - t0:.3f}')
#> without this change: ~6.0s
#> with this change: ~0.060s
```

### DIFF
--- a/pydantic/_internal/_validate_call.py
+++ b/pydantic/_internal/_validate_call.py
@@ -85,6 +85,8 @@ class ValidateCallWrapper:
             self.__return_pydantic_core_schema__ = None
             self.__return_pydantic_validator__ = None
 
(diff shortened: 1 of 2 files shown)


adriangb (approved):

jaraco:
This change fixes the reported issue, but introduces a new one on class attribute access. I'll open a report.

PR #6408, opened 2023-07-04
Fix list indentation

ofek:
Screenshot 2023-07-03 230701

Selected Reviewer: @lig

### DIFF
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -73,10 +73,10 @@ A few things to note on validators:
   of `UserModel`.
 * the second argument is always the field value to validate; it can be named as you please
 * you can also add any subset of the following arguments to the signature (the names **must** match):
-  * `values`: a dict containing the name-to-value mapping of any previously-validated fields
-  * `config`: the model config
-  * `field`: the field being validated. Type of object is `pydantic.fields.ModelField`.
-  * `**kwargs`: if provided, this will include the arguments above not explicitly listed in the signature
+   * `values`: a dict containing the name-to-value mapping of any previously-validated fields
+   * `config`: the model config
+   * `field`: the field being validated. Type of object is `pydantic.fields.ModelField`.
+   * `**kwargs`: if provided, this will include the arguments above not explicitly listed in the signature
 * validators should either return the parsed value or raise a `ValueError` or `AssertionError`
   (``assert`` statements may be used).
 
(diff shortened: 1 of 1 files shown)


lig:
@tpdorsey This is interesting.

- The preview on Github for `main` is ok https://github.com/pydantic/pydantic/blob/main/docs/usage/validators.md
- Same is the preview for this PR https://github.com/ofek/pydantic/blob/patch-1/docs/usage/validators.md

- Docs on the website aren't good https://docs.pydantic.dev/dev-v2/usage/validators/#field-validators

tpdorsey:
The approach is correct, but it should be 4 spaces per indent, including an additional indent on wrapping lines.

```
* validators are "class methods", so the first argument value they receive is the `UserModel` class, not an instance
    of `UserModel`.
* the second argument is always the field value to validate; it can be named as you please
* you can also add any subset of the following arguments to the signature (the names **must** match):
    * `values`: a dict containing the name-to-value mapping of any previously-validated fields
    * `config`: the model config
    * `field`: the field being validated. Type of object is `pydantic.fields.ModelField`.
    * `**kwargs`: if provided, this will include the arguments above not explicitly listed in the signature
* validators should either return the parsed value or raise a `ValueError` or `AssertionError`
    (``assert`` statements may be used).

!!! warning
    If you make use of `assert` statements, keep in mind that running
    Python with the [`-O` optimization flag](https://docs.python.org/3/using/cmdline.html#cmdoption-o)
    disables `assert` statements, and **validators will stop working**.

* where validators rely on other values, you should be aware that:
    * Validation is done in the order fields are defined.
        E.g. in the example above, `password2` has access to `password1` (and `name`),
        but `password1` does not have access to `password2`. See [Field Ordering](models.md#field-ordering)
        for more information on how fields are ordered

    * If validation fails on another field (or that field is missing) it will not be included in `values`, hence
        `if 'password1' in values and ...` in this example.
```

The difference in rendering is down to different markdown parsing engines. The mkdocs parser is more strict about indents.

ofek:
Where do you see those rules?

tpdorsey:
@ofek https://www.markdownguide.org/basic-syntax#adding-elements-in-lists

ofek:
I increased the indentation and also updated the text to reflect the actual state of v2

tpdorsey:
Thank you for the contribution!

PR #6410, opened 2023-07-04
💚 Use "result" instead of "status" on GitHub Action conditional

Kludex:
Let's pretend this never happened.

Ref.: https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context


Kludex:
please review

hramezani (approved):

PR #6436, opened 2023-07-04
Fix `GenerateJsonSchema.generate_definitions` signature

dmontagu:
This massively improves the `GenerateJsonSchema.generate_definitions` API — discovered the need for this while working with @tiangolo through FastAPI test failures with pydantic v2.

In particular, prior to this change, the first item returned by the `GenerateJsonSchema.generate_definitions` would _only_ include a mapping to schemas that were "pure" definition schemas. Now, it will include every set of inputs as a mapped output.

In particular, we were able to make this usable from FastAPI for the purpose of exposing the `mode='serialization'` capability, fixing some bugs with the FastAPI generation, and significantly reduce the overhead of generating JSON schema — it previously needed to be computed twice because the map returned by the function didn't have all (or even most) of the data necessary; now it does. (This is used on the `main-pv2-serialization-schemas`.)

If there are any objections to making this change to the `generate_definitions` method in pydantic, I would suggest we add a new method with a new name that does exactly this. However, considering how much more useful this method is and the fact that we currently view `GenerateJsonSchema` as "unstable", I would be in favor of just making this change to that method's signature. Open to disagreement on this.

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -309,7 +309,7 @@ class GenerateJsonSchema:
 
     def generate_definitions(
         self, inputs: Sequence[tuple[JsonSchemaKeyT, JsonSchemaMode, core_schema.CoreSchema]]
-    ) -> tuple[dict[tuple[JsonSchemaKeyT, JsonSchemaMode], DefsRef], dict[DefsRef, JsonSchemaValue]]:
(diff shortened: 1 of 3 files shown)


adriangb (approved):

tiangolo (approved):
Yessss! Thanks 🚀

adriangb (approved):

adriangb (approved):

samuelcolvin:
amazing 🚀 .

Issue #6438, opened 2023-07-04
'use_enum_values' not working in union types


schwingkopf:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

I observed that the `use_enum_values=True` setting does not correctly work for fields that use a union of `Enum` and another type.
Minimal reproducible example:
```
from enum import Enum
from pydantic import BaseModel, ConfigDict
from pydantic_settings import BaseSettings, SettingsConfigDict

class MyEnum(Enum):
    val = "val"

class Model1(BaseModel):
    model_config = ConfigDict(use_enum_values=True)
    x: MyEnum

class Model2(BaseModel):
    model_config = ConfigDict(use_enum_values=True)
    x: MyEnum | int

m = Model1(x=MyEnum.val)
print(m.x)

m = Model2(x=MyEnum.val)
print(m.x)
```
Output is:
```
val
MyEnum.val
```

Expected output would be:
```
val
val
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0
        pydantic-core version: 2.0.1 release build profile
                 install path: C:\Users\<...>\AppData\Local\pypoetry\Cache\virtualenvs\<...>-nZd32k8O-py3.10\Lib\site-packages\pydantic  
               python version: 3.10.9 (tags/v3.10.9:1dd9be6, Dec  6 2022, 20:01:21) [MSC v.1934 64 bit (AMD64)]
                     platform: Windows-10-10.0.19045-SP0
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @davidhewitt

PR #6440, opened 2023-07-04
Fix issue with union parsing of enums

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6438

### DIFF
--- a/pydantic/_internal/_std_types_schema.py
+++ b/pydantic/_internal/_std_types_schema.py
@@ -95,6 +95,12 @@ def get_enum_core_schema(enum_type: type[Enum], config: ConfigDict) -> CoreSchem
         update_json_schema(original_schema, updates)
         return json_schema
 
+    strict_python_schema = core_schema.is_instance_schema(enum_type)
+    if use_enum_values:
+        strict_python_schema = core_schema.chain_schema(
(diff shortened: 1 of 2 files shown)


pydantic/_internal/_std_types_schema.py
```diff
@@ -95,6 +95,12 @@ def get_json_schema(_, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
         update_json_schema(original_schema, updates)
         return json_schema
 
+    strict_python_schema = core_schema.is_instance_schema(enum_type)
+    if use_enum_values:
+        strict_python_schema = core_schema.chain_schema(
+            [strict_python_schema, core_schema.no_info_plain_validator_function(lambda x: x.value)]
```
dmontagu, line 101: I think what happens is that, when parsing unions, it tries to parse in strict mode first so that it gets the right union case, rather than the first one, when possible. Because the strict mode python schema wasn't applying the `use_enum_values` thing before, it didn't affect the parsing of unions.

I guess it was also a bug that if you were to validate an enum in strict mode with `use_enum_values=True` it wouldn't actually use the enum values either.

davidhewitt (approved):
LGTM 👍

Issue #6455, opened 2023-07-05
Ability to Pickle RootModels


Fokko:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

PyIceberg is used in some systems where Pydantic objects are being pickled. We don't always control the serialization framework, therefore we need to support this.

### Example Code

```Python
Python 3.9.6 (default, May  7 2023, 23:32:44) 
[Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pydantic import RootModel, ConfigDict
>>> from typing import Generic, TypeVar
>>> 
>>> T = TypeVar('T')
>>> 
>>> class FrozenRootModel(RootModel[T], Generic[T]):
...     model_config = ConfigDict(frozen=True)
...     
... 
>>> FrozenRootModel('yes') == FrozenRootModel('yes')
True
>>> import pickle
>>> FrozenRootModel('yes') == pickle.loads(pickle.dumps(FrozenRootModel('yes')))
False
```
```


### Python, Pydantic & OS Version

```Text
>>> import pydantic.version; print(pydantic.version.version_info())
             pydantic version: 2.0.1
        pydantic-core version: 2.0.2 release build profile
                 install path: /Users/fokkodriesprong/Library/Python/3.9/lib/python/site-packages/pydantic
               python version: 3.9.6 (default, May  7 2023, 23:32:44)  [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.4.1-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```
```


Selected Assignee: @Kludex

**Comment by davidhewitt** [comment]
It seems to me that the `False` arises because after the pickle round-trip, the `__dict__` picks up some extra members. (Which leads to `False` being returned from `BaseModel.__eq__`.)

```python
>>> a = FrozenRootModel('yes')
>>> b = pickle.loads(pickle.dumps(a))
>>> a.__dict__
{'root': 'yes'}
>>> b.__dict__
{'root': 'yes', '__pydantic_extra__': None, '__pydantic_private__': None}
```

PR #6457, opened 2023-07-05
Fix pickling of RootModel

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6455

### DIFF
--- a/tests/test_construction.py
+++ b/tests/test_construction.py
@@ -353,6 +353,18 @@ def test_pickle_undefined(create_module):
     assert not hasattr(m3, '_foo_')
 
 
+def test_pickle_root_model(create_module):
+    @create_module
+    def module():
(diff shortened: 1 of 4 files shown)


tests/test_construction.py
```diff
@@ -353,6 +353,18 @@ class PickleModelTwo(BaseModel):
     assert not hasattr(m3, '_foo_')
 
 
+def test_pickle_root_model(create_module):
```
lig, line 356: Maybe it would be better to keep consolidating `RootModel` tests in `tests/test_root_model.py`?

lig (approved):
I was under impression that I've tested that :(

Overall LGTM 👍🏻 

A module for the test is up to you.

Fokko (approved):
LGTM, thanks for the quick fix! 🚀

Issue #6463, opened 2023-07-05
Optional fields raise validation errors: Field required [type=missing...]


ybressler:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

Optional fields are being ignored when I build a model. This is true for fields annotated with `Field()` function, and those that are not.

These errors are being raised when instantiating a model as a class `Model(...)` as well as  parsing from a dict: `Model.model_validate(...)`

**A workaround:** Set the default value of the field to be `None`. This is a breaking change of behavior from V1 and doesn't seem correct. (Unless it is, then it's not covered in the documentation.) 

I've provided a reproducible example below.

### Example Code

```Python
# main.py
from typing import Optional
from pydantic import BaseModel, Field


class MyModel(BaseModel):
    a: Optional[int]
    b: Optional[int] = Field(description="Optional int annotated with a field")
    c: Optional[str]
    d: Optional[str] = Field(description="Optional str annotated with a field")


# --------------------------

# test_main.py

from typing import Optional
import pytest

from main import MyModel


@pytest.mark.parametrize(
    ('a', 'b', 'c', 'd'),
    [
        (1, 2, 'one', 'two'),
        (1, None, 'one', None),
        (None, 2, None, 'two'),
        (None, None, None, None)
    ]
)
def test_model_with_values(
        a: Optional[int],
        b: Optional[int],
        c: Optional[str],
        d: Optional[str]
):
    """Model succeeds with explicit None for optional fields"""
    my_model = MyModel(a=a, b=b, c=c, d=d)


@pytest.mark.xfail
def test_model_with_some_values():
    """This raises an error, and it shouldn't"""
    my_model = MyModel(a=1)



@pytest.mark.parametrize(
    "data",
    [
        # These work if I explicitly set the values to None
        ({"a": 1, "b": 2, "c": "one", "d": "three"}),
        ({"a": 1, "b": None, "c": "one", "d": None}),
        ({"a": 1, "b": 2, "c": "one", "d": "three"}),
        ({"a": None, "b": 2, "c": None, "d": "three"}),
        ({"a": None, "b": None, "c": None, "d": None}),
        # But they fail if I am missing values - no default is set
        pytest.param({"a": 1, "c": "one"}, marks=pytest.mark.xfail),
        pytest.param({"b": 2, "d": "two"}, marks=pytest.mark.xfail),
        pytest.param({}, marks=pytest.mark.xfail)
    ]
[...]

PR #6468, opened 2023-07-05
Revise the section on required / optional / nullable fields.

ybressler:
## Change Summary
Small enhancements to migration documentation – explicitly call out the breaking change of optional fields. (Related to https://github.com/pydantic/pydantic/issues/6463)

**Note:**
I created 2 alternative tables for demonstrating how fields are defined, cause I'm not sure which one we should go with. (Or, if we should keep it as a bullet pointed list. Or, if we should scrap it and just keep the code example below it.)

## Checklist

* [ ] ~Unit tests for the changes exist~
* [ ] ~Tests pass on CI and coverage remains at 100%~

Selected Reviewer: @lig

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -479,26 +479,18 @@ In Pydantic V2, we recognize that the value is an instance of one of the cases a
 
 Pydantic V2 changes some of the logic for specifying whether a field annotated as `Optional` is required
 (i.e., has no default value) or not (i.e., has a default value of `None`), and now more closely matches the
-behavior of `dataclasses`.
+behavior of `dataclasses`. Similarly, fields annotated as `Any` no longer have a default value of `None`.
 
 The following table describes the behavior of field annotations in V2:
 
-| State                        | Field Definition           |
-|------------------------------|----------------------------|
-| Required, cannot be None     | `f1: str`              |
-| Required, can be None        | `f2: Optional[str]`        |
-| Not required, can be None    | `f3: Optional[str] = None` |
-| Not required, cannot be None | `f4: str = 'Foobar'`       |
-
-
-
-| Required | Can be `None` | Field Definition |
-|----------|---------------|------------------|
-| ✔        | X             | `f1: str`        |
-| ✔         | ✔             | `f2: Optional[str]`        |
-| X        | ✔             | `f3: Optional[str] = None` |
-| X        | X             | `f4: str = 'Foobar'`       |
-
+| State                                            | Field Definition           |
+|--------------------------------------------------|----------------------------|
+| Required, cannot be `None`                       | `f1: str`                  |
+| Required, can be `None`                          | `f2: Optional[str]`        |
+| Not required, can be `None`                      | `f3: Optional[str] = None` |
+| Not required, cannot be `None`                   | `f4: str = 'Foobar'`       |
+| Required, can be any type (including `None`)     | `f5: Any`                  |
+| Not required, can be any type (including `None`) | `f6: Any = None`           |
 
 
 !!! note
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -483,22 +483,24 @@ behavior of `dataclasses`. Similarly, fields annotated as `Any` no longer have a
 
 The following table describes the behavior of field annotations in V2:
 
-| State                                            | Field Definition           |
-|--------------------------------------------------|----------------------------|
-| Required, cannot be `None`                       | `f1: str`                  |
+| State                                                 | Field Definition            |
+|-------------------------------------------------------|-----------------------------|
+| Required, cannot be `None`                            | `f1: str`                   |
 | Not required, cannot be `None`, is `'abc'` by default | `f3: str = 'abc'`           |
 | Required, can be `None`                               | `f2: Optional[str]`         |
 | Not required, can be `None`, is `None` by default     | `f3: Optional[str] = None`  |
 | Not required, can be `None`, is `'abc'` by default    | `f3: Optional[str] = 'abc'` |
-| Not required, cannot be `None`                   | `f4: str = 'Foobar'`       |
-| Required, can be any type (including `None`)     | `f5: Any`                  |
-| Not required, can be any type (including `None`) | `f6: Any = None`           |
+| Not required, cannot be `None`                        | `f4: str = 'Foobar'`        |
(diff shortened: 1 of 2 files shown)


ybressler:
please review

docs/migration.md
```diff
@@ -477,11 +477,35 @@ In Pydantic V2, we recognize that the value is an instance of one of the cases a
...
+| Not required, cannot be None | `f4: str = 'Foobar'`       |
+
+
+
+| Required | Can be `None` | Field Definition |
+|----------|---------------|------------------|
+| ✔        | X             | `f1: str`        |
+| ✔         | ✔             | `f2: Optional[str]`        |
+| X        | ✔             | `f3: Optional[str] = None` |
+| X        | X             | `f4: str = 'Foobar'`       |
```
ybressler, line 499: Here are 2 formats for expressing field definitions. An additional open would be to show as a bullet point list. (I'd rather have as text in addition to the code example below – since these are breaking changes.)

If we wanted to go overboard, we could show how this is different from V1: 
```
| State                        | V1 Field Definition           | V2 Field Definition           |
|------------------------------|----------------------------|---------------------|
| Required, cannot be None     | `f1: str`              | `f1: str`              |
| Required, can be None        | `f2: Optional[str] = Field(...)`        | `f2: Optional[str]`             |
| Not required, can be None    | `f3: Optional[str]` | `f3: Optional[str]` |
| Not required, cannot be None | `f4: str = 'Foobar'`       | `f4: str = 'Foobar'`       |

```

docs/migration.md
```diff
@@ -477,11 +477,35 @@ In Pydantic V2, we recognize that the value is an instance of one of the cases a
...
+
+
+
+!!! note
+    A field annotated as `typing.Optional[T]` will be required, and will allow for a value of `None`.
+    It does not mean that the field has a default value of `None`. _(This is a breaking change from V1.)_
+
+
+Code example demonstrating the above:
```
dmontagu, line 508:
```suggestion
Here is a code example demonstrating the above:
```

docs/migration.md
```diff
@@ -477,11 +477,35 @@ In Pydantic V2, we recognize that the value is an instance of one of the cases a
 
 #### Required, optional, and nullable fields
 
-Pydantic V1 had a somewhat loose idea about "required" versus "nullable" fields. In Pydantic V2, these concepts are
-more clearly defined.
+Pydantic V2 introduces stricter definitions for specifying if a field is `required` or `optional` – more closely
+matching the behavior `dataclasses`.
```
dmontagu, line 481:
```suggestion
Pydantic V2 changes some of the logic for specifying whether a field annotated as `Optional` is required
(i.e., has no default value) or not (i.e., has a default value of `None`), and now more closely matches the
behavior of `dataclasses`.
```

docs/migration.md
```diff
@@ -477,11 +477,35 @@ In Pydantic V2, we recognize that the value is an instance of one of the cases a
...
 
-Pydantic V2 will move to match `dataclasses`, thus you may explicitly specify a field as `required` or `optional` and whether the field accepts `None` or not.
+In V2, fields are specified as follows:
```
dmontagu, line 483:
```suggestion
The following table describes the behavior of field annotations in V2:
```

docs/migration.md
dmontagu: I think the table you have is good as is, don't think we need to expand it to have the "V1 Field Definition"

dmontagu (commented):
One other thing that I think would be good — in V1, all of this also applied for fields annotated as `Any`. It might be worth adding a note that this has changed as well — fields annotated as `Any` no longer have a default of `None` either.

docs/migration.md
```diff
@@ -477,11 +477,28 @@ In Pydantic V2, we recognize that the value is an instance of one of the cases a
...
+behavior of `dataclasses`. Similarly, fields annotated as `Any` no longer have a default value of `None`.
 
-Pydantic V2 will move to match `dataclasses`, thus you may explicitly specify a field as `required` or `optional` and whether the field accepts `None` or not.
+The following table describes the behavior of field annotations in V2:
 
+| State                                            | Field Definition           |
+|--------------------------------------------------|----------------------------|
+| Required, cannot be `None`                       | `f1: str`                  |
+| Required, can be `None`                          | `f2: Optional[str]`        |
+| Not required, can be `None`                      | `f3: Optional[str] = None` |
```
lig, line 490: I feel like this could give a slightly incorrect impression that exactly `= None` is needed to transform a required field into a not required one.

I think that it is worth mentioning here that any default value if provided makes a field not required.

Something like this maybe
```suggestion
| Not required, cannot be `None`, is `'abc'` by default | `f3: str = 'abc'`      |
| Required, can be `None`                               | `f2: Optional[str]`      |
| Not required, can be `None`, is `None` by default     | `f3: Optional[str] = None`  |
| Not required, can be `None`, is `'abc'` by default    | `f3: Optional[str] = 'abc'` |
```

docs/migration.md
```diff
@@ -477,11 +477,28 @@ In Pydantic V2, we recognize that the value is an instance of one of the cases a
 
 #### Required, optional, and nullable fields
 
-Pydantic V1 had a somewhat loose idea about "required" versus "nullable" fields. In Pydantic V2, these concepts are
-more clearly defined.
+Pydantic V2 changes some of the logic for specifying whether a field annotated as `Optional` is required
+(i.e., has no default value) or not (i.e., has a default value of `None`), and now more closely matches the
```
lig, line 481:
```suggestion
(i.e., has no default value) or not (i.e., has a default value of `None` or any other value of the corresponding type), and now more closely matches the
```

lig (changes requested):

lig (approved):
👍🏻 looks top-notch

dimbleby:
the second of these entries seems just to be a less-fully-explained copy of the second?

```markdown
| Not required, cannot be `None`, is `'abc'` by default | `f3: str = 'abc'`           |
| Not required, cannot be `None`                        | `f4: str = 'Foobar'`        |
```

lig:
@dimbleby Nice catch. Thanks! Would you like making a PR?;)

dimbleby:
negative, sorry!

tpdorsey:
I will follow up on this. Thanks for pointing it out @dimbleby Cheers

PR #6480, opened 2023-07-06
Fix the un-hashability of various annotation types, for use in caching generic containers

dmontagu:
skip change file check

Selected Reviewer: @hramezani

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -5074,9 +5074,14 @@ def test_handle_3rd_party_custom_type_reusing_known_metadata() -> None:
     ]
 
 
-def test_skip_validation():
+@pytest.mark.parametrize('optional', [True, False])
+def test_skip_validation(optional):
+    type_hint = SkipValidation[int]
+    if optional:
+        type_hint = Optional[type_hint]
+
     @validate_call
-    def my_function(y: Annotated[int, SkipValidation]):
+    def my_function(y: type_hint):
         return repr(y)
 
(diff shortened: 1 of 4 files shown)


tests/test_types.py
```diff
@@ -5107,6 +5112,16 @@ class A(BaseModel):
...
 
 
+def test_transform_schema():
+    ValidateStrAsInt = Annotated[str, TransformSchema(lambda _s: core_schema.int_schema())]
+
+    class Model(BaseModel):
+        x: Optional[ValidateStrAsInt]
+
+    assert Model(x=None).x is None
+    assert Model(x='1').x == 1
```
ybressler, line 5122: Transformation from a string repr of an int/float/decimal is no longer automatically executed in V2, right? I'm assuming you are overriding this with `TransformSchema(lambda _s: core_schema.int_schema())`?

Also, this test could be rewritten as parameterized if you wanted to go overboard. Perhaps some other edge cases like negative numbers, or floats? (Though would a float get transformed to an int?)

ybressler (commented):
> _**Note:** I'm trying to get in the habit of reviewing more code here. Apologies if my comments are n00b._

Code looks straightforward overall. I appreciate the addition of tests. Your PR description mentions inclusion of methods in caching, though I don't see any tests which relate to the caching of these methods. (Unless they are implicit?)

dmontagu:
please review

hramezani (approved):

Issue #6477, opened 2023-07-06
npt.NDArray with `arbitrary_types_allowed=True` raises `pydantic.errors.PydanticSchemaGenerationError`


pvardanis:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

I'm having the following `BaseModel` that was previously working with `v1`:

```python
CustomNDArrayType: TypeAlias = Union[npt.NDArray, List[npt.NDArray]]

class SomeModel(BaseModel):
    data: Optional[CustomNDArrayType]

    model_config = ConfigDict(
        arbitrary_types_allowed=True,
        extra="forbid",
        validate_assignment=True,
    )
```

but in `v2` raises the following:

```
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for numpy.ndarray[typing.Any, numpy.dtype[+ScalarType]]. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
For further information visit https://errors.pydantic.dev/2.0.2/u/schema-for-unknown-type
```

Strange thing is, that setting `arbitrary_types_allowed=True` doesn't ignore the error at all. Any ideas how to deal with this?

### Example Code

```Python
CustomNDArrayType: TypeAlias = Union[npt.NDArray, List[npt.NDArray]]

class SomeModel(BaseModel):
    data: Optional[CustomNDArrayType]

    model_config = ConfigDict(
        arbitrary_types_allowed=True,
        extra="forbid",
        validate_assignment=True,
    )
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0.2
        pydantic-core version: 2.1.2 release build profile
                 install path: /Users/panagiotisvardanis/miniforge3/envs/mac-tensorflow/lib/python3.8/site-packages/pydantic
               python version: 3.8.16 | packaged by conda-forge | (default, Feb  1 2023, 16:01:13)  [Clang 14.0.6 ]
                     platform: macOS-13.4.1-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @adriangb

[...]

PR #6484, opened 2023-07-06
Replace TransformSchema with GetPydanticSchema

dmontagu:
`GetPydanticSchema` is meant to be a convenience type that eliminates the need to make "marker classes" when doing simple things related to custom types.

I specifically worked on this to provide a good way to handle https://github.com/pydantic/pydantic/issues/6477 with versions of python/numpy where `numpy.typing.NDArray` is not actually a type, as I couldn't do this easily with our existing annotations. (I'll include a comment below demonstrating that this works for this, so I can more easily link to it in that issue.)

While I understand there may be some hesitancy to introduce another annotation type like this, I think this implementation will "stand the test of time" because it directly matches the custom hooks interface — this type will reduce boilerplate and not require breaking changes for as long as we don't introduce breaking changes into the custom hooks types.

---

Also, considering `TransformSchema` is not in `pydantic.types.__all__`, is completely undocumented, and migrating usage of it to `GetPydanticSchema` is essentially trivial, I think it's okay to remove it, which I am currently doing in this PR.

That said, we could definitely keep it and just add `GetPydanticSchema` as a new type, I just think it has been mostly rendered redundant.

skip change file check

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -1337,41 +1337,43 @@ __getattr__ = getattr_migration(__name__)
 class GetPydanticSchema:
     """A convenience class for creating an annotation that provides pydantic custom type hooks.
 
-    This class lets you replace the use of a custom type defining one or more of `__get_pydantic_core_schema__`,
-    `__get_pydantic_json_schema__`, and `__prepare_pydantic_annotations__` with a single annotation.
+    This class is intended to eliminate the need to create a custom "marker" which defines the
+     `__get_pydantic_core_schema__` and `__get_pydantic_json_schema__` custom hook methods.
+
+    For example, to have a field treated by type checkers as `int`, but by pydantic as `Any`, you can do:
+    ```python
+    from typing import Any
+
+    from typing_extensions import Annotated
+
+    from pydantic import BaseModel, GetPydanticSchema
+
+    HandleAsAny = GetPydanticSchema(lambda _s, h: h(Any))
+
+    class Model(BaseModel):
+        x: Annotated[int, HandleAsAny]  # pydantic sees `x: Any`
+
+    print(repr(Model(x='abc').x))
+    #> 'abc'
+    ```
     """
 
     get_pydantic_core_schema: Callable[[Any, _annotated_handlers.GetCoreSchemaHandler], CoreSchema] | None = None
     get_pydantic_json_schema: Callable[[Any, _annotated_handlers.GetJsonSchemaHandler], JsonSchemaValue] | None = None
-    prepare_pydantic_annotations: Callable[[Any, tuple[Any, ...], ConfigDict], tuple[Any, Iterable[Any]]] | None = None
+    # Note: if we find a use, we could uncomment the following as a way to specify `__prepare_pydantic_annotations__`:
+    # prepare_pydantic_annotations: Callable[
+    #   [Any, tuple[Any, ...], ConfigDict], tuple[Any, Iterable[Any]]
+    # ] | None = None
 
-    @staticmethod
-    def for_type(type_: Any) -> GetPydanticSchema:
-        """Returns an annotation that makes pydantic treat the annotated field as the provided type.
-
-        For example:
-        ```python
-        from pydantic import BaseModel, GetPydanticSchema
-
-        class Model(BaseModel):
-            x: Annotated[int, GetPydanticSchema.for_type(Any)]  # pydantic sees `x: Any`
-
-        print(repr(Model(x='abc').x))
-        #> 'abc'
-        ```
-        """
-        return GetPydanticSchema(
-            get_pydantic_core_schema=lambda _s, h: h(type_), get_pydantic_json_schema=lambda _s, h: h(type_)
-        )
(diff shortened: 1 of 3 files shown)


dmontagu:
Here is a code snippet showing how the type introduced by this PR can be used to address https://github.com/pydantic/pydantic/issues/6477:


```python
from typing import Union, List, Optional, Any

import numpy as np
import numpy.typing as npt
from typing_extensions import TypeAlias, Annotated

from pydantic import BaseModel, GetPydanticSchema, InstanceOf

VALIDATE_NDARRAY_INSTANCES = False
if VALIDATE_NDARRAY_INSTANCES:
    PydanticNDArray: TypeAlias = Annotated[npt.NDArray, GetPydanticSchema.for_type(InstanceOf[np.ndarray])]
else:
    PydanticNDArray: TypeAlias = Annotated[npt.NDArray, GetPydanticSchema.for_type(Any)]

CustomNDArrayType: TypeAlias = Union[PydanticNDArray, List[PydanticNDArray]]


class SomeModel(BaseModel):
    data: Optional[CustomNDArrayType]

print(SomeModel(data=None))
# data=None

print(SomeModel(data=np.array([1, 2, 3])))
# data=array([1, 2, 3])

print(SomeModel(data=[np.array([1, 2, 3])]))
# data=[array([1, 2, 3])]

print(SomeModel(data=[1]))
# If VALIDATE_NDARRAY_INSTANCES is False:
#> data=[1]  

# If VALIDATE_NDARRAY_INSTANCES is True:
"""
pydantic_core._pydantic_core.ValidationError: 2 validation errors for SomeModel
data.is-instance[ndarray]
  Input should be an instance of ndarray [type=is_instance_of, input_value=[1], input_type=list]
    For further information visit https://errors.pydantic.dev/2.1.2/v/is_instance_of
data.list[is-instance[ndarray]].0
  Input should be an instance of ndarray [type=is_instance_of, input_value=1, input_type=int]
    For further information visit https://errors.pydantic.dev/2.1.2/v/is_instance_of
"""
```

dmontagu:
please review (cc @adriangb regardless of who hooky picks)

pydantic/types.py
```diff
@@ -1333,12 +1334,45 @@ def encode_str(self, value: str) -> str:
...
+    `__get_pydantic_json_schema__`, and `__prepare_pydantic_annotations__` with a single annotation.
+    """
 
-    def __get_pydantic_core_schema__(
-        self, source_type: type[Any], handler: _annotated_handlers.GetCoreSchemaHandler
-    ) -> CoreSchema:
-        return self.transform(handler(source_type))
+    get_pydantic_core_schema: Callable[[Any, _annotated_handlers.GetCoreSchemaHandler], CoreSchema] | None = None
+    get_pydantic_json_schema: Callable[[Any, _annotated_handlers.GetJsonSchemaHandler], JsonSchemaValue] | None = None
+    prepare_pydantic_annotations: Callable[[Any, tuple[Any, ...], ConfigDict], tuple[Any, Iterable[Any]]] | None = None
```
adriangb, line 1346: Can we leave this one off for now?

pydantic/types.py
```diff
@@ -1333,12 +1334,45 @@ def encode_str(self, value: str) -> str:
...
+
+    @staticmethod
+    def for_type(type_: Any) -> GetPydanticSchema:
```
adriangb, line 1349: Can we leave this off for now? I'd like it to be requested/ used in more places before we introduce an abstraction.

adriangb (approved):
A couple suggestions but generally looks good!

pydantic/types.py
dmontagu: I'm okay with that, but I would suggest we leave this line there with a comment explaining that if we find that it would be useful, it could be uncommented.

adriangb: Sure fine by me

pydantic/types.py
dmontagu: If we don't include it, I would at least consider my proposed solution above to the numpy thing a request.

In particular, I think `GetPydanticSchema.for_type(InstanceOf[np.ndarray])` is a lot easier to understand than
`GetPydanticSchema(lambda _s, h: h(InstanceOf[np.ndarray]), lambda _s, h: h(InstanceOf[np.ndarray]))`.

I'm okay waiting until I have another example or two where this utility would be useful, but I think it's advanced enough that I'll be surprised if people request it even if they would find it useful; I sort of think we'd need documentation of it before people would realize it was an option. Basically, if we internally want it a few more times I think it would be worth including.

The best argument against including it now, in my opinion, is if we want/need to change the way it interacts with `__prepare_pydantic_annotations__`, and just suggesting users use the thing I said was "hard to understand" above.

dmontagu: hmm actually I think I made a mistake in the `for_type` method anyway, and you don't need to specify the `get_pydantic_json_schema` if you specify the core one. So maybe that's another good reason not to include it 😅. I'll remove it for now.

dmontagu:
please update (for me)

dmontagu:
please review

adriangb (approved):

PR #6487, opened 2023-07-06
Ensure docs tests always run in the right folder

dmontagu:
I frequently find myself annoyed because, if I use PyCharm to run the `test_docs.py` tests, it defaults to running in a different folder than it uses when run via `make test`, and it causes some tests to fail.

This PR fixes that and ensures no matter what the cwd is when pytest is running, it gets set properly and the tests pass.


dmontagu:
please review

dmontagu:
please review

lig (approved):
👍🏻

coditamar:
@CodiumAI-Agent please review

PR #6493, opened 2023-07-06
Add documentation of cyclic reference handling

dmontagu:
Took a stab at documenting this.

Happy to drop the examples of handling the recursion errors if they are overkill.

skip change file check

### DIFF
--- a/docs/usage/postponed_annotations.md
+++ b/docs/usage/postponed_annotations.md
@@ -91,7 +91,7 @@ When working with self-referencing recursive models, it is possible that you mig
 in validation inputs. For example, this can happen when validating ORM instances with back-references from
 attributes.
 
-Rather than raising a python `RecursionError` while attempting to validate data with cyclic references, Pydantic is able
+Rather than raising a Python `RecursionError` while attempting to validate data with cyclic references, Pydantic is able
 to detect the cyclic reference and raise an appropriate `ValidationError`:
 
 ```py
@@ -124,7 +124,7 @@ except ValidationError as exc:
     """
 ```
 
-Because this error is raised without actually exceeding the maximum recursion depth, it means that you can catch and
+Because this error is raised without actually exceeding the maximum recursion depth, you can catch and
 handle the raised `ValidationError` without needing to worry about the limited remaining recursion depth:
 
 ```python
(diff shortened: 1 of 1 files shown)


docs/usage/postponed_annotations.md
```diff
@@ -84,3 +84,186 @@ print(Foo())
...
 #> a=123 sibling=Foo(a=321, sibling=None)
 ```
+
+### Cyclic references
+
+When working with self-referencing recursive models, it is possible that you might encounter cyclic references
+in validation inputs. For example, this can happen when validating ORM instances with back-references from
+attributes.
+
+Rather than raising a python `RecursionError` while attempting to validate data with cyclic references, Pydantic is able
```
tpdorsey, line 94:
```suggestion
Rather than raising a Python `RecursionError` while attempting to validate data with cyclic references, Pydantic is able
```

docs/usage/postponed_annotations.md
```diff
@@ -84,3 +84,186 @@ print(Foo())
...
+except ValidationError as exc:
+    print(exc)
+    """
+    1 validation error for ModelB
+    a.b
+      Recursion error - cyclic reference detected [type=recursion_loop, input_value={'a': {'b': {...}}}, input_type=dict]
+    """
+```
+
+Because this error is raised without actually exceeding the maximum recursion depth, it means that you can catch and
```
tpdorsey, line 127:
```suggestion
Because this error is raised without actually exceeding the maximum recursion depth, you can catch and
```

docs/usage/postponed_annotations.md
```diff
@@ -84,3 +84,186 @@ print(Foo())
...
+# Create data with cyclic references representing the graph 1 -> 2 -> 3 -> 1
+node_data = {'id': 1, 'children': [{'id': 2, 'children': [{'id': 3}]}]}
+node_data['children'][0]['children'][0]['children'] = [node_data]
+
+print(Node.model_validate(node_data))
+#> id=1 children=[Node(id=2, children=[Node(id=3, children=[])])]
+```
+
+Similarly, if Pydantic encounters a recursive reference during _serialization_, rather than waiting for the maximum
+recursion depth to be exceeded a `ValueError` is raised immediately:
```
tpdorsey, line 184:
```suggestion
recursion depth to be exceeded, a `ValueError` is raised immediately:
```

tpdorsey (approved):

PR #6507, opened 2023-07-07
Fix ci

adriangb:
skip change file check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,7 +253,7 @@ jobs:
         python-version: ['3.8', '3.9', '3.10', '3.11']
         include:
           - typing-extensions-version: '4.6.1'
-            python-version: '3.8'
+            python-version: '3.7'
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -271,9 +271,9 @@ jobs:
     - name: install typing-extensions
       run: |
         if [ ${{ matrix.typing-extensions-version }} == 'main' ]; then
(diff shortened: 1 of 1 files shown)


adriangb:
I wasn't able to use `pdm add` because of https://github.com/pdm-project/pdm/issues/2093. I'd prefer to use `pdm add` because pip won't fail (or fail as nicely) if there are incompatibilities with other installed deps.

adriangb:
please review

.github/workflows/ci.yml
```diff
@@ -243,14 +243,17 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
         # test the oldest supported version and main
         typing-extensions-version: ['4.6.1', 'main']
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
+        include:
+          - typing-extensions-version: '4.6.1'
+            python-version: '3.7'
```
lig, line 256: Why do we even need 4.6.1 here? As far as I can see, we just do simple `make test`.

I think it's enough to run against `main` here. Thus this could be just a list of python versions and no `else` branch in the `install typing-extensions` step.

lig (changes requested):

.github/workflows/ci.yml
adriangb: The idea is to test the oldest version we support and the latest version on every python version we support

adriangb: `make test` could very well fail if .e.g we import a symbol that doesn't exist

PR #6514, opened 2023-07-07
Fix typing of model_validator

### LINKED ISSUE #6277: [PYD-136] ModelAfterValidator causing warnings from Pylance and mypy

adriangb:
Fixes #6277. Closes https://github.com/python/mypy/issues/15620

skip change file check

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/functional_validators.py
+++ b/pydantic/functional_validators.py
@@ -419,7 +419,7 @@ ModelAfterValidator = Callable[[_ModelType, _core_schema.ValidationInfo], _Model
 
 _AnyModelWrapValidator = Union[ModelWrapValidator, ModelWrapValidatorWithoutInfo]
 _AnyModeBeforeValidator = Union[ModelBeforeValidator, ModelBeforeValidatorWithoutInfo]
-_AnyModeAfterValidator = Union[ModelAfterValidator[_ModelType], ModelAfterValidatorWithoutInfo[_ModelType]]
+_AnyModelAfterValidator = Union[ModelAfterValidator[_ModelType], ModelAfterValidatorWithoutInfo[_ModelType]]
 
 
 @overload
@@ -443,7 +443,7 @@ def model_validator(
     *,
     mode: Literal['after'],
 ) -> Callable[
-    [_AnyModeAfterValidator[_ModelType]], _decorators.PydanticDescriptorProxy[_decorators.ModelValidatorDecoratorInfo]
+    [_AnyModelAfterValidator[_ModelType]], _decorators.PydanticDescriptorProxy[_decorators.ModelValidatorDecoratorInfo]
 ]:
     ...
 
--- a/pydantic/functional_validators.py
+++ b/pydantic/functional_validators.py
@@ -409,32 +409,17 @@ class ModelBeforeValidator(Protocol):
         ...
 
 
-class ModelAfterValidatorWithoutInfo(Protocol):
-    """A `@model_validator` decorated function signature. This is used when `mode='after'` and the function does not
-    have info argument.
-    """
+ModelAfterValidatorWithoutInfo = Callable[[_ModelType], _ModelType]
+"""A `@model_validator` decorated function signature. This is used when `mode='after'` and the function does not
+have info argument.
+"""
 
-    @staticmethod
-    def __call__(  # noqa: D102
-        self: _ModelType,  # type: ignore
-    ) -> _ModelType:
-        ...
-
-
-class ModelAfterValidator(Protocol):
-    """A `@model_validator` decorated function signature. This is used when `mode='after'`."""
-
-    @staticmethod
-    def __call__(  # noqa: D102
-        self: _ModelType,  # type: ignore
-        __info: _core_schema.ValidationInfo,
-    ) -> _ModelType:
-        ...
-
+ModelAfterValidator = Callable[[_ModelType, _core_schema.ValidationInfo], _ModelType]
+"""A `@model_validator` decorated function signature. This is used when `mode='after'`."""
 
 _AnyModelWrapValidator = Union[ModelWrapValidator, ModelWrapValidatorWithoutInfo]
 _AnyModeBeforeValidator = Union[ModelBeforeValidator, ModelBeforeValidatorWithoutInfo]
-_AnyModeAfterValidator = Union[ModelAfterValidator, ModelAfterValidatorWithoutInfo]
+_AnyModeAfterValidator = Union[ModelAfterValidator[_ModelType], ModelAfterValidatorWithoutInfo[_ModelType]]
 
 
 @overload
@@ -457,7 +442,9 @@ def model_validator(
 def model_validator(
     *,
     mode: Literal['after'],
-) -> Callable[[_AnyModeAfterValidator], _decorators.PydanticDescriptorProxy[_decorators.ModelValidatorDecoratorInfo]]:
+) -> Callable[
+    [_AnyModeAfterValidator[_ModelType]], _decorators.PydanticDescriptorProxy[_decorators.ModelValidatorDecoratorInfo]
+]:
     ...
 
 
--- a/docs/usage/dataclasses.md
+++ b/docs/usage/dataclasses.md
@@ -347,10 +347,10 @@ class User:
         return values
 
     @model_validator(mode='after')
-    def post_root(cls, values: Dict[str, Any]) -> Dict[str, Any]:
-        print(values)
+    def post_root(self) -> 'User':
+        print(self)
         #> User(birth=Birth(year=1995, month=3, day=2))
-        return values
+        return self
 
     def __post_init__(self):
         print(self.birth)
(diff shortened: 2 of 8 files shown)


adriangb:
@Kludex want to try and apply the same thing to the other protocols? We might be able to simplify quite a bit

adriangb:
please review

pydantic/functional_validators.py
```diff
@@ -409,32 +409,17 @@ def __call__(  # noqa: D102
...
-    @staticmethod
-    def __call__(  # noqa: D102
-        self: _ModelType,  # type: ignore
-        __info: _core_schema.ValidationInfo,
-    ) -> _ModelType:
-        ...
+ModelAfterValidatorWithoutInfo = Callable[[_ModelType], _ModelType]
+"""A `@model_validator` decorated function signature. This is used when `mode='after'` and the function does not
+have info argument.
+"""
```
ybressler, line 415: If I understand correctly, mypy didn't recognize the the previously annotated `Protocol` object as `callable`? (Even though it had `call` method? By switching to explicitly annotating with `Callable`, the annotations are correct?

ybressler: I like that we're annotating callable functions as callable though. (Though `Protocol` is supposed to solve this for us, no?)

ybressler (commented):
_**Note:** I am familiarizing myself with the pydantic codebase through reviews – my review is therefore a bit n00b._

Left a Q about the behavior of `Protocol` + `mypy`. (Otherwise, IMO, looks great.)

pydantic/functional_validators.py
adriangb: In theory you can use either, but it's a little bit weird because here we want to require the `self` argument, in particular because it helps distinguish from the other alternative signatures. But my hack to make that work does not seem to have been compatible with mypy so this is a simpler albeit less pedantic version that hopefully will be.

lord-haffi: > here we want to require the `self` argument, in particular because it helps distinguish from the other alternative signatures

I don't know if I'm understanding your thought correctly.
But if those validators work with other signatures as well (e.g. with classmethod signatures) it would be misleading imo. The implementation as far as I saw supports static-, class- and instance-signatures and even decorators. Therefore, I don't think you have to enforce an instance-signature here (i.e. with self argument).

Actually, even the [documentation](https://docs.pydantic.dev/latest/usage/validators/#model-validators) suggests a classmethod-signature for `model_validator` with `mode="after"`. Does the solution of this PR work with classmethod-signatures though, did you test it?

pydantic/functional_validators.py
```diff
@@ -409,32 +409,17 @@ def __call__(  # noqa: D102
...
 
+ModelAfterValidator = Callable[[_ModelType, _core_schema.ValidationInfo], _ModelType]
+"""A `@model_validator` decorated function signature. This is used when `mode='after'`."""
 
 _AnyModelWrapValidator = Union[ModelWrapValidator, ModelWrapValidatorWithoutInfo]
 _AnyModeBeforeValidator = Union[ModelBeforeValidator, ModelBeforeValidatorWithoutInfo]
-_AnyModeAfterValidator = Union[ModelAfterValidator, ModelAfterValidatorWithoutInfo]
+_AnyModeAfterValidator = Union[ModelAfterValidator[_ModelType], ModelAfterValidatorWithoutInfo[_ModelType]]
```
lord-haffi, line 422: Based on line 420: Is this a typo? (If so, maybe correct 421 as well)
```suggestion
_AnyModelAfterValidator = Union[ModelAfterValidator[_ModelType], ModelAfterValidatorWithoutInfo[_ModelType]]
```

lord-haffi (commented):
Just ran into this issue today and found the issue and this PR and thought why not telling my thoughts on it :)
I really like the idea of `Protocol`. But annotating callables with it with an explicit self-argument doesn't really seem to work with mypy. However, pyright seems to get it to work. I found this little discussion (more like an issue) regarding this topic if anyone is interested:
https://github.com/python/typing/discussions/1040

The PR looks good, but I had a question as well.
Thanks for the good work, appreciate it :)

pydantic/functional_validators.py
adriangb: > Actually, even the [documentation](https://docs.pydantic.dev/latest/usage/validators/#model-validators) suggests a classmethod-signature for model_validator with mode="after"

This is precisely what we wanted to avoid, the docs are wrong.

pydantic/functional_validators.py
adriangb: Thanks, fixed!

pydantic/functional_validators.py
lord-haffi: Ok, I tried around by myself but I really couldn't get a Protocol typing for instance methods to work together with the decorator. At least with mypy. Very sad, maybe it's possible in the future.

Btw, why do you even want to avoid classmethod signatures? Is there anywhere a discussion about this?

adriangb: > Btw, why do you even want to avoid classmethod signatures? Is there anywhere a discussion about this?

You're getting an instance of the model. Of course `(cls: Type['Model'], model: 'Model')` works but why do that instead of `(self)` which works great with type checkers, etc.

pydantic/functional_validators.py
```diff
@@ -409,32 +409,17 @@ def __call__(  # noqa: D102
...
+have info argument.
+"""
 
+ModelAfterValidator = Callable[[_ModelType, _core_schema.ValidationInfo], _ModelType]
+"""A `@model_validator` decorated function signature. This is used when `mode='after'`."""
 
 _AnyModelWrapValidator = Union[ModelWrapValidator, ModelWrapValidatorWithoutInfo]
 _AnyModeBeforeValidator = Union[ModelBeforeValidator, ModelBeforeValidatorWithoutInfo]
-_AnyModeAfterValidator = Union[ModelAfterValidator, ModelAfterValidatorWithoutInfo]
+_AnyModelAfterValidator = Union[ModelAfterValidator[_ModelType], ModelAfterValidatorWithoutInfo[_ModelType]]
```
dmontagu, line 422: Does this reparametrization do anything? It looks like the same typevar is used in the original definition of this thing. Is this for pseudo-documentation reasons or does it type check differently? Should it make use of a TypeAlias?

adriangb: It's because generic aliases require the type var reparametrization

dmontagu (approved):

lmrlyv:
Hi @adriangb,
Thanks for the fix!

Since this commit is already merged and addresses the typing issue, I don't want to silence it in our codebase. 
Could you please share the planned date for the next release?

Kludex:
I believe that either today, or tomorrow (likely).

Issue #6489, opened 2023-07-06
Computed field errors during declaration with forward referenced return type


zakstucke:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

Computed fields errors when used with forward referenced types. 

Pydantic errors on model declaration itself stating the type is not defined.

datetime module used for simple example, in reality hit when a forward reference was actually needed due to circular reference constraints.

```
Traceback (most recent call last):

  File "/tmp/ipykernel_44291/634715983.py", line 10, in <module>
    class Foo(BaseModel):

  File "/tmp/ipykernel_44291/634715983.py", line 11, in Foo
    @computed_field
     ^^^^^^^^^^^^^^

  File "python3.11/site-packages/pydantic/fields.py", line 995, in computed_field
    return dec(__f)
           ^^^^^^^^

  File "python3.11/site-packages/pydantic/fields.py", line 978, in dec
    return_type = _decorators.get_function_return_type(f, return_type)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "python3.11/site-packages/pydantic/_internal/_decorators.py", line 712, in get_function_return_type
    hints = get_function_type_hints(unwrap_wrapped_function(func), include_keys={'return'})
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "python3.11/site-packages/pydantic/_internal/_typing_extra.py", line 249, in get_function_type_hints
    type_hints[name] = typing._eval_type(value, globalns, None)  # type: ignore
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/typing.py", line 371, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/typing.py", line 877, in _evaluate
    eval(self.__forward_code__, globalns, localns),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "<string>", line 1, in <module>

NameError: name 'datetime' is not defined
```

[...]

PR #6516, opened 2023-07-07
Defer evaluation of return type hints for serializer functions

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6489

Also fixes the same issue for `PlainSerializer` etc.

skip change file check

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/functional_serializers.py
+++ b/pydantic/functional_serializers.py
@@ -42,7 +42,7 @@ class PlainSerializer:
         schema = handler(source_type)
         try:
             return_type = _decorators.get_function_return_type(
-                self.func, self.return_type, handler.get_types_namespace()
+                self.func, self.return_type, handler._get_types_namespace()
             )
         except NameError as e:
             raise PydanticUndefinedAnnotation.from_name_error(e) from e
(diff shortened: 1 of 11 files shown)


dmontagu:
need to add tests

please update

pydantic/functional_serializers.py
```diff
@@ -39,8 +40,13 @@ def __get_pydantic_core_schema__(
             The Pydantic core schema.
         """
         schema = handler(source_type)
-        return_type = _decorators.get_function_return_type(self.func, self.return_type)
-        return_schema = None if return_type is None else handler.generate_schema(return_type)
+        try:
+            return_type = _decorators.get_function_return_type(
+                self.func, self.return_type, handler.get_types_namespace()
+            )
```
adriangb, line 46: Hmm if this is the only place this is being used I'm inclined to make `_get_types_namespace` a private attribute and just `type: ignore` it, only exposing it when someone else asks for it.

pydantic/functional_serializers.py
```diff
@@ -136,7 +147,7 @@ def field_serializer(
 def field_serializer(
     *fields: str,
     mode: Literal['plain', 'wrap'] = 'plain',
-    return_type: Any = None,
+    return_type: Any = PydanticUndefined,
```
adriangb, line 150: Good call on this, obvious in retrospect

adriangb (approved):
Looks good, please add tests. please update

dmontagu:
please review

PR #6527, opened 2023-07-08
Update _redirects

Kludex:
skip change file check

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -1,6 +1,6 @@
 / /latest/ 301
-/2.0.1/:else/ /2.0/:else/ 301
-/2.0.2/:else/ /2.0/:else/ 301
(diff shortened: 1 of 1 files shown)


_redirects
```diff
@@ -1,6 +1,6 @@
 / /latest/ 301
-/2.0.1/ /2.0/ 301
-/2.0.2/ /2.0/ 301
+/2.0.1/* /2.0/* 301
+/2.0.2/* /2.0/* 301
```
samuelcolvin, line 3: I think this is wrong, should be `:splat`

samuelcolvin: For both

samuelcolvin (approved):

PR #6535, opened 2023-07-08
Add a `replace_types` option to `ConfigDict`

### LINKED ISSUE #6045: V2 - Validating int->str
Continue of the #5993

```Python
class Data(BaseModel):
    s: str
    f: str
    d: str


```

### Affected Components

vitalik:
## Related issue number

Fixes #6045

Introduced new config option `replace_types` that allows to patch fields annotations after class creation:

```Python
LaxStr = Annotated[str, BeforeValidator(str_validator)]

class LaxStrBase(BaseModel):
    model_config: ConfigDict = {'replace_types': {str: LaxStr}} # !!!!!!!!!

class Data(LaxStrBase):
    f: str
    i: str
    d: str

user = Data(f=1.1, i=1, d=Decimal('1.1'))
# will equal to {'f': '1.1', 'i': '1', 'd': '1.1'}
```

this also allows a pydantic global override string behaviour:
```Python
BaseModel.model_config['replace_types'] = {str: LaxStr}
```

## Checklist

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -9,6 +9,7 @@ from dirty_equals import HasRepr, IsPartialDict
 from pydantic_core import SchemaError
 
 from pydantic import (
+    AfterValidator,
     BaseConfig,
     BaseModel,
     BeforeValidator,
@@ -680,3 +681,15 @@ def test_replace_types_nested():
             'url': 'https://errors.pydantic.dev/2.1.2/v/string_type',
         }
     ]
+
+
+def test_replace_types_annotated():
+    DoubleStr = Annotated[str, AfterValidator(lambda x: x * 2)]
+
+    class Model(BaseModel):
+        x: Annotated[str, AfterValidator(str.lower)]
+        model_config = {'replace_types': {str: DoubleStr}}
+
+    user = Model(x='ABC')
+
+    assert user.model_dump() == {'x': 'abcabc'}
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -11,6 +11,7 @@ from pydantic_core import SchemaError
 from pydantic import (
     BaseConfig,
     BaseModel,
+    BeforeValidator,
     Field,
     PrivateAttr,
     PydanticDeprecatedSince20,
@@ -23,6 +24,7 @@ from pydantic._internal._config import ConfigWrapper, config_defaults
 from pydantic.config import ConfigDict
 from pydantic.dataclasses import dataclass as pydantic_dataclass
 from pydantic.errors import PydanticUserError
+from pydantic.v1.validators import str_validator
 
 if sys.version_info < (3, 9):
     from typing_extensions import Annotated
@@ -633,3 +635,19 @@ def test_config_inheritance_with_annotations():
         model_config: ConfigDict = {'str_to_lower': True}
 
     assert Child.model_config == {'extra': 'allow', 'str_to_lower': True}
+
+
+def test_replace_types():
+    LaxStr = Annotated[str, BeforeValidator(str_validator)]
+
+    class LaxStrBase(BaseModel):
+        model_config: ConfigDict = {'replace_types': {str: LaxStr}}
+
+    class Data(LaxStrBase):
+        f: str
+        i: str
+        d: str
+
+    user = Data(f=1.1, i=1, d=Decimal('1.1'))
+
+    assert user.model_dump() == {'f': '1.1', 'i': '1', 'd': '1.1'}
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -651,3 +651,32 @@ def test_replace_types():
     user = Data(f=1.1, i=1, d=Decimal('1.1'))
 
     assert user.model_dump() == {'f': '1.1', 'i': '1', 'd': '1.1'}
+
+
+def test_replace_types_nested():
+    LaxStr = Annotated[str, BeforeValidator(str_validator)]
+
+    class Inner(BaseModel):
+        x: str
+
+    class Outer(BaseModel):
+        x: str
+        inner: Inner
+        model_config = {'replace_types': {str: LaxStr}}
+
+    user = Outer.model_validate({'x': 1, 'inner': {'x': 'a'}})
+    assert user.model_dump() == {'x': '1', 'inner': {'x': 'a'}}
+
+    with pytest.raises(ValidationError) as exc_info:
+        Outer.model_validate({'x': 1, 'inner': {'x': 1}})
+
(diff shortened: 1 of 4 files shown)


adriangb:
I like this idea! My main concern is the implementation, in particular, that it is done by replacing fields and only for BaseModel. It seems to me like it could be made to work for dataclasses and typeddict without much more effort.

vitalik:
> replacing fields and only for BaseModel.

not sure I understand... - it should work for global (BaseModel), some base model (and only child classes) and a single class:

```Python

# 1 this changes globally
BaseModel.model_config['replace_types'] = {str: LaxStr}

# 2 this changes types only for child classes inherited from LaxStrBase

class LaxStrBase(BaseModel):
    model_config: ConfigDict = {'replace_types': {str: LaxStr}} 

# 3 and for one single model:

class MyModel(BaseModel):
    model_config: ConfigDict = {'replace_types': {str: LaxStr}} 
    a: str
    b: str
    x: OtherStr

```

adriangb:
I mean:

```python
from typing_extensions import TypedDict
from pydantic import ConfigDict

@dataclass
class Model1:
    x: str
    __pydantic_config__ = ConfigDict(replace_types={str: int})

class Model2(TypedDict):
    x: str
    __pydantic_config__ = ConfigDict(replace_types={str: int})
```

adriangb:
A couple more things to think about:

- Handling generic types (`replace_types={List: Dict}`), how can users handle replacing the origin vs. replacing a parametrized version
- Recursive replacements, both that `{str: Annotated[str, ...]}` should cause recursive replacement and that `{str: int, int: float}` should (or not?) lead to `str` being replaced with `float`.

vitalik:
> * Handling generic types (`replace_types={List: Dict}`), how can users handle replacing the origin vs. replacing a parametrized version

I do not see any practicality in such a case, but even if it used - it will not replace concrete annotation with origins because the equal statement is used, and it will not work for cases without passed origins:

```
List != List[str]
List != List[Any]
str != Annotated[str, 'meta']
ans so on
```

> * Recursive replacements, both that `{str: Annotated[str, ...]}` should cause recursive replacement and that `{str: int, int: float}` should (or not?) lead to `str` being replaced with `float`.

Again - practicality is questionable, but yeah why not let user loop all the way `str -> int -> float` for some unknown reason (just keep keys of the dict in a required sort order)

I think the main practicality here is in enhancing types with extra Annotated markers or dates with datetimes with timezones. I do not see any useful case where you replace all your strings with integers or any other unrelated to each other types.

I guess if you really need some magic with handling container types and origins - the `replace_types` can be a callback function that takes class and FieldInfo and patches is it in any imaginable way:


```Python

def patch_strings(cls: BaseModel, field_name: str, info: FieldInfo):
    if info.annotation == str or info.annotation.origin == str:
         info.annotation = LaxStr
         info.summary = 'it can take numbers ;)'
    if name == 'age':
         info.annotation = Annotated[int, Range(0, 100)]

class LaxStrBase(BaseModel):
    model_config: ConfigDict = { 'replace_types': patch_strings }
```

adriangb:
Thanks for thinking about these things. I agree these aren't dealbreakers for the feature, I just wanted us to think about them a bit, thank you for keeping the focus on the core use cases this is solving. Practically speaking I see the value in this and think it'd be totally okay to start with no generics or recursive replacement support. I do think we need to get this to work with `TypedDict`, `dataclasses` at least, those two can also have a `Config`. Do you want to give that a shot?

tests/test_config.py
```diff
@@ -633,3 +635,48 @@ class Child(Parent):
...
+        model_config: ConfigDict = {'replace_types': {str: LaxStr}}
+
+    class Data(LaxStrBase):
+        f: str
+        i: str
+        d: str
+
+    user = Data(f=1.1, i=1, d=Decimal('1.1'))
+
+    assert user.model_dump() == {'f': '1.1', 'i': '1', 'd': '1.1'}
```
ybressler, line 653: Consider parameterizing this test case – I would expect to see test examples for Nonetypes – ex, `None, 0, 'nan'`.

tests/test_config.py
```diff
@@ -633,3 +635,48 @@ class Child(Parent):
...
+def test_replace_types_nested():
+    LaxStr = Annotated[str, BeforeValidator(str_validator)]
+
+    class Inner(BaseModel):
+        x: str
+
+    class Outer(BaseModel):
+        x: str
+        inner: Inner
+        model_config = {'replace_types': {str: LaxStr}}
```
ybressler, line 665: Interesting. Is this the behavior for other config settings? That they don't bubble down? If so, perhaps add this to the test docstring.

Also, would be helpful to create a ticket for documenting this behavior as follow up work (just so this doesn't get lost in the mix).

ybressler (changes requested):
I really love the implementation here. Offered some small code suggestions with your tests.

Code aside, I wonder if this change is in alignment with the direction of Pydantic V2. I personally really like this feature and would use it myself. But I am curious to see what the core contributors' take is on this.

vitalik:
@adriangb back from #6045 

I think patching FieldInfo.annotation might not perfect - I'm trying to  do monkey patching on other large project but stuck with this:

```Python
from pydantic import BaseModel, Field, BeforeValidator
from typing import Annotated
from pydantic.v1.validators import str_validator
from pydantic._internal import _fields
from pydantic._internal import _model_construction


LaxStr = Annotated[str, BeforeValidator(str_validator)]

# patching orig_collect_model_fields

orig_collect_model_fields = _fields.collect_model_fields


def patched_collect_model_fields(*a, **kw):
    fields, classvars = orig_collect_model_fields(*a, **kw)
    for f in fields.values():
        if f.annotation == str:
            f.annotation = LaxStr
    return fields, classvars


_model_construction.collect_model_fields = patched_collect_model_fields


class Data(BaseModel):
    gender: str = Field('F', max_length=1)


print(Data(gender='M'))

```

gives:
```
gender
  Value should have at most 1 item after validation, not 1 [type=too_long, input_value='M', input_type=str]
    For further information visit https://errors.pydantic.dev/2.0.3/v/too_long
```
probably some extra thing need to be done ?

adriangb:
Indeed that does seem brittle. I think this needs to happen during core schema generation.

lig:
@vitalik is this PR still on progress or it's ready for review?

adriangb:
I think it may still need some work but primarily it requires our input

samuelcolvin:
closing as we're working on another solution.

fortzi:
@samuelcolvin can you please share what's planned to solve this issue?

Issue #6525, opened 2023-07-08
Discriminator needs literal error when validating literal field


asisloustau:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

When defining a `field_validator` for a `Literal` field that is used in a discriminated union, A `PydanticUserError` is raised, prompting that the field needs to be of type `Literal`.

If the 2 `field_validator`s are commented out  in the example below, the models will be created without issue.

Is this expected, or is this a bug? I couldn't find any documentation referring to this edge case.

Thank you!

### Example Code

```Python
import typing as t

from pydantic import BaseModel, Field, field_validator


class AnimalModel(BaseModel):
    age_years: int


class CatModel(AnimalModel):
    name: t.Literal["kitty", "cat"]
    favorite_lasagna_brand: str
    # comment out the 2 field validators and model will work!
    @field_validator("name", mode="after")
    def replace_name(cls, v):
        return "cat"


class SparrowModel(AnimalModel):
    name: t.Literal["sparrow", "birdie"]
    color: str
    # comment out the 2 field validators and model will work!
    @field_validator("name", mode="after")
    def replace_name(cls, v):
        return "sparrow"


AllowedAnimalTypes = t.Annotated[
    CatModel | SparrowModel,
    Field(discriminator="name"),
]


class ZooModel(BaseModel):
    zoo_name: str
    animals: t.List[AllowedAnimalTypes]


if __name__ == "__main__":
    cool_zoo = ZooModel(
        zoo_name="Zoolandia",
        animals=[
            SparrowModel(name="birdie", color="red", age_years=2),
            CatModel(name="kitty", favorite_lasagna_brand="Tony's", age_years=20),
        ],
    )

    # raises PydanticUserError(
    # pydantic.errors.PydanticUserError: Model 'CatModel' needs field 'name' to be of type `Literal`
    # For further information visit https://errors.pydantic.dev/2.0.2/u/discriminator-needs-literal
```

### Python, Pydantic & OS Version

```Text
❯ python -c "import pydantic.version; print(pydantic.version.version_info())"
             pydantic version: 2.0.2
        pydantic-core version: 2.1.2 release build profile
                 install path: /home/aloustau/sandbox/pydantic2_learning/.venv/lib/python3.10/site-packages/pydantic
               python version: 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0]
[...]

PR #6541, opened 2023-07-09
Fix building of discriminated union where field_validator wraps a Literal

adriangb:
skip change file check

Closes https://github.com/pydantic/pydantic/issues/6525

### DIFF
--- a/tests/test_discriminated_union.py
+++ b/tests/test_discriminated_union.py
@@ -8,7 +8,7 @@ from dirty_equals import HasRepr, IsStr
 from pydantic_core import SchemaValidator, core_schema
 from typing_extensions import Annotated, Literal
 
-from pydantic import BaseModel, ConfigDict, Field, TypeAdapter, ValidationError
+from pydantic import BaseModel, ConfigDict, Field, TypeAdapter, ValidationError, field_validator
 from pydantic._internal._discriminated_union import apply_discriminator
 from pydantic.errors import PydanticUserError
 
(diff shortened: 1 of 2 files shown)


tests/test_discriminated_union.py
```diff
@@ -1232,3 +1232,72 @@ class TestModel(BaseModel):
...
+            return 'sparrow'
+
+    Animal = Annotated[
+        CatModel | SparrowModel,
+        Field(discriminator='name'),
+    ]
+
+    ta = TypeAdapter(Animal)
+
+    assert ta.validate_python({'name': 'kitty'}) == CatModel(name='cat')
```
ybressler, line 1261: What if both models have a common value – I would expect the firstmost defined object to return. Worth testing?

tests/test_discriminated_union.py
```diff
@@ -1232,3 +1232,72 @@ class TestModel(BaseModel):
...
+            'mapping': {
+                'birdie': '#/$defs/SparrowModel',
+                'cat': '#/$defs/CatModel',
+                'kitty': '#/$defs/CatModel',
+                'sparrow': '#/$defs/SparrowModel',
+            },
+            'propertyName': 'name',
+        },
+        'oneOf': [{'$ref': '#/$defs/CatModel'}, {'$ref': '#/$defs/SparrowModel'}],
+    }
```
ybressler, line 1303: Thorough! But a bit tedious to write + replicate. Any way to simplify this part of the code?

If what we care about are the `$defs` then maybe:
```
    assert ta.json_schema() == {
        '$defs': {
            CatModel.json_schema(),
            SparrowModel.json_schema(),
        },
        'discriminator': {
            'mapping': {
                'birdie': '#/$defs/SparrowModel',
                'cat': '#/$defs/CatModel',
                'kitty': '#/$defs/CatModel',
                'sparrow': '#/$defs/SparrowModel',
            },
            'propertyName': 'name',
        },
        'oneOf': [{'$ref': '#/$defs/CatModel'}, {'$ref': '#/$defs/SparrowModel'}],
    }
```

ybressler (commented):
Code looks good! Left a few inquisitive Q's :)

tests/test_discriminated_union.py
adriangb: Yes indeed! But that's already tested elsewhere (in this file, I can look for the right line tomorrow; if it's not us should be!)

adriangb: (To be clear, I believe the actual behavior is to error during schema building)

tests/test_discriminated_union.py
adriangb: Indeed that'd be less verbose, but the `insert_assert` utility does snapshot testing so this is not hand written and doesn't need to be updated by hand 😀

adriangb:
please review cc @dmontagu

ybressler (approved):
Great work.

adriangb:
Closing in favor of #6570

Issue #5656, opened 2023-04-30
Use property docstrings as descriptions


rattrayalex:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this feature is missing

### Description

_Moving from [this comment](https://github.com/pydantic/pydantic/issues/2947#issuecomment-1067378353) on [a related issue](https://github.com/pydantic/pydantic/issues/2947), where people seemed interested in the feature._

I would like this:
```py
class Foo(BaseModel):
    bar: float
    """Short explanation."""
```

To be equivalent to this:

```py
class Foo(BaseModel):
    bar: float = Field(description="Short description.")
    """Short explanation."""
```

If you provide both, the `description` should win.

Reasons:

1. It's a nicer, cleaner, standard Pythonic syntax.
2. The docstring appears in hover in editors like VS Code, but the Field description does not (though note https://github.com/pydantic/pydantic/issues/638 would remove this limitation).

Would that be possible?

Note this feature request seems like the inverse of https://github.com/pydantic/pydantic/issues/638, where the docstrings are pulled from the description, which I think is also a good idea. Ideally we'd have both IMO, since this syntax 

### Affected Components

- [ ] [Compatibility between releases](https://docs.pydantic.dev/changelog/)
- [ ] [Data validation/parsing](https://docs.pydantic.dev/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://docs.pydantic.dev/usage/exporting_models/) - `.model_dump()` and `.model_dump_json()`
- [ ] [JSON Schema](https://docs.pydantic.dev/usage/schema/)
- [ ] [Dataclasses](https://docs.pydantic.dev/usage/dataclasses/)
- [X] [Model Config](https://docs.pydantic.dev/usage/model_config/)
- [ ] [Field Types](https://docs.pydantic.dev/usage/types/) - adding or changing a particular data type
- [ ] [Function validation decorator](https://docs.pydantic.dev/usage/validation_decorator/)
- [ ] [Generic Models](https://docs.pydantic.dev/usage/models/#generic-models)
- [ ] [Other Model behaviour](https://docs.pydantic.dev/usage/models/) - `model_construct()`, pickling, private attributes, ORM mode
- [ ] [Plugins](https://docs.pydantic.dev/) and integration with other tools - mypy, FastAPI, python-devtools, Hypothesis, VS Code, PyCharm, etc.

[...]

PR #6563, opened 2023-07-10
Extract attribute docstrings for `FieldInfo.description`

Viicos:
## Change Summary

This implements attribute docstring parsing when the `FieldInfo.description` attribute is `None` (that is it wasn't set by the user). Thanks to @danields761 for the original implementation (which is really clean) of which the following modifications have been made:
- Only use quotes docstrings ([`#` comments](https://github.com/danields761/pydantic-settings/#extracts-fields-documentation) not implemented), e.g:
```python
class A(BaseModel):
    a: int
    "doc"
    b: int
    """triple quotes supported"""
    c: int

    """Spaces supported as well"""
```
- Support spaces between assignment and docstring (unlike the [original implementation](https://github.com/danields761/class-doc/blob/9b122d85ce667d096ebee75a49350bbdbd48686d/class_doc.py#L232)) (this removes the need for a custom set of ast routines, and VSCode seems to [support spaces](https://github.com/pydantic/pydantic/assets/65306057/1bd6a6bf-dfc5-43aa-aee6-7edd7009a849) as well)

A few things to take into account:

- This feature won't be available in some specific environments (e.g. when in an interactive session). The reason is `inspect.getsource`.
- Maybe a `model_config` value should be added to activate this feature (with a default value of `False`), considering some people might set some "private" information in these docstrings that could get exposed with FastAPI for example.

I would understand if this feature doesn't make it into Pydantic, as it could be considered a breaking change, and relying on `inspect` feels a bit hacky to me. But it would be a great feature nonetheless, considering @tiangolo [had something similar](https://github.com/pydantic/pydantic/discussions/4281#discussioncomment-6006803) in mind.

I'll wait for a review before adding tests and documentation

## Related issue number

Closes https://github.com/pydantic/pydantic-settings/issues/2.
Related discussion: https://github.com/pydantic/pydantic/discussions/4281.

EDIT by @Kludex:
- Closes https://github.com/pydantic/pydantic/issues/5656

## Checklist


pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+        # `target` can only be `ast.Name` in the context of a `BaseModel` subclass definition
+        yield node.target.id
+
+
+def _extract_docs(cls_def: ast.ClassDef) -> dict[str, str]:
+
+    nodes_docs_pairs = [
+        (node, inspect.cleandoc(next_node.value.s))
+        for node, next_node in zip(cls_def.body, cls_def.body[1:])
+        if isinstance(node, (ast.Assign, ast.AnnAssign))  # e.g. a: int / a = 1...
```
Viicos, line 27: We could simplify even more by disregarding `ast.Assign`, which isn't possible in a Pydantic model:

```python
class A(BaseModel):
  a = 1
# throws PydanticUserError
```

Viicos:
Please review

pydantic/_internal/_docs_extraction.py
davidhewitt: If this is also true for dataclasses, I would be ok with taking the simpler route.

Viicos: dataclasses can have simple assignments (same goes for Pydantic dataclasses). But they aren't taken into account:

```python

@dataclass
class A:
  a: int = 1
  b = 2

A.__pydantic_fields__
#> {'a': FieldInfo(annotation=int, required=False, default=1, init_var=False, kw_only=False)}
# Same goes for A.__dataclass_fields__, which is used to build __pydantic_fields__
```

Viicos:
> If `Field(description="...")` is set as well as then should it error? I think no, probably best to use the `description` in that case (i.e. for cases when you want the Python and schema to diverge).

Yes currently `description` has a higher priority and is used even if docstring is set. Having an error would be a breaking change

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+        cls: The class of the Pydantic model to inspect.
+    Returns:
+        A mapping containing attribute names and their corresponding docstring.
+    """
+    try:
+        source = inspect.getsource(cls)
+    except OSError:
+        # Source can't be parsed (maybe because running in an interactive terminal)
+        # TODO should we issue a UserWarning to indicate that docstrings won't be
+        # taken into account?
```
Kludex, line 54: I think it is fine to have empty string as description in these cases. No need to warn.

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+    return doc_mapping
+
+
+def extract_docstrings_from_cls(cls: type[Any]) -> dict[str, str]:
+    """Returns a mapping of the model attributes and their corresponding docstring.
+
+    Args:
+        cls: The class of the Pydantic model to inspect.
+    Returns:
+        A mapping containing attribute names and their corresponding docstring.
```
Kludex, line 47:
```suggestion

    Returns:
        A mapping containing attribute names and their corresponding docstring.
```

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+
+    for node, docs in nodes_docs_pairs:
+        for target in _get_assign_targets(node):
+            doc_mapping[target] = docs
+
+    return doc_mapping
+
+
+def extract_docstrings_from_cls(cls: type[Any]) -> dict[str, str]:
+    """Returns a mapping of the model attributes and their corresponding docstring.
```
Kludex, line 42:
```suggestion
    """Map model attributes and their corresponding docstring.
```

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+        cls: The class of the Pydantic model to inspect.
+    Returns:
+        A mapping containing attribute names and their corresponding docstring.
+    """
+    try:
+        source = inspect.getsource(cls)
```
Kludex, line 50: From the [Python documentation](https://docs.python.org/3/library/inspect.html#inspect.getsource):

"A [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is raised if the object is a built-in module, class, or function."

We can't have this scenario here?

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+    except OSError:
+        # Source can't be parsed (maybe because running in an interactive terminal)
+        # TODO should we issue a UserWarning to indicate that docstrings won't be
+        # taken into account?
+        return {}
+
+    text = textwrap.dedent(source)
```
Kludex, line 57: Why is this necessary? The AST should be valid without it. 🤔

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+
+    tree = ast.parse(text).body[0]
+    if not isinstance(tree, ast.ClassDef):
+        raise TypeError(f"Expected '{ast.ClassDef.__name__}', but received '{cls}' instead")
```
Kludex, line 61: Is there a scenario in which the first (and only element) in the body is not a `ast.ClassDef`?

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+def _get_assign_targets(node: ast.Assign | ast.AnnAssign) -> Iterable[str]:
+    if isinstance(node, ast.Assign):
+        for target in node.targets:
+            if isinstance(target, ast.Tuple):
+                yield from (el.id for el in target.elts)
+            else:
+                yield target.id
+    else:
+        # `target` can only be `ast.Name` in the context of a `BaseModel` subclass definition
+        yield node.target.id
```
Kludex, line 19: What's the scenario that we need to supporrt `ast.Assign`? It's not possible to build a `BaseModel` with them, is it?

Kludex (commented):
Can't we instead create an `ast.Visitor`?

Something like:
```py
from _ast import AnnAssign, Expr
from typing import Any
import inspect
import ast


class Model(BaseModel):
    a: int
    """My docstring."""
    b: int = 1
    """Docstring"""
    c: int = 1
    # This doesn't get to the AST.


class Visitor(ast.NodeVisitor):
    def __init__(self) -> None:
        super().__init__()

        self.target: str | None = None
        self.attrs: dict[str, str] = {}

    def visit_AnnAssign(self, node: AnnAssign) -> Any:
        if isinstance(node.target, ast.Name):
            self.target = node.target.id

    def visit_Expr(self, node: Expr) -> Any:
        if isinstance(node.value, ast.Str):
            docstring = inspect.cleandoc(node.value.s)
            if self.target:
                self.attrs[self.target] = docstring
            self.target = None


source = inspect.getsource(Model)
tree = ast.parse(source)

visitor = Visitor()
visitor.visit(tree)
print(visitor.attrs)
```

EDIT: I guess we need to make sure the `Expr` is the next node? 🤔 That can be done implementing the `def visit()`.

pydantic/_internal/_docs_extraction.py
Viicos: `cls` is supposed to be a `BaseModel`, but I'll add the `TypeError` just to be safe

Kludex: I don't like defensive programming. If you know is a `BaseModel`, no need to add it.

pydantic/_internal/_docs_extraction.py
Viicos: I think it's a [leftover](https://github.com/danields761/class-doc/blob/9b122d85ce667d096ebee75a49350bbdbd48686d/class_doc.py#L266) from `class-doc`, I'll remove it

pydantic/_internal/_docs_extraction.py
Viicos: If the argument passed to `cls` isn't a class, but that shouldn't happen

Viicos:
Thanks @Kludex for the review, I'll apply the modifications and try using `NodeVisitor`, which seems like a good fit as well

Kludex:
> Thanks @Kludex for the review, I'll apply the modifications and try using `NodeVisitor`, which seems like a good fit as well

Cool. 👍 

You need to make sure the sequence of nodes is valid i.e. `ast.AnnAssign` comes before `ast.Expr(ast.Str)`.

adriangb:
Will this work for dataclasses? TypedDict? NamedTuple?

Viicos:
> Will this work for dataclasses? TypedDict? NamedTuple?

I've implemented it for dataclasses, and will look into TypedDict and NamedTuple

pydantic/_internal/_docs_extraction.py
Viicos: In fact it is required to avoid `IndentationError`s when models are defined in nested levels (e.g. in functions).

Kludex: Can you show me an example?

Viicos: <details><summary>MRE</summary>

```python
def test():
    class Model(BaseModel):
        a: int
        """doc"""

    source = inspect.getsource(Model)
    tree = ast.parse(source)

test()
#> IndentationError: unexpected indent
```

</details>

Kludex: Ah... Ok. 👍

Kludex:
Beautiful! 👏

Viicos:
> > Will this work for dataclasses? TypedDict? NamedTuple?
> 
> I've implemented it for dataclasses, and will look into TypedDict and NamedTuple

This will probably require some additionnal work for `TypedDict` and `NamedTuple`

tests/test_docs_extraction.py
```diff
@@ -0,0 +1,139 @@
...
+        b: str = '1'
+
+        """B docs"""
+
+    assert Model.model_fields['a'].description is None
+    assert Model.model_fields['b'].description is None
+
+
+def test_model_docs_extraction():
+    class Model(BaseModel):
```
Viicos, line 21: Unfortunately `inspect.getsource` isn't working as expected with classes. Tests are failing under 3.9 because prior to https://github.com/python/cpython/pull/10307, the following would happen:

<details><summary>MRE</summary>

```python
import inspect


def test1():
    class Model:
        a = 1

    print(inspect.getsource(Model))


def test2():
    class Model:
        a = 2

    print(inspect.getsource(Model))


test1()
#> uses Model with a = 1
test2()
#> uses Model with a = 1 (wrong!)
```
</details>

This has been fixed since then, but encountered this one: https://github.com/python/cpython/issues/106727

I think this isn't a big deal, as users will probably not defined two models with the same name in the same module.

adriangb: That kinda sounds like a dealbreaker to me. I don’t know why you think it’s not common, it seems to me like it would happen all the time in tests.

Viicos: Imo it's not common from a user perspective (as this only occurs if in the same module). But it is indeed way more common in tests 😕

adriangb: Users write tests…

Viicos: Ah yes didn't think about that 😅 That was what I was afraid of by using `inspect`, and I don't see any other path except shipping our own source parsing logic, which probably isn't a good idea

Viicos: I'll try to improve class detection on the Python side (see the related issue on cpython), but unfortunately it will not be able to cover all edge cases, such as:

```python
if cond:
    class A: pass
else:
    class A: pass        
```

As this feature would be off by default, users could be warned about the potential unexpected behavior in the documentation. I'll let you decide on the output of this PR

Viicos: So Python got some improvements since then:
- https://github.com/python/cpython/pull/106815
- https://github.com/python/cpython/pull/106968

It works way better if a method is defined on the class (thanks to `__code__.co_firstlineno`), but can still pick the wrong class otherwise.

dmontagu: Am I understanding correctly that this misbehavior is only on Python<=3.9?

Also, when it gets the wrong source, does it lead to actual errors, or does it just retrieve the docstrings incorrectly?

If it only affects <=3.9 and fails gracefully in the sense that it just doesn't get the docstrings correctly, then unless I'm missing something I think the issues with `inspect.getsource` maybe shouldn't be a dealbreaker since it would require all of the following conditions to be met for it to cause problems:

* You are using Python<=3.9
* You have opted into the new config setting for this functionality
* You have defined a model with the same name multiple times in the same module
* You require the description attribute for the `FieldInfo` must be correct
* You aren't happy to (or aren't aware that you can) explicitly specify the description in `Field(...)` for affected attributes

In particular, for the case of writing tests that you mentioned @adriangb I would think that this would very rarely cause problems for tests that aren't explicitly testing this functionality.

If nothing else, it seems reasonable to me to emit a warning if this particular config flag is set in `Python<=3.9`.

(That said, it looks to me like there are other test failures happening right now, not just caused by this issue.)

Viicos: The following PR:
- https://github.com/python/cpython/pull/10307

was introduced in Python 3.9, and fixes class detection when they are nested in other classes.

Up until now, it would still fail with the following:

```python
import inspect

class Model:
    a = 1

class Model:
    a = 2

print(Model.a)
#> 2
print(inspect.getsource(Model))
#> class Model:
#>     a = 1
```

This got improved with the following PR (it will now pick the last class definition):
- https://github.com/python/cpython/pull/106815

But still can get it wrong if you're doing something like:

```python
if False:
    class Model:
        a = 1
else:
    class Model:
        a = 2
# Which really is a uncommon edge case
```

> when it gets the wrong source, does it lead to actual exceptions being raised, or does it just retrieve the docstrings incorrectly?

It will not raise currently, as applying field docstrings in a safe way, but I could maybe raise if we encounter unexpected fields (I'll have to see if it's doable when inheritance is being used).

So to summarize, all of your conditions should indeed be met, except:
- You are using Python<=3.9
- it fails gracefully (can be fixed as I said)

I'll be off for a while and will continue working on this in September if this is still accepted

Viicos:
> Will this work for dataclasses? TypedDict? NamedTuple?

I've Implemented `TypedDict`, left `NamedTuple` as it doesn't seem to support pydantic config. Next I might see if I can raise an exception if the wrong class is being used (to be correct, if the retrieved field to docstring mapping does not match the model fields).

Viicos:
> Next I might see if I can raise an exception if the wrong class is being used (to be correct, if the retrieved field to docstring mapping does not match the model fields).

This ended up being harder than I thought, as currently `extract_docstrings_from_cls` will return any docstring from `AnnAssign` nodes, including private attrs and class vars. So trying to compare the result of this function to the `model_fields` isn't trivial.

alexmojaki:
You may be able to avoid relying on `inspect` to identify the correct class by walking up the stack to find the frame and thus line number where the class is being declared. Here's a proof of concept:

```python
import inspect


class Meta(type):
    def __new__(cls, name, bases, dct):
        result = super().__new__(cls, name, bases, dct)
        inner1(result)
        return result


def inner1(cls):
    inner2(cls)


def inner2(cls):
    frame = inspect.currentframe()
    while frame:
        if frame.f_code == Meta.__new__.__code__:
            lines, _ = inspect.findsource(cls)
            lnum = frame.f_back.f_lineno
            print("----")
            print("".join(inspect.getblock(lines[lnum - 1 : lnum + 1])))
            print("----")
            return

        frame = frame.f_back
    raise RuntimeError("not found")


class A(metaclass=Meta):
    """source 1"""


class A(metaclass=Meta):
    """source 2"""
```

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,74 @@
...
+                lnum = frame.f_back.f_lineno
+                source = inspect.getblock(lines[lnum - 1 : lnum + 1])
+                break
+            frame = frame.f_back
+
+        if not source:
+            return {}
+
+    # Required for nested class definitions, e.g. in a function block
+    dedent_source = textwrap.dedent(''.join(source))
```
alexmojaki, line 72: It's possible for the body of a class (or function) to contain some non-whitespace text at a lower indentation level than the class, in the form of a comment or the contents of some multiline string (docstring or otherwise). In that case `dedent` will still return something that's invalid syntax. A more robust method is to wrap the source in a function or class and then extract the body in the AST.

Viicos:
> You may be able to avoid relying on `inspect` to identify the correct class by walking up the stack to find the frame and thus line number where the class is being declared. Here's a proof of concept:

Thank you really much for this clever idea

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,76 @@
...
+    else:
+        source: Sequence[str] = []
+        frame = inspect.currentframe()
+
+        if frame is None:
+            return {}
+
+        while frame:
+            if frame.f_code is ModelMetaclass.__new__.__code__:
+                lnum = frame.f_back.f_lineno
```
alexmojaki, line 61: I think it'd be safer to keep walking up the stack from this point until you see a line such that `line.strip().startswith("class ")`. I'm worried someone might inherit from the metaclass and then call `super().__new__`.

pydantic/_internal/_docs_extraction.py
Viicos: Do you have an example when `dedent` would still not be enough? I'll add a test with it and try to implement the workaround

alexmojaki:
```python
def foo():
    class A:
        def bar(self):
            print(""""
An inconveniently dedented string...
            """)
```

pydantic/_internal/_docs_extraction.py
Viicos: Indeed, and with this solution we are able to make it work with dataclasses/TypedDicts

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,84 @@
...
+def _extract_source_from_frame(lines: list[str], cls_name: str) -> list[str] | None:
+    source: list[str] = []
+    frame = inspect.currentframe()
+
+    if frame is None:
+        return None
+
+    while frame and frame.f_back:
+        lnum = frame.f_back.f_lineno
+        if isinstance(lnum, int) and len(lines) >= lnum and re.match(fr"$class\s+{cls_name}", lines[lnum - 1].strip()):
```
alexmojaki, line 45:
```suggestion
        if isinstance(lnum, int) and len(lines) >= lnum and re.match(fr"^class\s+{cls_name}", lines[lnum - 1].strip()):
```

alexmojaki: actually with `match` the `^` is implied

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,89 @@
...
+    source: list[str] = []
+    frame = inspect.currentframe()
+
+    if frame is None:
+        return None
+
+    while frame and frame.f_back:
+        lnum = frame.f_back.f_lineno
+        if isinstance(lnum, int) and len(lines) >= lnum and re.match(fr"class\s+{cls_name}", lines[lnum - 1].strip()):
+            source = inspect.getblock(lines[lnum - 1 :])  # type: ignore
```
Viicos, line 46: I might be able to remove the `# type: ignore` with https://github.com/python/typeshed/pull/10920

Viicos:
Thanks @alexmojaki for the feedback, we are now able to use the correct class in any case (hopefully).

alexmojaki (changes requested):
Thoughts for tests:

1. In some tests, please add one or two decorators to the classes that don't do anything, just to make sure that the frame is pointing at the `class` line.
2. Please add a test where the class is defined inside `exec` so that the source code can't be retrieved.
3. Maybe add one or more tests involving model inheritance.

pydantic/_internal/_config.py
```diff
@@ -80,6 +80,7 @@ class ConfigWrapper:
     json_schema_serialization_defaults_required: bool
     json_schema_mode_override: Literal['validation', 'serialization', None]
     coerce_numbers_to_str: bool
+    use_attributes_docstring: bool
```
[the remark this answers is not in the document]
Viicos: Sounds better!

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,89 @@
...
+    def visit_AnnAssign(self, node: ast.AnnAssign) -> Any:
+        if isinstance(node.target, ast.Name):
+            self.target = node.target.id
+
+    def visit_Expr(self, node: ast.Expr) -> Any:
+        if isinstance(node.value, ast.Str) and self.previous_node_type is ast.AnnAssign:
+            docstring = inspect.cleandoc(node.value.s)
+            if self.target:
+                self.attrs[self.target] = docstring
+            self.target = None
```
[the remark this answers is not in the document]
Viicos: Before applying the suggestion/reverting to the old solution, I'll let maintainers decide what they prefer.

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _extract_source_from_frame(lines: list[str], cls_name: str) -> list[str] | None:
+    source: list[str] = []
+    frame = inspect.currentframe()
+
+    if frame is None:
+        return None
```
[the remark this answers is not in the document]
alexmojaki: What I mean is that the while loop below will be skipped and `source` will still be empty, and then `if not source: return None` will trigger, so it's the same either way.

On that note, you could also immediately `return source` instead of `break`.

tests/test_docs_extraction.py
```diff
@@ -18,6 +23,11 @@ class ModelNoDocs(BaseModel):
 
 
 def test_model_docs_extraction():
+
+    # Using a couple dummy decorators to make sure the frame is pointing at
+    # the `class` line:
+    @dec_noop
```
alexmojaki, line 29: I think this would be most useful in `test_model_docs_duplicate_class`, because if decorators are a problem for the frame method then this might only be passing because it's falling back to `inspect`.

pydantic/_internal/_generate_schema.py
```diff
@@ -1144,6 +1150,12 @@ def _typed_dict_schema(self, typed_dict_cls: Any, origin: Any) -> core_schema.Co
                         )[0]
 
                     field_info = FieldInfo.from_annotation(annotation)
+                    if (
+                        field_docstrings is not None
+                        and field_info.description is None
+                        and field_name in field_docstrings
+                    ):
+                        field_info.description = field_docstrings[field_name]
```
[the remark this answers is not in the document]
Viicos: I don't know if this is possible, considering the fact that `TypedDict`s are handled differently, and not with `FieldInfo` objects:
https://github.com/pydantic/pydantic/blob/93e4d233e3ca5bce1e4ac018cbade55f26513a6c/pydantic/_internal/_generate_schema.py#L1123
https://github.com/pydantic/pydantic/blob/93e4d233e3ca5bce1e4ac018cbade55f26513a6c/pydantic/_internal/_generate_schema.py#L1146-L1149

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,89 @@
...
+    # Required for nested class definitions, e.g. in a function block
+    dedent_source = textwrap.dedent(''.join(source))
+    if not dedent_source.startswith('class'):
+        # We are in the case where there's a dedented (usually multiline) string
+        # at a lower indentation level than the class itself. We wrap our class
+        # in a function as a workaround.
+        dedent_source = f'def _():\n{dedent_source}'
+
+    visitor = DocstringVisitor()
+    visitor.visit(ast.parse(dedent_source))
```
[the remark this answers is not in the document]
Viicos: Will it work if the source is indented with tabs and not spaces (or even not 4 spaces)?
(tried the suggestion and made some tests fail)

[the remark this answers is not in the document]
alexmojaki: I'm guessing it would work if the indentation is not 4 spaces, but not if it was tabs. So never mind. Good catch.

tests/test_docs_extraction.py
```diff
@@ -105,6 +132,26 @@ def bar(self):
...
+
+        model_config = ConfigDict(
+            use_attribute_docstrings=True,
+        )
+
+    class SecondModel(FirstModel):
+        a: int
+        """A overridden docs"""
+
+    assert SecondModel.model_fields['a'].description == 'A overridden docs'
```
alexmojaki, line 151: Please also check the field in `FirstModel`

alexmojaki:
Where possible/sensible, can you use the same name for all classes in the file? Just to be extra sure that it doesn't get affected by that.

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,89 @@
...
+def _extract_source_from_frame(lines: list[str], cls_name: str) -> list[str] | None:
+    source: list[str] = []
+    frame = inspect.currentframe()
+
+    if frame is None:
+        return None
+
+    while frame and frame.f_back:
+        lnum = frame.f_back.f_lineno
+        if isinstance(lnum, int) and len(lines) >= lnum and re.match(fr'class\s+{cls_name}', lines[lnum - 1].strip()):
```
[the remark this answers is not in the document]
Viicos: Good catch, this is why tests are currently failing by the way: coincidentally, when walking up the frames, `lnum` is equal to `226` at some point (somewhere in `pydantic/dataclasses.py`) which happen to be on the same line number where another class with the same name is defined in tests:

https://github.com/pydantic/pydantic/blob/4bc1c1a2974558c64145d48a8277bd700236bea0/tests/test_docs_extraction.py#L226

I still need to implement this

pydantic/config.py
```diff
@@ -857,5 +857,12 @@ class Model(BaseModel):
     ```
     """
 
+    use_attribute_docstrings: bool
+    """
+    Whether attributes docstring should be used for `FieldInfo.description`.
```
alexmojaki, line 862:
```suggestion
    Whether docstrings of attributes (bare string literals immediately following the attribute declaration)
    should be used for field descriptions.
    
    ```py
    
    class Model(BaseModel):
      model_config = ConfigDict(use_attribute_docstrings=True)
    
      x: str
      """
      Example of an attribute docstring
      """
    
      y: int = Field(description='Description in Field')
      """
      Description in Field overrides attribute docstring
      """
    
    print(Model.model_fields['x'].description)
    #> Example of an attribute docstring
    print(Model.model_fields['y'].description)
    #> Description in Field
    ```

    This requires the source code of the class to be available at runtime.
```

pydantic/_internal/_docs_extraction.py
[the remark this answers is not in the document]
alexmojaki: Resolved

pydantic/_internal/_generate_schema.py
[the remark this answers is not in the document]
alexmojaki: Resolved

pydantic/config.py
```diff
@@ -244,7 +244,7 @@ class User(BaseModel, validate_assignment=True):  # (1)!
         1 validation error for User
         name
           Input should be a valid string [type=string_type, input_value=123, input_type=int]
-        '''
+            '''
```
alexmojaki, line 247: revert?

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,79 @@
...
+    frame = inspect.currentframe()
+
+    while frame:
+        lnum = frame.f_lineno
+        try:
+            lines, _ = inspect.findsource(frame)
+        except OSError:
+            # Source can't be parsed (maybe because running in an interactive terminal),
+            # we don't want to error here.
+            continue
```
alexmojaki, line 46: I think the tests are currently stuck in an infinite loop at this point.

Are you able to run tests locally before you push?

tests/test_docs_extraction.py
```diff
@@ -0,0 +1,287 @@
...
+
+        class MyModelEnclosed(BaseModel):
+            b: int
+            """B docs"""
+
+            model_config = ConfigDict(
+                use_attribute_docstrings=True,
+            )
+
+    assert MyModelEnclosed.model_fields['a'].description == 'A docs'
```
alexmojaki, line 114:
```suggestion
        class MyModel(BaseModel):
            a: int
            """A docs"""

            model_config = ConfigDict(
                use_attribute_docstrings=True,
            )

    else:

        class MyModel(BaseModel):
            b: int
            """B docs"""

            model_config = ConfigDict(
                use_attribute_docstrings=True,
            )

    assert MyModel.model_fields['a'].description == 'A docs'
```

tests/test_docs_extraction.py
```diff
@@ -0,0 +1,287 @@
...
+            use_attribute_docstrings=True,
+        )
+
+    class SecondModel(FirstModel):
+        a: int
+        """A overridden docs"""
+
+    assert SecondModel.model_fields['a'].description == 'A docs'
+    assert SecondModel.model_fields['a'].description == 'A overridden docs'
+    assert SecondModel.model_fields['b'].description == 'B docs'
```
alexmojaki, line 153:
```suggestion
    class MyModel(BaseModel):
        a: int
        """A docs"""

        b: int
        """B docs"""

        model_config = ConfigDict(
            use_attribute_docstrings=True,
        )

    FirstModel = MyModel

    class MyModel(FirstModel):
        a: int
        """A overridden docs"""

    assert FirstModel.model_fields['a'].description == 'A docs'
    assert MyModel.model_fields['a'].description == 'A overridden docs'
    assert MyModel.model_fields['b'].description == 'B docs'
```

tests/test_docs_extraction.py
```diff
@@ -0,0 +1,287 @@
...
+        foo: int = 123
+        """Shouldn't be used"""
+
+    CreatedModel = create_model(
+        'CreatedModel',
+        foo=(int, 123),
+        __config__=ConfigDict(use_attribute_docstrings=True),
+    )
+
+    assert CreatedModel.model_fields['foo'].description is None
```
alexmojaki, line 270:
```suggestion
    class MyModel(BaseModel):
        foo: int = 123
        """Shouldn't be used"""

    assert MyModel.model_fields['foo'].description == "Shouldn't be used"

    MyModel = create_model(
        'MyModel',
        foo=(int, 123),
        __config__=ConfigDict(use_attribute_docstrings=True),
    )

    assert MyModel.model_fields['foo'].description is None
```

tests/test_docs_extraction.py
```diff
@@ -0,0 +1,287 @@
...
+            \"\"\"A docs\"\"\"
+
+            model_config = ConfigDict(use_attribute_docstrings=True)
+        """
+    )
+
+    locals_dict = {}
+
+    exec(source, globals(), locals_dict)
+    assert locals_dict['MyModelExec'].model_fields['a'].description is None
```
alexmojaki, line 287:
```suggestion
        class MyModel(BaseModel):
            a: int
            \"\"\"A docs\"\"\"

            model_config = ConfigDict(use_attribute_docstrings=True)
        """
    )

    locals_dict = {}

    exec(source, globals(), locals_dict)
    assert locals_dict['MyModel'].model_fields['a'].description is None
```

Viicos:
Considering how `inspect.findsource` has evolved across the different Python versions (that is the reason current tests are unstable depending on the version, and it might even be different in Python 3.13 with https://github.com/python/cpython/pull/106815): should I remove the fallback to the `inspect` method?

alexmojaki:
I see a test failure with dataclasses because the frame line is at the decorator instead of the class, which makes sense. I think that means that you can't separate getting the source from parsing it. You need to get the block starting at the current line, parse it, and see if you get a class definition with a matching name.

I say remove the inspect fallback for now (it'll be hard to keep it in combination with the above), see if all the tests pass, and then we can decide whether to put it back.

Viicos:
> I see a test failure with dataclasses because the frame line is at the decorator instead of the class, which makes sense. I think that means that you can't separate getting the source from parsing it. You need to get the block starting at the current line, parse it, and see if you get a class definition with a matching name.

After some debugging in 3.11/<3.11, it appears `frame.f_lineno` is indeed pointing at the decorator since 3.11, this probably comes from the new frame optimizations introduced in this version. Looking for a workaround.

alexmojaki:
> this probably comes from the new frame optimizations introduced in this version

Maybe it's because of that, but arguably the pre-3.11 behaviour of pointing at the class when the decorator is being executed is a bug that got fixed.

> Looking for a workaround.

That's what I described?

Viicos:
> That's what I described?

I meant implement the workaround you described 😅

alexmojaki:
The reason I suggested `ast` and not something like `line.strip().startswith('@')` is that you could have:

```python
@deco(
  ...
)
class ...
```

and if you're thinking about continuing to skip lines until you seem to reach a class definition, you might even have:

```python
@deco("""
   class ...
""")
class ...
```

> Unfortunately the fallback is used for `TypedDict`, which can only be used with `TypeAdapter`, and we can't walk back frames in this case.

Well that seems like a big problem since we don't trust `inspect`, I thought it was there as a theoretical last resort.

I wonder how useful this feature is for TypedDict. With this code:

```python
from typing import TypedDict


class A:
    foo: int
    "Foo doc"


class B(TypedDict):
    foo: int
    "Foo doc"


print(A().foo)

print(B(foo=1)["foo"])
```

My IDE (PyCharm) shows a doc when I hover over `A().foo` as expected, but there's nothing for `B` which is also not surprising. Unless it's different for other editors I don't see much reason for people to use attribute docstrings with TypedDict.

Viicos:
> The reason I suggested `ast` and not something like `line.strip().startswith('@')` is that you could have:

You're right my bad, I'll then switch to `ast`

> Unless it's different for other editors I don't see much reason for people to use attribute docstrings with TypedDict.

Well it is still beneficial for users to be able to use attribute docstrings instead of the following:

```python
from typing import Annotated, TypedDict

from pydantic import Field

class B(TypedDict):
    foo: Annotated[int, Field(description="Foo doc")]
```

With `BaseModel` and `pydantic.dataclass`, we "control" where the class is created so we can use frames successfully, but for `TypedDict`, it is either used with `TypeAdapter` or [as a field](https://github.com/pydantic/pydantic/pull/6563/files#diff-685ec8dd8bd3182be6a541d0af1d0ab37ab5abf3d80d9cb57f0e79efba8ac7a1R244-R255). But I agree this causes issues, such as the test failing under 3.7/3.8.

One warkaround would be to remove the workaround in the general case, and only use it if the call is coming from [`_typed_dict_schema`](https://github.com/pydantic/pydantic/blob/aa73451918a0a7f43c240c677ad4c9527a12e253/pydantic/_internal/_generate_schema.py#L1081). What do you think?

alexmojaki:
> One warkaround would be to remove the workaround in the general case, and only use it if the call is coming from [_typed_dict_schema](https://github.com/pydantic/pydantic/blob/aa73451918a0a7f43c240c677ad4c9527a12e253/pydantic/_internal/_generate_schema.py#L1081). What do you think?

I think that's OK. I was worried about the older review comments saying that the unreliability of `inspect` might be a dealbreaker, but after reading further I see that it's not so bad.

If we want to avoid `inspect` completely while still supporting TypedDict we could add an alternative API, e.g:

```python
@collect_docstrings
class B(TypedDict):
```

but I don't know if it's worth trying to do everything in this one PR.

alexmojaki:
I also think it's OK to remove support for TypedDict from this PR to keep the scope narrow and have something that can be merged with confidence.

pdm.lock
```diff
@@ -6,7 +6,7 @@ groups = ["default", "docs", "email", "linting", "memray", "mypy", "testing", "t
 cross_platform = true
 static_urls = false
 lock_version = "4.3"
-content_hash = "sha256:6f68a60ed5be207473a66fe2368d57b082d703e44f85c1b1767e625d81519dbb"
+content_hash = "sha256:712240a1a52356c85f693200f95828b353fe8adcc480fb443c6294f20b4dbde0"
```
alexmojaki, line 9: why has this changed?

Viicos: I'm not used to PDM, but one of the CI workflows was failing: https://github.com/pydantic/pydantic/actions/runs/6639854334/job/18039074733

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,117 @@
...
+    Args:
+        cls: The class of the Pydantic model to inspect.
+        use_inspect: Whether to skip usage of frames to find the object and use
+            the `inspect` module instead.
+
+    Returns:
+        A mapping containing attribute names and their corresponding docstring.
+    """
+    if use_inspect:
+        # Might not work as expected if two classes have the same name in the same source file.
```
alexmojaki, line 102: There needs to be some documentation about this for users, mentioning TypedDict and Python versions.

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,117 @@
...
+            pass
+        else:
+            if inspect.getmodule(frame) is inspect.getmodule(cls) and isinstance(lnum, int):
+                block_lines = inspect.getblock(lines[lnum - 1 :])
+                dedent_source = _dedent_source_lines(block_lines)  # type: ignore
+                # Because lnum can point to potential decorators before the class definition,
+                # we use ast to parse the block source:
+                try:
+                    block_tree = ast.parse(dedent_source)
+                except Exception:
```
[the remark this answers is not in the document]
Viicos: I had it happen, but it was before switching `if dedent_source.startswith('class'):` to `if dedent_source.startswith((' ', '\t')):`.

I can remove it, but I'm not sure we won't ever have any unexpected `IndentationError`s with this call

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,117 @@
...
+    while frame:
+        lnum = frame.f_lineno
+        try:
+            lines, _ = inspect.findsource(frame)
+        except OSError:
+            # Source can't be parsed (maybe because running in an interactive terminal),
+            # we don't want to error here.
+            pass
+        else:
+            if inspect.getmodule(frame) is inspect.getmodule(cls) and isinstance(lnum, int):
```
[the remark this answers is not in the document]
Viicos: According to typeshed:

```python
    # see discussion in #6769: f_lineno *can* sometimes be None,
    # but you should probably file a bug report with CPython if you encounter it being None in the wild.
    # An `int | None` annotation here causes too many false-positive errors.
    @property
    def f_lineno(self) -> int | Any: ...
```

pydantic/_internal/_docs_extraction.py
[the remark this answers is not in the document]
alexmojaki: Ok to keep it just in case, I was just wondering. But this should do:

```suggestion
                except SyntaxError:
```

tests/test_docs_extraction.py
```diff
@@ -0,0 +1,334 @@
...
+            use_attribute_docstrings=True,
+        )
+
+    assert MyModel.model_fields['a'].description == 'A docs'
+
+    class MyParameterizedModel(MyModel[int]):
+        a: int
+        """A parameterized docs"""
+
+    assert MyParameterizedModel.model_fields['a'].description == 'A parameterized docs'
```
Viicos, line 194: We hit an interesting behavior here. If you try to debug with a breakpoint L51 in `_docs_extraction.py`, you will see `_extract_source_from_frame` is called two times.

`BaseModel.__class_getitem__` will manually return a new class object, and will use `model_parametrized_name` to set the name of this new class. So when writing `class MyParameterizedModel(MyModel[int])`, a first class is created when `MyModel[int]` is executed (we can't find any class declaration because the model name is set to `MyModel[int]` by Pydantic,  so no docs are extracted). Then when `MyParameterizedModel(...)` is executed and docs are successfully extracted here.

alexmojaki: I was worried that the user overriding `model_parametrized_name` might be a problem, but it seems to be fine:

```python
import typing
from typing import Generic

T = typing.TypeVar("T")


class MyModel(BaseModel, Generic[T]):
    a: T
    """A docs"""

    model_config = ConfigDict(
        use_attribute_docstrings=True,
    )

    @classmethod
    def model_parametrized_name(cls, params):
        return "OtherName"


assert MyModel.model_fields["a"].description == "A docs"
assert MyModel[int].model_fields["a"].description == "A docs"  # maybe add this to the test
assert MyModel[int].__name__ == "OtherName"
assert MyModel.__name__ == "MyModel"


class MyParameterizedModel(MyModel[int]):
    a: int
    """A parameterized docs"""


assert MyParameterizedModel.model_fields["a"].description == "A parameterized docs"
assert MyParameterizedModel.__name__ == "MyParameterizedModel"
```

Viicos:
Thanks again @alexmojaki for the reviews and finding all the edge cases. I hope this will not be too hard to review; the implementation is a 100 lines but it can be tricky to follow. The tests take into account all the worst cases that could happen, so in the general case if people don't play with `exec` or similar it should be fine, and the solution with frames is more robust than `inspect.getsource`

tests/test_docs_extraction.py
```diff
@@ -0,0 +1,335 @@
...
+        b: str = '1'
+
+        """B docs"""
+
+    assert MyModel.__pydantic_fields__['a'].description is None
+    assert MyModel.__pydantic_fields__['b'].description is None
+
+
+def test_dataclass_docs_extraction():
+    @pydantic_dataclass(config=ConfigDict(use_attribute_docstrings=True))
```
alexmojaki, line 213:
```suggestion
    @pydantic_dataclass(
        config=ConfigDict(use_attribute_docstrings=True),
    )
```

so that we can feel justified about the more complicated implementation :laughing:

pydantic/_internal/_docs_extraction.py
```diff
@@ -0,0 +1,105 @@
...
+            lnum = frame.f_lineno
+            try:
+                lines, _ = inspect.findsource(frame)
+            except OSError:
+                # Source can't be parsed (maybe because running in an interactive terminal),
+                # we don't want to error here.
+                pass
+            else:
+                block_lines = inspect.getblock(lines[lnum - 1 :])
+                dedent_source = _dedent_source_lines(block_lines)  # type: ignore
```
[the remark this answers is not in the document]
Viicos: I'm waiting for pyright to release a new version that will include the typeshed changes so that I can remove the comments

alexmojaki (approved):
Great work! I suggest you resolve the merge conflicts and then re-request review from a maintainer.

Viicos:
Last merge commit fixes a couple typo in the newly added config values

alexmojaki:
@dmontagu ping

pdm.lock
```diff
@@ -1749,4 +1749,4 @@ summary = "Backport of pathlib-compatible object wrapper for zip files"
 files = [
     {file = "zipp-3.15.0-py3-none-any.whl", hash = "sha256:48904fc76a60e542af151aded95726c1a5c34ed43ab4134b597665c86d7ad556"},
     {file = "zipp-3.15.0.tar.gz", hash = "sha256:112929ad649da941c23de50f356a2b5570c954b65150642bccdd66bf194d224b"},
-]
+]
```
alexmojaki, line 1752:
```suggestion
]

```

Viicos:
I think I should probably make another PR to update `pyright` and fix the new issues

alexmojaki:
@dmontagu ping

Pwuts:
*bump*

Viicos:
There have been a lot of reviewing from @alexmojaki so I'm going to sum things up here so that it can be easier to review.

## Class definition parsing

This was first implemented in this commit: https://github.com/pydantic/pydantic/pull/6563/commits/29bbf3e13d45066475d6d560594a97aeadcfec50, inspired by  [danields761/pydantic-settings](https://github.com/danields761/pydantic-settings/). @Kludex then suggested using an `ast.NodeVisitor` instead, but @alexmojaki gave some arguments [here](https://github.com/pydantic/pydantic/pull/6563#discussion_r1369906061) to go back to the first solution. I'll let maintainers give input on which solution they want to keep.

## Retrieving the class source code

To retrieve the source code of a class so that it can be parsed, I was first using `inspect.getsource`, but I faced some issues described [here](https://github.com/pydantic/pydantic/pull/6563#discussion_r1262899854). After opening an issue on the cpython repo, class discovery got improved on Python 3.12, but can still be wrong in some cases.

Then @alexmojaki [suggested using stack frames](https://github.com/pydantic/pydantic/pull/6563#issuecomment-1769236460) to circumvent this issue. We got hit by a few issues going this way, in particular with decorated classes (depending on the Python version, the frame `f_lineno` would differ). We ended up parsing the current frame block source code as a temporary AST to see if it would match the correct class definition (see the implementation of `_extract_source_from_frame`).

I believe we added extensive testing with edge cases that will most probably never happen in real code. This is also gated behind a config flag, so it won't break existing code.

---

There has been some [recent discussion](https://discuss.python.org/t/36413) about [PEP 224 -- Attribute Docstrings](https://peps.python.org/pep-0224/), which would make this PR obsolete if implemented. However, it would only be available in the latest Python version, so I think this feature PR would still be relevant.

sydney-runkle:
@Viicos,

Sorry for all of the pings this week. Gearing up to do a final review + to merge this - could you please rebase when you have a chance? Thank you 🌟!

alexmojaki:
Thanks @Viicos!

Viicos:
Thanks for the reviews @alexmojaki, glad to see this one merged

pvcraven:
Should this work now?
```py


class MyTest(BaseModel):
    my_field_a: str = Field(description="Description 1")
    my_field_b: str
    """ Description 2 """


assert MyTest.model_fields["my_field_a"].description == "Description 1"
assert MyTest.model_fields["my_field_b"].description == "Description 2"
```
For me, the second assertion fails. Which I thought should work with this PR?

alexmojaki:
https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.use_attribute_docstrings

```python
class Model(BaseModel):
    model_config = ConfigDict(use_attribute_docstrings=True)
```

Issue #6534, opened 2023-07-08
Cannot get schema from model using string `$ref` as a field alias


Matthew-Burkard:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

Calling `model_json_schema`  raises an error if the model has a field with alias `$ref`.

### Example Code

```Python
from pydantic import BaseModel, Field


class MyModel(BaseModel):
    my_field: str = Field(alias="$ref")


print(MyModel.model_json_schema())
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0.2
        pydantic-core version: 2.1.2 release build profile
                 install path: /home/matthew/.cache/pypoetry/virtualenvs/sandbox-pwkskCd0-py3.9/lib/python3.9/site-packages/pydantic
               python version: 3.9.15 (main, Nov  6 2022, 17:01:32)  [GCC 11.3.0]
                     platform: Linux-6.2.6-76060206-generic-x86_64-with-glibc2.35
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @davidhewitt

Kludex:
@Matthew-Burkard Thanks for reporting it. Yep, it's a bug.

PR #6568, opened 2023-07-10
Make it work to use `$ref` as an alias

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6534

Note that while there are many references to the `'$ref'` key in the codebase that might otherwise appear to be subject to this issue, in most cases, there are not free-form user-specified keys in the values being handled. The exception to this are functions that are called recursively on dicts that happen to be JSON schemas. In these cases, I believe that is safe to assume that if the key is `'$ref'` _and_ the value has type `str` (as opposed to `dict` etc.) then that should be a _real_ JSON schema ref, as opposed to an incidental occurrence of the key `'$ref'`. This PR just starts checking the type of the value for the key `'$ref'` in those recursive functions.

skip change file check

Selected Reviewer: @lig


dmontagu:
please review

PR #6570, opened 2023-07-10
Handle function validators in a discriminated union

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6525

skip change file check


adriangb:
Sorry for never linking it to the original issue: https://github.com/pydantic/pydantic/pull/6541

tests/test_discriminated_union.py
```diff
@@ -1232,3 +1232,28 @@ class TestModel(BaseModel):
...
+        def replace_name(cls, v):
+            return 'dog'
+
+    AllowedAnimal = Annotated[Union[CatModel, DogModel], Field(discriminator='name')]
+
+    class Model(BaseModel):
+        x: AllowedAnimal
+
+    m = Model(x={'name': 'kitty'})
+    assert m.x.name == 'cat'
```
adriangb, line 1259: Might be worth confirming a discriminated union was used, either via an error case or a json schema (or both)

pydantic/_internal/_discriminated_union.py
```diff
@@ -430,6 +430,18 @@ def _infer_discriminator_values_for_inner_schema(
             # This will happen if the field has a default value; we ignore it while extracting the discriminator values
             return self._infer_discriminator_values_for_inner_schema(schema['schema'], source)
 
+        elif schema['type'] == 'function-after':
+            # After validators don't affect the discriminator values
+            return self._infer_discriminator_values_for_inner_schema(schema['schema'], source)
+
+        elif schema['type'] in {'function-before', 'function-wrap', 'function-plain'}:
```
adriangb, line 437: Not sure how you got around typing here but might be worth using `_core_utils.is_function_with_inner_schema(`

adriangb (approved):

pydantic/_internal/_discriminated_union.py
dmontagu: Note that it's not actually equivalent to that function call. We could do something like that, but this works because I'm not actually using any fields from inside `schema` to generate this error, so I don't need to do any type narrowing of `schema`.

The reason I did this was because each of those types of validator could affect what goes in, whereas `function-after` can't, so it's safe to rely on the input discriminator to select the union case to use for type-checking.

tests/test_discriminated_union.py
dmontagu: I was relying on the docs tests for that but I'll add it here too

tests/test_discriminated_union.py
```diff
@@ -1257,3 +1257,17 @@ class Model(BaseModel):
 
     m = Model(x={'name': 'kitty'})
     assert m.x.name == 'cat'
+
+    # Ensure a discriminated union is actually being used during validation
+    with pytest.raises(ValidationError) as exc_info:
+        Model(x={'name': 'invalid'})
+    assert exc_info.value.errors(include_url=False) == [
+        {
+            'ctx': {'discriminator': "'name'", 'expected_tags': "'kitty', 'cat', 'puppy', 'dog'", 'tag': 'invalid'},
```
adriangb, line 1266: These should be a list, it’s a 🐛

PR #6595, opened 2023-07-11
♻️ Use different error message on v1 redirects

Kludex:
Selected Reviewer: @hramezani


Kludex:
please review

hramezani:
Do we have a test for it? Or, Don't we need a test for it?

hramezani (approved):

Issue #6548, opened 2023-07-10
Using `None` as generic type errors in V2.


zakstucke:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

As stated, `type(None)` works as a generic, but `None` itself has stopped working, which I believe is a definite bug as mypy goes haywire when using `type(None)`.

```
Traceback (most recent call last):
  File "mypy_playground.py", line 12, in <module>
    Container[None](value=None)  # <-- errors in V2, works in V1
    ~~~~~~~~~^^^^^^
  File "venv/lib/python3.11/site-packages/pydantic/main.py", line 636, in __class_getitem__
    submodel = _generics.create_generic_submodel(model_name, origin, args, params)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/pydantic/_internal/_generics.py", line 142, in create_generic_submodel
    created_model = meta(
                    ^^^^^
  File "venv/lib/python3.11/site-packages/pydantic/_internal/_model_construction.py", line 172, in __new__
    complete_model_class(
  File "venv/lib/python3.11/site-packages/pydantic/_internal/_model_construction.py", line 420, in complete_model_class
    schema = cls.__get_pydantic_core_schema__(cls, handler)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/pydantic/main.py", line 533, in __get_pydantic_core_schema__
    return __handler(__source)
           ^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/pydantic/_internal/_schema_generation_shared.py", line 82, in __call__
    schema = self._handler(__source_type)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py", line 266, in generate_schema
    return self._generate_schema_for_type(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py", line 287, in _generate_schema_for_type
    schema = self._generate_schema(obj)
[...]

PR #6609, opened 2023-07-11
Make it work to use None as a generic parameter

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6548

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -702,8 +702,6 @@ class GenerateSchema:
         )
 
     def _common_field_schema(self, name: str, field_info: FieldInfo, decorators: DecoratorInfos) -> _CommonField:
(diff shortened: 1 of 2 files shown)


pydantic/_internal/_generate_schema.py
```diff
@@ -702,8 +702,6 @@ def _generate_dc_field_schema(
         )
 
     def _common_field_schema(self, name: str, field_info: FieldInfo, decorators: DecoratorInfos) -> _CommonField:
-        assert field_info.annotation is not None, 'field_info.annotation should not be None when generating a schema'
```
dmontagu, line 705: I thought this assertion was safe and that we couldn't hit this branch, but apparently we need to handle this case. Easy enough fix.

adriangb: Would that be an unreasonable breaking change? It's to fix a bug.

adriangb (approved):

PR #6633, opened 2023-07-12
Produce more accurate signatures for pydantic dataclasses

### LINKED ISSUE #6623: [PYD-179] Pydantic dataclasses have under-specified signature
This is used in FastAPI, and causes problems with using dataclasses in endpoints. Here's a fairly minimal reproduction showing that it's misbehaving in v2:

```
import inspect
from typing import Union
import pydantic
from pydantic.dataclasses import dataclass as pydantic_dc_v2
from pydantic.v1.dataclasses import dataclass as pydantic_dc_v1


@dataclass()
class CommonQueryParams:
    q: Union[str, None] = None
    skip: int = 0
    limit: int = 100


@pydantic_dc_v1()
class CommonQueryParamsV1:
    q: Union[str, None] = None
    skip: int = 0
    limit: int = 100


@pydantic_dc_v2()
class CommonQueryParamsV2:
    q: Union[str, None] = None
    skip: int = pydantic.Field(default=0)
    limit: int = 100


print(inspect.signature(CommonQueryParams))
#> (q: Optional[str] = None, skip: int = 0, limit: int = 100) -> None
print(inspect.signature(CommonQueryParamsV1))
#> (q: Optional[str] = None, skip: int = 0, limit: int = 100) -> None
print(inspect.signature(CommonQueryParamsV2))
#> (*args: 'Any', **kwargs: 'Any') -> 'None'
```
The main challenge is making sure that we get the right signature when there are `= pydantic.Field(...)`

<sub>[PYD-179](https://linear.app/pydantic/issue/PYD-179/pydantic-dataclasses-have-under-specified-signature)</sub>

dmontagu:
Closes #6623

This is necessary for FastAPI to to produce a collection of query param dependencies from a pydantic dataclass:

```python
from typing import Union

from fastapi import FastAPI, Depends

import pydantic
from pydantic.dataclasses import dataclass as pydantic_dataclass

app = FastAPI()


@pydantic_dataclass()
class QueryParams:
    q: Union[str, None] = None
    skip: int = pydantic.Field(default=0)
    limit: int = 100


@app.get("/")
def home(query_params: QueryParams = Depends()):
    return query_params
```

### DIFF
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -185,6 +185,7 @@ def dataclass(
 
         cls = dataclasses.dataclass(  # type: ignore[call-overload]
             cls,
+            # the value of init here doesn't affect anything except that it makes it easier to generate a signature
             init=True,
             repr=repr,
             eq=eq,
--- a/pydantic/dataclasses.py
+++ b/pydantic/dataclasses.py
@@ -185,7 +185,7 @@ def dataclass(
 
         cls = dataclasses.dataclass(  # type: ignore[call-overload]
             cls,
-            init=init,
+            init=True,
             repr=repr,
             eq=eq,
             order=order,
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -2,12 +2,14 @@
 from __future__ import annotations as _annotations
 
 import dataclasses
+import inspect
 import typing
 import warnings
 from functools import partial, wraps
+from inspect import Parameter, Signature, signature
 from typing import Any, Callable, ClassVar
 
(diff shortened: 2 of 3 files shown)


pydantic/dataclasses.py
```diff
@@ -185,7 +185,7 @@ def create_dataclass(cls: type[Any]) -> type[PydanticDataclass]:
 
         cls = dataclasses.dataclass(  # type: ignore[call-overload]
             cls,
-            init=init,
+            init=True,
```
Kludex, line 188: Should we add a note about this here as well?

Kludex (approved):
It works as expected with FastAPI 👍

pydantic/dataclasses.py
dmontagu: I added a comment just above it

Issue #6647, opened 2023-07-13
`anyOf` causing problems with Swagger/ReDoc on primitive types


Kludex:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

If we have an object, and a primite type, we are still creating an `anyOf`. That unfortunately doesn't work well with neither ReDoc nor Swagger.

From the code below, the following OpenAPI spec is generated:

<details><summary>Details</summary>
<p>

```json
{
    "components": {
        "schemas": {
            "HTTPValidationError": {
                "properties": {
                    "detail": {
                        "items": {
                            "$ref": "#/components/schemas/ValidationError"
                        },
                        "title": "Detail",
                        "type": "array"
                    }
                },
                "title": "HTTPValidationError",
                "type": "object"
            },
            "ValidationError": {
                "properties": {
                    "loc": {
                        "items": {
                            "type": ["string", "integer"]
                        },
                        "title": "Location",
                        "type": "array"
                    },
                    "msg": {
                        "title": "Message",
                        "type": "string"
                    },
                    "type": {
                        "title": "Error Type",
                        "type": "string"
                    }
                },
                "required": [
                    "loc",
                    "msg",
                    "type"
                ],
                "title": "ValidationError",
                "type": "object"
            }
        }
    },
    "info": {
        "title": "FastAPI",
        "version": "0.1.0"
    },
    "openapi": "3.1.0",
    "paths": {
        "/": {
            "get": {
                "operationId": "root__get",
                "parameters": [
                    {
                        "description": "Date to query",
                        "in": "query",
                        "name": "at",
                        "required": false,
                        "schema": {
                            "anyOf": [
                                {
                                    "enum": [
                                        "today",
                                        "tomorrow",
                                        "yesterday"
                                    ],
                                    "type": "string"
                                },
                                {
                                    "type": "null"
                                }
                            ],
                            "description": "Date to query",
                            "title": "At"
                        }
                    }
                ],
                "responses": {
                    "200": {
                        "content": {
                            "application/json": {
                                "schema": {}
                            }
                        },
                        "description": "Successful Response"
                    },
                    "422": {
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/HTTPValidationError"
                                }
                            }
                        },
                        "description": "Validation Error"
                    }
                },
                "summary": "Root"
            }
        }
    }
}
```

</p>
</details> 

Ideally, we'd have the following schema:

[...]

PR #6653, opened 2023-07-13
✨ Add `SkipJsonSchema` annotation

Kludex:
This PR solves #6647 on Pydantic side, in the sense that it allows the user to not generate the JSON schema on a type.

Although this is nice to have on Pydantic, it's not ideal for FastAPI users to do `T | SkipJsonSchema[None]`, mainly because of dev experience. That said, @dmontagu proposed that FastAPI would use a `FastAPIGeneratorJsonSchema` on their side which would skip nullable fields _if is not a `Body`_. - I'm going to push a PR in FastAPI to help on this.

Selected Reviewer: @lig

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -2085,7 +2085,7 @@ def models_json_schema(
     description: str | None = None,
     ref_template: str = DEFAULT_REF_TEMPLATE,
     schema_generator: type[GenerateJsonSchema] = GenerateJsonSchema,
-) -> tuple[dict[tuple[type[BaseModel] | type[PydanticDataclass], JsonSchemaMode], JsonSchemaValue,], JsonSchemaValue]:
+) -> tuple[dict[tuple[type[BaseModel] | type[PydanticDataclass], JsonSchemaMode], JsonSchemaValue], JsonSchemaValue]:
     """Utility function to generate a JSON Schema for multiple models.
 
     Args:
@@ -2245,7 +2245,7 @@ if TYPE_CHECKING:
     SkipJsonSchema = Annotated[AnyType, ...]
 else:
 
-    @_internal_dataclass.slots_dataclass
+    @dataclasses.dataclass(**_internal_dataclass.slots_true)
     class SkipJsonSchema:
         """Add this as an annotation on a field to skip generating a JSON schema for that field.
 
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -30,7 +30,7 @@ from typing import (
 import pydantic_core
 from pydantic_core import CoreConfig, CoreSchema, PydanticOmit, core_schema, to_jsonable_python
 from pydantic_core.core_schema import ComputedField
-from typing_extensions import Literal, assert_never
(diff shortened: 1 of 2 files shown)


Kludex:
please review

commonism:
I had a look and I think this will cause problems besides 

> [It is RECOMMENDED that a default value be valid against the associated schema.](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-default)

It creates a difference in the serialization done and communicated (to the consumers of the json document).

The test only work due to the use of None as default value
https://github.com/pydantic/pydantic/blob/070546862c1a693be272bd91c50de38b080436fd/tests/test_json_schema.py#L4840-L4843

https://github.com/pydantic/pydantic/blob/070546862c1a693be272bd91c50de38b080436fd/tests/test_json_schema.py#L4848-L4852

For consumers of the json description document, the resulting difference may be a real problem to overcome, as the description document does not match the protocol.

```python
    class A(BaseModel):
        x: int | SkipJsonSchema[None] = None

    class B(BaseModel):
        x: int | SkipJsonSchema[None] = 5

    class C(BaseModel):
        x: int | SkipJsonSchema[None]
```
For **consumers** of the json description document:
  * A(x=None) is invalid
    *  `{"x":None}` is invalid
  * A() will result in x being None anyway.
  * for B it is not possible to set x to None, neither via assignment or constructor due to type mismatch
    *  `{"x":None}` is invalid
  * for C it is not possible to set x to None
      *  `{"x":None}` is invalid

The proposal does not improve anything to the consumers of the description document created by pydantic, but makes the document even incompatible to the datamodel used by pydantic itself.
This degradation of the correctness of the description document is not limited to the consumers which fail at the v3.1 features used by pydantic, but even for clients which can do v3.1 perfectly, all consumers.

I'd prefer to have this compatibility gap bridged otherwise - using type:[…, none], as it does not harm the correctness of the description document used by consumers.
That said - I consider even just waiting for the ecosystem to catch up better.

Kludex:
please review

dmontagu (dismissed):

dmontagu:
I'm going to approve and merge this, because with https://github.com/pydantic/pydantic/pull/6798 you'll be able to do:
```python
from typing import Annotated

from pydantic.json_schema import SkipJsonSchema


class Model(BaseModel):
    x: Annotated[int | SkipJsonSchema[None], Field(json_schema_extra=lambda x: x.pop('default'))] = None

print(Model.model_json_schema())
#> {'properties': {'x': {'title': 'X', 'type': 'integer'}}, 'title': 'Model', 'type': 'object'}
```
which gives a way to have a reusable annotation that removes the default, which I hope gives you a way to address your point.

That said, I think everyone currently agrees the better solution to https://github.com/pydantic/pydantic/issues/6647 for FastAPI is on the FastAPI side, with the custom GenerateJsonSchema for non-body parameters that handles optional schemas differently. (I believe @Kludex has opened a PR for that.)

@commonism if you still have issues with this decision, please create a new issue about it and @-mention me, and we can try to make other improvements.

dmontagu (approved):

Issue #6426, opened 2023-07-04
not compatible with python 3.7.1


hampusek:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2 installed directly from the `main` branch, or equivalent

### Description

Using pydantic `BaseModel` or `dataclass` in python 3.7.1 causes the following error:

```
AttributeError: module 'typing' has no attribute 'OrderedDict'
```

complete error log:

```
Traceback (most recent call last):
  File "t.py", line 5, in <module>
    class Foo(BaseModel):
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_model_construction.py", line 177, in __new__
    types_namespace=types_namespace,
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_model_construction.py", line 420, in complete_model_class
    schema = cls.__get_pydantic_core_schema__(cls, handler)
  File "/usr/local/lib/python3.7/site-packages/pydantic/main.py", line 533, in __get_pydantic_core_schema__
    return __handler(__source)
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_schema_generation_shared.py", line 82, in __call__
    schema = self._handler(__source_type)
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_generate_schema.py", line 265, in generate_schema
    obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_generate_schema.py", line 277, in _generate_schema_for_type
    schema = self._generate_schema_from_prepare_annotations(obj)
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_generate_schema.py", line 383, in _generate_schema_from_prepare_annotations
    new_obj, new_annotations = self._prepare_annotations(obj, [])
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_generate_schema.py", line 1210, in _prepare_annotations
    res = self._get_prepare_pydantic_annotations_for_known_type(source_type, annotations)
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_generate_schema.py", line 1186, in _get_prepare_pydantic_annotations_for_known_type
    from ._std_types_schema import PREPARE_METHODS
  File "/usr/local/lib/python3.7/site-packages/pydantic/_internal/_std_types_schema.py", line 608, in <module>
[...]

PR #6664, opened 2023-07-13
Use OrderedDict from typing_extensions

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6426

Note that this is the relevant code in `typing_extensions`:
```python
# 3.7.2+
if hasattr(typing, 'OrderedDict'):
    OrderedDict = typing.OrderedDict
# 3.7.0-3.7.2
else:
    OrderedDict = typing._alias(collections.OrderedDict, (KT, VT))
```

Given that, I think this is safe to merge. @Kludex I understand where you are coming from with not fixing it but given its so simple I think it's worth just doing it.

Selected Reviewer: @adriangb


dmontagu:
Please review

adriangb (approved):

PR #6665, opened 2023-07-13
Prepare for v2.0.3 release

adriangb:


dmontagu (approved):

dosisod:
Just a head's up @adriangb this is commit is incorrectly tagged as `v2.03` instead of `v2.0.3` (notice the missing the dot). Dunno if this would cause any issues, but it might be good to re-tag it just in case.

Thanks!

adriangb:
Yup thank you!

PR #6679, opened 2023-07-14
Remove the Validate always section in validator docs

adriangb:
Selected Reviewer: @lig

### DIFF
--- a/docs/usage/validators.md
+++ b/docs/usage/validators.md
@@ -321,37 +321,6 @@ print(Model(a='invalid').a)
 #> 2000-01-01 00:00:00
 ```
 
-## Validate always
-
-**TODO this content is wrong!**
-
-For performance reasons, by default validators are not called for fields when a value is not supplied.
-However there are situations where it may be useful or required to always call the validator, e.g.
-to set a dynamic default value.
-
-```py
-from datetime import datetime
-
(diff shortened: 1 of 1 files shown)


adriangb:
please review

docs/usage/validators.md
```diff
@@ -321,37 +321,6 @@ print(Model(a='invalid').a)
 #> 2000-01-01 00:00:00
 ```
 
-## Validate always
-
-**TODO this content is wrong!**
-
-For performance reasons, by default validators are not called for fields when a value is not supplied.
-However there are situations where it may be useful or required to always call the validator, e.g.
-to set a dynamic default value.
```
ybressler, line 330: I would argue it is important to keep the section, but describe behavior instead. You can use `validate_defaults` to perform validation always...

adriangb: The reality of the matter is that (1) these docs as they currently are do not help and (2) many of the requests we get for this feature either are better solved in another manner or have big drawbacks. We can re-add docs when we find a simple example that this is the right fit for, once we get actual users requesting this that can only be solved by this option, etc.

ybressler: Got it. Valid.

lig (approved):
👌🏻 🤷🏻‍♂️

Issue #6672, opened 2023-07-14
Possible memory leak (again) in generic models after trying V2


MarkusSintonen:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

After taking Pydantic V2 into use some of our tests started to fail that try to ensure there are no leaks related to Pydantic generic models. Below is an example test that succeeds in Pydantic V1 but fails in V2 which might indicate memory leak issues as the models are not necessarily collected properly. Test checks the GC behaviour of a partially parametrized generic model.

(In V1 there was a fix for a similar issue but seems the issue might have popped up again: https://github.com/pydantic/pydantic/pull/5052)

Is V2 safe to take into use because of a possible regression here in garbage collection of models?

Thank you!

### Example Code

```Python
import gc
from typing import Generic, TypeVar

try:
    # v2
    from pydantic._internal._generics import _GENERIC_TYPES_CACHE as cache
    from pydantic import BaseModel as PydanticModel
except ImportError:
    # v1
    from pydantic.generics import GenericModel as PydanticModel, _generic_types_cache as cache


T = TypeVar("T")
C = TypeVar("C")


def test_collects():
    assert len(cache) == 0

    def fn():
        class Inner(PydanticModel, Generic[T, C]):
            a: T
            b: C

        class Outer(PydanticModel, Generic[C]):
            c: Inner[int, C]

        klass = Outer[str]
        assert len(cache) > 0
        assert klass in cache.values()

    fn()

    gc.collect(0)
    gc.collect(1)
    gc.collect(2)

    # Fails in V2 but succeeds in V1!:
    # AssertionError: [<class 'tests.test.Inner[int, ~C]'>, <class 'tests.test.Inner[int, ~C]'>]
    # assert 2 == 0
    assert len(cache) == 0, [*cache.values()]
```

### Python, Pydantic & OS Version

```Text
For V1 (doesnt fail):

             pydantic version: 1.10.11
            pydantic compiled: True
                 install path: /Users/markussintonen/Library/Caches/pypoetry/virtualenvs/leak-test-VvUvdXxD-py3.11/lib/python3.11/site-packages/pydantic
               python version: 3.11.1 (main, Dec 23 2022, 09:40:27) [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.4.1-x86_64-i386-64bit
     optional deps. installed: ['typing-extensions']


For V2 (fails):
             pydantic version: 2.0.2
[...]

PR #6681, opened 2023-07-14
Use WeakValueDictionary to fix generic memory leak

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6672


adriangb (approved):
Seems like a reasonable trade off to me to xfail the pypy tests if we can fix memory leaks on all other Python versions

pydantic/_internal/_model_construction.py
```diff
@@ -535,3 +538,42 @@ def generate_model_signature(
...
+
+class _PydanticWeakValueDictionary(weakref.WeakValueDictionary):
+    pass
+
+
+def build_lenient_weakvaluedict(d: dict[str, Any] | None) -> dict[str, Any] | None:
+    """Takes an input dictionary, and produces a new value that (invertibly) replaces the values with weakrefs.
+
+    We can't just use a WeakValueDictionary because many types (including int, str, etc.) can't be stored as values
+    in a WeakValueDictionary.
```
MarkusSintonen, line 551: What is the actual issue here?:
```
weak_dict = WeakValueDictionary()
weak_dict["test1"] = str
weak_dict["test2"] = int
print([v for v in ddd.values()])  # [<class 'str'>, <class 'int'>]
```

pydantic/_internal/_model_construction.py
```diff
@@ -535,3 +538,42 @@ def generate_model_signature(
...
+
+    The `unpack_lenient_weakvaluedict` function can be used to reverse this operation.
+    """
+    if d is None:
+        return None
+    result = {}
+    for k, v in d.items():
+        try:
+            proxy = _PydanticWeakValueDictionary(ref=v)
```
MarkusSintonen, line 560: If there is an issue using WeakValueDictionary could this instead use `weakref.ref` for `proxy` instead of wrapping it to a whole `_PydanticWeakValueDictionary` with a single key?

MarkusSintonen: Something like this would also work by implementing MutableMapping (without overriding WeakValueDictionary and keeping `__getitem__` to return the actual value instead of weakrefs from `__pydantic_parent_namespace__`):
```python
class ParentNamespaceDictionary(MutableMapping[str, Any]):
    def __init__(self) -> None:
        self._data: dict[str, Union[weakref.ReferenceType[Any], Any]] = {}
        self._needs_remove_dead = False

    def __setitem__(self, key: str, value: Any) -> None:
        self._remove_dead()
        try:
            self._data[key] = weakref.ReferenceType(value, self._on_item_finalized)
        except TypeError:
            self._data[key] = value  # Eg str instances

    def __getitem__(self, key: str) -> Any:
        self._remove_dead()
        val = self._data[key]
        if not isinstance(val, weakref.ReferenceType):
            return val
        if (unwrapped := val()) is not None:
            return unwrapped
        raise KeyError(key)

    def __delitem__(self, key: str) -> None:
        self._remove_dead()
        del self._data[key]

    def __len__(self) -> int:
        self._remove_dead()
        return len(self._data)

    def __iter__(self) -> Iterator[str]:
        self._remove_dead()
        return iter(self._data)

    def _remove_dead(self) -> None:
        if self._needs_remove_dead:
            self._needs_remove_dead = False
            dead_keys = [k for k, v in self._data.items() if isinstance(v, weakref.ReferenceType) and v() is None]
            for k in dead_keys:
                del self._data[k]

    def _on_item_finalized(self, _ref: weakref.ReferenceType) -> None:
        self._needs_remove_dead = True
```

MarkusSintonen:
@dmontagu it seems the leak is still present when model also defines a `field_validator`. So try to add the same test but this time change the `Outer` to include a `field_validator`:
```
        class Outer(PydanticModel, Generic[C]):
            c: Inner[int, C]

            @field_validator("c", mode="before")
            @classmethod
            def _validator(cls, val: Any, info: FieldValidationInfo) -> Any:
                return val
```
Test also fails when using `model_validator`.
Weird...

dmontagu:
@MarkusSintonen it looks like the latest version of pydantic-core should have addressed your latest leak issue.

I'm wondering now if maybe that PR will render this "fix" unnecessary.. @davidhewitt is there any reason I shouldn't be able to use the `main` branch of pydantic-core with this branch and see if this memory leak has gone away?

MarkusSintonen:
>  it looks like the latest version of pydantic-core should have addressed the leak described in your previous comment — any chance you could confirm? (If that's annoying to do don't worry about it, I can make sure.)
> I'm wondering now if maybe that PR will render this "fix" unnecessary..

I used [commit](https://github.com/pydantic/pydantic/commit/a787642276b8cf4587ffdb94ff54a2dcf1fb27bd) that bumps `pydantic-core` to run some of our tests. For me it partially fixes the issue but not fully. So it seems both are needed, the fix now in master having latest pydantic-core and this one fixing issues with `__pydantic_parent_namespace__`. The issues to me seem two different sources of possible leaks.

I used both this branch and above master commit in a fork. Verified that also `__pydantic_parent_namespace__` is required.

Would recommend you on verifying with the unit tests that issue gets fixed. The tests in this branch should not pass in the linked master commit.

davidhewitt:
In principle you can use `pydantic-core` main by updating the pdm dependency here and running in CI, but I think in practice we need to tweak CI a bit to make that work effortlessly.

However in this case, no need, as the fix is already landed in `main` courtesy of `pydantic-core==2.3.1`.

pydantic/_internal/_model_construction.py
dmontagu: I see how the `ParentNamespaceDict` would work, but ultimately I think it's a bit more complexity than just replacing the WeakValuesDict with a custom `weakref.ReferenceType` subclass.

I'd be open to a PR making this change. And would be open to doing one myself if there was more of a clear benefit for it, probably in a future where we do more with the parent namespace dict. I'd rather not worry about getting this exactly right though unless that dict gets used more broadly than it is now.

If you disagree I could put in the effort to implement it this way.

dmontagu:
This is approved and I think ready to merge. @MarkusSintonen is there any reason not to merge this now? If not, I think it can get into today's release.

MarkusSintonen:
> This is approved and I think ready to merge. MarkusSintonen is there any reason not to merge this now? If not, I think it can get into today's release.

@dmontagu good to go by me, thanks!

pydantic/_internal/_model_construction.py
```diff
@@ -162,8 +163,10 @@ def wrapped_model_post_init(self: BaseModel, __context: Any) -> None:
                 obj.__set_name__(cls, name)
 
             if __pydantic_reset_parent_namespace__:
-                cls.__pydantic_parent_namespace__ = parent_frame_namespace()
+                cls.__pydantic_parent_namespace__ = build_lenient_weakvaluedict(parent_frame_namespace())
             parent_namespace = getattr(cls, '__pydantic_parent_namespace__', None)
+            if isinstance(parent_namespace, dict):
+                parent_namespace = unpack_lenient_weakvaluedict(parent_namespace)
```
MarkusSintonen, line 169: This check maybe redundant as there is already the None check inside `unpack_lenient_weakvaluedict `

samuelcolvin (approved):
LGTM

samuelcolvin:
thank you all for this.

PR #6695, opened 2023-07-15
Update validator docs

adriangb:
Goals:
- Update examples for v2
- Emphasize use of Annotated
- Simplify examples to focus on one thing / feature at a time
- Add sections on error handling, ordering of validators and the different modes
- Include better information on the expected signatures and uses for model validators
- Remove some redundant examples/sections

Selected Reviewer: @dmontagu


adriangb:
please review

adriangb:
@tpdorsey apologies just saw you had https://github.com/pydantic/pydantic/pull/6660, I was unaware of it. I think this is a more wholesale rework with valuable changes to the examples and such. Can you bring over the improvements you made there to this PR? Feel free to just push to this branch.

docs/usage/validators.md
```diff
@@ -1,375 +1,244 @@
...
-* A single validator can be applied to multiple fields by passing it multiple field names.
-* A single validator can also be called on *all* fields by passing the special value `'*'`.
-* The keyword argument `mode='before'` will cause the validator to be called prior to other validation.
-* Using validator annotations inside of `Annotated` allows applying validators to items of collections.
+```py test="skip" lint="skip" upgrade="skip"
+MyVal = Annotated[int, AfterValidator(func1), BeforeValidator(func2)]
+```
 
-## Special Types
+## Field validators
```
dmontagu, line 143: I feel like it might make more sense to start with this section than the annotated validators, since 1, this is closer to existing usage, and 2, it doesn't introduce the concept of using `Annotated`. I feel like it introduces just the concept of a validator function, which then motivates the idea of using `Annotated` validators as a way to make them reusable by way of a reused type annotation (rather than needing to set up the validator on every model you want to use it with).

I'm fine if you disagree, I just think this topic is quicker to understand and lowers the barrier to understand the annotated ones.

docs/usage/validators.md
```diff
@@ -1,375 +1,244 @@
...
-    normalize_name = field_validator('name')(normalize)
-
+A few things to note on validators:
 
-jane_doe = Producer(name='JaNe DOE')
-john_doe = Consumer(name='joHN dOe')
-assert jane_doe.name == 'Jane Doe'
-assert john_doe.name == 'John Doe'
-```
+* validators are "class methods", so the first argument value they receive is the `UserModel` class, not an instance of `UserModel`. We recommend you use the `@classmethod` decorator on them below the `@field_validator` decorator to get proper type checking.
```
dmontagu, line 218: I think it's maybe worth drawing the distinction between field_validator and model_validator, or somehow accounting for the fact that `@model_validator(mode='after')` can be used on instance methods. I'm not saying we need to say that here in full detail, but I think maybe we should phrase this point in a way that feels less in-conflict with that fact.

dmontagu: (One way to achieve that might be to replace "validators" with "`field_validator`s" in this text

docs/usage/validators.md
```diff
@@ -1,375 +1,244 @@
...
+    disables `assert` statements, and **validators will stop working**.
 
-!!! tip
-    If you have a lot of fields that you want to validate, it usually makes sense to
-    define a help function with which you will avoid setting `allow_reuse=True` over and
-    over again.
+If you want to access values from another field you can use `FieldValidationInfo.data`, which is a dict of field name to field value.
+Validation is done in the order fields are defined, so you have to be careful when using `FieldValidationInfo.data` to not access a field that has not yet been validated/populated.
+In the example above you would not be able to access `info.data['id']` from within `name_must_contain_space`.
+For a more flexible approach to this we recommend you use `@model_validator`, which is discussed in the section below.
```
dmontagu, line 233: I think it might be good to phrase this even more aggressively, something like:
```suggestion
If you want to access values from another field inside a `@field_validator`, this may be possible using `FieldValidationInfo.data`, which is a dict of field name to field value.
Validation is done in the order fields are defined, so you have to be careful when using `FieldValidationInfo.data` to not access a field that has not yet been validated/populated — 
in the code above, for example, you would not be able to access `info.data['id']` from within `name_must_contain_space`.
However, in most cases where you want to perform validation using multiple field values, it is better to use `@model_validator` which is discussed in the section below.
```
But I leave it up to you.

docs/usage/validators.md
```diff
@@ -419,15 +292,120 @@ except ValidationError as e:
...
+
+Before model validators are passed the raw input which is often a `dict[str, Any]` but could also be an instance of the model itself (e.g. if `UserModel.model_validate(UserModel.construct(...))` is called) or anything else since you can pass arbitrary objects into `model_validate`.
+Because of this before validators are extremely flexible and powerful but can be cumbersome and error prone to implement.
+Before model validators should be class methods.
+The first argument should be `cls` (and we also recommend you use `@classmethod` below `@model_validator` for proper type checking), the second argument will be the input (you should generally type it as `Any` and use `isinstance` to narrow the type) and the third argument (if present) will be a `pydantic.ValidationInfo`.
+
+After validators are instance methods and always receive an instance of the model as the first argument.
+You should not use `(cls, ModelType)` as the signature, instead just use `(self)` and let type checkers infer the type of `self` for you.
+Since these are fully type safe they are often easier to implement than before validators.
```
dmontagu, line 456:
```suggestion
Since these are fully type safe they are often easier to implement than `mode='before'` model_validators.
```
It's easy for us to read the docs and understand when we are using the words "before", "after", "validator" as referencing specific concepts, but I think it might be a bit harder for people needing to go through these docs, so I would suggest trying to make it more clear when it's a reference to a specific concept (e.g., `mode='before'`, or `@model_validator`) rather than using "normal" language, since I think it will cost new readers some cognitive overhead to parse 
> After validators are instance methods ...
as
> `mode='after'` validators are instance methods ...
rather than
> After the point in time that validators are instance methods ...

Obviously they won't stay confused for long, I just think it's a pattern we can try to follow to make this all that much easier to read.

docs/usage/validators.md
```diff
@@ -419,15 +292,120 @@ except ValidationError as e:
...
+If any field fails to validate after validators will not be called.
```
dmontagu, line 457: Another good example that I think could be hard for a newcomer to make sense of right away. If you don't immediately realize that "after validators" need to be grouped together as a concept, it is easy to read one word at a time and think that it's about to explain to me that something might happen before a field fails to validate that would affect something later, and then think "was there a typo/grammar error here?".

I think the following prevents that misread:
```suggestion
If any field fails to validate, `mode='after'` validators will not be called.
```

docs/usage/validators.md
```diff
@@ -419,15 +292,120 @@ except ValidationError as e:
...
+m = Model(names=['foo', 'bar'])
+print(m)
+#> names=['foo', 'bar']
+
+m = Model(names=['foo', 123])  # (1)!
+print(m)
+#> names=['foo', 123]
+```
+
+1. Note that the validation of the second item is skipped.
```
dmontagu, line 560: I'd maybe add a note that if the value has the wrong type due to skipped validation, a warning may be produced during serialization.

docs/usage/validators.md
```diff
@@ -419,15 +292,120 @@ except ValidationError as e:
...
+If any field fails to validate after validators will not be called.
+
+## Handling errors in validators
 
-Field validation will not occur if `pre=True` root validators raise an error. As with field validators,
-"post" (i.e. `pre=False`) root validators by default will be called even if prior validators fail; this
-behaviour can be changed by setting the `skip_on_failure=True` keyword argument to the validator.
-The `values` argument will be a dict containing the values which passed field validation and
-field defaults where applicable.
+As mentioned in the previous sections validators can raise either a `ValueError` or `AssertionError` (including ones generated by `assert ...` statements).
```
dmontagu, line 461:
```suggestion
As mentioned in the previous sections you can raise either a `ValueError` or `AssertionError` (including ones generated by `assert ...` statements) within a validator to indicate validation failed.
```

dmontagu (approved):
LGTM, though I have made various suggestions you can take or leave.

PR #6711, opened 2023-07-17
Add pydantic core uuid

JeanArhancet:
## Change Summary

Integration test for this MR : https://github.com/pydantic/pydantic-core/pull/772

## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -1630,6 +1630,17 @@ class GenerateJsonSchema:
         self.update_with_validations(json_schema, schema, self.ValidationsMapping.string)
         return json_schema
 
+    def uuid_schema(self, schema: core_schema.UuidSchema) -> JsonSchemaValue:
+        """Generates a JSON schema that matches a schema that defines a UUID.
+
+        Args:
+            schema: The core schema.
+
+        Returns:
+            The generated JSON schema.
+        """
+        return {'type': 'string', 'format': 'uuid'}
+
     def multi_host_url_schema(self, schema: core_schema.MultiHostUrlSchema) -> JsonSchemaValue:
         """Generates a JSON schema that matches a schema that defines a URL that can be used with multiple hosts.
 
@@ -1946,6 +1957,9 @@ class GenerateJsonSchema:
             'ge': 'minimum',
             'lt': 'exclusiveMaximum',
             'gt': 'exclusiveMinimum',
+        }
+        uuid = {
+            'version': 'version',
         }
 
     def get_flattened_anyof(self, schemas: list[JsonSchemaValue]) -> JsonSchemaValue:
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -542,16 +542,7 @@ class UuidVersion:
     def __get_pydantic_core_schema__(
         self, source: Any, handler: _annotated_handlers.GetCoreSchemaHandler
     ) -> core_schema.CoreSchema:
-        return core_schema.general_after_validator_function(
-            cast(core_schema.GeneralValidatorFunction, self.validate), handler(source)
-        )
-
-    def validate(self, value: UUID, _: core_schema.ValidationInfo) -> UUID:
-        if value.version != self.uuid_version:
-            raise PydanticCustomError(
-                'uuid_version', 'uuid version {required_version} expected', {'required_version': self.uuid_version}
-            )
-        return value
+        return core_schema.uuid_schema(version=self.uuid_version)
 
     def __hash__(self) -> int:
         return hash(type(self.uuid_version))
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -48,7 +48,7 @@ class MyModel(BaseModel):
 @pytest.mark.parametrize(
     'ser_type,gen_value,json_output',
     [
-        (UUID, lambda: 'ebcdab58-6eb8-46fb-a190-d07a33e9eac8', b'"ebcdab58-6eb8-46fb-a190-d07a33e9eac8"'),
+        (UUID, lambda: UUID('ebcdab58-6eb8-46fb-a190-d07a33e9eac8'), b'"ebcdab58-6eb8-46fb-a190-d07a33e9eac8"'),
         (IPv4Address, lambda: '192.168.0.1', b'"192.168.0.1"'),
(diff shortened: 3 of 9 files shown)


davidhewitt:
Naturally it fails to build 😂. I think it would be fixed by installing a rust toolchain, similar to this: https://github.com/pydantic/pydantic-core/pull/782

JeanArhancet:
@davidhewitt  Do you have an idea of the CI issue?

davidhewitt:
Not off the top of my head, I'll follow up on this (probably tomorrow at this point, I got pulled sideways onto #6727).

JeanArhancet:
I had to make a lot of changes but I don't think it's any good: 

- `UUID` -> `uuid` in error messages
- `class` -> `class_name` in error messages
-  Add a field `version` in metadata
- ...

Also, I've changed the UUID1...etc types

@davidhewitt WDYT?

pydantic/json_schema.py
```diff
@@ -1630,6 +1630,17 @@ def url_schema(self, schema: core_schema.UrlSchema) -> JsonSchemaValue:
...
+    def uuid_schema(self, schema: core_schema.UuidSchema) -> JsonSchemaValue:
+        """Generates a JSON schema that matches a schema that defines a UUID.
+
+        Args:
+            schema: The core schema.
+
+        Returns:
+            The generated JSON schema.
+        """
+        return {'type': 'string', 'format': 'uuid'}
```
davidhewitt, line 1642: I think we might want to have the uuid version in the format (where applicable)?

tests/test_json.py
```diff
@@ -48,7 +48,7 @@ class MyModel(BaseModel):
 @pytest.mark.parametrize(
     'ser_type,gen_value,json_output',
     [
-        (UUID, lambda: 'ebcdab58-6eb8-46fb-a190-d07a33e9eac8', b'"ebcdab58-6eb8-46fb-a190-d07a33e9eac8"'),
+        (UUID, lambda: UUID('ebcdab58-6eb8-46fb-a190-d07a33e9eac8'), b'"ebcdab58-6eb8-46fb-a190-d07a33e9eac8"'),
```
davidhewitt, line 51: Why was this change necessary?

pydantic/types.py
```diff
@@ -527,46 +528,36 @@ def condecimal(
...
+
+    version: int
 
     def __hash__(self) -> int:
-        return hash(type(self.uuid_version))
+        return hash(self.version)
 
 
+UuidVersion = Annotated[int, Version]
+UUIDStrict = Annotated[UUID, Strict()]
```
davidhewitt, line 555: Personally I don't think we need to add this, as users already have `Annotated[UUID, Strict()]` recommended as https://docs.pydantic.dev/latest/usage/strict_mode/#strict-mode-with-annotated-strict

If we added this, users would also want `UUID1Strict` etc presumably, and it all gets a bit too much then IMO.

pydantic/types.py
```diff
@@ -527,46 +528,36 @@ def condecimal(
...
+
+        UuidVersion = Annotated[int, Version]
+        ```
+    """
+
+    version: int
 
     def __hash__(self) -> int:
-        return hash(type(self.uuid_version))
+        return hash(self.version)
```
davidhewitt, line 551: cc @adriangb I would be grateful to hear your input here on what you is appropriate to do with this type. My instinct is against adding a generic `Version` type and keeping it just as `UuidVersion`, maybe there's a technical reason I've missed for doing this. We should be careful not to create a breaking change.

davidhewitt (commented):
Thanks, this is helpful to see. I've spattered a few questions on the diff here.

I also think we probably do not want to have the `uuid_exact_type` error, to keep things simpler. I've posted a corresponding suggestion on the `pydantic-core` PR.

I got diverted by memory leaks yesterday, I am now returning to the decimal validator so will figure out the CI issue here and the benchmark example ASAP.

tests/test_json.py
JeanArhancet: To match the new UUID serializer https://github.com/pydantic/pydantic-core/pull/772/files#diff-3e8029d3656b4c545330496a646d8b651884a649d1a7934fd7f32c50e33fe8a7R10-R20

Do you think it's a regression?

tests/test_types.py
```diff
@@ -1288,7 +1289,7 @@ def __bool__(self) -> bool:
         ('float_check', b't', ValidationError),
         ('uuid_check', 'ebcdab58-6eb8-46fb-a190-d07a33e9eac8', UUID('ebcdab58-6eb8-46fb-a190-d07a33e9eac8')),
         ('uuid_check', UUID('ebcdab58-6eb8-46fb-a190-d07a33e9eac8'), UUID('ebcdab58-6eb8-46fb-a190-d07a33e9eac8')),
-        ('uuid_check', b'ebcdab58-6eb8-46fb-a190-d07a33e9eac8', UUID('ebcdab58-6eb8-46fb-a190-d07a33e9eac8')),
+        ('uuid_check', b'ebcdab58-6eb8-46fb-a190-d07a33e9eac8', ValidationError),
```
JeanArhancet, line 1292: If I run a python script to test, I get an error. So I think it's normal to have a `ValidatorError`

```print(str(UUID(b'a6cc5730-2261-11ee-9c43-2eb5a363657c')))``` -> `TypeError: a bytes-like object is required, not 'str'`

**Comment by JeanArhancet** [comment]
@davidhewitt I went back on the changes

File: pydantic/types.py (hunk 3 above)
Comment (JeanArhancet) [reply pydantic/types.py]: You're right, thanks

File: pydantic/json_schema.py (hunk 1 above)
Comment (JeanArhancet) [reply pydantic/json_schema.py:1642]: I think we deduplicate the information because if we have a uuid4 the json is : ` {'type': 'string', 'format': 'uuid4'}`

**Comment by JeanArhancet** [comment]
Sorry to ask again @davidhewitt 

But I seem to have a regression here: 

```
│  tests/test_types.py  │  test_union_subclass[10]    │  4706           │  4715        │  AssertionError  │
│  tests/test_types.py  │  test_union_subclass[None]  │  4706           │  4715        │  AssertionError  │
```

I don't understand why. Do you have any idea ?

JeanArhancet:
superseded by #6831

Issue #6663, opened 2023-07-13
`Json[type]` schema generated now returns `string` rather than `type`


zakstucke:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

With `Json[]` the schema type is now always `string`, rather than the type inside `Json[]`.

In `V1` the example would produce:

```JSON
{
   "title":"Foo",
   "type":"object",
   "properties":{
      "foo":{
         "title":"Foo",
         "type":"integer"
      }
   },
   "required":[
      "foo"
   ]
}
```

In `V2` it produces:
```JSON
{
   "properties":{
      "foo":{
         "format":"json-string",
         "title":"Foo",
         "type":"string"
      }
   },
   "required":[
      "foo"
   ],
   "title":"Foo",
   "type":"object"
}
```

I may be wrong, but given `"format": "json-string"` is added for clarity, there's no reason to change the type to `"string"` and this is a bug?

If this is intended and shouldn't be changed, is there any easy way to opt out without overriding `__get_pydantic_json_schema__` everywhere? The use case is most of our endpoints come in as json strings which are decoded, so this currently breaks a lot of the schema and validation externally.

### Example Code

```Python
from pydantic import BaseModel, Json


class Foo(BaseModel):
    foo: Json[int]


Foo.schema()
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0.2
        pydantic-core version: 2.1.2 release build profile
                 install path: venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.4 (main, Jun  7 2023, 12:45:48) [GCC 11.3.0]
                     platform: Linux-5.19.0-46-generic-x86_64-with-glibc2.35
     optional deps. installed: ['devtools', 'email-validator', 'typing-extensions']
```

Selected Assignee: @samuelcolvin

hramezani:
Thanks @zakstucke for reporting this 🙏 

Isn't the `"type":"string"` correct? because the model accepts strings, not int.

```py
Foo(foo='1')  # Ok

Foo(foo=1)
"""
pydantic_core._pydantic_core.ValidationError: 1 validation error for Foo
foo
  JSON input should be string, bytes or bytearray [type=json_type, input_value=1, input_type=int]
    For further information visit https://errors.pydantic.dev/2.0.3/v/json_type
"""
```

zakstucke:
@hramezani I would say no given a few reasons:
[...]

PR #6715, opened 2023-07-17
Use contentSchema keyword for JSON schema

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6663

Selected Reviewer: @lig

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -1609,7 +1609,7 @@ class GenerateJsonSchema:
         Returns:
             The generated JSON schema.
         """
-        content_core_schema = schema['schema'] if 'schema' in schema else core_schema.any_schema()
+        content_core_schema = schema.get('schema') or core_schema.any_schema()
         content_json_schema = self.generate_inner(content_core_schema)
         if self.mode == 'validation':
             return {'type': 'string', 'contentMediaType': 'application/json', 'contentSchema': content_json_schema}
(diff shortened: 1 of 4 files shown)


dmontagu:
please review

pydantic/json_schema.py
```diff
@@ -1609,13 +1609,13 @@ def json_schema(self, schema: core_schema.JsonSchema) -> JsonSchemaValue:
         Returns:
             The generated JSON schema.
         """
+        content_core_schema = schema['schema'] if 'schema' in schema else core_schema.any_schema()
```
lig, line 1612: I would like the following code more but it will always call `core_schema.any_schema()`

```suggestion
        content_core_schema = schema.get('schema', core_schema.any_schema())
```

It would be nice to have a constant like `core_schema.DEFAULT_SCHEMA` that will hold this value and allow the above to be as efficient as `if..in`.

lig (approved):
👍🏻

pydantic/json_schema.py
dmontagu, line 1612:
```suggestion
    content_core_schema = schema.get('schema') or core_schema.any_schema()
```
I think this is the "better" version of your suggestion, which works since all schemas must have a type key so `schema.get('schema')` will be falsy only if it is `None` (and having the value None is not meaningfully different from not being present)

PR #6720, opened 2023-07-17
Fix recursion error in json schema generation

### LINKED ISSUE #6708: Leak in `pydantic_js_functions` leading to `RecursionError`s with V2 generic model json schema handling
### Initial Checks

### Description

There seems to be a leak related to items in `pydantic_js_functions` which also leads to an `RecursionError` with `model_json_schema` handling. Seems `pydantic_js_functions` is getting lot of entries to the list with certain types of generic models. Below I tried to isolate the problem as best as possible but there might be even a simpler example of the failure.

### Example Code

```Python
from typing import Generic, Optional, TypeVar

try:
    # v2
    from pydantic._internal._generics import _GENERIC_TYPES_CACHE
    from pydantic import BaseModel as PydanticModel
    GenericModel = BaseModel
except ImportError:
    # v1
    from pydantic.generics import GenericModel as GenericModel


T = TypeVar("T")


def test_leak():
    class Inner(BaseModel):
        a: int

    class Outer(GenericModel, Generic[T]):
        b: Optional[T]  # Works when not Optional...

    for i in range(1000):
        class ModelTest(BaseModel):
            c: Outer[Inner]

        if core_schema := getattr(ModelTest, "__pydantic_core_schema__", None):
            # v2
            # pydantic_js_functions on "Inner" keeps on growing at each iteration: 5, 6, 7, 8, 9, 10...
            inner = next(d for d in core_schema["definitions"] if d["cls"].__name__ == "Inner")
            inner_pydantic_js_functions = inner["metadata"]["pydantic_js_functions"]
            print(f"Run {i}: len(pydantic_js_functions)={len(inner_pydantic_js_functions)}")
            # ... and then RecursionError at "Run 228: len(pydantic_js_functions)=232"
            ModelTest.model_json_schema()
        else:
            # v1
            # It works without issues
            ModelTest.schema()
```

```Text
pydantic version: 2.0.3
                 install path: /Users/markussintonen/Library/Caches/pypoetry/virtualenvs/leak-test-VvUvdXxD-py3.11/lib/python3.11/site-packages/pydantic
               python version: 3.11.1 (main, Dec 23 2022, 09:40:27) [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.4.1-x86_64-i386-64bit
```

Selected Assignee: @dmontagu

adriangb:
Fixes #6708

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -281,6 +281,17 @@ class GenerateSchema:
             obj, from_dunder_get_core_schema=from_dunder_get_core_schema, from_prepare_args=from_prepare_args
         )
 
+    def _add_js_function(self, metadata_schema: CoreSchema, js_function: Callable[..., Any]) -> None:
+        metadata = CoreMetadataHandler(metadata_schema).metadata
+        pydantic_js_functions = metadata.setdefault('pydantic_js_functions', [])
+        # because of how we generate core schemas for nested generic models
+        # we can end up adding `BaseModel.__get_pydantic_json_schema__` multiple times
+        # this check may fail to catch duplicates if the function is a `functools.partial`
+        # or something like that
+        # but if it does it'll fail by inserting the duplicate
+        if js_function not in pydantic_js_functions:
+            pydantic_js_functions.append(js_function)
+
     def _generate_schema_for_type(
         self,
         obj: Any,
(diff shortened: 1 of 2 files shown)


adriangb:
@commonism any concerns with this?

Kludex (approved):

dmontagu (commented):
I think this makes sense, though @adriangb and I both agree that there is probably a better fix out there.

Probably the best way to fix it would be to generate the JSON schema more fully-independently of the core schema, rather than storing JSON-modification functions on the core schema. Or at least rework the core schema generation so that there is less mutability. But this works now.

MarkusSintonen:
@adriangb amazing thank you for the quick fix! I tested this branch against some of our code which failed previously with random RecursionErrors and now it works with this branch changes

commonism:
> @commonism any concerns with this?

No.

PR #6730, opened 2023-07-18
Fix incorrect subclass check for secretstr

### LINKED ISSUE #6729: Subclass of SecretStr expects `bytes` as input
### Initial Checks

### Description

For my use-case I have defined a subclass of the `SecretStr` type. This functionality worked fine in pydantic V1. Since the migration to V2 the follow error appeared continuously:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for [redacted]
shared_secret
  Input should be a valid bytes [type=bytes_type, input_value='[redacted]', input_type=str]
```

While the type is defined as follows:

```
class EncryptedBase64Str(SecretStr):
    def get_decrypted_value(self, aws_key_id: Union[str, UUID], **kwargs) -> str:
        decrypt_with_key(self.get_secret_value(), key=aws_key_id)
```

I.e. it is counter intuitive that the expected type is "bytes" when the class inherits from the specialized str variant.

This is due to the too restrictive test in `pydantic.types._SecretFieldValidator.__get_pydantic_core_schema__` where type is evaluated based on the types defined by pydantic:

```
        self.inner_schema = handler(str if self.field_type is SecretStr else bytes)
        error_kind = 'string_type' if self.field_type is SecretStr else 'bytes_type'
```

Children of SecretStr are neglected here.

### Example Code

```Python
from pydantic import SecretStr, ValidationError
from pydantic_settings import BaseSettings
from typing import Union
from uuid import UUID


class EncryptedBase64Str(SecretStr):
    def get_decrypted_value(self, aws_key_id: Union[str, UUID], **kwargs) -> str:
        # simplified for demonstration purposes
        return self.get_secret_value()


class Settings(BaseSettings):
    some_decryptable_secret: EncryptedBase64Str

try:
    Settings(some_decryptable_secret="abc")
except ValidationError as error:
    print(error)
    exit(0)

raise Exception("bug is not reproducible")
```

```Text
pydantic version: 2.0.3
                 install path: /Users/alexandervandenbulcke/Library/Caches/pypoetry/virtualenvs/jarvis-api-5-3ao05Y-py3.9/lib/python3.9/site-packages/pydantic
               python version: 3.9.16 (main, Dec  7 2022, 10:02:13)  [Clang 14.0.0 (clang-1400.0.29.202)]
```

AlexVndnblcke:
Currently subclasses of SecretStr fail to evaluate `self.field_type is SecretStr` and default to `bytes_type`. This is not how it should go as the subclass explicitly subclasses the SecretStr type.

This fix tests if the type is of SecretStr with `issubclass` which states:

>  A class is considered a subclass of itself

Which makes is backwards compatible for the actual SecretStr as well.

## Change Summary

Allow pydantic scheme creation to work with subclasses of `SecretStr`

## Related issue number

fix #6729

## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -728,7 +728,7 @@ class _SecretFieldValidator:
     inner_schema: CoreSchema = _dataclasses.field(init=False)
 
     def validate(self, value: _SecretField[SecretType] | SecretType, _: core_schema.ValidationInfo) -> Any:
-        error_prefix: Literal['string', 'bytes'] = 'string' if self.field_type is SecretStr else 'bytes'
+        error_prefix: Literal['string', 'bytes'] = 'string' if issubclass(self.field_type, SecretStr) else 'bytes'
         if self.min_length is not None and len(value) < self.min_length:
             short_kind: core_schema.ErrorType = f'{error_prefix}_too_short'  # type: ignore[assignment]
             raise PydanticKnownError(short_kind, {'min_length': self.min_length})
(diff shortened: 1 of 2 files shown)


AlexVndnblcke:
I'm unsure why the tests fail (error occurs during installation of the dependencies), if anyone has an idea, I'm all ears.

AlexVndnblcke:
please review

davidhewitt (approved):
Thanks, this fix looks correct to me 👍

PR #6780, opened 2023-07-20
Improve API documentation, in particular more links between usage and API docs

samuelcolvin:
Improving docs, mostly API docs and links to API docs.

Selected Reviewer: @Kludex


samuelcolvin:
please review.

docs/api/base_model.md
```diff
@@ -0,0 +1,29 @@
+Pydantic models are simply classes which inherit from `BaseModel` and define fields as annotated attributes.
+
+::: pydantic.BaseModel
+    options:
+        show_root_heading: true
+        merge_init_into_class: false
+        group_by_category: false
+        # explicit members list so we can set order and include `__init__` easily
```
Kludex, line 8: You mean that if we add `__init__`, then it's going to be the last?

We need to remember to add methods here when we implement them, right?

pydantic/config.py
```diff
@@ -44,7 +44,9 @@ def __getattribute__(self, __name: str) -> Any:
 
 
 class ConfigDict(TypedDict, total=False):
-    """A dictionary-like class for configuring Pydantic models.
+    """usage docs: https://docs.pydantic.dev/2.0/usage/model_config/
```
Kludex, line 47:
```suggestion
    """Usage docs: https://docs.pydantic.dev/2.0/usage/model_config/
```

pydantic/fields.py
```diff
@@ -670,7 +675,9 @@ def Field(  # noqa: C901
     max_length: int | None = _Unset,
     **extra: Unpack[_EmptyKwargs],
 ) -> Any:
-    """Create a field for objects that can be configured.
+    """Usage docs: https://docs.pydantic.dev/dev-v2/usage/fields
```
Kludex, line 678: Why did you change some to 2.0 and some you added dev-v2?

Kludex (approved):

pydantic/fields.py
samuelcolvin: I think they are all `2.0`, should probably be changed to `2.1` if these changes are going to be released in the next minor release.

The point is we can't just use `latest` or `dev-v2` since it'll be wrong for someone reading 2.0.3 code in 6 months time, or 6 years time 🤯.

docs/api/base_model.md
samuelcolvin: yes we need to add them.

We can use some very special regex to include `__init__` but not other dunder attributes, but the order is still very confusing. I think since this is a very important bit of the API documentation, it's worth getting it as good as possible even if that means adding members manually.

PR #6790, opened 2023-07-21
🚧 Move type replacing into `GenerateSchema`

### LINKED ISSUE #6045: V2 - Validating int->str

lig:
## Change Summary

An attempt to move type replacing logic into `GenerateSchema` based on #6535 

TODO:
- Apply type replacement for types like `Dict[str, str]`.
- Get it working with `TypedDict`: as of now `replace_types` config has no effect.

## Related issue number

Fix #6045 and some other cases

## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -59,6 +59,7 @@ class ConfigWrapper:
     validate_return: bool
     protected_namespaces: tuple[str, ...]
     hide_input_in_errors: bool
+    replace_types: dict[type, type] | None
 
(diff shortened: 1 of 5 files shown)


lig:
We've decided to go with #6737 for now instead

samuelcolvin:
closing since we've decided to go another route, we have this code in case we need it in future.

Issue #6783, opened 2023-07-20
Creating dataclass(slots=True) instance fails with defaults


bluetech:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

The example code fails as follows:

```
Traceback (most recent call last):
  File "x.py", line 7, in <module>
    A()
  File "venv/lib/python3.11/site-packages/pydantic/_internal/_dataclasses.py", line 124, in __init__
    s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
AttributeError: 'A' object attribute 'a' is read-only
```

It does work with non-pydantic dataclasses.

### Example Code

```Python
from pydantic import dataclasses

@dataclasses.dataclass(slots=True)
class A:
    a: int = 0

A()
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0.3
        pydantic-core version: 2.3.0 release build profile
                 install path: /venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.4 (main, Jun  6 2023, 22:16:46) [GCC 12.3.0]
                     platform: Linux-6.4.1-arch2-1-x86_64-with-glibc2.37
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @samuelcolvin

hramezani:
Hi @bluetech ✋ . happy to see you here.

Thanks for reporting this 🙏 

I can confirm this.

PR #6796, opened 2023-07-21
Fix pydantic dataclasses that use slots with default values

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6783


samuelcolvin (approved):
LGTM

Issue #6745, opened 2023-07-19
Computed Fields that depend on private fields don't work properly


rizenfrmtheashes:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

When I try to fetch a computed field, or look it up from `.model_dump()`, and it uses private fields I get str representation of the field annotation rather than the str representation of the value itself. 

### Example Code

```Python
from pydantic import BaseModel, Field, UUID4, computed_field
import uuid

class TestClassUUID(BaseModel):
    _foo: UUID4 = Field(alias="foo")
    _bar: UUID4 = Field(alias="bar")

    @computed_field
    def pain(self) -> str:
        return f"{self._bar}#{self._foo}"

thing = TestClassUUID(foo=uuid.uuid4(), bar=uuid.uuid4())
thing.model_dump()
# > {'pain': "annotation=NoneType required=True alias='blah' alias_priority=2#annotation=NoneType required=True alias='foo' alias_priority=2"}

class TestClass2(BaseModel):
    foo_new: UUID4 = Field(alias="foo")
    bar_new: UUID4 = Field(alias="bar")

    @computed_field
    def pain(self) -> str:
        return f"{self.foo_new}#{self.bar_new}"

thing = TestClass2(foo=uuid.uuid4(), bar=uuid.uuid4())
thing.model_dump()
# > {'foo_new': UUID('a11d0645-e1e6-4aaf-a003-0adc560ae49f'),
#    'bar_new': UUID('2d05147a-4433-41f2-b111-554245ad2f78'),
#    'pain': 'a11d0645-e1e6-4aaf-a003-0adc560ae49f#2d05147a-4433-41f2-b111-554245ad2f78'}
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.0.3
        pydantic-core version: 2.3.0 release build profile
                 install path: /Users/dhruv/.venv/estuary_beam/lib/python3.11/site-packages/pydantic
               python version: 3.11.3 (main, Jul 15 2023, 20:48:19) [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.4-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @Kludex

adriangb:
It seems like this has nothing to do with computed fields:

```python
from pydantic import BaseModel, Field, computed_field

class Model(BaseModel):
    _foo: str = Field(alias="foo")

[...]

PR #6797, opened 2023-07-21
Error if an invalid field name is used with Field

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6745

### DIFF
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -308,16 +308,23 @@ def inspect_namespace(  # noqa C901
         elif isinstance(value, ModelPrivateAttr):
             if var_name.startswith('__'):
                 raise NameError(
-                    f'Private attributes "{var_name}" must not have dunder names; '
-                    'use a single underscore prefix instead.'
+                    'Private attributes must not use dunder names;'
+                    f' instead of {var_name!r} use a single underscore prefix instead.'
                 )
             elif is_valid_field_name(var_name):
                 raise NameError(
-                    f'Private attributes "{var_name}" must not be a valid field name; '
-                    f'use sunder names, e.g. "_{var_name}"'
+                    'Private attributes must not use valid field names;'
+                    f' instead of {var_name!r} use sunder names, e.g. {"_" + var_name!r}'
                 )
             private_attributes[var_name] = value
             del namespace[var_name]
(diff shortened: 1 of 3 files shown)


pydantic/_internal/_model_construction.py
```diff
@@ -308,16 +308,23 @@ def inspect_namespace(  # noqa C901
         elif isinstance(value, ModelPrivateAttr):
             if var_name.startswith('__'):
                 raise NameError(
-                    f'Private attributes "{var_name}" must not have dunder names; '
-                    'use a single underscore prefix instead.'
+                    'Private attributes must not use dunder names;'
+                    f' instead of {var_name!r} use a single underscore prefix instead.'
```
hramezani, line 312:
```suggestion
                    f' use a single underscore prefix instead of {var_name!r}.'
```

pydantic/_internal/_model_construction.py
```diff
@@ -308,16 +308,23 @@ def inspect_namespace(  # noqa C901
...
-                    f'use sunder names, e.g. "_{var_name}"'
+                    'Private attributes must not use valid field names;'
+                    f' instead of {var_name!r} use sunder names, e.g. {"_" + var_name!r}'
                 )
             private_attributes[var_name] = value
             del namespace[var_name]
+        elif isinstance(value, FieldInfo) and not is_valid_field_name(var_name):
+            suggested_name = var_name.lstrip('_') or 'my_field'  # don't suggest '' for all-underscore name
+            raise NameError(
+                f'Fields must not use names with leading underscores;'
```
hramezani, line 324:
```suggestion
                f'Field() must not use names with leading underscores;'
```

hramezani (approved):
Other than minor suggestions LGTM!

pydantic/_internal/_model_construction.py
dmontagu: We say `Private attributes` above, which I was trying to mirror here with the description of what it was, rather than the function used to create it. I actually think it's less confusing like this because people might not realize that calling `Field()` signals to pydantic that that thing should be a field, which is what I want to emphasize.

hramezani: Agree, I've edited my suggestion

hramezani (approved):

PR #6798, opened 2023-07-21
Make it so callable JSON schema extra works

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6647

Note that as [mentioned here](https://github.com/pydantic/pydantic/pull/6653#issuecomment-1646211152), the better solution for FastAPI is to use a custom `GenerateJsonSchema` that generates the JSON schema differently for `Optional` query params, etc.. But this at least makes it possible to express what you want on the pydantic side via:
```python
from typing import Annotated

from pydantic.json_schema import SkipJsonSchema


class Model(BaseModel):
    x: Annotated[int | SkipJsonSchema[None], Field(json_schema_extra=lambda x: x.pop('default'))] = None

print(Model.model_json_schema())
#> {'properties': {'x': {'title': 'X', 'type': 'integer'}}, 'title': 'Model', 'type': 'object'}
```

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -757,13 +757,15 @@ class GenerateSchema:
             'examples': field_info.examples,
         }
         json_schema_updates = {k: v for k, v in json_schema_updates.items() if v is not None}
-        if isinstance(field_info.json_schema_extra, dict):
-            json_schema_updates.update(field_info.json_schema_extra)
+
+        json_schema_extra = field_info.json_schema_extra
 
         def json_schema_update_func(schema: CoreSchemaOrField, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
             json_schema = {**handler(schema), **json_schema_updates}
-            if callable(field_info.json_schema_extra):
-                field_info.json_schema_extra(json_schema)
+            if isinstance(json_schema_extra, dict):
+                json_schema.update(json_schema_extra)
+            elif callable(json_schema_extra):
+                json_schema_extra(json_schema)
             return json_schema
 
         metadata = build_metadata_dict(js_annotation_functions=[json_schema_update_func])
@@ -1471,21 +1473,23 @@ class GenerateSchema:
             json_schema_update: JsonSchemaValue = {}
             if metadata.title:
                 json_schema_update['title'] = metadata.title
-            if isinstance(metadata.json_schema_extra, dict):
-                json_schema_update.update(metadata.json_schema_extra)
             if metadata.description:
                 json_schema_update['description'] = metadata.description
             if metadata.examples:
                 json_schema_update['examples'] = metadata.examples
-            if json_schema_update:
+
+            json_schema_extra = metadata.json_schema_extra
+            if json_schema_update or json_schema_extra:
 
                 def json_schema_update_func(
                     core_schema: CoreSchemaOrField, handler: GetJsonSchemaHandler
                 ) -> JsonSchemaValue:
                     json_schema = handler(core_schema)
                     json_schema.update(json_schema_update)
-                    if callable(metadata.json_schema_extra):
-                        metadata.json_schema_extra(json_schema)
+                    if isinstance(json_schema_extra, dict):
+                        json_schema.update(json_schema_extra)
+                    elif callable(json_schema_extra):
+                        json_schema_extra(json_schema)
                     return json_schema
 
                 CoreMetadataHandler(schema).metadata.setdefault('pydantic_js_annotation_functions', []).append(
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -757,10 +757,14 @@ class GenerateSchema:
             'examples': field_info.examples,
         }
         json_schema_updates = {k: v for k, v in json_schema_updates.items() if v is not None}
-        json_schema_updates.update(field_info.json_schema_extra or {})
+        if isinstance(field_info.json_schema_extra, dict):
+            json_schema_updates.update(field_info.json_schema_extra)
 
         def json_schema_update_func(schema: CoreSchemaOrField, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
-            return {**handler(schema), **json_schema_updates}
+            json_schema = {**handler(schema), **json_schema_updates}
+            if callable(field_info.json_schema_extra):
+                field_info.json_schema_extra(json_schema)
+            return json_schema
 
         metadata = build_metadata_dict(js_annotation_functions=[json_schema_update_func])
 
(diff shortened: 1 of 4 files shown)


dmontagu:
please review

pydantic/fields.py
```diff
@@ -356,14 +357,27 @@ class MyModel(pydantic.BaseModel):
...
+            flattened_field_infos.extend(x for x in field_info.metadata if isinstance(x, FieldInfo))
+            flattened_field_infos.append(field_info)
+        field_infos = tuple(flattened_field_infos)
         new_kwargs: dict[str, Any] = {}
+        metadata = {}
         for field_info in field_infos:
             new_kwargs.update(field_info._attributes_set)
+            for x in field_info.metadata:
+                if not isinstance(x, FieldInfo):
+                    metadata[type(x)] = x
```
dmontagu, line 376: I ran into a case where this was necessary to get proper merging of FieldInfos that use `gt`, `lt`, etc.

pydantic/_internal/_generate_schema.py
```diff
@@ -1480,6 +1462,37 @@ def json_schema_update_func(
             return maybe_updated_schema
         return original_schema
 
+    def apply_single_annotation_json_schema(
```
dmontagu, line 1465: I had to break this out as a separate function in order to ensure the callable JSON schema ran _after_ all modifications to the core schema that influence the json schema are performed. Otherwise, for example, you can remove the `default`.

pydantic/_internal/_generate_schema.py
```diff
@@ -1493,6 +1506,7 @@ def _get_wrapped_inner_schema(
         def new_handler(source: Any) -> core_schema.CoreSchema:
             schema = metadata_get_schema(source, get_inner_schema)
             schema = self.apply_single_annotation(schema, annotation)
+            schema = self.apply_single_annotation_json_schema(schema, annotation)
```
dmontagu, line 1509: As you can see here, everything related to JSON schema now runs _after_ everything not related to JSON schema.

tests/test_json_schema.py
```diff
@@ -5149,3 +5164,23 @@ class ModelTest(BaseModel):  # noqa: F811
...
+        s.pop('default')
+
+    class Model(BaseModel):
+        a: int = Field(default=1, json_schema_extra=pop_default)
+        b: Annotated[int, Field(json_schema_extra=pop_default)] = 2
+        c: Annotated[int, Field(json_schema_extra=pop_default), Field(default=3)]
+        d: Annotated[int, Field(default=4), Field(json_schema_extra=pop_default)]
+        # TODO: it doesn't work properly to have both annotation and assigned value of FieldInfo:
+        # e: Annotated[int, Field(json_schema_extra=pop_default)] = Field(default=5)
+        # f: Annotated[int, Field(default=6)] = Field(json_schema_extra=pop_default)
```
dmontagu, line 5180: I think it would be possible to make this and the previous case work (i.e., `e` and `f`), but I tried and it ended up being a significantly bigger change and I wasn't able to see it _all_ the way through and get all tests passing. So rather than wasting time on it for now, I just added a note here. Happy to create an issue for this if there isn't one already. (CC @adriangb)

While it might be tempting to try to make this an error, for various reasons, I think it would be preferable if assigning a `FieldInfo` was equivalent to adding it as the last annotation on the field, which _does_ work as shown above.

pydantic/fields.py
adriangb: This stuff is getting really annoying. @samuelcolvin not having to do this stuff is one of the big advantages of the Output enum thing we were discussing the other day.

pydantic/_internal/_generate_schema.py
adriangb: Typo?

tests/test_json_schema.py
adriangb: I do think we should make it work at some point. Is it not working now? I thought it was.

adriangb:
Can’t this go in https://github.com/pydantic/pydantic/blob/6ef959fa69d1ab1aedf2778b6c8f452e0fee37b6/pydantic/json_schema.py#L2174? To avoid making Field even more complex.

pydantic/_internal/_generate_schema.py
```diff
@@ -757,10 +757,14 @@ def set_discriminator(schema: CoreSchema) -> CoreSchema:
             'examples': field_info.examples,
         }
         json_schema_updates = {k: v for k, v in json_schema_updates.items() if v is not None}
-        json_schema_updates.update(field_info.json_schema_extra or {})
+        if isinstance(field_info.json_schema_extra, dict):
+            json_schema_updates.update(field_info.json_schema_extra)
```
samuelcolvin, line 761: why don't we call `field_info.json_schema_extra` if it's a function here?

pydantic/_internal/_generate_schema.py
```diff
@@ -757,10 +757,14 @@ def set_discriminator(schema: CoreSchema) -> CoreSchema:
...
 
         def json_schema_update_func(schema: CoreSchemaOrField, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
-            return {**handler(schema), **json_schema_updates}
+            json_schema = {**handler(schema), **json_schema_updates}
+            if callable(field_info.json_schema_extra):
+                field_info.json_schema_extra(json_schema)
```
samuelcolvin, line 766: again, I'm a bit lost why we don't update do `json_schema.update(field_info.json_schema_extra)` here if it's a dict?

If you sure these are fine, great.

samuelcolvin (approved):
Looks fine, I'm a bit lost on a few things but if you're confident, let's merge.

pydantic/_internal/_generate_schema.py
dmontagu: okay so, in the update func, we have:
```python
            json_schema = {**handler(schema), **json_schema_updates}
```
and `handler(schema)` may result in a different schema than just the updates; `json_schema_updates` is not the whole json schema. In particular, getting stuff like `"type": "object"` (or even other types for `RootModel` subclasses), you probably want that to be accessible in the `json_schema_extra` call, which is why that needs to happen after the call to `handler` in the `json_schema_update_func` below.

pydantic/_internal/_generate_schema.py
dmontagu: I think we can clean this up, I'll push a commit doing that shortly

pydantic/_internal/_generate_schema.py
dmontagu: No typo as far as I can tell — the `self.apply_single_annotation` does everything not related to JSON schema, then after that, `self.apply_single_annotation_json_schema` does everything related to json schema. I think that's what I said in the comment above?

tests/test_json_schema.py
dmontagu: It works for most things, it just doesn't work for `json_schema_extra` and `default` specifically, annoyingly.

dmontagu: Also, even on main, this doesn't work:
```python

class Model(BaseModel):
    a: Annotated[int, Field(default=1)] = Field()
Model()
```
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Model
a
  Field required [type=missing, input_value={}, input_type=dict]
```

dmontagu: I was able to fix this through an appropriate change in `FieldInfo.from_annotated_attribute` 💪

PR #6809, opened 2023-07-22
Fix annotated_types.MaxLen validator for custom sequence types

### LINKED ISSUE #6808: MaxLen annotation validator for custom sequences is incorrect
### Initial Checks

### Description

When using an `annotated_type.MaxLen` annotation with a custom sequence it rejects sequences that have precisely the maximum length when it should only reject if they are greater.

I'm pretty sure this is because [the validator function](https://github.com/pydantic/pydantic/blob/6ef959fa69d1ab1aedf2778b6c8f452e0fee37b6/pydantic/_internal/_validators.py#L266C27-L266C27) used for custom sequences currently checks for strict inequality rather than less or equal.

### Example Code

```Python
from typing import Annotated
from annotated_types import MaxLen

class MySequence(BaseModel):

    inner: list[int]

    def __len__(self):
        return len(self.inner)

    def __getitem__(self, __key: int) -> int:
        return self.inner[__key]


class MyModel(BaseModel):
    seq: Annotated[MySequence, MaxLen(3)]


MyModel.model_validate({"seq": {"inner": [1, 2, 3]}})
```

```Text
pydantic version: 2.0.3
python version: 3.11.4 | packaged by Anaconda, Inc. | (main, Jul  5 2023, 13:47:18) [MSC v.1916 64 bit (AMD64)]
optional deps. installed: ['typing-extensions'];
```

ImogenBits:
## Change Summary

Makes the validator function used for `annotated_types.MaxLen` with custom sequences use an inclusive upper bound.

## Related issue number

fix #6808

## Checklist

* [ ] Tests pass on CI

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_validators.py
+++ b/pydantic/_internal/_validators.py
@@ -263,7 +263,7 @@ def min_length_validator(x: Any, min_length: Any) -> Any:
 
 
 def max_length_validator(x: Any, max_length: Any) -> Any:
-    if not (len(x) <= max_length):
+    if len(x) > max_length:
         raise PydanticKnownError(
             'too_long',
             {'field_type': 'Value', 'max_length': max_length, 'actual_length': len(x)},
--- a/pydantic/_internal/_validators.py
+++ b/pydantic/_internal/_validators.py
@@ -263,7 +263,7 @@ def min_length_validator(x: Any, min_length: Any) -> Any:
 
 
 def max_length_validator(x: Any, max_length: Any) -> Any:
-    if not (len(x) < max_length):
+    if not (len(x) <= max_length):
         raise PydanticKnownError(
             'too_long',
             {'field_type': 'Value', 'max_length': max_length, 'actual_length': len(x)},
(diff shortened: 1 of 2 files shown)


ImogenBits:
please review

didn't add any unit tests as I'm not sure where they would go, can't find any similar ones.

pydantic/_internal/_validators.py
```diff
@@ -263,7 +263,7 @@ def min_length_validator(x: Any, min_length: Any) -> Any:
 
 
 def max_length_validator(x: Any, max_length: Any) -> Any:
-    if not (len(x) < max_length):
+    if not (len(x) <= max_length):
```
hramezani, line 266:
```suggestion
    if len(x) > max_length:
```

hramezani:
Thanks @ImogenBits for this patch 👍 


> didn't add any unit tests as I'm not sure where they would go, can't find any similar ones.

You can adjust the test in https://github.com/pydantic/pydantic/blob/53207640a308806373bdeb1f4dea432f9c005ce2/tests/test_types.py#L5379

please update

hramezani:
please review

samuelcolvin (approved):

PR #6810, opened 2023-07-23
Fix float -> Decimal coercion precision loss

### LINKED ISSUE #6807: Parsing JSON number to Decimal loose precision
### Initial Checks

### Description

I believe PR #6327 introduced Decimal validator regression.

https://github.com/pydantic/pydantic/blob/6ef959fa69d1ab1aedf2778b6c8f452e0fee37b6/pydantic/_internal/_std_types_schema.py#L189-L194
Such a schema union causes that in the first place the JSON literal data is cast to the float type. This in turn causes a loss of accuracy upon subsequent conversion to Decimal.

Attached script results are:

v2.0.0
```
python -> a=Decimal('0.1234')
json -> a=Decimal('0.12339999999999999580335696691690827719867229461669921875')
```
This result was reported as #6295 fixed by #6321

v2.0.2
```
python -> a=Decimal('0.1234')
json -> a=Decimal('0.1234')
```
I believe this result as a correct behaviour, after #6321 is merged. 

v2.0.3
```
python -> a=Decimal('0.12339999999999999580335696691690827719867229461669921875')
json -> a=Decimal('0.12339999999999999580335696691690827719867229461669921875')
```
This result is actual behaviour with v2.0.3 installed, after #6327 merged. Validation logic was moved to pydantic-core in order to achieve speedup, but precision lost is introduced as a side effect.

According to Contribution Guide I'm writing this issue, but I think that I'm also able to start PR to fix it. It should be simple as removing single
https://github.com/pydantic/pydantic/blob/6ef959fa69d1ab1aedf2778b6c8f452e0fee37b6/pydantic/_internal/_std_types_schema.py#L191
to avoid conversion to float and result in the same validation logic as in v2.0.2
EDIT: Also `str_schema` `strict` param should be set to `False`.

### Example Code

```Python
from decimal import Decimal

class TestModel(BaseModel):
    a: Decimal

data = {"a": 0.1234}
data_json = '{"a":0.1234}'

print('python ->',TestModel.model_validate(data))
print('json ->',TestModel.model_validate_json(data_json))
```

```Text
pydantic version: 2.0.3
                 install path: /home/kuba/pydanticDecimal/venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.3 (main, Apr  5 2023, 14:15:06) [GCC 9.4.0]
                     platform: Linux-5.15.0-75-generic-x86_64-with-glibc2.31
```

Selected Assignee: @Kludex

adriangb:
Fixes #6807

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_std_types_schema.py
+++ b/pydantic/_internal/_std_types_schema.py
@@ -180,7 +180,7 @@ class DecimalValidator:
     def __get_pydantic_core_schema__(self, _source_type: Any, _handler: GetCoreSchemaHandler) -> CoreSchema:
         Decimal = decimal.Decimal
 
-        def to_decimal(v: Any) -> decimal.Decimal:
+        def to_decimal(v: str) -> decimal.Decimal:
             try:
                 return Decimal(v)
             except decimal.DecimalException as e:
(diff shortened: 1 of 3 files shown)


adriangb:
please review cc @hramezani

hramezani (approved):

Issue #6375, opened 2023-07-02
Support `json_encoders` in Pydantic v2


adriangb:
We deprecated `json_encoders` from Config in v2, but I'm not sure that there's a good replacement.

> Due to performance overhead and implementation complexity, we have now removed support for specifying `json_encoders` in the model config. This functionality was originally added to achieve custom serialization logic, and we think the new serialization decorators are a better choice in most common scenarios. However, if your usage of `json_encoders` is not compatible with the new serialization decorators, please create a GitHub issue letting us know. See [Custom data types](usage/types/custom.md) for further details.

Firstly, this config option is not equivalent to custom data types because a custom data type would require replacing that type everywhere, adding a lot of migration friction. I've tried migrating several libraries and this is one of the trickiest bits.

Second, I don't think the performance justification is valid: if you need to run Python code to serialize that's going to be the case even with custom types. I don't see any reason why we can't add a `PlainSerializerFunctionSerSchema` whenever we encounter a type that matches `json_encoders` statically while building the `CoreSchema` unless there's some dynamic runtime behavior in v1 that I'm not understanding.

Selected Assignee: @dmontagu

adriangb:
We eventually want to support something similar for validation, perhaps allowing mapping `{pd.Timestamp: CustomPdTimestampType}`. I've seen `json_encoders` often used with `arbitrary_types_allowed`, allowing mapping to a custom type would be the right replacement for both.

We happen to [have a `serialization` field for definitions](https://github.com/pydantic/pydantic-core/blob/d285975d8634c793e2cf8809b4c7cd3ba013c3d2/python/pydantic_core/core_schema.py#L3634) so we can actually add serialization to any type without breaking references, but we can't do the same for validation in general. That is, there's no way to add a `greater than 0` constraint on a type without creating a new reference. In particular, if two libraries overrode a type in this way and they were both used within one model/schema we'd have no way to differentiate them unless we create two different references. That's a hard problem that we're not ready to solve yet.

tuukkamustonen:
Commeting on a closed ticket but... so the support has been added back, but there's still the deprecation note.

Is there already a better solution (field specific serialization adds so much noise) or should we tolerate (or silence) the warning, and trust that reincarnation of the feature will appear in V3?

WilliamStam:
running into this exact thing at the moment. since the depreciation warning is messing with pytest n stuff tried fixing it. 

used to use it for stuff like

```python
class BaseModel(ParentBaseModel):
[...]

PR #6811, opened 2023-07-23
Add back support for `json_encoders`

adriangb:
Closes https://github.com/pydantic/pydantic/issues/6726, closes https://github.com/pydantic/pydantic/issues/6375, 

Selected Reviewer: @dmontagu


tests/mypy/outputs/1.1.1/mypy-plugin_ini/pydantic_settings.py
```diff
@@ -10,6 +10,3 @@ class Settings(BaseSettings):
 s = Settings(foo='test', _case_sensitive=True, _env_prefix='test__', _env_file='test')
 
 s = Settings(foo='test', _case_sensitive=1, _env_prefix=2, _env_file=3)
-# MYPY: error: Argument "_case_sensitive" to "Settings" has incompatible type "int"; expected "Optional[bool]"  [arg-type]
-# MYPY: error: Argument "_env_prefix" to "Settings" has incompatible type "int"; expected "Optional[str]"  [arg-type]
-# MYPY: error: Argument "_env_file" to "Settings" has incompatible type "int"; expected "Optional[Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]]"  [arg-type]
```
adriangb, line 15: Ugh not sure why this changed, it seems totally unrelated to this PR... @dmontagu any ideas?

adriangb:
please review

tests/mypy/outputs/1.1.1/mypy-plugin-strict-no-any_ini/dataclass_no_any.py
```diff
@@ -0,0 +1,15 @@
...
+# MYPY: error: Expression type contains "Any" (has type overloaded function)  [misc]
+class Foo:
+    foo: int
+
+
+@dataclass(config=ConfigDict(title='Bar Title'))
+# MYPY: error: Expression type contains "Any" (has type "Type[ConfigDict]")  [misc]
+# MYPY: error: Expression type contains "Any" (has type "ConfigDict")  [misc]
+class Bar:
+    bar: str
```
adriangb, line 15: @hramezani did you intentionally add this error?

hramezani: No, It was generated by the script. Actually, we can say by your change.

dmontagu: The point of this test was specifically to make sure this error _didn't_ get created. However, I've discovered that using a type alias for the callable in `json_encoders` seems to make this go away. (At least with mypy 1.1.1...), which I've just pushed in a commit. So hopefully that resolves the issue.

pydantic/_internal/_generate_schema.py
```diff
@@ -210,6 +211,42 @@ def modify_model_json_schema(
...
+    # Decimal.__class__.__mro__ (and probably other cases) doesn't include Decimal itself
+    for base in (tp, *tp.__class__.__mro__[:-1]):
+        try:
+            encoder = json_encoders[base]
+        except KeyError:
+            continue
+
+        serialization = schema.get('serialization') or core_schema.plain_serializer_function_ser_schema(encoder)
+        if 'ref' in schema:
+            return core_schema.definition_reference_schema(schema_ref=schema['ref'], serialization=serialization)  # type: ignore
```
dmontagu, line 241: This seems problematic to me — it's not currently hit by any of the tests (as determined by putting `if 'ref' in schema: assert False`, and it seems like in principle it could result in a definition getting removed from the schema. Does that make sense? I'm trying to look further into it now, maybe it happens to be the case that this can't cause issues...

dmontagu:
Maybe enough of an edge case to ignore, but it might be nice to improve the way you can use a self-referential JSON encoder.

In particular, the following works:
```python


class Model1(BaseModel):
    model_config = {'json_encoders': {}}
    x: 'Model1 | None'


Model1.model_config['json_encoders'][Model1] = lambda x: x.x if x.x is not None else 1
del Model1.__pydantic_core_schema__
Model1.model_rebuild(force=True)

print(Model1(x=Model1(x=None)).model_dump_json())
#> {"x":1}
```
But obviously this is a lot worse than doing something like `model_config = {'json_encoders': {'Model1': lambda x: x.x}}`

I guess maybe we can postpone this as a feature request for the future.

pydantic/_internal/_generate_schema.py
```diff
@@ -210,6 +211,42 @@ def modify_model_json_schema(
...
+        return schema
+    # Check the class type and its superclasses for a matching encoder
+    # Decimal.__class__.__mro__ (and probably other cases) doesn't include Decimal itself
+    for base in (tp, *tp.__class__.__mro__[:-1]):
+        try:
+            encoder = json_encoders[base]
+        except KeyError:
+            continue
+
+        serialization = schema.get('serialization') or core_schema.plain_serializer_function_ser_schema(encoder)
```
dmontagu, line 239:
```suggestion
        serialization = core_schema.plain_serializer_function_ser_schema(encoder)
```
It already returns above if `schema.get('serialization') is not None`, so I think this or can be dropped

pydantic/_internal/_generate_schema.py
dmontagu: I haven't been able to reproduce it (though investigating did lead me to the other comment I made, about self-referencing encoders), but yeah this does seem to make the assumption that the original definition will be present elsewhere, I'm not sure whether that is a safe assumption or not, maybe you know based on the way definitions get stored in `GenerateSchema` or whatever.

dmontagu (approved):
Other than the comments I've made, this all looks good to me, so I'll approve for now so you can merge after reviewing those comments and making changes if appropriate.

pydantic/_internal/_generate_schema.py
```diff
@@ -210,6 +211,42 @@ def modify_model_json_schema(
...
+    json_encoders: JsonEncoders | None, tp: Any, schema: CoreSchema
+) -> CoreSchema:
+    """Iterate over the json_encoders and add the first matching encoder to the schema.
+
+    Args:
+        json_encoders (JsonEncoders | None): A dictionary of types and their encoder functions.
+        tp (Any): The type to check for a matching encoder.
+        schema (CoreSchema): The schema to add the encoder to.
+    """
+    if json_encoders is None:
```
dmontagu, line 227:
```suggestion
    if not json_encoders:
```
also handles the empty-dict case 🤷‍♂️

pydantic/_internal/_generate_schema.py
```diff
@@ -210,6 +211,42 @@ def modify_model_json_schema(
...
+        return schema
+    if 'serialization' in schema:
+        return schema
+    # Check the class type and its superclasses for a matching encoder
+    # Decimal.__class__.__mro__ (and probably other cases) doesn't include Decimal itself
+    for base in (tp, *tp.__class__.__mro__[:-1]):
+        try:
+            encoder = json_encoders[base]
+        except KeyError:
+            continue
```
dmontagu, line 237:
```suggestion
        encoder = json_encoders.get(base)
        if encoder is None:
            continue
```
fewer loc and since I think it should be faster since you'll get a KeyError more often than not (my understanding is that using `try: except:` is faster when you don't hit the exception, and slower when you do). I have no problem keeping it as-is if you prefer though.

dmontagu: Okay yeah, I was curious so actually tried, and the `.get` seems to be ~5x faster:
```python
import time

a = dict(enumerate(range(100)))
n = 10_000_000

t0 = time.time()
for _ in range(n):
    try:
        a[100]
    except KeyError:
        continue
t1 = time.time()
print(f'{t1 - t0:.3f}')
#> 2.540

t0 = time.time()
for _ in range(n):
    x = a.get(100)
    if x is None:
        continue
t1 = time.time()
print(f'{t1 - t0:.3f}')
#> 0.646
```

dmontagu: And if you change it to `a[99]` and `a.get(99)` instead of 100, the `try: except:` is ~2x faster, so, intuition confirmed.

I also confirmed these performance ratios hold if you make the dict have just a single item.

adriangb: Ok just keep in mind that this was copy pasta from V1

adriangb:
> Maybe enough of an edge case to ignore, but it might be nice to improve the way you can use a self-referential JSON encoder.
> I guess maybe we can postpone this as a feature request for the future.

Yeah I don't think this worked in v1 so let's table it.

pydantic/_internal/_generate_schema.py
```diff
@@ -1413,6 +1456,7 @@ def inner_handler(obj: Any) -> CoreSchema:
         if pydantic_js_annotation_functions:
             metadata = CoreMetadataHandler(schema).metadata
             metadata.setdefault('pydantic_js_annotation_functions', []).extend(pydantic_js_annotation_functions)
+        schema = _add_custom_serialization_from_json_encoders(self.config_wrapper.json_encoders, source_type, schema)
         return schema
```
samuelcolvin, line 1460:
```suggestion
        return _add_custom_serialization_from_json_encoders(self.config_wrapper.json_encoders, source_type, schema)
```

pydantic/_internal/_config.py
```diff
@@ -47,6 +47,7 @@ class ConfigWrapper:
     ignored_types: tuple[type, ...]
     allow_inf_nan: bool
     json_schema_extra: dict[str, object] | JsonSchemaExtraCallable | None
+    json_encoders: dict[type[object], JsonEncoder] | None
```
samuelcolvin, line 50: you need to add this to the docstring above, perhaps including the warning.

samuelcolvin (commented):
otherwise LGTM. Do we need to add, alter or remove something in the migration guide?

pydantic/_internal/_generate_schema.py
```diff
@@ -235,6 +238,13 @@ def _add_custom_serialization_from_json_encoders(
         if encoder is None:
             continue
 
+        from pydantic.version import VERSION
+
+        warnings.warn(
+            f'`json_encoders` is deprecated. See https://docs.pydantic.dev/{VERSION}/usage/serialization/#custom-serializers for alternatives',
+            PydanticDeprecatedSince20,
```
samuelcolvin, line 243: this is incorrect, it should only be the major and minor versions.

Maybe the solution is to have a proper warnings page like we have for user and validation errors. I can work on this today.

Fixing this should not block release.

samuelcolvin (changes requested):

pydantic/_internal/_generate_schema.py
adriangb: It actually redirects, e.g. https://docs.pydantic.dev/2.0.2/usage/serialization/#custom-serializers works

adriangb:
please review

pydantic/_internal/_config.py
adriangb: done

pydantic/_internal/_generate_schema.py
samuelcolvin: only for specific versions see #6527.

I'll provide a general fix.

adriangb: Ah I see. Okay then we can just strip the patch version.

samuelcolvin (approved):
LGTM.

PR #6820, opened 2023-07-24
Pydantic plugins - `entry-point` support

### LINKED ISSUE #6765: Support plugins
### Initial Checks

### Description

We want to be able to _know_ the "what", "when", and "how" things happens internally in Pydantic, and be able to perform some action according to it.

For that, we thought about a plugin system, using [entry points](https://packaging.python.org/en/latest/specifications/entry-points/), that will allow external packages (plugins) to _intercept_ events in Pydantic, like validation and schema creation.

### User side

Let's see an example on how this would look like for the Pydantic user's side:

```py

class Model(BaseModel, observe="all")
    a: int

Model.model_validate({"a": 1})
```

In this case, we allow `Model` to receive another class parameter called `observe`. On its hand, given this parameter, Pydantic will check if there's a plugin installed that uses this parameter, and the plugin will be able to execute the logic it wants, wrapping Pydantic events (like validation, and schema creation, as mentioned above).

### Pydantic side

On the Pydantic's side, we'll introspect the packages installed, looking for an entry point, similar as to what [flake8](https://flake8.pycqa.org/en/latest/plugin-development/index.html) and [pytest](https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html) are doing. If the entry point is found, we'll register that in a plugin list, and whenever an event happens, we'll check if the plugin wants to do something about it.

### Plugin side

On the plugin side, I imagine the implementation as simple as a callable. Let's see the minimal possible example:

```py
from typing import Any
from pydantic.plugin import Plugin, OnValidatePython


def enter(
    cls: type[Any],  # type: ignore
    json_data: str | bytes | bytearray,
    *,
    strict: bool | None = None,
    context: dict[str, Any] | None = None,
) -> None:
    print(cls)


plugin = Plugin(on_validate_python=OnValidatePython(enter=enter))
```

Also, they'll need to add the entry point on the `pyproject.toml`:

```toml
[project.entry-points."pydantic"]
observe = "pydantic_sdk.plugin:plugin"
```

### Affected Components

Kludex:
## Change Summary

Implement a plugin system using [Python entry points](https://packaging.python.org/en/latest/specifications/entry-points/).

## Related issue number

- Closes #6765

## Checklist

* [X] Tests pass on CI

## Missing

I'll add them if everything is fine with the current implementation.

- [x] Tests
- [x] Documentation
- [x] Other places where we can have `PluggableSchemaValidator`


Kludex:
please review

pydantic/_internal/_dataclasses.py
```diff
@@ -162,7 +164,11 @@ def __init__(__dataclass_self__: PydanticDataclass, *args: Any, **kwargs: Any) -
     # debug(schema)
     cls.__pydantic_core_schema__ = schema = _discriminated_union.apply_discriminators(flatten_schema_defs(schema))
     simplified_core_schema = inline_schema_defs(schema)
-    cls.__pydantic_validator__ = validator = SchemaValidator(simplified_core_schema, core_config)
+    if plugins:
+        validator = PluggableSchemaValidator(simplified_core_schema, core_config)
+    else:
+        validator = SchemaValidator(simplified_core_schema, core_config)
```
samuelcolvin, line 170: this should probably be a function in `plugins`.

pydantic/plugin/loader.py
```diff
@@ -0,0 +1,54 @@
+"""This module is used to load plugins for pydantic.
+
+Plugins are installed via Python entry points, for example in `pyproject.toml`:
+
+    [tool.poetry.plugins."pydantic"]
+    observer = "pydantic_sdk:main"
```
samuelcolvin, line 6: can you show the PEP 0621 way of doing it, e.g. [here](https://github.com/samuelcolvin/pytest-pretty/blob/af6cd38ea37e5a3879428f401804d1da72177c81/pyproject.toml#L55-L56), not poetry.

Or maybe both?

pydantic/plugin/loader.py
```diff
@@ -0,0 +1,54 @@
...
+
+On Pydantic, plugins are loaded by calling `load_plugins()`. Plugins are loaded in the order
+they are found, and the order they are found is not guaranteed.
+
+Consider that you have a plugin called `observer`, then you can use it like this:
+
+    from pydantic import BaseModel
+
+    class Foo(BaseModel, observer='all'):
+        ...
```
samuelcolvin, line 18: use code fences and the `py` language identifier.

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,86 @@
...
+    def on_error(self, error: ValidationError) -> None:
+        """Call `on_error` step for plugins."""
+        self._on_step('on_error')(error)
+
+    def _on_step(self, step: Literal['enter', 'on_success', 'on_error']):
+        """Call a step for plugins."""
+
+        def on_step(*args: Any, **kwargs: Any) -> None:
+            for plugin in plugins:
+                with contextlib.suppress(AttributeError, TypeError):
```
samuelcolvin, line 64: I'm not sure we should be suppressing these errors?

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,86 @@
...
+        self.schema_validator = SchemaValidator(*args, **kwargs)
+
+        self.instrumented_validate_json = _plug(self.schema_validator.validate_json)
+        self.instrumented_validate_python = _plug(self.schema_validator.validate_python)
+
+    def __getattr__(self, name: str) -> Any:
+        if name == 'validate_json':
+            return self.instrumented_validate_json
+        elif name == 'validate_python':
+            return self.instrumented_validate_python
```
samuelcolvin, line 85: if you rename `instrumented_validate_json` to `validate_json`, you shouldn't need these two.

samuelcolvin (changes requested):
otherwise I think this looks like a great start.

Obviously needs tests and documentation.

pydantic/plugin/loader.py
samuelcolvin: this is the code you would use if you were using poetry, I'd prefer the main example to be how to do it if you're using a PEP 621 pyproject.toml.

E.g:

```toml
[project.entry-points.pydantic]
pydantic_sdk = "pydantic_sdk:main"
```

or whatever it is.

Kludex: Oh! I didn't even notice I was using poetry there... 😅

Kludex:
@samuelcolvin Should I allow `BaseModel.__init_subclass__` to receive keyword arguments on this PR?

Kludex:
Only some tests missing here.

Kludex:
please review

pydantic/plugin/_loader.py
```diff
@@ -0,0 +1,58 @@
...
+    """Load plugins for pydantic.
+
+    Inspired by:
+    - https://github.com/pytest-dev/pytest/blob/5c0e5aa39916e6a49962e662002c23f578e897c9/src/pluggy/_manager.py#L352-L377
+    """
+    plugins: list[Plugin] = []
+    for dist in importlib_metadata.distributions():
+        for entry_point in dist.entry_points:
+            if entry_point.group == GROUP:
+                plugins.append(entry_point.load())
```
hramezani, line 54: This line causes a circular import error if the pluging uses `Pydantic` functionality e.g. `TypeAdapter`.

Here is a minimal plugin to reproduce the problem. It uses the `pydantic_sdk.Observe` and `Observe` import the `BaseSettings` and `TypeAdapter`

```py
from typing import Any

from rich.console import Console

from pydantic.plugin import OnValidatePython as _OnValidatePython
from pydantic.plugin import Plugin

console = Console()

from pydantic_sdk.observe import debug


class TestPlugin(_OnValidatePython):
    def enter(
        self,
        input: Any,
        *,
        strict: bool | None = None,
        from_attributes: bool | None = None,
        context: dict[str, Any] | None = None,
        self_instance: Any | None = None,
    ) -> None:

        debug('enter', input=input, strict=strict, from_attributes=from_attributes, context=context)


plugin = Plugin(on_validate_python=TestPlugin)

```

lig:
please review

lig:
Got an example of an ImportError. Please update.

lig:
Implemented deferred import for all plugins by default

docs/integrations/plugins.md
```diff
@@ -0,0 +1,86 @@
...
+
+In case you have a project called `pydantic-plugin`, you can create a plugin by adding the following
+to your `pyproject.toml`:
+
+```toml
+[project.entry-points.pydantic]
+pydantic_plugin = "pydantic_plugin:plugin"
+```
+
+The entry point group is `pydantic`, and the name of the entry point is the name of the plugin.
```
samuelcolvin, line 19: explain what `my_pydantic_plugin:plugin` means? I assume that `plugin` is the object in `my_pydantic_plugin` to load and use as a plugin?

docs/integrations/plugins.md
```diff
@@ -0,0 +1,86 @@
...
+
+Plugins are loaded in the order they are found, and the order they are found is not guaranteed.
+
+As a user, you can modify the behavior of the plugin in a `BaseModel` using the `plugin_settings`
+class keyword argument. This argument takes a dictionary of settings that will be passed to all plugins as is.
```
samuelcolvin, line 24:
```suggestion
config argument or class keyword argument. This argument takes a dictionary of settings that will be passed to all plugins as is.
```

samuelcolvin: add a link to config.

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,189 @@
...
+from .plugin import EventHandler, Plugin
+
+P = ParamSpec('P')
+R = TypeVar('R')
+
+
+def create_schema_validator(
+    schema: CoreSchema, config: CoreConfig | None = None, plugin_settings: dict[str, Any] | None = None
+) -> SchemaValidator:
+    """Get the schema validator class.
```
samuelcolvin, line 22:
```suggestion
    """Create a `SchemaValidator` or `PluggableSchemaValidator` if plugins are installed.
```

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,189 @@
...
+    ) -> None:
+        self.schema_validator = SchemaValidator(schema, config)
+
+        self.plugin_factory = _PluginFactory(schema, config, plugins, plugin_settings)
+
+        self.validate_json = self.plugin_factory(self.schema_validator.validate_json)
+        self.validate_python = self.plugin_factory(self.schema_validator.validate_python)
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.schema_validator, name)
```
samuelcolvin, line 52:
```suggestion
        return getattr(self._schema_validator, name)
```

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,189 @@
...
+
+        return wrapper
+
+    def gather_calls(self, callback_type: Literal['on_enter', 'on_success', 'on_error']) -> list[Callable[..., None]]:
+        calls: list[Callable[..., None]] = []
+
+        for handler in self.event_handlers:
+            if not hasattr(handler, callback_type):
+                continue
+            calls.append(getattr(handler, callback_type))
```
samuelcolvin, line 168:
```suggestion
        for handler_name in self.event_handlers:
            handler = getattr(handler, callback_type, None)
            if handler:
                calls.append(handler)
```

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,189 @@
...
+        self.plugin_factory = _PluginFactory(schema, config, plugins, plugin_settings)
+
+        self.validate_json = self.plugin_factory(self.schema_validator.validate_json)
+        self.validate_python = self.plugin_factory(self.schema_validator.validate_python)
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.schema_validator, name)
+
+
+class _PluginFactory:
```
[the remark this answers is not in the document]
samuelcolvin: Everything from here on seems like OO overkill.

Surely we can simplify it all done to single function `build_wrapper`?

It'll be easier to read and more performant.

pydantic/plugin/_loader.py
```diff
@@ -0,0 +1,81 @@
...
+    for dist in importlib_metadata.distributions():
+        for entry_point in dist.entry_points:
+            if entry_point.group != GROUP:
+                continue
+            if entry_point.value in _plugins:
+                continue
+            _plugins[entry_point.value] = entry_point.load()
+
+
+def __getattr__(attr_name: str) -> object:
```
samuelcolvin, line 66: this seems like a weird way to do things, why don't we just have a `get_plugins()` method?

pydantic/plugin/_loader.py
```diff
@@ -0,0 +1,81 @@
...
+    try:
+        load_plugins()
+    except ImportError:
+        warnings.warn(
+            'Import error while loading a Pydantic plugin could be caused by a circular import,'
+            ' avoid module level imports from the same package in your plugin'
+        )
+        raise
+
+    return set(_plugins.values())
```
samuelcolvin, line 81: returning a set here looses order and means order might change, better to return a `Sequence` and just return `_plugins.values()`.

samuelcolvin (changes requested):
Looking good, but I think we can do a lot of simplification.

pydantic/plugin/_loader.py
lig: Then we should test for order to be preserved starting from entrypoint configuration in `package.toml`.

pydantic/plugin/_loader.py
lig: well, no it will be an equivalent. it was an attempt to cache imported plugins but now it's being done in another way

pydantic/plugin/plugin.py
```diff
@@ -0,0 +1,81 @@
...
+        """Call `on_success` callback."""
+        ...
+
+    @abstractmethod
+    def on_error(self, error: ValidationError) -> None:
+        """Call `on_error` callback."""
+        ...
+
+
+class OnValidatePython(EventHandler, Protocol):
```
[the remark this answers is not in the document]
lig: Well, Protocol allows to have its implementation without inheritance. It's basically lighter and was invented to replace ABC. Why not using it then?

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,189 @@
...
+    """Get the schema validator class.
+
+    Returns:
+        If plugins are installed then return `PluggableSchemaValidator`, otherwise return `SchemaValidator`.
+    """
+    from ._loader import plugins
+
+    if plugins:
+        return PluggableSchemaValidator(schema, config, plugins, plugin_settings or {})  # type: ignore
+    return SchemaValidator(schema, config)
```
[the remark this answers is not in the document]
lig: Will do but I don't agree https://fosstodon.org/@lig/110372430220539347

pydantic/plugin/schema_validator.py
```diff
@@ -0,0 +1,189 @@
...
+        if _callback_key in self._in_call:
+            yield None
+            return
+
+        self._in_call.add(_callback_key)
+        yield func
+        self._in_call.remove(_callback_key)
+
+
+class _Event(str, Enum):
```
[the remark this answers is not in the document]
lig: We use both attr name and attr value here to map callback method names onto SchemaValidator method names. Also, we are using this enum to validate function name.

pydantic/plugin/schema_validator.py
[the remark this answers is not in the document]
lig: I see how `_PluginFactory` could be transformed into a function. However, it's a bit more complicated with `_PluginAPI` below as its instance is intentionally being created for each validation call.

docs/integrations/plugins.md
lig: Done, please check

docs/integrations/plugins.md
lig: Done, please check

pydantic/plugin/plugin.py
```diff
@@ -0,0 +1,76 @@
...
+        strict: bool | None = None,
+        context: dict[str, Any] | None = None,
+        self_instance: Any | None = None,
+    ) -> None:
+        """Call `enter` callback."""
+        pass
+
+
+@dataclass(frozen=True)
+class Plugin:
```
samuelcolvin, line 72: rename to `PydanticPlugin`, and make this a protocol, so it doesn't need to be imported.

docs/usage/plugins.md
```diff
@@ -0,0 +1,113 @@
...
+class Foo(BaseModel, plugin_settings={'my-plugin': {'observe': 'all'}}):
+    ...
+```
+
+## Build a plugin
+
+??? api "API Documentation"
+    [`pydantic.plugin.plugin`][pydantic.plugin.plugin]<br>
+
+Pydantic has an API for creating plugins. The API is exposed via the `pydantic.plugin` module.
```
lig, line 41: There is a similar sentence above

> Pydantic allows users to create plugins that can be used to extend the functionality of the library.

This one feels like a first sentence of a document but we already have one like that. It's stylistic and just felt like that. Feel free to ignore.

docs/usage/plugins.md
```diff
@@ -0,0 +1,113 @@
...
+
+On your plugin you can _wrap_ the following methods:
+
+* [`validate_python`][pydantic_core.SchemaValidator.validate_python]: Used to validate the data from a Python object.
+* [`validate_json`][pydantic_core.SchemaValidator.validate_json]: Used to validate the data from a JSON string.
+* [`validate_strings`][pydantic_core.SchemaValidator.validate_strings]: Used to validate the data from strings.
+
+For each method, you can implement the following callbacks:
+
+* `enter`: Called before the validation of a field starts.
```
lig, line 51:
```suggestion
* `on_enter`: Called before the validation of a field starts.
```

pydantic/main.py
```diff
@@ -742,7 +742,11 @@ def __getattr__(self, item: str) -> Any:
                 except KeyError as exc:
                     raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}') from exc
             else:
-                pydantic_extra = object.__getattribute__(self, '__pydantic_extra__')
+                try:
```
lig, line 745: Add a comment maybe that this is a guard for using a partially imported module in plugins?

Ok, there is one below. Maybe something short with "see below"?

tests/test_plugins.py
```diff
@@ -0,0 +1,177 @@
+from __future__ import annotations
```
lig, line 1: Shall this file be moved into `tests/plugin/` dir?

lig, line 1: There were some tests that reproduce `logfire` behavior. I think it would be useful porting those to this implementation.

lig (changes requested):

pydantic/main.py
samuelcolvin: sounds good.

docs/usage/plugins.md
samuelcolvin: feel free to tweak.

docs/usage/plugins.md
samuelcolvin: feel free to fix if you tweak this.

tests/test_plugins.py
samuelcolvin: I don't think we need this now, we can do that (via another simple package probably) in future.

samuelcolvin: the pytest shim stuff isn't necessarily testing the same stuff - we really need to write packages to file.

We could use `tmp_path`, create the different files, then run python code.

lig (approved):

samuelcolvin (approved):

Issue #6806, opened 2023-07-22
`__post_init__` executes before `@model_validator(mode="after")`


Expurple:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Disclaimer: this may be a documentation issue and not a bug. But it's still an issue in that case.

### Expected behavior

When using a Pydantic `dataclass`, I expected to see the behavior documented in [Dataclasses documentation](https://docs.pydantic.dev/latest/usage/dataclasses/#initialization-hooks) and [migration guide](https://docs.pydantic.dev/latest/migration/#changes-to-dataclasses):

>  The `__post_init__` in Pydantic dataclasses is called after validation, rather than before.

> The `__post_init__` in Pydantic dataclasses will now be called after validation, rather than before. As a result, the `__post_init_post_parse__` method would have become redundant, so has been removed.

### Actual behavior

As the title says,  `__post_init__` executes before `@model_validator(mode="after")`. See the code example.

Interestingly, `__post_init_post_parse__` works as expected. It hasn't been removed and it's definitely not redundant yet.

### Additional info

[Documentation for `functional_validators`](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.AfterValidator) doesn't mention `__post_init__` or `__post_init_post_parse__`, it only states that

> validation should be applied **after** the inner validation logic.

Also, [documentation for `dataclasses`](https://docs.pydantic.dev/latest/usage/dataclasses/#difference-with-stdlib-dataclasses) recommends:

> When substituting usage of `dataclasses.dataclass` with `pydantic.dataclasses.dataclass`, it is recommended to move the code executed in the `__post_init__` to methods decorated with `model_validator`.

This is a valid workaround, but I wonder why is this the recommened way. IMO, validators should not mutate the object and `__post_init__` is a more idiomatic choice.

### Example Code

```Python
from pydantic import model_validator
from pydantic.dataclasses import dataclass

@dataclass
class ExampleDataclass:
    a: int
    b: int

    @model_validator(mode="after")
    def _validate_relationship_between_a_and_b(self) -> "ExampleDataclass":
        # The nature of the check doesn't matter.
        if self.a == self.b:
            raise ValueError("invalid combination of a and b")
        return self

    def __post_init__(self) -> None:
        # The nature of the computation doesn't matter,
        # but it raises an error with invalid combinations of a and b.
        # In this case, ZeroDivisionError.
        self._c = 42 / (self.a - self.b)

[...]

PR #6822, opened 2023-07-24
Improve initialization hooks example

hramezani:
Closes https://github.com/pydantic/pydantic/issues/6806


hramezani:
Please review

Kludex (approved):
Do we mention that `__post_init__`` comes in the "middle" between the validators somewhere?

hramezani:
> Do we mention that `**post_init**`` comes in the "middle" between the validators somewhere?

There is [a line after this code block](https://docs.pydantic.dev/latest/usage/dataclasses/#initialization-hooks) that says

> 

> Do we mention that `**post_init**`` comes in the "middle" between the validators somewhere?

I've change it to make it more clear

docs/usage/dataclasses.md
```diff
@@ -378,25 +378,35 @@ class User:
...
-The `__post_init__` in Pydantic dataclasses is called _after_ validation, rather than before.
+The `__post_init__` in Pydantic dataclasses is called in the _middle_ of validators.
+Here is the order:
+
+* Before model validator.
+* Before field validator.
+* After field validator.
+* Field type validators. e.g. validation for types like `int`, `str`, ... .
+* `__post_init__`.
+* After model validator.
```
Kludex, line 409: I think `model_validator(mode='before')`, `field_validator(mode='before')`, etc. notation looks better. 👀 

Also, I think we call the "Field type validators" differently in the rest of the documentation? 🤔 Inner validator? I'm not sure...

hramezani: Ok, look better.
I changed them.

PR #6823, opened 2023-07-24
Add `config.defer_build` to optionally make model building lazy

### LINKED ISSUE #6742: Incorrect ForwardRef class lookup in inherited models
### Initial Checks

### Description

Let's suppose parent class has fields with `ForwardRef` and calls `model_rebuild`. It correctly build.
Let's imagine we have child class in different module. In this case for that class pydantic tries to lookup types of parent fields again, but uses current scope instead of scope of parent class.

In all files below it is expected that `Child.a` will have the type `fw_parent.A` as it is declared. But it isn't

When running `main.py` the error is:
```
pydantic.errors.PydanticUserError: `Child` is not fully defined; you should define `A`, then call `Child.model_rebuild()`.
```
It looks like pydantic tries to rebuild parent fields while it has info about them in parent class.

When running `main2.py` the error is:
```
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class '__main__.A'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
```
It looks like pydantic tries to use wrong A class from wrong file.

### Example Code

```Python
# fw_parent.py
from typing import Optional

class Model(BaseModel):
    a: Optional["A"]


class A(BaseModel):
    pass


Model.model_rebuild()

# main.py
from fw_parent import Model

class Child(Model):
    pass


Child(a=None)


# main2.py
from fw_parent import Model

class A:
    pass

class Child(Model):
    pass


Child(a=None)
```

```Text
pydantic version: 2.0.3
                 install path: /home/tishka17/src/tmp/venv/lib/python3.10/site-packages/pydantic
               python version: 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0]
                     platform: Linux-5.19.0-46-generic-x86_64-with-glibc2.35
```

Selected Assignee: @dmontagu

samuelcolvin:
## Change Summary

This proposes a new config flag `defer_model_build`, when true this means the core schema, schema validator and schema serializer are not constructed for a model when it's defined, instead they're only built when `model_rebuild()` is called, or when the model is first called.

Two reasons this can be useful:
* so models which are only ever used nested or within `TypeAdapter`s don't need all this logic calling - see #6768 and #6748
* so that the type namespace used when constructing models can be customised, fixes #6742

## Related issue number

fix #6742

## Checklist

* [x] Tests pass on CI
* [x] Documentation reflects the changes where applicable - **somewhat, I'm going to rewrite the docs for `ConfigDict` anyway, so I'll add more logic then**

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/_internal/_mock_validator.py
+++ b/pydantic/_internal/_mock_validator.py
@@ -23,7 +23,7 @@ class MockValidator:
         code: PydanticErrorCodes,
         attempt_rebuild: Callable[[], SchemaValidator | None] | None = None,
     ) -> None:
-        """Attempt rebuild."""
+        # debug(attempt_rebuild, code)
         self._error_message = error_message
         self._code: PydanticErrorCodes = code
         self._attempt_rebuild = attempt_rebuild
@@ -49,7 +49,9 @@ class MockValidator:
         return None
 
 
-def set_basemodel_mock_validator(cls: type[BaseModel], cls_name: str, undefined_name: str) -> None:
+def set_basemodel_mock_validator(
+    cls: type[BaseModel], cls_name: str, undefined_name: str = 'all referenced types'
+) -> None:
     undefined_type_error_message = (
         f'`{cls_name}` is not fully defined; you should define {undefined_name},'
         f' then call `{cls_name}.model_rebuild()`.'
--- a/pydantic/_internal/_mock_validator.py
+++ b/pydantic/_internal/_mock_validator.py
@@ -23,7 +23,6 @@ class MockValidator:
         code: PydanticErrorCodes,
         attempt_rebuild: Callable[[], SchemaValidator | None] | None = None,
     ) -> None:
-        # debug(attempt_rebuild, code)
         self._error_message = error_message
         self._code: PydanticErrorCodes = code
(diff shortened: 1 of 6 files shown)


adriangb:
Having this as a config flag sounds reasonable to me. For the reasons discussed in https://github.com/pydantic/pydantic/issues/6768 I don’t think it’s good as a default, but for large apps it seems like a reasonable trade-off.

Shouldn’t we make TypeAdapter lazy as well?

samuelcolvin:
Let's rename to `defer_build` and apply to `TypeAdapter` too.

pydantic/_internal/_mock_validator.py
```diff
@@ -23,7 +23,7 @@ def __init__(
         code: PydanticErrorCodes,
         attempt_rebuild: Callable[[], SchemaValidator | None] | None = None,
     ) -> None:
-        """Attempt rebuild."""
+        # debug(attempt_rebuild, code)
```
hramezani, line 26: I think you should remove it

hramezani (approved):

PR #6829, opened 2023-07-24
Fix serialization issue with InstanceOf

dmontagu:
Addresses a bug noticed in https://github.com/pydantic/pydantic/discussions/6819#discussioncomment-6531443 where `InstanceOf[T]` was serializing like `T` from JSON, but like `SerializeAsAny[T]` from python. 

With this PR, it always serializes like `T`, but you can still get the `SerializeAsAny` behavior by annotating with that, and either nesting works (`SerializeAsAny[InstanceOf[T]]` or `InstanceOf[SerializeAsAny[T]]`).

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/functional_validators.py
+++ b/pydantic/functional_validators.py
@@ -531,16 +531,20 @@ else:
             from pydantic import PydanticSchemaGenerationError
 
             # use the generic _origin_ as the second argument to isinstance when appropriate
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

Kludex (approved):

PR #6838, opened 2023-07-25
Update field annotations when building the schema

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6742

Currently, once you have rebuilt a model, the field annotations aren't updated from the original forward refs; as a result, when you declare subclasses in new modules, the same forward ref gets re-evaluated, even if the original model was rebuilt with a specific annotation in use.

This PR changes things so that the `FieldInfo`s in `cls.model_fields` _do_ get updated to resolve the `ForwardRef`s, ensuring that the resolved fields retain their resolved type even when new subclasses are defined in different modules.

I still need to add a test along the lines of what is described in https://github.com/pydantic/pydantic/issues/6742#issuecomment-1647856883, but I was able to confirm that files written in that way seem to work with these changes. Before I attempt to add such tests though, I was hoping to get confirmation that this approach would be accepted.

### DIFF
--- a/pydantic/_internal/_generics.py
+++ b/pydantic/_internal/_generics.py
@@ -284,7 +284,9 @@ def replace_types(type_: Any, type_map: Mapping[Any, Any] | None) -> Any:
     if origin_type is typing_extensions.Annotated:
         annotated_type, *annotations = type_args
         annotated = replace_types(annotated_type, type_map)
-        return typing_extensions.Annotated[annotated, *annotations]
+        for annotation in annotations:
+            annotated = typing_extensions.Annotated[annotated, annotation]
+        return annotated
 
     # Having type args is a good indicator that this is a typing module
     # class instantiation or a generic alias of some sort.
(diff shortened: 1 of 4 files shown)


pydantic/_internal/_generics.py
```diff
@@ -338,6 +336,34 @@ def replace_types(type_: Any, type_map: Mapping[Any, Any] | None) -> Any:
     return type_map.get(type_, type_)
 
 
+def has_instance_in_type(type_: Any, isinstance_target: Any) -> bool:
```
dmontagu, line 339: Although this doesn't really relate to pydantic generics functionality, I put this function in `_generics.py` because it is very closely related in implementation to `_generics.replace_types`, and because the interesting parts of the logic are mostly related to "walking" generic types.

dmontagu:
please review

samuelcolvin (approved):
LGTM

PR #6852, opened 2023-07-25
Deprecate `Field.include`

hramezani:
Selected Reviewer: @Kludex

### DIFF
--- a/docs/usage/serialization.md
+++ b/docs/usage/serialization.md
@@ -634,7 +634,7 @@ The same holds for the `model_dump_json` method.
 ### Model- and field-level include and exclude
 
 In addition to the explicit arguments `exclude` and `include` passed to `model_dump` and `model_dump_json` methods,
-we can also pass the `include`/`exclude` arguments directly to the `Field` constructor:
+we can also pass the `exclude` arguments directly to the `Field` constructor:
 
 Setting `exclude` on the field constructor (`Field(..., exclude=True)`) takes priority over the
 `exclude`/`include` on `model_dump` and `model_dump_json`:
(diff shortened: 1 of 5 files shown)


hramezani:
please review

docs/usage/serialization.md
```diff
@@ -705,12 +705,12 @@ print(t.model_dump(exclude={'value': True, 'user': {'username'}}))
...
 ```py
-from pydantic import BaseModel, Field, SecretStr
+from pydantic import BaseModel, SecretStr
 
 
 class User(BaseModel):
-    id: int = Field(..., include=True)
-    username: str = Field(..., include=True)  # overridden by explicit include
+    id: int
+    username: str  # overridden by explicit include
```
Kludex, line 713: Does this makes sense still?

pydantic/fields.py
```diff
@@ -793,6 +785,10 @@ def Field(  # noqa: C901
     if validation_alias in (_Unset, None):
         validation_alias = alias
 
+    include = extra.pop('include', None)  # type: ignore
+    if include is not None:
+        warn('`include` is deprecated and does nothing. It will be removed, use `exclude` instead', DeprecationWarning)
```
Kludex, line 790: AFAIU, this is not a deprecation... If the functionality is removed, it's a removal. The parameter itself is deprecated.

Also, how is `exclude` a replacement for `include`? 🤔

hramezani: > AFAIU, this is not a deprecation... If the functionality is removed, it's a removal. The parameter itself is deprecated.
> 

I think it is a deprecation as it did nothing before and still does nothing. So, no change in functionality. just warning the user.
 
> Also, how is `exclude` a replacement for `include`? 🤔

`include=False` can be replaced with `exclude=True`. That's why I suggested using `exclude`.

docs/usage/serialization.md
hramezani: No, I've just removed.

Kludex (approved):

PR #6881, opened 2023-07-26
Update TypedDict docs to correct min Python version

### LINKED ISSUE #6860: Pydantic V2 stopped validating TypedDict under Python 3.11
### Initial Checks

### Description

So I was using pydantic v2 to validate some json using `typing.TypedDict` on python 3.11; for versions below that I was using `typing_extensions`.
Now here comes the kicker, after reinstalling the thing, which puller new MINOR version of pydantic v2 - which apparently included this: https://github.com/pydantic/pydantic/commit/41371ff32377f521b9d3396a7746e305bc677f35#diff-2ebfa892ec1d09440eefd4b22f488ec38c1cde365394e5a4cea6edb484827ba5R75 all "validations" starter throwing an error.
So while I know what to do on my side, I just would love to avoid such unpleasant surprises in the future.

I love pydantic, and I'm grateful for the hard work put into it, but if possible
1) please, as minimum, when introducing such breaking change mark them as such in CHANGELOG (ideal would be to use semver)
2) couldn't just requirement be just a recommendation? pydantic worked for on python3.11 just fine, so I don't see why the change couldn't just be in docs and let the code work if possible as it was and just let it error out. 

### Example Code

_No response_

```Text
pydantic version: 2.1.0
        pydantic-core version: 2.4.0

#I'm working on a library with multiple python versions between 3.8 and 3.12
```

Selected Assignee: @adriangb

adriangb:
Closes #6860

Selected Reviewer: @dmontagu

### DIFF
--- a/docs/usage/types/dicts_mapping.md
+++ b/docs/usage/types/dicts_mapping.md
@@ -58,7 +58,7 @@ except ValidationError as e:
 
 !!! note
     This is a new feature of the Python standard library as of Python 3.8.
-    Because of various limitations and bugs in the standard library implementations before 3.12, [typing-extensions](https://pypi.org/project/typing-extensions/) package is required for Python <3.12. You'll need to import `TypedDict` from `typing_extensions` instead of `typing` and will get a build time error if you don't.
+    Because of limitations in `typing.TypedDict` before 3.12, the [typing-extensions](https://pypi.org/project/typing-extensions/) package is required for Python <3.12. You'll need to import `TypedDict` from `typing_extensions` instead of `typing` and will get a build time error if you don't.
 
 [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) declares a dictionary type that expects all of
 its instances to have a certain set of keys, where each key is associated with a value of a consistent type.
--- a/docs/usage/types/dicts_mapping.md
+++ b/docs/usage/types/dicts_mapping.md
@@ -58,9 +58,7 @@ except ValidationError as e:
 
 !!! note
     This is a new feature of the Python standard library as of Python 3.8.
-    Prior to Python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
-    But required and optional fields are properly differentiated only since Python 3.9.
-    We therefore recommend using [typing-extensions](https://pypi.org/project/typing-extensions/) with Python 3.8 as well.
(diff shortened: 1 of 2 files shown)


docs/usage/types/dicts_mapping.md
```diff
@@ -58,9 +58,7 @@ except ValidationError as e:
 
 !!! note
     This is a new feature of the Python standard library as of Python 3.8.
-    Prior to Python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
-    But required and optional fields are properly differentiated only since Python 3.9.
-    We therefore recommend using [typing-extensions](https://pypi.org/project/typing-extensions/) with Python 3.8 as well.
+    Because of various limitations and bugs in the standard library implementations before 3.12, [typing-extensions](https://pypi.org/project/typing-extensions/) package is required for Python <3.12. You'll need to import `TypedDict` from `typing_extensions` instead of `typing` and will get a build time error if you don't.
```
davidhewitt, line 61:
```suggestion
    Because of limitations in `typing.TypedDict` before 3.12, the [typing-extensions](https://pypi.org/project/typing-extensions/) package is required for Python <3.12. You'll need to import `TypedDict` from `typing_extensions` instead of `typing` and will get a build time error if you don't.
```

docs/usage/types/types_fields.md
```diff
@@ -44,10 +44,7 @@ except ValidationError as e:
 
 !!! note
     This is a new feature of the Python standard library as of Python 3.8.
-    Prior to Python 3.8, it requires the [typing-extensions](https://pypi.org/project/typing-extensions/) package.
-    But required and optional fields are properly differentiated only since Python 3.9.
-    We therefore recommend using [typing-extensions](https://pypi.org/project/typing-extensions/) with Python 3.8 as well.
-
+    Because of various limitations and bugs in the standard library implementations before 3.12, [typing-extensions](https://pypi.org/project/typing-extensions/) package is required for Python <3.12. You'll need to import `TypedDict` from `typing_extensions` instead of `typing` and will get a build time error if you don't.
```
davidhewitt, line 47: I think we can get away with just saying "limitations" rather than bashing the stdlib too hard.

```suggestion
    Because of limitations in `typing.TypedDict` before 3.12, the [typing-extensions](https://pypi.org/project/typing-extensions/) package is required for Python <3.12. You'll need to import `TypedDict` from `typing_extensions` instead of `typing` and will get a build time error if you don't.
```

davidhewitt (commented):
Just a suggestion to make the wording a bit shorter...

adriangb:
please review

davidhewitt (approved):

PR #6902, opened 2023-07-27
[FIX] model validation for instances of BaseModel with `from_attributes=True`

### LINKED ISSUE #6901: [BUG] ValidationError not raised when validating a Pydantic model (with a missing required field)
### Initial Checks

### Description

Validating a Pydantic model directly using the model object does not work as expected: no ValidationError when a required field is missing.

See the two following examples, using `model_validate` and `TypeAdapter`.

See Pull Request with changes suggestions and tests : https://github.com/pydantic/pydantic/pull/6902

### Example Code

```Python
# 1. `model_construct` then `model_validate`

class Model(BaseModel):
    x: str


obj = Model()  # ValidationError is raised: required field is missing (as expected)
obj = Model.model_construct()
obj.model_validate(obj, from_attributes=True)  # ValidationError should be raised but is not


# 2. `TypeAdapter`


class Model(BaseModel):
    x: str

validator = TypeAdapter(Model)
validator.validate_python({})   # ValidationError is raised: required field is missing (as expected)
m = Model.model_construct()
validator.validate_python(m, from_attributes=True) # ValidationError should be raised but is not
```

```Text
pydantic version: 2.1.1
        pydantic-core version: 2.4.0
                 install path: /home/marie/Projects/pydantic/pydantic
               python version: 3.10.12 (main, Jun  7 2023, 12:45:35) [GCC 9.4.0]
                     platform: Linux-5.15.0-78-generic-x86_64-with-glibc2.31
```

Selected Assignee: @Kludex

marielejeune:
## Change Summary

Validating a Pydantic model directly using the model object does not work as expected when object is an instance of BaseModel and `from_attributes = True`

## Related issue number

fix #6901 

## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -493,9 +493,10 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
         """
         # `__tracebackhide__` tells pytest and some other tools to omit this function from tracebacks
         __tracebackhide__ = True
-        return cls.__pydantic_validator__.validate_python(
-            obj, strict=strict, from_attributes=from_attributes, context=context
-        )
+        kwargs = {'strict': strict, 'from_attributes': from_attributes, 'context': context}
+        if from_attributes and isinstance(obj, BaseModel):
+            kwargs['self_instance'] = obj
+        return cls.__pydantic_validator__.validate_python(obj, **kwargs)
 
     @classmethod
     def model_validate_json(
--- a/pydantic/type_adapter.py
+++ b/pydantic/type_adapter.py
@@ -205,7 +205,10 @@ class TypeAdapter(Generic[T]):
         Returns:
             The validated object.
         """
-        return self.validator.validate_python(__object, strict=strict, from_attributes=from_attributes, context=context)
+        kwargs = {'strict': strict, 'from_attributes': from_attributes, 'context': context}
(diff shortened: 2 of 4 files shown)


dmontagu:
`model_validate` is not meant to be used as an instance method, and should never modify the input object.

It seems this is just meant to address the case where you do an invalid `model_construct` as brought up in https://github.com/pydantic/pydantic/issues/6901, but:
* Using `model_construct` to create improperly-initialized models is not how it's meant to be used, and is not a workflow we want to accommodate
* As a rule, passing an input to `model_validate`/`validate_python` shouldn't modify that input
* This adds some overhead to `model_validate`/`validate_python` for _everyone_ to avoid issues caused by invalid `model_construct`; and given how performant we've made validation through the use of rust, I suspect this overhead is significant for some cases (such as `TypeAdapter(int)`, for example).

As a result, I think this PR is not likely to be merged.

dmontagu:
FastAPI will never create an invalid model from data, which is what it is meant to handle.

If you just don’t use `model_construct` in your code you won’t have this problem.

lmignon:
> FastAPI will never create an invalid model from data, which is what it is meant to handle.
> 
> If you just don’t use `model_construct` in your code you won’t have this problem.

@dmontagu Thank you for all your explanations.

Issue #6911, opened 2023-07-27
A model field with contents identical to a created object are not equal if a custom type (rootmodel) is used


RS-Credentive:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Please see the sample code below for the smallest use case I could develop. The code below uses the objects defined in the sample code.

Note that there are two models. One uses custom field types based on RootModels, and the other uses simple str types for model fields.

## Custom Field Types

I can create a standalone object using the Field type, and the same content, and it appears identical to the field of the model. However, they are not considered the same object by the interpreter. 

```py
Python 3.11.4 (main, Jun  7 2023, 19:03:45) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import properties
>>> # create an instance of the Property Object
>>> property = properties.Property()
>>> property
Property(ns=Uri(root=Url('http://test.com/URI')))
>>> property.ns
Uri(root=Url('http://test.com/URI'))
>>> # Now create a Uri object
>>> default_ns = properties.Uri('http://test.com/URI')
>>>
>>> # They look the same
>>> default_ns
Uri(root=Url('http://test.com/URI'))
>>>
>>> # However, they are not considered equal by the interpreter
>>>
>>> property.ns == default_ns
False
```

## Standard Field Types

```py
$ python
Python 3.11.4 (main, Jun  7 2023, 19:03:45) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import properties
>>> # create an instance of the Property Object
>>> property_notypes = properties.PropertyNoTypes()
>>> property_notypes
PropertyNoTypes(ns='http://test.com/URI')
>>> property_notypes.ns
'http://test.com/URI'
>>> # Now create a str object
>>> default_ns_notypes="http://test.com/URI"
>>> 
>>> # They also look the same
>>> default_ns_notypes
'http://test.com/URI'
>>> 
>>> # They are considered equal by the interpreter
>>> 
>>> property_notypes.ns == default_ns_notypes
True
```

## What I expected

I expected that a custom object would be considered equal to a field of the same type and content by the interpreter.

Is this a bug in Pydantic v2,  or an issue with the way I have defined the objects?

### Example Code

```Python
from __future__ import annotations

from pydantic import BaseModel, RootModel, Field, AnyUrl


class Uri(RootModel[AnyUrl]):
    root: AnyUrl = Field(
[...]

PR #6918, opened 2023-07-27
Fix misbehavior related to copying of RootModel

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6911

Technically we could include a check in the `__copy__` directly in `BaseModel`, but I think this is just a bit more performant since it removes the `if not self.__pydantic_root_model__:` branch in the method for both classes.

Can refactor to modify the `BaseModel` methods if preferred.

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/root_model.py
+++ b/pydantic/root_model.py
@@ -3,11 +3,12 @@
 from __future__ import annotations as _annotations
 
 import typing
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

hramezani (approved):

Issue #6922, opened 2023-07-27
python==3.9, pydantic>=2.1.0 regression in support for ParamSpec, RecursionError in model definition


ringohoffman:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

This bug does not appear in `pydantic==2.1.0` in python 3.8. This seems to be related to the introduction of `__instancecheck__` in python 3.9.

In `pydantic<=2.0.3` in python==3.9, there was no problem using a field that was a Generic w.r.t. a `ParamSpec`, but after `pydantic==2.1.0`, a `RecursionError` is now raised when defining a `BaseModel` with a field that uses a `ParamSpec` (or a `Generic` `TypeAlias` that uses a `ParamSpec`).

```
│ /Users/ringo/opt/anaconda3/envs/glass39/lib/python3.9/site-packages/pydantic/_internal/_generics │
│ .py:363 in <genexpr>                                                                             │
│                                                                                                  │
│   360 │   # Handle special case for typehints that can have lists as arguments.                  │
│   361 │   # `typing.Callable[[int, str], int]` is an example for this.                           │
│   362 │   if isinstance(type_, (List, list)):                                                    │
│ ❱ 363 │   │   if any(has_instance_in_type(element, isinstance_target) for element in type_):     │
│   364 │   │   │   return True                                                                    │
│   365 │                                                                                          │
│   366 │   return False                                                                           │
│                                                                                                  │
│ /Users/ringo/opt/anaconda3/envs/glass39/lib/python3.9/site-packages/pydantic/_internal/_generics │
│ .py:363 in has_instance_in_type                                                                  │
│                                                                                                  │
│   360 │   # Handle special case for typehints that can have lists as arguments.                  │
│   361 │   # `typing.Callable[[int, str], int]` is an example for this.                           │
│   362 │   if isinstance(type_, (List, list)):                                                    │
│ ❱ 363 │   │   if any(has_instance_in_type(element, isinstance_target) for element in type_):     │
│   364 │   │   │   return True                                                                    │
│   365 │                                                                                          │
│   366 │   return False                                                                           │
│                                                                                                  │
│ /Users/ringo/opt/anaconda3/envs/glass39/lib/python3.9/site-packages/pydantic/_internal/_generics │
│ .py:363 in <genexpr>                                                                             │
│                                                                                                  │
│   360 │   # Handle special case for typehints that can have lists as arguments.                  │
[...]

PR #6923, opened 2023-07-27
Fix issue with recursion error caused by ParamSpec

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6922

Selected Reviewer: @lig


dmontagu:
please review

PR #6926, opened 2023-07-28
Add test for v1/v2 Annotated discrepancy

carlbordum:
It makes sense that `model_dump` only dumps the actual fields on the model. This test is here to ensure the v1 behavior isn't accidentally introduced again.


dmontagu:
Thanks for the contribution

dmontagu (approved):

PR #6937, opened 2023-07-28
Make xfailing test for root model extra stop xfailing

dmontagu:
Fixing this xfailing test seemed straightforward.

Happy to change the PR to removing the xfailing test if we think it's better not to change this behavior. Or we can just close this, leave it unmodified, and kick the can down the road but I'd rather resolve this sooner if possible.

### DIFF
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -900,7 +900,7 @@ except PydanticUserError as exc_info:
     assert exc_info.code == 'invalid_annotated_type'
 ```
 
-## `config` is unused with TypeAdapter {#type-adapter-config-unused}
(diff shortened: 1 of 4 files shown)


docs/errors/usage_errors.md
```diff
@@ -882,7 +882,7 @@ except PydanticUserError as exc_info:
     assert exc_info.code == 'multiple-field-serializers'
 ```
 
-## Invalid annotated type {#invalid_annotated_type}
```
dmontagu, line 885: Also noticed this, maybe it's a bad idea to change this now (or maybe we need to add it to a redirect in the cloudflare worker) but this was the only one using snake- rather than kebab-case.

lig (approved):
👍🏻 Looks nice

PR #6951, opened 2023-07-30
Handle constraints being applied to schemas that don't accept it

adriangb:
This handles the particular case of `pattern` not working when using a custom `GenerateSchema` (reported in https://github.com/pydantic/pydantic/issues/6045#issuecomment-1652440886) but also tries to be more general to handle other similar issues.

### DIFF
--- a/pydantic/_internal/_known_annotated_metadata.py
+++ b/pydantic/_internal/_known_annotated_metadata.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
+from collections import defaultdict
 from copy import copy
 from functools import partial
 from typing import Any, Iterable
@@ -17,7 +18,7 @@ SEQUENCE_CONSTRAINTS = {'min_length', 'max_length'}
 INEQUALITY = {'le', 'ge', 'lt', 'gt'}
 NUMERIC_CONSTRAINTS = {'multiple_of', 'allow_inf_nan', *INEQUALITY}
 
-STR_CONSTRAINTS = {*SEQUENCE_CONSTRAINTS, *STRICT, 'strip_whitespace', 'to_lower', 'to_upper'}
+STR_CONSTRAINTS = {*SEQUENCE_CONSTRAINTS, *STRICT, 'strip_whitespace', 'to_lower', 'to_upper', 'pattern'}
 BYTES_CONSTRAINTS = {*SEQUENCE_CONSTRAINTS, *STRICT}
 
 LIST_CONSTRAINTS = {*SEQUENCE_CONSTRAINTS, *STRICT}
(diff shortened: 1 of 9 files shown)


adriangb:
Open questions:
- Should we try to error fast if we _know_ there's going to be an issue, e.g. `Annotated[int, Field(pattern=...)]`. The only way I see to do this is essentially special case by checking `constraint = 'pattern' and schema['type'] in ('int', 'float', etc.)`
- Can we get rid of the "merging" of `FieldInfo`? I think for validation yes, for JSON schema we may still need to merge them, but not even sure about that.
- Do we still want to make a more performant version of this in pydantic-core?

adriangb:
please review

adriangb:
This is now working and allowing constraint re-ordering. The cost is running things in Python when they can't be "inlined" into Rust. There's also a slight inconsistency I belive since when we _do_ "inline" it the order is not preserved (it's whatever the order of the if statements is in Rust). I guess that's fine.

tests/test_types.py
```diff
@@ -1675,25 +1675,6 @@ class Model(BaseModel):
...
-def test_invalid_schema_constraints(kwargs, type_):
-    match = (
-        r'(:?Invalid Schema:\n.*\n  Extra inputs are not permitted)|(:?The following constraints cannot be applied to)'
-    )
-    with pytest.raises((SchemaError, TypeError), match=match):
-
-        class Foo(BaseModel):
-            a: type_ = Field('foo', title='A title', description='A description', **kwargs)
-
-
```
dmontagu, line 1696:
```suggestion
@pytest.mark.parametrize(
    'kwargs,type_',
    [
        ({'gt': 0}, conlist(int, min_length=4)),
        ({'gt': 0}, conset(int, min_length=4)),
        ({'gt': 0}, confrozenset(int, min_length=4)),
    ],
)
def test_invalid_schema_constraints(kwargs, type_):
    match = (
        r'(:?Invalid Schema:\n.*\n  Extra inputs are not permitted)|(:?The following constraints cannot be applied to)'
    )
    with pytest.raises((SchemaError, TypeError), match=match):

        class Foo(BaseModel):
            a: type_ = Field('foo', title='A title', description='A description', **kwargs)
```

dmontagu (approved):

tests/test_types.py
adriangb: The reason why the pattern one doesn't give an error anymore and the others do is that those go down the `__prepare_pydantic_annotations__` path. One of the important bits of this PR is that we might be able to get rid of that in the future, but I guess it makes sense to leave the tests in for now.

adriangb:
These two give different results:

```python

from pydantic import BaseModel, Field, AfterValidator

from pydantic.type_adapter import TypeAdapter


Tp = Annotated[int, AfterValidator(lambda x: x * 2), Field(gt=3)]

class A(BaseModel):
    x: Tp

print(A.__pydantic_core_schema__['definitions'][0]['schema']['fields']['x']['schema'])

ta = TypeAdapter(Tp)
print(ta.core_schema)
```

It seems like this is because BaseModel merges field infos when it's constructed. We'd have to get rid of all field-info merging to get the right behavior w.r.t. ordering of constraints. I'm inclined to leave fixing that for a future PR. WDYT @dmontagu?

pydantic/_internal/_known_annotated_metadata.py
```diff
@@ -28,15 +29,57 @@
...
+for constraint in TIMEDELTA_CONSTRAINTS:
+    CONSTRAINTS_TO_ALLOWED_SCHEMAS[constraint].update(('timedelta',))
+for constraint in TIME_CONSTRAINTS:
+    CONSTRAINTS_TO_ALLOWED_SCHEMAS[constraint].update(('time',))
+for schema_type in (*TEXT_SCHEMA_TYPES, *SEQUENCE_SCHEMA_TYPES, *NUMERIC_SCHEMA_TYPES, 'typed-dict', 'model'):
+    CONSTRAINTS_TO_ALLOWED_SCHEMAS['strict'].add(schema_type)
+for constraint in URL_CONSTRAINTS:
+    CONSTRAINTS_TO_ALLOWED_SCHEMAS[constraint].update(('url', 'multi-host-url'))
+for constraint in BOOL_CONSTRAINTS:
+    CONSTRAINTS_TO_ALLOWED_SCHEMAS[constraint].update(('bool',))
```
samuelcolvin, line 81: does this get called on `from pydantic import X`? If so we might need to worry about import time, if it's later/when require it shouldn't be a problem.

pydantic/_internal/_known_annotated_metadata.py
```diff
@@ -96,82 +139,150 @@ def apply_known_metadata(annotation: Any, schema: CoreSchema) -> CoreSchema | No
...
+                schema,
+            )
+        elif constraint == 'pattern':
+            # insert a str schema to make sure the regex engine matches
+            return cs.chain_schema(
+                [
+                    schema,
+                    cs.str_schema(pattern=value),
+                ]
+            )
```
samuelcolvin, line 165: these can probably be made more compact thereby making the function easier to read.

samuelcolvin (commented):
otherwise LGTM.

samuelcolvin:
please update.

pydantic/_internal/_known_annotated_metadata.py
adriangb: black controls formatting not me 😄

pydantic/_internal/_known_annotated_metadata.py
adriangb: I believe it is imported immediately. It's not all that complex though, just assigning times to a dict. I could hide it behind an lru_cache function or something, but that would probably make schema generation slower (one more dict lookup), and that seems to be a bigger problem right now than import time

pydantic/_internal/_validators.py
```diff
@@ -269,3 +270,12 @@ def max_length_validator(x: Any, max_length: Any) -> Any:
...
         )
     return x
+
+
+def forbid_inf_nan_check(x: Any) -> Any:
+    if not math.isfinite(x):
+        raise PydanticKnownError(
+            'finite_number',
+            {},
+        )
```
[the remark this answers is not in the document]
adriangb: yep thanks

adriangb:
please review

tests/test_types.py
adriangb: I'll add this but just to be clear it's more documenting current behavior than testing the desired / future behavior

adriangb:
please review

tests/test_serialize.py
```diff
@@ -896,7 +896,7 @@ class UserLogin(User):
         password: SecretStr
 
     class OuterModel(BaseModel):
-        maybe_as_any: Optional[SerializeAsAny[User]]
+        maybe_as_any: Optional[SerializeAsAny[User]] = None
```
adriangb, line 899: Not strictly necessary for this PR but the insert_assert() below doesn't work without it

pydantic/_internal/_validators.py
```diff
@@ -269,3 +270,9 @@ def max_length_validator(x: Any, max_length: Any) -> Any:
             {'field_type': 'Value', 'max_length': max_length, 'actual_length': len(x)},
         )
     return x
+
+
+def forbid_inf_nan_check(x: Any) -> Any:
+    if not math.isfinite(x):
+        raise PydanticKnownError('finite_number')
+    return x
```
adriangb, line 278: Added a test for the error case here

pydantic/_internal/_known_annotated_metadata.py
```diff
@@ -96,82 +139,150 @@ def apply_known_metadata(annotation: Any, schema: CoreSchema) -> CoreSchema | No
...
+            predicate_name = f'{annotation.func.__qualname__} ' if hasattr(annotation.func, '__qualname__') else ''
+
+            def val_func(v: Any) -> Any:
+                # annotation.func may also raise an exception, let it pass through
+                if not annotation.func(v):
+                    raise PydanticCustomError(
+                        'predicate_failed',
+                        f'Predicate {predicate_name} failed',  # type: ignore
+                        {},
+                    )
```
adriangb, line 281:
```suggestion
                    raise PydanticCustomError(
                        'predicate_failed',
                        f'Predicate {predicate_name} failed',  # type: ignore
                    )
```

dmontagu (approved):

PR #6970, opened 2023-07-31
Fix bug in dataclass fields using Annotated

### LINKED ISSUE #6962: `BeforeValidator` runs twice in case when dataclass
### Initial Checks

### Description

`BeforeValidator` runs twice in case when dataclass.

This is inconvinient when validate input value type before convert like #6045
Is there any way to avoid this issue?

### Example Code

`test.py`:

```Python
from typing import Annotated

from pydantic import BeforeValidator


def convert(value: int) -> str:
    print(f"{value=}")
    return str(value)


IntToStr = Annotated[str, BeforeValidator(convert)]


@dataclass
class A:
    a: IntToStr


A(123)
```

Result:

```console
$ python test.py
value=123
value='123'
```

```Text
$ python -c "import pydantic.version; print(pydantic.version.version_info())"
             pydantic version: 2.1.1
        pydantic-core version: 2.4.0
                 install path: /workspace/.venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.4 (main, Jul  4 2023, 05:25:16) [GCC 12.2.0]
                     platform: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.36
```

adriangb:
Fixes #6962

### DIFF
--- a/pydantic/_internal/_dataclasses.py
+++ b/pydantic/_internal/_dataclasses.py
@@ -270,4 +270,4 @@ def is_pydantic_dataclass(_cls: type[Any]) -> TypeGuard[type[PydanticDataclass]]
     Returns:
         `True` if the class is a pydantic dataclass, `False` otherwise.
     """
-    return dataclasses.is_dataclass(_cls) and '__pydantic_validator__' in _cls.__dict__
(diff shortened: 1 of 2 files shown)


adriangb:
please review

davidhewitt:
@adriangb I think #6998 may be a better fix? 😄

PR #6988, opened 2023-08-02
Fix generic computed fields

dmontagu:
Fix an issue where computed fields with a generic return type would not reflect the parametrized value of the generic parameter.


dmontagu:
please review

adriangb (approved):

PR #7026, opened 2023-08-07
remove unused "type ignores" with pyright

samuelcolvin:
I've wanted to do this since #5399 when we switched from mypy to pyright.

This will conflict with #7024 but should be simple to fix.

Selected reviewer: @adriangb


adriangb (approved):

adriangb:
looks good but some test failures, guessing some of those are needed. please fix then merge

PR #7028, opened 2023-08-07
tweak `CoreWalk` for marginal performance improvement

samuelcolvin:
Makes a small difference and I think it "more correct", but the time difference isn't noticable so if you disagree, close.

Selected reviewer: @dmontagu


samuelcolvin:
please review.

samuelcolvin:
not helping.

Issue #7045, opened 2023-08-08
Wrong JSON Schema generation with sequence like enums


Viicos:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

When using Enums with complex values, the generated JSON Schema is incorrect (e.g. [this validator](https://www.jsonschemavalidator.net/) raises an error).

By the way, I'm trying to find a good way of serializing such enums (by their name), similar to https://github.com/pydantic/pydantic/discussions/6466, but imho it it a bit cumbersome for a use case that could be pretty common. Do you think a helper could be integrated into Pydantic?

A (weird) alternative implementation:

```python
from typing import Annotated

from pydantic import BaseModel, PlainSerializer, WrapValidator

def serialize_with_name(EnumClass: type[Enum]) -> Enum:
    return Annotated[
        EnumClass,
        WrapValidator(lambda v, handler: EnumClass[v] if isinstance(v, str) else v),
        PlainSerializer(lambda el: el.name, return_type=Literal[EnumClass._member_names_], when_used="json"),   # The Literal is still valid Python
    ]

class Model(BaseModel):
    direction: serialize_with_name(Direction)
```

Related issue: https://github.com/pydantic/pydantic/issues/3174

### Example Code

```Python
from enum import Enum

from pydantic import BaseModel

class Direction(Enum):
    NORTH      = (0, 1)
    NORTH_EAST = (1, 1)
    EAST       = (1, 0)
    SOUTH_EAST = (1, -1)
    SOUTH      = (0, -1)
    SOUTH_WEST = (-1, -1)
    WEST       = (-1, 0)
    NORTH_WEST = (-1, 1)


class Model(BaseModel):
    direction: Direction

Model.model_json_schema()
#> {'$defs': {'Direction': {'enum': [(0, 1), (1, 1), (1, 0), (1, -1), (0, -1), (-1, -1), (-1, 0), (-1, 1)], 'title': 'Direction'}}, 'properties': {'direction': {'$ref': '#/$defs/Direction'}}, 'required': ['direction'], 'title': 'Model', 'type': 'object'}
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.1.1
pydantic-core version: 2.4.0
pydantic-core build: profile=release pgo=true mimalloc=true
python version: 3.10.9 (main, Jan 13 2023, 19:08:27) [GCC 11.3.0]
```

dmontagu:
[...]

PR #7056, opened 2023-08-09
Handle non-json native enum values

adriangb:
Fixes https://github.com/pydantic/pydantic/issues/7045

This isn't perfect, e.g. ideally if the value is always a two tuple of ints it would be reflected in the generated schema. But this is better than invalid json.

### DIFF
--- a/tests/test_json_schema.py
+++ b/tests/test_json_schema.py
@@ -5384,3 +5384,16 @@ def test_callable_json_schema_extra_dataclass():
     json_schema = adapter.json_schema()
     for key in 'abcdef':
         assert json_schema['properties'][key] == {'title': key.upper(), 'type': 'integer'}  # default is not present
+
+
+def test_enum_complex_value() -> None:
+    """https://github.com/pydantic/pydantic/issues/7045"""
+
+    class MyEnum(Enum):
+        foo = (1, 2)
+        bar = (2, 3)
+
+    ta = TypeAdapter(MyEnum)
+
+    # insert_assert(ta.json_schema())
+    assert ta.json_schema() == {'enum': [[1, 2], [2, 3]], 'title': 'MyEnum', 'type': 'array'}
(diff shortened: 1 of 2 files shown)


adriangb:
please review

adriangb:
I guess so! Better than generating invalid json and having something downstream crash right?

tests/test_json_schema.py
```diff
@@ -5384,3 +5384,16 @@ class MyDataclass:
...
+    """https://github.com/pydantic/pydantic/issues/7045"""
+
+    class MyEnum(Enum):
+        foo = (1, 2)
+        bar = (2, 3)
+
+    ta = TypeAdapter(MyEnum)
+
+    # insert_assert(ta.json_schema())
+    assert ta.json_schema() == {'enum': [[1, 2], [2, 3]], 'title': 'MyEnum', 'type': 'array'}
```
Viicos, line 5399: This would still fail to parse with JSON data matching this schema:

```
TypeAdapter(Direction).validate_json('[0,1]')
#> Input should be (0, 1),(1, 1),(1, 0),(1, -1),(0, -1),(-1, -1),(-1, 0) or (-1, 1) [type=enum, input_value=[0, 1], input_type=list]
```
But serializing complex enums by their value isn't a good idea anyway, as I mentioned in https://github.com/pydantic/pydantic/issues/7045

Viicos:
> I guess so! Better than generating invalid json and having something downstream crash right?

True! Just wanted to be sure pydantic "allowed" exceptions to be raised in this `model_json_schema`  method

tests/test_json_schema.py
adriangb: Yup not much we can do about that

dmontagu:
I think it should work to use the `WithJsonSchema` annotation if you want/need to explicitly specify something, for what it's worth.

tests/test_json_schema.py
dmontagu: I think it would work with an appropriate validator, perhaps annoying that you have to write it, but I think given how rare this is it's not unreasonable, and I think this change to the JSON schema generation is an improvement either way.

dmontagu (approved):

PR #7103, opened 2023-08-12
Do not require `validate_assignment` to use `Field.frozen`

Viicos:
## Change Summary

I find it a bit unusual to have the following working:

```python

class Model(BaseModel):
    a: int = 1
    model_config = {"frozen": True}

m = Model()
m.a = 2
#> Raises ValidationError
```

But not the following:

```python

class Model(BaseModel):
    a: int = Field(frozen=True)

m = Model(a=1)
m.a = 2
#> No ValidationError, unless model_config.validate_assignment is set to True
```

If you think the current behavior is as designed, I can revert my commit but I think a note on the `Field.frozen` docstring would be required.
## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -748,7 +748,7 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
                 'input': value,
             }
             raise pydantic_core.ValidationError.from_exception_data(self.__class__.__name__, [error])
-        elif name in self.model_fields and self.model_fields[name].frozen:
+        elif getattr(self.model_fields.get(name), "frozen", False)
             error: pydantic_core.InitErrorDetails = {
                 'type': 'frozen_field',
                 'loc': (name,),
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -748,7 +748,7 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
                 'input': value,
             }
             raise pydantic_core.ValidationError.from_exception_data(self.__class__.__name__, [error])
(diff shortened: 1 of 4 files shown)


hramezani:
It is intended behavior. The second example will work if `config['validate_assignment'] = True`. You can find it in the [example in doc](https://docs.pydantic.dev/latest/usage/fields/).

Probably we can improve the the docstring and documentation here

Viicos:
Then should the first example I gave follows the same idea (requiring `validate_assignment=True`)? As I feel this is a bit inconsistent

pydantic/main.py
```diff
@@ -748,6 +748,13 @@ def __setattr__(self, name: str, value: Any) -> None:
                 'input': value,
             }
             raise pydantic_core.ValidationError.from_exception_data(self.__class__.__name__, [error])
+        elif name in self.model_fields and self.model_fields[name].frozen:
```
samuelcolvin, line 751: let's not look up `self.model_fields` twice.

samuelcolvin (changes requested):
otherwise LGTM

pydantic/main.py
Viicos:
```suggestion
        elif getattr(self.model_fields.get(name), "frozen", False)
```

Would this be ok? Or else I need to make sure `name` is part of `self.model_fields`

Viicos:
Please review

hramezani (approved):

samuelcolvin (approved):
LGTM, thanks so much for the contribution.

PR #7115, opened 2023-08-14
Add `short_version` and use it in links

hramezani:


pydantic/version.py
```diff
@@ -7,6 +7,14 @@
 """The version of Pydantic."""
 
 
+def version_short() -> str:
```
hramezani, line 10: We can replace it with a hardcoded short version or cached one

samuelcolvin (approved):
LGTM

Issue #6980, opened 2023-08-01
Incorrect detection of generics with nested classes


M0dEx:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

When nesting generic classes, one of which is a subclass of `BaseModel` (as shown in the example), Pydantic incorrectly raises a `TypeError` with the explanation that one of the classes is not generic:
```
Traceback (most recent call last):
  File "/Users/***/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/232.8660.197/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
           ^^^^^^
  File "<input>", line 12, in <module>
  File "/Users/***/Library/Caches/pypoetry/virtualenvs/***/lib/python3.11/site-packages/pydantic/main.py", line 607, in __class_getitem__
    raise TypeError(f'{cls} is not a generic class')
TypeError: <class '__main__.BrokenB'> is not a generic class
```

According to [MyPy docs](https://mypy.readthedocs.io/en/stable/generics.html#defining-subclasses-of-generic-classes), the example should be valid (and without inheriting `BaseModel` works just fine.)

There is a workaround - by specifying the `Generic` keyword with the same `TypeVar` when inheriting the base class (`A` in the example code below), the check inside Pydantic succeeds.

### Example Code

```Python
from typing import Generic, TypeVar

from pydantic import BaseModel

T = TypeVar("T")


class A(BaseModel, Generic[T]):
    a: T
    

class BrokenB(A[T]):
    b: T


class BrokenC(BrokenB[int]):  # FAILS
    pass


class FixedB(A[T], Generic[T]):
    b: T


class FixedC(FixedB[int]):  # WORKS
    pass
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.1.1
pydantic-core version: 2.4.0
pydantic-core build: profile=release pgo=false mimalloc=true
install path: /Users/***/Library/Caches/pypoetry/virtualenvs/***-py3.11/lib/python3.11/site-packages/pydantic
python version: 3.11.2 (main, Mar 21 2023, 13:41:33) [Clang 14.0.0 (clang-1400.0.29.202)]
platform: macOS-13.5-x86_64-i386-64bit
optional deps. installed: ['typing-extensions']
```

Selected Assignee: @dmontagu

dmontagu:
[...]

PR #7119, opened 2023-08-14
Don't always require typing.Generic as a base for partially parametrized models

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/6980

I won't be surprised if there are still cases that aren't properly handled by this change, but this resolves the reported issue, and given the complexity of the generics stuff I think our best bet in the long run is to just rely on converting reported issues into tests, and making sure all tests pass.

Selected Reviewer: @adriangb


dmontagu:
please review

adriangb (approved):
Looks good I remember I wanted this but couldn't figure out how to do it, good stuff!

Issue #7134, opened 2023-08-16
NamedTuple default values are ignored in validation


ohad-weka:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

A validation for a named-tuple that has defaults still demands supplying all fields. 

This behavior seems to be similar to what v1 `parse_obj_as` does in this case.

### Example Code

```Python
from typing import NamedTuple

from pydantic import TypeAdapter


class NT(NamedTuple):
    x: int
    y: int = 33


# Fails - field y missing required argument
TypeAdapter(NT).validate_python([1])
# Fails - field y missing required argument
TypeAdapter(NT).validate_python({"x": 22})

# Ok
TypeAdapter(NT).validate_python({"x": 22, "y": 33})
# Ok
TypeAdapter(NT).validate_python([22, 33])
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.1.1
        pydantic-core version: 2.4.0
          pydantic-core build: profile=release pgo=false mimalloc=true
                 install path: /Users/ohad.shemesh/.pyenv/versions/3.7.10/envs/teka_5ff5dcadf4a78469da90f47fcec27a25/lib/python3.7/site-packages/pydantic
               python version: 3.7.10 (default, Jun 27 2022, 12:52:30)  [Clang 13.1.6 (clang-1316.0.21.2.5)]
                     platform: Darwin-22.3.0-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @Kludex

PR #7144, opened 2023-08-16
Reflect namedtuple default values during validation

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/7134

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1130,7 +1130,9 @@ class GenerateSchema:
 
             arguments_schema = core_schema.arguments_schema(
                 [
-                    self._generate_parameter_schema(field_name, annotation)
+                    self._generate_parameter_schema(
+                        field_name, annotation, default=namedtuple_cls._field_defaults.get(field_name, Parameter.empty)
+                    )
                     for field_name, annotation in annotations.items()
                 ],
                 metadata=build_metadata_dict(js_prefer_positional_arguments=True),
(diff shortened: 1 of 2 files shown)


dmontagu:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -1130,7 +1130,9 @@ def _namedtuple_schema(self, namedtuple_cls: Any, origin: Any) -> core_schema.Co
 
             arguments_schema = core_schema.arguments_schema(
                 [
-                    self._generate_parameter_schema(field_name, annotation)
+                    self._generate_parameter_schema(
+                        field_name, annotation, default=namedtuple_cls._field_defaults.get(field_name, Parameter.empty)
```
dmontagu, line 1134: Though this looks like a private attribute access, this `_field_defaults` attribute is explicitly present in the type stubs for `NamedTuple` — I think it's intended for public access, just has an underscore to prevent name collisions.

dmontagu: https://github.com/python/typeshed/blob/695d41f4870aede6e7027cf23bb65b07b59eedc6/stdlib/typing.pyi#L860 for reference

Viicos: And its documented: https://docs.python.org/3/library/collections.html#collections.somenamedtuple._field_defaults

adriangb (approved):

PR #7150, opened 2023-08-16
Update dependencies, fix pydantic-core usage, fix CI issues

dmontagu:
* Update pydantic-core to 2.6.0
	* The main changes in pydantic required were changes to handling of union schemas to handle the case where the choices field might have labels. I have not added tests for this, just gotten the code to type-check. I think we can add tests when we expose this functionality into pydantic somehow.
* Update all other dependencies (I just ran `make refresh-lockfiles`). This included `pyright` and `ruff`, which resulted in needing to change some stuff for type-checking/linting.
	* While addressing updates to pyright's type checking, I noticed that we no longer need to check for indirect schema references in the tagged union schema, so I removed that code.
* Fix the pydantic test that has been causing the pydantic-core CI to fail due to updated type serialization. (It was failing because it was checking that two instances of type `str` that were equal were actually the same object in memory, which clearly isn't guaranteed. It wasn't unreasonable to want them to be the same, because they were previously both the exact str instance `FooEnum.FOO.value`, but anyway I don't think it should be required and I changed the test to just check the types were exactly the same and the values were equal).

Selected Reviewer: @lig

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,7 +227,7 @@ jobs:
         node-version: '18'
 
     - name: install pyright
-      run: npm install -g pyright@1.1.302  # try to keep this in sync with .pre-commit-config.yaml
+      run: npm install -g pyright@1.1.322  # try to keep this in sync with .pre-commit-config.yaml
 
     - name: run pyright tests
       run: make test-pyright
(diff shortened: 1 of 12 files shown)


dmontagu:
please review

adriangb (approved):

davidhewitt (approved):

Issue #7097, opened 2023-08-12
The "preserve the type of the input whenever possible" union handling is inconsistent


stevegoss-komodo:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Code that relies on the strict left-to-right behavior of union validation/parsing is broken in V2, with no available workaround. Not only that, but the subtle change creates inconsistencies where identical data in python or JSON form parses differently. In the example code, all of those would have parsed as UUIDs in V1 (behavior that our code relies on), but now, it parses as string in certain scenarios, but not in others.

It is difficult to migrate with this highly inconsistent and frankly baffling behavior. Using unions like the one in the example with FastAPI as path parameters or query parameters used to be a useful way to detect UUIDs, but now, FastAPI always returns a string.

### Example Code

```Python
from uuid import UUID

from pydantic import TypeAdapter, BaseModel

IdOrSlug = UUID | str

print(type(TypeAdapter(IdOrSlug).validate_json("\"f4fe10b4-e0c8-4232-ba26-4acd491c2414\"")))
# <class 'uuid.UUID'>

print(type(TypeAdapter(IdOrSlug).validate_python("f4fe10b4-e0c8-4232-ba26-4acd491c2414")))
# <class 'str'>

class AModel(BaseModel):
    my_id: IdOrSlug

model = AModel.model_validate_json("""{"my_id":"f4fe10b4-e0c8-4232-ba26-4acd491c2414"}""")
print(type(model.my_id))
# <class 'uuid.UUID'>

model1 = AModel.model_validate({"my_id":"f4fe10b4-e0c8-4232-ba26-4acd491c2414"})
print(type(model1.my_id))
# <class 'str'>
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.1.1
        pydantic-core version: 2.4.0
          pydantic-core build: profile=release pgo=false mimalloc=true
                 install path: /Users/redacted/projects/platform-orchestration/.venv/lib/python3.10/site-packages/pydantic
               python version: 3.10.10 (main, May  8 2023, 10:19:34) [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.4.1-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

samuelcolvin:
I agree with @davidhewitt, but I would suggest adding `dumb_union` (default `False`) since:
1. Better to have `False` as the default
[...]

PR #7151, opened 2023-08-16
add union_mode='left_to_right'

davidhewitt:
## Change Summary

Adds a new `Field` option, `union_mode`, which can be used to alter the union validation. At present, there are only two options:
 - `smart` (the default), which is equivalent to the current V2 behaviour
 - `left_to_right`, which is just simple left-to-right validation, first success wins, like V1

This option can be used to opt out of smart union validation in cases when users want more easily understandable results.

In https://github.com/pydantic/pydantic/issues/7097#issuecomment-1677134952 we suggested a `dumb_union=True` boolean flag. In further discussion offline I raised the idea that using a string setting here (while a touch more verbose) is more descriptive and also gives us space to add more options backwards-compatibly later.

## Related issue number

Related #7097
Related #7110

## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -506,6 +506,9 @@ print(Model(x='1'))
 In Pydantic V1, the printed result would have been `x=1`, since the value would pass validation as an `int`.
 In Pydantic V2, we recognize that the value is an instance of one of the cases and short-circuit the standard union validation.
 
+To revert to the non-short-circuiting left-to-right behavior of V1, annotate the union with `Field(union_mode='left_to_right')`.
+See [Union Mode](./usage/types/unions.md#union-mode) for more details.
+
 #### Required, optional, and nullable fields
 
 Pydantic V2 changes some of the logic for specifying whether a field annotated as `Optional` is required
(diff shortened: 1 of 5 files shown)


docs/migration.md
```diff
@@ -506,6 +506,9 @@ print(Model(x='1'))
 In Pydantic V1, the printed result would have been `x=1`, since the value would pass validation as an `int`.
 In Pydantic V2, we recognize that the value is an instance of one of the cases and short-circuit the standard union validation.
 
+To revert to the non-short-circuiting left-to-right behavior of V1, annotate the union with `Field(union_mode='left_to_right')`.
+See [Union Mode](./usage/types/unions.md#union-mode) for more details.
```
davidhewitt, line 510: I'm not an expert on V1's union validation, this statement may not be true. I thought it might be worth calling out this or something similar here in the migration guide.

dmontagu: It's right

dmontagu (approved):

stevegoss-komodo:
@davidhewitt Thank you! This will greatly ease our transition to Pydantic V2!

wooparadog:
Hi, just found this gem while migrating to V2. Is there a way to make this behavior as default..?

Issue #7152, opened 2023-08-16
Pyright type error with @model_validator(mode="wrap")


object-Object:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Annotating a method with `@model_validator(mode="wrap")` results in the following type errors from Pyright.
```
wrap.py:7:6 - error: Argument of type "(cls: type[Self@Model], value: Any, handler: ModelWrapValidatorHandler[Self@Model], info: ValidationInfo) -> Self@Model" cannot be assigned to parameter of type "_AnyModelWrapValidator"
    Type "(cls: type[Self@Model], value: Any, handler: ModelWrapValidatorHandler[Self@Model], info: ValidationInfo) -> Self@Model" cannot be assigned to type "_AnyModelWrapValidator"
      Type "(cls: type[Self@Model], value: Any, handler: ModelWrapValidatorHandler[Self@Model], info: ValidationInfo) -> Self@Model" cannot be assigned to type "(cls: type[_ModelType@__call__], __value: Any, __handler: ModelWrapValidatorHandler[_ModelType@__call__], __info: ValidationInfo) -> _ModelType@__call__"
        Parameter 1: type "type[_ModelType@__call__]" cannot be assigned to type "type[Self@Model]"
          "type*" is incompatible with "Model"
        Parameter 3: type "ModelWrapValidatorHandler[_ModelType@__call__]" cannot be assigned to type "ModelWrapValidatorHandler[Self@Model]"
          "ModelWrapValidatorHandler[_ModelType@__call__]" is incompatible with "ModelWrapValidatorHandler[Self@Model]"
            Type parameter "_ModelTypeCo@ModelWrapValidatorHandler" is covariant, but "_ModelType@__call__" is not a subtype of "Self@Model"
        Function return type "Self@Model" is incompatible with type "_ModelType@__call__" (reportGeneralTypeIssues)
```

Initially I [opened an issue](https://github.com/microsoft/pyright/issues/5735) in the Pyright repo, but as discussed there it seems this error is actually correct for the below code, though it's entirely possible that I'm just using it wrong.

### Example Code

```Python
from typing import Any, Self
from pydantic import ValidationInfo
from pydantic import BaseModel, model_validator
from pydantic.functional_validators import ModelWrapValidatorHandler

class Model(BaseModel):
    @model_validator(mode="wrap")
    @classmethod
[...]

PR #7154, opened 2023-08-16
Make ModelWrapValidator protocols generic

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/7152

@samuelcolvin @adriangb is there any good way to add a test for this? Do we care enough to bother?

### DIFF
--- a/pydantic/functional_validators.py
+++ b/pydantic/functional_validators.py
@@ -351,7 +351,7 @@ class ModelWrapValidatorHandler(_core_schema.ValidatorFunctionWrapHandler, Proto
         ...
 
 
-class ModelWrapValidatorWithoutInfo(Protocol):
+class ModelWrapValidatorWithoutInfo(Protocol[_ModelType]):
     """A @model_validator decorated function signature.
     This is used when `mode='wrap'` and the function does not have info argument.
     """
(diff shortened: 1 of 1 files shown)


adriangb (approved):

pydantic/functional_validators.py
```diff
@@ -423,7 +423,7 @@ def __call__(  # noqa: D102
 ModelAfterValidator = Callable[[_ModelType, _core_schema.ValidationInfo], _ModelType]
 """A `@model_validator` decorated function signature. This is used when `mode='after'`."""
 
-_AnyModelWrapValidator = Union[ModelWrapValidator, ModelWrapValidatorWithoutInfo]
+_AnyModelWrapValidator = Union[ModelWrapValidator[_ModelType], ModelWrapValidatorWithoutInfo[_ModelType]]
```
adriangb, line 426: Doesn't this need to be parametrized later?

adriangb:
Unfortunately adding tests for these things is not easy. If the original example now works locally with mypy and pyright let's merge this.

pydantic/functional_validators.py
object-Object: From my testing, adding the `_ModelType` parameter to the overload return type on line 435 seems to give the right results for a correct and incorrect validator (and the `after` overload does the same thing already):
```py
) -> Callable[[_AnyModelWrapValidator[_ModelType]], _decorators.PydanticDescriptorProxy[_decorators.ModelValidatorDecoratorInfo]]:
```
That appears to be the only place where these types are used.

samuelcolvin (approved):
I'd agree with @adriangb.

LGTM.

PR #7155, opened 2023-08-16
Add ability to get namespace from module in TypeAdapter

dmontagu:
Related to https://github.com/pydantic/pydantic/issues/7111, but in its current form this PR cannot/should not be merged.

The problem with this approach is that the input to the `TypeAdapter` may be in a different module. The most obvious case of that is something like:
```python
from typing import List

TypeAdapter(List[forward.Model])
```

I do think it's a good idea to add the ability to provide a types namespace when creating a type adapter, which would at least make it _possible_ to resolve the issue in fastapi (e.g., by having fastapi capture the namespace in the initial call to the decorator). But I think, unlike what this PR currently does, we should probably not try to infer the module automatically, and should require it be explicitly specified.

@adriangb is well aware of this issue and may have opinions about how to proceed.

### DIFF
--- a/pydantic/type_adapter.py
+++ b/pydantic/type_adapter.py
@@ -28,7 +28,9 @@ if TYPE_CHECKING:
     IncEx = Union[Set[int], Set[str], Dict[int, Any], Dict[str, Any]]
 
 
-def _get_schema(type_: Any, config_wrapper: _config.ConfigWrapper, parent_depth: int) -> CoreSchema:
+def _get_schema(
+    type_: Any, config_wrapper: _config.ConfigWrapper, parent_depth: int, types_namespace: dict[str, Any] | None = None
+) -> CoreSchema:
     """`BaseModel` uses its own `__module__` to find out where it was defined
     and then look for symbols to resolve forward references in those globals.
     On the other hand this function can be called with arbitrary objects,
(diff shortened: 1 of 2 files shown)


pydantic/type_adapter.py
```diff
@@ -164,7 +167,12 @@ def __init__(self, type: Any, *, config: ConfigDict | None = None, _parent_depth
         try:
             core_schema = _getattr_no_parents(type, '__pydantic_core_schema__')
         except AttributeError:
-            core_schema = _get_schema(type, config_wrapper, parent_depth=_parent_depth + 1)
+            types_namespace = None
+            if hasattr(type, '__module__'):
+                types_namespace = sys.modules[type.__module__].__dict__
+            core_schema = _get_schema(
+                type, config_wrapper, parent_depth=_parent_depth + 1, types_namespace=types_namespace
+            )
```
dmontagu, line 175: I think if we just make `types_namespace` an argument to the `TypeAdapter.__init__` that might be enough for users/libraries to pick up the baton. In particular, I think it would be enough that FastAPI could resolve the issue by taking responsibility for collecting the namespace when the route decorators are called.

I think it might make sense to just add that argument to the `__init__` and merge this, though it wouldn't resolve the reported issue without more tweaks. But I think this change doesn't actually _resolve_ the reported issue anyway, just addresses one instance of it.

adriangb:
I'll also note that in theory this will be a non-issue for future python versions that implement PEP 649

dmontagu:
I don't think I'm going to be able to see this across the line any time soon, I suspect trying to get this right could lead to introducing new and painful-to-resolve bugs, and as @adriangb noted it will be massively simplified by PEP649, so I think probably best to abandon this effort and focus on other approaches that could resolve the FastAPI issue..

PR #7184, opened 2023-08-20
fix: Allow inheriting dataclasses with defaults

### LINKED ISSUE #6984: Allow inheriting `dataclasses.dataclass` with defaults
### Initial Checks

### Description

Currently, inheriting a pure dataclass to a subclass that is also a `BaseModel` works. This is great, and this feature has been used extensively in v1 in a code base i maintain. v2 seems to have the added feature of not requiring to re-declare the fields, which is awesome.

However, if the dataclass defines a default for a field, it crashes with a `NameError`

```console
...
  File "<input>", line 19, in <module>
  File ".../lib/python3.10/site-packages/pydantic/_internal/_model_construction.py", line 172, in __new__
    set_model_fields(cls, bases, config_wrapper, types_namespace)
  File ".../lib/python3.10/site-packages/pydantic/_internal/_model_construction.py", line 400, in set_model_fields
    fields, class_vars = collect_model_fields(cls, bases, config_wrapper, types_namespace, typevars_map=typevars_map)
  File ".../lib/python3.10/site-packages/pydantic/_internal/_fields.py", line 155, in collect_model_fields
    raise NameError(
NameError: Field name "a" shadows an attribute in parent "A"; you might want to use a different field name with "alias='a'".
```

### Example Code

```Python
import pydantic

import pydantic


@dataclass
class A:
    a: None

class B(pydantic.BaseModel, A): # Ok
    pass


@dataclass
class A:
    a: None = None
    

class B(pydantic.BaseModel, A): # Not ok
    pass
```

```Text
pydantic version: 2.1.1
        pydantic-core version: 2.4.0
                 install path: .../lib/python3.10/site-packages/pydantic
               python version: 3.10.2 (main, Mar  3 2022, 12:03:29) [GCC 9.3.0]
                     platform: Linux-5.14.0-1054-oem-x86_64-with-glibc2.35
```

Selected Assignee: @lig

tobni:
## Change Summary

This patch allows for a `pydantic.Basemodel` class to inherit a "pure" stdlib dataclass, even when the attribute name is set on the class instance, as is the case with `dataclasses.field` and default assignments.

## Related issue number

fix #6984

## Checklist

* [ ] Tests pass on CI


tobni:
please review

tobni:
Looks like https://github.com/pydantic/pydantic/pull/7193 addressed the immediate issue. I am however keen on letting the test and extra logic to avoid the warning remain to ensure compat.

tests/test_dataclasses.py
```diff
@@ -2487,3 +2487,25 @@ class A:
...
+
+def test_can_inherit_stdlib_dataclasses_with_dataclass_fields():
+    @dataclasses.dataclass
+    class Base:
+        a: int = dataclasses.field(default=5)
+
+    class Model(BaseModel, Base):
+        pass
+
+    assert Model().a == 5
```
adriangb, line 2511: Can we also add a test for `default_factory`? I'm assuming it doesn't work?

tobni: @adriangb I added a test that xfails. The internals of stdlib dataclasses does not assign default factory attributes to the class instance.

```python
@dataclasses.dataclass
class Base:
    a: int = dataclasses.field(defualt_factory=lambda: 5)

assert hasattr(Base, 'a')  # this fails
```

tobni: The inheritance works fine if you provide a value, so for my usecase this isn't an important feature to support.

tests/test_dataclasses.py
```diff
@@ -2500,3 +2500,50 @@ class StdLibDataclass:
...
+    """This test documents that default factories are not supported"""
+
+    @dataclasses.dataclass
+    class Base:
+        a: str = dataclasses.field(default_factory=lambda: 'TEST')
+
+    class Model(BaseModel, Base):
+        pass
+
+    assert Model().a == 'TEST'
```
adriangb, line 2527: I'd rather see `with pytest.raises(...)` since that's what we use elsewhere

adriangb (changes requested):

tobni:
@adriangb Done!

tobni:
please review

adriangb (approved):

PR #7196, opened 2023-08-21
add CPython 3.12 and PyPy 3.10 to CI

davidhewitt:
## Change Summary

Now that 3.12 is in release candidate I think we should start testing in CI.

Also added PyPy 3.10 to CI, as pydantic-core tests and releases with it.

## Related issue number

Related to #7181

## Checklist

* [ ] Tests pass on CI

Selected Reviewer: @adriangb


Kludex:
I was recommended to do this instead: https://github.com/encode/starlette/blob/a8b8856ce393a82ab4a714131085dbc4d658e34d/.github/workflows/test-suite.yml#L21-L24

"allow-prereleases"

gotmax23:
I'm working on packaging pydantic-core and pydantic v2 in Fedora. You can `git am` https://src.fedoraproject.org/fork/gotmax23/rpms/python-pydantic/raw/v2/f/Fix-Python-3.12-test-failures.patch on top of your PR to address the Python 3.12 test failures. There were changes to TypeVar reprs and ABC error messages that broke a couple tests.

docs/usage/models.md
```diff
@@ -1162,21 +1162,24 @@ To do this, you may want to use a `default_factory`.
...
 class Model(BaseModel):
     uid: UUID = Field(default_factory=uuid4)
-    updated: datetime = Field(default_factory=datetime.utcnow)
+    updated: datetime = Field(default_factory=datetime_now)
 
 
 m1 = Model()
 m2 = Model()
 assert m1.uid != m2.uid
-assert m1.updated != m2.updated
```
Kludex, line 1179: I don't know how this was passing before.

Kludex:
please review

docs/usage/models.md
```diff
@@ -738,7 +738,7 @@ Also, like `List` and `Dict`, any parameters specified using a `TypeVar` can lat
     `TypeVar` needs to be wrapped inside [`SerializeAsAny`](serialization.md#serializing-with-duck-typing)
     for Pydantic to serialize `ChildModel` as `ChildModel`.
 
-```py
+```py requires="3.12"
```
lig, line 741: Is it only for tests? Would it still show different example code for different Python versions?

docs/usage/models.md
lig: Well, it kinda should produce different time each time the default factory is being called.

The following passes

```py
>>> from datetime import datetime
>>> assert datetime.utcnow() != datetime.utcnow()
```

tests/test_abc.py
```diff
@@ -10,6 +11,7 @@ class Model(BaseModel, abc.ABC):
         some_field: str
 
 
+@pytest.mark.skipif(sys.version_info < (3, 12), reason='error value different on older versions')
```
lig, line 14: I'd rather don't throw out tests for everything except an unreleased Python version.

lig (changes requested):

tests/test_abc.py
Kludex: As discussed, we are fine with this change.

docs/usage/models.md
Kludex: We have a mock on `datetime.now` on the `test_docs.py`, that's why it doesn't pass. 😅

docs/usage/models.md
Kludex: No, it shows the same thing for every Python version.

Kludex:
please review

cclauss:
It might be time to revisit this with [PyPy3.11](https://pypy.org/download.html), which fails the tests when I try it here:
https://github.com/pydantic/pydantic/blob/58befdcc6e49d793b61bcadfb36b2a4e19b49bff/.github/workflows/ci.yml#L97-L98

davidhewitt:
Thanks for flagging, I'll investigate now

Issue #6899, opened 2023-07-27
Support `__get_validators__`


samuelcolvin:
I know it's "late", but should we provide backwards compatibility with `__get_validators__`?

It wouldn't be that hard and would help those migrating a lot.

Selected Assignee: @hramezani

Kludex:
@samuelcolvin Same thoughts still or things changed?

PR #7197, opened 2023-08-21
Support __get_validators__

hramezani:
Closes https://github.com/pydantic/pydantic/issues/6899

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -587,16 +587,21 @@ class GenerateSchema:
         else:
             ref_mode = 'to-def'
 
+        schema: CoreSchema
         get_schema = getattr(obj, '__get_pydantic_core_schema__', None)
         if get_schema is None:
-            return None
-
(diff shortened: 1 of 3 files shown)


samuelcolvin (commented):
we should add a deprecation warning, otherwise LGTM.

adriangb (commented):
Do we care how this interacts with `Annotated`, including with other classes that implement `__get_pydantic_core_schema__`? Might be worth testing

hramezani:
> Do we care how this interacts with `Annotated`, including with other classes that implement `__get_pydantic_core_schema__`? Might be worth testing

@adriangb, I think we already have a test for this. Is that enough?

https://github.com/pydantic/pydantic/blob/d110b47803eb484bbc85cf9a2105e24a5852c2e2/tests/test_deprecated.py#L498-L523

adriangb:
But that doesn't use this new / old `__get_valdiators__` right?

adriangb (approved):

Issue #7236, opened 2023-08-23
Generic type annotation breakes hashing at runtime


jgarvin:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

See the commented out line below -- switching the inheritance order will make the type unhashable and the script will fail with:

```
Traceback (most recent call last):
  File "/tmp/reproducer.py", line 18, in <module>
    print({B(x=2)})
          ^^^^^^^^
TypeError: unhashable type: 'B'
```

This seems undesirable since the base class is just a type annotation not actually intended to change runtime behavior, and the error message doesn't give you any hint what is wrong.

### Example Code

```Python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from pydantic import BaseModel, ConfigDict
from typing import Generic, TypeVar

class Config(BaseModel):
    model_config = ConfigDict(frozen=True)

ConfigT = TypeVar("ConfigT", bound=Config)

# class B(Generic[ConfigT], BaseModel): # this order doesn't work!!!
class B(BaseModel, Generic[ConfigT]):
    x: int

    model_config = ConfigDict(frozen=True)

print({B(x=2)})
```

### Python, Pydantic & OS Version

```Text
$ python3.11 -c "import pydantic.version; print(pydantic.version.version_info())"
             pydantic version: 2.1.1
        pydantic-core version: 2.4.0
          pydantic-core build: profile=release pgo=true mimalloc=true
                 install path: /home/xxxxx/.local/lib/python3.11/site-packages/pydantic
               python version: 3.11.4 (main, Jun  7 2023, 12:45:49) [GCC 9.4.0]
                     platform: Linux-5.18.12-051812-generic-x86_64-with-glibc2.31
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @davidhewitt

dmontagu:
Yeah, putting `Generic` before `BaseModel` causes other weird errors as well. I recently worked on an issue caused by that, but I guess it only updated the error message if it actually caused other misbehavior. We should probably just make it raise an error any time you have a generic `BaseModel` where `Generic` isn't the last thing.

@jgarvin do you _need_ the generic to not be the last item? Or would it be enough for us to just tell you to put it last if you don't (so you don't waste time on this in the future)? I think it might be a more challenging change to make everything work properly if `Generic` is not the last base class.

jgarvin:
[...]

PR #7274, opened 2023-08-29
Fix hash function generation for frozen models with unusual MRO

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/7236

Also fixes a bug with the way attributes were read from nested typeddicts (mro was being iterated over in reverse order), and adds a related test.

Selected Reviewer: @hramezani

### DIFF
--- a/tests/test_types_typeddict.py
+++ b/tests/test_types_typeddict.py
@@ -21,6 +21,7 @@ from pydantic import (
     PydanticUserError,
     ValidationError,
 )
+from pydantic._internal._decorators import get_attribute_from_bases
 from pydantic.functional_serializers import field_serializer, model_serializer
 from pydantic.functional_validators import field_validator, model_validator
 from pydantic.type_adapter import TypeAdapter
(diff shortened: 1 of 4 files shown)


tests/test_types_typeddict.py
```diff
@@ -905,3 +906,16 @@ class MySubTypedDict(MyMiddleTypedDict):
...
+    class A(TypedDict):
+        x = 1
+
+    class B(A):
+        x = 2
+
+    class C(B):
+        pass
+
+    assert get_attribute_from_bases(C, 'x') == 2
```
dmontagu, line 921: Currently on the `main` branch, this returns `1`

dmontagu:
please review

tests/test_types_typeddict.py
adriangb: Oh wow I’m very glad you caught this 🙈

adriangb (approved):
Amazing work @dmontagu !

Issue #7209, opened 2023-08-22
model serialization / validation schemas leaking into schema


commonism:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

As already discussed in https://github.com/tiangolo/fastapi/discussions/10019#discussioncomment-6737227 current pydantic defines different models in the schema when using both modes (validation/serialization) on a model - "{}Input" & "{}Output".
FastAPI started using mode=serialization for response models in 0.101 - bringing this issue up https://github.com/tiangolo/fastapi/discussions/10039 https://github.com/tiangolo/fastapi/discussions/10041.

This breaks all existing clients and invalidates every client code written due to changing the class name.
Current pydantic choose to change the name of the models again - adding a "-" before the Input/Output.

I understand there is a requirement to have different personalities, but the schema should be limited to a single personality - validation.

### Example Code

```Python
import io

import yaml

from pydantic import BaseModel
from pydantic.json_schema import models_json_schema


class Pet(BaseModel):
    name: str
    weight: int | None
    color: str | None = None

data = models_json_schema([(Pet,"serialization"),(Pet,"validation")])
f = io.StringIO()
for d in data:
    for k,v in d.items():
        print(yaml.dump(v, f, indent=4))
f.seek(0)
print(f.read())
```

```yaml
$ref: '#/$defs/PetOutput'
$ref: '#/$defs/PetInput'
PetInput:
    properties:
        color:
            anyOf:
            -   type: string
            -   type: 'null'
            default: null
            title: Color
        name:
            title: Name
            type: string
        weight:
            anyOf:
            -   type: integer
            -   type: 'null'
            title: Weight
    required:
    - name
    - weight
    title: Pet
    type: object
PetOutput:
    properties:
        color:
            anyOf:
            -   type: string
            -   type: 'null'
            default: null
            title: Color
        name:
            title: Name
            type: string
        weight:
            anyOf:
            -   type: integer
            -   type: 'null'
            title: Weight
    required:
    - name
    - weight
    - color
    title: Pet
    type: object
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.2.1
        pydantic-core version: 2.6.1
          pydantic-core build: profile=release pgo=true
                 install path: ~/venv/openapi3/lib/python3.10/site-packages/pydantic
               python version: 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
[...]

PR #7275, opened 2023-08-29
Make fields with defaults not required in the serialization schema by default

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/7209

Note this is not against main, but against @Kludex's branch for supporting ser json stuff.

This should have tests added before being merged, but I wanted to open now to get some initial review. Reading the discussion in the issue linked above may be helpful for understanding the motivation for this change.

### DIFF
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -76,6 +76,10 @@ class ConfigWrapper:
     hide_input_in_errors: bool
     defer_build: bool
     schema_generator: type[GenerateSchema] | None
+    json_schema_serialization_defaults_required: bool
+    json_schema_mode_override: Literal['validation', 'serialization', None]
+    json_schema_validation_suffix: str
+    json_schema_serialization_suffix: str
 
     def __init__(self, config: ConfigDict | dict[str, Any] | type[Any] | None, *, check: bool = True):
         if check:
@@ -239,6 +243,10 @@ config_defaults = ConfigDict(
     json_encoders=None,
     defer_build=False,
     schema_generator=None,
+    json_schema_serialization_defaults_required=False,
+    json_schema_mode_override=None,
+    json_schema_validation_suffix='-Input',
+    json_schema_serialization_suffix='-Output',
 )
 
 
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -207,5 +207,47 @@ class ConfigDict(TypedDict, total=False):
     Defaults to `None`.
     """
 
+    json_schema_serialization_defaults_required: bool
+    """
+    Whether fields with default values should be marked as required in the serialization schema.
+
+    This ensures that the serialization schema will reflect the fact a field with a default will always be present
+    when serializing the model, even though it is not required for validation.
+
+    However, there are scenarios where this may be undesirable — in particular, if you want to share the schema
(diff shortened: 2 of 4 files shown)


dmontagu:
please review

docs/usage/model_config.md
```diff
@@ -658,7 +658,7 @@ except NameError as e:
 
 ## Hide Input in Errors
 
-_Pydantic_ shows the input value and type when it raises `ValidationError` during the validation.
+Pydantic shows the input value and type when it raises `ValidationError` during the validation.
```
dmontagu, line 661: I think the italics here were a vestige of older pattern, just figured I'd address this while I was at it.

docs/usage/model_config.md
```diff
@@ -701,3 +701,100 @@ except ValidationError as e:
...
+    'required': ['a'],
+    'title': 'ForceInputModel',
+    'type': 'object',
+}
+"""
+```
+
+This can be useful when using frameworks (such as FastAPI) that may generate different schemas for validation
+and serialization that must both be referenced from the same schema; when this happens, we automatically append
+`-Input` to the definition reference for the validation schema and `-Output` to the definition reference for the
```
dmontagu, line 797: The reason I removed the ability to override the `-Input` and `-Output` values was twofold:
* Right now, these refs are computed after the config has been removed from the config stack. It was possible to work around this, but very ugly
* I think the most compelling use case was to be able to set one of suffixes to be the empty string, e.g. so the input would use the empty string and output would use `'-Output'`. Unfortunately, the way I've implemented it, it tries to use the first definition reference that is not used by any other schema. Because the serialization schema tries to use without the suffix before using the suffix, it means the empty string just doesn't work, because there are always two schemas that want the name `Model`.

I believe with enough effort it may be possible to handle this better, but I think the changes still made by this PR are higher priority than controlling the definition ref suffix so I figured it was worth getting them in and deferring further work on this.

pydantic/types.py
```diff
@@ -713,7 +713,7 @@ class Model(BaseModel):
 path.unlink()
 
 path = Path('directory')
-path.mkdir()
+path.mkdir(exist_ok=True)
```
dmontagu, line 716: Unrelated to this PR, but this resolved some local test failures for me, and I don't see any reason not to keep it there.

Kludex:
Assigning myself since I'm already in the loop here.

docs/usage/model_config.md
```diff
@@ -701,3 +701,100 @@ except ValidationError as e:
       Input should be a valid string [type=string_type]
     """
 ```
+
+## JSON schema customization
+
+#### `json_schema_serialization_defaults_required`
```
Kludex, line 707: There's a limit on characters. Also, I don't think we should use parameter name as the name of the section.

Screenshot 2023-09-18 at 11 10 35

docs/usage/model_config.md
```diff
@@ -701,3 +701,100 @@ except ValidationError as e:
...
+By default, the JSON schema generated for serialization will mark fields as not-required, even if they
```
Kludex, line 708:
```suggestion

By default, the JSON schema generated for serialization will mark fields as **not-required**, even if they
```

Kludex (approved):

PR #7299, opened 2023-08-30
Mark `Extra` as deprecated

disrupted:
## Change Summary

Similary to the `BaseConfig`, `Extra` is now marked using the `@deprecated` decorator for IDE support to advise against its usage.

## Checklist

* [x] Tests pass on CI

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/config.py
+++ b/pydantic/config.py
@@ -1,15 +1,12 @@
 """Configuration for Pydantic models."""
 from __future__ import annotations as _annotations
 
-import warnings
 from typing import TYPE_CHECKING, Any, Callable, Dict, Type, Union
 
 from typing_extensions import Literal, TypeAlias, TypedDict
 
 from ._migration import getattr_migration
-from .deprecated.config import BaseConfig
-from .deprecated.config import Extra as _Extra
-from .warnings import PydanticDeprecatedSince20
+from .deprecated.config import BaseConfig, Extra
 
 if TYPE_CHECKING:
     from ._internal._generate_schema import GenerateSchema as _GenerateSchema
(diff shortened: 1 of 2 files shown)


dmontagu (approved):

dmontagu:
I just pushed a commit that I think should make the test pass by suppressing the deprecation warning for the instantiation of the `Extra` class. Let me run this by the other maintainers, but LGTM.

Also, it would be good if you could confirm that with the changes I pushed it still shows the deprecation properly — I use PyCharm which doesn't reflect that decorator properly yet 😕.

pydantic/config.py
```diff
@@ -22,7 +24,12 @@
     Callable[[Dict[str, Any], Type[Any]], None],
 ]
 
-Extra = _Extra()
+with warnings.catch_warnings():
+    # The relevant deprecation warning is still raised on attribute access,
+    # so it is okay to suppress the one coming from the @deprecated decorator here.
+    warnings.filterwarnings('ignore', category=PydanticDeprecatedSince20)
+
+    Extra = _Extra()
```
Kludex, line 32: Why is it `Extra = _Extra()` and not `Extra = _Extra`?

Kludex (changes requested):
This doesn't work as expected on VSCode because of the reexport on `config.py`, and the instantiation of `_Extra()`.

pydantic/config.py
disrupted: I've completely removed the instantiation of `Extra` now so that it is only re-exported same as `BaseConfig`

disrupted: `getattr(Extra, attribute)` doesn't seem to work without instantiating it first

disrupted: I've reverted the change

dmontagu: @Kludex I _think_ we can make that work but then we need to override getattribute on the metaclass. Might help to go over it live so we can confirm it's working with display etc.

Kludex (approved):

pydantic/config.py
disrupted: @dmontagu @Kludex is there anything else I should do or are you taking this over?

Kludex: I'm taking over. Thanks @disrupted 🙏

Issue #7293, opened 2023-08-30
model_post_init not called on subclasses with a private attribute


jrapin:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

In the example below, I would expect the `Model1.model_post_init` to be called when instantiating `Model2` but it is not.

On the other hand, `Model1.model_post_init` is called:
- when instantiating `Model1`
- when instantiating `Model1` even if I add a private attribute
- when instantiating `Model2` if I change `_private` into `private`
- when instantiating `Model2` if I add a `model_post_init` in `Model2` which calls `Model1.model_post_init`

Is this an expected behavior?
Thanks for your help!

Note: `Model2.__pydantic_post_init__` remains set to `'model_post_init'`

### Example Code

```Python
import pydantic

class Model1(pydantic.BaseModel):
    a: int = 12

    def model_post_init(self, __context) -> None:
        print("Post init", self.__class__)

class Model2(Model1):
    _private: bool = True

Model2() # does not trigger inherited post init
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.3.0
        pydantic-core version: 2.6.3
          pydantic-core build: profile=release pgo=false
                 install path: /Users/jrapin/Applications/miniconda3/envs/test_env/lib/python3.10/site-packages/pydantic
               python version: 3.10.12 (main, Jul  5 2023, 15:02:25) [Clang 14.0.6 ]
                     platform: macOS-13.5.1-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

Selected Assignee: @Kludex

Viicos:
This coming from:

https://github.com/pydantic/pydantic/blob/b9d2a98e8f66bb0c8aab77d1ed200a2769d7eca4/pydantic/_internal/_model_construction.py#L106-L120

In your case, `private_attributes` evaluates to `True`, but as you do not define `model_post_init` in `Model2`, it is not present in the `namespace`, and gets overridden by `init_private_attributes`.

Maybe something like this could fix it?

```python
def get_model_post_init(namespace: dict[str, Any], bases: tuple[type[Any], ...]) -> Any | None:
    if 'model_post_init' in namespace:
        return namespace['model_post_init']

    for base in bases:
        model_post_init = getattr(base, 'model_post_init')
        if model_post_init is not None:
            return model_post_init

# then in ModelMetaclass__new__...

            if private_attributes:
                original_model_post_init = get_model_post_init(namespace, bases)
[...]

PR #7302, opened 2023-08-31
Do not override `model_post_init` in subclass with private attrs

Viicos:
## Related issue number

Fixes https://github.com/pydantic/pydantic/issues/7293

## Checklist

* [x] Tests pass on CI

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -272,7 +272,6 @@ def get_model_post_init(namespace: dict[str, Any], bases: tuple[type[Any], ...])
         return namespace['model_post_init']
 
     for base in bases:
-        # Or use cls.__mro__ instead of bases?
         model_post_init = getattr(base, 'model_post_init')
         if model_post_init is not None:
             return model_post_init
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -22,6 +22,7 @@ from ._decorators import (
     DecoratorInfos,
     PydanticDescriptorProxy,
     get_attribute_from_bases,
+    mro_for_bases,
 )
 from ._discriminated_union import apply_discriminators
 from ._fields import collect_model_fields, is_valid_field_name, is_valid_privateattr_name
@@ -274,7 +275,10 @@ def get_model_post_init(namespace: dict[str, Any], bases: tuple[type[Any], ...])
     from ..main import BaseModel
 
     model_post_init = get_attribute_from_bases(bases, 'model_post_init')
-    if model_post_init is not BaseModel.model_post_init:
+    for base in mro_for_bases(bases):
+        # Using `__dict__` instead of `getattr` to avoid getting the `model_post_init` from `BaseModel`
+        model_post_init = base.__dict__.get('model_post_init')
+    if model_post_init is not None and model_post_init is not BaseModel.model_post_init:
         return model_post_init
 
 
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -278,8 +278,8 @@ def get_model_post_init(namespace: dict[str, Any], bases: tuple[type[Any], ...])
     for base in mro_for_bases(bases):
         # Using `__dict__` instead of `getattr` to avoid getting the `model_post_init` from `BaseModel`
         model_post_init = base.__dict__.get('model_post_init')
-    if model_post_init is not None and model_post_init is not BaseModel.model_post_init:
-        return model_post_init
+        if model_post_init is not None and model_post_init is not BaseModel.model_post_init:
+            return model_post_init
 
 
 def inspect_namespace(  # noqa C901
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -271,10 +271,11 @@ def get_model_post_init(namespace: dict[str, Any], bases: tuple[type[Any], ...])
     if 'model_post_init' in namespace:
         return namespace['model_post_init']
 
-    for base in bases:
-        model_post_init = getattr(base, 'model_post_init', None)
-        if model_post_init is not None:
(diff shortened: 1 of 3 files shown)


Viicos:
Please review

`[PermissionError]: [Errno 13] Permission denied: '/home/runner/work/pydantic/pydantic/coverage'` 🤨

pydantic/_internal/_model_construction.py
```diff
@@ -266,6 +266,17 @@ def init_private_attributes(self: BaseModel, __context: Any) -> None:
     object_setattr(self, '__pydantic_private__', pydantic_private)
 
 
+def get_model_post_init(namespace: dict[str, Any], bases: tuple[type[Any], ...]) -> Callable[..., Any] | None:
+    """Get the `model_post_init` method from the namespace or the class bases, or `None` if not defined."""
+    if 'model_post_init' in namespace:
+        return namespace['model_post_init']
+
+    for base in bases:
+        model_post_init = getattr(base, 'model_post_init', None)
```
Viicos, line 275: This will actually always be defined for any base, as any subclass of `BaseModel` will inherit from the original empty method:

https://github.com/pydantic/pydantic/blob/c987bc6398b898d5c9e49606e3acb17b732359e8/pydantic/main.py#L417-L421

So if we have the following:

```python

class A(BaseModel):
    a: int = 1

class B(BaseModel):
    b: int = 1

    def model_post_init(self, __context):
        """Do custom things"""

class C(A, B):
    _private: bool = True
```

`get_model_post_init` will use the empty `model_post_init` available on `A`, and skip the custom one from `B`. Any ideas on how to prevent this? Maybe a custom flag could help, similar to:

https://github.com/pydantic/pydantic/blob/c987bc6398b898d5c9e49606e3acb17b732359e8/pydantic/main.py#L167-L168

?

dmontagu: If your concern is to make sure attribute lookup only considers things defined on the exact class, the way we achieve that elsewhere is to look in `base.__dict__`. So, if you do `if 'model_post_init' in base.__dict__`, that tells you if it was defined specifically in `base` vs. in one of its subclasses. The best way to do it might be to loop over the `__mro__` for the class looking for items in the classes' `__dict__`s.

Thinking this through, I think if we want to auto-generate a method but not have that auto-generated version factor into inheritance, using a flag like you've described is probably the best way. _But_, since in this case the method _is_ defined on BaseModel, I think the correct behavior would actually involve calling the BaseModel version in your code above, since that is consistent with standard python behavior for method resolution:

```python
class A:
    def my_method(self):
        print('A')

class B(A):
    pass

class C:
    def my_method(self):
        print('C')

class D(B, C):
    pass

D().my_method()
#> A
```

Not sure if I'm overlooking something important here; in general I'm open to other points of view but I think it will be important that whatever we do doesn't introduce breaking changes here. More generally, I won't be surprised if the best solution ends up being for users to just be more proactive about overriding `model_post_init` in subclasses when inheriting from multiple base models and not being able to put the one with the desired `model_post_init` first in the inheritance order.

dmontagu: Also, I think if all you are wanting to do is specifically ignore `BaseModel.model_post_init`, it might be best to add something like:
```python
if model_post_init is BaseModel.model_post_init:
    continue
```

pydantic/_internal/_model_construction.py
```diff
@@ -266,6 +266,17 @@ def init_private_attributes(self: BaseModel, __context: Any) -> None:
...
+        if model_post_init is not None:
+            return model_post_init
```
dmontagu, line 277:
```suggestion
    model_post_init = get_attribute_from_bases(bases, 'model_post_init')
    if model_post_init is not BaseModel.model_post_init:
        return model_post_init
```
I think this is probably the "most correct" version of this, since we are already using the `get_attribute_from_bases` function which handles mro the same as python, etc.

dmontagu: hmm I see this errors, I'll see if I can achieve the same thing in a better way..

dmontagu: The above is also closely related to what we do when setting `__pydantic_post_init__` (i.e., ignoring `BaseModel.model_post_init` and follows the same logic as we do when setting `__hash__` (i.e., using the mro).

Viicos: Unfortunately even using `get_attribute_from_bases`, the test is still failing (in fact `get_attribute_from_bases` is doing something quite similar, i.e. using `hasattr` and `getattr`, so we face the same issue). Maybe doing something like the following:

```python
for base in mro_for_bases(bases):
    model_post_init = base.__dict__.get('model_post_init')
    if model_post_init is not None and model_post_init is not BaseModel.model_post_init:
        return model_post_init
```

dmontagu: @Viicos this is a good catch — but actually I think this is a bug in the `get_attribute_from_bases` function — it shouldn't be using `hasattr` and `getattr` for exactly this reason. I think looking in the class `__dict__` should fix this and allow us to use that function, if it does, I would propose we make that change. Any objections? If not, I can push a commit and merge this.

Viicos: Sure, do you want to do it in a separate PR? Might be easier if it breaks some tests. Then I'll rebase on `main` once merged

dmontagu: Doesn't break any tests — I think my preference would be to just put it in this PR since the test you've added is actually a good test of that change.

To reduce the back-and-forth I'll just push the commit in hopes you are okay with it; I can revert that if you prefer though, and move it elsewhere.

dmontagu (approved):

pydantic/_internal/_model_construction.py
Viicos: Totally ok with it, thanks for the last push!

PR #7347, opened 2023-09-05
Get rid of `__prepare_pydantic_annotations__`

adriangb:
This simplifies schema generation a bit and (hopefully) enables further refactors down the road that speed up schema generation.


adriangb:
please review

tests/test_annotated.py
```diff
@@ -328,7 +328,7 @@ class Model(BaseModel):
         Model(x=2)
     # insert_assert(exc_info.value.errors(include_url=False))
     assert exc_info.value.errors(include_url=False) == [
-        {'type': 'less_than', 'loc': ('x',), 'msg': 'Input should be less than 4', 'input': 2, 'ctx': {'lt': '4'}}
+        {'type': 'less_than', 'loc': ('x',), 'msg': 'Input should be less than 4', 'input': 2, 'ctx': {'lt': 4}}
```
adriangb, line 331: We used to stringify errors, now no longer doing it which makes the divergent code paths better match.

tests/test_types.py
```diff
@@ -4029,11 +4028,11 @@ class Foobar(BaseModel):
...
         {
-            'type': 'string_too_short',
+            'type': 'too_short',
             'loc': ('password',),
-            'msg': 'String should have at least 6 characters',
+            'msg': 'Value should have at least 6 items after validation, not 0',
             'input': '',
-            'ctx': {'min_length': 6},
+            'ctx': {'field_type': 'Value', 'min_length': 6, 'actual_length': 0},
         }
```
adriangb, line 4036: Since `SecretStr` is a composite type it will no longer get the `string_too_short` error and instead will get the more general `too_short` error. It seems like the latter is maybe even better, it includes the actual length in the context for starters. I'd be +1 for getting rid of `string_too_short` altogether.

tests/test_types.py
```diff
@@ -5276,27 +5275,22 @@ def ___repr__(self) -> str:
...
+            return core_schema.no_info_after_validator_function(PdDecimal, handler.generate_schema(Decimal))
 
     class Model(BaseModel):
         x: Annotated[PdDecimal, PdDecimalMarker(), annotated_types.Gt(0)]
 
     assert isinstance(Model(x=1).x, PdDecimal)
     with pytest.raises(ValidationError) as exc_info:
         Model(x=-1)
-    # insert_assert(exc_info.value.errors(include_url=False))
+
```
adriangb, line 5286:
```suggestion
    # insert_assert(exc_info.value.errors(include_url=False))
```

tests/test_types.py
```diff
@@ -5369,61 +5362,103 @@ class Model(BaseModel):
...
-            'msg': 'String should have at least 1 characters',
-            'type': 'string_too_short',
+            'msg': 'Value should have at least 1 item after validation, not 0',
+            'input': '',
+            'ctx': {'field_type': 'Value', 'min_length': 1, 'actual_length': 0},
+        }
+    ]
+
+
+def test_constraint_dataclass() -> None:
```
adriangb, line 5377: I added this test just because I found it as part of developing this PR and fixed the bug (see changes in`_generate_schema.py`) but I could move it out to another PR if desired.

tests/test_types.py
```diff
@@ -5369,61 +5362,103 @@ class Model(BaseModel):
...
+            'ctx': {'gt': MyDataclass(0)},
         }
     ]
 
 
 def test_transform_schema_for_third_party_class():
     # Here, third party means you can't define methods on the class directly, so have to use annotations.
 
-    class DatetimeWrapper:
+    class IntWrapper:
```
adriangb, line 5407: I had to change from datetime to int because `PydanticKnownError` w/ `gt` only accepts numeric types. We can either use a `PydanticCustomError` here which is mostly API compatible but might take a performance hit (when I first implemented the functions in `_validators.py` @samuelcolvin pushed me to use `PydanticKnownError` over `PydanticCustomError` for this reason) or we can change `PydanticKnownError` in `pydantic-core` to not be as pedantic.

dmontagu (approved):

PR #7363, opened 2023-09-07
Attempt to fix docs deploy job

dmontagu:


dmontagu:
The only thing left that I can think of would be changing from python 3.10 to 3.8 (as used by the cloudflare docs builder) since that seems to be working based on the docs previews. If that doesn't fix the deployment issue, then I would think it would have to be something with `mike deploy` vs. `mkdocs build`

sydney-runkle (approved):
Looks good to me @dmontagu, thanks for digging into this issue!

PR #7370, opened 2023-09-08
📝 Remove `string_types.md` page

Kludex:
Small improvements :)

- Related https://github.com/pydantic/pydantic/issues/7322


docs/usage/types/string_types.md
```diff
@@ -9,16 +9,65 @@ enums inheriting from `str` are converted using `v.value`, and all other types c
...
+    To use this type, you need to install the optional [`email-validator`][email-validator] package:
+
+    ```bash
+    pip install email-validator
+    ```
+
+
+The input string must be a valid email address as specified by the [RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4).
+
+The `NameEmail` has two properties: `name` and `email`. In case the name is not provided, it's inferred from the email address.
```
hramezani, line 49:
```suggestion
The `NameEmail` has two properties: `name` and `email`. In case the `name` is not provided, it's inferred from the email address.
```

hramezani (approved):

docs/usage/types/string_types.md
Kludex: Complied. 👍

Issue #7399, opened 2023-09-07
mypy thinks a config variable can never be `None` when nested in inherited BaseModel class


Greesb:
Hello,

I just switched a project to pydantic v2 and encountered a bug with mypy thinking that some variable can never be `None` when the typing and the default value clearly indicate otherwise.

Here's a minimal code that reproduce the error:
```python
 
import pydantic


class ModelNested(pydantic.BaseModel):
    my_var: str | None = pydantic.Field(default=None)


class Model(ModelNested):
    pass

m = Model()
if m.my_var is None:
    print("hello there")  # mypy thinks we can never go inside this if
```

Here's the mypy error: `test.py:14: error: Statement is unreachable  [unreachable]`

hramezani:
related to https://github.com/pydantic/pydantic/issues/7398

PR #7411, opened 2023-09-11
Fix mishandling of unions while freezing types in the mypy plugin

dmontagu:
This addresses the specific issue brought up in https://github.com/pydantic/pydantic/issues/7399, but I am concerned that there is a bigger problem that I'm not adequately addressing here.

I will raise an issue in the mypy repository asking for more information, as I think the problem that the pydantic mypy plugin is hitting has something to do with a warning comment in code I adapted from the dataclasses plugin.

If we are going to merge this, we should update the mypy plugin tests to include the example from #7399 (including adding the `unreachable` warning).

Closes https://github.com/pydantic/pydantic/issues/7399


dmontagu:
I think if this still passes CI let's just merge it @sydney-runkle, I created the mypy issue https://github.com/python/mypy/issues/16087 but haven't heard anything back so I think it makes sense to address the case people are actually hitting and worry about doing a better job in the future if more issues get reported.

sydney-runkle (approved):
Nice. Waiting on approval of the `v2.5.0` release, then will merge this later today :).

PR #7414, opened 2023-09-12
Remove `bytesize.md` page

hramezani:
- Related to https://github.com/pydantic/pydantic/issues/7322

Selected Reviewer: @lig


hramezani:
please review

docs/usage/types/strict_types.md
```diff
@@ -93,3 +93,19 @@ m = Model(license_plate='ABC-123')
 print(m)
 #> license_plate='ABC-123'
 ```
+
+However, for the sake of improved integration with type checkers, we now discourage the use of `conbytes` and other
+function calls used to return constrained types. Instead, you can use `pydantic.types.Strict` and `annotated_types.Len`
```
Kludex, line 98:
```suggestion
function calls used to return constrained types. Instead, you can use [`pydantic.types.Strict`][pydantic.types.Strict] and [`annotated_types.Len`][<LINK>]
```

Kludex (changes requested):

hramezani:
please review

Issue #7407, opened 2023-09-11
Standadize EmailStr validator __get_pydantic_core_schema__ method


aasay:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Currently the __get_pydantic_core_schema__ method for the EmailStr validator only accepts a single argument for "source".  All of the documentation for __get_pydantic_core_schema__ references that it should accept both a source and a handler argument.  This specifically is a problem when working with third party packages and knowing how many arguments to pass.

### Example Code

Current Code /pydantic/networks.py

```python
if TYPE_CHECKING:
    EmailStr = Annotated[str, ...]
else:

    class EmailStr:
        """Validate email addresses.

        Example:
            ```py
            from pydantic import BaseModel, EmailStr

            class Model(BaseModel):
                email: EmailStr

            print(Model(email='contact@mail.com'))
            #> email='contact@mail.com'
            ```
        """

        @classmethod
        def __get_pydantic_core_schema__(
            cls,
            source: type[Any]
        ) -> core_schema.CoreSchema:
            import_email_validator()
            return core_schema.general_after_validator_function(cls._validate, core_schema.str_schema())
```

Proposed

```python
if TYPE_CHECKING:
    EmailStr = Annotated[str, ...]
else:

    class EmailStr:
        """Validate email addresses.

        Example:
            ```py
            from pydantic import BaseModel, EmailStr

            class Model(BaseModel):
                email: EmailStr

            print(Model(email='contact@mail.com'))
            #> email='contact@mail.com'
            ```
        """

        @classmethod
        def __get_pydantic_core_schema__(
            cls,
            source: type[Any],
            handler: Callable[[Any], core_schema.CoreSchema] = None
        ) -> core_schema.CoreSchema:
            import_email_validator()
            return core_schema.general_after_validator_function(cls._validate, core_schema.str_schema())

```

hramezani:
Thanks @aasay for this issue 🙏 
I've merged a [PR](https://github.com/pydantic/pydantic/commit/b46010bf4a73d439e3faf98e6e2e16e2faf7e6e2) to fix this.

PR #7415, opened 2023-09-12
Standardize `__get_pydantic_core_schema__` signature

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/7407

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/networks.py
+++ b/pydantic/networks.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING, Any
 from pydantic_core import MultiHostUrl, PydanticCustomError, Url, core_schema
 from typing_extensions import Annotated, TypeAlias
 
-from ._internal import _fields, _repr, _schema_generation_shared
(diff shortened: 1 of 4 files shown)


hramezani:
please review

pydantic/networks.py
```diff
@@ -360,6 +363,7 @@ def __get_pydantic_json_schema__(
     def __get_pydantic_core_schema__(
         cls,
         source: type[Any],
+        handler: _annotated_handlers.GetCoreSchemaHandler,
```
Kludex, line 366: Are those needed here?

hramezani: Actually, not needed because they are working without them. This is just to make all the `__get_pydantic_core_schema__` signature similar.

Kludex: I see why you changed the other files, but since this is optional here, I would leave without it.

Anyway, not a blocker. Do as you think is more appropriate. 🙏

Kludex (approved):

adriangb (approved):

Issue #7409, opened 2023-09-11
Reduce import time


ofek:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this feature is missing

### Description

I am investigating switching various CLIs to use Pydantic over custom configuration logic. Currently, I unfortunately cannot due to the overhead of merely importing. I think `pydantic_core` cannot be optimized further but I believe `pydantic` can be. The following are taken from Python 3.11:

### `pydantic_core`

```
❯ python -m timeit -n 1 -r 1 "import pydantic_core"
1 loop, best of 1: 17.1 msec per loop
❯ python -X importtime -c "import pydantic_core"
import time: self [us] | cumulative | imported package
import time:        92 |         92 | winreg
import time:       182 |        182 |   _io
import time:        41 |         41 |   marshal
import time:       162 |        162 |   nt
import time:       630 |       1014 | _frozen_importlib_external
import time:       335 |        335 |   time
import time:       219 |        554 | zipimport
import time:        51 |         51 |     _codecs
import time:       362 |        412 |   codecs
import time:      1071 |       1071 |   encodings.aliases
import time:      1504 |       2987 | encodings
import time:       601 |        601 | encodings.utf_8
import time:       630 |        630 | encodings.cp1252
import time:        45 |         45 | _signal
import time:        30 |         30 |     _abc
import time:       227 |        256 |   abc
import time:       260 |        515 | io
import time:        85 |         85 |       _stat
import time:       163 |        247 |     stat
import time:       720 |        720 |     _collections_abc
import time:        34 |         34 |       genericpath
import time:        77 |         77 |       _winapi
import time:       414 |        525 |     ntpath
import time:       490 |       1980 |   os
import time:        63 |         63 |   _sitebuiltins
import time:      1877 |       1877 |   _distutils_hack
import time:       697 |        697 |   types
import time:       665 |        665 |       warnings
import time:       726 |       1391 |     importlib
import time:       685 |        685 |     importlib._abc
import time:       111 |        111 |         itertools
import time:       447 |        447 |         keyword
import time:       100 |        100 |           _operator
import time:       667 |        767 |         operator
import time:       471 |        471 |         reprlib
import time:        68 |         68 |         _collections
import time:      1597 |       3460 |       collections
import time:       110 |        110 |         _functools
import time:       854 |        964 |       functools
import time:       987 |       5410 |     contextlib
[...]

PR #7423, opened 2023-09-13
experiment with reducing import time

samuelcolvin:
ref #7409

Experiment with reducing import time.

Removing dataclasses is pretty hard since:
1. it's used by `annotated-types`, I think persuading @adriangb to remove dataclasses from there and using a proxy/hack like we do here is pretty unlike
2. it's used in dataclass transform like `@dataclass_transform(field_specifiers=(dataclasses.field, Field))`, while this is only required by type checkers, I think it would be quite to remove them and keep `dataclass_transform` behaviour the same, maybe it's possible
3. We have a lot of dataclasses with pydantic, I've got rid of or deferred most of them here, but the remainders are hard

From my experience here the low hanging fruit to improve performance times are:
* defer everything related to `deprecated`, using a mixture of local imports and `pydantic.__init__.__getattr__`
* defer importing `pydantic.types`, and add it all it's exports to `pydantic.__init__.__getattr__`
* defer import of `pydantic_core.core_schema`
* hard to tell, but perhaps the `deferred_dataclasses` provides some improvements even through it doesn't yet avoid dataclasses being imported

### DIFF
--- a/pydantic/__init__.py
+++ b/pydantic/__init__.py
@@ -19,10 +19,10 @@ from ._internal._annotated_handlers import (
 )
 from ._internal._generate_schema import GenerateSchema as GenerateSchema
 from ._migration import getattr_migration
-from .config import ConfigDict, Extra
-from .deprecated.class_validators import root_validator, validator
-from .deprecated.config import BaseConfig
-from .deprecated.tools import *
+from .config import ConfigDict
+# from .deprecated.class_validators import root_validator, validator
+# from .deprecated.config import BaseConfig
+# from .deprecated.tools import *
 from .errors import *
 from .fields import AliasChoices, AliasPath, Field, PrivateAttr, computed_field
 from .functional_serializers import PlainSerializer, SerializeAsAny, WrapSerializer, field_serializer, model_serializer
@@ -40,7 +40,7 @@ from .json_schema import WithJsonSchema
 from .main import *
 from .networks import *
 from .type_adapter import TypeAdapter
-from .types import *
+# from .types import *
 from .validate_call import validate_call
 from .version import VERSION
 from .warnings import *
@@ -66,8 +66,8 @@ __all__ = [
     'PlainValidator',
     'WrapValidator',
     # deprecated V1 functional validators
-    'root_validator',
-    'validator',
+    # 'root_validator',
+    # 'validator',
     # functional serializers
     'field_serializer',
     'model_serializer',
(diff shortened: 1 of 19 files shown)


samuelcolvin:
Worth noting that the big win from not importing `dataclasses` is not importing `inspect`, but to get that win you need to avoid importing `typing_extensions` too. At that point, I'm not really sure it's worth it.

ofek:
FYI the current reduction is huge on Windows compared to before for me:

Before:

```
❯ python -m timeit -n 1 -r 1 "import pydantic"
1 loop, best of 1: 95.7 msec per loop
```

After:

```
❯ python -m timeit -n 1 -r 1 "import pydantic"
1 loop, best of 1: 65.9 msec per loop
```

ofek:
I don't know how much it would help here but in case you're unaware since Python 3.7 there exists a way to dynamically set module attributes: https://peps.python.org/pep-0562/

For example, in Hatch I have core functionality that lets people migrate their build configuration and project metadata automatically and I use that to patch setuptools: https://github.com/pypa/hatch/blob/hatch-v1.7.0/src/hatch/cli/new/migrate.py#L317-L330

ofek:
Anything I can do to help or test?

ofek:
Actually, I found out that there a well defined way to do lazy imports based on the PEP I mentioned that has already become a standard within the scientific Python community:

- https://scientific-python.org/specs/spec-0001/
- https://github.com/scientific-python/lazy_loader

I'm going to begin introducing that at work in fact soon based on successful preliminary tests I did today

samuelcolvin:
I think we should do the first two things on my list, anything else if going to take longer. Probably easiest via a new pr.

Of course anyone else is welcome to have a dig and see if they can improve import time.

samuelcolvin:
I've made some progress on this in #7590, this should be released in v2.4 today.

If that goes well and we don't receive any issues about problems with module `__getattr__` on v2.4, I would propose we put everything in `pydantic/__init__.py` in `__getattr__` at runtime.

samuelcolvin:
BTW, @adiangb has done lots of work for v2.4 improve the time taken to build pydantic models, see #7565 #7536 #7535 #7529 #7528 #7527 #7523.

While this won't affect import time, it'll make the time taken to import modules which include pydantic models much smaller.

ofek:
Is there anything I can do to help or test?

samuelcolvin:
@ofek see #7947 which should help a lot.

Closing this as it's not going to happen this way.

Issue #7437, opened 2023-09-13
TypedDict key - unable to validate against Required constraint


baderdean:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

When trying to validate a Schema.org JSON parsed from the Web into its equivalent TypedDict with a required key (here Name), I got this error about Required[str] not being a Python type. While I undertand that Required is not a class, it's still a builtin and the only way to make a TypedDict key required.

So I don't understand how could I validate a JSON against TypedDict keys using Pydantic? Sorry, I'm not even 100% sure it's a bug or that's just that TypedDict keys are not as supported as Dataclass but worth asking since I didn't found something about TypedDict keys not being fully supported.

### Example Code

```Python
from typing_extensions import TypedDict, Required
from pydantic import TypeAdapter
Thing = TypedDict("Thing", {"name" : Required[str]})
TypeAdapter(Thing.__annotations__['name'], config=dict(arbitrary_types_allowed=True)).validate_python("My Name")

/home/bader/code/python3/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py:364: UserWarning: typing.Required is not a Python type (it may be an instance of an object), Pydantic will allow any object with no validation since we cannot even enforce that the input is an instance of the given type. To get rid of this error wrap the type with `pydantic.SkipValidation`.
  warn(
'My Name'
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.1.1
        pydantic-core version: 2.4.0
          pydantic-core build: profile=release pgo=true mimalloc=true
                 install path: /home/bader/code/python3/lib/python3.11/site-packages/pydantic
               python version: 3.11.4 (main, Jun  9 2023, 07:59:55) [GCC 12.3.0]
                     platform: Linux-6.2.0-26-generic-x86_64-with-glibc2.37
     optional deps. installed: ['email-validator', 'typing-extensions']
```

dmontagu:
I opened a PR #7440 to address this, though I'm not sure it will be accepted.

If you want a function that will remove the `Required`/`NotRequired` from your annotation, as I alluded in the PR body you could take an approach like:
```python
from typing import Any

from typing_extensions import TypedDict, Required, NotRequired, get_args, get_origin
from pydantic import TypeAdapter

Thing = TypedDict("Thing", {"name" : Required[str]})


def prepare_typeddict_annotation(annotation: Any) -> Any:
    if get_origin(annotation) in (Required, NotRequired):
[...]

PR #7440, opened 2023-09-14
Generate a schema for Required/NotRequired

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/7437

If there are good objections to handling (and ignoring) the `Required`/`NotRequired` things, we can close this. I think it's reasonable to ignore them though; admittedly it is an edge case but I think it's nice if it works, in particular for the use case described in #7437.

@baderdean if this PR gets rejected or you aren't patient enough to wait for a new pydantic release, you should be able to take a similar approach to this PR to pass your annotation into a function that will "strip off" the `Required`/`NotRequired` to achieve the same thing in your own code without needing the changes here.

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -33,6 +33,7 @@ from typing import (
 )
 from warnings import warn
 
+import typing_extensions
 from pydantic_core import CoreSchema, PydanticUndefined, core_schema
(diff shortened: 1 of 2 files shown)


dmontagu:
Closing for the reasons described in https://github.com/pydantic/pydantic/issues/7437#issuecomment-1808431997

PR #7449, opened 2023-09-14
Slightly improve schema generation performance

adriangb:
Chip away at #6768 

This takes the benchmark from `10.91s` to `9.46s` on my laptop, a respectable ~10% improvement. But still not good enough.


adriangb:
please review

adriangb:
The thing that comes to mind that would give us the order of magnitude performance improvement we are looking for is caching the generated schemas.
We should definitely be able to cache "simple" types to avoid re-generating their schema. You'd think it might be as easy as saying `if type_ in SOME_GLOBAL_CACHE` but it's not because if any of (1) the type is a generic type (remember this includes things as simple as `list[]`) or (2) we are within a model then the state of type var scopes and model configs comes into play, and those things are far from trivial to use as a key in a cache.

adriangb:
Closing in favor of #7473

PR #7473, opened 2023-09-17
Remove flattening of schemas before inlining and add schema walking caches

adriangb:
Needs https://github.com/pydantic/pydantic-core/pull/965

Selected Reviewer: @dmontagu


adriangb:
This brings `tests/benchmarks/test_fastapi_startup.py` from `10.66s` on my laptop to `7.49s`

adriangb:
This required some changes to FastAPI's assertions about OpenAPI schemas, we should review them and confirm they're good before moving forward: https://github.com/tiangolo/fastapi/pull/10272/files

adriangb:
This is what the profiling looks like now:

image

In other words, most of the time is now spent in Rust!

adriangb:
please review

adriangb:
> This required some changes to FastAPI's assertions about OpenAPI schemas, we should review them and confirm they're good before moving forward: https://github.com/tiangolo/fastapi/pull/10272/files

Ended up not, not sure what CI is running (locally I still get failures without these changes) but this PR is the same as `main`.

adriangb:
This is now running in `3.31s` vs. `10.49s` on back-to-back runs on my laptop. So we're pretty close to v1 performance!

This is the current flamegraph:

image

The vast majority of the time is spent in Rust, second comes generating the schema (which is mostly because with generics we create an entire model for every parametrization, that could be changed in theory), the ref stuff and discriminated union stuff is now a distant third.

tests/benchmarks/test_fastapi_startup_generics.py
```diff
@@ -132,7 +133,9 @@ def bench():
     print(f'Python version: {sys.version}')
     if sys.argv[-1] == 'cProfile':
         cProfile.run(
-            'test_fastapi_startup_perf(lambda f: f())', sort='tottime', filename=__file__.strip('.py') + '.cprof'
+            'test_fastapi_startup_perf(lambda f: f())',
+            sort='tottime',
+            filename=Path(__file__).name.strip('.py') + '.cprof',
```
adriangb, line 138: Misc improvement unrelated to this PR, happy to move to another one if requested.

tests/test_dataclasses.py
```diff
@@ -2070,8 +2070,8 @@ class GenericDataclass(Generic[T]):
 
     # verify that generic parameters are showing up in the type ref for generic dataclasses
     # this can probably be removed if the schema changes in some way that makes this part of the test fail
-    assert '[int:' in validator1.core_schema['schema']['schema_ref']
-    assert '[str:' in validator2.core_schema['schema']['schema_ref']
+    assert '[int:' in validator1.core_schema['ref']
+    assert '[str:' in validator2.core_schema['ref']
```
adriangb, line 2074: These are now getting inlined before saving to `TypeAdapter.core_schema`

tests/test_internal.py
```diff
@@ -4,18 +4,27 @@
...
 from pydantic._internal._repr import Representation
 
 
+def remove_metadata(schema: CoreSchema) -> CoreSchema:
+    def inner(s: CoreSchema, recurse: Walk) -> CoreSchema:
+        s = s.copy()
+        s.pop('metadata', None)
+        return recurse(s, inner)
+
+    return walk_core_schema(schema, inner)
```
adriangb, line 20: This is necessary to ignore the cache stuff we're putting in metadata now.

tests/test_internal.py
```diff
@@ -25,24 +34,14 @@
                     cs.int_schema(ref='int'),
                 ],
             ),
-            cs.definitions_schema(
-                cs.list_schema(cs.definition_reference_schema('list_of_ints')),
-                definitions=[
-                    cs.list_schema(cs.definition_reference_schema('int'), ref='list_of_ints'),
-                    cs.int_schema(ref='int'),
-                ],
-            ),
```
adriangb, line 34: Removing the second item in each test since there's no longer two steps, just one.

tests/test_internal.py
```diff
@@ -25,24 +34,14 @@
...
-            cs.list_schema(cs.list_schema(cs.int_schema())),
+            cs.list_schema(cs.list_schema(cs.int_schema(ref='int'), ref='list_of_ints')),
```
adriangb, line 37: We're now preserving the `ref` even when we inline.

tests/test_model_signature.py
```diff
@@ -38,7 +38,11 @@ def test_generic_model_signature():
     class Model(BaseModel, Generic[T]):
         a: T
 
-    sig = signature(Model[int])
+    IntModel = Model[int]
+
+    assert IntModel.model_validate({'a': '1'}).a == 1
+
+    sig = signature(IntModel)
```
adriangb, line 45: This is really weird! If I don't call `model_validate` the signature is different. It seems like some weirdness with generics. @dmontagu any thoughts?

tests/test_root_model.py
```diff
@@ -47,11 +47,6 @@ def check_schema(schema: CoreSchema) -> None:
     # we assume the shape of the core schema here, which is not a guarantee
     # pydantic makes to its users but is useful to check here to make sure
     # we are doing the right thing internally
-    assert schema['type'] == 'definitions'
-    inner = schema['schema']
-    assert inner['type'] == 'definition-ref'
-    ref = inner['schema_ref']  # type: ignore
-    schema = next(s for s in schema['definitions'] if s['ref'] == ref)  # type: ignore
     assert schema['type'] == 'model'
```
adriangb, line 50: These are now getting inlined before saving to `BaseModel.__pydantic_core_schema__`

pydantic/_internal/_core_utils.py
```diff
@@ -189,13 +204,13 @@ def _build_schema_type_to_method(self) -> dict[core_schema.CoreSchemaType, Recur
         return mapping
 
     def walk(self, schema: core_schema.CoreSchema, f: Walk) -> core_schema.CoreSchema:
-        return f(schema.copy(), self._walk)
+        return f(schema, self._walk)
```
davidhewitt, line 207: Does this have behavioural changes (e.g. walking schema now mutates inplace rather than copying)?

pydantic/_internal/_core_utils.py
```diff
@@ -206,11 +221,15 @@ def _handle_other_schemas(self, schema: core_schema.CoreSchema, f: Walk) -> core
 
     def _handle_ser_schemas(self, ser_schema: core_schema.SerSchema, f: Walk) -> core_schema.SerSchema:
         schema: core_schema.CoreSchema | None = ser_schema.get('schema', None)
+        ser_schema = ser_schema
```
davidhewitt, line 224: assignment possibly not necessary?

```suggestion
```

pydantic/_internal/_core_utils.py
```diff
@@ -228,7 +247,7 @@ def handle_definitions_schema(self, schema: core_schema.DefinitionsSchema, f: Wa
             # This means we'd be returning a "trivial" definitions schema that just wrapped the inner schema
             return new_inner_schema
 
-        new_schema = schema.copy()
+        new_schema = schema
```
davidhewitt, line 250: Looks like similar question / concern about removal of copying here. I understand it's probably great for perf so just need to understand implications.

I would be somewhat glad if these could be split into a separate PR (followup?) given they are an orthogonal problem to the schema ref inlining changes.

pydantic/_internal/_core_utils.py
```diff
@@ -416,92 +435,122 @@ def walk_core_schema(schema: core_schema.CoreSchema, f: Walk) -> core_schema.Cor
...
+    """Simplify schema references by:
+    1. Inlining any definitions that are only referenced in one place and are not involved in a cycle.
+    2. Removing any unused `ref` references from schemas.
+    """
+    state = _DefinitionsState(
+        definitions={},
+        ref_counts=defaultdict(int),
+        involved_in_recursion={},
+        current_recursion_ref_count=defaultdict(int),
+    )
```
davidhewitt, line 455: What's the motivation for moving into a state? I guess it makes closures smaller, but causes lots of dictionary lookups?

pydantic/_internal/_core_utils.py
```diff
@@ -416,92 +435,122 @@ def walk_core_schema(schema: core_schema.CoreSchema, f: Walk) -> core_schema.Cor
...
+        current_recursion_ref_count=dict(state['current_recursion_ref_count']),
+    )
+
+    def can_be_inlined(s: core_schema.DefinitionReferenceSchema, ref: str) -> bool:
+        if state['ref_counts'][ref] > 1:
+            return False
+        if state['involved_in_recursion'].get(ref, False):
+            return False
+        if 'serialization' in s:
+            return False
```
davidhewitt, line 532: What's the motivation for this check? Seems somewhat surprising to me that serialization override prevents inlining.

davidhewitt (commented):
I think this PR is quite complex so I would like to move the removal of copying to another PR if possible. Some questions...

pydantic/_internal/_core_utils.py
adriangb: Not really, I think we don’t mutate in place anyway

pydantic/_internal/_core_utils.py
adriangb: Happy to split this PR up

pydantic/_internal/_core_utils.py
adriangb: So that it can be easily stored in a single metadata key

pydantic/_internal/_core_utils.py
adriangb: We use this to modify serialization on a type without globally modifying a type. That is, you can wrap a type with a serializer

adriangb:
I’ll be splitting this PR up into multiple smaller PRs:
- Update pydantic-core
- Remove ref flattening step
- Caches
- Discriminated union improvements
- Remove unused copies

adriangb:
Supplanted by #7565, #7536, #7535, #7529, #7528, #7527, #7524, #7523 and #7522

PR #7475, opened 2023-09-17
Added JsonType to config

### LINKED ISSUE #6348: `ConfigDict.json_schema_extra` has type `Dict[str, object]`
### Initial Checks

### Description

The `json_schema_extra` is defined as `dict[str, object] | JsonSchemaExtraCallable | None` so it's (theoretically) not possible to add a simple string->string mapping.

I would consider this as a bug as the functionality is working but the types are wrong and it should probably be `dict[str, Any]`

### Example Code

_No response_

```Text
pydantic version: 2.0
                 install path: <REDACTED>
               python version: 3.11.4 (main, Jun  7 2023, 00:34:59) [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.4.1-arm64-arm-64bit
```

vladNed:
## Change Summary

Added `JsonType` to be used for `json_schema_extra` property on `ConfigDict`

## Related issue number

Closes #6348 

## Checklist

* [ ] Tests pass on CI


vladNed:
please review

pydantic/config.py
```diff
@@ -25,6 +25,7 @@
 Extra = _Extra()
 
 ExtraValues = Literal['allow', 'ignore', 'forbid']
+JsonType = Dict[str, Union[str, int, float, bool, None, List[Any], Dict[str, Any]]]
```
hramezani, line 28: `JsonType` is not the same as [Samuel suggestion](https://github.com/pydantic/pydantic/issues/6348#issuecomment-1618135577) on the issue.

hramezani:
Thanks @vladNed for this.

Please fix the type for all `json_schema_extra` in the code base. You can find them by searching for `json_schema_extra:`.
for example: 
https://github.com/pydantic/pydantic/blob/6f527f3005d9b87dbd2114e8f50f893f8bb66bc9/pydantic/fields.py#L61

hramezani:
please update

pydantic/config.py
vladNed: checked once again. I added the types as mentioned in the issue.

vladNed:
@hramezani of course. I updated all `json_schema_extra` types

vladNed:
please review

hramezani:
@vladNed I've pushed a commit and fixed the type

vladNed:
@hramezani looks good, still feels that doesn't approach the first suggested type in the story, as that was a recursive type. But.. if you find this will suite I feel that is more explicit this way.

samuelcolvin:
If we do this, (which I'm not sure is worthwhile):
1. It should be call `JsonValue` - it's not a type
2. It should be truly recursive.

Something like:
```py
JsonValue: TypeAlias = Union[None, bool, int, float, str, List[JsonValue], Dict[str, JsonValue]]'
```

Please update.

alexmojaki:
@sydney-runkle can this be closed since it was replaced by https://github.com/pydantic/pydantic/pull/7803?

Issue #5893, opened 2023-05-26
Present Field Name Rather Than Alias When Using Intellisense


BTripp1986:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this feature is missing

### Description

Thank you for making such an amazing tool.  

I have a feature request in regards to aliases. My use case is that I need to use aliases for several of my models to interact with the AWS API. I am representing AWS resources such as:  

```python
class Ec2Instance(BaseModel):
  arn: Optional[str] = Field(alias="Arn")
  """Resource ID for the Instance"""
  instance_type: str = Field(alias="InstanceType")
  """The instance type such as t2.micro, t3.xlarge, etc."""
  ...more fields...

  class Config:
    allow_population_by_field_name = True
```

I need these aliases because this library is used to fetch data via the AWS API, which returns dictionaries with PascalCase keys and the aliases are much more Pythonic than converting each key or using PascalCase for field names.  

In addition to fetching data, these models are also used for creating new resources. With the `Ec2Instance` example above, a user would define all required fields when creating the resource. This library is used by users without expertise of the domain, and the entire point of the library is to make things easier by adding another layer of abstraction where they can use intellisense and inline documentation instead of needing to open the AWS docs in a browser.  

Currently, it is almost there, they can hover over parameters when instantiating an object from the model to see the inline docs, but the issue is that intellisense will suggest the PascalCase aliases instead of the field names. Since the Python convention is to use lower_case_with_underscores, I would like to change this behavior.  

The workaround the I currently have implemented is ugly and violates the DRY principal:  
```python
def aliasing(alias: str) -> str:
    return alias

class Ec2Instance(BaseModel):
  arn: Optional[str] = Field(alias=aliasing("Arn"))
  """Resource ID for the Instance"""
  instance_type: str = Field(alias=aliasing("InstanceType"))
  """The instance type such as t2.micro, t3.xlarge, etc."""
  ...more fields...

  class Config:
    allow_population_by_field_name = True
```

With this workaround, I get the desired behavior. Is this something that could be added to the model Config? I know you are hesitant to increase the parameters in the model Config, but I would argue that this change will help keep code Pythonic in certain circumstances.  

If you are willing to accept a PR then I can make some time to implement this over the next several weeks.

### Affected Components

- [ ] [Compatibility between releases](https://docs.pydantic.dev/changelog/)
[...]

PR #7477, opened 2023-09-18
Update fields.md

etrotta:
Fix a code block under the "VSCode and Pyright users" section
Trivial change to the documentation

Selected Reviewer: @Kludex

### DIFF
--- a/docs/usage/fields.md
+++ b/docs/usage/fields.md
@@ -191,7 +191,7 @@ print(user.model_dump(by_alias=True))  # (2)!
     class User(BaseModel):
         model_config = ConfigDict(populate_by_name=True)
 
-        name: str = Field(..., alias='username')
+        name: str = Field(..., alias=str('username'))
     ```
 
      This is discussed in more detail in [this issue](https://github.com/pydantic/pydantic/issues/5893).


docs/usage/fields.md
```diff
@@ -191,7 +191,7 @@ print(user.model_dump(by_alias=True))  # (2)!
     class User(BaseModel):
         model_config = ConfigDict(populate_by_name=True)
 
-        name: str = Field(..., alias='username')
+        name: str = Field(..., alias=str('username'))
```
hramezani, line 194: we don't need this change.

What is the reason for this change here?

etrotta: If that is not required, then this part of the same document is wrong then?
> To "trick" VSCode into preferring the field name, you can use the `str` function to wrap the alias value

I found it weird that there are two code blocks essentially equal to each other and the code does not matches the description above it

hramezani:
please review

Kludex:
What's the issue for this PR?

etrotta:
> @etrotta What's the issue for this PR?
> 
> I do remember something about this...

I was hoping it would be just a trivial change to the documentation so did not open one, but from that section of the documentation it mentions https://github.com/pydantic/pydantic/issues/5893

samuelcolvin:
this conflicts with main since #7322 was merged. please update.

Issue #7481, opened 2023-09-18
loc_by_alias is not reflecting alias names on integer validations


alkimake:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

if pydantic raises error types such as `less_than_equal` or `greater_than_equal`, `loc_by_alias` does not reflecting to the error definitions or `loc` value. 

Example code will print;

```
[{'ctx': {'le': 10},
  'input': 12,
  'loc': ('integer_value',),
  'msg': 'Input should be less than or equal to 10',
  'type': 'less_than_equal',
  'url': 'https://errors.pydantic.dev/2.3/v/less_than_equal'}]
```

here is the `loc` value should be `IntegerValue` instead of `integer_value`.

There might be many more error types that `loc_by_alias` might not work but I only tried for integer.

### Example Code

```Python
from pprint import pprint

import pydantic


class AliasClass(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(
        loc_by_alias=True,
        populate_by_name=True
    )
    integer_value: int = pydantic.Field(alias='IntegerValue', le=10, ge=1)

try:
    AliasClass(integer_value=12)
except pydantic.ValidationError as exc:
    pprint(exc.errors())
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.3.0
        pydantic-core version: 2.6.3
          pydantic-core build: profile=release pgo=false
               python version: 3.10.12 (main, Jun 26 2023, 20:42:23) [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.5.2-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

samuelcolvin:
`loc_by_alias` is perhaps a slight misnomer, although I don't know of a better name.

It means `loc` should always use the actual key provided in the data, rather than the field name.

I think therefore it's behaving as I would expect it to, although we should definitely improve the docs on this.

alkimake:
Let's ignore the documentation for a moment, if there is no behavioral change why there is this option `loc_by_alias`? I am a bit confused actually.

if `loc` needs to be consistent with the key that is provided, having an option to populate the model with both alias and field name (`populate_by_name`) is enough. and `loc` will always return the given key. (it will ruin our man in the middle but that's my problem for sure). Therefor there is no need for `loc_by_alias` as far as I can see.

hramezani:
[...]

PR #7483, opened 2023-09-18
Improve `loc_by_alias` doc

hramezani:
Closes https://github.com/pydantic/pydantic/issues/7481

### DIFF
--- a/pydantic/_internal/_config.py
+++ b/pydantic/_internal/_config.py
@@ -54,7 +54,7 @@ class ConfigWrapper:
     validate_assignment: bool
     arbitrary_types_allowed: bool
     from_attributes: bool
-    # whether to use the actual key provided in the data(e.g. alias or first alias for "field required" errors) instead of field_names
+    # whether to use the actual key provided in the data (e.g. alias or first alias for "field required" errors) instead of field_names
     # to construct error `loc`s, default `True`
     loc_by_alias: bool
     alias_generator: Callable[[str], str] | None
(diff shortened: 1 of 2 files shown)


hramezani:
please review

pydantic/config.py
```diff
@@ -93,7 +93,7 @@ class ConfigDict(TypedDict, total=False):
     """
 
     loc_by_alias: bool
-    """Whether to use the alias for error `loc`s rather than the field's name. Defaults to `True`."""
+    """Whether to use the actual key provided in the data(e.g. alias) for error `loc`s rather than the field's name. Defaults to `True`."""
```
Kludex, line 96: This may need a new line?
```suggestion
    """Whether to use the actual key provided in the data (e.g. alias) for error `loc`s rather than the field's name. Defaults to `True`."""
```

pydantic/_internal/_config.py
```diff
@@ -54,8 +54,8 @@ class ConfigWrapper:
     validate_assignment: bool
     arbitrary_types_allowed: bool
     from_attributes: bool
-    # whether to use the used alias (or first alias for "field required" errors) instead of field_names
-    # to construct error `loc`s, default True
+    # whether to use the actual key provided in the data(e.g. alias or first alias for "field required" errors) instead of field_names
```
Kludex, line 57:
```suggestion
    # whether to use the actual key provided in the data (e.g. alias or first alias for "field required" errors) instead of field_names
```

Kludex (commented):
Hmm... I'm not sure this simplifies things 👀

hramezani:
> Hmm... I'm not sure this simplifies things 👀

The goal is to make it clear that by enabling the `loc_by_alias` flag you will get `the actual key provided in input data` in the error `loc` instead of `alias` all the time.

ref: https://github.com/pydantic/pydantic/issues/7481#issuecomment-1723071503

samuelcolvin (approved):

Issue #7502, opened 2023-09-19
Missing spaces between enum members in ValidationError


jaap3:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

The error message that is raised when validating input against an enum, joins the members with just a `,` (comma) instead of `, ` (comma space).

I.e. using the example code below the error is:

`Input should be 'red','green' or 'blue'`

I would've expected the error to read:

`Input should be 'red', 'green' or 'blue'`

### Example Code

```Python
import enum
import pydantic

class Color(enum.StrEnum):
    RED = "red"
    GREEN = "green"
    BLUE = "blue"

class ColorModel(pydantic.BaseModel):
    color: Color

ColorModel(color="yellow")
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.3.0
        pydantic-core version: 2.6.3
          pydantic-core build: profile=release pgo=false
                 install path: ~/.pyenv/versions/3.11.3/envs/pydantic2/lib/python3.11/site-packages/pydantic
               python version: 3.11.3 (main, May 24 2023, 16:21:31) [Clang 14.0.3 (clang-1403.0.22.14.1)]
                     platform: macOS-13.5.2-arm64-arm-64bit
     optional deps. installed: ['typing-extensions']
```

PR #7506, opened 2023-09-19
Improve enum error message

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/7502


jaap3:
Thanks for the quick follow up, I was looking into patching this myself but you are much quicker 🙂.

Seems fastapi is checking against this error message. I had a similar test for our code based using pydantic 1, that's actually how I noticed that the formatting was off. Not sure that this failing test means the formatting cannot be changed?

davidhewitt (approved):

PR #7518, opened 2023-09-20
Remove `dicts_mapping.md`, `sequence_iterable.md` and `set_types.md` …

hramezani:
- Related to https://github.com/pydantic/pydantic/issues/7322

Selected Reviewer: @lig


hramezani:
please review

docs/api/standard_library_types.md
```diff
@@ -386,6 +386,412 @@ except ValidationError as e:
...
+
+## Frozenset
+
+### `frozenset`
+
+Allows `list`, `tuple`, `set`, `frozenset`, `deque`, or generators and casts to a `frozenset`.
+When a generic parameter is provided, the appropriate validation is applied to all items of the frozen set.
+
+### `typing.FrozenSet`
+Handled the same as `frozenset` above.
```
Kludex, line 427:
```suggestion
### `typing.FrozenSet`

Handled the same as `frozenset` above.
```

docs/api/standard_library_types.md
```diff
@@ -386,6 +386,412 @@ except ValidationError as e:
...
+
+
+## Support for iterable types
+
+### `typing.Sequence`
+This is intended for use when the provided value should meet the requirements of the `Sequence` protocol, and it is
+desirable to do eager validation of the values in the container. Note that when validation must be performed on the
+values of the container, the type of the container may not be preserved since validation may end up replacing values.
+We guarantee that the validated value will be a valid `typing.Sequence`, but it may have a different type than was
+provided (generally, it will become a `list`).
```
Kludex, line 461:
```suggestion
### `typing.Sequence`

This is intended for use when the provided value should meet the requirements of the `Sequence` protocol, and it is
desirable to do eager validation of the values in the container. Note that when validation must be performed on the
values of the container, the type of the container may not be preserved since validation may end up replacing values.
We guarantee that the validated value will be a valid `typing.Sequence`, but it may have a different type than was
provided (generally, it will become a `list`).
```

docs/api/standard_library_types.md
```diff
@@ -386,6 +386,412 @@ except ValidationError as e:
...
+
+### `typing.Iterable`
+This is intended for use when the provided value may be an iterable that shouldn't be consumed.
+See [Infinite Generators](#infinite-generators) below for more detail on parsing and validation.
+Similar to `typing.Sequence`, we guarantee that the validated result will be a valid `typing.Iterable`,
+but it may have a different type than was provided. In particular, even if a non-generator type such as a `list`
+is provided, the post-validation value of a field of type `typing.Iterable` will be a generator.
```
Kludex, line 468:
```suggestion
### `typing.Iterable`

This is intended for use when the provided value may be an iterable that shouldn't be consumed.
See [Infinite Generators](#infinite-generators) below for more detail on parsing and validation.
Similar to `typing.Sequence`, we guarantee that the validated result will be a valid `typing.Iterable`,
but it may have a different type than was provided. In particular, even if a non-generator type such as a `list`
is provided, the post-validation value of a field of type `typing.Iterable` will be a generator.
```

docs/api/standard_library_types.md
```diff
@@ -386,6 +386,412 @@ except ValidationError as e:
...
+    sequence_of_ints: Sequence[int] = None
+
+
+print(Model(sequence_of_ints=[1, 2, 3, 4]).sequence_of_ints)
+#> [1, 2, 3, 4]
+print(Model(sequence_of_ints=(1, 2, 3, 4)).sequence_of_ints)
+#> (1, 2, 3, 4)
+```
+
+### Strings aren't Sequences
```
Kludex, line 488: Hmm... This is more a concept. But let's keep it here for now.

docs/api/standard_library_types.md
```diff
@@ -386,6 +386,412 @@ except ValidationError as e:
...
+except ValidationError as e:
+    print(e)
+    """
+    1 validation error for ValidatorIterator
+    2
+      Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='a', input_type=str]
+    """
+```
+
+## Dicts and Mapping Types
```
Kludex, line 626:
```suggestion
## Mapping Types
```

docs/api/standard_library_types.md
```diff
@@ -386,6 +386,412 @@ except ValidationError as e:
...
+    """
+```
+
+### TypedDict
+
+!!! note
+    This is a new feature of the Python standard library as of Python 3.8.
+    Because of limitations in `typing.TypedDict` before 3.12, the [typing-extensions](https://pypi.org/project/typing-extensions/)
+    package is required for Python <3.12. You'll need to import `TypedDict` from `typing_extensions` instead of `typing` and will
+    get a build time error if you don't.
```
Kludex, line 688: Is this still true in Pydantic V2?

Kludex (changes requested):

docs/api/standard_library_types.md
hramezani: we will get a `pydantic.errors.PydanticUserError` if we user `typing.TypedDict` with Python < 3.12

hramezani:
please review

Kludex (approved):

PR #7523, opened 2023-09-20
Simplify flatteining and inlining of Coreschema

adriangb:

### DIFF
--- a/tests/test_model_signature.py
+++ b/tests/test_model_signature.py
@@ -38,7 +38,11 @@ def test_generic_model_signature():
     class Model(BaseModel, Generic[T]):
         a: T
 
-    sig = signature(Model[int])
+    IntModel = Model[int]
+
+    assert IntModel.model_validate({'a': '1'}).a == 1
+
+    sig = signature(IntModel)
     assert sig != signature(BaseModel)
     assert _equals(map(str, sig.parameters.values()), ('a: int',))
     assert _equals(str(sig), '(*, a: int) -> None')
--- a/tests/test_model_signature.py
+++ b/tests/test_model_signature.py
@@ -32,17 +32,14 @@ def test_model_signature():
     assert _equals(str(sig), '(*, a: float, b: int = 10) -> None')
 
 
+@pytest.mark.xfail(reason='https://github.com/pydantic/pydantic/pull/7523#discussion_r1331765978')
 def test_generic_model_signature():
     T = TypeVar('T')
 
     class Model(BaseModel, Generic[T]):
         a: T
 
-    IntModel = Model[int]
-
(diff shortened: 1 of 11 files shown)


adriangb:
please review

tests/test_model_signature.py
```diff
@@ -38,7 +38,11 @@ def test_generic_model_signature():
     class Model(BaseModel, Generic[T]):
         a: T
 
-    sig = signature(Model[int])
+    IntModel = Model[int]
+
+    assert IntModel.model_validate({'a': '1'}).a == 1
+
+    sig = signature(IntModel)
```
davidhewitt, line 45: It looks like the cause of this is that the following code is firing on this branch from `_model_construction.py`:

```
    if collect_invalid_schemas(schema):
        set_model_mocks(cls, cls_name)
        return False
```

I suggest we understand this further - smells like a possible bug.

davidhewitt: Let's xfail instead of modify, I will dig into this in parallel.

JrooTJunior:
@dmontagu, @samuelcolvin, this changes breaks compatibility of https://github.com/aiogram/aiogram with pydantic 2.4

```
  File "/home/jrootjunior/work/aiogram/aiogram3/.hatch/dev/lib/python3.11/site-packages/pydantic/_internal/_core_utils.py", line 515, in count_refs
    recurse(state['definitions'][ref], count_refs)
            ~~~~~~~~~~~~~~~~~~~~^^^^^
KeyError: 'aiogram.types.input_text_message_content.InputTextMessageContent:94580207914144'
```

([full traceback](https://gist.github.com/JrooTJunior/7e3f6d3100184e6a1381eab8302eb94d))

You can check it by installing `pip install aiogram==3.1.0` and then just import it - `import aiogram`.
I have temporary workaround - restrict to install pydantic 2.4 with aiogram, so, when you install `pip install aiogram=3.1.1` and then import it - it works as well.

davidhewitt:
Thanks for the report, we're investigating.

Olegt0rr:
Just leave it here

Screenshot 2023-09-25 at 20 15 16

tests/test_root_model.py
```diff
@@ -47,11 +47,6 @@ def check_schema(schema: CoreSchema) -> None:
     # we assume the shape of the core schema here, which is not a guarantee
     # pydantic makes to its users but is useful to check here to make sure
     # we are doing the right thing internally
```
lig, line 49: Remove related comment as well then?

adriangb:
> Just leave it here
> 
> Screenshot 2023-09-25 at 20 15 16

What was the intention of posting this? Do you feel this was an _intentional_ breaking change?

PR #7527, opened 2023-09-20
Add caches for collecting definitions and invalid schemas from a CoreSchema

adriangb:
Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -116,6 +116,12 @@ def collect_definitions(schema: core_schema.CoreSchema) -> dict[str, core_schema
     defs: dict[str, CoreSchema] = {}
 
     def _record_valid_refs(s: core_schema.CoreSchema, recurse: Recurse) -> core_schema.CoreSchema:
+        if 'metadata' in s:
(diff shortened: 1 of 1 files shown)


pydantic/_internal/_core_utils.py
```diff
@@ -157,11 +163,21 @@ def _record_refs(s: core_schema.CoreSchema, recurse: Recurse) -> core_schema.Cor
...
+        if invalid is True:
             invalid_schemas.append(s)
-        return recurse(s, _is_schema_valid)
+        elif invalid is False:
+            return s
+        s = recurse(s, _is_schema_valid)
+        if invalid is True:
+            child_invalid = True
+        metadata['invalid'] = child_invalid
+        return s
```
davidhewitt, line 180: Can you elaborate on the motivation of this change? Maybe needs a comment, it's quite winding logic.

adriangb: The `invalid` thing already existed. What this is doing is marking a schema as "there are no invalid schemas in this subtree" (which is the most common case) so that we can skip recursing into it after the first time we recurse through it.

adriangb: I'll wait to make that change until #7529 is merged so I can re-use `post_process_schema` and such

davidhewitt (approved):

PR #7535, opened 2023-09-21
Cache invalid CoreSchema discovery

adriangb:

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -42,6 +42,14 @@ _LIST_LIKE_SCHEMA_WITH_ITEMS_TYPES = {'list', 'tuple-variable', 'set', 'frozense
 _DEFINITIONS_CACHE_METADATA_KEY = 'pydantic.definitions_cache'
 
 NEEDS_APPLY_DISCRIMINATED_UNION_METADATA_KEY = 'pydantic.internal.needs_apply_discriminated_union'
+"""Used to mark a schema that has a discriminated union that needs to be checked for validity at the end of
+schema building because one of it's members refers to a definition that was not yet defined when the union
+was first encountered.
+"""
+HAS_INVALID_SCHEMAS_METADATA_KEY = 'pydantic.internal.invalid'
+"""Used to mark a schema that is invalid because it refers to a definition that was not yet defined when the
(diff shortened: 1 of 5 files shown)


adriangb:
please review

pydantic/_internal/_core_utils.py
```diff
@@ -136,20 +136,13 @@ def _record_valid_refs(s: core_schema.CoreSchema, recurse: Recurse) -> core_sche
 
 def define_expected_missing_refs(
     schema: core_schema.CoreSchema, allowed_missing_refs: set[str]
-) -> core_schema.CoreSchema:
+) -> tuple[core_schema.CoreSchema, bool]:
```
davidhewitt, line 139: Potentially document or comment what the bool is here, I had to look at the callsite.

One option to avoid `bool` (and intermediate `tuple`, if you care about micro-optimizing) is to return `core_schema.CoreSchema | None` and return `None` in the case when there are no missing refs.

pydantic/_internal/_generate_schema.py
```diff
@@ -644,14 +653,21 @@ def _get_first_two_args_or_any(self, obj: Any) -> tuple[Any, Any]:
         return args[0], args[1]
 
     def _post_process_generated_schema(self, schema: core_schema.CoreSchema) -> core_schema.CoreSchema:
-        schema.setdefault('metadata', {})[
-            NEEDS_APPLY_DISCRIMINATED_UNION_METADATA_KEY
-        ] = self._needs_apply_discriminated_union
+        metadata = schema.setdefault('metadata', {})
+        metadata[NEEDS_APPLY_DISCRIMINATED_UNION_METADATA_KEY] = self._needs_apply_discriminated_union
+        metadata['invalid'] = self._has_invalid_schema
```
davidhewitt, line 658: Should `'invalid'` maybe also become a constant?

pydantic/_internal/_core_utils.py
adriangb: Great call, done

pydantic/_internal/_generate_schema.py
adriangb: Done

davidhewitt (approved):

PR #7536, opened 2023-09-21
Replace dict.get and dict.setdefault with more verbose versions in CoreSchema building hot paths

adriangb:
I've benchmarked this to be slightly faster. Some of these, especially the ones during walking of core schemas, get hit *a lot*.

Selected Reviewer: @Kludex


adriangb:
please review

whisller:
@adriangb would you be able to include those benchmarks please? I am curious what was the difference in both approaches

adriangb:
Very minor. Mainly creating a new empty dict

PR #7542, opened 2023-09-21
Allow access to `field_name` and `data` in all validators if there is data and a field name

samuelcolvin:
Companion to pydantic/pydantic-core#980.

Also results in significant simplification of the types in pydantic-core.

**Note:** because of this, some type annotations and function from `pydantic_core.core_schema` have been changed.

Fix #7448
fix #6794
Fix #6345

TODO:
* [x] release pydantic-core with the PR merged
* [x] update the docs
* [x] check the above issues really are solved by this

Selected Reviewer: @adriangb

PR #7550, opened 2023-09-21
Update pydantic-core to 2.10.0

adriangb:
(waiting for release)


adriangb:
Looks like we'll need to update FastAPI as well cc @tiangolo

samuelcolvin:
replaced by #7542 which cherry-picks some of these commits.

PR #7559, opened 2023-09-22
Remove the `model_config.md` page

Kludex:
- Related to #7322


docs/concepts/alias.md
```diff
@@ -0,0 +1,36 @@
+## Alias Precedence
```
hramezani, line 1: As it is `alias` concept page, don't we need to add a description about what is an alias?

hramezani (approved):

PR #7565, opened 2023-09-22
Allow disabling CoreSchema validation for faster startup times

adriangb:
Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_core_utils.py
+++ b/pydantic/_internal/_core_utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from collections import defaultdict
 from typing import (
     Any,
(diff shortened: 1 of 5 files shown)


hramezani (approved):

mrtolkien:
Just wanna say that using `PYDANTIC_SKIP_VALIDATING_CORE_SCHEMAS=true` **combined with** `model_config = ConfigDict(defer_build=True)` in my AWS Lambda reduced startup time from 4s to 1s.

I have a common library of Pydantic objects for multiple Lambdas, which mean a given one only uses a small subset of them despite them being imported with the library.

I'm not exactly sure what both those things do tbh, but it worked for my use-case 😬

MarkusSintonen:
@adriangb would it be possible to change the logic here so that the env would be boolified? So we could use `PYDANTIC_SKIP_VALIDATING_CORE_SCHEMAS=false` to do the validation eg in CI but omit it in local tests (via pytest_env in pyproject would use `PYDANTIC_SKIP_VALIDATING_CORE_SCHEMAS = "true"`)

PR #7589, opened 2023-09-25
revert plugin performance regression

samuelcolvin:
Avoid loading plugins, and thereby iterating over `importlib_metadata.distributions()` ever time a model is created.

Instead we only look for plugins is we haven't looked before, or loading plugins previously failed.

Selected Reviewer: @lig

### DIFF
--- a/pydantic/plugin/_loader.py
+++ b/pydantic/plugin/_loader.py
@@ -35,8 +35,8 @@ def get_plugins() -> Iterable[PydanticPluginProtocol]:
         # this happens when plugins themselves use pydantic, we return no plugins
         return ()
     elif _plugins is None:
-        # plugins already loaded
         _plugins = {}
+        # set _loading_plugins so any plugins that use pydantic don't themselves use plugins
         _loading_plugins = True
         try:
             for dist in importlib_metadata.distributions():
--- a/pydantic/plugin/_loader.py
+++ b/pydantic/plugin/_loader.py
@@ -33,7 +33,7 @@ def get_plugins() -> Iterable[PydanticPluginProtocol]:
     global _plugins, _loading_plugins
     if _loading_plugins:
         # this happens when plugins themselves use pydantic, we return no plugins
-        return {}.values()
+        return ()
     elif _plugins is None:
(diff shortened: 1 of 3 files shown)


samuelcolvin:
please review

pydantic/plugin/_loader.py
```diff
@@ -18,36 +18,41 @@
...
-                error_type = e.__class__.__name__
-                warnings.warn(
-                    f'{error_type} while loading the `{entry_point.name}` Pydantic plugin, this could be caused '
-                    f'by a circular import issue (e.g. the plugin importing Pydantic), Pydantic will attempt to '
-                    f'import this plugin again each time a Pydantic validators is created. {e}'
-                )
+    global _plugins, _loading_plugins
+    if _loading_plugins:
+        # this happens when plugins themselves use pydantic, we return no plugins
+        return {}.values()
```
lig, line 36: Just an empty tuple should be enough as `Iterable`

tests/plugin/example_plugin.py
```diff
@@ -5,6 +5,10 @@ class MyModel(BaseModel):
     x: int
 
 
+m = MyModel(x='10')
+if m.x != 10:
+    raise Exception('m.x should be 10')
```
lig, line 10: `assert` ?

lig (approved):
just a couple of nuances...

pydantic/plugin/_loader.py
samuelcolvin: yes, but I wanted to keep the type exactly the same.

tests/plugin/example_plugin.py
samuelcolvin: I didn't want any confusion from pytest's assert rewriting logic.

pydantic/plugin/_loader.py
samuelcolvin: I'll add a comment.

samuelcolvin: I think you're right, it doesn't need to be dict values.

Issue #7611, opened 2023-09-25
Recurse KeyError when using several BaseModels 


hoorelbeke-jimmy:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

The following code is working well with pydantic 2.3, but after upgrading it to 2.4 it fails with the following error:

```bash
Traceback (most recent call last):
  File "/home/ahoorelbeke/generator_v1_detailed/src/generator_source/parameters/parameters_equipment_design.py", line 23, in <module>
    class E(BaseModel):
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_model_construction.py", line 184, in __new__
    complete_model_class(
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_model_construction.py", line 495, in complete_model_class
    schema = apply_discriminators(simplify_schema_references(schema))
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 519, in simplify_schema_references
    schema = walk_core_schema(schema, count_refs)
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 439, in walk_core_schema
    return f(schema, _dispatch)
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 515, in count_refs
    recurse(state['definitions'][ref], count_refs)
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 215, in walk
    return f(schema, self._walk)
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 503, in count_refs
    return recurse(s, count_refs)
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 218, in _walk
    schema = self._schema_type_to_method[schema['type']](schema, f)
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 227, in _handle_other_schemas
    schema['schema'] = self.walk(sub_schema, f)  # type: ignore
  File "/home/ahoorelbeke/venv/lib/python3.10/site-packages/pydantic/_internal/_core_utils.py", line 215, in walk
[...]

PR #7619, opened 2023-09-25
✅ Add test for the case from #7611

lig:
Not a fix for #7611

### DIFF
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -2612,3 +2612,25 @@ def test_custom_exception_handler():
         MyModel(**data)
 
     assert len(traceback_exceptions) == 1
+
+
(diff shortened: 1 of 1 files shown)


tests/test_edge_cases.py
```diff
@@ -2612,3 +2612,25 @@ def __exit__(self, _exception_type, exception, exception_traceback):
...
 
     assert len(traceback_exceptions) == 1
+
+
+def test_recursive_walk_fails_on_double_diamond_composition():
+    class A(BaseModel):
+        pass
+
+    class B(BaseModel):
+        a_1: A = None
```
adriangb, line 2622: Can we minimize this? I don’t think the default values are relevant.

PR #7625, opened 2023-09-26
Fix an issue where JSON schema extras weren't JSON encoded

dmontagu:
Closes https://github.com/tiangolo/fastapi/issues/10321

As far as I can tell, in v1, we would json encode anything extra (which included the examples). So I think this is restoring intended behavior (and as far as I can imagine is objectively better behavior in all circumstances where it might affect things).

@Kludex for reference

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -33,7 +33,7 @@ from typing import (
 )
 from warnings import warn
 
-from pydantic_core import CoreSchema, PydanticUndefined, core_schema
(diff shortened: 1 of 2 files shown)


dmontagu:
Note this doesn't affect the result of a mutating `json_schema_extra`. To make that work would be a bit more complex because it makes in-place changes, so a straightforward call to `to_jsonable_python` won't work out of the box as that won't mutate the input. But it could be made to work, I'm just not sure it's worth the overhead it would add.

Kludex (approved):

PR #7646, opened 2023-09-26
Fix schema references in discriminated unions

adriangb:
Fixes https://github.com/pydantic/pydantic/issues/7611#issuecomment-1735538979 (I hope)

### DIFF
--- a/pydantic/_internal/_discriminated_union.py
+++ b/pydantic/_internal/_discriminated_union.py
@@ -6,7 +6,12 @@ from pydantic_core import CoreSchema, core_schema
 
 from ..errors import PydanticUserError
 from . import _core_utils
-from ._core_utils import NEEDS_APPLY_DISCRIMINATED_UNION_METADATA_KEY, CoreSchemaField, collect_definitions
+from ._core_utils import (
+    NEEDS_APPLY_DISCRIMINATED_UNION_METADATA_KEY,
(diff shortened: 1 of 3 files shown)


adriangb:
@Fokko can you try this branch? I tried it locally and can import files I could not before but don't know how to get your unit tests to run (missing env vars, etc.)

Fokko:
@adriangb Thanks for the quick turnaround, this fixed it: https://github.com/apache/iceberg/pull/8646

tnx

dmontagu (approved):

dmontagu:
I think the following example reproduces the issue that this PR is trying to solve:

```python

from pydantic import BaseModel, Field, RootModel


class Model1(BaseModel):
    kind: Literal['1'] = '1'
    two: 'Model2 | None'


class Model2(BaseModel):
    kind: Literal['2'] = '2'
    one: Model1 | None


class Root1(RootModel[Model1]):
    pass


class Root2(RootModel[Model2]):
    pass


class Outer(BaseModel):
    first: Annotated[Root1 | Root2, Field(discriminator='kind')]
    second: Annotated[Root1 | Root2, Field(discriminator='kind')]
```

Fokko:
@dmontagu Yes, that looks like it. We use the discriminator to distinguish between the different versions.

PR #7647, opened 2023-09-26
Fix bug with JSON schema for sequence of discriminated union

dmontagu:
Discovered while working with @adriangb on discriminated union bugs


adriangb (approved):

adriangb:
Do we have similar issues for `Mapping` and co?

PR #7648, opened 2023-09-26
Add test for complicated scenario that was producing schema generation errors

dmontagu:
The test added in this PR fails before https://github.com/pydantic/pydantic/pull/7646, and I think is a fairly minimal example of the issue resolved in that PR.

### DIFF
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -2640,11 +2640,11 @@ def test_recursive_walk_fails_on_double_diamond_composition():
 def test_recursive_root_models_in_discriminated_union():
     class Model1(BaseModel):
         kind: Literal['1'] = '1'
-        two: 'Model2 | None'
+        two: Optional['Model2']
 
     class Model2(BaseModel):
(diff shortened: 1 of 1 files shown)


tests/test_edge_cases.py
```diff
@@ -2634,3 +2635,71 @@ class E(BaseModel):
...
+        pass
+
+    class Root2(RootModel[Model2]):
+        pass
+
+    class Outer(BaseModel):
+        first: Annotated[Root1 | Root2, Field(discriminator='kind')]
+        second: Annotated[Root1 | Root2, Field(discriminator='kind')]
+
+    assert Outer.model_json_schema() == {
```
dmontagu, line 2659: Prior to #7646, this test fails to even create the `Outer` class due to a `SchemaError`. I added the `model_json_schema()` call though to make sure the generated JSON schema is also correct (which it appears to be), as an additional layer of protection against bugs that might get introduced in the future.

adriangb (approved):

PR #7675, opened 2023-09-27
Update Pydantic V1 to 1.10.13

hramezani:


pydantic/v1/datetime_parse.py
```diff
@@ -147,7 +147,7 @@ def parse_time(value: Union[time, StrBytesIntFloat]) -> time:
     number = get_numeric(value, 'time')
     if number is not None:
         if number >= 86400:
-            # doesn't make sense since the time loop back around to 0
+            # doesn't make sense since the time time loop back around to 0
```
samuelcolvin, line 150: typo

lig: We need to fix it in v1 first:)

lig (approved):
👍🏻

pydantic/v1/datetime_parse.py
hramezani: It got merged. we need to fix it in `v1`

Issue #7689, opened 2023-09-28
Argument 'bases' has incorrect type


joostlek:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

So last night 1.10.13 got released. And apparently it breaks on runtime,

The context is that there are some integrations in HomeAssistant that use pydantic, and they all seem to raise this issue when 1.10.13 is installed. This isn't raised in 1.10.12 (https://github.com/home-assistant/core/pull/101044). Linked issue on the HA repo (https://github.com/home-assistant/core/issues/101042).

```
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/loader.py", line 818, in get_component
    ComponentProtocol, importlib.import_module(self.pkg_path)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/src/homeassistant/homeassistant/components/radio_browser/__init__.py", line 5, in <module>
    from radios import RadioBrowser, RadioBrowserError
  File "/usr/local/lib/python3.11/site-packages/radios/__init__.py", line 8, in <module>
    from .models import Country, Language, Station, Stats, Tag
  File "/usr/local/lib/python3.11/site-packages/radios/models.py", line 12, in <module>
[...]

PR #7696, opened 2023-09-28
Fix Cython build

hramezani:
fixes https://github.com/pydantic/pydantic/issues/7689

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,7 +268,7 @@ jobs:
     - name: set up python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: '3.11'
 
     - name: install
       run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -268,7 +268,7 @@ jobs:
     - name: set up python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: '3.11'
```
hramezani, line 271: I've updated this because of [failure on Python 3.10](https://github.com/pydantic/pydantic/actions/runs/6343083711/job/17230801075?pr=7696)

hramezani:
Please review

samuelcolvin:
closing as no updates on this. Happy to reconsider if we get more reports and more details.

cp2004:
PiWheels removed the broken wheel builds, and home assistant patched their wheel builds (https://github.com/home-assistant/core/pull/101976) so the original reporters 'effectively stopped' running into this issue would be my understanding. I would think the root of the issue still needs to be fixed however?

PR #7719, opened 2023-10-02
fix: update typevar handling when default is not set

pmmmwh:
## Change Summary

I believe the behaviour change in #7606 is raising false positives when the `default` argument is supported but not specified.

Ref: https://peps.python.org/pep-0696/#implementation

## Related issue number

#7606

## Checklist

* [x] Tests pass on CI

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -1449,17 +1449,14 @@ class GenerateSchema:
 
         bound = typevar.__bound__
         constraints = typevar.__constraints__
-        not_set = object()
-        default = getattr(typevar, '__default__', not_set)
-        if default is None:
-            default = not_set
+        default = getattr(typevar, '__default__', None)
 
-        if (bound is not None) + (len(constraints) != 0) + (default is not not_set) > 1:
+        if (bound is not None) + (len(constraints) != 0) + (default is not None) > 1:
             raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


pmmmwh:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -1451,6 +1451,8 @@ def _unsubstituted_typevar_schema(self, typevar: typing.TypeVar) -> core_schema.
         constraints = typevar.__constraints__
         not_set = object()
         default = getattr(typevar, '__default__', not_set)
+        if default is None:
+            default = not_set
```
adriangb, line 1455: Can the default not be `None`? How does this differentiate `default=None` from no default?

pmmmwh: According to the PEP:
- when default is unset, `__default__` would be `None`
- if `default=None`, `__default__` should be `NoneType`

adriangb: Can we get rid of our internal `unset` thing then?

pmmmwh: Hmm, yes - let me try and see if that works

pmmmwh: Done :)

adriangb (approved):

PR #7802, opened 2023-10-11
Use generated alias for aliases that are not specified otherwise

alexmojaki:
## Change Summary

All three aliases now fallback to the alias generator if there is one and no other value was specified for that alias.

## Related issue number

Closes https://github.com/pydantic/pydantic/issues/7781

## Checklist

* [x] Tests pass on CI

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_generate_schema.py
+++ b/pydantic/_internal/_generate_schema.py
@@ -997,21 +997,20 @@ class GenerateSchema:
         # apply alias generator
         alias_generator = self._config_wrapper.alias_generator
         if alias_generator:
-            if field_info.alias_priority is None or field_info.alias_priority <= 1:
+            if field_info.alias_priority is None or field_info.alias_priority <= 1 or field_info.alias is None:
                 alias = alias_generator(name)
                 if not isinstance(alias, str):
                     raise TypeError(f'alias_generator {alias_generator} must return str, not {alias.__class__}')
+                if field_info.alias is None:
+                    if field_info.serialization_alias is None:
+                        field_info.serialization_alias = alias
+                    if field_info.validation_alias is None:
+                        field_info.validation_alias = alias
+                else:
+                    field_info.serialization_alias = alias
+                    field_info.validation_alias = alias
+                    field_info.alias_priority = 1
                 field_info.alias = alias
-                field_info.validation_alias = alias
-                field_info.serialization_alias = alias
-                field_info.alias_priority = 1
-            elif field_info.alias is None:
-                alias = alias_generator(name)
-                if not isinstance(alias, str):
-                    raise TypeError(f'alias_generator {alias_generator} must return str, not {alias.__class__}')
-                field_info.alias = alias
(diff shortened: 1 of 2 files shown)


alexmojaki:
please review

pydantic/_internal/_generate_schema.py
```diff
@@ -996,14 +996,22 @@ def json_schema_update_func(schema: CoreSchemaOrField, handler: GetJsonSchemaHan
...
+                field_info.validation_alias = alias
+                field_info.serialization_alias = alias
+                field_info.alias_priority = 1
+            elif field_info.alias is None:
+                alias = alias_generator(name)
+                if not isinstance(alias, str):
+                    raise TypeError(f'alias_generator {alias_generator} must return str, not {alias.__class__}')
+                field_info.alias = alias
+                field_info.serialization_alias = field_info.serialization_alias or alias
+                field_info.validation_alias = field_info.validation_alias or alias
```
dmontagu, line 1014: I think in principle you can use `field_info.serialization_alias = ''`, and we've had that request before, but with these `or`s that would get replaced. So I think this should be changed to something like:
```python
if field_info.serialization_alias is None:
    field_info.serialization_alias = alias
```
and similar for `validation_alias`.

I recognize you could maybe fit this in one line but I think it gets long enough with an `is None` check instead of just an `or` that it's not worth it any more.

pydantic/_internal/_generate_schema.py
```diff
@@ -996,14 +996,22 @@ def json_schema_update_func(schema: CoreSchemaOrField, handler: GetJsonSchemaHan
...
-            field_info.serialization_alias = alias
-            field_info.alias_priority = 1
+        if alias_generator:
+            if field_info.alias_priority is None or field_info.alias_priority <= 1:
+                alias = alias_generator(name)
+                if not isinstance(alias, str):
+                    raise TypeError(f'alias_generator {alias_generator} must return str, not {alias.__class__}')
+                field_info.alias = alias
+                field_info.validation_alias = alias
+                field_info.serialization_alias = alias
```
dmontagu, line 1006: These 6 lines seem exactly copied between the two if branches. I think it would be better to not duplicate them, and just include the `field_info.alias_priority = 1` line when `field_info.alias is not None`. So this would end up looking more like:

```python
            if field_info.alias_priority is None or field_info.alias_priority <= 1 or field_info.alias is None:
                alias = alias_generator(name)
                if not isinstance(alias, str):
                    raise TypeError(f'alias_generator {alias_generator} must return str, not {alias.__class__}')
                field_info.alias = alias
                field_info.validation_alias = alias
                field_info.serialization_alias = alias
                if field_info.alias is not None:
                    field_info.alias_priority = 1
```

dmontagu (changes requested):
Seems good, two small change requests

sydney-runkle (approved):
LGTM 👍 Thanks so much for the contribution! 🌟

dmontagu (approved):

PR #7806, opened 2023-10-12
Allow plugins to catch non ValidationError errors

adriangb:
Selected Reviewer: @Kludex

### DIFF
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -5,7 +5,7 @@ from typing import Any, Generator
 
 from pydantic_core import ValidationError
 
-from pydantic import BaseModel
+from pydantic import BaseModel, field_validator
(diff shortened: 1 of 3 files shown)


tests/test_plugins.py
```diff
@@ -58,7 +60,7 @@ class Model(BaseModel, plugin_settings={'observe': 'all'}):
 
 def test_on_validate_json_on_error() -> None:
     class CustomOnValidateJson:
-        def enter(
+        def on_enter(
```
adriangb, line 63: None of these were getting called 😢. In my opinion, this is a strong argument for using a single method with a context manager instead of multiple optional callbacks.

adriangb:
please review

samuelcolvin (approved):

PR #7861, opened 2023-10-18
Make path of the item to validate available in plugin

hramezani:
## Checklist

* [x] Tests pass on CI

### DIFF
--- a/docs/concepts/plugins.md
+++ b/docs/concepts/plugins.md
@@ -89,6 +89,7 @@ class Plugin(PydanticPluginProtocol):
     def new_schema_validator(
         self,
         schema: CoreSchema,
+        source_type: str,
         type_path: str,
         item_type: str,
         config: Union[CoreConfig, None],
--- a/docs/concepts/plugins.md
+++ b/docs/concepts/plugins.md
@@ -89,6 +89,7 @@ class Plugin(PydanticPluginProtocol):
     def new_schema_validator(
         self,
         schema: CoreSchema,
+        type_path: str,
         config: Union[CoreConfig, None],
         plugin_settings: Dict[str, object],
     ) -> NewSchemaReturns:
--- a/docs/concepts/plugins.md
+++ b/docs/concepts/plugins.md
@@ -58,10 +58,12 @@ Let's see an example of a plugin that _wraps_ the `validate_python` method of th
 from typing import Any, Dict, Optional, Union
 
 from pydantic_core import CoreConfig, CoreSchema, ValidationError
+from typing_extensions import Literal
 
 from pydantic.plugin import (
     NewSchemaReturns,
     PydanticPluginProtocol,
+    SchemaTypePath,
     ValidatePythonHandlerProtocol,
 )
 
@@ -89,9 +91,15 @@ class Plugin(PydanticPluginProtocol):
     def new_schema_validator(
         self,
         schema: CoreSchema,
-        source_type: str,
-        type_path: str,
-        item_type: str,
+        schema_type: str,
+        schema_type_path: SchemaTypePath,
+        schema_kind: Literal[
+            'BaseModel',
+            'TypeAdapter',
+            'dataclass',
+            'create_model',
+            'validate_call',
+        ],
         config: Union[CoreConfig, None],
         plugin_settings: Dict[str, object],
     ) -> NewSchemaReturns:
--- a/docs/concepts/plugins.md
+++ b/docs/concepts/plugins.md
@@ -91,7 +91,7 @@ class Plugin(PydanticPluginProtocol):
     def new_schema_validator(
         self,
         schema: CoreSchema,
-        schema_type: str,
+        schema_type: Any,
         schema_type_path: SchemaTypePath,
         schema_kind: Literal[
             'BaseModel',
--- a/docs/concepts/plugins.md
+++ b/docs/concepts/plugins.md
@@ -58,11 +58,11 @@ Let's see an example of a plugin that _wraps_ the `validate_python` method of th
 from typing import Any, Dict, Optional, Union
 
 from pydantic_core import CoreConfig, CoreSchema, ValidationError
-from typing_extensions import Literal
 
 from pydantic.plugin import (
     NewSchemaReturns,
     PydanticPluginProtocol,
+    SchemaKind,
     SchemaTypePath,
     ValidatePythonHandlerProtocol,
 )
@@ -93,13 +93,7 @@ class Plugin(PydanticPluginProtocol):
         schema: CoreSchema,
         schema_type: Any,
         schema_type_path: SchemaTypePath,
-        schema_kind: Literal[
-            'BaseModel',
-            'TypeAdapter',
-            'dataclass',
-            'create_model',
-            'validate_call',
-        ],
+        schema_kind: SchemaKind,
         config: Union[CoreConfig, None],
         plugin_settings: Dict[str, object],
     ) -> NewSchemaReturns:
--- a/docs/concepts/plugins.md
+++ b/docs/concepts/plugins.md
@@ -90,6 +90,7 @@ class Plugin(PydanticPluginProtocol):
         self,
         schema: CoreSchema,
         type_path: str,
+        item_type: str,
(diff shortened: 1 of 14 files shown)


hramezani:
> ```python
> def foo():
>     class Model(BaseModel): pass
> 
> def bar():
>     class Model(BaseModel): pass
> 
> foo()
> bar()
> ```

I've added a test for this.

hmm, this is also an idea to make the class itself available on plugin. The current solution suggested by @samuelcolvin. So, what do you think here @samuelcolvin ?

pydantic/_internal/_model_construction.py
```diff
@@ -451,6 +455,7 @@ def complete_model_class(
         config_wrapper: The config wrapper instance.
         raise_errors: Whether to raise errors.
         types_namespace: Optional extra namespace to look for types in.
+        is_dynamic_model: Whether the model is a dynamic model (function is called from `create_model`).
```
samuelcolvin, line 458: no need for this argument, and logic here. Instead put the logic in `create_model`

pydantic/_internal/_dataclasses.py
```diff
@@ -170,7 +170,7 @@ def __init__(__dataclass_self__: PydanticDataclass, *args: Any, **kwargs: Any) -
 
     cls.__pydantic_core_schema__ = schema
     cls.__pydantic_validator__ = validator = create_schema_validator(
-        schema, core_config, config_wrapper.plugin_settings
+        schema, f'{cls.__module__}:{cls.__qualname__}', core_config, config_wrapper.plugin_settings
```
samuelcolvin, line 173: I think it would be useful to have a function in the plugin code:

```py
def build_type_path(module: str, name: str) -> str:
    return f'{module}:{name}'
```

pydantic/main.py
```diff
@@ -1425,6 +1425,7 @@ def create_model(
     if resolved_bases is not __base__:
         ns['__orig_bases__'] = __base__
     namespace.update(ns)
+    kwds['is_dynamic_model'] = True
```
samuelcolvin, line 1428: I don't think we need `is_dynamic_model` here, I think the `__module__: str = __name__,` parameter should work.

If for some reason that doesn't work, we'll need to use `f = sys._getframe(3)` here.

pydantic/plugin/__init__.py
```diff
@@ -27,6 +27,7 @@ class PydanticPluginProtocol(Protocol):
     def new_schema_validator(
         self,
         schema: CoreSchema,
+        path: str,
```
samuelcolvin, line 30: rename to `type_path` for clarity.

pydantic/plugin/_schema_validator.py
```diff
@@ -18,7 +18,7 @@
 
 
 def create_schema_validator(
-    schema: CoreSchema, config: CoreConfig | None = None, plugin_settings: dict[str, Any] | None = None
+    schema: CoreSchema, path: str, config: CoreConfig | None = None, plugin_settings: dict[str, Any] | None = None
```
samuelcolvin, line 21: and everywhere here.

pydantic/type_adapter.py
```diff
@@ -245,7 +245,10 @@ def __init__(self, type: Any, *, config: ConfigDict | None = None, _parent_depth
         try:
             validator = _getattr_no_parents(type, '__pydantic_validator__')
         except AttributeError:
-            validator = create_schema_validator(core_schema, core_config, config_wrapper.plugin_settings)
+            f = sys._getframe(1)
+            validator = create_schema_validator(
+                core_schema, f'{f.f_globals["__name__"]}:{type}', core_config, config_wrapper.plugin_settings
+            )
```
samuelcolvin, line 251: instead of this, I think it would be better to add a parameter

```py
module: str = __name__,
```
argument if that works.

samuelcolvin (commented):
otherwise LGTM.

please update.

pydantic/type_adapter.py
hramezani: If we have `TypeAdapter(List[str])` in `test.test_plugins`:

- Current `type_path` = `tests.test_plugins:typing.List[str]`
- By your suggested change, `type_path` = `pydantic.type_adapter:typing.List[str]`

Should we do the change?

samuelcolvin: no shit, no I see the problem, using `__name__` doesn't work - either here or in `create_model` - you will need to use `get_frame()`.

hramezani:
@samuelcolvin I've addressed your comments.

Also, I've pushed a new commit and added `item_type` param that contains the type of the item to be validated. e.g `BaseModel`, `create_model`, `dataclass`, ...


please review

hramezani:
I've added `source_type` as well.

Here are the params:

- source_type: str -> The model, dataclass or type itself.
- type_path: str -> It's path e.g. `foo.bar.MyModel`
- item_type: str -> it's type e.g `BaseModel`, `create_model`, `dataclass`

TBH, I am not happy with param names. Any suggestions?

pydantic/plugin/__init__.py
```diff
@@ -27,6 +27,9 @@ class PydanticPluginProtocol(Protocol):
     def new_schema_validator(
         self,
         schema: CoreSchema,
+        source_type: str,
+        type_path: str,
+        item_type: str,
```
samuelcolvin, line 32:
```suggestion
        schema_type: Any,
        schema_type_path: tuple[str, str],
        schema_kind: Literal['BaseModel', 'create_mode', 'validate_call', 'TypeAdapter'],
```
Maybe I'm confused, but `source_type` should be `Any` I think?

Also:
* `item_type` is a very confusing name, better to use `schema_kind` and make it a literal
* I think it's more pythonic/correct ot make `type_path`, now `schema_type_path` an enum of `module`, `name` - you could even used a `NamedTuple`? Sorry for confusion.

pydantic/plugin/_schema_validator.py
```diff
@@ -17,8 +17,18 @@
...
+    return f'{module}:{name}'
+
+
 def create_schema_validator(
-    schema: CoreSchema, config: CoreConfig | None = None, plugin_settings: dict[str, Any] | None = None
+    schema: CoreSchema,
+    source_type: Any,
+    module: str,
+    type_name: str,
+    item_type: str,
```
samuelcolvin, line 29: rename and change type types here to. You can leave `schema_type_module` and `schema_type_name` as separate parameters if you like.

samuelcolvin (changes requested):
Sorry, some more changes to make it clearer. Hope this is makes sense to you.

hramezani:
@samuelcolvin I've addressed your comments. please review

docs/concepts/plugins.md
```diff
@@ -89,6 +91,15 @@ class Plugin(PydanticPluginProtocol):
     def new_schema_validator(
         self,
         schema: CoreSchema,
+        schema_type: str,
```
samuelcolvin, line 94:
```suggestion
        schema_type: Any,
```

???

pydantic/_internal/_model_construction.py
```diff
@@ -76,6 +76,7 @@ def __new__(
         namespace: dict[str, Any],
         __pydantic_generic_metadata__: PydanticGenericMetadata | None = None,
         __pydantic_reset_parent_namespace__: bool = True,
+        cls_module: str | None = None,
```
samuelcolvin, line 79: Are you sure it's valid to pass `cls_module` as a kwarg to `__new__`?

samuelcolvin: if module has to be passed this way for `create_model`, maybe better to use `_cls_module` to avoid clashes with config?

samuelcolvin (commented):
otherwise LGTM.

please update.

pydantic/_internal/_model_construction.py
hramezani: Yes, it's there for `create_model`. I renamed all of them to `_cls_module`

hramezani:
Ok, @samuelcolvin I've addressed your comments

Please review again

samuelcolvin (approved):
LGTM

Issue #7895, opened 2023-10-21
Inconsistent `model_dump` behavior with dataclasses and `field(init=False)`


treykeown:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Models will only dump correctly if a field with `init=False` uses `default_factory`.

### Example Code

```Python
import uuid
from pydantic import RootModel
from pydantic.dataclasses import dataclass
from dataclasses import field

@dataclass
class Works:
    id: str = field(init=False, default_factory=lambda: str(uuid.uuid4()))

@dataclass
class DoesntWork1:
    id: str = field(init=False, default="DEFAULT_ID")

@dataclass
class DoesntWork2:
    id: str = field(init=False)

    def __post_init__(self):
        self.id = str(uuid.uuid4())

w = Works()
dw1 = DoesntWork1()
dw2 = DoesntWork2()

print(RootModel[Works](w).model_dump_json())
print(RootModel[DoesntWork1](dw1).model_dump_json())
print(RootModel[DoesntWork2](dw2).model_dump_json())

# Result:
# {"id":"0d984922-443f-4b0d-999e-e91185791ef2"}
# {}
# {}
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.4.2
        pydantic-core version: 2.10.1
          pydantic-core build: profile=release pgo=false
                 install path: .../.venv/lib/python3.12/site-packages/pydantic
               python version: 3.12.0 (main, Oct  2 2023, 22:40:56) [Clang 15.0.0 (clang-1500.0.40.1)]
                     platform: macOS-13.6-arm64-arm-64bit
             related packages: mypy-1.6.1 typing_extensions-4.8.0 fastapi-0.104.0
```

treykeown:
@PrettyWood could this be somehow related to #7821 ?

PrettyWood:
No nothing to do with that
It comes from https://github.com/pydantic/pydantic/blob/main/pydantic/_internal/_fields.py#L251
We should probably work on it

hramezani:
Thanks @treykeown for reporting this and Thanks @PrettyWood  for checking.

I've created a [PR](https://github.com/pydantic/pydantic/pull/7898) to fix the case that `default` exists(DoesntWork1).

Not sure what should we do in the case with `__post_init__`(DoesntWork2)

PrettyWood:
[...]

PR #7898, opened 2023-10-23
Fix pydantic dataclass problem with dataclasses.field default

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/7895

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -248,7 +248,11 @@ def collect_dataclass_fields(
         if is_classvar(ann_type):
             continue
 
-        if not dataclass_field.init and dataclass_field.default_factory == dataclasses.MISSING:
+        if (
+            not dataclass_field.init
+            and dataclass_field.default_factory == dataclasses.MISSING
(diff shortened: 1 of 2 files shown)


hramezani:
Please review

pydantic/_internal/_fields.py
```diff
@@ -248,7 +248,11 @@ def collect_dataclass_fields(
         if is_classvar(ann_type):
             continue
 
-        if not dataclass_field.init and dataclass_field.default_factory == dataclasses.MISSING:
+        if (
+            not dataclass_field.init
+            and dataclass_field.default_factory == dataclasses.MISSING
+            and dataclass_field.default == dataclasses.MISSING
```
PrettyWood, line 254: I would have swapped the order because `default` is used more often that `default_factory`

PrettyWood (approved):
LGTM! Thank you

pydantic/_internal/_fields.py
hramezani: Sure! Done

PrettyWood (approved):

PR #7929, opened 2023-10-25
Fix `display_as_type` for `TypeAliasType` in python 3.12

dmontagu:
@adriangb any issues with this? I think you added those xfails, this change seems reasonable to me since it only affects cases which would currently raise an AttributeError when trying to display a type. And the comment I added should be enough to indicate why it was added in case it becomes unnecessary in the future and anyone otherwise wants to remove it.

Related to https://github.com/pydantic/pydantic/pull/7893; if this gets merged then probably the xfail on the test added there will need to be removed before merging (or that xfail will need to be removed as part of this PR if that PR is merged first).


sydney-runkle (approved):
Nice 🏆

adriangb (approved):

PR #8001, opened 2023-11-02
Make union case tags affect union error messages

dmontagu:
This PR makes it so that you can get cleaner error messages when working with unions by providing `Tag`s for union cases, even if you aren't using a discriminated union. 

If this is accepted, we'll want to improve the docs related to `Tag`.


samuelcolvin (approved):
LGTM, needs documentation.

docs/concepts/unions.md
```diff
@@ -513,4 +515,81 @@ assert m == DiscriminatedModel(
 assert m.model_dump() == data
 ```
 
-You can also simplify error messages with a custom error, like this:
+You can also simplify error messages by labeling each case with a `Tag`. This is especially useful
```
Kludex, line 518:
```suggestion
You can also simplify error messages by labeling each case with a [`Tag`][??]. This is especially useful
```

samuelcolvin (approved):
LGTM

Issue #7971, opened 2023-10-30
`'utf-8' codec can't decode byte` error when serializing `SecretBytes` object


mauvilsa:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Getting `'utf-8' codec can't decode byte` when serializing a `SecretBytes` object. In v1 casting to `str` the result is `'**********'` at least for all secret values I have tried. But in v2, depending on the secret value, it can fail with the decode error.

### Example Code

```Python
>>> from pydantic import SecretBytes
>>> value=b"\x89PNG\r\n\x1a\n"
>>> secret=SecretBytes(value)
>>> str(secret)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/.../lib/python3.10/site-packages/pydantic/types.py", line 1412, in __str__
    return str(self._display())
  File "/.../lib/python3.10/site-packages/pydantic/types.py", line 1523, in _display
    return _secret_display(self.get_secret_value()).encode()
  File "/.../lib/python3.10/site-packages/pydantic/types.py", line 1474, in _secret_display
    value = value.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.4.2
        pydantic-core version: 2.10.1
          pydantic-core build: profile=release pgo=true
                 install path: /.../lib/python3.10/site-packages/pydantic
               python version: 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
                     platform: Linux-6.2.0-35-generic-x86_64-with-glibc2.35
             related packages: typing_extensions-4.8.0
```

PR #8012, opened 2023-11-04
Don't decode bytes (which may not be UTF8) when displaying SecretBytes

alexmojaki:
## Change Summary

Removes unneeded `.decode()`.

I couldn't find any justification for decoding the bytes. The code was introduced in [this commit](https://github.com/pydantic/pydantic/pull/5674/commits/b904b986b6ef7d7fd8b59caf55ade9ae48945d94#diff-f7a381c0918da7ec905bfbe6aff2ccc72c2d0fd6da0dd3011ef1da892e42d57aR497-R498) in https://github.com/pydantic/pydantic/pull/5674 and I don't see any PR comments explaining it. @adriangb any idea?

There are some encodings where a non-empty byte string can decode to any empty string:

```python
>>> ''.encode('utf16')
b'\xff\xfe'
>>> ''.encode('utf16').decode('utf16')
''
>>> ''.encode('utf16').decode('utf16') == ''
True
```

but `.decode()` without arguments should always use UTF8 in Python 3 regardless of locale settings.

I've checked that any non-empty byte string with up to 3 bytes decodes to a non-empty string if it can be decoded at all:

```python
import itertools

count = 0
total = 0
for length in [1, 2, 3]:
    for b in itertools.product(range(256), repeat=length):
        total += 1
        b = bytes(b)
        assert b
        assert len(b) == length > 0
        try:
            s = b.decode()
        except UnicodeDecodeError:
            continue
        assert s
        assert len(s) > 0
        assert b == s.encode()
        count += 1

print(count)  # 2668544
print(total)  # 16843008
```

I also think that any non-empty byte string should be treated as non-empty when displaying even if it can be decoded to an empty unicode string. Although I'm also not sure about revealing whether *any* secret value is empty.

## Related issue number

Closes https://github.com/pydantic/pydantic/issues/7971

## Checklist

* [x] Tests pass on CI

Selected Reviewer: @lig

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -1495,8 +1495,6 @@ class _SecretField(Generic[SecretType]):
 
 
 def _secret_display(value: str | bytes) -> str:
-    if isinstance(value, bytes):
-        value = value.decode()
     return '**********' if value else ''
 
 
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -4174,7 +4174,9 @@ def test_secretbytes():
         empty_password: SecretBytes
 
     # Initialize the model.
-    f = Foobar(password=b'wearebytes', empty_password=b'')
+    # Use bytes that can't be decoded with UTF8 (https://github.com/pydantic/pydantic/issues/7971)
+    password = b'\x89PNG\r\n\x1a\n'
+    f = Foobar(password=password, empty_password=b'')
 
     # Assert correct types.
     assert f.password.__class__.__name__ == 'SecretBytes'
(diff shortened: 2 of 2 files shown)


alexmojaki:
please review

alexmojaki:
> I couldn't find any justification for decoding the bytes. The code was introduced in [this commit](https://github.com/pydantic/pydantic/pull/5674/commits/b904b986b6ef7d7fd8b59caf55ade9ae48945d94#diff-f7a381c0918da7ec905bfbe6aff2ccc72c2d0fd6da0dd3011ef1da892e42d57aR497-R498) in https://github.com/pydantic/pydantic/pull/5674 and I don't see any PR comments explaining it. @adriangb any idea?

Pinging @adriangb because I wrote the above in an edit and I don't know if that triggers a notification.

adriangb (approved):
I don’t see any problem with this. In not sure why I did that in that commit, may have been carryover from moving files or something. I don’t see the decode() call in v1

adriangb:
Mind adding a small release note as a bug fix?

alexmojaki:
> Mind adding a small release note as a bug fix?

Where would I add that? I can't see anything in https://docs.pydantic.dev/latest/contributing/ and adding to `HISTORY.md` doesn't look right.

adriangb:
I guess we just use a tag on the PR now

PR #8023, opened 2023-11-06
Clearer error on invalid Plugin

samuelcolvin:
I got an error from a residual `example_plugin` install, seemed useful to give more details on the plugin that causes the error.


sydney-runkle (approved):
LGTM as well, I pulled in the main branch as well so we could get tests passing 👍

PR #8083, opened 2023-11-10
Correct Documentation Regarding Parsing Datetime Objects From Strings

parkerhancock:
## Change Summary

Changing documentation of `datetime.datetime` parsing to indicate that acceptable string values are only those supported by `datetime.datetime.fromisoformat`.

The current documentation says that a datetime can be in a string formatted like this (i.e. ISO-8601):
`YYYY-MM-DD[T]HH:MM[:SS[.ffffff]][Z or [±]HH[:]MM]`

But in reality, Pydantic relies on the python standard library `datetime.datetime.fromisoformat` in order to convert non-`int`/`float` strings into `datetime.datetime` objects. 

It seems like a trivial difference, but the behavior of `datetime.datetime.fromisoformat` changed in Python v3.11 to be more permissive and to support, among other things, the trailing `Z` to indicate UTC time. Earlier versions of Python **do not support** this format. Until Pydantic deprecates support for Python versions below 3.11, this change is necessary to prevent a "gotcha" that I ran into when trying to convert datetimes from strings and then ran tests on Python v3.10.

As an additional consideration, you may want to include putting into the docs a preferred practice for converting non-standard datetime formats. For example, a highly generic way do it it would be like this:

```python
import datetime
from dateutil.parser import parse as parse_dt

from pydantic import BeforeValidator

ParseDateTime = Annotated(datetime.datetime, BeforeValidator(lambda x: parse_dt(x)))
```

since `dateutil.parser.parse` supports a very wide array of formats, this should work for most formats. Another way to ensure Pydantic handles ISO-8601 datetime strings for all python versions would be to check Python versions, and if it is below 3.11, use `dateutil.parser.isoparse`, which supports ISO-8601 strings. But that would add a Pydantic dependency.

## Related issue number

No related issue

## Checklist

* [X] Tests pass on CI

Selected Reviewer: @adriangb

### DIFF
--- a/docs/api/standard_library_types.md
+++ b/docs/api/standard_library_types.md
@@ -58,7 +58,7 @@ types:
     * `int` or `float`; assumed as Unix time, i.e. seconds (if >= `-2e10` and <= `2e10`) or milliseconds
       (if < `-2e10`or > `2e10`) since 1 January 1970
     * `str`; the following formats are accepted:
-        * `YYYY-MM-DD[T]HH:MM[:SS[.ffffff]][Z or [±]HH[:]MM]`
+        * `str` in a format pareseable by `datetime.datetime.fromisoformat`
         * `int` or `float` as a string (assumed as Unix time)
 
 ```py


parkerhancock:
please review

adriangb:
In v2 we use speedate: https://github.com/pydantic/speedate

Maybe you want to edit the v1 docs? https://github.com/pydantic/pydantic/tree/1.10.X-fixes/docs

Also your example can be `ParseDateTime = Annotated(datetime.datetime, BeforeValidator(parse_dt))`, no need for a lambda.

parkerhancock:
Then something else is going on. Here's what I'm seeing:

image

Patent Client is tested against 3.11, which has the new `isoparse`, and all tests pass. 

The CoLab notebook is running Python 3.10, which doesn't have the updated `datetime.datetime.isoparse` code. And Pydantic v2.4.2 is failing to parse a string like "2018-09-06T00:00:00Z". That is explainable if Pydantic v2.4.2 is using the old `isoparse` internally. But that isn't explained if it is using speedate.

Looking at the speeddate repo, the very first example is a datetime string in *exactly* the same format, with the 'T' separator between date and time, and the trailing Z to indicate UTC time:

```rust
use speedate::{DateTime, Date, Time};

fn main() {
    let dt = DateTime::parse_str("2022-01-01T12:13:14Z").unwrap();
```

Any other ideas?

parkerhancock:
Found the problem. it was with my code. Date handling is hard, and there was a call to datetime.datetime.isoformat tucked in a place I couldn't find...

PR #8209, opened 2023-11-22
Add eval_type_backport to handle union operator and builtin generic subscripting in older Pythons

alexmojaki:
Closes https://github.com/pydantic/pydantic/issues/7873

(previously #7958)

Replaces all calls (outside v1) to `typing._eval_type` with the new `typing_extra.eval_type_backport` which falls back to https://github.com/alexmojaki/eval_type_backport if installed.


alexmojaki:
please review

pydantic/_internal/_typing_extra.py
```diff
@@ -221,12 +221,36 @@ def eval_type_lenient(value: Any, globalns: dict[str, Any] | None, localns: dict
...
+    This would also be the place to add support for `list[int]` instead of `List[int]` etc.
+    """
+    try:
+        return typing._eval_type(  # type: ignore
+            value, globalns, localns
+        )
+    except TypeError as e:
+        if not (isinstance(value, typing.ForwardRef) and is_unsupported_types_for_union_error(e)):
+            raise
+        from eval_type_backport import eval_type_backport
```
dmontagu, line 249: It seems to me that right now if you don't have the `eval_type_backport` package installed, you'll get a `ModuleNotFoundError` if you hit this code branch. I think that's more confusing than it needs to be; I think it would be better if you could add a `try: except ImportError:` to provide a nicer error message explaining to the user what they've done wrong (use the `|` notation for unions in older python version) and how to resolve it (use `typing.Union` or install the `eval_type_backport` package).

pyproject.toml
```diff
@@ -20,57 +20,6 @@ include = [
     '/requirements',
 ]
 
-[project]
```
dmontagu, line 23: Does this whole section need to be moved? It makes it harder to confirm the small diff, and will affect the git blame for this file for future reference.

tests/test_types.py
```diff
@@ -4960,7 +4960,7 @@ class Model(BaseModel):
 
 def test_smart_union_compounded_types_edge_case():
     class Model(BaseModel):
-        x: Union[List[str], List[int]]
+        x: 'List[str] | List[int]'
```
dmontagu, line 4963: You didn't change _all_ unions did you? We should probably make sure we keep at least some instances of `Union[X, Y]` in tests to make sure we can catch any issues that might occur for that but not for `X | Y`. (Even if that shouldn't be an issue today, it might end up being an issue with future code changes.)

dmontagu (changes requested):
Suggestions made above, I think at least the ImportError should be handled, if you disagree or I'm misunderstanding something though let me know.

pyproject.toml
alexmojaki: No, PDM did this automatically and I didn't want to interfere because until now I thought this change was in `pdm.lock` :laughing:

pydantic/_internal/_typing_extra.py
alexmojaki: Right, this was my plan.

> This PR isn't fully ready (it just assumes that the package is installed) but I want to check that this is going in the right direction.

tests/test_types.py
alexmojaki: No, there's still loads of normal Unions left.

alexmojaki:
How should I fix the lint action, i.e. install the package for pyright's sake? Add a new optional dependency group similar to `email`, or add the package to the `linting` dev dependency group?

a-recknagel:
One question, how will this work on the user-site? Will the new lib just be a dependency for pydantic when installing with python >=3.9?

pydantic/_internal/_typing_extra.py
```diff
@@ -221,12 +221,36 @@ def eval_type_lenient(value: Any, globalns: dict[str, Any] | None, localns: dict
...
 
     try:
-        return typing._eval_type(value, globalns, localns)  # type: ignore
+        return eval_type_backport(value, globalns, localns)
     except NameError:
         # the point of this function is to be tolerant to this case
         return value
 
 
+def is_unsupported_types_for_union_error(e: TypeError) -> bool:
```
samuelcolvin, line 230: let's move this below the function that uses it.

pydantic/_internal/_typing_extra.py
```diff
@@ -221,12 +221,36 @@ def eval_type_lenient(value: Any, globalns: dict[str, Any] | None, localns: dict
...
+    return str(e).startswith('unsupported operand type(s) for |: ')
+
+
+def eval_type_backport(
+    value: Any, globalns: dict[str, Any] | None = None, localns: dict[str, Any] | None = None
+) -> Any:
+    """Like `typing._eval_type`, but falls back to the `eval_type_backport` package if it's
+    installed to let older Python versions use newer typing features.
+    Currently this just means that `X | Y` is converted to `Union[X, Y]` if `X | Y` is not supported.
+    This would also be the place to add support for `list[int]` instead of `List[int]` etc.
```
samuelcolvin, line 240: I think it makes sense to support at least the most well known cases of `type`, `list`, `dict`, `set` when we release this.

Presumably there's a set list somewhere and implementing more things is a simple as a lookup dict?

I think supporting pipe unions but nothing else is confusing.

pydantic/_internal/_typing_extra.py
```diff
@@ -403,28 +427,13 @@ def get_type_hints(  # noqa: C901
...
                 )
-            value = typing._eval_type(value, globalns, localns)  # type: ignore
+            value = eval_type_backport(value, globalns, localns)
             if name in defaults and defaults[name] is None:
                 value = typing.Optional[value]
             hints[name] = value
         return hints if include_extras else {k: typing._strip_annotations(t) for k, t in hints.items()}  # type: ignore
 
 
-if sys.version_info < (3, 9):
```
samuelcolvin, line 413: why don't we need this anymore?

pyproject.toml
samuelcolvin: let's revert all these changes.

tests/test_dataclasses.py
```diff
@@ -1346,7 +1346,7 @@ class B:
 
     @pydantic.dataclasses.dataclass
     class Top:
-        sub: Union[A, B] = dataclasses.field(metadata=dict(discriminator='l'))
```
samuelcolvin, line 1349: I don't think we want to change most tests, just add some new tests specifically for the new behaviour.

samuelcolvin (commented):
otherwise I'm :+1: for this change.

We'll need to add something to docs too.

pydantic/_internal/_typing_extra.py
alexmojaki: AFAICT this could always have been replaced with `typing._eval_type`. Now it has to be, with the backport.

tests/test_dataclasses.py
alexmojaki: I've reverted the random test changes, updated existing tests more strategically where it seemed natural (i.e. they already used `from __future__ import annotations` or they were skipped for Python <= 3.9), and added new tests in various files (all called `test_eval_type_backport`) to test the different places where `eval_type_backport` is called.

bswck:
You might want to check out the current discussion on built-in generic's subclasses:
https://github.com/alexmojaki/eval_type_backport/issues/9

pydantic/_internal/_typing_extra.py
```diff
@@ -213,20 +213,52 @@ def get_cls_type_hints_lenient(obj: Any, globalns: dict[str, Any] | None = None)
...
+        return typing._eval_type(  # type: ignore
+            value, globalns, localns
+        )
+    except TypeError as e:
+        if not (isinstance(value, typing.ForwardRef) and is_backport_fixable_error(e)):
+            raise
+        try:
+            from eval_type_backport import eval_type_backport
+        except ImportError:
+            raise RuntimeError(
```
samuelcolvin, line 249:
```suggestion
            raise ImportError(
```
???

samuelcolvin (approved):

dmontagu (approved):

PR #8222, opened 2023-11-24
Add types.OnErrorOmit

adriangb:


pydantic/types.py
```diff
@@ -2800,3 +2804,21 @@ class Model(BaseModel):
...
+    def __get_pydantic_core_schema__(cls, source_type: Any, handler: GetCoreSchemaHandler) -> CoreSchema:
+        # there is no actual default value here but we use with_default_schema since it already has the on_error
+        # behavior implemented and it would be no more efficient to implement it on every other validator
+        # or as a standalone validator
+        return core_schema.with_default_schema(schema=handler(source_type), on_error='omit')
+
+
+OnErrorOmit = Annotated[T, _OnErrorOmit]
+"""
+When used as an item in a list, the key type in a dict, optional values of a TypedDict, etc.
```
samuelcolvin, line 2820: key or value? we should probably add a test.

pydantic/types.py
```diff
@@ -2800,3 +2804,21 @@ class Model(BaseModel):
             _AllowAnyJson,
         ],
     )
+
+
+class _OnErrorOmit:
```
samuelcolvin, line 2809: I wonder if we should we make this a dataclass?

```py
@dataclass
class OnError(Generic[T]):
    on_error: Literal['omit', 'raise', 'default']
    default: Any | Callable[[], T] = None
```
?

tests/test_types.py
```diff
@@ -6129,3 +6133,38 @@ class MyModel(BaseModel):
...
+    actual = Model(
+        a_list=[1, 2, 'a', 3],
+        a_dict={1: 1, 2: 2, 'a': 'a', 'b': 0, 3: 'c', 4: 4},
+        a_typed_dict=MyTypedDict(a=1, b='xyz'),  # type: ignore
+    )
+
+    expected = Model(a_list=[1, 2, 3], a_dict={1: 1, 2: 2, 4: 4}, a_typed_dict=MyTypedDict(a=1))
+
+    assert actual == expected
+
```
samuelcolvin, line 6159: can we add a quick test for list, dict key and dict value?

pydantic/types.py
```diff
@@ -2800,3 +2804,21 @@ class Model(BaseModel):
...
+        # behavior implemented and it would be no more efficient to implement it on every other validator
+        # or as a standalone validator
+        return core_schema.with_default_schema(schema=handler(source_type), on_error='omit')
+
+
+OnErrorOmit = Annotated[T, _OnErrorOmit]
+"""
+When used as an item in a list, the key type in a dict, optional values of a TypedDict, etc.
+this annotation omits the item from the iteration if there is any error validating it.
+That is, instead of a ValidationError being propagated up and the entire iterable being discarded
```
samuelcolvin, line 2822:
```suggestion
That is, instead of a [`ValidationError`][???] being propagated up and the entire iterable being discarded
```

samuelcolvin (commented):
otherwise LGTM.

pydantic/types.py
adriangb: I did

pydantic/types.py
adriangb: It sounds like you want to fully expose the `with_default_value` options, but IMO even if we do that later on this use case is simple and common enough to warrant it's own annotation with no options so you don't need to do `Annotated[..., OnError(on_error='omit')]`.

tests/test_types.py
adriangb: That's already in this test, I'm not sure what you're asking for

samuelcolvin (approved):

pydantic/types.py
samuelcolvin: PR welcome for anyone to add `OnError` as described by me above.

Issue #7327, opened 2023-09-04
field defined with `Field()` makes pydantic to have a "RecursionError: maximum recursion depth exceeded"


luminoso:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

pydantic 2.x leads to a recursion error when running the example code.

```
(....)

  File "/home/luminoso/PycharmProjects/pydantic2datamodel-bug-report/venv/lib/python3.11/site-packages/pydantic/_internal/_repr.py", line 55, in __repr_str__
    return join_str.join(repr(v) if a is None else f'{a}={v!r}' for a, v in self.__repr_args__())
                                                                            ^^^^^^^^^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded
```

to test, try to parse an object: 
```python
Model.parse_obj({"Settings": {"name": "foobar"}})
```

(Despite pydantic 2 deprecated `parse_obj()` the problem also happens with pydantic v2 `model_validate()`)

A workaround is not to use the `Field()`, as so:

```python
class Model(BaseModel):
    Settings: Settings
```

Where the problem doesn't happen.

And the solution is to downgrade pydantic `1.10.12`, where the problem does not occur with both with or without the `Field()` definition.

### Example Code

```Python
from __future__ import annotations

from pydantic.fields import Field
from pydantic.main import BaseModel


class Settings(BaseModel):
    name: str


class Model(BaseModel):
    Settings: Settings = Field(..., title="a nice title")
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.3.0
        pydantic-core version: 2.6.3
          pydantic-core build: profile=release pgo=true
                 install path: /var/home/luminoso/PycharmProjects/pydantic2datamodel-bug-report/venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.4 (main, Jun  7 2023, 00:00:00) [GCC 13.1.1 20230511 (Red Hat 13.1.1-2)]
                     platform: Linux-6.1.50-200.fc38.x86_64-x86_64-with-glibc2.37
     optional deps. installed: ['email-validator', 'typing-extensions']
```

Selected Assignee: @dmontagu

luminoso:
(https://github.com/pydantic/pydantic/issues/7309 may be related)

hramezani:
Thanks @luminoso for reporting this 🙏 

Yes, It is a duplicate of https://github.com/pydantic/pydantic/issues/7309

[...]

PR #8243, opened 2023-11-28
Explicitly raise an error if field names clashes with types

Viicos:
## Change Summary
Linked: https://github.com/pydantic/pydantic/issues/8240.

I'm raising inside `from_annotated_attribute` so I don't have to duplicate the logic for dataclass, BaseModel etc. But unfortunately I can't provide info to the user regarding the field causing the issue :/

## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -1023,7 +1023,7 @@ except PydanticUserError as exc_info:
 Because type annotations are evaluated *after* assignments, you might get unexpected results when using a type annotation name
 that clashes with one of your fields. We raise an error in the following case:
 
-```py
+```py test="skip"
 from datetime import date
 
 from pydantic import BaseModel, Field
@@ -1035,9 +1035,8 @@ class Model(BaseModel):
 
 As a workaround, you can either use an alias or change your import:
 
-```py
+```py lint="skip"
 import datetime
-
 # Or `from datetime import date as _date`
 
 from pydantic import BaseModel, Field
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -1018,5 +1018,31 @@ except PydanticUserError as exc_info:
     assert exc_info.code == 'root-model-extra'
 ```
 
+## Cannot evaluate type annotation (#unevaluable-type-annotation)
+
+Because type annotations are evaluated *after* assignments, you might get unexpected results when using a type annotation name
+that clashes with one of your fields. We raise an error in the following case:
+
+```py
+from datetime import date
+
+from pydantic import BaseModel
+
+class Model(BaseModel):
+    date: date = Field(description="A date")
+```
+
+As a workaround, you can either use an alias or change your import:
+
+```py
+import datetime
+# Or `from datetime import date as _date`
+
+from pydantic import BaseModel
+
+class Model(BaseModel):
+    date: datetime.date = Field(description="A date")
+
+```
 
 {% endraw %}
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -1026,7 +1026,7 @@ that clashes with one of your fields. We raise an error in the following case:
 ```py
 from datetime import date
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class Model(BaseModel):
@@ -1045,7 +1045,6 @@ from pydantic import BaseModel, Field
 
 class Model(BaseModel):
     date: datetime.date = Field(description='A date')
-
 ```
 
 {% endraw %}
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -1037,9 +1037,11 @@ As a workaround, you can either use an alias or change your import:
 
 ```py
 import datetime
+
 # Or `from datetime import date as _date`
 
 from pydantic import BaseModel
+
 
 class Model(BaseModel):
     date: datetime.date = Field(description="A date")
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -1028,6 +1028,7 @@ from datetime import date
 
 from pydantic import BaseModel
 
+
 class Model(BaseModel):
     date: date = Field(description="A date")
 ```
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -1018,7 +1018,7 @@ except PydanticUserError as exc_info:
     assert exc_info.code == 'root-model-extra'
 ```
 
-## Cannot evaluate type annotation (#unevaluable-type-annotation)
+## Cannot evaluate type annotation {#unevaluable-type-annotation}
 
 Because type annotations are evaluated *after* assignments, you might get unexpected results when using a type annotation name
 that clashes with one of your fields. We raise an error in the following case:
--- a/docs/errors/usage_errors.md
+++ b/docs/errors/usage_errors.md
@@ -1030,7 +1030,7 @@ from pydantic import BaseModel
 
 
 class Model(BaseModel):
-    date: date = Field(description="A date")
+    date: date = Field(description='A date')
 ```
 
 As a workaround, you can either use an alias or change your import:
@@ -1040,11 +1040,11 @@ import datetime
 
 # Or `from datetime import date as _date`
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class Model(BaseModel):
-    date: datetime.date = Field(description="A date")
+    date: datetime.date = Field(description='A date')
 
 ```
 
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -321,12 +321,11 @@ class FieldInfo(_repr.Representation):
         Returns:
             A field object with the passed values.
         """
-
         if annotation is default:
-            raise RuntimeError(
+            raise PydanticUserError(
                 'Error when building FieldInfo from annotated attribute. '
-                "Make sure you don't have any field name clashing with a type annotation "
-                '(e.g. `date: date = Field(None)`).'
+                "Make sure you don't have any field name clashing with a type annotation ",
+                code='unevaluable-type-annotation',
             )
 
         final = False
--- a/pydantic/fields.py
+++ b/pydantic/fields.py
@@ -321,6 +321,14 @@ class FieldInfo(_repr.Representation):
         Returns:
             A field object with the passed values.
         """
+
+        if annotation is default:
+            raise RuntimeError(
+                'Error when building FieldInfo from annotated attribute. '
+                "Make sure you don't have any field name clashing with a type annotation "
+                '(e.g. `date: date = Field(None)`).'
+            )
+
         final = False
         if _typing_extra.is_finalvar(annotation):
             final = True
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -2434,10 +2434,10 @@ def test_invalid_forward_ref_model():
     """
     # The problem:
     if sys.version_info >= (3, 11):
-        error = RecursionError
-    else:
-        # See PR #8243, this was a TypeError raised by Python, but is now catched on the Pydantic side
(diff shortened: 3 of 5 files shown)


sydney-runkle:
Do we want to consider a case like second one here as well?

```py
from datetime import date
from typing import Union

class Foo(BaseModel):
    date: date = Field(default_factory=date.today)

f = Foo()
#> RecursionError: maximum recursion depth exceeded while getting the repr of an object

class FooWithUnion(BaseModel):
    date: Union[date, None] = Field(default=None)

f = FooWithUnion()
#> RecursionError: maximum recursion depth exceeded while getting the repr of an object

```

Viicos:
Thanks for the feedback, I'll try to come up with a solution covering all cases. Hopefully it won't involves hacky stuff as this might get tricky

Viicos:
> Do we want to consider a case like second one here as well?

@sydney-runkle where you able to run this code snippet? `Union` is a special form and the arguments are checked to be sure they are a type. I got the following on my end: `TypeError: Union[arg, ...]: each arg must be a type. Got FieldInfo ...`

Note that setting a default value of `None` [will pass](https://github.com/python/cpython/blob/c1c6bedfd3ee66ad208e0cd9cdd732374c95b83d/Lib/typing.py#L136-L137):

```python
from datetime import date
from typing import Union

class FooWithUnion(BaseModel):
    date: Union[date, None] =None

FooWithUnion.model_json_schema()
#> {'properties': {'date': {'default': None, 'title': 'Date', 'type': 'null'}}, 'title': 'B', 'type': 'object'}
# ^ type: null!
``` 
And in this case the issue is silently ignored, so having an explicit error will be beneficial in this case as well

sydney-runkle:
@Viicos,

You're right, I get the following:

```py
from datetime import date
from typing import Union

class FooWithUnion(BaseModel):
    date: Union[date, None] = Field(default=None)

f = FooWithUnion()
#> Union[arg, ...]: each arg must be a type. Got FieldInfo(annotation=NoneType, required=False).
```

Agreed, a more explicit error would be helpful 😄

sydney-runkle:
Makes sense. Curious to hear what @alexmojaki thinks as well.

The case that I mentioned isn't as urgent. As long as we cover the original case and the case that @samuelcolvin mentioned above, that's a pretty significant improvement over what we have now.

alexmojaki:
I'm flattered :smile: 

Here's an idea of how it might be possible to actually fix the problem: https://github.com/pydantic/pydantic/compare/main...alexmojaki:pydantic:field-type-clash

Viicos:
Again, @alexmojaki really clever solution. I've played a bit and this looks promising. I'll continue experimenting and will try to come up with an implementation. Thanks again!

Viicos:
So I've played a bit with it, here is an implementation following @alexmojaki's idea: https://github.com/pydantic/pydantic/compare/main...Viicos:pydantic:allow-name-clashes

Main differences are:
- We need a custom `__setitem__` in case class attributes are accessed in the body:
```python
class Model(BaseModel):
    def func(self):
        pass

    func2 = func  # Python will try to get "func" from our custom namespace, where "func" is still prefixed/mangled
```

- Do not "override" the custom namespace by a plain dict in dynamic model creations.

All tests passes with this implementation, but testing that it can actually use name that clashes is harder as it does not work inside functions (see [here](https://github.com/pydantic/pydantic/issues/7327#issuecomment-1830399262)).

I'll let maintainers decide if they want to support this and continue with this solution, otherwise we can continue on this PR and try to see how we could raise better errors.

alexmojaki:
Does this work with postponed/string annotations? Maybe to support using non-field variables in the class body (which I didn't) you could keep `localns` in `get_cls_type_hints_lenient` but exclude `model_fields`.

> We need a custom `__setitem__` in case class attributes are accessed in the body:

Do you mean `__getitem__`? If so, does this mean the `__getitem__` is called when evaluating normal expressions in the class body, but not when evaluating annotations? That's weird.

> All tests passes with this implementation, but testing that it can actually use name that clashes is harder as it does not work inside functions (see https://github.com/pydantic/pydantic/issues/7327#issuecomment-1830399262).

Is it fair to say that in every case where it doesn't work inside functions, it has never worked, and it also doesn't work with plain classes or standard dataclasses?

You *could maybe* make it work inside functions if you're willing to take the black magic further:

```python
def foo():
  class A:
    pass

  class Meta(type):
    @classmethod
    def __prepare__(metacls, name, bases):
      import inspect
      frame = inspect.currentframe().f_back
      return dict(frame.f_locals)

  class B(metaclass=Meta):
    A: A

  return B


print(foo().__annotations__)
```

Or just use one of those test fixtures that makes a module or something.

alexmojaki:
```python
import inspect
from datetime import date


class D(dict):
    def __getitem__(self, key):
        if not key[:2] == '__' == key[-2:]:
            frame = inspect.currentframe().f_back
            assert frame.f_locals is self
            for namespace in [frame.f_back.f_locals, frame.f_globals, frame.f_builtins]:
                try:
                    value = namespace[key]
                except KeyError:
                    pass
                else:
                    if key in self:
                        print(f'Found {key=} in outer namespace, using that instead of value in class')
                    return value
        return super().__getitem__(key)


class M(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        return D()


def foo():
    class A:
        pass

    class C(metaclass=M):
        class B:
            pass

        A: A = 1
        b: B = 2
        date: date = 3
        int: int = 4
        x: int = 5
        y: int = x

    # print(C.A, C.b, C.date, C.int, C.x, C.y)
    # print(C.__annotations__)
    assert C.A == 1
    assert C.b == 2
    assert C.date == 3
    assert C.int == 4
    assert C.x == 5
    assert C.y == 5
    assert C.__annotations__ == {
        'A': A,
        'b': C.B,
        'date': date,
        'int': int,
        'x': int,
        'y': int,
    }


foo()
```

alexmojaki:
To explain the above code (which I've just updated): we can use the custom namespace to override Python's usual scoping rules and check for a local class variable last (i.e. after enclosing scopes, globals, and builtins) rather than first. Of course that could be weird and surprising in some cases, but we can easily emit warnings/errors when relevant so that the user can see what's happening, and this can be as configurable as we want so that upgrading is still safe. I can explain more but right now I need to go to sleep :laughing:

Viicos:
This looks interesting but I don't know if this level of complexity will be accepted? :/

pydantic/fields.py
```diff
@@ -321,6 +321,14 @@ class MyModel(pydantic.BaseModel):
                 spam: Annotated[int, pydantic.Field(gt=4)] = 4  # <-- or this
             ```
         """
+
+        if annotation is default:
+            raise RuntimeError(
```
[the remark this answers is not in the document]
sydney-runkle: @Viicos,

Circling back to this solution - I do think we should go ahead and implement something along the lines of what you have here. Could you use a PydanticUserError here instead, as Samuel mentioned?

Happy to review once you make that change 👍

sydney-runkle (changes requested):

pydantic/fields.py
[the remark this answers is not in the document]
Viicos: Yes I think for now this would be the best we can do. It doesn't cover all cases (e.g. [this](https://github.com/pydantic/pydantic/pull/8243#issuecomment-1830485561) and [this](https://github.com/pydantic/pydantic/pull/8243#issuecomment-1830510003)) but this is still better than nothing. Will apply changes soon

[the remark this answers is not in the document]
sydney-runkle: Thanks! Yep, let's stick with this for now, then can consider a more complex solution in the future.

sydney-runkle:
@Viicos,

I know you're busy (helping us with other things, partially 🚀 ), but if you have any time to make a minor change here, we'd love to get this change in with the `2.6.0` release early next week 💯.

Viicos:
> I know you're busy (helping us with other things, partially 🚀 ), but if you have any time to make a minor change here, we'd love to get this change in with the `2.6.0` release early next week 💯.

Will take a look this evening/weekend!

tests/test_edge_cases.py
```diff
@@ -1375,19 +1375,17 @@ class FooBar:
...
     class Model(BaseModel):
-        a: int = int
-        b: Type[int]
-        c: Type[int] = int
-        d: FooBar = FooBar
-        e: Type[FooBar]
-        f: Type[FooBar] = FooBar
-        g: Sequence[Type[FooBar]] = [FooBar]
-        h: Union[Type[FooBar], Sequence[Type[FooBar]]] = FooBar
-        i: Union[Type[FooBar], Sequence[Type[FooBar]]] = [FooBar]
```
Viicos, line 1386: As said [here](https://github.com/pydantic/pydantic/pull/8243#issuecomment-1830437786): I just removed the now invalid cases

docs/errors/usage_errors.md
```diff
@@ -1018,5 +1018,31 @@ except PydanticUserError as exc_info:
     assert exc_info.code == 'root-model-extra'
 ```
 
+## Cannot evaluate type annotation (#unevaluable-type-annotation)
+
+Because type annotations are evaluated *after* assignments, you might get unexpected results when using a type annotation name
+that clashes with one of your fields. We raise an error in the following case:
```
sydney-runkle, line 1024: Great, thanks for adding these docs 👍

docs/errors/usage_errors.md
```diff
@@ -1018,5 +1018,31 @@ except PydanticUserError as exc_info:
     assert exc_info.code == 'root-model-extra'
 ```
 
+## Cannot evaluate type annotation (#unevaluable-type-annotation)
```
sydney-runkle, line 1021:
```suggestion
## Cannot evaluate type annotation {#unevaluable-type-annotation}
```

Viicos:
Regarding the other mentioned use cases, it could also be implemented (more easily) in a linter plugin, e.g. with flake8 or ruff. There are some interesting rules that could also be applied, e.g. enforcing `field: Annotated[<type>, Field(...)]` instead of `field: <type> = Field(...)`, or any use case where people with large code bases would like to enforce consistency.

docs/errors/usage_errors.md
```diff
@@ -1018,5 +1018,31 @@ except PydanticUserError as exc_info:
...
+Because type annotations are evaluated *after* assignments, you might get unexpected results when using a type annotation name
+that clashes with one of your fields. We raise an error in the following case:
+
+```py
+from datetime import date
+
+from pydantic import BaseModel
+
+class Model(BaseModel):
+    date: date = Field(description="A date")
```
sydney-runkle, line 1032:
```suggestion
from datetime import date

from pydantic import BaseModel


class Model(BaseModel):
    date: date = Field(description="A date")
```

docs/errors/usage_errors.md
```diff
@@ -1018,5 +1018,31 @@ except PydanticUserError as exc_info:
...
+
+```py
+import datetime
+# Or `from datetime import date as _date`
+
+from pydantic import BaseModel
+
+class Model(BaseModel):
+    date: datetime.date = Field(description="A date")
+
```
sydney-runkle, line 1045:
```suggestion
import datetime

# Or `from datetime import date as _date`

from pydantic import BaseModel


class Model(BaseModel):
    date: datetime.date = Field(description="A date")

```

sydney-runkle:
@Viicos,

Looks like a test is still failing - could you please take a look at that before we merge?

sydney-runkle:
@Viicos,

Hmm, still some linting and testing issues

Viicos:
@samuelcolvin @sydney-runkle 

I created [`flake8-pydantic`](https://github.com/Viicos/flake8-pydantic) yesterday, with a bunch of handy rules. [One of them](https://github.com/Viicos/flake8-pydantic?tab=readme-ov-file#pyd005---field-name-overrides-annotation) hopefully covers most of the use cases we encountered.

Let me know if you have other rules in mind that could be implemented!

sydney-runkle:
@Viicos,

Nice! That's great!

PR #8294, opened 2023-12-03
Fix usage of `@deprecated`

Viicos:
## Change Summary

The new version now emits runtime warnings, so no need to duplicate logic.

## Checklist

* [ ] Tests pass on CI

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -242,10 +242,11 @@ class ModelMetaclass(ABCMeta):
         return field_names, class_vars, private_attributes
 
     @property
-    @deprecated(
-        'The `__fields__` attribute is deprecated, use `model_fields` instead.', category=PydanticDeprecatedSince20
-    )
+    @deprecated('The `__fields__` attribute is deprecated, use `model_fields` instead.', category=None)
     def __fields__(self) -> dict[str, FieldInfo]:
+        warnings.warn(
+            'The `__fields__` attribute is deprecated, use `model_fields` instead.', PydanticDeprecatedSince20
+        )
         return self.model_fields  # type: ignore
 
 
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -246,7 +246,6 @@ class ModelMetaclass(ABCMeta):
         'The `__fields__` attribute is deprecated, use `model_fields` instead.', category=PydanticDeprecatedSince20
     )
     def __fields__(self) -> dict[str, FieldInfo]:
-        warnings.warn('The `__fields__` attribute is deprecated, use `model_fields` instead.', DeprecationWarning)
         return self.model_fields  # type: ignore
 
 
--- a/pydantic/deprecated/decorator.py
+++ b/pydantic/deprecated/decorator.py
@@ -1,4 +1,3 @@
-import warnings
 from functools import wraps
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Mapping, Optional, Tuple, Type, TypeVar, Union, overload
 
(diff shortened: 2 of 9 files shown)


sydney-runkle (changes requested):
LGTM overall, but I think there are more cases where we need to change this, ex:

https://github.com/pydantic/pydantic/blob/8bbb3c1fa0f360cf153ec82273b422f7ea64a789/pydantic/main.py#L985-L991

Will wait on a stable version of `typing_extensions`, but excited to adopt this once it comes out.

Thanks for your contribution 🚀 ⭐

Viicos:
> Will wait on a stable version of `typing_extensions`, but excited to adopt this once it comes out.

`typing_extensions==4.6.1` already support runtime warnings, but yes might be better to wait until `4.9.0` is released

sydney-runkle:
Looks good overall. Will approve once we rerun tests with `4.9.0`. Thanks :)

pydantic/deprecated/decorator.py
```diff
@@ -1,8 +1,7 @@
-import warnings
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Mapping, Optional, Tuple, Type, TypeVar, Union, overload
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Mapping, Optional, Tuple, Type, TypeVar, Union
 
-from typing_extensions import deprecated
+from typing_extensions import deprecated, overload
```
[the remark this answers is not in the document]
Viicos: Outdated, see new comment

tests/test_deprecated.py
```diff
@@ -718,49 +718,51 @@ class Model(BaseModel):
...
-        PydanticDeprecatedSince20, match='custom_pydantic_encoder is deprecated, use BaseModel.model_dump instead.'
+        PydanticDeprecatedSince20, match='`custom_pydantic_encoder` is deprecated, use `BaseModel.model_dump` instead.'
     ):
         custom_pydantic_encoder({int: lambda x: str(x)}, Model(x=1))
 
     assert hasattr(timedelta_isoformat, '__deprecated__')
     with pytest.warns(PydanticDeprecatedSince20, match='timedelta_isoformat is deprecated.'):
         timedelta_isoformat(timedelta(seconds=1))
 
-    assert all(hasattr(func, '__deprecated__') for func in get_overloads(validate_arguments))
```
Viicos, line 763: For this assertion to be true, each decorator would have to be marked as deprecated, which is unnecessary. See https://discuss.python.org/t/how-should-warnings-deprecated-be-handled-with-typing-overload/40641 for more details.

Viicos:
`typing-extensions` was updated to latest in the lock file. I will bump the minimum required version in the deprecated fields PR.

pydantic/_internal/_model_construction.py
```diff
@@ -246,7 +246,6 @@ def _collect_bases_data(bases: tuple[type[Any], ...]) -> tuple[set[str], set[str
         'The `__fields__` attribute is deprecated, use `model_fields` instead.', category=PydanticDeprecatedSince20
     )
     def __fields__(self) -> dict[str, FieldInfo]:
-        warnings.warn('The `__fields__` attribute is deprecated, use `model_fields` instead.', DeprecationWarning)
```
dmontagu, line 249: The reason we have this warning is actually because this is how PyCharm decides how to show the strikethrough for deprecated attributes:
image
If this line is removed, PyCharm stops showing the strikethrough on usage, even after updating `typing_extensions`.

If there are downsides to having both I'm happy to try to resolve it but I think showing the deprecated stuff as deprecated in PyCharm is a significant benefit, considering we are trying to advise people against using deprecated stuff anyway (so I don't mind terribly if it emits two warnings, unless there is some significant downside to that).

Viicos: Sure thing I can add back the warning for properties, and as a workaround set `category=None` in the decorator to avoid having the warning emitted twice.

However, this is most likely an issue with PyCharm. Do you happen to know if there's an already existing issue tracking this?

The recently defined typing spec [states properties should also be marked as deprecated](https://typing.readthedocs.io/en/latest/spec/directives.html#deprecated) by type checkers (see provided example).

dmontagu: Yeah it's definitely a PyCharm issue — I found this in their issue tracker https://youtrack.jetbrains.com/issue/PY-61651/Deprecation-highlighting-with-PEP-702-deprecated-decorator and added a comment on it about our usage within Pydantic.

Viicos: @dmontagu, is this also the case for standard methods/functions, or is it only the not working for properties?

dmontagu: I think for everything not just properties

Viicos: Ok, then I will add back the actual `warn` calls, and use `category=None` on the decorator. I can also create an issue to remind us to remove them again once PyCharm will have implemented support for the decorator.

Viicos:
Please review

Pypy seems to be flaky, I don't know if this is related to the proposed changes?

pydantic/_internal/_model_construction.py
sydney-runkle: @Viicos,

That would be fantastic if you could create an issue as a reminder for us to remove the `warn` calls

sydney-runkle (approved):
Changes look great, thank you!

pydantic/_internal/_model_construction.py
Viicos: Sure thing: https://github.com/pydantic/pydantic/issues/8524

PR #8305, opened 2023-12-05
Fix type annotation of `ModelMetaclass.__prepare__`

slanzmich:
## Change Summary

This fixes a type annotation that was fixed in the typeshed in https://github.com/python/typeshed/pull/11093. `type.__prepare__` (and thus `ModelMetaclass.__prepare__`) must return a mutable mapping. This is because during class creation, that namespace is populated when the class body is executed.

## Related issue number

As this change looks rather trivial, I have not created an issue.

## Checklist

* [ ] Tests pass on CI

Selected Reviewer: @hramezani

### DIFF
--- a/pydantic/_internal/_model_construction.py
+++ b/pydantic/_internal/_model_construction.py
@@ -8,7 +8,7 @@ import weakref
 from abc import ABCMeta
 from functools import partial
 from types import FunctionType
-from typing import Any, Callable, Generic, Mapping
+from typing import Any, Callable, Generic
 
 import typing_extensions
 from pydantic_core import PydanticUndefined, SchemaSerializer
@@ -216,7 +216,7 @@ class ModelMetaclass(ABCMeta):
             raise AttributeError(item)
 
     @classmethod
-    def __prepare__(cls, *args: Any, **kwargs: Any) -> Mapping[str, object]:
+    def __prepare__(cls, *args: Any, **kwargs: Any) -> dict[str, object]:
         return _ModelNamespaceDict()
 
     def __instancecheck__(self, instance: Any) -> bool:


slanzmich:
I'm not too familiar with how quickly changes from the typeshed end up in the static type checkers, though. Other than that this is ready, please review.

pydantic/_internal/_model_construction.py
```diff
@@ -216,7 +216,7 @@ def __getattr__(self, item: str) -> Any:
             raise AttributeError(item)
 
     @classmethod
-    def __prepare__(cls, *args: Any, **kwargs: Any) -> Mapping[str, object]:
+    def __prepare__(cls, *args: Any, **kwargs: Any) -> MutableMapping[str, object]:
```
Viicos, line 219:
```suggestion
    def __prepare__(cls, *args: Any, **kwargs: Any) -> _ModelNamespaceDict:
```
And make `_ModelNamespaceDict` parametrized? Could be `_ModelNamespaceDict(dict[str, object])` but that doesn't play nice with Python  < 3.10, so maybe something like `_ModelNamespaceDict(dict, typing.MutableMapping[str, object])`?

Viicos:
> I'm not too familiar with how quickly changes from the typeshed end up in the static type checkers, though.

pyright is checked on CI, and we should expect a new release very soon (probably before next week).

pydantic/_internal/_model_construction.py
slanzmich:
```python
from typing import MutableMapping

class D(dict, MutableMapping[str, float]):
    pass

d = D()
d["k"] = 100
d[2] = 20
```
does not seem to work, `mypy` finds no issue with that, while it complains as expected for
```python
class D(dict[str, float]):
    pass

d = D()
d["k"] = 100
d[2] = 20
```
`dict[str, object]` works starting with Python 3.9, but not 3.8, which is still supported by `pydantic`.

slanzmich: Sorry, I'm not too familiar with the code base, so I only noticed this now: `_model_construction.py` already uses [PEP 585](https://peps.python.org/pep-0585/) type annotations. For consistency, I assume we can annotate the return type as `dict[str, object]` without using `typing.Dict`.

sydney-runkle (approved):
@slanzmich,

Looks great, thanks for your contribution! :)

Viicos:
Pyright got updated to `1.1.339` yesterday but typeshed doesn't seem to have been synced in this release. I think it's still fine to merge, pyright can be updated later once a version ships with the typeshed changes

PR #8376, opened 2023-12-15
Add a test demonstrating why we don't reset the parent namespace in `create_model`

dmontagu:
Adds a test showing why the line of code removed in https://github.com/pydantic/pydantic/pull/8011 was there


pydantic/main.py
```diff
@@ -133,6 +133,7 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
         # pydantic._internal._generate_schema.GenerateSchema.model_schema to work for a plain BaseModel annotation
         model_fields = {}
         __pydantic_decorators__ = _decorators.DecoratorInfos()
+        __pydantic_parent_namespace__ = None
```
dmontagu, line 136: I noticed this attribute could fail to be defined in some places where it was accessed. Explicitly setting it to `None` here fixed that.

sydney-runkle (approved):
Looks great, thanks!

PR #8437, opened 2023-12-24
Introducing classproperty decorator for model_computed_fields

### LINKED ISSUE #7986: should 'model_computed_fields' be a class attribute not an instance property?
### Initial Checks

### Description

Hopefully title is quite self explanatory, but here is a bit of extra detail....

if I have a Model subclass "Schema", then I expected to be able to get to computed_fields via "model_computed_fields" attribute e.g. Schema.model_computed_fields, but this returns a "property" instance

looking under the hood, model_computed_fields returns data from `self.__pydantic_decorators__.computed_fields` - but `__pydantic_decorators__` is a class attribute and could presumably be accessed from the class without having a model instance

So in summary, this works:
`Schema.__pydantic_decorators__.computed_fields`

I expected this to work, but it doesn't:
`Schema.model_computed_fields`

Thanks!

### Example Code

_No response_

```Text
pydantic version: 2.2.0
        pydantic-core version: 2.6.0
               python version: 3.10.2 (main, Jan 28 2022, 16:19:08) [GCC 10.3.0]
                     platform: Linux-5.15.0-87-generic-x86_64-with-glibc2.31
```

Jocelyn-Gas:
## Change Summary

A `classproperty` decorator has been added in order to access `model_computed_fields` as a class attribute and an instance property.

## Related issue number
fix #7986

## Checklist

* [x] Tests pass on CI

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -122,6 +122,9 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
         __pydantic_serializer__: ClassVar[SchemaSerializer]
         __pydantic_validator__: ClassVar[SchemaValidator]
 
+        model_computed_fields: ClassVar[dict[str, ComputedFieldInfo]]
+        """A dictionary of computed field names and their corresponding `ComputedFieldInfo` objects."""
+
         # Instance attributes
         # Note: we use the non-existent kwarg `init=False` in pydantic.fields.Field below so that @dataclass_transform
         # doesn't think these are valid as keyword arguments to the class initializer.
@@ -129,12 +132,12 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
         __pydantic_fields_set__: set[str] = _Field(init=False)  # type: ignore
         __pydantic_private__: dict[str, Any] | None = _Field(init=False)  # type: ignore
 
-        model_computed_fields: ClassVar[dict[str, ComputedFieldInfo]]
     else:
         # `model_fields` and `__pydantic_decorators__` must be set for
         # pydantic._internal._generate_schema.GenerateSchema.model_schema to work for a plain BaseModel annotation
         model_fields = {}
         model_computed_fields = {}
+
         __pydantic_decorators__ = _decorators.DecoratorInfos()
         __pydantic_parent_namespace__ = None
         # Prevent `BaseModel` from being instantiated directly:
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -58,14 +58,6 @@ __all__ = 'BaseModel', 'create_model'
 _object_setattr = _model_construction.object_setattr
 
 
-class classproperty:
-    def __init__(self, function) -> None:
-        self.fget = function
-
-    def __get__(self, instance, owner) -> Any:
-        return self.fget(owner)
-
-
 class BaseModel(metaclass=_model_construction.ModelMetaclass):
     """Usage docs: https://docs.pydantic.dev/2.6/concepts/models/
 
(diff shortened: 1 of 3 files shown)


Jocelyn-Gas:
Please review the above PR to check it suits the need you described @hottwaj @sydney-runkle

Jocelyn-Gas:
It's my first contribution attempt to this repo. I'm a bit surprised but the `release notes` check that is not passing. If anyone can help me please, I'd be more than happy to fix it if pointed in the right direction 😁

tests/test_computed_fields.py
```diff
@@ -62,6 +62,13 @@ def double_width(self) -> int:
     assert rect.model_dump() == {'width': 10, 'length': 5, 'area': 50, 'area2': 50}
     assert rect.model_dump_json() == '{"width":10,"length":5,"area":50,"area2":50}'
 
+    assert set(Rectangle.model_fields) == {'width', 'length'}
+    assert set(Rectangle.model_computed_fields) == {'area', 'area2'}
+
+    assert Rectangle.model_computed_fields['area'].description == 'An awesome area'
+    assert Rectangle.model_computed_fields['area2'].title == 'Pikarea'
+    assert Rectangle.model_computed_fields['area2'].description == 'Another area'
```
Viicos, line 70: I would also check what happens when trying to set the `model_computed_fields` attribute to some value (for reference, having the chain of `@classmethod` + `@property` was deprecated in 3.11 partly because of unsupported behavior when setting values on those properties).

pydantic/main.py
```diff
@@ -58,6 +58,14 @@
 _object_setattr = _model_construction.object_setattr
 
 
+class classproperty:
+    def __init__(self, function) -> None:
+        self.fget = function
+
+    def __get__(self, instance, owner) -> Any:
+        return self.fget(owner)
```
Viicos, line 66: This could benefit from proper type checking on the arguments. What happens currently when `model_computed_fields` is accessed with `mypy` and `pyright`?

Viicos:
> I'm a bit surprised but the `release notes` check that is not passing. If anyone can help me please, I'd be more than happy to fix it if pointed in the right direction 😁

I think this is just a matter of maintainers adding a label on this PR

hottwaj:
Hey while I'm not the right person to comment on what is the best choice
for pydantic, I just wanted to offer my opinion that I'm not sure about
adding a new decorator class (classproperty) just for the purpose of this
one-off usage for model_computed_fields.  Doesn't this overcomplicate
things?

I think it would be better to look at what is done when
__pydantic_decorators__ is set and set model_computed_fields at the same
time?

Thanks!

Viicos:
> I'm not sure about adding a new decorator class (classproperty) just for the purpose of this one-off usage for model_computed_fields.

In kind of agree, taking a look at it it seems this can be set during class construction (this property doesn't make use of any "dynamic" attribute that could change).

Jocelyn-Gas:
Thanks for the review! I agree with you guys! I just removed the `@classproperty` descriptor for a class attribute that is set in the metaclass.

pydantic/main.py
Jocelyn-Gas: I think it should be solved since I removed this class. Thanks

tests/test_computed_fields.py
Jocelyn-Gas: Is it supposed to be settable ?

Viicos:
> I think Viicos' comment was more for when the classproperty implementation was being used.

Indeed, this attribute shouldn't be set and with `classproperty` I wanted to make sure it was the case.

sydney-runkle (approved):
@Jocelyn-Gas,

This looks great, thanks a ton for your contribution.

Apologies for the delayed review - things got a bit backlogged with the holidays.

Jocelyn-Gas:
> @Jocelyn-Gas,
> 
> This looks great, thanks a ton for your contribution.
> 
> Apologies for the delayed review - things got a bit backlogged with the holidays.

No worries for the delay, holidays first! 

I'm glad I could help, this was a first on this project! 😁

PR #8442, opened 2023-12-27
Update docstrings for secret types

dmontagu:
Not sure if we should also add a note in the usage docs, etc.

@samuelcolvin might make sense to add a note somewhere justifying the behavior for empty string (namely, in HTML forms it can be useful to see the empty string when debugging).

It might be worth noting more explicitly somewhere that this type should not be used if you need to protect the use of the empty string as a value, and/or that you should ensure elsewhere that it is not used for user input / etc.

### DIFF
--- a/pydantic/types.py
+++ b/pydantic/types.py
@@ -1552,7 +1552,7 @@ class SecretStr(_SecretField[str]):
     print(user.password.get_secret_value())
     #> password1
     print((SecretStr('password'), SecretStr('')))
-    #> ('**********', '')
+    #> (SecretStr('**********'), SecretStr(''))
     ```
     """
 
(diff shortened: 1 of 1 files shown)


dmontagu:
I marked as relnotes-ignore but it might be nice to have a relnotes-docs or similar.

samuelcolvin (approved):
Lgtm

pydantic/types.py
```diff
@@ -1550,6 +1551,8 @@ class User(BaseModel):
     #> username='scolvin' password=SecretStr('**********')
     print(user.password.get_secret_value())
     #> password1
+    print(SecretStr('password'), SecretStr(''))
+    #> '**********' ''
```
Kludex, line 1555:
```suggestion
    print((SecretStr('password'), SecretStr('')))
    #> ('**********', '')
```

dmontagu: It shows the repr when you make it a tuple, but anyway I still made this change

Issue #8320, opened 2023-12-07
`TypeAliasType` ref `KeyError`


samuelcolvin:
See https://github.com/samuelcolvin/FastUI/compare/pydantic-ref-bug?expand=1

To reproduce:
* check that out
* install deps
* run `python -c 'import fastui'`

Exception:

```traceback
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/samuel/code/fastui/src/python-fastui/fastui/__init__.py", line 11, in <module>
    class FastUI(pydantic.RootModel):
  File "/Users/samuel/code/fastui/env311/lib/python3.11/site-packages/pydantic/_internal/_model_construction.py", line 182, in __new__
    complete_model_class(
  File "/Users/samuel/code/fastui/env311/lib/python3.11/site-packages/pydantic/_internal/_model_construction.py", line 501, in complete_model_class
    schema = gen_schema.clean_schema(schema)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samuel/code/fastui/env311/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py", line 405, in clean_schema
    schema = _discriminated_union.apply_discriminators(schema)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samuel/code/fastui/env311/lib/python3.11/site-packages/pydantic/_internal/_discriminated_union.py", line 60, in apply_discriminators
    return simplify_schema_references(_core_utils.walk_core_schema(schema, inner))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samuel/code/fastui/env311/lib/python3.11/site-packages/pydantic/_internal/_core_utils.py", line 480, in simplify_schema_references
    schema = walk_core_schema(schema, count_refs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samuel/code/fastui/env311/lib/python3.11/site-packages/pydantic/_internal/_core_utils.py", line 431, in walk_core_schema
    return f(schema.copy(), _dispatch)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #8526, opened 2024-01-09
Fix schema-building bug with TypeAliasType for types with refs

dmontagu:
Fixes an issue with building schemas that was affecting FastUI.

In particular, closes https://github.com/pydantic/pydantic/issues/8320.

There are definitely still many misbehaviors related to complex usage of `TypeAliasType` that I've discovered while playing with it, but given how small this is and that it unblocks FastUI, I think we should merge this and think about fixing those once we have minimal reproductions.


dmontagu:
Also, note that this will fail CI until we update pydantic-core to 2.15.0.

sydney-runkle (approved):
Looks great! Thanks for digging into this @dmontagu 🚀

PR #8636, opened 2024-01-25
Use pydantic-settings 2.1 docs

hramezani:
fixes https://github.com/pydantic/pydantic/issues/8618


sydney-runkle (approved):
Thanks, @hramezani. Looks great.

PR #8651, opened 2024-01-27
Fix type-safety of attribute access in `BaseModel`

### LINKED ISSUE #8643: v2 regression: `__setattr__` hides bugs from type checkers
### Initial Checks

### Description

Apparently the type safety of Pydantic models has deteriorated significantly from v1 to v2 due to the exposure of a `__setattr__` method that accepts `Any`. This opens the door for a lot of serious bugs, because the type checkers fail to catch fundamental bugs like the following.

### Example Code

```py

class Foo(BaseModel):
    some_field: int

def patch_object(foo: Foo) -> None:
    # Ohoh, we got the field name wrong...
    foo.some_feeld = 42

foo = Foo(some_field=1)
patch_object(foo)
assert foo.some_field == 42
```

With Pydantic V1 this obvious bug was detected by both mypy and pyright. Mypy would for instance produce a very clear and helpful type error here:

```
error: "Foo" has no attribute "some_feeld"; maybe "some_field"?  [attr-defined]
```

With Pydantic V2, the type checker completely misses the bug.

It should be quite straightforward to make the field access of `BaseModel` type safe again, by simply not exposing the `__setattr__` method towards type checkers. Basically just this pattern ([example on mypy playground](https://mypy-play.net/?mypy=latest&python=3.12&gist=97dbc3dc8d51e14e3067d74ac32e6948)):

```py
from typing import Any, TYPE_CHECKING

class BaseModel:
    if not TYPE_CHECKING:
        def __setattr__(self, name: str, value: Any) -> None:
            ...

class Foo(BaseModel):
    some_field: int

def patch_object(foo: Foo) -> None:
    # Now the type checker will see the bug...
    foo.some_feeld = 42
```

Of course, in a completely strict sense this would be a "breaking" change, but since it looks rather like an bug/accident compared to V1, it would be really nice if this could be made type safe again. We just ran into such a fundamental bug resulting from this regression. This behavior is highly bug prone and also totally unexpected: Any type annotated dataclass-like type (including Pydantic V1 models) is naturally expected to respect the field annotations. After all, getting a field name wrong is one of the most basic of all bugs, and a primary reason for people to use type annotations and type checkers. If the code does not work at runtime (it crashes!), it should be flagged at type checking time.

I tried a range of pydantic versions including the upcoming 2.6.0b1. It looks like this used to work in 1.X but has started to fail in 2.X.

```Text
             pydantic version: 2.6.0b1
        pydantic-core version: 2.16.1
                 install path: /home/ekf2abt/git/scripts/pi_23/debug_pydantic_v2/venv/lib/python3.8/site-packages/pydantic
               python version: 3.8.10 (default, Nov 22 2023, 10:22:35)  [GCC 9.4.0]
                     platform: Linux-5.4.0-166-generic-x86_64-with-glibc2.29
             related packages: typing_extensions-4.9.0 mypy-1.8.0
                       commit: unknown
```

bluenote10:
## Change Summary

This PR fixes the type-safety of attribute access of `BaseModel`.

I noticed that in addition to `__setattr__`, `__delattr__` should get the same treatment. Otherwise the same issue would arise with `del model.non_existing_field` (albeit deleting fields is probably a bit strange to begin with).

## Related issue number

fix #8643

## Checklist

* [x] Tests pass on CI

-------------------

Regarding test: Are you familiar with the notion of type tests via `# type: ignore`?

I wanted to add a such a type test to make sure there cannot be a regression again. The idea is to leverage mypy's `warn_unused_ignores` feature, which you have already enabled as far as I can see. The nice thing about the feature is that `# type: ignore`s are only allowed _if and only if_ a particular line does not type check. This enables library authors to verify their interface in terms of typing guarantees. What I wanted to add is something like this to the `test_main.py`:

```py
if typing.TYPE_CHECKING:
    # The following test(s) are purely type-checking-time tests; no runtime execution needed.

    def test_typesafety_of_attributes():

        class Model(BaseModel):
            foo: int

        model = Model(foo=42)

        model.non_existing_field  # type: ignore[attr-defined]
        model.non_existing_field = ...  # type: ignore[attr-defined]
        del model.non_existing_field  # type: ignore[attr-defined]
```

Checking this snippet with mypy would fail on the `main` branch, because the last two lines actually don't produce type errors. After the change in `main.py` they do behave correctly and the type ignores become mandatory as one would expect. Therefore the snippet could guarantee that `__getattr__`/`__setattr__`/`__delattr__` don't accidentally leak towards the type checker again.

Unfortunately it looks like the tests are not typed checked though, i.e., the tests does something only when I do `mypy tests/test_main.py` explicitly, but running just `make` seems to ignore all type errors in the test.

Could you advice on how to proceed with the test?

Wouldn't it make sense to type check the tests as well? Currently they seem to have quite a lot of type errors, so this would be a bit of work, but could add some value. In general it is always nice to see that a test that does something with `pytest.raises` actually does require some `# type: ignore`, because it indicates that the runtime behavior and the type-checker time behavior are in line! So by forcing oneself to place these `# type: ignore` on these "expected errors", it gets pretty clear how well the typings match with actual behavior. What do you think?

Selected Reviewer: @adriangb

### DIFF
--- a/pydantic/main.py
+++ b/pydantic/main.py
@@ -733,6 +733,7 @@ class BaseModel(metaclass=_model_construction.ModelMetaclass):
 
     if not typing.TYPE_CHECKING:
         # We put `__getattr__` in a non-TYPE_CHECKING block because otherwise, mypy allows arbitrary attribute access
+        # The same goes for __setattr__ and __delattr__, see: https://github.com/pydantic/pydantic/issues/8643
 
         def __getattr__(self, item: str) -> Any:
             private_attributes = object.__getattribute__(self, '__private_attributes__')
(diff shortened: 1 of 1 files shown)


bluenote10:
please review

adriangb:
Personally, I like type-checking tests but that would be a big change on this repo that should be discussed separately.

Maybe we can add this to the mypy or pyright tests?
https://github.com/pydantic/pydantic/tree/main/tests/pyright
https://github.com/pydantic/pydantic/tree/main/tests/mypy

It's a bit hard to tell from the diff but I think all you did was indent these functions into a `TYPE_IGNORE` right? There were no changes to the function bodies?

bluenote10:
> It's a bit hard to tell from the diff but I think all you did was indent these functions into a TYPE_IGNORE right? There were no changes to the function bodies?

Yes, no other changes, just the indentation of the two methods. Unfortunately, GitHub's diff algorithm is a bit suboptimal, and makes the diff look worse than it is (it associates some `else:` statements with different `else:` statements because they happen to have the same indentation after the indentation change). To verify that it is purely a whitespace change, you can enable it here in the review:

image

> Maybe we can add this to the mypy or pyright tests?

I will look into that to see where it would fit.

bluenote10:
> Maybe we can add this to the mypy or pyright tests?

I had a closer look now, but it feels like this is not quite the right place for it, no? On first glance these seem more like "meta tests" that are responsible for the execution of the type checkers, managing versions and handling results. It is not quite the right place to actually type check certain aspects of API interface like `BaseModel`. That would be better suited to be placed alongside the actual (runtime) tests of these API interfaces, which are the normal test modules like `test_main.py` in the case of `BaseModel`.

Also, the modules in `tests/mypy`/`tests/pyright` do not seem to be type checked, so just moving the snippet there wouldn't work out-of-the-box either I guess.

adriangb:
@samuelcolvin how would you feel about enabling type-checking for tests and using includes/excludes to either (1) exclude all current files or (2) exclude all files and whitelist only some files (I'm not sure this is possible).

bluenote10:
> (2) exclude all files and whitelist only some files (I'm not sure this is possible).

This should be possible in general. At least I've successfully used something like this in mypy.ini's (pseudo-code, didn't look up the exact syntax):

```ini
# blacklist tests
[mypy-tests.*]
ignore_errors = True

# whitelist specific modules/packages
[mypy-tests.test_some_module.py]
ignore_errors = False
[mypy-tests.core.*]
ignore_errors = False

```

I'm not quite sure though if mypy's filtering requires proper "package" semantics to work (which may require placing `__init__.py` inside the `tests` folder structure, which is a subtle topic).

samuelcolvin:
I really don't want to add type checking to tests.

We have mypy tests, let's use them and if they're not sufficient, improve them.

adriangb:
Okay then let’s add relnotes and merge this without adding type checking tests

bluenote10:
> Okay then let’s add relnotes and merge this without adding type checking tests

@adriangb I couldn't figure out what exactly I have to do. If it refers to the GitHub labels I don't seem to have the rights to add one. Or should I write something in a change log somewhere?

adriangb (approved):

PR #8666, opened 2024-01-29
Fix bug with mypy plugin and `no_strict_optional = True`

dmontagu:
Fixes https://github.com/pydantic/pydantic/issues/8665

The issue was originally introduced in #7411 (which, if you look at the body, I noted that I was concerned that there was a deeper issue not being addressed). I still have those concerns, but at the very least I think with the changes in this PR, the  `_DeferAnalysis` exception should always get handled, even if it does something that causes the mypy plugin to otherwise misbehave. (That exception was the source of the issue reported in #8665.)

There were some code paths that could raise a `_DeferAnalysis` exception but which were not included in the `try: except:`.

I was able to reproduce the error in #8665 with the code sample provided, and this change fixed it. I also checked and I am pretty confident that there are no code paths that can raise that error now that aren't covered by the updated `try: except:`.

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -475,9 +475,9 @@ class PydanticModelTransformer:
                 return False
 
         is_settings = any(base.fullname == BASESETTINGS_FULLNAME for base in info.mro[:-1])
-        self.add_initializer(fields, config, is_settings, is_root_model)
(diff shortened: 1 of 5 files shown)


sydney-runkle (approved):
Thanks for the fix, @dmontagu!

PR #8923, opened 2024-02-29
Make TypeAdapter's typing compatible with special forms

adriangb:
See https://github.com/python/typing/pull/1618


sydney-runkle (approved):
Nice, thanks for adding the comment as well 👍

PR #9039, opened 2024-03-18
Remove unnecessary warning for config in plugin

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/9021

### DIFF
--- a/tests/mypy/test_mypy.py
+++ b/tests/mypy/test_mypy.py
@@ -87,7 +87,6 @@ cases = (
             'plugin_fail.py',
             'plugin_success_baseConfig.py',
             'plugin_fail_baseConfig.py',
-            'plugin_optional_inheritance.py',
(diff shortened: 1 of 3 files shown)


tests/mypy/test_mypy.py
```diff
@@ -105,6 +104,8 @@ def build(self) -> List[Union[Tuple[str, str], Any]]:
     # One-off cases
     + [
         ('mypy-plugin.ini', 'custom_constructor.py'),
+        ('mypy-plugin.ini', 'config_conditional_extra.py'),
+        ('mypy-plugin.ini', 'plugin_optional_inheritance.py'),
```
dmontagu, line 108: I just moved this down here to reduce the number of mypy tests by 1, since I added this yesterday and forgot I could add it down here instead of above to reduce the number of runs.

PR #7193, opened 2023-08-21
Make shadowing attributes a warning instead of an error

adriangb:
Closes https://github.com/pydantic/pydantic/issues/7108

Selected Reviewer: @samuelcolvin

PR #9105, opened 2024-03-26
Improve warning message when a field name shadows a field in a parent model

chan-vince:
## Change Summary

Added in #7193, a helpful warning message is displayed when a subclass contains a field with the same name as the parent class.

If you have many models that inherit from a parent class, it would be helpful to be able to see in the warning message which child model contains the duplicate field name - not just the parent class.

So I've simply tweaked the warning message text to allow the user to easily see which model they should inspect.

Before:

```
UserWarning: Field name "foo" shadows an attribute in parent "ParentModel";
```

After:

```
UserWarning: Field name "foo" in "ChildModel" shadows an attribute in parent "ParentModel"
```

I think this is clearer for cases where you have tens or hundreds of ChildModels.

I also removed the semi-colon because I'm not sure why it's there, but happy to put that back, I may have missed a convention or other warning message or code style rule.

## Related issue number

None

## Checklist

* [x] Tests pass on CI

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/_internal/_fields.py
+++ b/pydantic/_internal/_fields.py
@@ -193,7 +193,8 @@ def collect_model_fields(  # noqa: C901
                     # on the class instance.
                     continue
                 warnings.warn(
-                    f'Field name "{ann_name}" shadows an attribute in parent "{base.__qualname__}"; ',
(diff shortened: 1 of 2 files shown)


chan-vince:
please review

sydney-runkle (approved):
Great work, thank you!

PR #9135, opened 2024-03-29
Ensure json schema generator handles Literal null types

### LINKED ISSUE #8905: `Literal` annotation does not render type to json-schema

bruno-f-cruz:
## Change Summary

This PR aims to complete #8944 and #8905 by also handling null types when generating a json-schema from a pydantic model.

## Related issue number

fix #8905

## Checklist

* [ ] Tests pass on CI

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -751,6 +751,8 @@ class GenerateJsonSchema:
             result['type'] = 'boolean'
         elif types == {list}:
             result['type'] = 'array'
+        elif types == {type(None)}:
+            result['type'] = 'null'
         return result
 
     def enum_schema(self, schema: core_schema.EnumSchema) -> JsonSchemaValue:


bruno-f-cruz:
@sydney-runkle Not sure why those 2 tests are failing, can you take a look? thanks!

PR #9169, opened 2024-04-04
Update mkdocs_material

dmontagu:
Update the version of mkdocs-material, and enable various features.

I think some things, such as the instant previews, won't work on the preview site because they rely on the site_url being right.


sydney-runkle (approved):
Great :)

PR #9291, opened 2024-04-19
Fixes loss when serializing/deserializing large decimals #9180

### LINKED ISSUE #9180: Decoding Decimal from JSON number is lossy
### Initial Checks

### Description

When decoding a JSON number into a Python Decimal the precision seems to be limited. After a certain number of digits the value is cut off. This is something I would expect for a fixed precision float, but not for an arbitrary precision Decimal. This only happens with numbers that contain a decimal point. I assume what happens is that the number is internally converted to a float and then a Decimal. 

As a user, this lossy internal conversion is an unexpected and unwelcome surprise. Both the initial JSON string and the final Decimal can contain the full precision of the value without loss, so I would have expected the conversion to be lossless as well. 

PS: I'm not sure if this is a bug per se because I could not find any documentation that explicitly states what the expected behavior is. However, based on what was written in the documentation it certainly was unexpected to me.

### Example Code

```Python
from decimal import Decimal

class Test(BaseModel):
    value: Decimal

print(Test.model_validate_json('{"value": 1.234567890123456789012345678901234567890}'))
# Expected output: value=Decimal('1.234567890123456789012345678901234567890')
# Actual output: value=Decimal('1.2345678901234567')

print(Test.model_validate_json('{"value": 12345678901234567890123456789012345678.9}'))
# Expected output: value=Decimal('12345678901234567890123456789012345678.9')
# Actual output: value=Decimal('12345678901234568000000000000000000000')
```

```Text
             pydantic version: 2.6.4
        pydantic-core version: 2.16.3
                 install path: <redacted>
             related packages: fastapi-0.109.2 mypy-1.3.0 typing_extensions-4.8.0
                       commit: unknown
```

ybressler:
## Change Summary
Serializing/deserializing large decimal values is lossy. These changes enable expected behavior - no loss in information with large decimals.

### Help Needed
I was able to solve the deserialization part. But I'm having trouble identifying how/where to control for model export to json _(serialization is still lossy)_. 

## Related issue number
fix #9180

## Checklist

* [X ] The pull request title is a good summary of the changes - it will be used in the changelog
* [ ] Tests pass on CI

### DIFF
--- a/pydantic/_internal/_std_types_schema.py
+++ b/pydantic/_internal/_std_types_schema.py
@@ -192,6 +192,7 @@ class DecimalValidator:
                 # but if it's not make it a string
                 # we don't use JSON -> float because parsing to any float will cause
                 # loss of precision
+                core_schema.float_schema(),
                 core_schema.int_schema(strict=True),
                 core_schema.str_schema(strict=True, strip_whitespace=True),
                 core_schema.no_info_plain_validator_function(str),
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -248,6 +248,47 @@ def test_con_decimal_encode() -> None:
     assert Obj.model_validate_json(json_data) == Obj(id=1)
 
 
+@pytest.mark.parametrize(
+    ('json_data', 'expected'),
+    [
+        pytest.param(
+            '{"value": 1.234567890123456789012345678901234567890}',
+            Decimal(1.234567890123456789012345678901234567890),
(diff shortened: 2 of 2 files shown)


ybressler:
**Q:** `we don't use JSON -> float because parsing to any float will cause loss of precision`<br>
This doesn't seem right.

ybressler:
Closing, made a new branch up to date from the last release.

PR #9296, opened 2024-04-20
Fix bullets in Strict Mode docs

swelborn:
Current StrictMode documentation does not show this as a bulleted list. This should fix it.

image

Needed a blank line...


hramezani (approved):

hramezani:
ُThanks @swelborn

PR #9298, opened 2024-04-21
Implement deprecated field in json schema

NeevCohen:
## Change Summary

Reflect model depreciation in json schema

## Related issue number

Related to #8922

## Checklist

* [x] Tests pass on CI

Selected Reviewer: @Kludex

### DIFF
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -1445,6 +1445,9 @@ class GenerateJsonSchema:
                 f"model_config['json_schema_extra']={json_schema_extra} should be a dict, callable, or None"
             )
 
+        if hasattr(cls, '__deprecated__'):
+            json_schema['deprecated'] = True
+
         return json_schema
 
     def resolve_schema_to_update(self, json_schema: JsonSchemaValue) -> JsonSchemaValue:
--- a/pydantic/json_schema.py
+++ b/pydantic/json_schema.py
@@ -1445,10 +1445,14 @@ class GenerateJsonSchema:
                 f"model_config['json_schema_extra']={json_schema_extra} should be a dict, callable, or None"
             )
 
-        if hasattr(cls, '__deprecated__'):
+        if self._should_mark_class_as_deprecated(cls):
             json_schema['deprecated'] = True
 
         return json_schema
+
+    @staticmethod
+    def _should_mark_class_as_deprecated(cls) -> bool:
+        return hasattr(cls, '__deprecated__') and getattr(cls, '__deprecated__')
 
     def resolve_schema_to_update(self, json_schema: JsonSchemaValue) -> JsonSchemaValue:
         """Resolve a JsonSchemaValue to the non-ref schema if it is a $ref schema.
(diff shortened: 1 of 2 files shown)


NeevCohen:
Please review

pydantic/json_schema.py
```diff
@@ -1445,6 +1445,9 @@ def _update_class_schema(
                 f"model_config['json_schema_extra']={json_schema_extra} should be a dict, callable, or None"
             )
 
+        if hasattr(cls, '__deprecated__'):
+            json_schema['deprecated'] = True
+
```
sydney-runkle, line 1450: Just one thought here - we probably want to check if `cls.__deprecated__` is `True` or a `str`, right? You can imagine a case where someone sets `__deprecated__` to `False` on a model and this would lead to unexpected behavior. We maybe also want to check for `None`.

sydney-runkle (approved):
@NeevCohen,

Nice work, as always! Just one minor change request, but looks great otherwise!

sydney-runkle (changes requested):
Oops, meant to `'request changes'`

NeevCohen:
@sydney-runkle I added a check for a truthy value in the `__deprecated__` attribute

pydantic/json_schema.py
```diff
@@ -1445,8 +1445,15 @@ def _update_class_schema(
...
             )
 
+        if self._should_mark_class_as_deprecated(cls):
+            json_schema['deprecated'] = True
+
         return json_schema
 
+    @staticmethod
+    def _should_mark_class_as_deprecated(cls) -> bool:
+        return hasattr(cls, '__deprecated__') and getattr(cls, '__deprecated__')
```
PrettyWood, line 1455: getattr is enough but you can make it more explicit

NeevCohen: @PrettyWood Sure. Do you think `bool(getattr(cls, '__deprecated__'))` is fine?

sydney-runkle: LGTM once we decide on the syntax for this line!

PrettyWood: I just made a remark regarding the code with `hasattr` + `getattr`.
I now took a bit of time to understand the context and reading https://peps.python.org/pep-0702/ it seems you only care about the presence of the field
I can decide to write

```
@deprecated("")
class A: ...
```

and I'll have `A.__deprecated__ = ""`, which will be falsy and still valid.

So in the end I guess the initial approach with just `return hasattr(cls, '__deprecated__')` is the best

PrettyWood: Screenshot 2024-05-15 at 15 58 46

sydney-runkle: What about a `hasattr` and then a false check? @NeevCohen @PrettyWood

PrettyWood: +1. As I said earlier "So in the end I guess the initial approach with just return hasattr(cls, '__deprecated__') is the best"

sydney-runkle (approved):
Looks great, thanks @NeevCohen!

Issue #9221, opened 2024-04-12
Use of __package__ in pydantic/__init__.py fails if imported from zip


johanandersson-tietoevrycom:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

If pydantic folder is placed in zip, then when doing for example 'from pydantic import Field' and \_\_init\_\_.py is run to resolve to 'pydantic.fields' this fails because \_\_package\_\_ is not set. As \_\_package\_\_ is discouraged in docs: 
https://docs.python.org/3/reference/import.html, it would be nice if \_\_init\_\_.py used '\_\_spec\_\_.parent' instead as this then works in zip files too. At least do:
    if not \_\_package\_\_:
       \_\_package\_\_ = \_\_spec\_\_.parent
at some point before using \_\_package\_\_

### Python, Pydantic & OS Version

```Text
Tested in 2.5.3 -> 2.7.0
            pydantic version: 2.7.0
        pydantic-core version: 2.18.1
          pydantic-core build: profile=release pgo=true
                 install path: C:\Python311\Lib\site-packages\pydantic
               python version: 3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)]
                     platform: Windows-10-10.0.22631-SP0
             related packages: typing_extensions-4.11.0
                       commit: unknown
```

samuelcolvin:
I'm a bit lost, please can you format your issue using markdown so I can read the code.

johanandersson-tietoevrycom:
Hi,

I am unsure of what you want me todo. The issue was really only about following recommendation in the python docs and not use the identifier __package__ in your code.

We run Python embedded and prepackage all modules needed in a zip-file. We can not currently put pydantic in such a zip-file.

Regards,
Johan

Mvh
Johan Andersson
________________________________
Från: Samuel Colvin ***@***.***>
Skickat: Friday, April 12, 2024 2:57:20 PM
Till: pydantic/pydantic ***@***.***>
Kopia: Johan Andersson ***@***.***>; Author ***@***.***>
Ämne: Re: [pydantic/pydantic] Use of __package__ in pydantic/__init__.py fails if imported from zip (Issue #9221)


I'm a bit lost, please can you format your issue using markdown so I can read the code.

hramezani:
Thanks @johanandersson-tietoevrycom for reporting this issue.

[...]

PR #9331, opened 2024-04-26
Replace `__spec__.parent` with `__package__`

hramezani:
Fixes https://github.com/pydantic/pydantic/issues/9221

Also, it is [recommended](https://docs.python.org/3/reference/import.html#package__) to use `__spec__` instead of `__package__`


sydney-runkle (approved):
Thanks @hramezani,

This looks great!

PR #9353, opened 2024-04-30
Fix typing of `_frame_depth`

Viicos:
https://github.com/pydantic/pydantic/pull/8939 added a `_frame_depth` decorator that would not retain signature of the decorated methods


sydney-runkle (approved):
@Viicos,

This looks good -- approving and merging. Thanks for the awesome work - definitely a complex level of typing knowledge on display here :).

Linking as a reference for others: https://peps.python.org/pep-0612/.

PR #9421, opened 2024-05-09
Fix labeler permission problem

hramezani:
It is failing for PR from a fork: https://github.com/pydantic/pydantic/actions/runs/8992867651/job/24766726818?pr=9410

This is the fix based on this doc: https://github.com/actions/labeler?tab=readme-ov-file#permissions

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -2,6 +2,7 @@ name: release notes
 
 on:
   pull_request_target:
+    types: [opened]
 
 jobs:
   auto-labeler:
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -1,8 +1,7 @@
 name: release notes
 
 on:
-  pull_request:
-    types: [opened]
+  pull_request_target:
```
davidhewitt, line 4: I think we still want to filter just to opened events.

```suggestion
  pull_request_target:
    types: [opened]
```

davidhewitt (approved):

Issue #9461, opened 2024-05-20
RootModel.model_construct() typing


ornariece:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

The signature of the `model_construct` method of `RootModel` does not match the signature in the supertype `BaseModel`, causing mypy to disfunction when type checking.
Using ``my_root_model.model_construct('smth')`` results in 2 errors:
```python
error: Missing named argument "root" for "model_construct" of "MyRootModel"  [call-arg]
error: Argument 1 to "model_construct" of "MyRootModel" has incompatible type "str"; expected "Optional[Set[str]]"  [arg-type]
```
because mypy is trying to match the call with the signature of the supertype method.

I see no easy fix, because there is no way of forcing mypy to type check the call against the signature of the subtype method. And, after all, mypy is right? The signature of a subtype should match that of the supertype. Perhaps this is an indication that ``RootModel`` should **not** inherit from ``BaseModel``? While I see why you implemented it that way, I think there is a case to be made against it. What are your thoughts on this?
In any case, the current type checking is very annoying (especially with a lot of tests) because one has to manually ignore those 2 errors each time `Root.model_construct` is called.

### Python, Pydantic & OS Version

```Text
pydantic version: 2.7.1
        pydantic-core version: 2.18.2
          pydantic-core build: profile=release pgo=true
                 install path: -
               python version: 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)]
                     platform: Windows-10-10.0.19045-SP0
             related packages: mypy-1.2.0+dev.456dcbdd6a2f9b36249e0e7d31e521f5678a8235 typing_extensions-4.11.0
                       commit: unknown
```

Viicos:
I'm not able to reproduce, could you please throw a `reveal_type(my_root_model)` in your snippet to see the actual type?

ornariece:
```python
import pydantic


class MyRootModel(pydantic.RootModel[str]):
    pass


if __name__ == '__main__':
    reveal_type(MyRootModel)
    reveal_type(MyRootModel.model_construct('yo'))
```
results in
```python
note: Revealed type is "def (root: builtins.str) -> scratch_11.MyRootModel"
error: Missing named argument "root" for "model_construct" of "MyRootModel"  [call-arg]
note: Revealed type is "scratch_11.MyRootModel"
[...]

PR #9480, opened 2024-05-21
Do not alter `RootModel.model_construct` signature in the mypy plugin

Viicos:
Fixes https://github.com/pydantic/pydantic/issues/9461.

No need to tweak the signature as no `**values` argument is present


sydney-runkle (approved):
This looks like a good fix, thanks!

PR #9587, opened 2024-06-06
Support PEP 746

adriangb:
https://discuss.python.org/t/pep-746-typedmetadata-for-type-checking-of-pep-593-annotated/53834


sydney-runkle (approved):
Woohoo!!

Issue #9580, opened 2024-06-06
mypy plugin reports unsafe variance error on frozen models


DetachHead:
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

when a model is frozen, using a covariant type parameter on a field is safe, since it cannot be mutated. however the mypy plugin treats it the same as if it was not frozen

### Example Code

```Python
from typing import Generic, TypeVar

from pydantic import BaseModel

out_T = TypeVar("out_T", covariant=True)


class Foo(BaseModel, Generic[out_T]):
    model_config = {"frozen": True}
    value: out_T # error: Cannot use a covariant type variable as a parameter  [misc]
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.7.3
        pydantic-core version: 2.18.4
          pydantic-core build: profile=release pgo=true
                 install path: C:\Users\user\project\.venv\Lib\site-packages\pydantic
               python version: 3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)]
                     platform: Windows-10-10.0.19045-SP0
             related packages: typing_extensions-4.12.1
                       commit: unknown
```

dmontagu:
I suspect that the issue is that we are using the annotation on the field as a type hint in the signature of other methods, in particular `model_construct`. I'm thinking probably we need to replace covariant typevars with regular typevars when creating that signature.

Can you share the version of mypy you are using to ensure I can reproduce the issue?

dmontagu:
I was able to reproduce this, it looks like it's actually in the `__init__`, to my surprise.

I think I have a fix, but I'm a bit concerned about why we are running into this issue when the dataclass plugin does not. I'll open my PR with a fix but I think if we run into more issues like this we may want to dive into why the dataclass plugin behaves differently (I don't expect that to be a fun thing to look into though).

PR #9606, opened 2024-06-07
Fix bug with mypy plugin's handling of covariant typevar fields

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/9580 and https://github.com/pydantic/pydantic/issues/9581.

As noted elsewhere (including in comments in the code affected by this PR), I am not confident this is the best way to fix this bug, and I think it's suspicious that dataclasses doesn't have this issue since the pydantic mypy plugin is based on the dataclasses plugin. But I think it makes sense to just merge this and look into it more carefully/thoroughly if we continue to run into issues with typevars and/or typechecking of generic basemodels.

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -14,6 +14,7 @@ from mypy.nodes import (
     ARG_OPT,
     ARG_POS,
     ARG_STAR2,
+    INVARIANT,
     MDEF,
     Argument,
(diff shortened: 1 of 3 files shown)


sydney-runkle (approved):
@dmontagu,

Thanks for such a quick turnaround on this. Great work!

PR #9659, opened 2024-06-14
Allow a field specifier on __pydantic_extra__

dmontagu:
Closes https://github.com/pydantic/pydantic/issues/9187


sydney-runkle (approved):
Awesome, thanks @dmontagu !

PR #9680, opened 2024-06-17
Fix TypeAdapter typing

adriangb:


tests/pyright/pyright_example.py
```diff
@@ -56,3 +57,16 @@ def area(self) -> float:
...
+ta1 = TypeAdapter(int)
+assert ta1.validate_json('123') + 1 == 124
+# But currently cannot infer more complex "special forms" and defaults to TypeAdapter[Any]
+ta2 = TypeAdapter(Annotated[int, Field(gt=0)])
+assert ta2.validate_python(999) + 1 == 1000
+# this won't get type checked because ta2.validate_python() returns Any!
+assert ta2.validate_python(999) + 'a' == ':('
+# If you want these to be typed, you can use a type hint as follows:
+ta3: TypeAdapter[int] = TypeAdapter(Annotated[int, Field(gt=0)])
+assert ta3.validate_python(999) + 1 == 1000
```
adriangb, line 72: If running this with strict mode (added above) before this PR you get:
```
pyright_example.py:66:1 - error: Type of "ta2" is partially unknown
    Type of "ta2" is "TypeAdapter[Unknown]" (reportUnknownVariableType)
```

Viicos: Is it still possible to do `ta3 = TypeAdapter[int](Annotated[int, Field(gt=0)])`?

adriangb: Yeah I don't see why not. I can add a test if you'd like.

Viicos: I believe with the [new spec update](https://typing.readthedocs.io/en/latest/spec/constructors.html#new-method) regarding constructors it doesn't :/ :

> If the class is generic, it is possible for a `__new__` method to override the specialized class type and return a class instance that is specialized with different type arguments.
> ```python
> class MyClass[T]:
>     def __new__(cls, *args, **kwargs) -> "MyClass[list[T]]":
>         ...
>
> assert_type(MyClass[int](), MyClass[list[int]])
> ```

See [pyright playground](https://pyright-play.net/?pyrightVersion=1.1.364&strict=true&code=GYJw9gtgBA%2BjwFcAuCQFM5QJYQA5hCSgEMA7UsJYpLMUgZwChHRIokBPXLUgc2zwEiAQVIcANFFEUqSNABNJAcTSk0ILAGNJAFS5oAasRCSwAN3UAbMMXnMdUALxQ9uQ8YAUAIh1eAlMyalsT09C7CHipqGpoA2joAun4AXIxQ6VAAAuZWNnYZUPJowLAwagDucB5B9JKcbsns%2BvFJUAC0AHzhLY0AdP1pGdkWINa2g%2BlFJXAVVTV1%2Bo2iHH7tXTrCscsJfQPMSACMTuEePEgBSABMxxse0pTUCrFnkv29SYxIAMw3m2cJd3IDzk8mepCQr36HyAA)

adriangb: I just tested it. It _kind of_ works: `TypeAdapter[int](...)` returns `TypeAdapter[Any]`. That's unfortunate. I'd hope we had a way to make both work...

adriangb:
Since this is not clearly better I'm going to close it. We'll have to wait until TypeForm.

PR #9685, opened 2024-06-18
setup: remove upper bound from python_requires

### LINKED ISSUE #9663: Should we cap the required_python version in v1?
### Initial Checks

### Description

It was asked why we should cap the version in Pydantic V1 to an upper bound of 3.12: https://github.com/pydantic/pydantic/issues/9622#issuecomment-2165288361

My personal response was:

> Its "explicit support" vs "implied support"
>
> It may work in most usecases in 3.13. It may even pass the full test suite, though I have strong doubts on the cythonized builds working. In the end, just because it works doesn't mean its supported.
>
> It may be nontrivial to add full 3.13 support or maybe they don't want to be on the hook for dealing with another full life cycle of a release because there may be compatibility issues between bugfix versions like we just had with 3.12.4 and they need to focus developer time elsewhere, like pydantic v2 or a new product, instead of on a pydantic 1.10.x thats in maintenance mode.

To add additional detail to the "new version support", the challenges of ensuring that pydantic work on a new version are, at a minimum:
* leveraged private functions in the stdlib haven't changed within python. Private functions are under no contract to maintain compatibility
* dependencies used by pydantic for runtime and testing are also available or buildable for the target version
* For cythonized builds: cython has support for the target version (Cython 0.x will almost certainly not see Python 3.13 support.)
  * Pure python based builds (sdist, wheels) may still be released without the cython requirement

After responding to the original question, however, I decided to dig into this a bit and I think the unfortunate reality is that the Python packaging environment is _not at all friendly to upper bound Python versions_.

For some "light reading", see:
* https://discuss.python.org/t/requires-python-upper-limits/12663
* https://github.com/scipy/scipy/issues/14738#issuecomment-997236377
* https://github.com/numpy/numpy/issues/24810
* https://github.com/scipy/scipy/pull/16553
* https://github.com/python-poetry/poetry/issues/6884
* https://github.com/python-poetry/poetry/issues/6921
* https://github.com/numba/numba/blob/0994f97c33a19d7684471cc98b8c36573762204b/setup.py#L44-L47

So, while our intention may be to cap the upper bounds, puzzle solvers may just **fall back to an older version** and install that since it was released without a cap, which may be undesired. It may also have unintended consequences for dependency lockfiles for certain package managers (like poetry).

Until the semantics are worked out with the packaging group, it may make sense to undo the cap I added in the v1 branch and instead add a build time check for the python version and issue a `RuntimeError` like numba has done. This would allow attempts at using the sdist install on an unsupported version to throw an error, though it may be good to cover __init__ for the pure python wheel too maybe... With the caveat that the copied sources in the v2 branch will need have a fixup via `update_v1.sh` to handle this by dropping the runtime error.

The last thing I want to do is actually cause more headaches for packages that use Pydantic.

### Example Code

_No response_

```Text
pydantic 1.10.16+, any
```

### Affected Components

- [X] [Compatibility between releases](https://docs.pydantic.dev/changelog/)

vfazio:
A cap was recently added to denote the upper bound of python support for pydantic v1.

Further research, however, has determined that `python_requires` was not developed with an upper bound in mind [^0] and may actually cause some dependency solvers to silently fall back to older versions where there is no cap, and thus presumed support for the user's python.

For now, remove the upper bound.

[^0]: https://discuss.python.org/t/requires-python-upper-limits/12663

## Related issue number
fix #9663 

## Checklist

* [x] Tests pass on CI

skip change file check

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -134,7 +134,7 @@ setup(
     license='MIT',
     packages=['pydantic'],
     package_data={'pydantic': ['py.typed']},
-    python_requires='>=3.7, <3.13',
+    python_requires='>=3.7',
     zip_safe=False,  # https://mypy.readthedocs.io/en/latest/installed_packages.html
     install_requires=[
         'typing-extensions>=4.2.0'


sydney-runkle (approved):
Looks great, assuming CI is happy. Thank you!

PR #9779, opened 2024-06-27
Fix documentation formatting

ofek:
image


sydney-runkle (approved):
Great, thanks so much!

PR #9838, opened 2024-07-03
Fix issue with assertion caused by pluggable schema validator

dmontagu:
Fix issue with type adapters present when plugins are installed.

### DIFF
--- a/pydantic/type_adapter.py
+++ b/pydantic/type_adapter.py
@@ -38,7 +38,7 @@ from .json_schema import (
     JsonSchemaMode,
     JsonSchemaValue,
 )
-from .plugin._schema_validator import create_schema_validator
+from .plugin._schema_validator import PluggableSchemaValidator, create_schema_validator
 
 T = TypeVar('T')
 R = TypeVar('R')
@@ -253,7 +253,7 @@ class TypeAdapter(Generic[T]):
             self._module_name = module
 
         self._core_schema: CoreSchema | None = None
-        self._validator: SchemaValidator | None = None
+        self._validator: SchemaValidator | PluggableSchemaValidator | None = None
         self._serializer: SchemaSerializer | None = None
 
         if not self._defer_build():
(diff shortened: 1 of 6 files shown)


pydantic/type_adapter.py
```diff
@@ -311,11 +311,11 @@ def core_schema(self) -> CoreSchema:
...
     @cached_property
     @_frame_depth(2)  # +2 for @cached_property + validator(self)
-    def validator(self) -> SchemaValidator:
+    def validator(self) -> SchemaValidator | PluggableSchemaValidator:
         """The pydantic-core SchemaValidator used to validate instances of the model."""
-        if not isinstance(self._validator, SchemaValidator):
+        if not isinstance(self._validator, (SchemaValidator, PluggableSchemaValidator)):
             self._init_core_attrs(rebuild_mocks=True)  # Do not expose MockValSer from public function
-        assert isinstance(self._validator, SchemaValidator)
+        assert isinstance(self._validator, (SchemaValidator, PluggableSchemaValidator))
```
dmontagu, line 318: This is where the problem is — because we were asserting that it was a SchemaValidator, but it could be a PluggableSchemaValidator. While we could just drop this assert, @adriangb wanted to just stop lying in the `create_schema_validator` return type annotation.

dmontagu: (And I agree with that choice, given I wasn't sure where else there might be problems.)

Ideally we would make it so that a PluggableSchemaValidator was actually a subclass of SchemaValidator, or similar, but anyway this works for now.

tests/test_plugins.py
```diff
@@ -341,7 +341,8 @@ def new_schema_validator(self, schema, schema_type, schema_type_path, schema_kin
 
     plugin = Plugin()
     with install_plugin(plugin):
-        TypeAdapter(List[str])
+        adapter = TypeAdapter(List[str])
+        adapter.validate_python(['a', 'b'])
```
dmontagu, line 345: This raises an error on main, so obviously it's a problem lol

dmontagu:
One other super annoying thing that we should fix soon is that you can't run the plugins tests on their own. I'm not sure where we are affecting state that causes them to pass when you run `pytest tests` but to fail when you run `pytest tests/test_plugins.py` but ... I think that should get fixed.

I'll note that if you change the `_plugins: ... = None` to `_plugins: ... = {}` this goes away and all tests pass. BUT I think we have special logic when `_plugins is None`, so maybe we want to add a test that actually ensures this behavior is respected?

sydney-runkle (approved):
Great, thanks for this quick fix!

PR #9913, opened 2024-07-17
Prevent type error for exotic BaseModel/RootModel inheritance

### LINKED ISSUE #9861: Mypy error: incompatible definition of `model_construct()` in multiple inheritance from `BaseModel` and `RootModel`
### Initial Checks

### Description

The code below (I know it doesn't really make sense on its own terms, but it's a minimal reproduction of a case in which the multiple inheritance has a certain logic) yields the following mypy error on pydantic 2.8.2, but it used to be ok on 2.7.4:

`test.py:7: error: Definition of "model_construct" in base class "Foo" is incompatible with definition in base class "RootModel"  [misc]`

I'm guessing it's an issue with the mypy plugin.

Potentially related:
* https://github.com/pydantic/pydantic/issues/8729
* https://github.com/pydantic/pydantic/issues/9461

### Example Code

```Python

class Foo(BaseModel):
    pass


class Bar(Foo, RootModel):
    pass
```

```Text
pydantic version: 2.8.2
        pydantic-core version: 2.20.1
                 install path: .../venv/lib/python3.8/site-packages/pydantic
               python version: 3.8.19 (default, Apr  6 2024, 17:58:10)  [GCC 11.4.0]
                     platform: Linux-6.5.0-1025-oem-x86_64-with-glibc2.35
             related packages: fastapi-0.111.0 pydantic-settings-2.3.4 mypy-1.10.1 pydantic-extra-types-2.8.2 typing_extensions-4.12.2
                       commit: unknown
```

dmontagu:
Closes #9861

### DIFF
--- a/pydantic/mypy.py
+++ b/pydantic/mypy.py
@@ -345,6 +345,7 @@ class PydanticModelField:
         use_alias: bool,
         api: SemanticAnalyzerPluginInterface,
         force_typevars_invariant: bool,
+        is_root_model_root: bool,
     ) -> Argument:
         """Based on mypy.plugins.dataclasses.DataclassAttribute.to_argument."""
         variable = self.to_var(current_info, api, use_alias, force_typevars_invariant)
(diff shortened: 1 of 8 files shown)


dmontagu:
@Viicos I guess this is going to conflict with #9480 — but it seems that introduced the bug reported in #9861. I'll do a bit of digging..

dmontagu:
Okay I think I may have found a "real" fix now. Still needs to add the test case from #9861 before merging, assuming this passes CI.]

sydney-runkle (approved):
Thanks @dmontagu!

PR #10052, opened 2024-08-05
Enable more lint rules

Viicos:
## Change Summary

Following https://github.com/pydantic/pydantic/pull/10030#discussion_r1702074028, I figured we could enable more lint rules (the `PYI007` would have caught the bad version check).

## Checklist

* [ ] Tests pass on CI


pyproject.toml
```diff
@@ -186,8 +186,20 @@ target-version = 'py38'
 extend-exclude = ['pydantic/v1', 'tests/mypy/']
 
 [tool.ruff.lint]
-extend-select = ['Q', 'RUF100', 'C90', 'UP', 'I', 'D', 'T']
-extend-ignore = ['D105', 'D107', 'D205', 'D415']
```
Viicos, line 190: `extend-ignore` is deprecated and can be replaced by `ignore`.

pyproject.toml
```diff
@@ -186,8 +186,20 @@ target-version = 'py38'
...
+select = [
```
Viicos, line 189: I went for `select` so that we can more clearly see which rules are enabled.

pydantic/_internal/_known_annotated_metadata.py
Viicos, line 1: @sydney-runkle https://github.com/pydantic/pydantic/issues/10036 will relate to this:

I added `#noqas` for `B023` for now but I believe there is a genuine late binding bug here.

sydney-runkle (approved):
LGTM :)

PR #10085, opened 2024-08-08
Update tests for changes to core

BoxyUwU:
In pydantic/pydantic-core#1377 the warning when providing a wrongly typed value for serialization was changed. This PR updates pydantic's tests to check for the new warning 

cc @sydney-runkle


sydney-runkle (changes requested):
Looks great - let's mark these tests with `@pytest.mark.xfail(reason='blah blah blah')` with a reason explaining that these will pass once we use the next `pydantic-core` version :).

sydney-runkle (approved):
Nice work @BoxyUwU, first contribution to `pydantic`!!!

PR #10176, opened 2024-08-19
Remove docs about pydantic plugins

alexmojaki:
The current plugins API has some problems and we're likely to make breaking changes to it in the future. In the meantime it should no longer be documented. Fortunately it was always marked as experimental.


sydney-runkle (approved):
Awesome, thanks @alexmojaki!!

PR #10338, opened 2024-09-06
Simplify unions involving `Any` or `Never` when replacing type variables

Viicos:
## Change Summary

I could have implemented this simplification for all cases (perhaps in `GenerateSchema._apply_single_annotation` or similar) but I don't want to break any unusual scenarios, plus this kind of situation is pretty uncommon in the general case (you're not really going to annotate a field as `Any | <other>`), it only makes sense when parameterizing `T | <other>` with `Any`/`Never`.

This has the added benefit of creating simplified JSON Schemas, as shown in the test. 

## Checklist

* [ ] Tests pass on CI


sydney-runkle (approved):
Awesome! Thanks for the inline comments, makes this super easy to review / understand for future devs.

PR #10369, opened 2024-09-10
Fix nested type statement

kc0506:
## Change Summary

Fix the following:
```py
class A(BaseModel):
    type Int = int  # currently will throws an error because `TypeAliasType` is not in `default_ignored_types()`
    a: Int
```
See: https://docs.python.org/3/reference/simple_stmts.html#type

## Checklist

* [x] Tests pass on CI


sydney-runkle (approved):
Great! Contributing to https://github.com/pydantic/pydantic/issues/9782 :)

